PR #11, opened 2018-05-25
Workaround for VS15.6 Hosted Agents source generation

jeromelaban:
This works around a strange issue where the source generator, in the context of a VSTS VS15.6 hosted agent, would not pick up the Page item group.


jeromelaban:
The 15.7 release is pending on https://github.com/Microsoft/vsts-image-generation/issues/133

PR #23, opened 2018-06-04
Ensure default Style isn't cleared

jeromelaban:
If Style changes from default to a new style, ensure the default isn't cleared. This fixes issues with default Style being absent such as no ControlTemplate, observed in iOS ListView.

### DIFF
--- a/src/Uno.UI/UI/Xaml/FrameworkElement.cs
+++ b/src/Uno.UI/UI/Xaml/FrameworkElement.cs
@@ -268,7 +268,10 @@ namespace Windows.UI.Xaml
 			{
 				ApplyDefaultStyle();
 
-				if (FeatureConfiguration.FrameworkElement.ClearPreviousOnStyleChange)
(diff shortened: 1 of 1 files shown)


src/Uno.UI/UI/Xaml/FrameworkElement.cs
```diff
@@ -268,7 +268,10 @@ protected virtual void OnStyleChanged(Style oldStyle, Style newStyle)
 			{
 				ApplyDefaultStyle();
 
-				if (FeatureConfiguration.FrameworkElement.ClearPreviousOnStyleChange)
+				if (FeatureConfiguration.FrameworkElement.ClearPreviousOnStyleChange && 
+					// Don't clear the default Style, which should always be present.
+					!(bool)(oldStyle == Style.DefaultStyleForType(GetType()))
```
benventive: @Dissolute Why is this cast as a bool?

Dissolute: Good question... I had it as a variable and then pressed the 'inline' tool, and evidently didn't check after

PR #70, opened 2018-06-26
Various UWP Alignments

jeromelaban:
This update provides various compatibility updates for WASM:

- Add support for `Application.Start()` to provide a proper SynchronizationContext for error management
- Fix for ImplicitStyles support in XamlReader
- Add support for the `Pivot` control using the default UWP Xaml style
- Adjust body background color after the splashscreen removal
- Adjust the materialization of Control templates to not be lazy

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/Pivot/Pivot.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Pivot/Pivot.cs
@@ -1,37 +1,222 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Text;
-using Windows.UI.Xaml.Markup;
+using System.Linq;
+using System.Runtime.InteropServices.WindowsRuntime;
+using Uno.Extensions;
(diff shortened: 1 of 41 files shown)


src/Uno.UI/UI/Xaml/Controls/Pivot/Pivot.cs
```diff
@@ -1,37 +1,181 @@
...
+						headerItem.SetBinding(
+							ContentControl.ContentTemplateProperty,
+							new Binding
+							{
+								Path = "HeaderTemplate",
+								RelativeSource = RelativeSource.TemplatedParent
+							}
+						);
+
+						headerItem.PointerPressed += OnItemPointerPressed;
```
Dissolute: Is this going to leak on iOS?

jeromelaban: It will, I've adjusted this already :)

PR #77, opened 2018-06-29
Pivot, AutoSuggestBox, TextBox, XamlReader updates

jeromelaban:
- Added missing `TransformGroup` `ContentProperty`
- Fixed invalid namespace attribution of attached properties in `XamlReader`
- Fixed `BitmapImage.UriSource` udpates not being applied on Wasm
- Add basic implementation of `AutoSuggestBox`
- Fixed focus stealing issues with inactive `PivotItem` content
- Add `ThreadPoolTimer` support
- Fix for iOS popup not appearing
- Fix for Wasm textbox not properly updating while not loaded

### DIFF
--- a/src/Uno.UI/Generated/3.0.0.0/Windows.UI.Xaml.Controls/AutoSuggestBox.cs
+++ b/src/Uno.UI/Generated/3.0.0.0/Windows.UI.Xaml.Controls/AutoSuggestBox.cs
@@ -7,7 +7,7 @@ namespace Windows.UI.Xaml.Controls
 	#endif
 	public  partial class AutoSuggestBox : global::Windows.UI.Xaml.Controls.ItemsControl
 	{
-		#if __ANDROID__ || __IOS__ || NET46 || __WASM__
+#if false
 		[global::Uno.NotImplemented]
 		public  string PlaceholderText
 		{
@@ -20,8 +20,8 @@ namespace Windows.UI.Xaml.Controls
 				this.SetValue(PlaceholderTextProperty, value);
 			}
 		}
-		#endif
-		#if __ANDROID__ || __IOS__ || NET46 || __WASM__
+#endif
+#if false
 		[global::Uno.NotImplemented]
 		public  double MaxSuggestionListHeight
 		{
@@ -34,8 +34,8 @@ namespace Windows.UI.Xaml.Controls
 				this.SetValue(MaxSuggestionListHeightProperty, value);
 			}
 		}
-		#endif
-		#if __ANDROID__ || __IOS__ || NET46 || __WASM__
+#endif
+#if false
 		[global::Uno.NotImplemented]
 		public  bool IsSuggestionListOpen
 		{
@@ -48,8 +48,8 @@ namespace Windows.UI.Xaml.Controls
 				this.SetValue(IsSuggestionListOpenProperty, value);
 			}
 		}
-		#endif
-		#if __ANDROID__ || __IOS__ || NET46 || __WASM__
+#endif
+#if false
 		[global::Uno.NotImplemented]
 		public  object Header
 		{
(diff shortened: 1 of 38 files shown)


src/Uno.UI/UI/Xaml/Controls/Pivot/Pivot.cs
```diff
@@ -190,10 +190,12 @@ private void SynchronizeSelectedItem()
 					{
 						pivotItem.Opacity = 0;
 						pivotItem.IsHitTestVisible = false;
+						pivotItem.Visibility = Visibility.Collapsed;
```
Dissolute: Does it still need to set Opacity and IsHitTestVisible if it's setting visibility to Collapsed?

src/Uno.UI/Generated/3.0.0.0/Windows.UI.Xaml.Controls/AutoSuggestBox.cs
```diff
@@ -7,7 +7,7 @@ namespace Windows.UI.Xaml.Controls
 	#endif
 	public  partial class AutoSuggestBox : global::Windows.UI.Xaml.Controls.ItemsControl
 	{
-		#if __ANDROID__ || __IOS__ || NET46 || __WASM__
+#if false
```
carldebilly: Why not just remove it ? ;-)

src/Uno.UI/UI/Xaml/Controls/AutoSuggestBox/AutoSuggestBox.cs
```diff
@@ -0,0 +1,89 @@
...
+
+			if (IsLoaded)
+			{
+				RegisterEvents();
+			}
+		}
+
+		void RegisterEvents()
+		{
+			_textBox.KeyDown += OnTextBoxKeyDown;
```
carldebilly: No null check if template badly designed.

src/Uno.UI/UI/Xaml/Controls/AutoSuggestBox/AutoSuggestBox.cs
```diff
@@ -0,0 +1,89 @@
...
+			{
+				SubmitSearch();
+			}
+		}
+
+		private static void OnTextChanged(DependencyObject dependencyObject, DependencyPropertyChangedEventArgs args)
+		{
+			if(dependencyObject is AutoSuggestBox tb)
+			{
+				tb.TextChanged.Invoke(tb, new AutoSuggestBoxTextChangedEventArgs() { Reason = AutoSuggestionBoxTextChangeReason.UserInput });
```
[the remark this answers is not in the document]
jeromelaban: Definitely can :)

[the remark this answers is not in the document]
weitzhandler: Should be `tb.TextChanged?.Invoke` then.

src/Uno.UI/UI/Xaml/Controls/Pivot/Pivot.cs
weitzhandler: 1. Same [here](https://github.com/nventive/Uno/pull/77/files#diff-b075b6b5f6cb0c0eb8643ddfd5035f1bR197)
2. I'm not so involved, but setting the opacity and hit-test-visibility here might take away custom configuration (e.g. lower opacity) from the user and will reset it on selection synchronization.

src/Uno.UI/Generated/3.0.0.0/Windows.UI.Xaml.Controls/AutoSuggestBox.cs
jeromelaban: This is for easier change tracking when the source generator is executed.

src/Uno.UI/UI/Xaml/Controls/Pivot/Pivot.cs
jeromelaban: Fair enough, this is an extrapolation of the behavior coming from UWP, where the opacity seems to be changed as well, but this is not required for the current implementation.

Collapsed-only will be enough.

PR #93, opened 2018-07-06
Misc fixes rollup

jeromelaban:
- Fixed iOS `SimpleOrientationSensor` default queue management
- Fixed multiple memory leaks in `ListView`, `ScrollViewer`
- Implemented `CacheLength` for Android `ListViewBase`
- Fixed for `DependencyObject` properties inheritance race condition
- Fix for empty Path reporting an infinite size
- Fix Title  not appearing in CommandBar


Uno.UI.Samples/Content/UITests/ListViewTests/DelayedImagePresenter/DelayedImagePresenter.cs
```diff
@@ -0,0 +1,104 @@
+﻿using System;
```
Dissolute: Did you mean to include the samples?

jeromelaban: Not yet, right :)

PR #102, opened 2018-07-12
Disable artifacts publish for forks

jeromelaban:


src/Uno.UI.Wasm/WasmScripts/Uno.UI.js
```diff
@@ -352,16 +352,15 @@ var Uno;
                 return "ok";
             }
             /**
-                * Issue a browser alert to user
-                * @param message message to display
+                * Sets the browser window title
```
carldebilly: Maybe there's a file missing in your PR, because the change should be in the `.ts` too

jeromelaban: It should not have been there

PR #113, opened 2018-07-24
Misc binding engine alignments

jeromelaban:
- Use Portable symbols for Xamarin debugging stability
- Enable x:Name reference in x:Bind markup. This requires for a failed BindableMetadata lookup to fall through reflection lookup.
- Assume ".Value" binding path on a primitive is equivalent to self, to enable nullable bindings.
- Adjust unit tests logging
- Enables auto "LogicalChild" treatment to allow for DependencyObjectCollection members to be databound
- Enable parent reset for "LogicalChild" assignations

### DIFF
--- a/src/Uno.UI/DataBinding/BindingPropertyHelper.cs
+++ b/src/Uno.UI/DataBinding/BindingPropertyHelper.cs
@@ -318,6 +318,9 @@ namespace Uno.UI.DataBinding
 		/// earlier, the highest match would be returned, but the .NET core 
 		/// implementation found in Mono 4.2+ throws an ambiguous match exception.
 		/// This requires a recursive search using the <see cref="BindingFlags.DeclaredOnly"/> flag.
+		///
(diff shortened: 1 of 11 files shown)


src/Uno.UI/DataBinding/BindingPropertyHelper.cs
```diff
@@ -328,7 +331,7 @@ private static PropertyInfo GetPropertyInfo(Type type, string name)
 					BindingFlags.Instance |
 					BindingFlags.Static |
 					BindingFlags.Public |
-					BindingFlags.DeclaredOnly
+					BindingFlags.NonPublic
```
Dissolute: Does this mean that bindings in general are now going to (incorrectly) bind to non-public properties?

jeromelaban: Indeed, and DeclaredOnly should still be here. I'll add some new tests,

PR #127, opened 2018-08-06
Update addin details for marketplace publication

jeromelaban:


carldebilly:
Should add a link to the marketplace from the README.md

PR #152, opened 2018-08-28
Auto-generate list of views implemented in Uno

davidjohnoliver:


jeromelaban:
Pretty cool feature! Should be generated during the build, so it can be picked up the documentation generator.

Also, it may be interesting to split the file in multiple sections, per namespaces maybe ?

PR #156, opened 2018-08-29
Make attached properties localizable

miguelrochefort:


src/Uno.UI/Services/AndroidResourceNameEncoder.cs
```diff
@@ -0,0 +1,26 @@
...
+		/// <summary>
+		/// Encode a resource name to remove characters that are not supported on Android.
+		/// </summary>
+		/// <param name="key">The original resource name from the UWP Resources.resw file.</param>
+		/// <returns>The encoded resource name for the Android Strings.xml file.</returns>
+		public static string Encode(string key)
+		{
+			// These characters are used by the attached property localization syntax.
+			// Example: "MyUid.[using:Windows.UI.Xaml.Automation]AutomationProperties.Name"
+			return key
```
jeromelaban: We could probably add a fast path for this one, particularly if none of the replace characters can be found.

miguelrochefort: Something like this?

```csharp
private static char[] UnsupportedCharacters = new char[] { '[', ']', ':' };
...
if (key.IndexOfAny(UnsupportedCharacters) == -1)
{
    return key;
}
```

jeromelaban: Yes

PR #160, opened 2018-08-30
Add support for routed events

miguelrochefort:
# TASKS
- [X] Unify pointer event handling (remove partial/native overrides)
- [X] Add support for `AddHandler` and `RemoveHandler`
- [X] Add support for event routing


carldebilly:
Don't worry, still a work in progress. A lot of tests are required for this feature.

doc/articles/routed-events.md
```diff
@@ -0,0 +1,48 @@
+# Uno Support for _Routed Events_
+
+## Implemented Routed Events
+
+| Routed Event          | Android | iOS    | Wasm   |     |
+| --------------------- | ------- | ------ | ------ | --- |
+| _tapped events_
+| `Tapped `             | Yes     | Yes    | Yes    | [Documentation](https://docs.microsoft.com/uwp/api/windows.ui.xaml.uielement.tapped) |
+| `DoubleTapped`        | Yes     | ??     | No     | [Documentation](https://docs.microsoft.com/uwp/api/windows.ui.xaml.uielement.doubletapped) |
```
davidjohnoliver: > ??

??

doc/articles/routed-events.md
```diff
@@ -0,0 +1,48 @@
...
+| `PointerEntered`      | Yes     | No     | Yes    | [Documentation](https://docs.microsoft.com/uwp/api/windows.ui.xaml.uielement.pointerentered) |
+| `PointerExited`       | Yes     | No     | Yes    | [Documentation](https://docs.microsoft.com/uwp/api/windows.ui.xaml.uielement.pointerexited) |
+| `PointerMoved`        | Yes     | No     | Yes    | [Documentation](https://docs.microsoft.com/uwp/api/windows.ui.xaml.uielement.pointermoved) |
+| `PointerPressed`      | Yes     | No     | Yes    | [Documentation](https://docs.microsoft.com/uwp/api/windows.ui.xaml.uielement.pointerpressed) |
+| `PointerReleased`     | Yes     | No     | Yes    | [Documentation](https://docs.microsoft.com/uwp/api/windows.ui.xaml.uielement.pointerreleased) |
+
+Notes:
+
+1. **Focus** events::
+   * iOS: The concept of _focus_ is emulated because not supported by the platform.
```
davidjohnoliver: > The concept of _focus_ is emulated because not supported by the platform.

Does this have any practical implications? This could apply to large parts of Uno, I don't think it necessarily needs to be mentioned.

doc/articles/routed-events.md
```diff
@@ -0,0 +1,48 @@
...
+   * Wasm: Current implementation is not totally reliable and doesn't support _lost focus_ most of the time.
+2. **Keyboard** events:
+   * iOS: `KeyDown` & `KeyUp` events are generated from only a `TextBox`. Only characted-related keyboard events are generated.
+
+### Unsupported _Routed Events_
+
+Those _routed events_ are not supported right now in Uno:
```
davidjohnoliver: These routed events

src/SamplesApp/SamplesApp.Shared/App.xaml.cs
```diff
@@ -72,7 +72,7 @@ protected override void OnLaunched(LaunchActivatedEventArgs e)
 					// When the navigation stack isn't restored navigate to the first page,
 					// configuring the new page by passing required information as a navigation
 					// parameter
-					rootFrame.Navigate(typeof(MainPage), e.Arguments);
+					rootFrame.Navigate(typeof(RoutedEventsPage), e.Arguments);
```
davidjohnoliver: Don't merge this

src/SamplesApp/SamplesApp.Shared/RoutedEventsPage.xaml
```diff
@@ -0,0 +1,48 @@
+﻿<Page
```
davidjohnoliver: This should go in Uno.UI.Samples

src/Uno.UI.Wasm/WasmScripts/Uno.UI.d.ts
```diff
@@ -167,7 +167,7 @@ declare namespace Uno.UI {
             * Add an event handler to a html element.
             *
             * @param eventName The name of the event
-            * @param onCapturePhase true means "on trickle down", false means "on bubble up". Default is false.
+            * @param onCapturePhase true means "on trickle down" (going down to target), false means "on bubble up" (bubbling back to ascestors). Default is false.
```
davidjohnoliver: ancestors

src/Uno.UI.Wasm/ts/WindowManager.ts
```diff
@@ -369,7 +369,7 @@
 			* Add an event handler to a html element.
 			*
 			* @param eventName The name of the event
-			* @param onCapturePhase true means "on trickle down", false means "on bubble up". Default is false.
+			* @param onCapturePhase true means "on trickle down" (going down to target), false means "on bubble up" (bubbling back to ascestors). Default is false.
```
davidjohnoliver: ancestors

src/Uno.UI.sln
```diff
@@ -3193,186 +3193,177 @@ Global
 		{21206058-D5EF-466D-8E4D-AC62357FB479}.Release|x64.Build.0 = Release|Any CPU
```
davidjohnoliver: Needs to be merged?

src/Uno.UI/KeyboardListener.Android.cs
```diff
@@ -0,0 +1,102 @@
...
+		public void SetRootContent(UIElement rootContent)
+		{
+			if (_rootContent == rootContent)
+			{
+				return; // nothing to do
+			}
+
+			if (_rootContent != null)
+			{
+				rootContent.GotFocus -= OnFocus;
```
davidjohnoliver: I'm guessing these should be `_rootContent`?

src/Uno.UI/KeyboardListener.Android.cs
```diff
@@ -0,0 +1,102 @@
...
+		private void OnFocus(UIElement focusedView)
+		{
+			if (_focusedElement == focusedView)
+			{
+				return; // nothing to do
+			}
+
+			if (_focusedElement != null)
+			{
+				//_focusedElement.KeyPress -= OnKeyPressed;
```
davidjohnoliver: Commented code, and below

src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
```diff
@@ -146,6 +146,27 @@ protected override void OnLoaded()
 			SetUpdateControlTemplate();
 
 			base.OnLoaded();
+
+			PointerPressed += OnPointerPressed;
+			PointerReleased += OnPointerReleased;
```
davidjohnoliver: It's actually not correct to call the virtual method from the event. From the [UWP docs](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.control.onpointerpressed) (emphasis mine):

> Control.OnPointerPressed(PointerRoutedEventArgs) Method
> Called **before** the PointerPressed event occurs.

If the overridden OnPointerPressed method sets Handled=true, then the event won't be raised. That's the reason for all the OnPointerXInternal methods in the existing code. Doing it this way will introduce a regression because the virtual method may be called before or after event handlers.

src/Uno.UI/UI/Xaml/Controls/Primitives/Thumb.iOS.cs
```diff
@@ -1,165 +0,0 @@
...
-		}
-
-		internal void DeltaDrag(CGPoint location)
-		{
-			DragDelta?.Invoke(this, new DragDeltaEventArgs(location.X - _startlocation.X, location.Y - _startlocation.Y));
-		}
-
-		/// <summary>
-		/// This GestureRecognizer is used to eat the touch events sent to the Thumb. Without such a system,
-		/// the events can be captured by a ScrollView if wrapped around a Thumb.
```
davidjohnoliver: The comment explains why this was originally there... does [the sample](https://nventive.visualstudio.com/Umbrella/_git/Uno.UI?path=%2FUno.UI.Samples%2FContent%2FUITests%2FSlider%2FSlider_Inside_ScrollViewer.xaml&version=GBmaster) still work correctly? (On Android as well)

doc/articles/routed-events.md
carldebilly: Not tested.  Good catch.

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.Android.cs
```diff
@@ -27,7 +27,7 @@ public partial class TextBox : View.IOnFocusChangeListener, IOnEditorActionListe
 	{
 		private int _keyboardAccessDelay = 50;
 		private TextBoxView _textBoxView;
-		private readonly SerialDisposable _keyPressDisposable = new SerialDisposable();
+		//private readonly SerialDisposable _keyPressDisposable = new SerialDisposable();
```
davidjohnoliver: commented code

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.Android.cs
```diff
@@ -52,8 +52,7 @@ protected override void OnUnloaded()
 		{
 			base.OnUnloaded();
 
-			_keyPressDisposable.Disposable = null;
-			PointerPressed -= OnPointerPressed;
+			//_keyPressDisposable.Disposable = null;
```
davidjohnoliver: commented code

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.Android.cs
```diff
@@ -411,7 +402,7 @@ private void SetupTextBoxView()
 		{
 			if (_textBoxView != null)
 			{
-				_keyPressDisposable.Disposable = _textBoxView.RegisterKeyPress(OnKeyPress);
+				//_keyPressDisposable.Disposable = _textBoxView.RegisterKeyPress(OnKeyPress);
```
davidjohnoliver: commented code

doc/articles/routed-events.md
carldebilly: Is it documented elsewhere?

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBoxView.Android.cs
```diff
@@ -32,8 +32,17 @@ public TextBoxView()
...
 
-            // This ensures the TextBoxView gets resized as Text changes
-            LayoutParameters = new Android.Views.ViewGroup.LayoutParams(
+			this.KeyDown += (s, e) =>
+			{
+				Console.WriteLine("!");
+			};
+			this.KeyUp += (s, e) =>
+			{
+				Console.WriteLine("!");
```
davidjohnoliver: Exciting though it may be, don't merge this :p

src/SamplesApp/SamplesApp.Shared/RoutedEventsPage.xaml
carldebilly: It's more a development surface than a test control.

src/Uno.UI/UI/Xaml/GestureHandler.Android.cs
```diff
@@ -12,135 +12,44 @@ namespace Windows.UI.Xaml
...
-		{
-			PointerCanceled += handler;
-			SetShouldHandleCancel(true);
-		}
-
-		internal void UnregisterPointerCanceled(PointerEventHandler handler)
-		{
-			PointerCanceled -= handler;
-			SetShouldHandleCancel(PointerCanceled != null);
+			// TODO: How expensive is this?
```
davidjohnoliver: Is the comment still valid? (If it's referring to the 'if else' chain, I think we can live with it)

src/Uno.UI/UI/Xaml/UIElement.cs
```diff
@@ -39,194 +39,247 @@ public partial class UIElement : DependencyObject
...
-			add => RegisterEventHandler(
-				"keydown",
-				value,
-				eventExtractorScript: "(evt instanceof KeyboardEvent) ? evt.key : \"0\"",
-				payloadConverter: keyStr => new KeyRoutedEventArgs { Key = VirtualKeyHelper.FromKey(keyStr), OriginalSource = this});
-			remove => UnregisterEventHandler("keydown", value);
+			// TODO: WPF reads Handled directly on RoutedEventArgs. 
+			switch (args)
+			{
+				case PointerRoutedEventArgs pointer:
```
davidjohnoliver: Can the ICancellableRoutedEventArgs interface be used here?

src/SamplesApp/SamplesApp.Shared/RoutedEventsPage.xaml
davidjohnoliver: Alright... but does it belong in the Uno.UI repo?

doc/articles/routed-events.md
carldebilly: Fixed.

src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
```diff
@@ -146,6 +146,27 @@ protected override void OnLoaded()
 			SetUpdateControlTemplate();
 
 			base.OnLoaded();
+
+			PointerPressed += OnPointerPressed;
```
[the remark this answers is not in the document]
carldebilly: That was one of the points to check and we decided to test it on real app.

src/Uno.UI.sln
carldebilly: Yep because I removed useless platforms in projects.

src/Uno.UI/KeyboardListener.Android.cs
carldebilly: Fixed.

src/Uno.UI/UI/Xaml/GestureHandler.Android.cs
carldebilly: It's reffering of one point moved from the PR to issue #322 :
> _Implement a WPF-inspired internal virtual RoutedEventArgs.InvokeEventHandler(Delegate handler, object target) to avoid switch cases in UIElement.InvokeHandler_

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBoxView.Android.cs
carldebilly: Good catch.

src/SamplesApp/SamplesApp.Shared/RoutedEventsPage.xaml
jeromelaban: Yes, it should. We'll move it to a "SampleControl" like UI.

doc/articles/routed-events.md
```diff
@@ -0,0 +1,48 @@
...
+| `PointerExited`       | Yes     | No     | Yes    | [Documentation](https://docs.microsoft.com/uwp/api/windows.ui.xaml.uielement.pointerexited) |
+| `PointerMoved`        | Yes     | No     | Yes    | [Documentation](https://docs.microsoft.com/uwp/api/windows.ui.xaml.uielement.pointermoved) |
+| `PointerPressed`      | Yes     | No     | Yes    | [Documentation](https://docs.microsoft.com/uwp/api/windows.ui.xaml.uielement.pointerpressed) |
+| `PointerReleased`     | Yes     | No     | Yes    | [Documentation](https://docs.microsoft.com/uwp/api/windows.ui.xaml.uielement.pointerreleased) |
+
+Notes:
+
+1. **Focus** events::
+   * iOS: The concept of _focus_ is emulated because not supported by the platform.
+   * Wasm: Current implementation is not totally reliable and doesn't support _lost focus_ most of the time.
```
MatFillion: Do we have the information to be more specific than "most of the time" and "not reliable" ? It's not the end of the world, but if we have the info, here is the place to write it down :)

src/Uno.UI/UI/Xaml/Controls/Primitives/Thumb.iOS.cs
carldebilly: Need to test it.

carldebilly: Need to test it more thoroughly, but didn't see such bug in my tests.

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.Android.cs
carldebilly: Will be solved soon.

src/Uno.UI/UI/Xaml/UIElement.cs
carldebilly: Yep!  But it's a new concept and this code was written before the interface.  Good catch.  Will be refactored in next update.

src/Uno.UI/KeyboardListener.Android.cs
carldebilly: The class has been removed.

src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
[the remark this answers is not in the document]
carldebilly: There's a new mecanism in place and only those overriding the methods are getting subscribed in the native platform.

src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
carldebilly: The called method is static and the overriden version is always enlisted as first in the invocation list, so it should be called first too.

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.Android.cs
carldebilly: Removed.

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBoxView.Android.cs
carldebilly: Fixed

src/Uno.UI/UI/Xaml/UIElement.iOS.cs
```diff
@@ -736,66 +608,56 @@ IEnumerable<UIView> GetMatchesInChildren(UIView parent)
...
-						PointerDeviceType = PointerDeviceType.Touch,
-					});
-				})
+				var recognizer = new UITapGestureRecognizer(OnGesture)
 				{
 					NumberOfTapsRequired = 1,
 				};
 
-				recognizer.ShouldReceiveTouch += (_, touch) => (touch.View == owner);
+				recognizer.ShouldReceiveTouch += (_, touch) => true;
```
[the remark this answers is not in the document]
carldebilly: To be honest, I don't know.

src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
```diff
@@ -606,6 +746,11 @@ protected virtual void OnBorderBrushChanged(Brush oldValue, Brush newValue)
 
 		protected virtual void OnFocusStateChanged(FocusState oldValue, FocusState newValue)
 		{
+			if (newValue == FocusState.Unfocused && oldValue == newValue)
```
davidjohnoliver: This is a DP callback, under what circumstances is it called with an unchanged value?

carldebilly: During the initialization on Android it's happening a lot of time.

doc/articles/routed-events.md
```diff
@@ -0,0 +1,179 @@
...
+ 8. **Returned for native bubbling**: let the event bubbling in the platform.
+ 9. **Any Parent Interested?**: check if any parent control is interested by a "handled" version of this event
+    (using `AddEventHandler(<evt>, <handler>, handledEventsToo: true)`).
+ 10. **Bubbling in managed code**: propagate the event in Uno to parents (for events not bubbling natively
+    or `handledEventsToo: true`).
+ 11. **Parent defined?**: if the element is connected to any parent element.
+ 12. **Processing finished**: no more handlers is interested by this event. Propagation is stopped.
+    Native bubbling is stopped too, because the event is fully handled.
+
+## Native bubbling vs Managed bubbling
```
davidjohnoliver: It's unfortunate that consumers have to worry about this distinction IMO. It's probably acceptable for 'RoutedEvents v1', but ideally we would find a way to have native views in the tree (which is pretty common) and still support handledEventsToo.

src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
```diff
@@ -614,36 +759,210 @@ protected virtual void OnFocusStateChanged(FocusState oldValue, FocusState newVa
...
+
+			if (__GetIsIOverrideImplemented(type, "OnGotFocus", routedArgs))
+			{
+				result |= RoutedEventFlag.GotFocus;
+			}
+
+			return ImplementedRoutedEvents[type] = result;
+		}
+
+		private static bool __GetIsIOverrideImplemented(Type type, string name, Type[] args)
```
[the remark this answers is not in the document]
carldebilly: Do you consider "_because_" as a valid explanation?

doc/articles/routed-events.md
carldebilly: I agree...  but we're limited by the fact we are dealing with platforms where this feature doesn't exist.
Maybe we'll find another way to support it in the future.

src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
```diff
@@ -146,6 +147,145 @@ protected override void OnLoaded()
 			SetUpdateControlTemplate();
 
 			base.OnLoaded();
+
+			var implementedEvents = GetImplementedRoutedEvents(GetType());
+
+			if (implementedEvents.HasFlag(RoutedEventFlag.PointerPressed))
+			{
+				PointerPressed += OnPointerPressed;
```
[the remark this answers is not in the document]
carldebilly: Nope. It's a visual illusion :-).

src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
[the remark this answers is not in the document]
carldebilly: Done.

jeromelaban:
Is there a test that validates the performance of worst propagation case ? (so we can follow it through time with test run time)

carldebilly:
I would like a review on the names for some properties like `EventsBubblingInManagedCode` because that won't be easy to change in the future.

PR #163, opened 2018-08-31
Include SHA1 in Uno.UI.Tasks msbuild tasks

jeromelaban:
This avoids versioning issues when running multiple versions on Uno.UI side-by-side in the same solution, or in sequentially in the same VS instance.

This is a workaround for https://github.com/Microsoft/msbuild/issues/1754


carldebilly:
Why `_v0` ?

jeromelaban:
To be found by the string replace :)

PR #175, opened 2018-09-06
Expose ItemsControl.OnItemsChanged properly

### LINKED ISSUE #173: About OnItemsChanged method of ItemsControl have not Implemented
In Uno the OnItemsChanged method of ItemsControl has not Implemented,but then Items property has    ItemsVectorChanged event; In UWP, the OnItemsChanged method has been implemented,but the ItemsVectorChanged event of Items property not work! why not use ItemsVectorChanged event to implement OnItemsChanged method in Uno? so we can use OnItemsChanged also in Uno!

jeromelaban:
Fixes #173

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/ItemsControl/ItemsControl.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/ItemsControl/ItemsControl.cs
@@ -85,6 +85,7 @@ namespace Windows.UI.Xaml.Controls
 			OnDisplayMemberPathChangedPartial(string.Empty, this.DisplayMemberPath);
 
 			_items.VectorChanged += (s, e) => {
+				OnItemsChanged(null);
 				SetNeedsUpdateItems();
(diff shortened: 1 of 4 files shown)


src/Uno.UI/UI/Xaml/Controls/ItemsControl/ItemsControl.cs
```diff
@@ -85,6 +85,7 @@ void Initialize()
 			OnDisplayMemberPathChangedPartial(string.Empty, this.DisplayMemberPath);
 
 			_items.VectorChanged += (s, e) => {
+				OnItemsChanged(null);
```
davidjohnoliver: should the parameter be null?

jeromelaban: Documentation says that the value is implementation specific, and is effectively null in the current runtime.

Issue #190, opened 2018-09-12
Control templates are applied too early


dr1rrb:
## I'm submitting a...
- Bug report (I searched for similar issues and did not find one) 

## Current behavior
If we instantiate a control and set the `Template` property, then this template is applied immediately (which leads to invoke the `OnApplyTemplate` in the control).

## Expected behavior
On Windows the template of a control is being applied only when this control is in the visual tree and either we explicitly invoke the `ApplyTemplate` method, or we wait for the first measure pass.

_Note: To get the full behavior of windows, we should be able to do:_
```csharp
var myControl = new MyControl { Template = _myControlTemplate };
_panelInTheVisualTree.Children.Add(myControl); // Template was not applied yet
myControl.ApplyTemplate(); // Will invoke `OnApplyTemplate` synchronously and returns `true`
_panelInTheVisualTree.Children.Remove(myControl); // `myControl` was not measured
```
_But if the panel is not in the visual tree, `myControl.ApplyTemplate()` should not do anything and returns `false`_

## Minimal reproduction of the problem with instructions
```csharp
var myControl = new MyControl { Template = _myControlTemplate };
```

## Environment
```
Nuget Package: Uno.UI
Package Version(s): 1.25.0-dev.42

Affected platform(s):
- [x] iOS
- [x] Android
- [?] WebAssembly
- [ ] Windows
- [ ] Build tasks

Visual Studio : _Irrelevant_
Relevant plugins: _Irrelevant_
```

dr1rrb:
This got mitigated by #192, but it's still not properly aligned with UWP.

davidjohnoliver:
`ApplyTemplate()` doesn't require the view to be in the visual tree, it just requires `Template` to be non-null. If the `Control` has an implicit style then its `Template` will be null on UWP until it's in the visual tree because the `Style` can't be resolved until that point. The root cause then is #119.

For now the best workaround is to set `FeatureConfiguration.Control.UseLegacyLazyApplyTemplate = true`.

peternary:
The specific issue we're encountering involves a custom control that wraps a ListView and accepts a DataTemplateSelector as a wrapper but includes a default DataTemplateSelector.  By the time we have received the DataTemplateSelector the ListView has already been rendered using the default DataTemplateSelector.

dr1rrb:
This is causing some troubles again with this:
```xaml
<Setter Property="DisplayTemplate" Value="{StaticResource templTextDisplay}" />
[...]

Issue #119, opened 2018-07-31
Align Styles behavior with UWP


jeromelaban:
The current implementation of Styles in Uno Platform is lacking in some areas, such as Implicit styles, default styles merging, etc.. this issue will track other issues related to the alignment of the Style behavior with Windows' UWP implementation.

jeromelaban:
Here's an assessment of what's missing, with description of some implicit behaviors and related questions: 

- How do implicit Styles behave? 
  - How does resolution behave?
    - Standard 'leafmost first.' Styles can be defined in Resources local to a control or programmatically. Applied correctly to views added to tree programmatically.
- How do resources behave?
  - How does FrameworkElement.Resources work under the hood?
    - Explicit resources are seemingly resolved (or at least their scope is fixed) at parse-time. Resources added programmatically, or belonging to ancestors, can't be referenced in xaml. (Compared to implicit Styles which seem more like they are resolved at runtime.)
     - Resources added by x:Name are available as StaticResources and are also set as fields on the control, hence can be accessed by x:Bind.
  - What programmatic interventions are available?
    - Limited, see above.
  - What can be bound to?
    - Both x:Key and x:Name resources. 
  - How does ResourceDictionary behave at low level? (Eg what does Contains return if key is only in merged dictionary, etc)
- Do Uno and UWP Styles match in code?
  - Same comportment? Same contract?
    - ResourceDictionary doesn't support MergedDictionary; will resolve global defaults if local values not found.
    - UWP seems to resolve event handlers and x:Bind using IComponentConnector and IComponentConnector2 interfaces respectively. 
- How does DefaultStyleKey work?
  - Need to test class libraries
- What's the unset value of a custom control?
    - Typically the type of the base class, including classes with protected constructors (Control, ListViewBase).
    - What if it's modified at runtime?
- How do resources from class libraries behave?
  - Can a Style from class library X be the default for a type from class library Y?
  - Can a Style from a class library be used in the default layer for a framework type? If so, local to that class library, or in consuming external apps/libraries?

dr1rrb:
Linking #190 which is blocked until this got completed.

PR #192, opened 2018-09-12
Wait for a control to have a parent before loading its template

dr1rrb:
Issue: #190 

## PR Type
- Bugfix

Templates of controls are loaded as soon as the `Template` property is set

## What is the new behavior?
We check if the control has a parent before loading its template.

_Note: for now we only check if the control has a parent, but we should also check if the control is in the visual tree. See the #190 for full behavior description._

## PR Checklist
- [x] Tested code with current [supported SDKs](../readme.md#supported)
- [x] Docs have been added/updated which fit [documentation template](https://github.com/nventive/Uno/blob/master/docs/.feature-template.md). (for bug fixes / features)

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
@@ -91,10 +91,10 @@ namespace Windows.UI.Xaml.Controls
 		internal void SetUpdateControlTemplate(bool forceUpdate = false)
 		{
 			if (
-				!FeatureConfiguration.Control.UseLegacyLazyApplyTemplate
-				|| forceUpdate
-				|| this.HasParent()
-				|| CanCreateTemplateWithoutParent
+				(!FeatureConfiguration.Control.UseLegacyLazyApplyTemplate
+					|| forceUpdate
+					|| CanCreateTemplateWithoutParent)
+				&& this.HasParent() // Instead we should check if we are in a the visual tree
 			)
 			{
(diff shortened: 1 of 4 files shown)


src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
```diff
@@ -91,10 +91,10 @@ private void OnTemplateChanged(DependencyPropertyChangedEventArgs e)
...
 		{
 			if (
-				!FeatureConfiguration.Control.UseLegacyLazyApplyTemplate
-				|| forceUpdate
-				|| this.HasParent()
-				|| CanCreateTemplateWithoutParent
+				(!FeatureConfiguration.Control.UseLegacyLazyApplyTemplate
+					|| forceUpdate
+					|| CanCreateTemplateWithoutParent)
+				&& this.HasParent() // Instead we should check if we are in a the visual tree
```
jeromelaban: Add a reference to #119

dr1rrb: I also linked the source bug of this PR into your epic since the bug cannot be resolved properly until the epic got completed.

davidjohnoliver:
This isn't the right fix, even as a temporary one. It breaks `ApplyTemplate()` and it breaks `CanCreateTemplateWithoutParent`.

The good fix to restore lazy template loading is to set the flag for that, `FeatureConfiguration.Control.UseLegacyLazyApplyTemplate`.

PR #206, opened 2018-09-19
Blog post - Code generation part 2

davidjohnoliver:


doc/blog/201809-UnoUnderTheHoodDependencyObjectGeneration.md
```diff
@@ -0,0 +1,76 @@
...
+public MyBindableObject : DependencyObject { 
+    // Custom dependency properties... 
+}
+````
+ 
+There are non-view framework types, too, which inherit from `DependencyObject`, like `Transforms` and `Brushes`. 
+ 
+We face a weaker form of this problem - wanting to have two base types - in other cases is well. In a few places in the framework we inherit from a more derived native view type. For example, `ScrollContentPresenter` inherits from the native scroll view on Android and iOS. But we also want `ScrollContentPresenter` to expose the methods and properties of `FrameworkElement`. 
+ 
+## Mixing it up
```
MatFillion: as discussed, before or after this subtitle I would mention that we successfully solved that problem using code gen.

davidjohnoliver:
@ameliemorency Updated with changes requested by FT.

PR #207, opened 2018-09-19
added 'global::' tag preceeding a type cast, this will fix any potent…

mohaider:
Issue: #
 114921

## PR Type
fixing potential issues arising if projects have a similar namespace as uno.

 If a project who's namespace ends with uno, then conflicts will emerge. 

## What is the new behavior?
Fixes namespace conflicts


davidjohnoliver:
It seems to break Generic.xaml. Does it build correctly locally?

mohaider:
> It seems to break Generic.xaml. Does it build correctly locally? https://uno-platform.visualstudio.com/Uno%20Platform/_build/results?buildId=645&view=logs

I have realized I made a big mistake with this, adding a global:: tag to primitives so this will break generated code if they do have those types. I'm currently investigating this and will create a new pr.

PR #220, opened 2018-09-27
Control - fix ApplyTemplate()

davidjohnoliver:
Fix bug where Control.ApplyTemplate() would only work when view was attached to the visual tree. UWP only requires that Template be non-null.

Restore correct behaviour of Control.CanCreateTemplateWithoutParent.

Issue: #133790

## PR Type

## PR Checklist

### DIFF
--- a/src/Uno.UI.Tests/ControlTests/Given_Control.cs
+++ b/src/Uno.UI.Tests/ControlTests/Given_Control.cs
@@ -13,46 +13,64 @@ namespace Uno.UI.Tests.ControlTests
 	public partial class Given_Control
 	{
 		[TestMethod]
-		public void When_ManualyApplyTemplate()
+		public void When_ManuallyApplyTemplate()
 		{
-			var templatedRoot = default(UIElement);
-			var sut = new MyControl
+			var current = FeatureConfiguration.Control.UseLegacyLazyApplyTemplate;
+			try
 			{
-				Template = new ControlTemplate(() => templatedRoot = new Grid())
-			};
+				FeatureConfiguration.Control.UseLegacyLazyApplyTemplate = true;
+				var templatedRoot = default(UIElement);
+				var sut = new MyControl
+				{
+					Template = new ControlTemplate(() => templatedRoot = new Grid())
+				};
 
-			Assert.IsNull(sut.TemplatedRoot);
-			Assert.IsNull(templatedRoot);
+				Assert.IsNull(sut.TemplatedRoot);
+				Assert.IsNull(templatedRoot);
 
-			new Grid().Children.Add(sut); // This kind-of simulate that the control is in the visual tree.
+				new Grid().Children.Add(sut); // This kind-of simulate that the control is in the visual tree.
 
-			Assert.IsNull(sut.TemplatedRoot);
-			Assert.IsNull(templatedRoot);
+				Assert.IsNull(sut.TemplatedRoot);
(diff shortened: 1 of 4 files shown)


src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
```diff
@@ -91,10 +91,10 @@ private void OnTemplateChanged(DependencyPropertyChangedEventArgs e)
 		internal void SetUpdateControlTemplate(bool forceUpdate = false)
 		{
 			if (
```
dr1rrb: If we rollback the whole change, it means that we will have to use the legacy mode in all our applications... Invoke the `ApplyTemplate` just before the `OnLoaded` on Android did not do the trick?

src/Uno.UI.Tests/ControlTests/Given_Control.cs
```diff
@@ -13,46 +13,62 @@ namespace Uno.UI.Tests.ControlTests
...
-			var applied = sut.ApplyTemplate();
+				new Grid().Children.Add(sut); // This kind-of simulate that the control is in the visual tree.
 
-			Assert.IsTrue(applied);
-			Assert.IsNotNull(sut.TemplatedRoot);
-			Assert.AreSame(templatedRoot, sut.TemplatedRoot);
+				Assert.IsNull(sut.TemplatedRoot);
+				Assert.IsNull(templatedRoot);
+				
+				sut.Visibility = Visibility.Collapsed;
```
dr1rrb: We are testing a Uno side effect, at least we should add a comment, and keep valid tests as ignored

src/Uno.UI.Tests/DependencyProperty/Given_DependencyProperty.Propagation.cs
```diff
@@ -476,8 +476,6 @@ public void When_ControlTemplate_And_Animation()
 			});
 
 			SUT.Template = template;
-
```
dr1rrb: Same thing here : the behavior tested here is not valid on windows, we should not enforce that in tests

dr1rrb: Oups here as we set the template, you're right, it should work as is

src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
davidjohnoliver: Yes, if you need lazy template loading for the new activities you'll need to set the UseLegacyLazyApplyTemplate to true, until #119 is addressed.

src/Uno.UI.Tests/ControlTests/Given_Control.cs
davidjohnoliver: The original tests were invalid regarding UWP behaviour (ApplyTemplate() will work without a parent), I'll try to restore them in a valid form.

davidjohnoliver: @dr1rrb How about now?

src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
dr1rrb: Yeah I thought this flag would make our features incompatible, but it will break the `DataGrid`. I assume that no matter the name we can live with it for now :)

PR #238, opened 2018-10-03
Add WPF hosted mode

jeromelaban:
## PR Type
- Feature

This PR adds the ability to run the WebAssembly-scoped netstandard2.0 code of Uno.UI inside of the WPF host to allow for easier debugging.


src/Uno.UWP/UI/Core/CoreDispatcher.netstd.cs
```diff
@@ -10,9 +10,17 @@ public sealed partial class CoreDispatcher
     {
 		private Timer _timer;
 
+		/// <summary>
+		/// Provide a action that will delegate the dispach of CoreDispatcher work
+		/// </summary>
+		public static Action<Action> DispatchOverride;
```
davidjohnoliver: hmm... maybe this should at least be marked EditorBrowsable never

davidjohnoliver:
Interesting! Does the JS layer work the same way as on WASM?

jeromelaban:
It should work the same way yes, with a little catch about the async dispatch, but not sure yet how this is going to affect the behavior.

src/Uno.UWP/UI/Core/CoreDispatcher.netstd.cs
jeromelaban: Sure, I'll add that.

PR #250, opened 2018-10-04
Rework CoreBounds on Android like iOS does

YGuerin:
## Issue:

https://nventive.visualstudio.com/Umbrella/_workitems/edit/134157

## PR Type

- Bugfix

When RaiseNativeSizeChanged is called
1. RaiseSizeChanged is called
2. The visible bounds are updated
3. Once visible bounds are updated, OnConfigurationChanged is called and the native size is changed (Back to step 1)

## What is the new behavior?
1. Visible bounds are updated only if new bounds are different from current
2. RaiseSizeChanged is called
3. End of process

## PR Checklist


src/Uno.UI/UI/Xaml/Window.Android.cs
```diff
@@ -54,38 +54,25 @@ private static Window InternalGetCurrentWindow()
...
 			var newBounds = ViewHelper.PhysicalToLogicalPixels(new Rect(0, 0, screenWidth, screenHeight));
 
 			if (Bounds != newBounds)
 			{
 				Bounds = newBounds;
 
+				var applicationView = ApplicationView.GetForCurrentView();
+				if (applicationView != null)
+				{
+					applicationView.SetCoreBounds(newBounds);
```
davidjohnoliver: This will only be called if Bounds changes. What if the VisibleBounds change without Bounds changing? Eg if the StatusBar is hidden/displayed?

YGuerin: I've extracted it from newBounds and check for its own change

Issue #67, opened 2018-06-26
Maps Support


jeromelaban:
How can I have a screen and use a Map and put markers, have the map moved events, user click, etc, using UNO, for Android, iOS and Windows 10?
It would be good to have this included in the sample app.

Moved from https://github.com/nventive/Uno.Playground/issues/8

SetTrend:
To me it looks this issue is stalled (NuGet gallery, however, shows high activity for this control).

I'm planning to write an instant meeting app which intrinsically needs full maps support. - Could you please update documentation and above checklist to elaborate on what is available in Uno Maps today?

nnikos123:
I must admit that I was almost convinced myself to use UNO for my new project but after a closer look on map support I switched back to xamarin.
I think map support is a must have feature for many applications. 
Please update the support of mapcontrol
Thanks

francoistanguay:
Have you explored using Native Maps directly in the meantime (Per platform) and continue to benefit from all Uno has to offer? With Uno, you can always mix & match UWP/WinUI controls with Native as needed if there's something we haven't yet fully implemented.

What type of app are you building?

MartinZikmund:
I have used this map control in UWP before - https://github.com/ClemensFischer/XAML-Map-Control. It may be ported to Uno plus we can even take it as inspiration to implement the UWP's `MapControl`

DevTKSS:
@Jen-Uno seen you tiding up some older ones 👍 can you tell if here has been any proceeding? I only know the LiveCharts2 for using Maps in Uno 🤔 so option 1: Uno says thats all we get, or option 2: any update?

Jen-Uno:
@DevTKSS still in the backlog, but it is low priority on the roadmap.

MartinZikmund:
In general, it is preferrable to use [MapsUI](https://github.com/Mapsui/Mapsui) for Uno Platform apps. Currently WinUI `MapControl` is a webview-based control as well, so it feels this is not worth investigating until WinUI provides a proper native control as well. For that reason, I think we can even close this issue until the situation changes @Jen-Uno

DevTKSS:
@MartinZikmund could someone make a short how to for the usage? searching in the docs for "Map" only sends us to the mobile only map control 🤔

bricelam:
[...]

PR #253, opened 2018-10-09
Add basic MapControl, GeoLocator support

jeromelaban:
Issue: #67

## PR Type
- Feature 

## What is the new behavior?
Add basic support for `MapControl` and `GeoLocator`.

## PR Checklist


src/Uno.UI/UI/Xaml/Maps/MapControl.cs
```diff
@@ -0,0 +1,727 @@
...
+		Windows.UI.Xaml.DependencyProperty.Register(
+			"Layers", typeof(IList<MapLayer>),
+			typeof(MapControl),
+			new FrameworkPropertyMetadata(default(IList<MapLayer>)));
+		public static DependencyProperty RegionProperty { get; } =
+		Windows.UI.Xaml.DependencyProperty.Register(
+			"Region", typeof(string),
+			typeof(MapControl),
+			new FrameworkPropertyMetadata(default(string)));
+		public IReadOnlyList<MapElement> FindMapElementsAtOffset(global::Windows.Foundation.Point offset)
```
davidjohnoliver: Everything that's not implemented should be [NotImplemented]

src/Uno.UI.Maps/themes/Generic.xaml
```diff
@@ -0,0 +1,39 @@
...
+					xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+					xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+					xmlns:converters="using:Uno.UI.Converters"
+					xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+					xmlns:maps="using:Windows.UI.Xaml.Controls.Maps" 
+					xmlns:mapsPresenter="using:Windows.UI.Xaml.Controls.Maps.Presenter" 
+					mc:Ignorable="d">
+
+    <!-- Default style for Windows.UI.Xaml.Controls.Maps.MapControl -->
+    <Style TargetType="maps:MapControl">
```
davidjohnoliver: So... the default MapControl Style is defined in a different NuGet package? I have a feeling this will no longer work once Styles handling is aligned with UWP.

src/Uno.UI/UI/Xaml/Maps/MapControl.cs
jeromelaban: It can't that's thing for this control. The fact that it's implemented or not comes from the `Presenter` part of it.

src/Uno.UI.Maps/themes/Generic.xaml
jeromelaban: Why would it ? There can be a default style for a control in another package, as long as there is only one, and the map app's precedence will take over.

src/Uno.UI/UI/Xaml/Maps/MapControl.cs
davidjohnoliver: Not sure I follow... I don't see any of those members delegating implementation to the presenter.

jeromelaban: They will, ultimately, but we don't know which will be in the loaded presenter. At this time, we're probably going to have to manage that at runtime.

PR #254, opened 2018-10-09
Adjust invalid Wasm initialization

jeromelaban:


src/Uno.UI.Wasm/ts/WindowManager.ts
```diff
@@ -3,13 +3,16 @@
 	export class WindowManager {
 
 		public static current: WindowManager;
+		private static _isHosted: boolean = false;
```
carldebilly: The `_` prefix is not standard in JS/TS ;-)

carldebilly:
Watch-out, you did a merge of master instead of a rebase.

jeromelaban:
That's why I do this afterwards: 
image

I wish the update button wasn't there, though...

PR #263, opened 2018-10-11
[iOS] Improve Image performance with remote urls

davidjohnoliver:
Internal issue https://nventive.visualstudio.com/Umbrella/_workitems/edit/133247

Image performance improvements on iOS:
 - don't dispatch unnecessarily
 - don't download images on UI thread in debug builds

### DIFF
--- a/src/Uno.UI/UI/Xaml/Media/ImageSource.iOS.cs
+++ b/src/Uno.UI/UI/Xaml/Media/ImageSource.iOS.cs
@@ -102,6 +102,11 @@ namespace Windows.UI.Xaml.Media
 			   )
 			)
 			{
+				if (ct.IsCancellationRequested)
+				{
+					return null;
+				}
+
 				if (Stream != null)
 				{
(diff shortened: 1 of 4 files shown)


src/Uno.UI/UI/Xaml/Media/ImageSource.iOS.cs
```diff
@@ -152,8 +155,7 @@ internal async Task<UIImage> Open(CancellationToken ct)
 
 					if (SupportsAsyncFromBundle)
 					{
-						// Since iOS9, UIImage.FromBundle is thread safe, so we need to disable 
-						// the Xamarin binding check, which is unconditional.
+						// Since iOS9, UIImage.FromBundle is thread safe.
```
jeromelaban: True, but only in release mode. This will fail in debug (the calls to UIApplication.CheckForIllegalCrossThreadCalls are made regardless of the thread safety-ness of the API)

PR #266, opened 2018-10-12
Add a CommandBarHelper that allows apps to use CommanBar without using a Frame

dr1rrb:
## PR Type
 - Feature

`CommandBar` are tightly coupled to the `Frame` and cannot be used without one.

## What is the new behavior?
We are now able to render a `CommandBar` using a `UINavigationController` not managed by a `Frame`.

## PR Checklist

### DIFF
--- a/doc/articles/working-with-commandbar.md
+++ b/doc/articles/working-with-commandbar.md
@@ -295,6 +295,12 @@ To ensure everything works properly, you must follow a few rules:
 * The `CommandBar` must be accessible as soon as the page is being navigated to (i.e., don't put it inside a `DataTemplate` or an `AsyncValuePresenter`).
 * There can only be one `CommandBar` per page.
 
+## Extensibility
+
+The `CommandBar` it automatically managed by the `Frame` control, however you can still use the "native" mode of the `CommandBar` with your own navigation mechanisim.
+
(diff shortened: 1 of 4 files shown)


src/Uno.UI/UI/Xaml/DependencyObjectStore.cs
```diff
@@ -1572,7 +1572,7 @@ public bool Equals(WeakReference x, WeakReference y)
 
 			public int GetHashCode(WeakReference obj)
 			{
-				return obj?.GetHashCode() ?? 0;
+				return obj?.Target?.GetHashCode() ?? 0;
```
davidjohnoliver: Relevant to this PR? Should be a separate commit at least

davidjohnoliver:
You should update https://github.com/nventive/Uno/blob/master/doc/articles/working-with-commandbar.md

src/Uno.UI/UI/Xaml/DependencyObjectStore.cs
dr1rrb: Right, It's not relevant, I removed it and will do another PR for that.

dr1rrb:
> You should update https://github.com/nventive/Uno/blob/master/doc/articles/working-with-commandbar.md

Right, I searched (a bit ;) ) for the doc but I didn't find it. We should have a convention to name all the files by the control name and avoid prefix or suffix like "working with".

PR #276, opened 2018-10-17
Add build-time macOS platform support

### LINKED ISSUE #229: Instructions to build on Mac
## I'm submitting a...

I would like to have instructions on how to build Uno on a Mac, there does not seem to be a way of compiling the solution on a Mac, the out-of-the-box experience crashes on startup.

jeromelaban:
fixes #229

## PR Type
- Feature
- Bugfix

Building the Uno.UI solution in VS for Mac does not work.

## What is the new behavior?
Enables parts of the Uno.UI solution to build in VS for mac.
Add the bases for the macOS platform support.

## PR Checklist


davidjohnoliver:
Pretty impressive. How much actually works at the moment?

src/Uno.UI/Controls/BindableNSView.macOS.cs
```diff
@@ -0,0 +1,118 @@
...
+
+		/// <summary>
+		/// Moves a view from one position to another position, without unloading it.
+		/// </summary>
+		/// <param name="oldIndex">The old index of the item</param>
+		/// <param name="newIndex">The new index of the item</param>
+		/// <remarks>
+		/// The trick for this method is to move the child from one position to the other
+		/// without calling RemoveView and AddView. In this context, the only way to do this is
+		/// to call BringSubviewToFront, which is the only available method on ViewGroup that manipulates
```
davidjohnoliver: Comment refers to ViewGroup, and BringSubviewToFront is commented out below

jeromelaban: That's most probably the case on iOS too :) I'll fix it there as well.

jeromelaban:
>Pretty impressive. How much actually works at the moment?

It shows the main window, and probably other things too, considering a lot can be reused from iOS. But there are lots to do, even a new ListView 😉

davidjohnoliver:
> > Pretty impressive. How much actually works at the moment?
> 
> It shows the main window, and probably other things too, considering a lot can be reused from iOS. But there are lots to do, even a new ListView 😉

Nice :)

PR #277, opened 2018-10-17
ManagedWeakReference - ignore natively-collected targets

davidjohnoliver:
Don't treat a weak-ref target as alive if it's a managed peer whose underlying object has been collected.

This prevents errors from attempting to apply bindings to a collected object, which seem to be particularly prevalent on Android 8 for ImplicitTextBlocks created by ContentPresenter.

Issue: #
[Android][ListView] Group headers can disappear when using ScrollIntoView behavior. https://nventive.visualstudio.com/Umbrella/_workitems/edit/135202
[Android] Contents of button can disappear after a refresh https://nventive.visualstudio.com/Umbrella/_workitems/edit/131884

### DIFF
--- a/src/Uno.UI/DataBinding/ManagedWeakReference.cs
+++ b/src/Uno.UI/DataBinding/ManagedWeakReference.cs
@@ -4,6 +4,11 @@ using System.Runtime.InteropServices;
 using System.Text;
 using Uno.Extensions;
 using Uno.Logging;
+#if __ANDROID__
+using _NativeObject = Android.Runtime.IJavaObject;
+#elif __IOS__
+using _NativeObject = ObjCRuntime.INativeObject;
+#endif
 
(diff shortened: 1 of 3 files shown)


src/Uno.UI/DataBinding/ManagedWeakReference.cs
```diff
@@ -106,5 +149,13 @@ public void Dispose()
 		{
 			_disposed = true;
 		}
+
+#if !__ANDROID__ && !__IOS__
+		// Dummy
```
carldebilly: Dummy?

davidjohnoliver: Explicated

davidjohnoliver:
@jeromelaban Good point, I'll set up some tests.

PR #279, opened 2018-10-18
[iOS] ImageSource - use NSUrlSession

davidjohnoliver:
When downloading remote urls and no external downloader is set, use NSUrlSession instead of NSData.FromUrl. This enables response caching, download interruption, etc.

Issue: #
Internal: https://nventive.visualstudio.com/Umbrella/_workitems/edit/133247

## PR Type

Feature

Improved performance with downloaded images on iOS.

### DIFF
--- a/src/Uno.UI/UI/Xaml/Media/ImageSource.iOS.cs
+++ b/src/Uno.UI/UI/Xaml/Media/ImageSource.iOS.cs
@@ -13,6 +13,7 @@ using UIKit;
 using Uno.Logging;
 using Windows.UI.Core;
 using Uno.Disposables;
+using Microsoft.Extensions.Logging;
 
 namespace Windows.UI.Xaml.Media
 {
@@ -27,6 +28,17 @@ namespace Windows.UI.Xaml.Media
 		{
 			SupportsAsyncFromBundle = UIDevice.CurrentDevice.CheckSystemVersion(9, 0);
 			SupportsFromBundle = UIDevice.CurrentDevice.CheckSystemVersion(8, 0);
+		}
+
+		private static NSUrlSession _defaultSession;
+		/// <summary>
+		/// The <see cref="NSUrlSession"/> to use for remote url downloads, if using the platform downloader.
+		/// By default <see cref="NSUrlSession.SharedSession"/> will be used.
+		/// </summary>
+		public static NSUrlSession DefaultSession
+		{
+			get => _defaultSession ?? NSUrlSession.SharedSession;
+			set => _defaultSession = value;
 		}
 
 		protected ImageSource()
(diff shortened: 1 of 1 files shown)


src/Uno.UI/UI/Xaml/Media/ImageSource.iOS.cs
```diff
@@ -190,44 +202,65 @@ internal async Task<UIImage> Open(CancellationToken ct)
...
-					this.Log().Debug($"Loading image from [{WebUri.OriginalString}]");
-				}
+					NSUrlSessionDataTask task;
+					var awaitable = DefaultSession.CreateDataTaskAsync(request, out task);
+					ct.Register(OnCancel);
+					try
+					{
+						task.Resume(); // We need to call this manually https://bugzilla.xamarin.com/show_bug.cgi?id=28425#c3
+						var result = await awaitable;
+						task = null;
```
jeromelaban: What's this for ?

src/Uno.UI/UI/Xaml/Media/ImageSource.iOS.cs
```diff
@@ -190,44 +202,65 @@ internal async Task<UIImage> Open(CancellationToken ct)
...
+						var response = result.Response as NSHttpUrlResponse;
+
+						if (ct.IsCancellationRequested)
+						{
+							return;
+						}
+						else if (!IsSuccessful(response.StatusCode))
+						{
+							if (this.Log().IsEnabled(LogLevel.Error))
```
jeromelaban: We're not doing this yet, but maybe we should add the raise of [ImageFailed](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.image.imagefailed).

src/Uno.UI/UI/Xaml/Media/ImageSource.iOS.cs
```diff
@@ -190,44 +202,65 @@ internal async Task<UIImage> Open(CancellationToken ct)
...
+						if (ct.IsCancellationRequested)
+						{
+							if (this.Log().IsEnabled(LogLevel.Debug))
+							{
+								this.Log().LogDebug(e.ToString());
+							}
+						}
+						else if (this.Log().IsEnabled(LogLevel.Error))
+						{
+							this.Log().LogError(e.ToString());
```
jeromelaban: Same here for ImageFailed

src/Uno.UI/UI/Xaml/Media/ImageSource.iOS.cs
davidjohnoliver: Partly to avoid a leak if the CT is long-lived, partly so that cancelling the CT after the download completes is a no-op. (Mainly caution in other words.)

src/Uno.UI/UI/Xaml/Media/ImageSource.iOS.cs
davidjohnoliver: ImageFailed will be raised later if ImageData is null when Open() returns. It could be interesting to include the error in the event args, though.

src/Uno.UI/UI/Xaml/Media/ImageSource.iOS.cs
```diff
@@ -29,6 +30,17 @@ static ImageSource()
...
+		private static NSUrlSession _defaultSession;
+		/// <summary>
+		/// The <see cref="NSUrlSession"/> to use for remote url downloads, if using the platform downloader.
+		/// By default <see cref="NSUrlSession.SharedSession"/> will be used.
+		/// </summary>
+		public static NSUrlSession DefaultSession
+		{
+			get => _defaultSession ?? NSUrlSession.SharedSession;
+			set => _defaultSession = value;
+		}
```
carldebilly: Alternative code:
``` csharp
public static NSUrlSession DefaultSession { get; set; } = NSUrlSession.SharedSession;
```

src/Uno.UI/UI/Xaml/Media/ImageSource.iOS.cs
```diff
@@ -190,44 +202,65 @@ internal async Task<UIImage> Open(CancellationToken ct)
...
-					else
+
+					void OnCancel()
 					{
-						ImageData = UIImage.LoadFromData(data);
+						// Cancel the current download
+						task?.Cancel();
 					}
+
+					bool IsSuccessful(nint status) => status < 300 || status == 304;
```
carldebilly: That would be a strange behavior if your get a `304` here, because it means you have no payload to display.

davidjohnoliver: Fixed

PR #283, opened 2018-10-19
Include navigation bar in the visible bounds calculation

YGuerin:
## Issue: 
On few devices, the navigation can be docked / undocked causing bad visible bounds calculation

https://nventive.visualstudio.com/Umbrella/_workitems/edit/132002

## PR Type

- Bugfix

The navigation bar height / visibility is ignored in the visible bounds

## What is the new behavior?
The navigation bar height / visibility is included in the visible bounds

## PR Checklist

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/BorderLayerRenderer.Android.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/BorderLayerRenderer.Android.cs
@@ -44,7 +44,8 @@ namespace Windows.UI.Xaml.Controls
 			Thickness borderThickness,
 			Brush borderBrush,
 			CornerRadius cornerRadius,
-			Thickness padding
+			Thickness padding,
+			bool willUpdateMeasures = false
 			)
 		{
 			// This is required because android Height and Width are hidden by Control.
@@ -88,7 +89,14 @@ namespace Windows.UI.Xaml.Controls
 					}
 				}
 
-				view.Invalidate();
+				if (willUpdateMeasures)
+				{
+					view.RequestLayout();
+				}
+				else
+				{
+					view.Invalidate();
+				}
 
 				_currentState = newState;
 			}
--- a/src/Uno.UI/UI/Xaml/Window.Android.cs
+++ b/src/Uno.UI/UI/Xaml/Window.Android.cs
@@ -1,15 +1,9 @@
 #if XAMARIN_ANDROID
-using System;
-using System.Drawing;
-using System.Linq;
-using System.Runtime.InteropServices;
-using System.Threading;
 using Android.App;
-using Android.Content.Res;
+using Android.Util;
 using Android.Views;
 using Uno.UI;
 using Windows.Foundation;
-using Windows.Foundation.Metadata;
 using Windows.UI.Core;
 using Windows.UI.ViewManagement;
 
@@ -26,10 +20,9 @@ namespace Windows.UI.Xaml
 			CoreWindow = new CoreWindow();
 		}
 
-		private void InternalActivate()
-		{
+		internal int SystemUiVisibility { get; set; }
 
-		}
+		private bool IsNavigationBarVisible => (SystemUiVisibility & (int)SystemUiFlags.HideNavigation) == 0;
 
 		private void InternalSetContent(UIElement value)
 		{
@@ -56,18 +49,19 @@ namespace Windows.UI.Xaml
 		{
 			var newBounds = ViewHelper.PhysicalToLogicalPixels(new Rect(0, 0, screenWidth, screenHeight));
 			var statusBarHeight = GetLogicalStatusBarHeight();
+			var navigationBarHeight = GetLogicalNavigationBarHeight();
 
 			var newVisibleBounds = new Rect(
 				x: newBounds.X,
 				y: newBounds.Y + statusBarHeight,
 				width: newBounds.Width,
-				height: newBounds.Height - statusBarHeight
+				height: newBounds.Height - statusBarHeight - navigationBarHeight
 			);
 
 			var applicationView = ApplicationView.GetForCurrentView();
 			if (applicationView != null && applicationView.VisibleBounds != newVisibleBounds)
 			{
-				applicationView.SetCoreBounds(newBounds);
+				applicationView.SetCoreBounds(newVisibleBounds);
 			}
 
 			if (Bounds != newBounds)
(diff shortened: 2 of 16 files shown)


src/Uno.UI/UI/Xaml/Controls/BorderLayerRenderer.Android.cs
```diff
@@ -88,7 +88,7 @@ Thickness padding
 					}
 				}
 
-				view.Invalidate();
+				view.RequestLayout();
```
davidjohnoliver: No, this is a big deoptimization. In general we don't want to re-layout unless we know we have to, and we can enter this code path for changes that don't affect the measured size of the view (eg Background colour changes).

Try calling RequestLayout() from OnPaddingChanged instead.

src/Uno.UI/UI/Xaml/Window.Android.cs
```diff
@@ -26,9 +20,21 @@ public Window()
 			CoreWindow = new CoreWindow();
 		}
 
-		private void InternalActivate()
+		public int SystemUiVisibility { get; internal set; }
```
davidjohnoliver: should be internal since it's not part of UWP contract

src/Uno.UI/UI/Xaml/Window.Android.cs
```diff
@@ -26,9 +20,21 @@ public Window()
...
+			get
+			{
+				var decorView = (ContextHelper.Current as Activity)?.Window?.DecorView;
 
+				return (SystemUiVisibility & (int)SystemUiFlags.HideNavigation) == 0
+					&& (decorView == null || ((int)decorView.SystemUiVisibility & (int)SystemUiFlags.HideNavigation) == 0);
+			}
+		}
+
+		private void InternalActivate()
```
davidjohnoliver: ?

src/Uno.UI/UI/Xaml/Window.Android.cs
```diff
@@ -56,18 +62,19 @@ internal void RaiseNativeSizeChanged(int screenWidth, int screenHeight)
...
 				width: newBounds.Width,
-				height: newBounds.Height - statusBarHeight
+				height: newBounds.Height - statusBarHeight - navigationBarHeight
 			);
 
 			var applicationView = ApplicationView.GetForCurrentView();
 			if (applicationView != null && applicationView.VisibleBounds != newVisibleBounds)
 			{
-				applicationView.SetCoreBounds(newBounds);
+				applicationView.SetCoreBounds(newVisibleBounds);
```
davidjohnoliver: This was using the wrong value before?

src/Uno.UI/Uno.UI.csproj
```diff
@@ -32,6 +32,7 @@
 
 	<ItemGroup>
 		<Compile Remove="Mixins\**\*.cs" />
+		<None Remove="UI\OnSystemUiVisibilityChangeListener.Android.cs" />
```
davidjohnoliver: don't modify this file

src/Uno.UI/UI/Xaml/Window.Android.cs
```diff
@@ -95,12 +101,37 @@ private int GetLogicalStatusBarHeight()
...
+			if (defaultDisplay != null && IsNavigationBarVisible)
+			{
+				defaultDisplay.GetMetrics(metrics);
+				var usableHeight = metrics.HeightPixels;
+
+				defaultDisplay.GetRealMetrics(metrics);
+				var realHeight = metrics.HeightPixels;
+
+				logicalNavigationBarHeight = realHeight > usableHeight
+					? ViewHelper.PhysicalToLogicalPixels(realHeight - usableHeight)
```
davidjohnoliver: so neither GetMetrics nor GetRealMetrics include the status bar?

src/Uno.UI/Uno.UI.csproj
YGuerin: Happens every time I add a file. Is there any way to avoid that ?

src/Uno.UI/UI/Xaml/Window.Android.cs
YGuerin: Yes. @fcantin-nventive noticed that one

src/Uno.UI/UI/Xaml/Window.Android.cs
YGuerin: Existing code (Line 29 on the original version)

src/Uno.UI/UI/Xaml/Controls/BorderLayerRenderer.Android.cs
YGuerin: I moved the logic here just setting a bool in OnPaddingChanged. Could be useful if we need to set it in OnBorderThicknessChanged and OnCornerRadiusChanged

src/Uno.UI/UI/Xaml/Window.Android.cs
```diff
@@ -26,9 +20,21 @@ public Window()
 			CoreWindow = new CoreWindow();
 		}
 
-		private void InternalActivate()
+		public int SystemUiVisibility { get; internal set; }
+
+		public bool IsNavigationBarVisible
```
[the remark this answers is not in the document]
YGuerin: Done

src/Uno.UI/Uno.UI.csproj
davidjohnoliver: Not as far as I know, it seems to be a quirk of Visual Studio when confronted with Uno's wildcard-based scheme. Just make sure not to include changes to the csproj file in your commits.

src/Uno.UI/UI/Xaml/Window.Android.cs
YGuerin: Actually it's set to 0 by default. That's why I'm combining this SystemUiVisibility  to the decorView one. If one of them hide the navigation, so we will not measure navigation bar. OnSystemUiVisibilityChange is not called on page initialization.

src/Uno.UI/UI/Xaml/Window.Android.cs
YGuerin: Need to confirm this one but I think neither or both are handling the status bar. On it

src/Uno.UI/UI/Xaml/Window.Android.cs
davidjohnoliver: Ok, well, remove it ;)

src/Uno.UI/UI/Xaml/Window.Android.cs
MatFillion: I suppose it should be initialized in OnCreate?

src/Uno.UI/UI/Xaml/Window.Android.cs
YGuerin: Done

src/Uno.UI/UI/Xaml/Window.Android.cs
YGuerin: It looks like the status bar is always included in both.
See this : https://medium.com/@elye.project/an-important-note-when-managing-different-screen-height-3140e26e381a

YGuerin: According to my tests, it is included only if the status bar is translucent, it is not otherwise.
Should be the same for the navigation bar (on it)

PR #285, opened 2018-10-22
Implement MediaPlayer.Dispose()

fcantin-nventive:
## PR Type
Feature implementation

MediaPlayer.Dispose() is wrongly marked as NotImplemented, which generates a warning in debug/an error in release at build time.

## What is the new behavior?
MediaPlayer.Dispose() can be called, in order to stop the video when you leave its page.

## Other information
Internal work item: 135799

### DIFF
--- a/src/Uno.UWP/Media/Playback/MediaPlayer.iOS.cs
+++ b/src/Uno.UWP/Media/Playback/MediaPlayer.iOS.cs
@@ -380,5 +380,10 @@ namespace Windows.Media.Playback
 		{
 			_videoLayer.VideoGravity = gravity;
 		}
+
+		protected override void Dispose(bool disposing)
+		{
+			TryDisposePlayer();
+		}
 	}
 }
--- a/doc/ReleaseNotes/_ReleaseNotes.md
+++ b/doc/ReleaseNotes/_ReleaseNotes.md
@@ -4,6 +4,8 @@
 
 ### Features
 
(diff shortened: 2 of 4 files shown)


davidjohnoliver:
"in order to stop the video when you leave its page" Does it... do that? Given that the override you added for iOS literally does nothing, and the one on Android seems to do something else?

fcantin-nventive:
> "in order to stop the video when you leave its page" Does it... do that? Given that the override you added for iOS literally does nothing, and the one on Android seems to do something else?

It does, yes. For one thing, I've updated the code of the PR to add cleanup code on the iOS side. Furthermore, it is important to call Dispose so that the video stops on both Android and UWP, otherwise the video's audio track keeps playing.

src/Uno.UWP/Media/Playback/MediaPlayer.iOS.cs
```diff
@@ -379,6 +379,12 @@ private void OnSeekCompleted(bool finished)
 		internal void UpdateVideoGravity(AVLayerVideoGravity gravity)
 		{
 			_videoLayer.VideoGravity = gravity;
+			this.Dispose();
```
davidjohnoliver: Why do you call Dispose() here?

fcantin-nventive: There's no good reason, it must be a leftover from tests. I'll remove it.

PR #290, opened 2018-10-24
[iOS] Ensures the back gesture is enabled and disabled properly when the CommandBar is visible, collapsed, visible with a navigation command and collapsed with a navigation command.

jeremiethibeault:
[iOS] Ensures the back gesture is enabled and disabled properly when the CommandBar is visible, collapsed, visible with a navigation command and collapsed with a navigation command.

Issue:
https://nventive.visualstudio.com/Umbrella/_workitems/edit/134819
https://nventive.visualstudio.com/Umbrella/_workitems/edit/134828


davidjohnoliver:
Don't you also need to revert https://github.com/nventive/Uno/pull/260?

jeremiethibeault:
> Don't you also need to revert #260?

PR #291, opened 2018-10-25
[WASM] ScrollViewer - implement ViewChanged + Vertical/HorizontalOffset

davidjohnoliver:
## PR Type

Feature: Enables the ViewChanged event and VerticalOffset/HorizontalOffset properties on the WASM ScrollViewer.


carldebilly:
This will conflict with changes for RoutedEvent... but that's ok ;-).

carldebilly:
You should throttle the scroll using following to this guidelines:
<https://developer.mozilla.org/en-US/docs/Web/Events/scroll#Example>

src/Uno.UI/UI/Xaml/Controls/ScrollContentPresenter/ScrollContentPresenter.netstd.cs
```diff
@@ -112,6 +112,30 @@ protected override Foundation.Size ArrangeOverride(Foundation.Size finalSize)
...
+		protected override void OnUnloaded()
+		{
+			base.OnUnloaded();
+			UnregisterEventHandler("scroll", (EventHandler)OnScroll);
+		}
+
+		private void OnScroll(object sender, EventArgs args)
+		{
+			int.TryParse(GetProperty("scrollLeft"), out var horizontalOffset);
+			int.TryParse(GetProperty("scrollTop"), out var verticalOffset);
```
carldebilly: You can extract those information in the `eventExtractorScript` to prevent interop.

davidjohnoliver: Nice, I'll try that.

PR #297, opened 2018-10-26
[iOS] Fix Popup bug where children lose TemplatedParent

fcantin-nventive:
## PR Type

Bugfix

ComboBox on iOS does not display its Popup

## What is the new behavior?
ComboBox on iOS now displays its Popup

## Other information
Internal issues: 136093  & 136172

### DIFF
--- a/doc/ReleaseNotes/_ReleaseNotes.md
+++ b/doc/ReleaseNotes/_ReleaseNotes.md
@@ -26,7 +26,9 @@
  * 131768 [iOS] Improve ListView.ScrollIntoView() when ItemTemplateSelector is set
  * 135202, 131884 [Android] Content occasionally fails to show because binding throws an exception
  * 135646 [Android] Binding MediaPlayerElement.Source causes video to go blank
+ * 136093, 136172 [iOS] ComboBox does not display its Popup
  * 134819, 134828 [iOS] Ensures the back gesture is enabled and disabled properly when the CommandBar is visible, collapsed, visible with a navigation command and collapsed with a navigation command. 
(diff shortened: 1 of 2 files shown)


davidjohnoliver:
Why is the PopupPanel losing its TemplatedParent? Is this a recent regression?

davidjohnoliver:
Also in the current code it looks like we create PopupPanel as a side-effect of the Popup.MainWindow property?? How does that even work?

fcantin-nventive:
> Why is the PopupPanel losing its TemplatedParent? Is this a recent regression?

The PopupPanel never had a TemplatedParent, it was created directly by code in Popup.iOS.cs. The Popup's original children did have a TemplatedParent. When PopupPanel is inserted as the direct child of Popup, and the original children are set as children of PopupPanel, they inherit its TemplatedParent (which is null).

This has been the case for a while. The reason we recently saw a regression, is that ItemsPresenter used to not care about losing its TemplatedParent - as soon as it received it, it assumed it was loaded and set itself as the child of the ItemsControl. However, this assumption is no longer true and we added a check in ItemsPresenter to make sure that it is both loaded and has an ItemsControl PemplatedParent. This introduced a bug, because in the case of ComboBox, these two conditions were never simultaneously true.

fcantin-nventive:
> Also in the current code it looks like we create PopupPanel as a side-effect of the Popup.MainWindow property?? How does that even work?

That is a mysterious behavior from the past! It seemed to be working OK, but I agree it's bizarre. I will push the changes we worked on together to move this creation to a more logical spot.

PR #319, opened 2018-10-31
Fix WebView raising too many events when cancelling navigation on iOS.

jeanplevesque:
## PR Type
- Bugfix

WebView's NavigationFailed and NavigationCompleted events are raised when we cancel a navigation on iOS.

## What is the new behavior?
WebView doesn't raise NavigationFailed and NavigationCompleted when navigation is cancelled.

## PR Checklist

https://nventive.visualstudio.com/Umbrella/_workitems/edit/135839

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/WebView/UnoWKWebView.iOS.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/WebView/UnoWKWebView.iOS.cs
@@ -23,6 +23,7 @@ namespace Windows.UI.Xaml.Controls
 	public partial class UnoWKWebView : WKWebView, INativeWebView
 	{
 		private WebView _parentWebView;
+		private bool _isCancelling;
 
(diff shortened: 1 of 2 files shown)


src/Uno.UI/UI/Xaml/Controls/WebView/UnoWKWebView.iOS.cs
```diff
@@ -293,6 +298,8 @@ private void OnError(WKWebView webView, WKNavigation navigation, NSError error)
 
 				_parentWebView.OnComplete(uri, false, status);
 			}
+
+			_isCancelling = false;
```
davidjohnoliver: put this in a finally block

davidjohnoliver: Never mind, I see that not much is called if _isCancelled is true

PR #324, opened 2018-11-02
[Android] Fix Bottom Padding Mask on an element inside a ScrollViewer

YGuerin:

## PR Type

- Bugfix

On Android, if a PaddingMask is set inside an element contained by a ScrollViewer and the ScrollViewer Content is smaller than the screen size, the layout just get some weird measurements.

## What is the new behavior?

Everything is well measured when a PaddingMask is set

## PR Checklist

## Other information

https://nventive.visualstudio.com/Umbrella/_workitems/edit/136188

### DIFF
--- a/src/Uno.UI/KeyboardRectProvider.Android.cs
+++ b/src/Uno.UI/KeyboardRectProvider.Android.cs
@@ -2,6 +2,7 @@
 using Android.App;
 using Android.Graphics;
 using Android.Graphics.Drawables;
+using Android.Util;
 using Android.Views;
 using Android.Widget;
 
@@ -65,17 +66,22 @@ namespace Uno.UI
 		/// </summary>
 		void ViewTreeObserver.IOnGlobalLayoutListener.OnGlobalLayout()
 		{
-			var screenSize = new Point();
-			_activity.WindowManager.DefaultDisplay.GetSize(screenSize);
+			var metrics = new DisplayMetrics();
+			_activity.WindowManager.DefaultDisplay.GetMetrics(metrics);
+
+			var realMetrics = new DisplayMetrics();
+			_activity.WindowManager.DefaultDisplay.GetRealMetrics(realMetrics);
 
 			var popupRect = new Rect();
 			_popupView.GetWindowVisibleDisplayFrame(popupRect);
 
+			// We are only interested in the height above the navigation bar. If the navigation bar visibility changes, it will also raise OnGlobalLayout.
+			// While the keyboard is visible the navigation bar will always also be visible (enabling the close keyboard button).
 			var keyboardRect = new Rect(
 				0,
-				popupRect.Bottom,
-				screenSize.X,
-				screenSize.Y
+				Math.Min(popupRect.Bottom, metrics.HeightPixels),
+				realMetrics.WidthPixels,
+				realMetrics.HeightPixels
 			);
 
 			_onKeyboardRectChanged?.Invoke(keyboardRect);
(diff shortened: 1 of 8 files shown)


src/Uno.UI/UI/Xaml/Controls/Border/Border.Android.cs
```diff
@@ -102,7 +102,7 @@ protected override void OnBackgroundChanged(DependencyPropertyChangedEventArgs e
 
         partial void OnPaddingChangedPartial(Thickness oldValue, Thickness newValue)
         {
-            UpdateBorder(true);
+            UpdateBorder(oldValue != newValue);
```
davidjohnoliver: Why do you need this? In general the DependencyProperty callback will only be raised if the DP is really different.

src/Uno.UI/UI/Xaml/ApplicationActivity.Android.cs
```diff
@@ -61,6 +61,9 @@ private void OnInputPaneVisibilityChanged(InputPane sender, InputPaneVisibilityE
 				// using either SoftInput.AdjustResize or SoftInput.AdjustPan.
 				args.EnsuredFocusedElementInView = true;
 			}
+
+			// This handle when the soft bar is showed / hidden on some devices and recalculate bounds
+			RaiseConfigurationChanges();
```
davidjohnoliver: This is an event handler for the keyboard showing or hiding. If it's being called when the bottom bar shows/hides, then there's a bug somewhere.

src/Uno.UI/UI/Xaml/Controls/Border/Border.Android.cs
YGuerin: It was just to only redraw instead of requesting layout but indeed, the callback is raised only if the DP really changed.
I will remove this.

src/Uno.UI/UI/Xaml/ApplicationActivity.Android.cs
YGuerin: This is raised because of ViewTreeObserver.IOnGlobalLayoutListener.OnGlobalLayout from KeyboardRectProvider.Android which measures the difference between the screen size and the popupView.
It actually triggers on some devices (such as LG G7 ThinQ) because the user can choose to hide or show the bottom soft bar.
I will try to fix that.

davidjohnoliver: I can see why OnGlobalLayout might be raised, but it seems like a bug if InputPane.OccludedRect is also changing.

YGuerin: Yes, but it's beacause of the keyboardRect calculated in OnGlobalLayout. The value changes when the soft bar is showed / hidden and then it changes the InputPane.OccludedRect value.

YGuerin: Should be fixed :)

src/Uno.UI/KeyboardRectProvider.Android.cs
```diff
@@ -65,20 +66,25 @@ public void Stop()
...
-				popupRect.Bottom,
-				screenSize.X,
-				screenSize.Y
+				Math.Min(popupRect.Bottom, metrics.HeightPixels),
+				realMetrics.WidthPixels,
+				realMetrics.HeightPixels
 			);
 
 			_onKeyboardRectChanged?.Invoke(keyboardRect);
 		}
```
davidjohnoliver: It seems like you're no longer calling RaiseConfigurationChanged() when OnGlobalLayout() is called, do you not need that any more?

YGuerin: No because OnSystemUiVisibilityChangeListener handles it

PR #335, opened 2018-11-08
workaround for DatePickerSelector not propagating coerced initial value

XiaotianNetlift:

## PR Type
> Bugfix

On iOS, when the `DatePicker.Date` value falls outside of the range defined by `MaxYear` and `MinYear`, the `UIDatePicker` will coerce the date to the closest value within the valid range. This is done without raising the `ValueChanged` event which is used to sync the date with the `DatePickerSelector`.

For example, with today as `MaxYear` and tomorrow as `Date`, the `UIDatePicker` will show today as selected. If the user clicks outside of the flyout or on <kbd>accept</kbd>, the `DatePicker.Date` is not updated to the selected value (today in this example).

## What is the new behavior?
Instead of letting `UIDatePicker` to coerce the date, we clamp the date between the valid range before passing it to `DatePickerSelector`. This ensures the `DatePicker` will receive the selected value.

## PR Checklist

## Other information
The value coercion happens without raising `UIDatePicker.ValueChanged`. This is why the coerced initial date is not propagated back to `DatePickerSelector`. However, when the user manually selects an invalid date, the coercion occurs before `ValueChanged` is raised, and we don't have a problem in this case.

> https://nventive.visualstudio.com/Umbrella/_workitems/edit/137137


src/Uno.UI/UI/Xaml/Controls/DatePicker/DatePickerFlyout.iOS.cs
```diff
@@ -31,9 +31,16 @@ private void InitializeContent()
 
 		private void DatePickerFlyout_Opening(object sender, EventArgs e)
 		{
+			// The date coerced by UIDatePicker doesn't propagate back to DatePickerSelector (#137137)
+			// unless the user selected an invalid date, which an ValueChanged will be raised after coercion.
```
MatFillion: I think the sentence "Which an ValueChanged" doesn't work.

src/Uno.UI/UI/Xaml/Controls/DatePicker/DatePickerFlyout.iOS.cs
```diff
@@ -31,9 +31,16 @@ private void InitializeContent()
...
+			// As a workaround, we clamp the date between the valid range.
```
MatFillion: I think we can remove "As a workaround" (or replace with "To address this" or something similar), if this is the best we can do it's not really a workaround :)

doc/ReleaseNotes/_ReleaseNotes.md
```diff
@@ -20,6 +20,7 @@
  * 132004 [Android] Window bounds incorrect for screen with rounded corners
  * #312 [Wasm] Text display was chopped on Wasm.
  * 135839 `WebView` No longer raises NavigationFailed and NavigationCompleted events when navigation is cancelled on iOS.
+ * 137137 [iOS] added workaround for `DatePickerSelector` not propagating coerced initial value
```
MatFillion: use the work item title here.

PR #343, opened 2018-11-09
[iOS] Update CommandBar measurement for both iPhones and iPads

YGuerin:

## PR Type

- Bugfix

On certain devices, the command bar has a bad height

## What is the new behavior?

The command bar has the good height on every device

## PR Checklist

## Other information

https://nventive.visualstudio.com/Umbrella/_workitems/edit/134573

### DIFF
--- a/src/Uno.UI/Controls/NativeCommandBarPresenter.iOS.cs
+++ b/src/Uno.UI/Controls/NativeCommandBarPresenter.iOS.cs
@@ -21,8 +21,10 @@ namespace Uno.UI.Controls
 		private readonly SerialDisposable _statusBarSubscription = new SerialDisposable();
 		private readonly SerialDisposable _orientationSubscription = new SerialDisposable();
 
+		// See https://ivomynttinen.com/blog/ios-design-guidelines#nav-bar
 		private static readonly double _defaultCommandBarHeight = 44;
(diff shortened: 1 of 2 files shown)


src/Uno.UI/Controls/NativeCommandBarPresenter.iOS.cs
```diff
@@ -69,33 +71,47 @@ void OrientationChanged(DisplayInformation displayInformation, object args)
...
+				{
+					return _defaultCommandBarHeight;
 				}
 			}
 			else
 			{
-				return _defaultCommandBarHeight;
+				if (UIDevice.CurrentDevice.CheckSystemVersion(12, 0))
+				{
+					return _iPad12CommandBarHeight;
```
davidjohnoliver: What about if it's a page with a Large Title, won't it be bigger?

Is there no way to extract the real height at runtime? Relying on hardcoded values like this seems to be brittle.

YGuerin: Large titles are not supported by the CommandBar.

And I've tried [this solution](https://stackoverflow.com/questions/7312059/programmatically-get-height-of-navigation-bar) to get height on runtime but the UINavigationBar is always null in our case.

PR #345, opened 2018-11-12
[iOS][TextBlock] Invalidate the full layout when the measure is invalidate

YGuerin:

## PR Type

- Bugfix

When you go back to an app set on background, TextBlock alignments are not always applied

## What is the new behavior?

When a TextBlock gets its measurements invalidated, it is fully invalidate (view and subviews).

## PR Checklist

## Other information

https://nventive.visualstudio.com/Umbrella/_workitems/edit/137945
https://nventive.visualstudio.com/Umbrella/_workitems/edit/138043

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.iOS.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.iOS.cs
@@ -84,7 +84,7 @@ namespace Windows.UI.Xaml.Controls
 		protected internal override void OnInvalidateMeasure()
 		{
 			base.OnInvalidateMeasure();
-			SetNeedsDisplay();
+			SetNeedsLayout();
 			_measureInvalidated = true;
 		}
 
--- a/doc/ReleaseNotes/_ReleaseNotes.md
+++ b/doc/ReleaseNotes/_ReleaseNotes.md
@@ -37,6 +37,8 @@
  * Ensure the SplitView templated parent is propagated properly for FindName
  * Fix infinite loop when parsing empty Attached Properties on macOS
  * 103116 [iOS] Navigating to a _second_ local html file with `WebView` doesn't work.
+ * 137945 [iOS 12] Some UI elements are misaligned after device goes from landscape back to portrait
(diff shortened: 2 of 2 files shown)


YGuerin:
Note that I wasn't able to reproduce the first issue since it's random but it looks like it is highly related to the second one.

src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.iOS.cs
```diff
@@ -84,7 +84,7 @@ private CGRect GetDrawRect(CGRect rect)
 		protected internal override void OnInvalidateMeasure()
 		{
 			base.OnInvalidateMeasure();
-			SetNeedsDisplay();
+			SetNeedsLayout();
```
davidjohnoliver: I'm very puzzled as to how this could have any effect. OnInvalidateMeasure is called from UIElement.InvalidateMeasure();  InvalidateMeasure() is already calling SetNeedsLayout().

YGuerin: It's because it's not the same SetNeedsLayout() method.
UIElement.InvalidateMeasure(); calls UIView.SetNeedsLayout() and this one calls FrameworkElement.SetNeedsLayout() (which one call UIView.SetNeedsLayout if _inLayoutSubviews is false).
Should I modify the IFrameworkElementHelper.InvalidateMeasure instead of TextBlock.OnInvalidateMeasure ?

YGuerin: And do we need SetNeedsDisplay in that case ? SetNeedsLayout will request for a redraw, right ?

davidjohnoliver: It's a virtual method, it's the same method.

YGuerin: Since it's too random to reproduce, I will just close the PR.

PR #357, opened 2018-11-15
[iOS] Implements custom navigation bar

YGuerin:

## PR Type

- Bugfix

The command bar has not the good on every device

## What is the new behavior?

The command bar height is set to the device one

## PR Checklist

## Other information

https://nventive.visualstudio.com/Umbrella/_workitems/edit/134573

### DIFF
--- a/src/Uno.UI/Controls/CommandBarRenderer.iOS.cs
+++ b/src/Uno.UI/Controls/CommandBarRenderer.iOS.cs
@@ -47,6 +47,15 @@ namespace Uno.UI.Controls
 				new[] { BackButtonForegroundProperty },
 				new[] { BackButtonIconProperty }
 			);
+
+			if (Native is UnoNavigationBar unoNavigationBar)
+			{
+				unoNavigationBar.SizeChanged += Invalidate;
+
+				yield return Disposable.Create(() =>
+					unoNavigationBar.SizeChanged -= Invalidate
+				);
+			}
 		}
 
 		protected override void Render()
@@ -111,6 +120,10 @@ namespace Uno.UI.Controls
 				: null;
 			Native.BackIndicatorImage = backButtonIcon;
 			Native.BackIndicatorTransitionMaskImage = backButtonIcon;
+
+			Element.Presenter.Height = Native.Hidden
+				? 0
+				: Native.Frame.Size.Height;
 		}
 	}
 }
--- a/src/Uno.UI/UI/Xaml/Controls/CommandBar/CommandBar.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/CommandBar/CommandBar.cs
@@ -8,6 +8,12 @@ using Windows.Foundation.Collections;
 using Windows.UI.Xaml.Markup;
 using Windows.UI.Xaml.Controls.Primitives;
 using Uno.Logging;
+using Uno.UI.Controls;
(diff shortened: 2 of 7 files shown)


src/Uno.UI/Controls/CommandBarRenderer.iOS.cs
```diff
@@ -111,7 +121,12 @@ protected override void Render()
 				: null;
 			Native.BackIndicatorImage = backButtonIcon;
 			Native.BackIndicatorTransitionMaskImage = backButtonIcon;
+
+
```
davidjohnoliver: extra space

src/Uno.UI/UI/Xaml/Controls/CommandBar/CommandBar.cs
```diff
@@ -8,6 +8,8 @@
 using Windows.UI.Xaml.Markup;
 using Windows.UI.Xaml.Controls.Primitives;
 using Uno.Logging;
+using UIKit;
```
davidjohnoliver: This won't compile on other platforms. You'll need to add conditionals. (FindFirstChild is in the Uno.UI namespace on Android, I think on other platforms it's a normal method on FrameworkElement.)

src/Uno.UI/UI/Xaml/Controls/CommandBar/CommandBar.cs
```diff
@@ -36,6 +38,8 @@ public CommandBar()
 			CommandBarTemplateSettings = new CommandBarTemplateSettings(this);
 		}
 
+		internal NativeCommandBarPresenter _presenter { get; set; }
```
davidjohnoliver: Presenter

src/Uno.UWP/UI/ViewManagement/ApplicationView.iOS.cs
```diff
@@ -17,17 +17,16 @@ internal void SetVisibleBounds(UIKit.UIWindow keyWindow, Foundation.Rect windowB
...
 			// So we need to update the top inset depending of the status bar visibilty on other devices
-			if (UIDevice.CurrentDevice.UserInterfaceIdiom != UIUserInterfaceIdiom.Phone)
-			{
-				inset.Top = UIApplication.SharedApplication.StatusBarHidden
+			var statusBarHeight = UIApplication.SharedApplication.StatusBarHidden
 					? 0
 					: UIApplication.SharedApplication.StatusBarFrame.Size.Height;
-			}
+
+			inset.Top = (nfloat)Math.Max(inset.Top, statusBarHeight);
```
davidjohnoliver: different bug?

YGuerin: On iOS 11 and iPad, the status bar is not included in the safearea so we will set the visible bounds under the status bar for these devices. It's not the same one but related. I've noticed it when I was trying my stuff

src/Uno.UI/UI/Xaml/Controls/CommandBar/CommandBar.cs
YGuerin: Actually the NativeCommandBarPresenter doesn't seems to exist on other platform. I just set it for Android and iOS. We can see that point tomorrow if needed

PR #380, opened 2018-11-21
fixed list view scrolls up when tapping an item at the bottom of screen

XiaotianNetlift:
## PR Type
> Bugfix

On android, `ListView` will scroll up when you tap an item at the bottom of the screen. This also happens if you just hold the press, and the scrolling will occur before you release.

## What is the new behavior?
The above described behavior should only happens when the soft input panel (keyboard) is shown and you focus an item that is blocked by the keyboard.

## PR Checklist

## Other information
This issue started happening with 123bb68362e72c6c331991135fa471a163193f87 from #324

https://nventive.visualstudio.com/Umbrella/_workitems/edit/138099/
https://nventive.visualstudio.com/Umbrella/_workitems/edit/138463/

### DIFF
--- a/src/Uno.UI/KeyboardRectProvider.Android.cs
+++ b/src/Uno.UI/KeyboardRectProvider.Android.cs
@@ -2,6 +2,7 @@
 using Android.App;
 using Android.Graphics;
 using Android.Graphics.Drawables;
+using Android.Util;
 using Android.Views;
 using Android.Widget;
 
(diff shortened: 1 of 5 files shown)


src/Uno.UI/KeyboardRectProvider.Android.cs
```diff
@@ -66,25 +68,28 @@ public void Stop()
...
 		{
-			var metrics = new DisplayMetrics();
-			_activity.WindowManager.DefaultDisplay.GetMetrics(metrics);
+			var realMetrics = Get<DisplayMetrics>(_activity.WindowManager.DefaultDisplay.GetRealMetrics);
+			var displayRect = Get<Rect>(_activity.WindowManager.DefaultDisplay.GetRectSize);
+			var usableRect = Get<Rect>(_popupView.GetWindowVisibleDisplayFrame);
 
-			var realMetrics = new DisplayMetrics();
-			_activity.WindowManager.DefaultDisplay.GetRealMetrics(realMetrics);
+			// we assume that the keyboard and the navigation bar always occupy the bottom area, with the keyboard being above the navigation bar
```
MatFillion: This comment made me think about the floating keyboards on samsung phones, I wonder how it behaves.

XiaotianNetlift: It looks like an overlay to me. Given that "You can move it and re-size it with intuitive drag and drop", I doubt it would be taken into account for `DefaultDisplay.GetRectSize`.

*Needs to be confirmed:*
> Likely we will get an empty rect for keyboard here, resulting in `InputPaneVisible` being false. This would in turn disable the "scroll into focused element" behavior, 

... which should be the expected behavior since that the user is in control of the overlay keyboard.

XiaotianNetlift: Confirmed:
screencapture-file-c-users-xiaotian-gu-downloads-asd-html-2018-11-22-12_41_40

It is strange that on samsung S8+ with android 7.0, `displayRect` have the space for nav bar reserved even when it is hidden. However, `occupiedRect` still stands correct.

PR #418, opened 2018-12-05
Prevent itemscontrol items to be reloaded too many times

### LINKED ISSUE #382: ItemsControl: template recreated multiple times
## I'm submitting a...
Bug report 

## Current behavior
When using an ItemsControl, item templates are created multiple times. 
Resources used by extra items seem to be not properly released.

ListView seems to be affected as well.

## Expected behavior
Item template should be instantiated only one time.
If recreated, resources need to be properly released.

## Environment
Nuget Package: 
Uno.UI 1.42.0-dev.392

Affected platform(s):
- [X] iOS
- [X] Android
- [ ] WebAssembly
- [ ] Windows
- [ ] Build tasks

Visual Studio
- [X] 2017 (version: )
- [X] 2017 Preview (version: )
- [ ] for Mac (version: )

Relevant plugins
- [ ] Resharper (version: )

Internal issue: https://nventive.visualstudio.com/Umbrella/_workitems/edit/139686

baptiste-nv:
GitHub Issue (If applicable): Fixes #382 

## PR Type
Bugfix

When using an ItemsControl, item templates are created multiple times.

## What is the new behavior?
Item container is reused if possible, which limit reload.

## PR Checklist

## Other information
None

https://nventive.visualstudio.com/Umbrella/_workitems/edit/139686

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/ItemsControl/ItemsControl.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/ItemsControl/ItemsControl.cs
@@ -593,11 +593,9 @@ namespace Windows.UI.Xaml.Controls
 			int remainder = GetDisplayGroupCount(section) % itemsPerLine;
 			return remainder == 0 ? itemsPerLine : remainder;
 		}
-
+		
 		protected virtual void OnItemsSourceChanged(DependencyPropertyChangedEventArgs e)
 		{
-			Items?.Clear();
-
 			IsGrouping = (e.NewValue as ICollectionView)?.CollectionGroups != null;
 			SetNeedsUpdateItems();
 			ObserveCollectionChanged();
@@ -865,7 +863,7 @@ namespace Windows.UI.Xaml.Controls
 						PrepareContainerForIndex(container, index);
 						return container;
 					});
-
+				
 				var results = ItemsPanelRoot.Children.UpdateWithResults(containers.OfType<View>(), comparer: new ViewComparer());
 
 				foreach (var removed in results.Removed)
@@ -1041,9 +1039,16 @@ namespace Windows.UI.Xaml.Controls
 		{
 			var item = ItemFromIndex(index);
 
-			var container = IsItemItsOwnContainerOverride(item)
-				? item as DependencyObject
-				: GetContainerForItemOverride();
+			var container = default(DependencyObject);
+
+			if (IsItemItsOwnContainerOverride(item))
+			{
+				container = item as DependencyObject;
+			}
+			else
+			{
+				container = ItemsPanelRoot?.Children?.OfType<ContentPresenter>()?.FirstOrDefault(c => c.Content?.Equals(item) ?? false) ?? GetContainerForItemOverride();
+			}
 
 			container.SetValue(ItemsControlForItemContainerProperty, new WeakReference<ItemsControl>(this));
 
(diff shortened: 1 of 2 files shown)


src/Uno.UI/UI/Xaml/Controls/ItemsControl/ItemsControl.cs
```diff
@@ -1041,9 +1039,16 @@ internal DependencyObject GetContainerForIndex(int index)
...
-				: GetContainerForItemOverride();
+			var container = default(DependencyObject);
+
+			if (IsItemItsOwnContainerOverride(item))
+			{
+				container = item as DependencyObject;
+			}
+			else
+			{
+				container = ItemsPanelRoot?.Children?.OfType<ContentPresenter>()?.FirstOrDefault(c => c.Content?.Equals(item) ?? false) ?? GetContainerForItemOverride();
```
jeromelaban:
```suggestion
				container = ItemsPanelRoot?.Children
				  ?.OfType<ContentPresenter>()
				  ?.FirstOrDefault(c => object.ReferenceEquals(c.Content, item) ?? false) 
				  ?? GetContainerForItemOverride();
```

jeromelaban: We need to compare references, not using the Equals method (for which an Type may have the `Equals` overridden)

src/Uno.UI/UI/Xaml/Controls/ItemsControl/ItemsControl.cs
```diff
@@ -593,11 +593,9 @@ internal int GetItemsOnLastLine(int section, int itemsPerLine)
 			int remainder = GetDisplayGroupCount(section) % itemsPerLine;
 			return remainder == 0 ? itemsPerLine : remainder;
 		}
-
+		
 		protected virtual void OnItemsSourceChanged(DependencyPropertyChangedEventArgs e)
 		{
-			Items?.Clear();
```
jeromelaban: We may want to move this to the else of the outermost "if" in `UpdateItems` or clear the items only if the source becomes null.

src/Uno.UI/UI/Xaml/Controls/ItemsControl/ItemsControl.cs
davidjohnoliver: So... an ItemsControl will recycle containers now? That's a pretty sharp departure from UWP. And normally the template shouldn't be recreated, that suggests a bug regarding the FrameworkTemplate pooling.

Also to add to Jerome's comment, what's going to happen if the same item (by ref) occurs twice in the ItemsSource?

baptiste-nv: I will cancel that PR until we get a clearer view  of what should be done. Anyway, there is a perf issue to fix here ;)

src/Uno.UI/UI/Xaml/Controls/ItemsControl/ItemsControl.cs
baptiste-nv: There is no sync between Items and ItemsSource anywhere but here. And this call causes UpdateItems to be called twice (once due to VectorChanged event on Items, once in ItemSourceChanged). 
In any case, GetItems method (which is used to select the correct item source) will use ItemsSource first, and if empty (not an ICollection in fact), it will use Items.

baptiste-nv:
PR closed as it will require more investigation

PR #424, opened 2018-12-06
[iOS] Fix app freezes if an Image without width is set in the CommandBar

YGuerin:

## PR Type

- Bugfix

The app freezes if the Content of a CommandBar is an Image without specified width

## What is the new behavior?

The app no longer freezes

## PR Checklist

## Other information

https://nventive.visualstudio.com/Umbrella/_workitems/edit/138537/

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/Image/Image.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Image/Image.cs
@@ -293,8 +293,10 @@ namespace Windows.UI.Xaml.Controls
 						switch (ImageControl.Stretch)
 						{
 							case Stretch.Uniform:
-								//If sourceSize is empty, aspect ratio is undefined so we return 0
-								desiredSize.Height = sourceSize == default(Size) ? 0 : (knownWidth / aspectRatio);
+								// If sourceSize is empty, aspect ratio is undefined so we return 0.
+								// Since apsect ratio can have a lot of decimal, iOS ceils Image size to 0.5 if it's not a precise size (like 111.111111111)
+								// so the desiredSize will never match the actual size causing an infinite measuring and can freeze the app
+								desiredSize.Height = sourceSize == default(Size) ? 0 : Math.Ceiling((knownWidth / aspectRatio) * 2) / 2;
 								break;
 							case Stretch.None:
 								desiredSize.Height = sourceSize.Height;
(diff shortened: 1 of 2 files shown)


src/Uno.UI/UI/Xaml/IFrameworkElementImplementation.iOS.tt
```diff
@@ -450,7 +450,8 @@ namespace <#= mixin.NamespaceName #>
 					SizeChanged?.Invoke(this, new SizeChangedEventArgs(previousSize, _actualSize));
 
 #if <#= mixin.IsFrameworkElement #>
-					if (Superview != null && !(Superview is DependencyObject))
+					// Require measurement from CommandBar elements just cause apps to crash
```
MatFillion: Does the comment mean that we should never remeasure UINavigationBar ?  (I figured it out because of the diff but I'm not sure it's clear. Do we know why?

davidjohnoliver: Apart from being a monstrous hack, this will cause a regression when the CommandBar content is text that's too long to display in portrait.

YGuerin: I've updated my PR. Please see new modifications

src/Uno.UI/UI/Xaml/Controls/Image/Image.cs
```diff
@@ -351,8 +355,10 @@ protected override Size MeasureOverride(Size availableSize)
 						case Stretch.Uniform:
 							var desiredSize = new Size();
 							var aspectRatio = sourceSize.Width / sourceSize.Height;
-							desiredSize.Width = Math.Min(knownWidth, knownHeight * aspectRatio);
-							desiredSize.Height = Math.Min(knownHeight, knownWidth / aspectRatio);
+							// Since apsect ratio can have a lot of decimal, iOS ceils Image size to 0.5 if it's not a precise size (like 111.111111111)
```
MatFillion: Ceiling the aspect ratio to .5 seems rather radical, we're sure they're actually doing this? I mean 16:9 (1.7777777) and 16:10 (1.60000) would have the same value.

YGuerin: It ceils the whole measure, not the AspectRatio

PR #479, opened 2019-01-07
[Android][FlipView] Check if items have changed to notify the adapter

YGuerin:

## PR Type

- Bugfix

Changing selected index with tabs on a FlipView cause it to reload data every time. The SwipeGesture does not refresh the data.

## What is the new behavior?

The data are not refreshed every time when a new tab is selected

## PR Checklist

## Other information

https://nventive.visualstudio.com/Umbrella/_workitems/edit/143176

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/FlipView/FlipView.iOS.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/FlipView/FlipView.iOS.cs
@@ -63,13 +63,12 @@ namespace Windows.UI.Xaml.Controls
 			}
 		}
 
-		protected override void UpdateItems()
+		protected override bool UpdateItems()
 		{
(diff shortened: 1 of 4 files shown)


src/Uno.UI/UI/Xaml/Controls/FlipView/FlipView.iOS.cs
```diff
@@ -87,6 +86,8 @@ protected override void UpdateItems()
 			}
 			base.UpdateItems();
```
MatFillion: the return value of base.UpdateItems() wouldn't be appropriate?

YGuerin: It would be, my bad

PR #484, opened 2019-01-08
ItemsStackPanel wasn't propagating correctly measure/arrange phases after invalidation

carldebilly:
## Bugfix
Controls calling `.InvalidateMeasure()` or `.InvalidateArrange()` when hosted in a _virtualized list view_ (ItemsControl using a `ItemsStackPanel`) were not _remeasured_ nor _rearranged_.

The `Measure` phase of an `ItemsStackPanel` is not actually measuring its children. The method is
used to calculate the virtualization and check if more items are required. Only newly added items
are getting measured (and arranged).

This faulty behavior is preventing images from stretching correctly. Since
their stretch values are calculated once the image is loaded (because we
need to know the physical dimension of the image), an `InvalidateMeasure`
were issued on the `loaded` event. Hosting an image in a virtualized list were
preventing correct stretching of images, because the `Arrange` method were
never called after the invalidation.

## What is the new behavior?
New behaviors are added in this PR:
1. On re-`Measure`, **If the size is changed**, the list is cleared and recreated.
   THIS IS OBVIOUSLY A HACK.
2. On re-`Measure`, **If the size DID NOT changed**, the children items are simply remeasured.
3. On initial `Arrange` (when an item is added to the list), the calculated `Rect` is persisted.
4. On re-`Arrange`, all children items are being rearranged with the same size.

## PR Checklist

- [X] Tests for the changes have been added (for bug fixes / features) (if applicable)

## Internal Issue (If applicable):
* <https://nventive.visualstudio.com/Umbrella/_workitems/edit/143596> Wasm Images stretching is incorrect
* <https://nventive.visualstudio.com/Umbrella/_workitems/edit/143595> Wasm ListView Resizing is not working


carldebilly:
@davidjohnoliver Your input on this PR would be important.

davidjohnoliver:
> @davidjohnoliver Your input on this PR would be important.

Agreed, it's a hack. :p

PR #490, opened 2019-01-11
Handle the load events from managed code to reduce interop

dr1rrb:
## PR Type
 - Other: Performance improvement on WASM, reduces interop

On WASM `Loading`, `Loaded` and `Unloaded` events are raised by the Javascript then marshaled to the managed code.

## What is the new behavior?
They are now raised directly by the managed code

## PR Checklist

## Other information
Old behavior can be restored by setting to `false` the feature flag `FeatureConfiguration.FrameworkElement.WasmUseManagedLoadedUnloaded`

https://nventive.visualstudio.com/DefaultCollection/Umbrella/_workitems/edit/144039


dr1rrb:
> 
> 
> Just to be sure (it's not obvious in the code):
> Usually the _Loading_ is called before actually connect the element to the DOM and the _Loaded_ is after. Did you keep this behavior?

Yes, I invoke the native `Uno.UI.Xaml.WindowManagerInterop.AddView` (which will insert the native element in the DOM) between the `Loading` and the `Loaded`. As it runs synchronously, the behavior is correct.

PR #517, opened 2019-01-22
Adjust linker definitions

jeromelaban:
## PR Type
- Bugfix

## What is the new behavior?
- Wasm Animations depending on `ManagedObject` typescript API when the linker is enabled.
- Adjust linker file visibility
- Adjust CoreDispatcher.DispatcherCallback signature for AOT builds
- Adjust Wasm template main method return value

### DIFF
--- a/src/Uno.UI.Wasm/ts/Interop/ManagedObject.ts
+++ b/src/Uno.UI.Wasm/ts/Interop/ManagedObject.ts
@@ -1,25 +1,10 @@
 ﻿namespace Uno.Foundation.Interop {
 	export class ManagedObject {
 		private static assembly: UI.Interop.IMonoAssemblyHandle;
-		private static dispatchMethod: UI.Interop.IMonoMethodHandle;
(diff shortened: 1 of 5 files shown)


src/Uno.UI.Wasm/ts/Interop/ManagedObject.ts
```diff
@@ -1,41 +1,18 @@
 ﻿namespace Uno.Foundation.Interop {
 	export class ManagedObject {
 		private static assembly: UI.Interop.IMonoAssemblyHandle;
-		private static dispatchMethod: UI.Interop.IMonoMethodHandle;
+		private static dispatchMethod: any;
```
carldebilly: <http://chandermani.com/2017/typescript-anti-pattern-the-any-type/>

jeromelaban: It seems that exports are not made for delegate-alike, but for full objects. I changed it to be a signature instead.

PR #521, opened 2019-01-23
Update editorconfig file

jeromelaban:
## PR Type
- Code style update (formatting)

## What is the new behavior?
Adjust the location of the `.editorconfig` file, add other formatting rules.


.editorconfig
```diff
@@ -1,30 +0,0 @@
...
-tab_width = 2
-indent_size = 2
-
-[*.js]
-indent_style = space
-
-[*.yml]
-indent_style = space
-
-[*.md]
```
carldebilly: You are removing this from documentation...

src/.editorconfig
```diff
@@ -0,0 +1,223 @@
...
+csharp_prefer_braces = true:error
+
+# http://kent-boogaart.com/blog/editorconfig-reference-for-c-developers#csharp_prefer_simple_default_expression
+csharp_prefer_simple_default_expression = true:error
+
+# http://kent-boogaart.com/blog/editorconfig-reference-for-c-developers#csharp_preserve_single_line_blocks
+csharp_preserve_single_line_blocks = true
+
+# http://kent-boogaart.com/blog/editorconfig-reference-for-c-developers#csharp_preserve_single_line_statements
+csharp_preserve_single_line_statements = false
```
dr1rrb: I would not enforce somethings that changes the current code

src/.editorconfig
```diff
@@ -0,0 +1,223 @@
...
+csharp_space_between_square_brackets = false
+
+# http://kent-boogaart.com/blog/editorconfig-reference-for-c-developers#csharp_style_conditional_delegate_call
+csharp_style_conditional_delegate_call = true:error
+
+# http://kent-boogaart.com/blog/editorconfig-reference-for-c-developers#csharp_style_expression_bodied_accessors
+csharp_style_expression_bodied_accessors = true:error
+
+# http://kent-boogaart.com/blog/editorconfig-reference-for-c-developers#csharp_style_expression_bodied_constructors
+csharp_style_expression_bodied_constructors = false:error
```
dr1rrb: Does this will fail the build ?

src/.editorconfig
```diff
@@ -0,0 +1,223 @@
...
+dotnet_style_null_propagation = true:error
+
+# http://kent-boogaart.com/blog/editorconfig-reference-for-c-developers#dotnet_style_object_initializer
+dotnet_style_object_initializer = true:error
+
+# http://kent-boogaart.com/blog/editorconfig-reference-for-c-developers#dotnet_style_predefined_type_for_locals_parameters_members
+dotnet_style_predefined_type_for_locals_parameters_members = true:error
+
+# http://kent-boogaart.com/blog/editorconfig-reference-for-c-developers#dotnet_style_predefined_type_for_member_access
+dotnet_style_predefined_type_for_member_access = false:error
```
dr1rrb: According to our current rules, this should be true
```suggestion
dotnet_style_predefined_type_for_member_access = true:error
```

src/.editorconfig
jeromelaban: No, but I made it a suggestion instead.

src/.editorconfig
jeromelaban: True.

PR #565, opened 2019-02-05
Fix animations always using top-left center when using `Forever` repeat mode on Android

dr1rrb:
Fix animations always using top-left center when using `Forever` repeat mode on Android (and fix invalid rounding).

## PR Type
- Bugfix

- When using the repeat mode `Forever` with `RenderTransform = (.5, .5)`, the animation use the right center only for the first time, then it will use `(0, 0)` as center
- If the animated control has an actual size like `(41.2, 34.8)` the center might be offseted by 1px compared to Windows

## What is the new behavior?
- Animations always use the right center
- Center is the same as Windows

## PR Checklist

https://nventive.visualstudio.com/Umbrella/_workitems/edit/146261
https://nventive.visualstudio.com/Umbrella/_workitems/edit/146326

### DIFF
--- a/src/Uno.UI/Extensions/ViewHelper.Android.cs
+++ b/src/Uno.UI/Extensions/ViewHelper.Android.cs
@@ -168,6 +168,34 @@ namespace Uno.UI
 			return (int)((value * FontScale) + .5f);
 		}
 
+		/// <summary>
+		/// Gets the physical representation of the provided logical pixels, 
+		/// for the <see cref="View.PivotX"/> and <see cref="View.PivotY"/> of a View (cf. Remarks)
+		/// </summary>
+		/// <remarks>Compared to <see cref="LogicalToPhysicalPixels"/>, this won't apply any rounding and returns a float.</remarks>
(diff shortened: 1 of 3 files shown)


src/Uno.UI/Extensions/ViewHelper.Android.cs
```diff
@@ -168,6 +168,32 @@ public static float LogicalToPhysicalFontPixels(float value)
 			return (int)((value * FontScale) + .5f);
 		}
 
+		/// <summary>
+		/// Gets the physical representation of the provided logical pixels.
+		/// </summary>
+		/// <param name="value">The logical value</param>
+		/// <returns>The pixels value</returns>
+		[MethodImpl(MethodImplOptions.AggressiveInlining)]
```
MatFillion: what is the difference between this one and LogicalToPhysicalPixels ? Should it be noted in the summary/remarks?

dr1rrb: Right, updated!

PR #568, opened 2019-02-06
Timepicker - minute increment (ios/android)

AlexTrepanier:
## PR Type
Added time increment to ios and android

## PR Checklist
- [x] Tested code on ios, android and uwp
- [x] Docs have been added/updated [TimePicker.md](https://github.com/nventive/Uno/blob/master/src/Uno.UI/UI/Xaml/Controls/TimePicker/TimePicker.md)

143859

### DIFF
--- a/src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/TimePicker/Sample1.xaml
+++ b/src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/TimePicker/Sample1.xaml
@@ -11,28 +11,21 @@
       xmlns:controls="using:Uno.UI.Samples.Controls"
       mc:Ignorable="xamarin android ios native_ios native_android">
 
-    <controls:SampleControl SampleDescription="Sample1">
-        <controls:SampleControl.SampleContent>
-            <DataTemplate>
-                <Grid Background="White">
-                    <Grid.RowDefinitions>
-                        <RowDefinition Height="Auto" />
-                        <RowDefinition Height="*" />
-                    </Grid.RowDefinitions>
-                    	<Button Content="Set To Current Time"
-							Command="{Binding SetToCurrentTime}" Margin="10" HorizontalAlignment="Center" Visibility="Collapsed"/>
-                    <TimePicker Grid.Row="1"
+	<controls:SampleControl SampleDescription="style:default - clock:24h - increment:1 - time:now">
+		<controls:SampleControl.SampleContent>
+			<DataTemplate>
+				<Grid Background="White">
+					<TimePicker Grid.Row="1"
                                 Margin="15,10"
                                 ios:FlyoutPlacement="Full"
                                 HorizontalAlignment="Center"
                                 VerticalAlignment="Top"
-                                Time="{Binding Time, Mode=TwoWay}"
                                 MinuteIncrement="1"
                                 ClockIdentifier="24HourClock">
-                    </TimePicker>
-                </Grid>
-            </DataTemplate>
-        </controls:SampleControl.SampleContent>
-    </controls:SampleControl>
+					</TimePicker>
+				</Grid>
+			</DataTemplate>
+		</controls:SampleControl.SampleContent>
+	</controls:SampleControl>
 
 </Page>
--- a/src/Uno.UI/UI/Xaml/Controls/Flyout/FlyoutBase.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Flyout/FlyoutBase.cs
@@ -73,10 +73,7 @@ namespace Windows.UI.Xaml.Controls.Primitives
 			Closed?.Invoke(this, EventArgs.Empty);
 		}
 
-		internal protected virtual void Close()
-		{
-			throw new NotImplementedException("This method has no base class implementation and must be overridden in a derived class.");
-		}
+		internal protected virtual void Close(){}
 
 		public void ShowAt(FrameworkElement placementTarget)
 		{
@@ -103,12 +100,11 @@ namespace Windows.UI.Xaml.Controls.Primitives
 
 		internal protected virtual void Open()
 		{
-			throw new NotImplementedException("This method has no base class implementation and must be overridden in a derived class.");
 		}
 
 		protected virtual Control CreatePresenter()
 		{
-			throw new NotImplementedException("This method has no base class implementation and must be overridden in a derived class.");
+			return null;
 		}
 	}
 }
--- a/src/Uno.UI/UI/Xaml/Controls/TimePicker/TimePicker.md
+++ b/src/Uno.UI/UI/Xaml/Controls/TimePicker/TimePicker.md
@@ -4,7 +4,11 @@
 
 TimePicker is use to select a specific time of the day in hour and minute (AM/PM).
 
-Time button open the time picker flyout popup. 
+Button showing time open the time picker popup. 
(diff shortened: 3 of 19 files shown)


src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/TimePicker/Sample1.xaml
```diff
@@ -11,28 +11,21 @@
...
+			<DataTemplate>
+				<Grid Background="White">
+					<TimePicker Grid.Row="1"
                                 Margin="15,10"
                                 ios:FlyoutPlacement="Full"
                                 HorizontalAlignment="Center"
                                 VerticalAlignment="Top"
-                                Time="{Binding Time, Mode=TwoWay}"
-                                MinuteIncrement="1"
+                                MinuteIncrement="13"
```
MatFillion: Shouldn't Sample1 have a minute increment of 1 or just no specifications at all ?  (be more like default)

MatFillion: SampleDescription says 1

src/Uno.UI/UI/Xaml/Controls/Flyout/FlyoutBase.cs
```diff
@@ -73,10 +73,7 @@ internal void Hide(bool canCancel)
 			Closed?.Invoke(this, EventArgs.Empty);
 		}
 
-		internal protected virtual void Close()
```
MatFillion: Why is this necessary?

src/Uno.UI/UI/Xaml/Controls/TimePicker/TimePicker.md
```diff
@@ -23,6 +27,7 @@ Native time picker is wrapped in the flyout.
 Timepicker flyout appear centered to screen.
 You can change the flyout button by copying and modifying TimePickerFlyoutButtonStyle.
 You can change the flyout button by copying and modifying TimePickerFlyoutPresenterStyle.
+When time picker use minute increment more than 1 minutes, his style default to "spinner mode"
```
MatFillion:
```suggestion
If `MinuteIncrement` is not 1, `TimePicker` will show in "spinner mode"
```

src/Uno.UI/UI/Xaml/Controls/TimePicker/TimePickerFlyout.Android.cs
```diff
@@ -18,16 +18,15 @@ internal protected override void Open()
...
-			if (Build.VERSION.SdkInt == BuildVersionCodes.M &&
-			    Build.Manufacturer.ToLower().IndexOf("samsung") >= 0)
-			{
-				ShowUsingAlertDialog();
-			}
-			else
-			{
-				ShowUsingTimePickerDialog();
-			}
+			//Minute increments are not supported with clock and Samsung devices with Marshmellow
```
MatFillion:
```suggestion
			//Minute increments are not supported with clock and Samsung devices with marshmallow (Android 6.0)
```

src/Uno.UI/UI/Xaml/Controls/TimePicker/TimePickerFlyout.Android.cs
```diff
@@ -42,65 +41,44 @@ private void SaveTime(int hourOfDay, int minutes)
...
 
 		private void OnDismiss(object sender, EventArgs e)
 		{
 			Hide(canCancel: false);
 		}
 
-		partial void OnTimeChangedPartialNative(TimeSpan oldTime, TimeSpan newTime){}
+		partial void OnTimeChangedPartialNative(TimeSpan oldTime, TimeSpan newTime) { }
+
+		public class OnSetTimeListener : Java.Lang.Object, IOnTimeSetListener
```
MatFillion: Instead of creating a new class, could you implement IOnTimeSetListener on TimePickerFlyout directly?  (and pass `this` to the dialog).

MatFillion: I'm not entirely sure of the best practice. @jeromelaban

src/Uno.UI/UI/Xaml/Controls/TimePicker/TimeSpanExtensions.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+
+namespace Windows.UI.Xaml.Controls
+{
+	internal static class TimeSpanExtensions
+	{
+		internal static TimeSpan RoundToInterval(this TimeSpan time, int interval)
```
MatFillion: Add /// summary.

MatFillion:
```suggestion
		internal static TimeSpan RoundToInterval(this TimeSpan time, int minuteInterval)
```
a way to clarify that the unit is minute.

carldebilly: The description should specify the unit for the `interval` parameter.

src/Uno.UI/UI/Xaml/Controls/Flyout/FlyoutBase.cs
AlexTrepanier: need to be implemented in derived classes ..

I just removed the useless and dangerous 
throw new NotImplementedException("This method has no base class implementation and must be overridden in a derived class.");

since base.close() would crash

note that i havent cleaned Open() and CreatePresenter either.. but they should be cleaned..

flyoutbase should actually be abstract and these should be abstract method or virtual but empty. 

here where close() is used.. 
internal void Hide(bool canCancel)
		{
			if (!_isOpen)
			{
				return;
			}

			if (canCancel)
			{
				var closing = new FlyoutBaseClosingEventArgs();
				Closing?.Invoke(this, closing);
				if (closing.Cancel)
				{
					return;
				}
			}

			Close();
			_isOpen = false;
			Closed?.Invoke(this, EventArgs.Empty);
		}

src/Uno.UI/UI/Xaml/Controls/TimePicker/TimePickerFlyout.Android.cs
AlexTrepanier: class OnSetTimeListener need to implement Java.Lang.Object interface so i think it's better as internal class like now. 

public class OnSetTimeListener : Java.Lang.Object, IOnTimeSetListener

src/Uno.UI/UI/Xaml/Controls/TimePicker/TimeSpanExtensions.cs
AlexTrepanier: summary to the class or to the method.?

i can add but i don't see why for both the class and the method the name is self explanatory. 

also i'm moving the method to Uno.UI.Extensions.TimeSpanExtensions

AlexTrepanier: yes good point, i'll rename the method to RoundToNextMinuteInterval

MatFillion: on the method.

PR #594, opened 2019-02-13
[Wasm] Improve arrange and transform performance

jeromelaban:
## PR Type
- Performance

The current arrange implementation uses a lot of string formatting to send data to Javascript.

## What is the new behavior?
The new implementation uses TSBindings to improve the performance, and here are some results for 1000 arrange calls, using the interpreter:
- Before 590ms
- After, 360ms

## PR Checklist

### DIFF
--- a/src/Uno.UI.Wasm/WasmScripts/Uno.UI.js
+++ b/src/Uno.UI.Wasm/WasmScripts/Uno.UI.js
@@ -676,7 +676,7 @@ var Uno;
                     throw `Element id ${params.HtmlId} not found.`;
                 }
                 var style = htmlElement.style;
-                style.transform = `scale(${params.ScaleX}, ${params.ScaleY}) translate(${params.TranslateX}px, ${params.TranslateY}px)`;
+                style.transform = `matrix(${params.M11},${params.M12},${params.M21},${params.M22},${params.M31},${params.M32})`;
                 htmlElement.classList.remove(WindowManager.unoUnarrangedClassName);
                 return true;
             }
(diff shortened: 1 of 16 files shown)


src/Uno.UI.Wasm/WasmScripts/Uno.UI.js
```diff
@@ -1540,6 +1632,28 @@ class WindowManagerSetContentHtmlParams {
     }
 }
 /* TSBindingsGenerator Generated code -- this code is regenerated on each build */
+class WindowManagerSetElementTransformParams {
+    static unmarshal(pData) {
+        let ret = new WindowManagerSetElementTransformParams();
+        {
+            ret.ScaleX = Number(Module.getValue(pData + 0, "double"));
```
dr1rrb: Not sure that we use this, transform on WASM should only rely on matrix

src/Uno.UI.Wasm/ts/WindowManager.ts
```diff
@@ -476,6 +497,61 @@
...
+			}
+
+
+			htmlElement.classList.remove(WindowManager.unoUnarrangedClassName);
+
+			return true;
+		}
+
+		/**
+		* Arrange and clips a native elements
```
dr1rrb: Comment does not match

src/Uno.UI/UI/Xaml/UIElement.wasm.cs
```diff
@@ -172,6 +172,11 @@ protected internal void SetStyleArranged(params (string name, string value)[] st
 #endif
 		}
 
+		protected internal void SetNativeTransform(double scaleX, double scaleY, double translateX, double translateY)
```
dr1rrb: `Transform` on WASM are not using this (and cannot since they are using `matrix`) ... we should either change the name of this method (it seems to be used only by shapes to ensure the stretching), either accept a `matrix` as input and make the `Transform` use it.

jeromelaban: Indeed, I'll take a look.

dr1rrb: You should look at here :) : https://github.com/nventive/Uno/blob/f93828c65af3584929b27183f2e9641773d0f09d/src/Uno.UI/Media/NativeRenderTransformAdapter.wasm.cs#L29

jeromelaban: There you go :) Should be a lot better indeed.

src/Uno.UI/UI/Xaml/WindowManagerInterop.Wasm.cs
```diff
@@ -614,6 +694,81 @@ private struct WindowManagerSetContentHtmlParams
...
+						CultureInfo.InvariantCulture,
+						"rect({0}px,{1}px,{2}px,{3}px",
+						clipRect.Value.Top,
+						clipRect.Value.Right,
+						clipRect.Value.Bottom,
+						clipRect.Value.Left
+					)
+					: "";
+
+				SetStyles(
```
carldebilly: Using this won't remove the class `uno-unarranged` anymore... leading to showing nothing.

jeromelaban: That's what the `true` below is for.

PR #599, opened 2019-02-14
[Wasm] Fix rectangle clipping regression

jeromelaban:
## PR Type
- Bugfix

Rectangles are clipped incorrectly.

## What is the new behavior?
Inner rectange SVG is now clipped properly since #594, but the rectangle needs a larger clipping area. The clipping property is now applied properly, if there's one.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/FrameworkElement.Layout.wasm.cs
+++ b/src/Uno.UI/UI/Xaml/FrameworkElement.Layout.wasm.cs
@@ -209,11 +209,23 @@ namespace Windows.UI.Xaml
 					throw new InvalidOperationException($"{this}: Invalid frame size {newRect}. No dimension should be NaN or negative value.");
 				}
 
-				// Disable clipping for Scrollviewer (edge seems to disable scrolling if 
-				// the clipping is enabled to the size of the scrollviewer, even if overflow-y is auto)
-				var clip = this is Controls.ScrollViewer ? (Rect?)null : new Rect(0, 0, newRect.Width, newRect.Height);
+				Rect? getClip()
+				{
+					// Disable clipping for Scrollviewer (edge seems to disable scrolling if 
+					// the clipping is enabled to the size of the scrollviewer, even if overflow-y is auto)
+					if(this is Controls.ScrollViewer)
+					{
+						return null;
+					}
+					else if(Clip != null)
+					{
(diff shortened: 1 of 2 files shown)


src/Uno.UI/UI/Xaml/FrameworkElement.Layout.wasm.cs
```diff
@@ -209,11 +209,23 @@ private void ArrangeNative(Point offset, Size oldRenderSize)
 					throw new InvalidOperationException($"{this}: Invalid frame size {newRect}. No dimension should be NaN or negative value.");
 				}
 
-				// Disable clipping for Scrollviewer (edge seems to disable scrolling if 
-				// the clipping is enabled to the size of the scrollviewer, even if overflow-y is auto)
-				var clip = this is Controls.ScrollViewer ? (Rect?)null : new Rect(0, 0, newRect.Width, newRect.Height);
+				Rect? getClip()
```
dr1rrb: Why do you create a delegate for that ?

carldebilly: ... more than a delegate: a _display class_ capturing `this`.

jeromelaban: It's not a delegate, it's a local function. But it's that or a triple-if with a local variable... I want C# expression pattern matching :(

dr1rrb: Yes you're right it's a local function ... which will be translated as a display class by the compiler, like if you have declared a delegate 🤓

When do we update to C# 8 ?

jeromelaban: Nope, it's not a like a delegate, there's no indirection, there's a display class, but that's it.

I wrote something like this originally:

```
var clip = this is Controls.ScrollViewer ? (Rect?)null : Clip != null ? Clip.Rect : new Rect(0, 0, newRect.Width, newRect.Height);
```

because having to write `Rect clip;` then a bunch of if else, is not particularly nice either...

dr1rrb: 👌

PR #600, opened 2019-02-14
Review keyboardRect calculation

YGuerin:

## PR Type

- Bugfix

On Android, Keyboard measurement can be negative when it's hidden

## What is the new behavior?

Keyboard measurement is limited to positive values only

## PR Checklist

## Other information

https://nventive.visualstudio.com/Umbrella/_workitems/edit/145363
https://nventive.visualstudio.com/Umbrella/_workitems/edit/145370
https://nventive.visualstudio.com/Umbrella/_workitems/edit/145379

### DIFF
--- a/src/Uno.UI/UI/ViewManagement/InputPane/InputPane.cs
+++ b/src/Uno.UI/UI/ViewManagement/InputPane/InputPane.cs
@@ -84,7 +84,11 @@ namespace Windows.UI.ViewManagement
 
 		internal void OnOccludedRectChanged()
 		{
+#if __ANDROID__
+			var args = new InputPaneVisibilityEventArgs(KeyboardRect);
+#else
 			var args = new InputPaneVisibilityEventArgs(OccludedRect);
+#endif
 
 			if (Visible)
 			{
--- a/src/Uno.UI/KeyboardRectProvider.Android.cs
+++ b/src/Uno.UI/KeyboardRectProvider.Android.cs
@@ -78,8 +78,14 @@ namespace Uno.UI
 			// [rect] displayRect	: display area = screen - (bottom: nav_bar)
 			// [rect] usableRect	: usable area = screen - (top: status_bar) - (bottom: keyboard + nav_bar)
 			var navigationRect = new Rect(0, displayRect.Bottom, realMetrics.WidthPixels, realMetrics.HeightPixels);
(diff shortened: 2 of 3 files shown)


src/Uno.UI/UI/ViewManagement/InputPane/InputPane.cs
```diff
@@ -84,7 +84,11 @@ public bool TryHide()
 
 		internal void OnOccludedRectChanged()
 		{
+#if __ANDROID__
```
MatFillion: it seems odd to me that we return KeyboardRect. For Android are we using OccludedRect at all then?

jeromelaban:
This PR needs a UI Test

src/Uno.UI/UI/ViewManagement/InputPane/InputPane.cs
agneszitte-nventive: Was thinking the same, NavigationBarRect is not needed at all for android for the inputpane ?

agneszitte-nventive: Was thinking the same, NavigationBarRect is not needed at all for android for the inputpane ?

YGuerin: I think you right. I will add some UI tests but it's probably the wrong place to fix it

agneszitte-nventive: Thanks in advance for the UI tests ;) !

XiaotianNetlift: Both the navigation bar and keyboard may be blocking the bottom area. Hence why they are both being propagated from here. OccludedRect should be used to calculate bottom offset. And, keyboard and navigation rects are used to check their visibility.

YGuerin: Indeed. As I said above, this is the wrong place to fix this bug.
I've found that NavigationRect is always equals to 48, regardless of navigation bar visibility (I've found that thanks to the UI Tests). I didn't have time to fix it this week but I hope I'll be able to do it the next one.

YGuerin:
Creating a new pull request for that one.
https://github.com/nventive/Uno/pull/674

PR #609, opened 2019-02-19
[Wasm] Fix support for GotFocus routed event

jeromelaban:
## PR Type

- Bugfix

The `UIElement.GotFocus` event receives a null `RoutedEventArgs`.

## What is the new behavior?
The `UIElement.GotFocus` event now has a proper `RoutedEventArgs` instance.

## PR Checklist

### DIFF
--- a/src/Uno.UI/UI/Xaml/UIElement.wasm.cs
+++ b/src/Uno.UI/UI/Xaml/UIElement.wasm.cs
@@ -264,6 +264,7 @@ namespace Windows.UI.Xaml
 			PointerEventExtractor, // See PayloadToPointerArgs
 			TappedEventExtractor,
 			KeyboardEventExtractor,
+			FocusEventExtractor,
 		}
 
 		private class EventRegistration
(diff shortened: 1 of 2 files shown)


src/Uno.UI/UI/Xaml/UIElement.wasm.cs
```diff
@@ -367,6 +367,8 @@ public bool Dispatch(EventArgs eventArgs, string nativeEventPayload)
 						args = _payloadConverter(nativeEventPayload);
 					}
 
+					args = args ?? new RoutedEventArgs { OriginalSource = this };
```
dr1rrb: Not sure to understand how does `args` may be `null` here

carldebilly: I thought the changer where going to be in `WindowManager.ts`, in `getEventExtractor()`.
Here it should be an exception to have a null `args`.

jeromelaban: It would en up in being a null string, as there are no parameters for that event. What would you put there ?

jeromelaban: Should be better now :)

PR #613, opened 2019-02-20
[iOS] Fix transform animations which uses 'BeginTime'

dr1rrb:
[iOS] Fix transform animations which uses 'BeginTime' will wait the delay without any transform

## PR Type
- Bugfix

On iOS, when animating a `Transform` with a `BeginTime != TimeSpan.Zero` and the target property is not 0 (e.g. `TranslateTransform.Y == 150`), then while waiting for the delay, the current transform is not applied on the view (i.e. the view goes back to its original position like if `TranslateTransform.Y = 0`)

## What is the new behavior?
We wait for the first frame of the animation to disable the managed transform (cf. `RenderTransformAdapter.iOS.cs`)

## PR Checklist

https://nventive.visualstudio.com/Umbrella/_workitems/edit/147309

### DIFF
--- a/src/Uno.UWP/UI/Composition/CoreAnimation.iOS.cs
+++ b/src/Uno.UWP/UI/Composition/CoreAnimation.iOS.cs
@@ -152,6 +152,11 @@ namespace Windows.UI.Composition
 
 			_onAnimationStarted = (s, e) =>
 			{
+				// This will disable the transform while the native animation handles it
+				// It must be the first thing we do when the animation starts
+				// (However we have to wait for the first frame in order to not remove the transform while waiting for the BeginTime)
+				_prepare?.Invoke();
+
 				var anim = s as CAAnimation;
 
 				if (anim == null)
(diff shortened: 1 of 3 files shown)


src/Uno.UWP/UI/Composition/CoreAnimation.iOS.cs
```diff
@@ -152,6 +152,11 @@ private void Animate(CALayer layer, float from, float to, float delayMillisecond
 
 			_onAnimationStarted = (s, e) =>
 			{
+				// This will disable the transform while the native animation handle it
```
MatFillion: handles*

src/Uno.UWP/UI/Composition/CoreAnimation.iOS.cs
```diff
@@ -152,6 +152,11 @@ private void Animate(CALayer layer, float from, float to, float delayMillisecond
...
+				// It must be the first thing we do when the animation started
+				// (However we have to wait for the first frame in order to not remove the transform while waiting for the BeginTime)
```
MatFillion: Good note!

PR #615, opened 2019-02-20
Patch SplitView opening issue

dr1rrb:
## PR Type
- Other: Patch

In some applications it may occurs that the menu stay partially visible when closed

## What is the new behavior?
The menu is fully opened / closed

## PR Checklist

https://nventive.visualstudio.com/Umbrella/_workitems/edit/146335

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/SplitView/SplitView.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/SplitView/SplitView.cs
@@ -384,6 +384,9 @@ namespace Windows.UI.Xaml.Controls
 				PaneOpening?.Invoke(this, null);
 			}
 
+#if __IOS__
+			PatchInvalidFinalState(stateName);
+#endif
 			VisualStateManager.GoToState(this, stateName, useTransitons);
 
 			if (!IsPaneOpen)
--- a/src/Uno.UI/UI/Xaml/Controls/SplitView/SplitView.iOS.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/SplitView/SplitView.iOS.cs
@@ -1,0 +1,78 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using Windows.UI.Xaml.Media;
+using Windows.UI.Xaml.Media.Animation;
+
+namespace Windows.UI.Xaml.Controls
+{
+	partial class SplitView
+	{
+		/*
+
+			About the PatchInvalidFinalState:
+				On iOS for some strange reasons, the burger menu may not complete its open/close animation,
+				but instead either does not appear at all (when opened), either stays partialy visible (when closed).
+				This seems to be related to the fact that  animations of SplitView are altering the Grid.Column and the Visibility 
+				of some controls.
+				Usually it will become fully visible/hidden when a rendering cycle occurs, but this causes
+				a delay for the users which seems aukward.
+				The patch here only ensure to invalidate the view to force a layouting cycle.
+			
+		*/
+		
+		private VisualStateGroup _patchDisplayModeStatesGroup;
+		
+		private void PatchInvalidFinalState(string targetStateName)
+		{
+			// Get the DisplayModeStates visual state group from the root element of the template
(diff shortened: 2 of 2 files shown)


src/Uno.UI/UI/Xaml/Controls/SplitView/SplitView.cs
```diff
@@ -384,6 +384,9 @@ private void UpdateVisualStates(bool useTransitons)
 				PaneOpening?.Invoke(this, null);
 			}
 
+#if __IOS__
+			PatchInvalidFinalState(stateName);
```
carldebilly: Create a partial method and you won't need the `#if __IOS__` :-).

src/Uno.UI/UI/Xaml/Controls/SplitView/SplitView.iOS.cs
```diff
@@ -0,0 +1,78 @@
...
+using Windows.UI.Xaml.Media.Animation;
+
+namespace Windows.UI.Xaml.Controls
+{
+	partial class SplitView
+	{
+		/*
+
+			About the PatchInvalidFinalState:
+				On iOS for some strange reasons, it may occures that the burger menu does not completes its open/close animation,
```
MatFillion: "it may occures" doesn't work
```suggestion
				On iOS for some strange reasons, the burger menu may not complete its open/close animation,
```

src/Uno.UI/UI/Xaml/Controls/SplitView/SplitView.iOS.cs
```diff
@@ -0,0 +1,78 @@
...
+				but instead either does not appear at all (when opened), either stays partialy visible (when closed).
```
MatFillion: Good explanation!

src/Uno.UI/UI/Xaml/Controls/SplitView/SplitView.iOS.cs
```diff
@@ -0,0 +1,78 @@
...
+				This seems to be related to the fact that  animations of SplitView are altering the Grid.Column and the Visibility 
+				of some controls.
+				Usually it will finally becomes fully visible/hidden when a rendering cycle occures, but this causes
+				a delay for the users which seems aukward.
```
MatFillion:
```suggestion
				awkward
```

MatFillion:
```suggestion
				a delay for the users which seems awkward.
```

src/Uno.UI/UI/Xaml/Controls/SplitView/SplitView.iOS.cs
```diff
@@ -0,0 +1,78 @@
...
+	partial class SplitView
+	{
+		/*
+
+			About the PatchInvalidFinalState:
+				On iOS for some strange reasons, it may occures that the burger menu does not completes its open/close animation,
+				but instead either does not appear at all (when opened), either stays partialy visible (when closed).
+				This seems to be related to the fact that  animations of SplitView are altering the Grid.Column and the Visibility 
+				of some controls.
+				Usually it will finally becomes fully visible/hidden when a rendering cycle occures, but this causes
```
MatFillion:
```suggestion
				Usually it will become fully visible/hidden when a rendering cycle occurs, but this causes
```

src/Uno.UI/UI/Xaml/Controls/SplitView/SplitView.cs
MatFillion: but it's really a patch which should never be implemented on other platforms. Generally I'd agree with partial method but in this case it might be better to keep the #if __ IOS __ ? I'm ok with either.

dr1rrb: Indeed I asked myself if I should create a partial method for that or not, but I think that partial methods should be see as "optional feature", like `OnLoadedPartial` that you **can** implement to do something platform specific. 
Here we **must** invoke this method on iOS, and it's not something that you may implement if needed on another platform.

PR #622, opened 2019-02-22
Fix end of animations

dr1rrb:
- Make sure to set the final value of an animated property, so if a frame is missing at the end (which may happend on iOS) we are sure to complete the animation
- Ensure to reset the value set using the `Animations` precendence, so are able to change it later from code

## PR Type
 - Bugfix 

At the end of a `DoubleAnimation` we rely on the fact that the property was altered by the native animation.

## What is the new behavior?
Now we ensure to set the **expected** final value, using the `Local` precedence

## PR Checklist

## Other information
https://nventive.visualstudio.com/Umbrella/_workitems/edit/147416
https://nventive.visualstudio.com/Umbrella/_workitems/edit/147515

### DIFF
--- a/src/Uno.UI/DataBinding/BindingPropertyHelper.cs
+++ b/src/Uno.UI/DataBinding/BindingPropertyHelper.cs
@@ -753,9 +753,6 @@ namespace Uno.UI.DataBinding
 				}
 			}
 
-			var propertyInfo = Uno.Funcs.CreateMemoized(() => GetPropertyInfo(type, property, allowPrivateMembers: false));
-			var propertyType = Uno.Funcs.CreateMemoized(() => GetPropertyOrDependencyPropertyType(type, property));
-
 			// Start by using the provider, to avoid reflection
 			if (BindableMetadataProvider != null)
 			{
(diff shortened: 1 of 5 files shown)


src/Uno.UI/DataBinding/BindingPropertyHelper.cs
```diff
@@ -753,9 +753,6 @@ private static ValueSetterHandler InternalGetValueSetter(Type type, string prope
 				}
 			}
 
-			var propertyInfo = Uno.Funcs.CreateMemoized(() => GetPropertyInfo(type, property, allowPrivateMembers: false));
```
jeromelaban: Why do you need to remove this ?

dr1rrb: It's just a minor optimization I did : I moved it in the inner scope to reduce useless allocation, and I also removed the memoized deferring as it was used only once directly in those scope, again to reduce useless allocation.

jeromelaban: The first one is probably some leftover optimization, you moved the second one which is fine

src/Uno.UI/UI/Xaml/Media/Animation/DoubleAnimation.cs
```diff
@@ -342,15 +342,19 @@ private void OnEnd()
 
 			if (FillBehavior == FillBehavior.HoldEnd)//Two types of fill behaviors : HoldEnd - Keep displaying the last frame
 			{
-				HoldValue();
+				// We set the final value using the "Local" precedence
```
MatFillion: so this is the key change? the PR description mentions setting the value with animations precedence but here it is local precedence?

PR #629, opened 2019-02-26
Added Set Cursor support for Wasm

Firefly74940:

## PR Type

Feature 

Nothing happend as it's "NotImplemented" part

## What is the new behavior?
Support of different cursors on WASM

## PR Checklist

## Other information

It's a draft, code hasn't been tested yet ( need some downloads to be able to build uno.ui and test it )
for now draft is mostly about having feedback on : 

CoreCursorType : Should it be on it's own file ? 

TS setCursor: i did put it in WindowManager ( as it hold many features ) for now, but is it the right place ?


jeromelaban:
Thanks for your contribution! You should be able to test it right inside the Samples app that is in the Uno.UI solution. You can [add a sample](https://github.com/nventive/Uno/blob/master/doc/articles/debugging-uno-ui.md#running-the-samples-applications) that sets the pointer type so you can validate that it works properly.

For the typescript, the WindowManager file is getting large and we'll be refactoring it to split it as it gets too large, but for now, you can leave it there.

As for the duplicate type, you can simple move the file from the generated folder to its proper place in the rest of the folder structure.

src/Uno.UI.Wasm/ts/WindowManager.ts
```diff
@@ -1340,6 +1340,14 @@
 			}
 			return rootElement === element || rootElement.contains(element);
 		}
+
+		public setCursor( cssCursor: string) {
+			const unoBody = document.getElementById("uno-body");
```
carldebilly: `uno-body`: this element name is supplied in the constructor as `containerElementId`.

src/Uno.UI.Wasm/ts/WindowManager.ts
```diff
@@ -1340,6 +1340,14 @@
...
+
+			if (unoBody) {
+				unoBody.style.cursor = cssCursor;
```
carldebilly: The biggest problem I see with this approach is it won't work for controls deriving from `ButtonBase` because of this css: https://github.com/nventive/Uno/blob/master/src/Uno.UI.Wasm/WasmCSS/Uno.UI.css#L46

The goal was to set the pointer as "pointer" for buttons to improve usuability of such controls in wasm. So changing it globally on the _window_ won't affect buttons and other controls deriving from `ButtonBase`.

Maybe we should also add a property to the root element to temporary deactivate the custom css on buttons while the pointer is _customized_.

src/Uno.UI.Wasm/ts/WindowManager.ts
Firefly74940: i simply copied setupSplashScreen method, i will update it ! 

should i also update https://github.com/nventive/Uno/blob/0a54298294d493c1c822f790e54c847523fc04b0/src/Uno.UI.Wasm/ts/WindowManager.ts#L156 in the same commit ?

carldebilly: Yes please ;-)

src/Uno.UI.Wasm/ts/WindowManager.ts
Firefly74940: Using PointerExited / PointerEntered in button code would work ( but then you have to handle it cleverly.. ) [UWP toolkit](https://github.com/windows-toolkit/WindowsCommunityToolkit/blob/master/Microsoft.Toolkit.Uwp.UI/Extensions/Mouse/Mouse.cs) provide a way to solve it. 

For my use case, having the property to work everywhere except on button / hyperlink would be ok 

else, spawning a style node and removing it when done would work i guess: 
https://codepen.io/anon/pen/gEbKPB

Firefly74940:
wow, https://github.com/nventive/Uno/pull/629/commits/32f0802e9bdd9cd56bac4831c3020c5d08f0445a contain way too much changes, i will do a clean branch when i've solved my compiles issues

src/Uno.UI.Wasm/ts/WindowManager.ts
Firefly74940: well, turn out as it's a static function, it can't access it, but i guess when the WindowManager code will be refactored, it might move to a static const to solve that particular issue

Firefly74940:
Ok,
i've added a sample page
and i've added the technique illustrated here : https://codepen.io/anon/pen/gEbKPB  to override the behaviour pointed by @carldebilly  https://github.com/nventive/Uno/pull/629#discussion_r260398559
i wasn't able to move the CoreCursorType file without getting build error, so i reverted that change

is there anything else i should do before moving the draft to a pull request ?

jeromelaban:
This looks good to me, @carldebilly does it address your concern ?

Firefly74940:
if i may up this pr for review @carldebilly 
not that it is urgent or anything, but as @jeromelaban said WindowManager would get a refactor, it would avoid the trouble of merging conflits on that ( or redo pr )

PR #637, opened 2019-02-28
Fix timepicker bugs with clockmode in android and minute increments value are coerced to be between 1 to 30 #636

AlexTrepanier:
## PR Type
By default Time increment is set to 1
if you assign a negative value or 0, it will use 1 minute increments
if you assign a value over 30, it will use 30 minute increments
[Android] In case clock mode still appear for some reason picked value will be rounded to minute increment intervals..

## PR Checklist

148228 [Android] Right theme (clock or spinner) is selected for specific time increments 
148229 [Android] Right time is picked and rounded to nearest time increment in clock mode 
148241 [Android] won't open if `MinuteIncrement` is not set
AB#148228 
AB#148229 
AB#148241


src/Uno.UI/UI/Xaml/Controls/TimePicker/UnoTimePickerDialog.Android.cs
```diff
@@ -74,6 +76,8 @@ private void SetMinuteIncrement()
 
 			if (minutePicker != null)
 			{
+				IsInSpinnerMode = true;
```
MatFillion: it's a little weird that IsInSpinnerMode is set as a side-effect of calling SetMinuteIncrement.

MatFillion: Its ok for now I suppose, because it doesn't yet support changing the MinuteIncrement on TimePickerDialog, if it changes on the TimePicker, it would create a new dialog anyway on next "open", correct?

AlexTrepanier: yes it could be refactored a little better.. lets keep it like this for now.. i'll refactor and test another way after

PR #646, opened 2019-03-04
PR fix Time picker initial time when using time increment is using initial time seconds when rounding.. it should ignore seconds.. and [iOS] TimePicker is clipped off screen when ios:FlyoutPlacement isnt set

AlexTrepanier:
PR fix Time picker initial time when using time increment is using initial time seconds when rounding.. it should ignore seconds.. and [iOS] TimePicker is clipped off screen when ios:FlyoutPlacement isnt set

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/TimePicker/TimePicker.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/TimePicker/TimePicker.cs
@@ -30,6 +30,32 @@ namespace Windows.UI.Xaml.Controls
 
 		public TimePicker() { }
 
+		#region Time DependencyProperty
+
+		public TimeSpan Time
+		{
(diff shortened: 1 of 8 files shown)


src/Uno.UI/UI/Xaml/Controls/TimePicker/TimePicker.cs
```diff
@@ -123,7 +160,7 @@ private void SetupFlyoutButton()
 			UpdateDisplayedDate();
 		}
 
-		partial void OnTimeChangedPartial(TimeSpan oldTime, TimeSpan newTime)
+		void OnTimeChangedPartial(TimeSpan oldTime, TimeSpan newTime)
```
dr1rrb: Please always specify the access modifier
```suggestion
		private void OnTimeChangedPartial(TimeSpan oldTime, TimeSpan newTime)
```

PR #658, opened 2019-03-08
Bump dev version and update to latest dev to realign after stable release

MatFillion:


jeromelaban:
The alignment is missing the changes in the Directory.props file

build/Uno.UI.WpfHost.nuspec
```diff
@@ -16,7 +16,7 @@
 	<dependencies>
 
 	  <group targetFramework="net462">
-		<dependency id="Uno.UI" version="1.42.0-dev.395" />
+		<dependency id="Uno.UI" version="1.43.0-dev.965" />
```
agneszitte-nventive: Should be 1.43.1-dev.XXX instead of 1.43.0-dev.XXX

jeromelaban: This file is automatically updated by the build:

https://github.com/nventive/Uno/blob/4ad2327e8f0189bbe75f7b0e08365f7e73331c65/build/Uno.UI.proj#L135

PR #715, opened 2019-03-27
Added workaround for inverted ListView fling issue on Android P

XiaotianNetlift:

## PR Type
Bugfix

On Android P if a ListView is inverted, trying to use the fling gesture will cause it to scroll in the opposite of the desired direction.

## What is the new behavior?
Fling will scroll in the desired direction.

## PR Checklist

## Other information
https://issuetracker.google.com/u/1/issues/112385925
https://android-review.googlesource.com/c/platform/frameworks/native/%2B/800996
> In P, we switched VelocityTracker to use raw coordinates when
calculating the velocity. While that helped solve the issues with views
being moved while touched, this has broken some other assumptions in the
view hierarchy. Currently during dispatch, MotionEvent coordinates are
getting adjusted as events are passed to the child views.
>
> One example is a rotated ListView. When setRotation is called on a
ListView, ListView continues to use getYVelocity to determine the fling
speed. However, after the view is rotated, it's action is in the
horizontal direction. This causes flings on ListView to not work
properly.

https://nventive.visualstudio.com/Umbrella/_workitems/edit/140564/


src/Uno.UI/UI/Xaml/Controls/ListViewBase/NativeListViewBase.Android.cs
```diff
@@ -314,5 +317,60 @@ bool ILayoutConstraints.IsHeightConstrained(View requester)
...
+			// See: https://issuetracker.google.com/u/1/issues/112385925
+			velocityX = Recalibrate(velocityX, _deltaX);
+			velocityY = Recalibrate(velocityY, _deltaY);
+
+			return base.Fling(velocityX, velocityY);
+
+			int Recalibrate(int value, float? hint)
+			{
+				// note: the opposite sign should be used to recalibrate the velocity
+				return hint.HasValue && hint != 0
```
MatFillion: so since we check the sign, we _should_ be futureproof even if android fixes their bug in android Q or later, correct?

jeromelaban:
The issue has been fixed (https://www.google.com/url?q=https://android-review.googlesource.com/c/platform/frameworks/native/%2B/800996&sa=D&usg=AFQjCNEnROS6Ez5VUl1R1FF2hHI8dimKKA) can we detect whether this is fixed or not so we don't apply the workaround incorrectly ? Does it still make sense to have this fix ?

XiaotianNetlift:
Indeed, we should be checking the sdk version to toggle this.
It would still make sense to keep this workaround in case the phone was not updated.

src/Uno.UI/UI/Xaml/Controls/ListViewBase/NativeListViewBase.Android.cs
```diff
@@ -56,6 +60,12 @@ public NativeListViewBase() : base(ContextHelper.Current)
 			InitializeSnapHelper();
 
 			MotionEventSplittingEnabled = false;
+
+#if __ANDROID_28__
+			_shouldRecalibrateFlingVelocity = Android.OS.Build.VERSION.SdkInt >= Android.OS.BuildVersionCodes.P;
+#else
+			_shouldRecalibrateFlingVelocity = Android.OS.Build.VERSION.SdkInt >= 28;
```
XiaotianNetlift: Would it still be necessary to check whether we are targeting `MonoAndroid90` or not? The problem was that `BuildVersionCodes.P` isn't defined when targeting `MonoAndroid80`.

XiaotianNetlift: @jeromelaban @MatFillion

MatFillion: we can probaly just use `_shouldRecalibrateFlingVelocity = Android.OS.Build.VERSION.SdkInt >= 28;`

MatFillion: by reading the bug report, are we sure the fix is introduced with android Q ? we could say `Android.OS.Build.VERSION.SdkInt == 28` instead of >=28

XiaotianNetlift: Yeah, we will have to wait until the fix is released to update that.

PR #733, opened 2019-04-02
fixed status-bar, keyboard, nav-bar layout on android

XiaotianNetlift:
GitHub Issue (If applicable): n/a

## PR Type
Bugfix

Keyboard & navigation bar height is incorrectly measured on certain device with soft navigation bar, like HuaWei & Pixel. This mess up app using `SoftInput.AdjustNothing` with custom padding xaml-behavior, causing text-box to semi-covered behind the keyboard is opened, or behind the navigation bar when the keybobard is closed.

## What is the new behavior?
The above does not happen.

## PR Checklist

## Other information

https://nventive.visualstudio.com/Umbrella/_workitems/edit/150233


XiaotianNetlift:
Setting the `TranslucentNavigation` flag seems to cause the ui to display under the status bar, despite `TranslucentStatus` is not set. And, their integer value don't overlaps. This also occurs before this PR.

src/Uno.UI/UI/Xaml/ApplicationActivity.Android.cs
```diff
@@ -96,19 +97,18 @@ public void ExitFullscreen()
 			Window.ClearFlags(WindowManagerFlags.Fullscreen);
 		}
 
-		private void OnLayoutChanged(Rect occludedRect)
+		private void OnLayoutChanged(Rect statusBar, Rect keyboard, Rect navigationBar)
```
YGuerin: Why do we need to pass statusBar and navigationBar rects if we don't use them ?

src/Uno.UI/UI/Xaml/ApplicationActivity.Android.cs
```diff
@@ -96,19 +97,18 @@ public void ExitFullscreen()
...
 		{
-			if(_inputPane.OccludedRect != occludedRect)
-			{
-				_inputPane.OccludedRect = ViewHelper.PhysicalToLogicalPixels(occludedRect);
-			}
+			Xaml.Window.Current?.RaiseNativeSizeChanged();
+			_inputPane.OccludedRect = ViewHelper.PhysicalToLogicalPixels(keyboard);
```
YGuerin: Shouldn't we check if the sizes are differents (keyboard vs Occluded) ?
In the case of when you just hide / show your navigation bar, it will not change right ? I don't know if it will cost us some performances to check it.

src/Uno.UI/UI/Xaml/Window.Android.cs
```diff
@@ -136,9 +136,13 @@ private double GetLogicalNavigationBarHeightExcluded()
 		{
 			// The real metrics excluded the NavigationBar only if it is plain.
 			// We want to substract it if it is translucent. Otherwise, it will be like we substract it twice.
-			return NavigationBarHelper.IsNavigationBarVisible && NavigationBarHelper.IsNavigationBarTranslucent
-				? NavigationBarHelper.LogicalNavigationBarHeight
+			return ContextHelper.Current is Activity activity && ShouldBeExcluded()
+				? ViewHelper.PhysicalToLogicalPixels(LayoutProvider.Instance.NavigationBarLayout.Height())
 				: 0;
+
+			bool ShouldBeExcluded() => false
```
YGuerin: false || ? Did you forget to remove this ?

YGuerin:
Great one ! Very useful to review that. I left you some comments (it's more like questions that I've asked to myself to fully understand this).
Last comment : Shouldn't we include a .md ?

Otherwise, very very good one.

MatFillion:
@XiaotianNetlift  do you think we could add the linqpad script and .md with images.

src/Uno.UI/UI/Xaml/ApplicationActivity.Android.cs
XiaotianNetlift: It is so that you can listen to the changes, in case, you do need it.

src/Uno.UI/UI/Xaml/ApplicationActivity.Android.cs
XiaotianNetlift: `InputPane.OccludedRect::set` has built-in value changed guard. Also, the existing value-changed guard is useless here, since it is checking physical against logical values which are never the same, exception when it is 0-sized, or the device has a display scaling of 1:1.

src/Uno.UI/UI/Xaml/Window.Android.cs
XiaotianNetlift: Intended to align the `or` conditions below, otherwise we end up with something hard to read, like:
```
var predicate = 
    condition1
    || condition2;
```

XiaotianNetlift:
@MatFillion Will be in #736

src/Uno.UI/LayoutProvider.Android.cs
```diff
@@ -0,0 +1,133 @@
...
+using Android.Graphics.Drawables;
+using Android.Util;
+using Android.Views;
+using Android.Widget;
+using Windows.Foundation;
+using Rect = Android.Graphics.Rect;
+
+namespace Uno.UI
+{
+	public class LayoutProvider
```
davidjohnoliver: Should this be public? Preferably consumer code gets these values via the UWP contract rather than a platform-specific class.

src/Uno.UI/LayoutProvider.Android.cs
```diff
@@ -0,0 +1,133 @@
...
+	{
+		public delegate void LayoutChangedListener(Rect statusBar, Rect keyboard, Rect navigationBar);
+
+		public static LayoutProvider Instance { get; private set; }
+
+		public event LayoutChangedListener LayoutChanged;
+
+		public Rect StatusBarLayout { get; private set; } = new Rect(0, 0, 0, 0);
```
davidjohnoliver: Don't like the '-Layout' suffix. I would go with 'StatusBarBounds' etc

davidjohnoliver: Rect is a struct, there's no need to initialize these values.

src/Uno.UI/LayoutProvider.Android.cs
```diff
@@ -0,0 +1,133 @@
...
+				_adjustResizeLayoutProvider.Start(view);
+			}
+		}
+		internal void Stop()
+		{
+			_adjustNothingLayoutProvider.Stop();
+			_adjustResizeLayoutProvider.Stop();
+		}
+
+		private void MeasureLayout(PopupWindow sender)
```
davidjohnoliver: You tested landscape mode right? Is the navigation bar invariably at the bottom of the screen?

src/Uno.UI/LayoutProvider.Android.cs
```diff
@@ -0,0 +1,133 @@
...
+			{
+				var result = new T();
+				getter(result);
+
+				return result;
+			}
+		}
+
+
+		private class GlobalLayoutProvider : PopupWindow, ViewTreeObserver.IOnGlobalLayoutListener
```
davidjohnoliver: Minor Git nitpick: this should've been committed as a rename of the KeyboardRectProvider file, to preserve the code history.

src/Uno.UI/LayoutProvider.Android.cs
```diff
@@ -0,0 +1,133 @@
...
+using Windows.Foundation;
+using Rect = Android.Graphics.Rect;
+
+namespace Uno.UI
+{
+	public class LayoutProvider
+	{
+		public delegate void LayoutChangedListener(Rect statusBar, Rect keyboard, Rect navigationBar);
+
+		public static LayoutProvider Instance { get; private set; }
```
davidjohnoliver: Ditch this property. There's already one pseudo-singleton in the form of ApplicationActivity.Instance, just access it via that.

src/Uno.UI/UI/Xaml/Window.Android.cs
davidjohnoliver: If readers are wondering why there's a pointless `false`, it's still hard to read.

src/Uno.UI/LayoutProvider.Android.cs
XiaotianNetlift: The keyboard layout detection was not support in landscape mode, and still isn't. On the few phones that I checked the soft navigation bar are always placed on the side:
- On pixel, s8: always on the side of where earplug jack/charging port is normally
- On HuaWei, lg nexeus: always on the right side of current orientation

PR #743, opened 2019-04-05
Set IsPointerCaptured flag on iOS accordingly to the Pointer capture state

MakanWG:
On iOS, when manipulating a slider, the pointer is not properly captured, resulting in unpredictable behavior when releasing your finger outside of the slider bounds.

## What is the new behavior?
The pointer is properly captured and you can keep manipulating it as long as you don't release your finger.

## PR Checklist

Internal Issue :
[https://nventive.visualstudio.com/Umbrella/_workitems/edit/150453](https://nventive.visualstudio.com/Umbrella/_workitems/edit/150453)

### DIFF
--- a/src/Uno.UI/UI/Xaml/UIElement.cs
+++ b/src/Uno.UI/UI/Xaml/UIElement.cs
@@ -292,6 +292,9 @@ namespace Windows.UI.Xaml
 			else
 			{
 				_pointCaptures.Add(value);
+#if  __IOS__
+				IsPointerCaptured = true;
+#endif
 #if __WASM__
 				CapturePointerNative(value);
 #endif
@@ -304,7 +307,9 @@ namespace Windows.UI.Xaml
 			if(_pointCaptures.Contains(value))
 			{
 				_pointCaptures.Remove(value);
-
+#if __IOS__ 
+				IsPointerCaptured = false;
+#endif
 #if __WASM__
 				ReleasePointerCaptureNative(value);
 #endif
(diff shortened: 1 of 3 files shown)


MakanWG:
@carldebilly The IsPointerCaptured property was removed as part of the RoutedEvents PR a while ago on commit 807a4f5761c1bf9a7f07a7e70654bf30836ee968 (Changes on UIElement.cs). I feel like it should have been added back as it is being read on iOS but is never set to true.
Could you take a look at it just to be sure ?

src/Uno.UI/UI/Xaml/UIElement.cs
```diff
@@ -292,6 +292,9 @@ public bool CapturePointer(Pointer value)
 			else
 			{
 				_pointCaptures.Add(value);
+#if  __IOS__
```
davidjohnoliver: should this apply to Android as well?

MatFillion: @MakanWG wasn't sure. We only observed regressions on iOS and lookign at the code it is only used 
[thumb.cs](src\Uno.UI\UI\Xaml\Controls\Primitives\Thumb.cs) for .net and MacOS.  We could apply it everywhere except wasm to be safe.

MakanWG: Checked with @carldebilly , ended up just mirroring the behavior of wasm on iOS to avoid inconsistencies and regressions.
There is definitely work to do to harmonize the behavior for all platforms though.

PR #781, opened 2019-04-17
[iOS] Fix incorrect offset when setting RenderTransform

davidjohnoliver:
When view.Layer.AnchorPoint is set on initializing a transform, reapply the Frame to correct for the offset potentially introduced. This fixes a spurious visual offset when a RenderTransform is set after a view is already visible.

## PR Type

Bugfix 

## PR Checklist

## Other information

https://nventive.visualstudio.com/Umbrella/_workitems/edit/150882

### DIFF
--- a/src/Uno.UI/Media/NativeRenderTransformAdapter.iOSmacOS.cs
+++ b/src/Uno.UI/Media/NativeRenderTransformAdapter.iOSmacOS.cs
@@ -15,7 +15,10 @@ namespace Uno.UI.Media
 		{
 			// On íOS and MacOS Transform are applied by default on the center on the view
 			// so make sure to reset it when the transform is attached to the view
-			Apply(isSizeChanged: false, isOriginChanged: true);
+			InitializeOrigin();
+
+			// Apply the transform as soon as its been declared
+			Update();
 		}
 
 		private CATransform3D _transform = CATransform3D.Identity;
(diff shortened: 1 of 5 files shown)


src/Uno.UI/Media/NativeRenderTransformAdapter.iOSmacOS.cs
```diff
@@ -64,6 +67,17 @@ partial class NativeRenderTransformAdapter
...
 
+		private void InitializeOrigin()
+		{
+			var oldFrame = Owner.Layer.Frame;
+
+			Owner.Layer.AnchorPoint = CurrentOrigin;
+
+			// Restore the old frame to correct the offset potentially introduced by changing AnchorPoint. This is safe to do since we know
+			// that the transform is currently identity.
+			Owner.Layer.Frame = oldFrame;
```
carldebilly: You're not altering it...  why do you need such restore?

dr1rrb: The `Transform` may be something else than identity.

davidjohnoliver: Setting AnchorPoint alters it. Actually Frame is a 'composed property' that depends on Bounds, AnchorPoint, Position.

davidjohnoliver: It should normally be identity unless I'm misreading something. We've just removed the previous transform if any, and haven't yet applied the new transform.

dr1rrb: Depends if we speak about the `Layer.Transform` or the `Transform` property of the object. I agree that the `Layer.Transform` should be identity (except if someone altered it using the native API instead of the Windows/UNO APIs, but it's probably not a concern). Conversely the `Transform` property is usually not identity at this point.

In fact I think that your comment is about the `Layer.Transform`, so it's  right, but we should probably explicit this.

davidjohnoliver: Gotcha. Yes, it's the `Layer.Transform' property we're concerned about, since setting `Frame` when `Layer.Transform` is non-identity gives undefined behaviour according to Apple's docs.

PR #785, opened 2019-04-17
XamlGeneration - exclude 'not_wasm' from WASM

davidjohnoliver:
GitHub Issue (If applicable): #784 

## PR Type

Bugfix 

`not_wasm` prefix in XAML wasn't taken into account.

## What is the new behavior?

Now taken into account.

## PR Checklist

## Other information


carldebilly:
We should rely on a namespace instead of just the prefix. The Uno approach is not XML compliant.

davidjohnoliver:
> We should rely on a namespace instead of just the prefix. The Uno approach is not XML compliant.

Yeah the way we do this could be revisited at some point. The 'target', 'not_target' syntax is also a bit limited now that we have 5 different targets (UWP/Android/iOS/WASM/MacOS).

jeromelaban:
Yes, now that UWP supports conditional xaml we should be adding support for that instead.

Issue #747, opened 2019-04-07
Implement ShowComposeSmsMessageAsync


MartinZikmund:
## I'm submitting a...
- Feature request

## Current behavior
`NotImplementedException`

## Expected behavior
Implement `Windows.ApplicationModel.Chat.ChatMessageManager.ShowComposeSmsMessageAsync` and related APIs so that the following code works:

```
public async Task PrepareSmsAsync( string recipientNumber, string text )
{
    var chatMessage = new Windows.ApplicationModel.Chat.ChatMessage();
    chatMessage.Body = text;
    chatMessage.Recipients.Add( recipientNumber );
    await Windows.ApplicationModel.Chat.ChatMessageManager.ShowComposeSmsMessageAsync( chatMessage );
}
```
## Environment

Nuget Package: Latest

Package Version(s): Latest

Affected platform(s):
- [x] iOS
- [x] Android
- [ ] WebAssembly
- [ ] Windows
- [ ] Build tasks

Visual Studio
- [x] 2017 (version: )
- [ ] 2017 Preview (version: )
- [ ] for Mac (version: )

Relevant plugins
- [ ] Resharper (version: )
```

MartinZikmund:
I will take care of implementing this.

davidjohnoliver:
You may want to take a look at [Xamarin Essentials](https://github.com/xamarin/Essentials).

PR #787, opened 2019-04-17
ChatMessageManager.ShowComposeSmsMessageAsync

MartinZikmund:
#747 

## PR Type

- Feature

`ChatMessageManager.ShowComposeSmsMessageAsync` is not implemented

## What is the new behavior?
`ChatMessageManager.ShowComposeSmsMessageAsync` is implemented 😉 . Also required `ChatMessage` and its `Recipients` and `Body` properties.

## PR Checklist

## Other information
Needed to introduce `NotNullList` which mimics the UWP behavior of not allowing `null` entries in the `Recipients` property. Should be useful for other places as well (for example `FileOpenPicker's` `Extensions` list if I recall correctly. Not sure about the namespace where to have it though. If some other location would be preferred, I can change it.


carldebilly:
Suggestion : having a sample testing for availability of SMS API (implemented and supported on current hardware).

src/Uno.UWP/ApplicationModel/Chat/ChatMessageManager.iOS.cs
```diff
@@ -0,0 +1,43 @@
...
+				throw new ArgumentNullException(nameof(message));
+			}
+
+			var window = UIApplication.SharedApplication.KeyWindow;
+			var controller = window.RootViewController;
+
+			var messageController = new MFMessageComposeViewController();
+
+			messageController.Body = message.Body;
+			messageController.Recipients = message?.Recipients?.ToArray() ?? new string[] { };
```
davidjohnoliver: Same question here. From the [docs](https://developer.apple.com/documentation/messageui/mfmessagecomposeviewcontroller/1614071-recipients) the Apple API seems to be more restricted:

> Each string in the array should contain the phone number of the intended recipient.

src/Uno.UWP/ApplicationModel/Chat/ChatMessageManager.iOS.cs
```diff
@@ -0,0 +1,43 @@
...
+
+			var tcs = new TaskCompletionSource<bool>();
+			messageController.Finished += (sender, e) =>
```
davidjohnoliver: What happens if the user cancels sending the message? What if there's an error? (No network, invalid recipient, etc.) The task should still complete.

src/Uno.UWP/ApplicationModel/Chat/ChatMessageManager.Android.cs
```diff
@@ -0,0 +1,57 @@
...
+				throw new ArgumentNullException(nameof(message));
+			}
+
+			var intent = CreateIntent(message.Body, message.Recipients.ToArray())
+				.SetFlags(ActivityFlags.ClearTop)
+				.SetFlags(ActivityFlags.NewTask);
+
+			ContextHelper.Current.StartActivity(intent);
+			
+			return Task.FromResult(true).AsAsyncAction();
```
davidjohnoliver: On iOS the task completes once the message has been sent, on Android it completes synchronously as soon as the intent is launched... which is the correct behaviour? ('Correct' == matching UWP)

MartinZikmund:
> Suggestion : having a sample testing for availability of SMS API (implemented and supported on current hardware).

@carldebilly How would you suggest implementing this? UWP docs use the API `ApiInformation.IsTypePresent("Windows.ApplicationModel.Chat.ChatMessageManager")` to check if the device is SMS capable, but I am not sure that can be utilized in Uno (apart from branching the logic in the `IsTypePresent` implementation?

src/Uno.Foundation/Uno.Foundation.csproj
```diff
@@ -35,6 +35,10 @@
 	<ItemGroup Condition="'$(TargetFrameworkIdentifier)'=='.NETStandard'">
 		<PackageReference Include="System.Runtime.InteropServices.WindowsRuntime" Version="4.3.0" />
 	</ItemGroup>
+
+	<ItemGroup>
+	  <UpToDateCheckInput Remove="Collections\NonNullList.cs" />
```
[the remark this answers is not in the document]
MartinZikmund: Oh! Missed that, will fix

src/Uno.UWP/ApplicationModel/Chat/ChatMessageManager.Android.cs
```diff
@@ -0,0 +1,57 @@
...
+		private const string RecipientsSeparator = ";";
+
+		public static IAsyncAction ShowComposeSmsMessageAsync(ChatMessage message)
+		{
+			if (message == null)
+			{
+				throw new ArgumentNullException(nameof(message));
+			}
+
+			var intent = CreateIntent(message.Body, message.Recipients.ToArray())
```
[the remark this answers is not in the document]
MartinZikmund: Actually I have tested with arbitrary strings and it did work and threw no exceptions - the "non-number" recipients appear as normal text in the _To_ field of the SMS app. Not sure if it is useful, but it is probably the same thing as in UWP

src/Uno.UWP/ApplicationModel/Chat/ChatMessageManager.iOS.cs
MartinZikmund: Similarly on iOS it outputs non-numeric recipients as normal string in the _To_ field

src/Uno.UWP/ApplicationModel/Chat/ChatMessageManager.Android.cs
MartinZikmund: That is a good point. I will try to dig up my old Lumia to check. On desktop this just opens the deprecated SMS app, so it is not very useful and can't really be used to check.

src/Uno.UWP/ApplicationModel/Chat/ChatMessageManager.iOS.cs
MartinZikmund: The truth is I have closely followed the implementation of SMS in Xamarin.Essentials and this was the approach chosen there. I will check with UWP and try to do the same thing

jeromelaban:
> How would you suggest implementing this?

@MartinZikmund for now, testing with `IsTypePresent` in the sample is probably enough. The supported platforms will raise their own exceptions there, and that's fine.

MartinZikmund:
> > How would you suggest implementing this?
> 
> @MartinZikmund for now, testing with `IsTypePresent` in the sample is probably enough. The supported platforms will raise their own exceptions there, and that's fine.

@jeromelaban I'm sorry, but I am not sure what you meant. `IsTypePresent` will always return `true` on Android and iOS, so it is a different behavior than in case of UWP where it returns false in case device does not have SMS capability. Is that ok?

jeromelaban:
@MartinZikmund What I mean is that the behavior of the IsTypePresent is appropriate and does not need to change. 

When the running device does not provide the proper support, there will be exceptions raised when sending an SMS, which is also appropriate. Is it what happens when sending an sms on windows desktop ?

MartinZikmund:
> @MartinZikmund What I mean is that the behavior of the IsTypePresent is appropriate and does not need to change.
> 
> When the running device does not provide the proper support, there will be exceptions raised when sending an SMS, which is also appropriate. Is it what happens when sending an sms on windows desktop ?

Surprisingly not 😊 - the API actually works on desktop as well and it just opens the deprecated Messages app and which shows "nothing". It works well on Lumia though.

PR #798, opened 2019-04-22
Dev/pako/fix android textbox unfocus

MakanWG:
The Android platform is not able to dismiss a keyboard that was shown using the [FORCE](https://developer.android.com/reference/android/view/inputmethod/InputMethodManager.html#SHOW_FORCED) flag. 
We try to force the hide method as well but it does not work when the Navigation already took place.

## What is the new behavior?
Using the [IMPLICIT](https://developer.android.com/reference/android/view/inputmethod/InputMethodManager.html#SHOW_IMPLICIT) flag instead when showing the keyboard to allow Android to dismiss it when needed.

## PR Checklist

[151524](https://nventive.visualstudio.com/Umbrella/_workitems/edit/151524)

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.Android.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.Android.cs
@@ -427,6 +427,9 @@ namespace Windows.UI.Xaml.Controls
 
 		private void ProcessFocusChanged(bool hasFocus)
 		{
+			//We get the view token early to avoid nullvalues when the view has already been detached
+			var viewWindowToken = _textBoxView.WindowToken;
+
 			_keyboardDisposable.Disposable = CoreDispatcher.Main
 				//The delay is required because the OnFocusChange method is called when the focus is being changed, not when it has changed.
 				//If the focus is moved from one TextBox to another, the CurrentFocus will be null, meaning we would hide the keyboard when we shouldn't.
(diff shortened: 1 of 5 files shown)


src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.Android.cs
```diff
@@ -448,8 +451,7 @@ private void ProcessFocusChanged(bool hasFocus)
 						//When a TextBox gains focus, we want to show the keyboard
 						if (hasFocus && needsKeyboard)
 						{
-							inputManager?.ShowSoftInput(_textBoxView, Android.Views.InputMethods.ShowFlags.Forced);
-							inputManager?.ToggleSoftInput(Android.Views.InputMethods.ShowFlags.Forced, Android.Views.InputMethods.HideSoftInputFlags.ImplicitOnly);
```
MakanWG: Here I couldn't get why we were sending a "show" request immediately followed by a "toggle" request.

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.Android.cs
```diff
@@ -427,6 +427,9 @@ void IOnFocusChangeListener.OnFocusChange(View v, bool hasFocus)
 
 		private void ProcessFocusChanged(bool hasFocus)
 		{
+			//We get the view token early to avoid nullvalues when the view has already been detached
+			var viewWindowToken = _textBoxView.WindowToken;
```
davidjohnoliver: Is it correct to keep this indefinitely? Is it still valid, eg, after the activity has been destroyed and recreated? (You can test this by pressing the hardware back button to exit the app, then return to it by pressing its icon.)

davidjohnoliver: Never mind - I was misreading this, it looks fine.

Issue #721, opened 2019-03-28
Add support for Windows.Globalization.Language.CurrentInputMethodLanguageTag


yowl:
## I'm submitting a...

  
 - Feature request 

## Current behavior
Attempts to get the `CurrentInputMethodLanguageTag` fail as its not supported

## Expected behavior
Add some means (maybe just put a setter) so that this property can be called.  This would be beneficial in getting closer to supporting Telerik's UWP controls as they (I think the numeric input) uses this.

## Minimal reproduction of the problem with instructions

Attempt to call the property.  Its marked as not supported https://github.com/nventive/Uno/blob/25d0c4adb3b98b0f201a46c2336adb78d69067a5/src/Uno.UWP/Generated/3.0.0.0/Windows.Globalization/Language.cs#L60-L63

## Environment

```
Nuget Package: 

Package Version(s): 

Affected platform(s):
- [ ] iOS
- [ ] Android
- [x ] WebAssembly
- [ ] Windows
- [ ] Build tasks

Visual Studio
- [ x] 2017 (version: )
- [ ] 2017 Preview (version: )
- [ ] for Mac (version: )

Relevant plugins
- [ ] Resharper (version: )
```

MartinZikmund:
I can do this one for Android and iOS, but not sure about WebAssembly. I think there is no way to get current input language in JavaScript. Should WebAssebly have a setter then?

yowl:
That would be ok for me.

MartinZikmund:
@yowl Can this be closed?

yowl:
@MartinZikmund Yes thanks.

PR #800, opened 2019-04-23
CurrentInputMethodLanguageTag

MartinZikmund:
GitHub Issue (If applicable): #721 

## PR Type

- Feature 

Missing implementation of `CurrentInputMethodLanguageTag`

## What is the new behavior?
Implemented `CurrentInputMethodLanguageTag` and `TrySetInputMethodLanguageTag`. 

## PR Checklist

## Other information
**OS support** - I have noticed that Android tends not to report the Input Keyboard Language reliably and it has been confirmed based on what I read. The newer `LanguageTag` property I am using seems to be even less reliable than the deprecated `Locale` property, unfortunately.

**Code duplication** - I had to resort to code duplication in `Language.iOS.cs` - `UIView FindFirstResponder(UIView view)` method is the same code as in `Uno.UI.Extensions.UIViewExtensions`, which I could not use here as it is in the `Uno.UI` library and hence not accessible. I could move `UIViewExtensions` to `Uno.UWP`, but that would mean breaking backward compatibility. Please advise on the best course of action.


MartinZikmund:
@carldebilly Is the code duplication mentioned in the `Other Information` section OK or should it be addressed somehow?

carldebilly:
I'll let @jeromelaban express himself on this one :-).

But I'm curious: what is the need/use case for this feature?

MartinZikmund:
@carldebilly this feature is needed for Telerik controls support (#721), so I thought it would be nice to implement it for Android and iOS as well, although arguably getting current keyboard input language is usually not too useful 😊

yowl:
@carldebilly seems like there is something special about french keyboards and the minus sign.  This code is used in Telerik's numeric text box control to allow negative numbers
```
        private static bool IsAzertyKeyboard
        {
            get
            {
                return Windows.Globalization.Language.CurrentInputMethodLanguageTag.StartsWith("fr-");
            }
        }
```
and
```
            if (RadNumericBox.IsAzertyKeyboard)
            {
                isNegativeSign = isNegativeSign || (key == VirtualKey.Number6 && DeviceTypeHelper.GetDeviceType() != DeviceType.Phone);
            }
```

jeromelaban:
Indeed, there's duplication there, but that's not a big issue. We'll eventually consolidate, but for now, you can put that method in a `ViewExtensions.iOSmacOS.cs` class in the Extensions folder, so we know that it is reused at that level. We'll call one from the other to avoid binary breaking changes.

I've made changes to that class for macOS support here, to make it available on both: https://github.com/nventive/Uno/pull/802/files#diff-cce17a30f244931e5ac882cfed84883d

src/Uno.UWP/Globalization/Language.Android.cs
```diff
@@ -0,0 +1,24 @@
...
+		{
+			get
+			{
+				var inputMethodManager = ContextHelper.Current.GetSystemService(Context.InputMethodService).JavaCast<InputMethodManager>();
+				var currentInputMethod = inputMethodManager.CurrentInputMethodSubtype;
+				return currentInputMethod.LanguageTag;
+			}
+		}
+
+		public static bool TrySetInputMethodLanguageTag(string languageTag) => false;
```
jeromelaban: We may want to log that it is not supported.

src/Uno.UWP/Globalization/Language.wasm.cs
```diff
@@ -0,0 +1,19 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace Windows.Globalization
+{
+	public partial class Language
+	{
+		public static string CurrentInputMethodLanguageTag { get; private set; } = "";
+
+        public static bool TrySetInputMethodLanguageTag(string languageTag)
```
jeromelaban: We'll probably need to adjust that to get the browser cultures, but for now, not throwing is best :)

src/Uno.UWP/Globalization/Language.Android.cs
MartinZikmund: The `false` should actually be ok - according to docs - https://docs.microsoft.com/en-us/uwp/api/windows.globalization.language.trysetinputmethodlanguagetag - it is expected that the method does nothing on all devices except IoT.

jeromelaban: That makes sense, thanks.

src/Uno.UWP/Globalization/Language.Android.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Windows.Globalization
+{
+	public partial class Language
+	{
+		public static string CurrentInputMethodLanguageTag
+		{
+			get
+			{
+				var inputMethodManager = ContextHelper.Current.GetSystemService(Context.InputMethodService).JavaCast<InputMethodManager>();
```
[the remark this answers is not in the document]
MartinZikmund: Fixed :-)

MartinZikmund:
Resolved the `UIViewExtensions` code duplication, although now there is a potential name clash between `Uno.Extensions.UIViewExtensions` and `UIKit.UIViewExtensions`.

MartinZikmund:
I merged latest master, are any additional changes needed here? @carldebilly @jeromelaban

ghuntley:
Thank-you @MartinZikmund. There was some CI issues which have since been resolved.

PR #650, opened 2019-03-06
Fix touch events not dispatched to UIElement that have a RenderTransform

dr1rrb:
## PR Type
- Bugfix

## What is the current behavior?
Touch events are not dispatched to `UIElement` which has a `RenderTransform`

## What is the new behavior?
We can also tap on the part of a `UIElement`  which does not overlap it original location

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] Tested code with current [supported SDKs](../README.md#supported)
- [ ] Docs have been added/updated which fit [documentation template](https://github.com/nventive/Uno/blob/master/doc/.feature-template.md). (for bug fixes / features)
- [x] [Unit Tests and/or UI Tests](doc/articles/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [x] Contains **NO** breaking changes
- [ ] Updated the [Release Notes](https://github.com/nventive/Uno/tree/master/doc/ReleaseNotes)
- [x] Associated with an issue (GitHub or internal)

## Other information
This is only a partial fix, cf. https://github.com/nventive/Uno/issues/649

Internal Issue (If applicable):
https://nventive.visualstudio.com/Umbrella/_workitems/edit/147978
https://nventive.visualstudio.com/Umbrella/_workitems/edit/148298
https://nventive.visualstudio.com/Umbrella/_workitems/edit/148300
https://nventive.visualstudio.com/Umbrella/_workitems/edit/148303
https://nventive.visualstudio.com/Umbrella/_workitems/edit/148375
https://nventive.visualstudio.com/Umbrella/_workitems/edit/148521

PR #817, opened 2019-04-25
 fix android touch event dispatched to invisible view

XiaotianNetlift:
GitHub Issue (If applicable): #N/A

## PR Type
Bugfix

Currently on android, non-visible views can receive and handle touch events, when they weren't supposed to.

## What is the new behavior?
Non-visible views can no longer receive touch events.

## PR Checklist

## Other information
```
<Grid
    <Grid x:Name="layer0" Visibility="Visible"
    <Grid x:Name="layer1" Visibility="Collapsed"
    <Grid x:Name="layer2" Visibility="Collapsed"
        <Grid.RenderTransform
            <RenderTransform ...
```
Given this view, touch events will be first dispatched `layer2`. If `layer2` or one of its nested views choose to handle the event, then `layer0` won't be able to process that event.

Regression from: #650
The visibility of the children were not considered when dispatching those event:
> https://github.com/nventive/Uno/blob/e0ef795fdd31cc2cbf8d8288d4455f4406c251f8/src/Uno.UI.BindingHelper.Android/Uno/UI/UnoViewGroup.java#L405-L440

https://nventive.visualstudio.com/Umbrella/_workitems/edit/151430


src/Uno.UI.BindingHelper.Android/Uno/UI/UnoViewGroup.java
```diff
@@ -421,12 +424,17 @@ public boolean dispatchTouchEvent(MotionEvent e)
...
+					continue;
+				}
+
 				final Matrix transform = _childrenTransformations.get(child);
 				final float offsetX = getScrollX() - child.getLeft();
 				final float offsetY = getScrollY() - child.getTop();
 
 				if (transform == null || transform.isIdentity()) {
-					// No meaningfull transformation on this child, instead of cloning the MetionEvent,
+					// No meaningful transformation on this child, instead of cloning the MetionEvent,
```
MatFillion:
```suggestion
					// No meaningful transformation on this child, instead of cloning the MotionEvent,
```

src/Uno.UI.BindingHelper.Android/Uno/UI/UnoViewGroup.java
```diff
@@ -421,12 +424,17 @@ public boolean dispatchTouchEvent(MotionEvent e)
 			for (int i = getChildCount() - 1; i >= 0; i--) { // Inverse enumeration in order to prioritize controls that are on top
 				View child = getChildAt(i);
 
+				if (child.getVisibility() != View.VISIBLE)
```
MatFillion: seems odd it is the only place in all this logic where we check visibility.

src/Uno.UI.BindingHelper.Android/Uno/UI/UnoViewGroup.java
```diff
@@ -364,8 +366,9 @@ public boolean dispatchTouchEvent(MotionEvent e)
 
 		if (!_isHitTestVisible || !_isEnabled)
```
MatFillion: should visibility be checked here instead ?

src/Uno.UI.BindingHelper.Android/Uno/UI/UnoViewGroup.java
```diff
@@ -313,7 +315,7 @@ public final void setAlpha(float opacity)
 	public boolean dispatchTouchEvent(MotionEvent e)
 	{
 		String originalIndent = _indent;
-		Log.i("", _indent + "    + " + this.toString());
+		Log.i(LOGTAG, _indent + "    + " + this.toString());
```
carldebilly: Do we need to keep this logging in production code?

davidjohnoliver: +1 
(Hint: we do not.)

src/Uno.UI.BindingHelper.Android/Uno/UI/UnoViewGroup.java
davidjohnoliver: Normally the visibility check would be handled by super.dispatch..., but this is the code path where we don't call it (because it doesn't account for the transform).

src/Uno.UI.BindingHelper.Android/Uno/UI/UnoViewGroup.java
MatFillion: > Normally the visibility check would be handled by super.dispatch..., but this is the code path where we don't call it (because it doesn't account for the transform).

so it is ok as it is now.

src/Uno.UI.BindingHelper.Android/Uno/UI/UnoViewGroup.java
dr1rrb: This code is here to ease debug as debug Android is not really easy, and the whole mothod is commented-out. I would keep it for future debug, however the `LOGTAG` should be put here, as this method is trying here to dump the indent the log to follow the visual tree.

src/Uno.UI.BindingHelper.Android/Uno/UI/UnoViewGroup.java
```diff
@@ -364,8 +366,9 @@ public boolean dispatchTouchEvent(MotionEvent e)
 
 		if (!_isHitTestVisible || !_isEnabled)
 		{
-			// Log.i(this.toString(), "!_isHitTestVisible: " + !_isHitTestVisible);
-			// Log.i(this.toString(), "!_isEnabled: " + !_isEnabled);
+			// Log.i(LOGTAG, _indent + "!_isHitTestVisible: " + !_isHitTestVisible);
```
dr1rrb: Same thing here, do not use the `LOGTAG` (and all Log  in this method)

src/Uno.UI.BindingHelper.Android/Uno/UI/UnoViewGroup.java
davidjohnoliver: > and the whole method is commented-out.

?

(PS: Debugging Java code is actually really easy :p )

XiaotianNetlift: Those `Log` lines were already there. I just added a tag, so they can be easily filtered/searched.
And, this entire method is commented out with `/**/` outside of this snippet.

PR #835, opened 2019-05-01
WASM: flyout is dismissing too fast

carldebilly:
## Bugfix
A flyout will dismiss too fast.

The routed event "pointer pressed" will close the flyout popup before
any content can process it.

## What is the new behavior?
The flyout now requires a full _pressed_ + _released_ sequence
(non-handled events) before closing.

## PR Checklist

- [X] Contains **NO** breaking changes: _the new behavior is coherent with UWP_.

* https://nventive.visualstudio.com/Umbrella/_workitems/edit/152477

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/Popup/Popup.wasm.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Popup/Popup.wasm.cs
@@ -61,23 +61,36 @@ namespace Windows.UI.Xaml.Controls
 				}
 			}
 
+			bool pressed = false;
+
 			if (previousPanel != null)
 			{
-				previousPanel.PointerPressed -= Panel_PointerPressed;
+				previousPanel.PointerPressed -= pointerPressed;
+				previousPanel.PointerReleased -= pointerReleased;
 			}
 			if (newPanel != null)
 			{
-				newPanel.PointerPressed += Panel_PointerPressed;
+				newPanel.PointerPressed += pointerPressed;
+				newPanel.PointerReleased += pointerReleased;
 			}
-		}
 
-		private void Panel_PointerPressed(object sender, Input.PointerRoutedEventArgs e)
-		{
-			if (IsLightDismissEnabled)
+			void pointerPressed(object sender, Input.PointerRoutedEventArgs args)
 			{
-				IsOpen = false;
+				// Both pressed & released must reach
+				// the popup to close it.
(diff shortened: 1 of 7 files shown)


src/Uno.UI/UI/Xaml/Controls/Popup/Popup.wasm.cs
```diff
@@ -61,23 +61,36 @@ protected override void OnIsOpenChanged(bool oldIsOpen, bool newIsOpen)
...
-		private void Panel_PointerPressed(object sender, Input.PointerRoutedEventArgs e)
-		{
-			if (IsLightDismissEnabled)
+			void pointerPressed(object sender, Input.PointerRoutedEventArgs args)
 			{
-				IsOpen = false;
+				// Both pressed & released must reach
+				// the popup to close it.
+				// (and, obviously, the popup must be light dismiss!)
+				pressed = IsLightDismissEnabled;
```
davidjohnoliver: You can't just use Tapped intead of PointerReleased?

carldebilly: Tapped is not a "true" routed event and is the result of the interpretation of pointer events.
I want to prevent source events from generating the tap.

src/Uno.UI/UI/Xaml/Controls/Popup/Popup.wasm.cs
```diff
@@ -61,23 +61,36 @@ protected override void OnIsOpenChanged(bool oldIsOpen, bool newIsOpen)
...
-				newPanel.PointerPressed += Panel_PointerPressed;
+				newPanel.PointerPressed += pointerPressed;
+				newPanel.PointerReleased += pointerReleased;
 			}
-		}
 
-		private void Panel_PointerPressed(object sender, Input.PointerRoutedEventArgs e)
-		{
-			if (IsLightDismissEnabled)
+			void pointerPressed(object sender, Input.PointerRoutedEventArgs args)
```
dr1rrb: If you put it here as nested method with a capture, I'm not sure that the `-=` will work properly (as the target object will now be the object that captured the `pressed`.

carldebilly: Good question.

carldebilly: Changed.

Issue #864, opened 2019-05-14
[iOS] Fast tap while subscribed to `Tapped` event, prevents Pointer events


dr1rrb:
## I'm submitting a...
- Bug report

## Current behavior
If a `GestureRecognizer` is set on a `UIView`, then when it recognize a gesture, the "touches" are cancelled (i.e. the overridable `TouchesCancelled` in fired instead of the `TouchesEnded` in the `UIElement`). Sometimes, it may even prevent the `TouchesBegan` (if UI is big enough?).

This has multiple effects:
1. You may not receive the `PointerReleased` (has it's hooked to the `TouchesEnded`) (usually affect "pressed" visual state)
1. `Tapped` vs pointers events are not fired in the same order as Windows
1. `Tapped` may be fired while no pointers are raised

## Expected behavior
As of windows, we should _always_ (depending of capture, pointer moves, etc.) received events in the following sequence:
1. Pressed
1. Released
1. Tapped

## Minimal reproduction of the problem with instructions
Cf. sample `RoutedEvent_TappedAndRelease`

## More info / how to diagnose
You are probably affected by this bug if:
* Your `UIElement` does not react to "fast" tap, but reacts to "long tap" (keep your finger half a second on screen ... but lower than the "hold" delay).
* Your  `Control` stays in "pressed" state.

## Environment
Nuget Package: Uno.UI
Package Version(s): 1.45.0-dev.1450
Affected platform(s):
- [x] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] Windows
- [ ] Build tasks

Visual Studio: _irrelevant_
Relevant plugins: _none_

PR #865, opened 2019-05-14
Misc fixes - Temp fix PointerRelease not raised on iOS if subscribed to Tapped

dr1rrb:
* Temp fix `PointerRelease` event not raised on iOS if subscribed to `Tapped` (Mitigates: https://github.com/nventive/Uno/issues/864)
* Avoid useless warning when using enum `FlowDirection`
* Add dummy implementation of `CoreWindow.Get[Async]KeyState`

GitHub Issue (If applicable): #864

## PR Type
- Feature
- Other: mitigation

The `PointerReleased` is never fired on a `UIElement` if als subscribed to the `Tapped` event. The most visible effect of this, is that controls stay in pressed state.

## What is the new behavior?
The event is fired ... if we press long enough to trigger `TouchesBegan` (i.e. the `PointerPressed`) (cf. #864)

## PR Checklist

- [x] [Wasm UI Tests](doc/articles/working-with-the-samples-apps.md#running-the-webassembly-ui-tests-snapshots) are not showing unexpected any differences

## Other information
This is only a mitigation : In some cases only the `Tapped` event is being raised without any pointer events. Is seems that the `GestureRecognizer` triggers fast enough to even prevent the `TouchesBegan` (and so the `Touches<Cancelled|Ended>`)


src/SamplesApp/UITests.Shared/UITests.Shared.projitems
```diff
@@ -2633,381 +2642,7 @@
...
-    <Compile Update="C:\s\nv.github\uno\src\SamplesApp\UITests.Shared\Windows_UI_Xaml_Controls\CommandBar\CommandBar_Padding.xaml.cs">
-      <DependentUpon>CommandBar_Padding.xaml</DependentUpon>
-    </Compile>
-    <Compile Update="C:\s\nv.github\uno\src\SamplesApp\UITests.Shared\Windows_UI_Xaml_Controls\CommandBar\CommandBar_Uppercase_Resource.xaml.cs">
-      <DependentUpon>CommandBar_Uppercase_Resource.xaml</DependentUpon>
-    </Compile>
-    <Compile Update="C:\s\nv.github\uno\src\SamplesApp\UITests.Shared\Windows_UI_Xaml_Controls\CommandBar\CommandBar_With_Long_Sentences.xaml.cs">
-      <DependentUpon>CommandBar_With_Long_Sentences.xaml</DependentUpon>
-    </Compile>
-  </ItemGroup>
```
carldebilly: Removed a lot of file nesting declarations...

src/Uno.UWP/UI/Core/CoreWindow.cs
```diff
@@ -6,9 +6,17 @@
...
 {
-	public partial class CoreWindow 
+	public partial class CoreWindow
 	{
-		public global::Windows.UI.Core.CoreDispatcher Dispatcher
+		public CoreDispatcher Dispatcher
 			=> CoreDispatcher.Main;
+
+		[Uno.NotImplemented]
+		public CoreVirtualKeyStates GetAsyncKeyState(System.VirtualKey virtualKey)
```
carldebilly: Seems unrelated to the pr

src/Uno.UI/UI/Xaml/UIElement.iOS.cs
```diff
@@ -459,9 +459,26 @@ public override void TouchesCancelled(NSSet touches, UIEvent evt)
...
 					OriginalSource = this
 				};
-				var isHandledInManaged = RaiseEvent(PointerCanceledEvent, args);
 
-				if (!isHandledInManaged)
+				// As soon as a gesture is recognized by a 'GestureRecognizer' (cf. Tap and DoubleTap events)
+				// the touches are "cancelled" and this method is invoked. So until we remove the GestureRecognizers
+				// for those events, we handle the "Cancelled" similarly as the "Ended" and we raise the "Release".
+				var pointerEventIsHandledInManaged = false;
+				if (IsPointerCaptured || evt.IsTouchInView(this))
```
davidjohnoliver: The test for this one is to put a Button inside a ScrollViewer, press on the Button, and then start to scroll. The Button's Click shouldn't be raised.

davidjohnoliver:
Lots of unrelated files changed..

src/SamplesApp/UITests.Shared/UITests.Shared.projitems
dr1rrb: Removed a lot of invalid absolute path ;)

src/Uno.UWP/UI/Core/CoreWindow.cs
dr1rrb: Actually this PR contains multiple minor changes made to support the SyncFusion DataGrid, however in comment I focused on the most important changes instead of writing useless comment like :
## Current Behavior
`VirtualKeyStates` throw exceptin

## New behavior
`VirtualKeyStates` always return `None`

src/Uno.UI/Generated/3.0.0.0/Windows.UI.Xaml/FlowDirection.cs
```diff
@@ -2,7 +2,7 @@
 #pragma warning disable 114 // new keyword hiding
 namespace Windows.UI.Xaml
 {
-	#if __ANDROID__ || __IOS__ || NET461 || __WASM__ || __MACOS__
+	#if false
```
MatFillion: im confused about this change. Why?

src/Uno.UWP/UI/Core/CoreWindow.cs
dr1rrb: Splitted !

src/Uno.UI/Generated/3.0.0.0/Windows.UI.Xaml/FlowDirection.cs
dr1rrb: Expand the code ;) : one for the class, and one for the attribute. Actually this is generated code, I just manually updated it to match the newly implemented APIs

src/Uno.UI/UI/Xaml/UIElement.iOS.cs
davidjohnoliver: Did you make this test?

dr1rrb: As the `Click` is linked to the `Tapped`, which is raised by the `GestureRecognizer`, it won't be affected by this commit. However yes I've validated, it works properly.

BTW I noticed that we don't even receive the `PointerPressed` nor `PointerReleased` on a `Button` on Windows (FYI @carldebilly ). This should be included in final fix of #864

davidjohnoliver:
@dr1rrb  This is causing Button.Click to be fired twice (once from Tapped, then once from PointerReleased). This breaks CheckBox/ToggleButton amongst other things - IsChecked gets set to !(!IsChecked), which isn't very helpful. (I noticed because the menu button in SamplesApp stopped working.)

davidjohnoliver:
And I just tried the Button-inside-ScrollViewer test, and Click *is* getting called!

dr1rrb:
> 
> 
> And I just tried the Button-inside-ScrollViewer test, and Click _is_ getting called!

Actually it seems to depend of how long you press of the `Button` before "scrolling" and how planets are aligned (sometimes we get all events - possibly even a double `Click` - and sometimes we get the right event sequence). It's probably due to both the `GestureRecognizer` which [delays](https://developer.apple.com/documentation/uikit/uigesturerecognizer/1624209-delaystouchesended) and [alter](https://stackoverflow.com/questions/5818692/how-to-avoid-touches-cancelled-event) touches events, and to the `UIScrollView` which [adds a delay to propagate touches](https://developer.apple.com/documentation/uikit/uiscrollview/1619398-delayscontenttouches). So will revert this commit as it was only a workaround to avoid to fix the full issue (i.e. removing the gesture recognizer on iOS). 

Note: we will however create a custom branch with this commit (or a new commit which uses the flag [cancelstouchesinview](https://developer.apple.com/documentation/uikit/uigesturerecognizer/1624218-cancelstouchesinview) and/or a a feature flag ?) until we fix it properly.

Issue #863, opened 2019-05-14
Implement Windows.System.Power.PowerManager


MartinZikmund:
## I'm submitting a...
 - Feature request 

## Current behavior
The `Windows.System.Power.PowerManager` APIs are not supported.

## Expected behavior
Add support for `Windows.System.Power.PowerManager` on iOS, Android.

## Environment
```
Nuget Package: latest

Package Version(s): latest

Affected platform(s):
- [x] iOS
- [x] Android
- [x] WebAssembly
- [ ] Windows
- [ ] Build tasks

Visual Studio
- [ ] 2017 (version: )
- [x] 2019 (version: )
- [ ] for Mac (version: )

Relevant plugins
- [ ] Resharper (version: )

MartinZikmund:
I will implement this 😉

PR #877, opened 2019-05-19
Windows.System.PowerManager API

MartinZikmund:
GitHub Issue (If applicable): #863 

## PR Type

- Feature 

`PowerManager` class and its features are not implemented.

## What is the new behavior?
Most properties and events of `Windows.System.Power.PowerManager` are implemented on iOS and Android:

    - BatteryStatus
    - EnergySaverStatus
    - PowerSupplyStatus
    - RemainingChargePercent
    - PowerSupplyStatusChanged
    - EnergySaverStatusChanged
    - RemainingChargePercentChanged
    - BatteryStatusChanged

## PR Checklist

## Other information
The `PowerSupplyStatusChanged` event does not behave 100% on Android - the system sometimes sends the broadcast but the property is still not yet updated. That is a Android OS issue though.

The two remaining APIs related to `RemainingDischargeTime` are not implementable on iOS and Android as system does not provide such information.


src/Uno.UWP/System/Power/PowerManager.cs
```diff
@@ -0,0 +1,112 @@
...
+			{
+				_batteryStatusChanged -= value;
+				if (_batteryStatusChanged == null)
+				{
+					EndBatteryStatusMonitoring();
+				}
+			}
+		}
+
+		internal static void RaiseRemainingChargePercentChanged() => _remainingChargePercentChanged?.Invoke(null, null);
```
davidjohnoliver: UWP returns null for args as well as sender?

src/Uno.UWP/System/Power/PowerManager.iOS.cs
```diff
@@ -0,0 +1,150 @@
...
+		private static NSObject _batteryStateChangeSubscription;
+		private static NSObject _powerStateChangeSubscription;
+
+		private static readonly UIDevice _device;
+		private static readonly bool _isSimulator;
+
+		static PowerManager()
+		{
+			_device = UIDevice.CurrentDevice;
+			_isSimulator = _device.Model == DeviceModelSimulator;
```
davidjohnoliver: Hmm what does `Model` return if you launch on a simulated iPad?

src/Uno.UWP/System/Power/PowerManager.iOS.cs
```diff
@@ -0,0 +1,150 @@
...
+
+		private static void StartBatteryStatusMonitoring() =>
+			StartIosBatteryStateMonitoring();
+
+		private static void EndBatteryStatusMonitoring() =>
+			EndIosBatteryStateMonitoring();
+
+		private static void StartIosBatteryStateMonitoring()
+		{
+			if (!_isSimulator) return;
```
davidjohnoliver:
```suggestion
			if (_isSimulator) return;
```

src/Uno.UWP/System/Power/PowerManager.iOS.cs
```diff
@@ -0,0 +1,150 @@
...
+			if (_batteryStatusChanged == null &&
+				_powerSupplyStatusChanged == null)
```
davidjohnoliver: seems like copy-paste of end condition

src/Uno.UWP/System/Power/Internal/BatteryChangedBroadcastReceiver.cs
```diff
@@ -0,0 +1,24 @@
...
+	internal class BatteryChangedBroadcastReceiver : BroadcastReceiver
+	{
+		public override void OnReceive(Context context, Intent intent)
+		{
+			if (Intent.ActionBatteryLow.Equals(
+				   intent.Action, StringComparison.InvariantCultureIgnoreCase) ||
+				Intent.ActionBatteryOkay.Equals(
+					intent.Action, StringComparison.InvariantCultureIgnoreCase))
+			{
+				Windows.System.Power.PowerManager.RaiseBatteryStatusChanged();
```
davidjohnoliver: Seems like this will raise BatteryStatusChanged a lot, if it's fired every time the charge percentage changes? Should it check if BatteryStatus is actually different?

src/Uno.UWP/System/Power/PowerManager.cs
MartinZikmund: Will check :-)

src/Uno.UWP/System/Power/PowerManager.iOS.cs
MartinZikmund: I originally found a resource that said there is only iPhone Simulator value returned, but now I found otherwise. Will fix

src/Uno.UWP/System/Power/PowerManager.iOS.cs
MartinZikmund: Yes, refactor typo... I need to retest iOS again.

src/Uno.UWP/System/Power/Internal/BatteryChangedBroadcastReceiver.cs
MartinZikmund: In fact Android does not offer continuous battery percentage monitoring, so it only reports for "low" and "okay" whatever it is. So it will actually fire very seldom. But I can add some "last value" cache (to all properties), so that the events actually trigger only when changes occur 👌👍

davidjohnoliver: Sounds good. That limitation then (and any others) should be documented as well. You can add an .md file in doc/articles/features/

MartinZikmund: I see, will do that as well. Thanks for the tips 😉 !

src/Uno.UWP/System/Power/PowerManager.cs
MartinZikmund: image

Seems that's the case for these events. `sender` is `null` as the class is static, and the second parameter is `null` as well :-D .

src/Uno.UWP/System/Power/Internal/BatteryChangedBroadcastReceiver.cs
MartinZikmund: Fixed :-) . Also improved to use a `lock` statement around event `add` and `remove` to avoid race conditions

src/Uno.UWP/System/Power/PowerManager.cs
```diff
@@ -0,0 +1,178 @@
+﻿#if __MOBILE__
```
jeromelaban: Let's use `__ANDROID__` and `__IOS__` explicitly, as we don't know how marzipan is going to be handled.

MartinZikmund: Done :-)

PR #884, opened 2019-05-22
VisualStateManager & Storyboard "Completed" callback fixes

carldebilly:
## Bugfix
The `Completed` callback of a `Storyboard` were not properly called
in various situations:
1. When there was no children in the Storyboard
2. When it was triggered using a StateTrigger

## What is the new behavior?
Behavior is now coherent with UWP.

## PR Checklist

- [X] [Unit Tests and/or UI Tests](doc/articles/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)


src/Uno.UI/UI/Xaml/AdaptiveTrigger.cs
```diff
@@ -22,25 +22,22 @@ private void UpdateState()
...
-			var isMinWidthSet = MinWindowWidth != -1;
-			var isMinHeightSet = MinWindowHeight != -1;
+			var w = size.Width;
+			var h = size.Height;
+			var mw = MinWindowWidth;
+			var mh = MinWindowHeight;
 
-			var widthIsActive = isMinWidthSet && size.Width >= MinWindowWidth;
-			var heightIsActive = isMinHeightSet && size.Height >= MinWindowHeight;
+			var isActive = (w > mw) && (h > mh);
```
dr1rrb: indeed condition is much simpler :) However not sure about the variables names

src/Uno.UI/UI/Xaml/Media/Animation/Storyboard.cs
```diff
@@ -165,15 +156,15 @@ public void Resume()
 					eventId: TraceProvider.StoryBoard_Resume,
 					opCode: EventOpcode.Send,
 					activity: _traceActivity,
-					payload: new[] { Target?.GetType().ToString(), PropertyInfo?.Path }
+					payload: new object[] { Target?.GetType().ToString(), PropertyInfo?.Path }
```
dr1rrb: Not sure, is this is really useful or it's only for readability ?

src/Uno.UI/UI/Xaml/VisualStateManager.cs
```diff
@@ -59,7 +59,7 @@ private static void OnVisualStateGroupsChanged(object sender, DependencyProperty
 				{
 					foreach (VisualStateGroup group in oldGroups)
 					{
-						group.SetParent(fe);
+						group.SetParent(null);
```
dr1rrb: 😱

src/Uno.UI/UI/Xaml/Media/Animation/Storyboard.cs
```diff
@@ -119,15 +102,24 @@ public void Begin()
...
+				foreach (ITimeline child in Children)
 				{
 					_runningChildren++;
 					child.Completed += Child_Completed;
 					child.Begin();
 				}
 			}
+			else
+			{
+				Dispatcher.RunAsync(CoreDispatcherPriority.Normal, () =>
```
davidjohnoliver: why does this need to be dispatched? (And if it does, the 'immediately' in the comment isn't accurate)

src/Uno.UI/UI/Xaml/AdaptiveTrigger.cs
```diff
@@ -22,25 +22,22 @@ private void UpdateState()
 		{
 			var size = Window.Current.Bounds;
 
-			var isMinWidthSet = MinWindowWidth != -1;
```
davidjohnoliver: this check isn't necessary?

carldebilly: Nope ;-)

1. It was comparing against scalar value, which is risky when using a floated point value
2. The check was useless in this circonstances and the result is the same.

src/Uno.UI/UI/Xaml/Media/Animation/Storyboard.cs
carldebilly: Using is not raising the `Completed` synchronously, so we're doing the same.

Here "Immediately" means we have nothing to wait for. But I agree it could be confusing.

src/Uno.UI/UI/Xaml/AdaptiveTrigger.cs
carldebilly: more tests has been added to check those cases.

Issue #907, opened 2019-05-23
Support Windows.System.Display.DisplayRequest


MartinZikmund:
# I'm submitting a...
 - Feature request

## Current behavior
`DisplayRequest` class not supported on Android and iOS

## Expected behavior
`DisplayRequest` class supported on Android and iOS 😉

## Environment

```
Nuget Package:

Package Version(s):

Affected platform(s):
- [ ] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] Windows
- [ ] Build tasks

Visual Studio
- [ ] 2017 (version: )
- [ ] 2019 (version: )
- [ ] for Mac (version: )

Relevant plugins
- [ ] Resharper (version: )

MartinZikmund:
I will implement this.

PR #908, opened 2019-05-23
Windows.System.Display.DisplayRequest

MartinZikmund:
GitHub Issue (If applicable): #907 

## PR Type

 - Feature -->

`Windows.System.Display.DisplayRequest` is not implemented

## What is the new behavior?
`Windows.System.Display.DisplayRequest` is implemented on iOS and Android. Its behavior matches UWP and it allows creating multiple instances as well as throws when release is attempted when there is no active request.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UWP/System/Display/DisplayRequest.cs
+++ b/src/Uno.UWP/System/Display/DisplayRequest.cs
@@ -1,0 +1,50 @@
+﻿#if __ANDROID__ || __IOS__
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Windows.System.Display
+{
+	public partial class DisplayRequest
+	{
+		private static object _syncLock = new object();
+		private static int _globalRequestCount = 0;
+
+		private int _instanceRequestCount = 0;
+
+		public void RequestActive()
+		{
+			lock (_syncLock)
+			{
+				_instanceRequestCount++;
+				_globalRequestCount++;
+				if (_globalRequestCount == 1)
+				{
+					//first global request, activate screen lock
+					ActivateScreenLock();
+				}
+			}
+		}
+
+		public void RequestRelease()
+		{
+			lock (_syncLock)
+			{
+				if (_instanceRequestCount == 0)
+					throw new InvalidOperationException("No active display request to release");
+				_instanceRequestCount--;
+				_globalRequestCount--;
+				if ( _globalRequestCount == 0)
+				{
+					//last global active request, deactivate
+					DeactivateScreenLock();
+				}
+			}
+		}
+
+		partial void ActivateScreenLock();
+
+		partial void DeactivateScreenLock();
+	}
+}
+#endif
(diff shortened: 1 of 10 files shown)


src/Uno.UWP/System/Display/DisplayRequest.cs
```diff
@@ -0,0 +1,50 @@
...
+
+		private int _instanceRequestCount = 0;
+
+		public void RequestActive()
+		{
+			lock (_syncLock)
+			{
+				_instanceRequestCount++;
+				_globalRequestCount++;
+				if (_globalRequestCount == 1)
```
carldebilly: I think there's a way to implement this in a lockless way using `Interlocked.Increment` / `Interlocked.Decrement`

MartinZikmund: Not sure how to approach this though - as I need to increment/decrement two variables, but the most important thing is that the `ActivateScreenLock` and `DeactivateScreenLock` must run only when the `_globalRequestCount` has an appropriate value. I think I can make sure to safely increment with `Interlocked`, but that won't solve the problem when I check `_globalRequestCount` value (in `if`), enter the block, and at that moment another thread changes the `_globalRequestCount` again, I may for example then deactivate the screen lock although another thread just activated it. Do you have an idea how to resolve this problem?

src/Uno.UWP/System/Display/DisplayRequest.cs
```diff
@@ -0,0 +1,50 @@
...
+					ActivateScreenLock();
+				}
+			}
+		}
+
+		public void RequestRelease()
+		{
+			lock (_syncLock)
+			{
+				if (_instanceRequestCount == 0)
```
jeromelaban: Missing braces

src/Uno.UWP/System/Display/DisplayRequest.cs
```diff
@@ -0,0 +1,50 @@
...
+					throw new InvalidOperationException("No active display request to release");
+				_instanceRequestCount--;
+				_globalRequestCount--;
+				if ( _globalRequestCount == 0)
```
jeromelaban: Formatting

src/Uno.UWP/System/Display/DisplayRequest.cs
MartinZikmund: Sorry, power of habit, at work we omit braces for single `throw` statements

src/Uno.UWP/System/Display/DisplayRequest.cs
MartinZikmund: Subtle, was confused where the problem is the first time I saw it :-D .

src/Uno.UWP/System/Display/DisplayRequest.cs
MartinZikmund: I think the `DisplayRequest` class will not commonly be used in high-perf scenarios, most apps will just use it once per launch, for example to keep screen on when playing video, so the cost of the `lock` feels manageable and it is easy to show that the code works correctly as compared to a more complex solution

PR #735, opened 2019-04-03
Fix Combobox, Popup & Flyout placements

carldebilly:
GitHub Issue (If applicable): #688

## Bugfix
ComboBox, Popup and Flyout were wrongly calculated on all platforms.

## Fixes/Changes:
* Flyout `Placement` is now working well on Wasm
* New samples for ComboBox (retrieved from private repo)
* Implemented LayoutInformation.GetLayoutSlot()
* Added support for `VisibleBounds` for Wasm (the value is initialized to browser size for now)
* Implemented `RectHelper`

## PR Checklist

Please check if your PR fulfills the following requirements:

- [X] Tested code with current [supported SDKs](../README.md#supported)
- [x] Docs have been added/updated which fit [documentation template](https://github.com/nventive/Uno/blob/master/doc/.feature-template.md). (for bug fixes / features)
- [X] [Unit Tests and/or UI Tests](doc/articles/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [X] Contains **NO** breaking changes
- [ ] Updated the [Release Notes](https://github.com/nventive/Uno/tree/master/doc/ReleaseNotes)
- [ ] Associated with an issue (GitHub or internal)

Internal Issue (If applicable):
* https://nventive.visualstudio.com/Umbrella/_workitems/edit/146096

PR #987, opened 2019-05-30
fixed button flyout placement for android

XiaotianNetlift:
## PR Type
- Bugfix

On Android, button flyout is placed lower than where it should be:
image

## What is the new behavior?
Button flyout is placed where it should be.

## PR Checklist
- [ ] ~~Docs have been added/updated which fit [documentation template](https://github.com/nventive/Uno/blob/master/doc/.feature-template.md). (for bug fixes / features)~~
- [ ] ~~[Unit Tests and/or UI Tests](doc/articles/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)~~
- [ ] ~~[Wasm UI Tests](doc/articles/working-with-the-samples-apps.md#running-the-webassembly-ui-tests-snapshots) are not showing unexpected any differences~~

## Other information
https://nventive.visualstudio.com/Umbrella/_workitems/edit/153874

In #735, the reference point was changed from
```
var targetTransform = (MatrixTransform)target.TransformToVisual(this);
var targetRect = new Rect(targetTransform.Matrix.OffsetX, targetTransform.Matrix.OffsetY, target.ActualWidth, target.ActualHeight);
```
to
```
var anchorRect = anchor.GetAbsoluteBoundsRect();
```

The former takes visible bounds into consideration and cancels it, while the later doesn't.
Thus the padding for status bar is applied twice:
- once from `FlyoutPopupPanel` (the container); and
- once again from `finalRect` because this is relative to `anchorRect` which is offsetted already

image
> While the shadowed area appears to be correct for `finalRect`, it was rendered as an absolute coordinates instead of being relative to its container.


src/Uno.UI/UI/Xaml/Controls/Flyout/FlyoutPopupPanel.cs
```diff
@@ -88,7 +88,8 @@ private Rect CalculateFlyoutPlacement(Size desiredSize)
 			}
 
 			var visibleBounds = ApplicationView.GetForCurrentView().VisibleBounds;
-			var anchorRect = anchor.GetAbsoluteBoundsRect();
+			var anchorTransform = (MatrixTransform)anchor.TransformToVisual(this);
```
carldebilly: The goal of the previous PR was to remove this cast.

The bug **must** be fixed in `.GetAbsoluteBoundsRect()`.

We had a hard time in the past because of those casts... we must remove them.

jeromelaban: Agreed. Let's find why this happens first.

dr1rrb: What you probably need if this https://github.com/nventive/Uno/blob/master/src/Uno.UI/UI/Xaml/Media/GeneralTransform.cs#L16 to get `Flyout` location 

Note: There is also https://github.com/nventive/Uno/blob/master/src/Uno.UI/UI/Xaml/Media/GeneralTransform.cs#L31 but I think that you don't want to apply the transform on the `Flyout` (for instance if you have a `ScaleX=.1` you probably don't want to scale down your flyout ... to validate with Windows :) )

XiaotianNetlift: image

`.GetAbsoluteBoundsRect()` returns a rect with the render size of the button and the delta between the on-screen locations of the button (`thisRect`) and `Window.Current.Content` (`otherRect`). This causes a problem because the `Window.Current.Content` doesn't align with the `FlyoutPopupPanel`, since the latter takes visible padding into consideration.

Therefore, instead of using `Window.Current.Content` as reference point, we should use `FlyoutPopupPanel`.

PR #936, opened 2019-05-27
Fix localization not applied if property is not set in Xaml

davidjohnoliver:
Ensure properties localized with x:Uid are applied even if set explicitly or implicitly in Xaml. (Xaml values are overridden)

GitHub Issue (If applicable): #

#921 

## PR Type
What kind of change does this PR introduce?

Bugfix

## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Tested code with current [supported SDKs](../README.md#supported)
- [ ] Docs have been added/updated which fit [documentation template](https://github.com/nventive/Uno/blob/master/doc/.feature-template.md). (for bug fixes / features)
- [x] [Unit Tests and/or UI Tests](doc/articles/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [ ] [Wasm UI Tests](doc/articles/working-with-the-samples-apps.md#running-the-webassembly-ui-tests-snapshots) are not showing unexpected any differences
- [ ] Contains **NO** breaking changes
- [x] Updated the [Release Notes](https://github.com/nventive/Uno/tree/master/doc/ReleaseNotes)
- [x] Associated with an issue (GitHub or internal)

## Other information

Internal Issue (If applicable):

PR #1024, opened 2019-06-05
fixed AppBarButton rendering on ios/android

XiaotianNetlift:
GitHub Issue (If applicable): #N/A

## PR Type
Bugfix

On Android, `AppBarButton.Label` takes precedence over `AppBarButton.Content` when used as `PrimaryCommands` which doesn't align with UWP.

## What is the new behavior?
On Android, `AppBarButton.Label` is ignored when used as `PrimaryCommands`.
Minor refactoring to ensure that the unused properties are properly cleared.

See tables below for the details of changes: `Rendered`(before this pr) vs `Rendered2`(this pr)
image
image
> `(undefined)` means the setter was not called.

## PR Checklist

- [ ] ~~[Unit Tests and/or UI Tests](doc/articles/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)~~
- [ ] ~~[Wasm UI Tests](doc/articles/working-with-the-samples-apps.md#running-the-webassembly-ui-tests-snapshots) are not showing unexpected any differences~~
- [ ] **Contains breaking changes**:
  > [Android] Unless nested under `SecondaryCommands`, the `AppBarButton.Label` property will no longer be used for the title of menu item, instead use the `AppBarButton.Content` property.

## Other information
With the changes in #936, localization can now be applied even if the property is not set in xaml. This caused some unintended side-effect with `AppBarButton` on android. When the `AppBarButton.Label` is localized in the localization (*.resw) files, it will now be set and will take precedence over `AppBarButton.Content`.

The `Label` keys were once used for menu-item tool-tip on android, and is still used for accessibility on ios, which could explain why they exist in the first place.

https://nventive.visualstudio.com/Umbrella/_workitems/edit/154794


davidjohnoliver:
> I think we should rename working-with-commandbar to commandbar.md and use only one file.
> 
> Other than that looks good!

Agreed, I prefer one file. Not to say it couldn't use some reorganization. A lot of links to internal UserEcho issues which should be removed.

Also agreed, nice work. :)

XiaotianNetlift:
Ok, I deleted my `CommandBar.md`, since `working-with-commandbar.md` already contains the same info and much more.
I did some cleaned up in the other md too:
- moved **Known Issues** above **FAQ** to the bottom
- removed **Remarks** section of **FAQ** that links to the internal user-echo
- converted the tier-2 headers (`##`) under **FAQ** to list
- code alignment in snippets

Issue #808, opened 2019-04-24
wasm: button content sized incorrectly in ListView ItemTemplate


yowl:
## I'm submitting a...

- - Bug report (I searched for similar issues and did not find one) 
 

## Current behavior
wasm buttons in ListView ItemTemplate seems to have their content width calculated incorrectly

## Expected behavior
Should looks the same as UWP and the button content should be allowed to expand to the size of the button as it does when not in a ListView ItemTemplate

## Minimal reproduction of the problem with instructions
wasm:
```
<Page
    x:Class="UnoStandardRefTest2.MainPage"
    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
    xmlns:local="using:UnoStandardRefTest2"
    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
    mc:Ignorable="d">
    <Page.Resources>
        <DataTemplate x:Key="dataTemplate">
            <Button Margin="5" Height="50" Width="250">
                <Button.Content>
                    <TextBlock Text="Wrap text only when long" TextWrapping="Wrap"/>
                </Button.Content>
            </Button>
        </DataTemplate>
    </Page.Resources>
    <StackPanel Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">
        <TextBlock Text="Hello, world !" Margin="20" FontSize="30" />
        <Button Margin="5" Height="50" Width="250">
            <Button.Content>
                <TextBlock Text="Wrap text only when long" TextWrapping="Wrap"/>
            </Button.Content>
        </Button>
        <ListView ItemsSource="{Binding Items}" ItemTemplate="{StaticResource dataTemplate}">
        </ListView>
    </Stack
```
Code behind:
```
    public sealed partial class MainPage : Page
    {
        public MainPage()
        {
            Items = Enumerable.Range(0, 5).Select(i => i.ToString()).ToList();
            //            Items = Enumerable.Range(0, 2000).Select(i => i.ToString()).ToList();
            this.InitializeComponent();
            DataContext = this;
        }

        public List<string> Items { get; set; }
    }
```
repo:https://github.com/yowl/UnoStandardRefTest2/tree/buttoninitemtemplate

## Environment

```
Nuget Package: 
    <PackageReference Include="Uno.UI" Version="1.45.0-dev.1333" />
[...]

PR #1034, opened 2019-06-07
[Wasm] Text measurement bug when ancestor contains a RenderTransform

carldebilly:
GitHub Issue (If applicable): #808 

## Bugfix
Text measurement refactoring.

Sometimes a TextBlock were wrapping even if the parent were provinding enough space to prevent it.

The problem was caused by a parent element containing a _transform_ in the HTMLDOM. The cause
is directly in the spec itself:

<https://www.w3.org/TR/css-transforms-1/>
> For elements whose layout is governed by the CSS box model, any value other than none
> for the transform property also causes the element to establish a containing block for all
> descendants. Its padding box will be used to layout for all of its absolute-position
> descendants, fixed-position descendants, and descendant fixed background attachments.

**BONUS**: The investigations also revealed another one, never reported: a TextBlock width was hard limited to viewport width.

## What is the new behavior?
This limitation is now used to remove any kind of [known yet] DOM ancestor constrains.
The parent element is now set to almost-infinite size to remove any size contention.

## PR Checklist

- [X] Docs have been added/updated which fit [documentation template](https://github.com/nventive/Uno/blob/master/doc/.feature-template.md). (for bug fixes / features)
- [X] [Unit Tests and/or UI Tests](doc/articles/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)

* https://nventive.visualstudio.com/Umbrella/_workitems/edit/154814

### DIFF
--- a/doc/ReleaseNotes/_ReleaseNotes.md
+++ b/doc/ReleaseNotes/_ReleaseNotes.md
@@ -80,6 +80,7 @@
 	- [iOS] Add support of `INotifyCollectionChanged` in the `Picker`
 	- [iOS] Remove the arbitrary `null` item added at the top of the `Picker`
 	- [iOS] Fix infinite layouting cycle in the iOS picker (Removed workaround which is no longer necessary as the given method is invoked properly on each measure/arrange phases)
+* [Wasm] Refactored the way the text is measured in Wasm. Wasn't working well when a parent with a RenderTransform.
 
 ### Breaking Changes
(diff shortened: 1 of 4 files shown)


jeromelaban:
There's a rendering difference in `SamplesApp.Windows_UI_Xaml_Controls.ListView.ListViewChangeHeight.png`:
image

is that expected ?

carldebilly:
@jeromelaban It is expected, but I'll check to make sure the result is what we want.  Good catch!

PR #1090, opened 2019-06-18
Fix and enhance the combo box drop down placement

dr1rrb:
- Improve the top/centered/bottom placement logic to be closer to UWP
- Avoid flicker when opening the drop down

GitHub Issue (If applicable):
https://github.com/nventive/calculator/issues/145
https://github.com/nventive/calculator/issues/146

## Bugfix
Improve the ComboBox' s drop down placement

- The first time that a ComboBox which should render its dorp down above it is opened on WASM, the drop down displays only one item far from the combo itself (Weird behaviuor like that also occurs when we resize the window)
- When we re-open it later, we sometimes still observe some flickers
- The drop down Height is hard coded to be at max 60% of the screen height

## What is the new behavior?
- The drop down is always placed at the right location
- The height of drop down is now at max 9 time the height of the `ComboBox` (approximation of the UWP behavior which sizes the drop down to display 9 items)
- When the drop down is opened, its location is now customized accordingly to the `SelectedItem` (closer to the UWP behavior which aligns the selected item of the drop down with combo itself)
  * If the selected item is close the the begining of the collection, drop down will appear below the combo
  * If the selected item is close the the end of the collection, drop down will appear above the combo
  * Otherwise the drop down will appear centered

## PR Checklist
- [x] [Wasm UI Tests](doc/articles/working-with-the-samples-apps.md#running-the-webassembly-ui-tests-snapshots) are not showing unexpected any differences
- ~~[ ] Contains **NO** breaking changes~~

## Breaking
On Android the `Popup.Anchor` is now internal (was public)


src/Uno.UI/UI/Xaml/Controls/ComboBox/ComboBox.cs
```diff
@@ -50,10 +52,20 @@ protected override void OnApplyTemplate()
...
+				oldPopup.CustomLayouter = null;
+			}
+
 			_popup = this.GetTemplateChild("Popup") as IPopup;
 			_popupBorder = this.GetTemplateChild("PopupBorder") as Border;
 			_contentPresenter = this.GetTemplateChild("ContentPresenter") as ContentPresenter;
 
+			if (_popup is PopupBase popup)
+			{
+				popup.CustomLayouter = new DropDownLayouter(this, popup);
```
carldebilly: You need to create a new one each time?

dr1rrb: As it has a reference to `this` and the `popup` yes. We could find some way to avoid that, but the cost is really small, especially as we usually don't change the template of controls.

PR #1100, opened 2019-06-19
Check for dataContext to avoid recyclingTemplate when unneeded

NicolasChampagne:
## PR Type
 Bugfix 

Would recycle template even when unneeded

## What is the new behavior?
Won't recycle template even when unneeded

## PR Checklist

https://nventive.visualstudio.com/Umbrella/_workitems/edit/144025


src/Uno.UI/UI/Xaml/FrameworkTemplatePool.cs
```diff
@@ -240,9 +240,12 @@ private void OnParentChanged(object instance, object key, DependencyObjectParent
 
 		internal static void PropagateOnTemplateReused(object instance)
 		{
-			if (instance is IFrameworkTemplatePoolAware a)
+			if (instance is IFrameworkTemplatePoolAware a && instance is FrameworkElement b)
```
davidjohnoliver: Give these variables more meaningful names

src/Uno.UI/UI/Xaml/FrameworkTemplatePool.cs
```diff
@@ -240,9 +240,12 @@ private void OnParentChanged(object instance, object key, DependencyObjectParent
...
 			{
-				a.OnTemplateRecycled();
+				if (b.DataContext == null)
```
davidjohnoliver:
```suggestion
				// If DataContext is not null, it means it has been explicitly set (not inherited). Resetting the view could push an invalid value through 2-way binding in this case.
				if (b.DataContext == null)
```

src/Uno.UI/UI/Xaml/FrameworkTemplatePool.cs
davidjohnoliver: Use `IFrameworkElement` rather than `FrameworkElement`

src/Uno.UI/UI/Xaml/FrameworkTemplatePool.cs
jeromelaban: This is a case that may need to be tested by a unit test. Also, testing for null may require determining the precedence of the value.

davidjohnoliver: Agreed for the test. 

@jeromelaban  Can you elaborate on why we would test for precedence in this case? At this point the template is detached from the visual tree, so if `DataContext` is non-null then we know it has been explicitly set.

NicolasChampagne:
@davidjohnoliver any other necessary changes?

davidjohnoliver:
> @davidjohnoliver any other necessary changes?

Jerome's comment:
> This is a case that may need to be tested by a unit test.

src/Uno.UI/UI/Xaml/FrameworkTemplatePool.cs
```diff
@@ -240,9 +240,10 @@ private void OnParentChanged(object instance, object key, DependencyObjectParent
 
 		internal static void PropagateOnTemplateReused(object instance)
 		{
-			if (instance is IFrameworkTemplatePoolAware a)
+			// If DataContext is not null, it means it has been explicitly set (not inherited). Resetting the view could push an invalid value through 2-way binding in this case.
```
carldebilly: What happen if the DataContext is set directly on one of the children?

davidjohnoliver: `OnTemplateRecycled` is called individually for each view in the subtree, so it shouldn't matter.

NicolasChampagne:
Bad merge, dropping pr

PR #1213, opened 2019-07-09
docs: information where to get pre-release builds

ghuntley:
## PR Type


doc/articles/contributing/pre-release-builds.md
```diff
@@ -0,0 +1,4 @@
+# pre-release builds
```
ghuntley: 🎗 needs to be added to the toc. this could be a section within an existing document.

PR #1190, opened 2019-07-04
Fix possible crash with AudioToolbox

dr1rrb:
## Bug fix tentative
**Tries** to prevent crash: 
```
Crashed: AVAudioSession Notify Thread
0  AudioToolbox                   0x1a96762c0 AudioSessionPropertyListeners::CallPropertyListenersImp(AudioSessionPropertyListeners const&, unsigned int, unsigned int, void const*) + 456
1  AudioToolbox                   0x1a976c100 AudioSessionPropertyListeners::CallPropertyListeners(unsigned int, unsigned int, void const*) + 220
2  AudioToolbox                   0x1a97aa43c HandleCFPropertyListChange(unsigned int, unsigned int, unsigned long, unsigned char*, unsigned int) + 456
3  AudioToolbox                   0x1a97aea34 HandleAudioSessionCFTypePropertyChangedMessage(unsigned int, unsigned int, void*, unsigned int) + 268
4  AudioToolbox                   0x1a97adcf8 ProcessDeferredMessage(unsigned int, __CFData const*, unsigned int, unsigned int) + 1312
5  AudioToolbox                   0x1a97ad4e4 ASCallbackReceiver_AudioSessionPingMessage + 588
6  AudioToolbox                   0x1a9666ec0 _XAudioSessionPingMessage + 52
7  AudioToolbox                   0x1a99ae1a8 mshMIGPerform + 232
8  CoreFoundation                 0x1a56af690 __CFRUNLOOP_IS_CALLING_OUT_TO_A_SOURCE1_PERFORM_FUNCTION__ + 56
9  CoreFoundation                 0x1a56aeddc __CFRunLoopDoSource1 + 440
10 CoreFoundation                 0x1a56a9c00 __CFRunLoopRun + 2096
11 CoreFoundation                 0x1a56a90b0 CFRunLoopRunSpecific + 436
12 AVFAudio                       0x1ab591334 GenericRunLoopThread::Entry(void*) + 156
13 AVFAudio                       0x1ab5bbc60 CAPThread::Entry(CAPThread*) + 88
14 libsystem_pthread.dylib        0x1a533c2c0 _pthread_body + 128
15 libsystem_pthread.dylib        0x1a533c220 _pthread_start + 44
16 libsystem_pthread.dylib        0x1a533fcdc thread_start + 4
```

## PR Checklist
- [x] Tested code with current [supported SDKs](../README.md#supported)
[...]

PR #1214, opened 2019-07-09
Complete the tentative to avoid crash in AudioToolBox

dr1rrb:
This PR is the same as https://github.com/unoplatform/uno/pull/1190, it only adds missed callbacks


src/Uno.UWP/Media/Playback/MediaPlayer.iOS.cs
```diff
@@ -87,6 +91,18 @@ public override void ObserveValue(NSString keyPath, NSObject ofObject, NSDiction
 						return;
 				}
 			}
+
+			public EventHandler<AVPlayerItemErrorEventArgs> OnMediaFailed { get; }
+			private void OnMediaFailedCore(object sender, AVPlayerItemErrorEventArgs args)
```
jeromelaban: Explain why this is indirect.

PR #1043, opened 2019-06-10
Misc fixes for the ComboBox

dr1rrb:
- Remove dependency to a "Background" template part which is unnecessary and not required on UWP
- Make sure that the `PopupPanel` hides itself if collapsed (special cases as it's at the top of the `Window`)
- [iOS] Add support of `INotifyCollectionChanged` in the `Picker`
- [iOS] Remove the arbitrary `null` item added at the top of the `Picker`
- [iOS] Fix infinite layouting cycle in the iOS picker (Removed workaround which is no longer necessary as the given method is invoked properly on each measure/arrange phases)

## PR Type
What kind of change does this PR introduce?
- Bugfix

## What is the current behavior?
The opposite to what is described in commit comment above

## What is the new behavior?
What is described in commit comment above

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] Tested code with current [supported SDKs](../README.md#supported)
- [ ] Docs have been added/updated which fit [documentation template](https://github.com/nventive/Uno/blob/master/doc/.feature-template.md). (for bug fixes / features)
- [ ] [Unit Tests and/or UI Tests](doc/articles/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [ ] [Wasm UI Tests](doc/articles/working-with-the-samples-apps.md#running-the-webassembly-ui-tests-snapshots) are not showing unexpected any differences
- [x] Contains **NO** breaking changes
- [x] Updated the [Release Notes](https://github.com/nventive/Uno/tree/master/doc/ReleaseNotes)
- [x] Associated with an issue (GitHub or internal)

## Other information
Internal Issue (If applicable):
https://dev.azure.com/nventive/_workitems/edit/155101
https://dev.azure.com/nventive/_workitems/edit/155106
https://dev.azure.com/nventive/_workitems/edit/155140

PR #1226, opened 2019-07-11
Fix to iOS ComboBox initial selected value not null/empty

NicolasChampagne:
## PR Type

- Bugfix

iOS initial value is the first item in the list 

## What is the new behavior?
iOS initial value is null

## PR Checklist

https://nventive.visualstudio.com/Umbrella/_workitems/edit/157456

### DIFF
--- a/src/SamplesApp/SamplesApp.UITests/Windows_UI_Xaml_Controls/ComboBoxTests/UnoSamples_Tests.ComboBoxTests.cs
+++ b/src/SamplesApp/SamplesApp.UITests/Windows_UI_Xaml_Controls/ComboBoxTests/UnoSamples_Tests.ComboBoxTests.cs
@@ -1,18 +1,33 @@
-﻿using System;
+﻿using NUnit.Framework;
+using SamplesApp.UITests.TestFramework;
+using System;
 using System.Collections.Generic;
+using System.Drawing;
 using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
-using NUnit.Framework;
 using Uno.UITest.Helpers;
+using Uno.UITest.Helpers.Queries;
 
-namespace SamplesApp.UITests
+namespace SamplesApp.UITests.Windows_UI_Xaml_Controls.ComboBoxTests
 {
 	[TestFixture]
-	partial class UnoSamples_Tests : SampleControlUITestBase
+	public partial class ComboBoxTests_Tests : SampleControlUITestBase
 	{
 		[Test]
-		public void ComboBox_Kidnapping()
+		[ActivePlatforms(Platform.iOS)]
+		public void ComboBoxTests_PickerDefaultValue()
+		{
+			Run("SamplesApp.Wasm.Windows_UI_Xaml_Controls.ComboBox.ComboBox_Picker");
+
+			_app.WaitForElement(_app.Marked("theComboBox"));
+			var theComboBox = _app.Marked("theComboBox");
+
+			Assert.IsNull(theComboBox.GetDependencyPropertyValue("SelectedItem"));
+		}
+
+		[Test]
+		public void ComboBoxTests_Kidnapping()
 		{
 			Run("UITests.Shared.Windows_UI_Xaml_Controls.ComboBox.ComboBox_ComboBoxItem_Selection");
 
@@ -26,16 +41,16 @@ namespace SamplesApp.UITests
 			_app.Tap(button);
 
 			var first = _app.FindWithin("_tb1", presenter);
-			Assert.AreEqual("Item 1", GetText(first));
+			Assert.AreEqual("Item 1", first.GetDependencyPropertyValue<string>("Text"));
 
 			_app.Tap(comboBox);
 			_app.TapCoordinates(300, 100);
 			first = _app.FindWithin("_tb1", presenter);
-			Assert.AreEqual("Item 1", GetText(first));
+			Assert.AreEqual("Item 1", first.GetDependencyPropertyValue<string>("Text"));
 
 			_app.Tap(button);
 			var second = _app.FindWithin("_tb2", presenter);
-			Assert.AreEqual("Item 2", GetText(second));
+			Assert.AreEqual("Item 2", second.GetDependencyPropertyValue<string>("Text"));
 
 			Configuration.AttemptToFindTargetBeforeScrolling = scrollingInitial;
 		}
--- a/src/Uno.UI/UI/Xaml/Controls/Picker/Picker.iOS.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Picker/Picker.iOS.cs
@@ -63,7 +63,7 @@ namespace Windows.UI.Xaml.Controls
 		}
 
 		public object[] Items { get; private set; } = new object[] { null }; // ensure there's always a null present to allow deselection
-		
+
 		partial void OnItemsSourceChangedPartialNative(object oldItemsSource, object newItemsSource)
 		{
 			if (oldItemsSource is INotifyCollectionChanged oldObservableCollection)
@@ -81,7 +81,17 @@ namespace Windows.UI.Xaml.Controls
 
 		private void OnItemSourceChanged(object collection, NotifyCollectionChangedEventArgs _)
 		{
-			Items = (collection as IEnumerable)?.ToObjectArray() ?? new object[0];
+			if (SelectedItem == null)
+			{
+				Items = new[] { (object)null }
+				  .Concat((collection as IEnumerable)?.ToObjectArray() ?? new object[0])
+				  .ToObjectArray();
+			}
+			else
+			{
+				Items = (collection as IEnumerable)?.ToObjectArray() ?? new object[0];
+			}
+
 			ReloadAllComponents();
 
 			if (!Items.Contains(SelectedItem))
@@ -101,6 +111,13 @@ namespace Windows.UI.Xaml.Controls
 					SelectedItem = firstItem;
 					return;
 				}
+			}
+			else if (newSelectedItem != null && Items[0] == null)
+			{
+				// On ItemSelection remove initial null item at Items[0]
+				Items = Items
+					.Skip(1)
+					.ToObjectArray();
 			}
 
 			Select(row, component: 0, animated: true);
--- a/build/android-uitest-run.sh
+++ b/build/android-uitest-run.sh
@@ -15,10 +15,6 @@ nohup $ANDROID_HOME/emulator/emulator -avd xamarin_android_emulator -skin 1280x8
 
 export IsUiAutomationMappingEnabled=true
 
-export IsUiAutomationMappingEnabled=true
-
-export IsUiAutomationMappingEnabled=true
-
 # build the sample and tests, while the emulator is starting
 msbuild /r /p:Configuration=Release $BUILD_SOURCESDIRECTORY/src/SamplesApp/SamplesApp.Droid/SamplesApp.Droid.csproj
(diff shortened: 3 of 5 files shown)


yowl:
Is this actually the UWP behavior?

Take this xaml:
```
        <ComboBox ItemsSource="{Binding Items}" SelectedItem="{Binding SelectedItem, Mode=TwoWay}">
        </ComboBox>
```
and bind to this:

```
    public sealed partial class MainPage : Page, INotifyPropertyChanged
    {
        List<string> items = new List<string>{"a", "b"};
        object selectedItem;

        public MainPage()
        {
            this.InitializeComponent();
            DataContext = this;
            SelectedItem = "a";
            Items = new List<string> { "c", "b" };
        }

        public List<string> Items
        {
            get { return items; }
            set
            {
                items = value;
                OnPropertyChanged();
            }
        }

        public object SelectedItem
        {
            get => selectedItem;
            set
            {
                selectedItem = value;
                OnPropertyChanged();
            }
        }

        public event PropertyChangedEventHandler PropertyChanged;

        void OnPropertyChanged([CallerMemberName] string propertyName = null)
        {
            PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(propertyName));
        }
    }
```
Run it and the combo is empty, it doesn't show the first item.

NicolasChampagne:
> Is this actually the UWP behavior?
> 
> Take this xaml:
> 
> ```
>         <ComboBox ItemsSource="{Binding Items}" SelectedItem="{Binding SelectedItem, Mode=TwoWay}">
>         </ComboBox>
> ```
> 
> and bind to this:
> 
> ```
>     public sealed partial class MainPage : Page, INotifyPropertyChanged
>     {
>         List<string> items = new List<string>{"a", "b"};
>         object selectedItem;
> 
>         public MainPage()
>         {
>             this.InitializeComponent();
>             DataContext = this;
>             SelectedItem = "a";
>             Items = new List<string> { "c", "b" };
>         }
> 
>         public List<string> Items
>         {
>             get { return items; }
>             set
>             {
>                 items = value;
>                 OnPropertyChanged();
>             }
>         }
> 
>         public object SelectedItem
>         {
>             get => selectedItem;
>             set
>             {
>                 selectedItem = value;
>                 OnPropertyChanged();
>             }
>         }
> 
>         public event PropertyChangedEventHandler PropertyChanged;
> 
>         void OnPropertyChanged([CallerMemberName] string propertyName = null)
>         {
>             PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(propertyName));
>         }
>     }
> ```
> 
> Run it and the combo is empty, it doesn't show the first item.

Yes that's exactly it, in iOS the bug was that the initial was the first item

src/SamplesApp/SamplesApp.UITests/Windows_UI_Xaml_Controls/ComboBoxTests/UnoSamples_Tests.ComboBoxTests.cs
```diff
@@ -0,0 +1,36 @@
...
+		public void ComboBoxTests_DefaultValue()
+		{
+			Run("SamplesApp.Wasm.Windows_UI_Xaml_Controls.ComboBox.ComboBox_ItemsSource");
+
+			_app.WaitForElement(_app.Marked("theComboBox"));
+			var theComboBox = _app.Marked("theComboBox");
+
+			// Assert initial state
+			var myDevice = _app.Device.GetType();
+			if (_app.Device.GetType().Name.Contains("Android") || _app.Device.GetType().Name.Contains("iOS"))
```
davidjohnoliver: It seems strange that you need this, is this because Uno behaves differently on WASM, or because the `GetDependencyPropertyValue()` method behaves differently? Can you add a comment explaining please.

src/SamplesApp/SamplesApp.UITests/Windows_UI_Xaml_Controls/ComboBoxTests/UnoSamples_Tests.ComboBoxTests.cs
```diff
@@ -0,0 +1,36 @@
...
+
+namespace SamplesApp.UITests.Windows_UI_Xaml_Controls.ComboBoxTests
+{
+	[TestFixture]
+	public partial class ComboBoxTests_Tests : SampleControlUITestBase
+	{
+		[Test]
+		public void ComboBoxTests_DefaultValue()
+		{
+			Run("SamplesApp.Wasm.Windows_UI_Xaml_Controls.ComboBox.ComboBox_ItemsSource");
```
davidjohnoliver: How can this test be testing the change you made? This sample isn't using the `Picker`.

src/Uno.UI/UI/Xaml/Controls/Picker/Picker.iOS.cs
```diff
@@ -84,7 +84,7 @@ private void OnItemSourceChanged(object collection, NotifyCollectionChangedEvent
 			Items = (collection as IEnumerable)?.ToObjectArray() ?? new object[0];
 			ReloadAllComponents();
 
-			if (!Items.Contains(SelectedItem))
+			if (Items[0] != null && !Items.Contains(SelectedItem))
```
davidjohnoliver: This is going to throw an exception if Items is empty. 

Also, I really can't tell what it's intended to achieve.

NicolasChampagne: If collection is empty first items[0] is set to null but items is never null, therefore there will never be an exception thrown.

The behaviour is that when a combobox is initialized the initial selected item is isnt the first item in the collection. Like in other platforms

NicolasChampagne:
Bad Merge

src/Uno.UI/UI/Xaml/Controls/Picker/Picker.iOS.cs
davidjohnoliver: Can you show me where in the code it's ensured that `Items` is never empty? I can see it's initialized as `{ null }`, but it seems like all bets are off after `ItemsSource` changes.

I'm guessing the regression appeared because this line was recently removed:

```csharp
Items = new[] { (object)null }.Concat(items).ToArray(); // ensure there's always a null present to allow deselection
```

Did you confirm with @dr1rrb that it wasn't removed accidentally?

src/Uno.UI/UI/Xaml/Controls/Picker/Picker.iOS.cs
```diff
@@ -81,7 +81,17 @@ public override CGSize SizeThatFits(CGSize size)
 
 		private void OnItemSourceChanged(object collection, NotifyCollectionChangedEventArgs _)
 		{
-			Items = (collection as IEnumerable)?.ToObjectArray() ?? new object[0];
+			if (Items == null)
```
davidjohnoliver: shouldn't it rather be checking if `SelectedItem` is null?

src/Uno.UI/UI/Xaml/Controls/Picker/Picker.iOS.cs
```diff
@@ -102,6 +112,10 @@ private void OnItemSourceChanged(object collection, NotifyCollectionChangedEvent
 					return;
 				}
 			}
+			else if (newSelectedItem != null && Items[0] == null)
+			{
+				Items = Items.Where(c => c != null).ToObjectArray();
```
davidjohnoliver: Need comment here explaining why this code is needed.

src/Uno.UI/UI/Xaml/Controls/Picker/Picker.iOS.cs
```diff
@@ -102,6 +112,11 @@ private void OnItemSourceChanged(object collection, NotifyCollectionChangedEvent
 					return;
 				}
 			}
+			else if (newSelectedItem != null && Items[0] == null)
+			{
+				// On ItemSelection remove initial null item at Items[0]
+				Items = Items.Where(c => c != null).ToObjectArray();
```
jeromelaban: You'll probably want to use Skip(1), otherwise this may remove all null values.

src/Uno.UI/UI/Xaml/Controls/Picker/Picker.iOS.cs
```diff
@@ -81,7 +81,17 @@ public override CGSize SizeThatFits(CGSize size)
 
 		private void OnItemSourceChanged(object collection, NotifyCollectionChangedEventArgs _)
 		{
-			Items = (collection as IEnumerable)?.ToObjectArray() ?? new object[0];
+			if (SelectedItem == null)
+			{
+				Items = new[] { (object)null }
+				  .Union((collection as IEnumerable)?.ToObjectArray() ?? new object[0])
```
jeromelaban: This is most likely a `.Concat()` that you need here.

src/SamplesApp/SamplesApp.UITests/Windows_UI_Xaml_Controls/ComboBoxTests/UnoSamples_Tests.ComboBoxTests.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.Threading.Tasks;
+using Uno.UITest.Helpers;
+using Uno.UITest.Helpers.Queries;
+
+namespace SamplesApp.UITests.Windows_UI_Xaml_Controls.ComboBoxTests
+{
+	[TestFixture]
+	public partial class ComboBoxTests_Tests : SampleControlUITestBase
+	{
+		[Test]
```
jeromelaban: To run the test on iOS only, use the `ActivePlatform` attribute: 
https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md#selectively-ignore-tests-per-platform

(It's been introduced after you submitted your PR)

src/Uno.UI/UI/Xaml/Controls/Picker/Picker.iOS.cs
dr1rrb: As discussed yesterday, yes it was intentional because it was allowing the user to select `null` which is not possible on UWP (and maybe even erasing the selected value on loaded, not sure, to validate). If we want to avoid auto select of the first value, we should insert a `null` only until a selection is made, and then remove it.

dr1rrb:
Something bugs me ... does the issue was not introduced by this line https://github.com/unoplatform/uno/commit/d4291271#diff-809c81ed5e6dc3663228d12c3fff1ceaR45 ?

davidjohnoliver:
> Something bugs me ... does this issue was not introduced by this line [d429127#diff-809c81ed5e6dc3663228d12c3fff1ceaR45](https://github.com/unoplatform/uno/commit/d4291271#diff-809c81ed5e6dc3663228d12c3fff1ceaR45) ?

It's worth to double-check, but note that this method is only called from the `ctor`.

NicolasChampagne:
> > Something bugs me ... does this issue was not introduced by this line [d429127#diff-809c81ed5e6dc3663228d12c3fff1ceaR45](https://github.com/unoplatform/uno/commit/d4291271#diff-809c81ed5e6dc3663228d12c3fff1ceaR45) ?
> 
> It's worth to double-check, but note that this method is only called from the `ctor`.

When first investigating this issue the regression only showed up after @dr1rrb changes with pr: https://github.com/unoplatform/uno/pull/1043

NicoChampagne:
> /azp run

jeromelaban:
Can you rebase your PR on master? The history contains unrelated changes.

NicolasChampagne:
> Make sure to rebase as Jerome mentioned, there shouldn't be anything in the 'Commits' tab that isn't associated with the PR.

I know I've revised on master and this was the result, I'll try once again

NicolasChampagne:
> Can you rebase your PR on master? The history contains unrelated changes.
done

Issue #1167, opened 2019-06-30
[Wasm] Disabled textbox can be edited


Wouterdek:
## On which page?
https://playground.platform.uno/#textbox

## What's wrong?
I can edit the second, disabled, textbox by selecting the first one and then pressing tab. I would expect pressing tab to focus the next enabled element. I'm using Firefox 67.0.4 on Windows.

## Any feedback?
I'm not sure whether this is a sample issue or a framework bug.

jeromelaban:
Thank you for the report!

I can confirm this issue happens on uno latest master.

ghuntley:
@Wouterdek we would welcome a contribution that fixes this. Can you send in a pull-request? 👍

Wouterdek:
Well, the issue seems to be that the HTML `disabled` attribute is not applied to the element even though `IsEnabled="False"` was specified in XAML. After a quick inspection it seems to me like [this line](https://github.com/unoplatform/uno/blob/7f4938b91ee3f172573611dcfcbde2cf1209e43c/src/Uno.UI/UI/Xaml/Controls/TextBox/TextBoxView.wasm.cs#L86) might have its values the wrong way around? It should be that `IsEnabled` <=> `!disabled`. There must be another error though, as all `IsEnabled="True"` textboxes do work at the moment, so I suspect the disabled attribute is simply never applied at all. However, I'm not familiar with the framework (literally saw this project first 4 days ago and tried the demo out of curiosity) so I can't really work it out further than that.

jeromelaban:
@Wouterdek thanks for the troubleshooting. Indeed that property is inverted, but it also not applied, most probably caused by some `IsEnabled` inheritance issue.

davidjohnoliver:
![](https://github.trello.services/images/mini-trello-icon.png) [WASM Incorrectly disabled TextBox https://github.com/unoplatform/uno/issues/1167](https://trello.com/c/isRXnzmT/35-wasm-incorrectly-disabled-textbox-https-githubcom-unoplatform-uno-issues-1167)

davidjohnoliver:
The Uno bug is fixed by #1295, once Uno.UI is updated in Uno.Playground I'll close this issue.

carldebilly:
Has been fixed by PR #1799

PR #1295, opened 2019-07-23
[WASM] Apply TextBox.IsEnabled correctly

davidjohnoliver:
GitHub Issue (If applicable): #1167

## PR Type

- Bugfix

Disabled TextBox could be tabbed to.

## What is the new behavior?

Disabled TextBox is really, truly disabled

## PR Checklist

## Other information

### DIFF
--- a/src/SamplesApp/SamplesApp.UITests/UnoSamples_Tests.Keyboard.cs
+++ b/src/SamplesApp/SamplesApp.UITests/UnoSamples_Tests.Keyboard.cs
@@ -369,6 +369,7 @@ namespace SamplesApp.UITests
 		}
 
 		[Test]
+		[AutoRetry]
 		public void TextBox_Disable()
 		{
 			Run("UITests.Shared.Windows_UI_Xaml_Controls.TextBoxTests.TextBox_Disabled");
--- a/src/SamplesApp/SamplesApp.UITests/UnoSamples_Tests.Keyboard.cs
+++ b/src/SamplesApp/SamplesApp.UITests/UnoSamples_Tests.Keyboard.cs
@@ -368,6 +368,21 @@ namespace SamplesApp.UITests
 			Assert.AreEqual(Final2, GetText(textBlock));
 		}
 
+		[Test]
+		public void TextBox_Disable()
+		{
+			Run("UITests.Shared.Windows_UI_Xaml_Controls.TextBoxTests.TextBox_Disabled");
+
+			const string Entered = "Catfish";
+			const string Final = "Cat";
+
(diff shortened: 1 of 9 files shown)


src/SamplesApp/SamplesApp.UITests/UnoSamples_Tests.Keyboard.cs
```diff
@@ -365,6 +365,21 @@ public void TextBox_BeforeTextChanging()
 			Assert.AreEqual(Final2, GetText(textBlock));
 		}
 
+		[Test]
```
carldebilly: `[AutoRetry]` ? ;-)

src/Uno.UI.Wasm/ts/WindowManager.ts
```diff
@@ -372,7 +372,16 @@
 
 			for (const name in properties) {
 				if (properties.hasOwnProperty(name)) {
-					(element as any)[name] = properties[name];
+					var setVal = properties[name];
+					if (setVal === "true") {
+						(element as any)[name] = true;
```
carldebilly: What happen if we try to set the string value `"true"` here?
+ maybe we should handle `null` the same way too, no?

davidjohnoliver: * Setting `input["disabled"]` to any string (including `"false"`) is equivalent to setting it to `true`, ie the `TextBoxView` stays disabled.
 * There may be other cases where we want to convert strings to literals, yes. Great candidate for a separate PR. ;)

src/SamplesApp/SamplesApp.UITests/UnoSamples_Tests.Keyboard.cs
davidjohnoliver: Why?

jeromelaban: It’s now a requirement, tests will fail if the attribute is not present.

davidjohnoliver: ...why??

carldebilly: Because tests on CI are not stable enough and it's a way to minimize broken builds.

davidjohnoliver: Sure but it doesn't make much sense to have an attribute in this scenario - just retry all the tests.

Issue #1135, opened 2019-06-26
Support Windows.UI.StartScreen.JumpList


MartinZikmund:
## What would you like to be added:

Support for `Windows.UI.StartScreen.JumpList` APIs.

## Why is this needed:

Apps could then add quick actions to the app context menu on all platforms. Would be useful in many cases including Uno Calc :-) .

## For which Platform:

- [x] iOS
- [x] Android
- [ ] WebAssembly
- [x] Windows

MartinZikmund:
I have already started working on this

ghuntley:
Thank-you!

MartinZikmund:
It's in!

PR #1306, opened 2019-07-26
Windows.UI.StartScreen.JumpList

MartinZikmund:
GitHub Issue (If applicable): #1135 

## PR Type

- Feature

Feature not supported.

## What is the new behavior?

`Windows.UI.StartScreen.JumpList` partially supported on iOS and Android.

## PR Checklist


MartinZikmund:
@jeromelaban @davidjohnoliver @carldebilly This is still a draft, as I need feedback on the Android implementation:

1.  How to properly implement the `Drawables` handling. My current "workaround" is that I have added `DrawablesHelper` to Uno.UWP project, but if the developer wants to use JumpList on Android, `DrawablesHelper` must be initialized. It would be cool if this could be done automagically as when XAML is generated, but this is not UI, so I am not sure. Also - Is it necessary to maintain multiple implementations of this drawables locator (I have seen it is in two places currently in Uno.UI project, so this would be a third), cannot there be a single drawables source which is shared among all classes which need it? Or are there any technical limitations?

2. I don't much like the way I had to handle Android launch in `NativeApplication.OnActivityStarted`. Do you have a suggestion how this could be improved? I need to check the intent to know if it comes from a jumplist item, so that I know how `OnLaunched` should be called.

src/SamplesApp/SamplesApp.Shared/App.xaml.cs
```diff
@@ -48,7 +49,7 @@ public App()
 		/// will be used such as when the application is launched to open a specific file.
 		/// </summary>
 		/// <param name="e">Details about the launch request and process.</param>
-		protected override void OnLaunched(LaunchActivatedEventArgs e)
+		protected override async void OnLaunched(LaunchActivatedEventArgs e)
```
carldebilly: You should launch an "async void" instead of converting the `OnLaunched` to async.

... anyway it looks like you have no `await` in your method.

src/SamplesApp/UITests.Shared/Helpers/BindableBase.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Runtime.CompilerServices;
+using System.Text;
+
+namespace UITests.Shared.Helpers
+{
+	[Windows.UI.Xaml.Data.Bindable]
+	public class BindableBase : INotifyPropertyChanged
```
carldebilly: You may use `ViewModelBase` for that ;-)

src/SamplesApp/UITests.Shared/UITests.Shared.projitems
```diff
@@ -4152,6 +4159,9 @@
     <Compile Update="$(MSBuildThisFileDirectory)Resources\StaticResource\StaticResource_Simple.xaml.cs">
       <DependentUpon>StaticResource_Simple.xaml</DependentUpon>
     </Compile>
+    <Compile Update="D:\Dev\Uno\src\SamplesApp\UITests.Shared\Windows_UI_StartScreen\JumpListTests.xaml.cs">
```
carldebilly: There's a bug in VisualStudio putting absolute paths here.  You need to manually put `$(MSBuildThisFileDirectory)` here.

src/SamplesApp/UITests.Shared/Windows_UI_StartScreen/JumpListTestsViewModel.cs
```diff
@@ -0,0 +1,116 @@
...
+			get => _newItem;
+			set
+			{
+				_newItem = value;
+				RaisePropertyChanged(nameof(NewItem));
+			}
+		}
+
+		public ICommand AddItemCommand =>
+			_addItemCommand ?? (_addItemCommand = new DelegateCommand(AddItemAsync));
```
carldebilly: There's a `base.GetOrCreateCommand()` you can use instead of DelegateCommand...
<https://github.com/unoplatform/uno/blob/master/src/SamplesApp/UITests.Shared/Helpers/ViewModelBase.cs#L55>

I think we should have an overload to support async methods.

src/SamplesApp/SamplesApp.Shared/App.xaml.cs
MartinZikmund: fixed :-)

src/Uno.UWP/ApplicationModel/Activation/LaunchActivatedEventArgs.cs
```diff
@@ -18,7 +18,6 @@ internal LaunchActivatedEventArgs(ActivationKind kind, string arguments)
 			Kind = kind;
 		}
 
-		[NotImplemented]
```
carldebilly: Still `[NotImplemented]` for iOS & Wasm...

src/SamplesApp/UITests.Shared/Helpers/BindableBase.cs
MartinZikmund: In this case, the class is more lightweight (there is more additional logic in `ViewModelBase`) and`ViewModelBase` requires passing `Dispatcher` in the constructor, so for these reasons it seemed useful to have an alternative. But can refactor to `ViewModelBase` if you prefer :-) .

src/SamplesApp/UITests.Shared/UITests.Shared.projitems
MartinZikmund: I see, thanks for noticing :-) !

src/SamplesApp/UITests.Shared/Windows_UI_StartScreen/JumpListTestsViewModel.cs
MartinZikmund: Cool! will do

src/SamplesApp/UITests.Shared/Helpers/BindableBase.cs
carldebilly: It's not required.

src/Uno.UWP/ApplicationModel/Activation/LaunchActivatedEventArgs.cs
carldebilly: ,... at least Wasm

src/SamplesApp/UITests.Shared/Helpers/BindableBase.cs
MartinZikmund: I will keep it for now then 😉

src/SamplesApp/UITests.Shared/Windows_UI_StartScreen/JumpListTestsViewModel.cs
MartinZikmund: Done 😊

src/Uno.UWP/ApplicationModel/Activation/LaunchActivatedEventArgs.cs
MartinZikmund: Fixed

davidjohnoliver:
Nice work once more Martin!

> 1. How to properly implement the `Drawables` handling. My current "workaround" is that I have added `DrawablesHelper` to Uno.UWP project, but if the developer wants to use JumpList on Android, `DrawablesHelper` must be initialized. It would be cool if this could be done automagically as when XAML is generated, but this is not UI, so I am not sure. Also - Is it necessary to maintain multiple implementations of this drawables locator (I have seen it is in two places currently in Uno.UI project, so this would be a third), cannot there be a single drawables source which is shared among all classes which need it? Or are there any technical limitations?

Where else is this already implemented, apart from ImageSource? You mean this class? https://github.com/unoplatform/uno/blob/master/src/Uno.UI/DrawableHelper.Android.cs

We should definitely keep the logic in one place as much as possible, and it should definitely work without additional user code. 

I expressed [concern earlier](https://gitter.im/uno-platform/Lobby?at=5d3b1813223775758cfab354) about a binary breaking change if it were moved from `ImageSource`, but now I'm thinking it might not be a big deal, since `App.xaml` generally won't be defined in 3rd-party libraries. ((I don't think there's a problem with initializing it in the code generated from `App.xaml`, even if it's 'not UI'.)

If it still were a concern, there are multiple other workable options to maximise code reuse, eg defining a `GetResourceId` delegate in `Uno.UWP/DrawableHelper` and setting it either at runtime or compile time.

davidjohnoliver:
> 2. I don't much like the way I had to handle Android launch in `NativeApplication.OnActivityStarted`. Do you have a suggestion how this could be improved? I need to check the intent to know if it comes from a jumplist item, so that I know how `OnLaunched` should be called.

What don't you like about it?

src/Uno.UI/UI/Xaml/NativeApplication.cs
```diff
@@ -27,7 +30,16 @@ public override void OnCreate()
 
 		private void OnActivityStarted(Activity activity)
 		{
-			_app.OnLaunched(new LaunchActivatedEventArgs());
+			if (_lastHandledIntent != activity.Intent &&
```
davidjohnoliver: Why do you need to check here that the intent has changed?

MartinZikmund: As Activity Started is called every time when app goes to foreground (for example home button and then reopen via task view), the OnLaunched with the arguments would be used each time. But the jump list item arguments are relevant only the first time

MartinZikmund:
@davidjohnoliver 2. - I didn't like the fact that the logic of "handling" of the intent (with the check if it has changed) had to be part of the `NativeApplication` class, but as I think about it, it may be appropriate as it is handling the app launch, so it probably makes sense

MartinZikmund:
@davidjohnoliverI 1. - found it also here - src/Uno.UI/Controls/BindableImageSource.Android.cs, seems like a separate implementation. I kind of like the solution of moving this into a single DrawableHelper in Uno.UWP, which would be initialized by the App.xaml codegen. Can I try and do that in this PR or a separate?

MartinZikmund:
@davidjohnoliver Rather done it separately, so that it is clear it is a separate issue, please see #1308

MartinZikmund:
Alright! Merged, refactored, ready to for review 👍

doc/articles/features/windows-ui-startscreen.md
```diff
@@ -0,0 +1,13 @@
+# Uno Support for Windows.UI.StartScreen
```
davidjohnoliver: Could you add this to toc.yml https://github.com/unoplatform/uno/blob/master/doc/articles/toc.yml#L65 - that way it shows up in the [generated docs](https://platform.uno/docs/articles/intro.html) sidebar.

doc/articles/features/windows-ui-startscreen.md
```diff
@@ -0,0 +1,13 @@
...
+
+The APIs are implemented on iOS and Android. Other platforms return `false` when calling the `JumpList.IsSupported()` method.
+
+`JumpListItem` supports `DisplayName`, `Description`, `Arguments` and `Logo` properties on both Android and iOS.
+
+`Logo` property can be initialized only with `ms-appx:`-based images. This behavior matches UWP and actually matches native support on both platforms as well.
+
+To handle `JumpListItem` activation, check the `LaunchActivatedEventArgs.Arguments` in `App.OnLaunched` method. Note, that when the application is running, the method will still be called again (this behavior matches UWP).
+
+The API supports interoperability with native "app shortcuts". This means items you add using native API instead of `JumpList` API will not be overwritten by `JumpList` and will not be shown in `JumpList.Items`. To identify Uno-specific app shortcuts, a `UnoShortcut` key is used and set in `ShortcutInfo.Extras` on Android and `UIApplicationShortcutItem.UserInfo` on iOS.
```
davidjohnoliver: Would be great to have a couple of screenshots here :)

PR #1308, opened 2019-07-27
Refactored Drawable helper to Uno project

MartinZikmund:
GitHub Issue (If applicable): #1306

## PR Type

Drawable helper is available through `ImageSource.Android` in Uno.UI.

## What is the new behavior?

A dedicated `DrawableHelper` was added to `Uno` project so that drawables can be retrieved from there as well (required for `JumpListItem` icons in #1306)

## PR Checklist


davidjohnoliver:
`BindableImageSource` looks like straight-up dead code - internal class, not referenced anywhere else, go ahead and delete it entirely. :)

MartinZikmund:
@davidjohnoliver done 👍

MartinZikmund:
Fixed the problems, now should build

ghuntley:
Like magic, it works!

MartinZikmund:
@ghuntley Seems the build doesn't start 🤔

jeromelaban:
It will now, I missed an option. Can you retry ?

jeromelaban:
Not sure `/azp run` will work, but try pushing an update, or a rebase.

MartinZikmund:
Oh there it is 😃 ! Thank you both!

src/Uno.UI/UI/Xaml/Media/ImageSource.Android.cs
```diff
@@ -321,63 +323,6 @@ private async Task<Bitmap> FetchResourceWithDownsampling(CancellationToken ct, i
...
-					p => (int)p.GetValue(null)
-				);
-		}
-
-		/// <summary>
-		/// Returns the Id of the bundled image.
-		/// </summary>
-		/// <param name="imageName">Name of the image</param>
-		/// <returns>Resource's id</returns>
-		public static int? FindResourceId(string imageName)
```
jeromelaban: Those methods and properties cannot be removed, as the methods are used by existing users of Uno. The body of the methods can redirect to the new location, though. You can mark those as editor hidden, though.

MartinZikmund: @jeromelaban  Should be fixed now - returned both these methods and the previous `DrawableHelper`. One more question though - should the `DrawableHelper` stay directly in the `Uno` namespace or would `Uno.Android` or `Uno.Helpers` be more appropriate?

jeromelaban: Let's keep them in the `Uno.UI` namespace, Uno is too broad for this assembly. Using `Android` as a namespace will cause conflicts :)

MartinZikmund: The problem is there is already the Uno.UI.DrawableHelper in the Uno.UI library (we now have `src/Uno.UWP/DrawableHelper.Android.cs` and `src/Uno.UI/DrawableHelper.Android.cs`), so this should probably be something else or it would clash

MartinZikmund: Alternative would also be to rename the new one to something else than `DrawableHelper`

MartinZikmund: Alternative, could also be to remove DrawableHelper from Uno.UI completely, but I guess that would be the same problem as with the ImageSource methods, as it was public and some users may be using.

MartinZikmund: @jeromelaban Just pinging, do you have some suggestion what to do with this problem, please? I would like to close this one soon so I can use it in #1306

MartinZikmund: @davidjohnoliver I learned Jerome is currently out of office, any chance you could help resolve this namespace problem?

davidjohnoliver: Sure thing - the main concern in this case is not introducing a binary breaking change, beyond that the namespace isn't such a big deal, since most consumers won't be using this code directly. 

My call on this:
 - leave `Uno.UI.DrawableHelper` and Uno.DrawableHelper` as you currently have, with `Uno.UI.DrawableHelper` internally redirecting to `Uno.DrawableHelper`
 - mark `Uno.DrawableHelper` and its methods as `EditorBrowsableState.Never`, to minimise confusion

I think that should do it.

MartinZikmund: So I would keep `Uno.DrawableHelper` public, in this general namespace? I wouldn't want to clutter it though, as it is currently empty otherwise I think. Key thing is it has to be public and ideally it should be the "go to helper for Drawables", so I would think it could be better to make the existing UI.DrawableHelper editor hidden instead

davidjohnoliver: >  I would think it could be better to make the existing UI.DrawableHelper editor hidden instead (to cover the unlikely scenario when some library references Uno dll but not Uno.UI dll)

Fair enough, let's hide the `Uno.UI` one then.

For the class in the Uno dll, let's call it `Uno.Helpers.DrawableHelper`.

MartinZikmund: Alright, sounds good :-)

MartinZikmund: @davidjohnoliver pushed the requested changes

src/Uno.UWP/Helpers/DrawableHelper.Android.cs
```diff
@@ -0,0 +1,88 @@
...
+		/// Returns the Id of the bundled image.
+		/// </summary>
+		/// <param name="imageName">Name of the image</param>
+		/// <returns>Resource's id</returns>
+		public static int? FindResourceId(string imageName)
+		{
+			var key = System.IO.Path.GetFileNameWithoutExtension(imageName);
+			if (_drawablesLookup == null)
+			{
+				throw new Exception("You must initialize drawable resources by invoking this in your main Module (replace \"GenericApp\"):\nWindows.UI.Xaml.Media.ImageSource.Drawables = typeof(GenericApp.Resource.Drawable);");
```
davidjohnoliver:
```suggestion
				throw new Exception("Drawable resources were not initialized.");
```

That error is way out of date 😆

MartinZikmund: True that :-D

MartinZikmund: Can the PR be merged even with the open change request from Jerome? It would be great, I could then finish the JumpList PR 👍 ✔

davidjohnoliver: Done :)

Issue #1350, opened 2019-08-08
Vertical Slider gets set to inverted value when tapped


pdecostervgls:
## Current behavior

When a vertical slider gets tapped at the 20% point, the thumb moves towards the 80% point

## Expected behavior

The vertical slider value reflects the position it was tapped at.

## How to reproduce it (as minimally and precisely as possible)

1. Goto Uno samples > Slider > Slider_Simple
2. Tap the vertical slider at around the 20% position from the bottom
3. Observe the value being set to the 80% point from the bottom

## Environment

Uno.UI Samples App

Affected platform(s):
- [x] iOS
- [x] Android
- [ ] WebAssembly
- [ ] Windows
- [ ] Build tasks

Visual Studio
- [ ] 2017 (version: )
- [x] 2019 (version: 16.1.6)
- [ ] for Mac (version: )

ghuntley:
❤️ @pdecostervgls, thanks for jumping in and opening up a pull-request.

carldebilly:
@pdecostervgls @ghuntley @davidjohnoliver Is this issue fixed?  The PR is merged.

pdecostervgls:
@carldebilly Yes, I believe it is fixed with the merging of that PR.

carldebilly:
Perfect.  I'll close this issue.  Reopen it (or open a new one) if you see a regression.

PR #1351, opened 2019-08-08
Fix inverted vertical slider

pdecostervgls:

#1350

## PR Type

- Bugfix

The value the vertical slider get set to when tapped is inverted.
e.g.: 60% -> 40%, 80% -> 20%

## What is the new behavior?
The vertical slider is set to the value that reflects the position its been tapped at.

## PR Checklist

(https://github.com/nventive/Uno/blob/master/doc/.feature-template.md). (for bug fixes / features)

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/Slider/Slider.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Slider/Slider.cs
@@ -352,7 +352,7 @@ namespace Windows.UI.Xaml.Controls
 
 			var newOffset = Orientation == Orientation.Horizontal ?
 				point.X / container.ActualWidth :
-				point.Y / container.ActualHeight;
+				1 - (point.Y / container.ActualHeight);
 
 			ApplySlideToValue(newOffset);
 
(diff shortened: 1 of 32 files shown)


src/Uno.UI/UI/Xaml/Controls/Slider/Slider.cs
```diff
@@ -352,7 +352,7 @@ private void OnSliderContainerPressed(object sender, PointerRoutedEventArgs e)
 
 			var newOffset = Orientation == Orientation.Horizontal ?
 				point.X / container.ActualWidth :
-				point.Y / container.ActualHeight;
+				1 - (point.Y / container.ActualHeight);
```
pdecostervgls: So the easiest way to fix this, is to subtract the result from 1, inverting the inverted value. However, one could argue that the actual bug resides in the `ApplySlideToValue` method.

davidjohnoliver: Looks fine to me this way, I would prefer to keep the `ApplySlideToValue` logic the same for both horizontal and vertical orientations.

pdecostervgls:
@davidjohnoliver Seems that the UI test failed on wasm. I haven't been able to test using wasm yet.

davidjohnoliver:
Thanks for adding the test :)

Issue #1586, opened 2019-09-23
Implement Gyrometer


MartinZikmund:
##  What would you like to be added:

`Windows.Devices.Sensors.Gyrometer`

## Why is this needed:

🔁

## For which Platform:

- [x] iOS
- [x] Android
- [x] WebAssembly
- [ ] Windows

MartinZikmund:
Done :-)

PR #1587, opened 2019-09-23
Windows.Devices.Sensors.Gyrometer + WASM sensor fixes

MartinZikmund:
GitHub Issue (If applicable): #1586

## PR Type

- Feature

No `Gyrometer` support

## What is the new behavior?

`Gyrometer` supported on Android, iOS and WASM

## PR Checklist

- [x] [Wasm UI Tests](doc/articles/working-with-the-samples-apps.md#running-the-webassembly-ui-tests-snapshots) are not showing unexpected any differences. Validate PR `Screenshots Compare Test Run` results.

##  Breaking change

I found out the `MagnetometerReading` constructor was accidentally left `public`, while it in fact does not exist in UWP. I have now marked it as `internal`. This should not affect any users, because constructing `MagnetometerReading` directly is not a supported scenario in UWP.

## Other information

The PR also contains some WASM fixes for `Magnetometer` as it had a typo which prevented it from actually working.


MartinZikmund:
@jeromelaban Do you have a tip why the PackageDiffIgnore change is not effective and the build still fails?

jeromelaban:
@MartinZikmund the diff is made against the latest stable published package, so your changes should still be in the existing block.

MartinZikmund:
@jeromelaban oh I see, will try

doc/articles/features/windows-devices-sensors.md
```diff
@@ -32,10 +32,16 @@ On Android, when both `ReadingChanged` and `Shaken` events are attached and the
 
 * No barometer API is currently available in JS
 
+## `Gyrometer`
+
+### Implementation notes
+
+`ReportInterval` property on WASM is currently not supported directly and we use an approximation in the form of raising the `ReadingChanged` event only when enough time has passed since the last report. The event is actually raised a bit more often to make sure the gap caused by the filter is not too large, but this is in-line with the behavior of UWP `Gyrometer`.
```
jeromelaban:
```suggestion
`ReportInterval` property on WASM is currently not supported directly. Uno uses an approximation in the form of raising the `ReadingChanged` event, only when enough time has passed since the last report. The event is raised a bit more often to make sure the gap caused by the filter is not too large, but this is in-line with the behavior of Windows' `Gyrometer`.
```

doc/articles/features/windows-devices-sensors.md
```diff
@@ -32,10 +32,16 @@ On Android, when both `ReadingChanged` and `Shaken` events are attached and the
...
+
 ## `Magnetometer`
 
 ### Implementation notes
 
-`ReportInterval` property on WASM is currently not supported directly and we use an approximation in the form of raising the `ReadingChanged` event only when enough time has passed since the last report. The event is actually raised a bit more often to make sure the gap caused by the filter is not too large, but this is in-line with the behavior of UWP Magnetometer.
+`ReportInterval` property on WASM is currently not supported directly and we use an approximation in the form of raising the `ReadingChanged` event only when enough time has passed since the last report. The event is actually raised a bit more often to make sure the gap caused by the filter is not too large, but this is in-line with the behavior of UWP `Magnetometer`.
```
jeromelaban:
```suggestion
`ReportInterval` property on WASM is currently not supported directly. Uno uses an approximation in the form of raising the `ReadingChanged` event only when enough time has passed since the last report. The event is raised a bit more often to make sure the gap caused by the filter is not too large, but this is in-line with the behavior of UWP `Magnetometer`.
```

src/Uno.UWP/Devices/Sensors/Gyrometer.Android.cs
```diff
@@ -0,0 +1,97 @@
...
+			if (sensor != null)
+			{
+				return new Gyrometer(sensor);
+			}
+			return null;
+		}
+
+		private void StartReading()
+		{
+			_listener = new GyrometerListener(this);
```
jeromelaban: Can you add debug logging on all those methods? this will help diagnosing issues.

MartinZikmund: Will do

MartinZikmund: Added the logging to the common Gyrometer.cs file - OnReadingChanged and in the event add/remove

MartinZikmund:
> Very nice! Could you try and see if a UI Test / Runtime Test could work ?

Done :-)

jeromelaban:
Thanks! Does the UI Test fail on wasm ? Or maybe it asks for some permission ?

MartinZikmund:
> Thanks! Does the UI Test fail on wasm ? Or maybe it asks for some permission ?

The sensor must be present on the device, so it would need to run on a mobile device, otherwise, it just shows Gyrometer is not available and the buttons are disabled. So it would probably fail the first status check

jeromelaban:
Agreed. It looks like there's a way to enable mobile emulation, but it's not mentioned if gyroscope is enabled in that context.

PR #1588, opened 2019-09-23
feat: Check launch URL for fragments if no querystring exists

Redth:
Querystrings were already being checked for and passed along to the launch event args, however in some cases we might want to use fragments instead. The best use case I can think of is that they don't get sent to from the client in web requests, and they don't participate in redirects.  Anyway, I think it's a simple case that would be useful to include.

GitHub Issue (If applicable): # N/A

## PR Type

- Bugfix

URL Fragments are ignored if there is no querystring (`?`) before the fragment start (`#`) and don't get passed to the launch event args.

## What is the new behavior?

If there's no querystring (`?`) in the url, the url is checked for a fragment (`#`) and captures it to pass to the launch event args.

## PR Checklist

### DIFF
--- a/src/Uno.UI/ts/WindowManager.ts
+++ b/src/Uno.UI/ts/WindowManager.ts
@@ -204,18 +204,34 @@ namespace Uno.UI {
 			* 
 			*/
 		static findLaunchArguments(): string {
-			if (typeof URLSearchParams === "function") {
-				return new URLSearchParams(window.location.search).toString();
-			}
-			else {
-				const queryIndex = document.location.search.indexOf('?');
+			/*
+			* Concatenate the query/search with any url fragments into one string
+			*/
+			const searchAndHash = window.location.search + window.location.hash;
+			
+			/*
+			* Look for the presence of ? for the querystring, if found return
+			* everything after it, which will include a fragment if also present
+			* since fragments always proceed querystring params if they exist
+			*/
+			const queryIndex = document.location.search.indexOf('?');
 
-				if (queryIndex !== -1) {
-					return document.location.search.substring(queryIndex + 1);
+			if (queryIndex !== -1) {
+				return document.location.search.substring(queryIndex + 1);
+			} else {
+				/*
+				* Fall back to trying a url fragment if no querystring args were found
+				* since there may still be a url fragment we want to include
+				*/
+				const fragmentIndex = document.location.search.indexOf('#');
+					
+				if (fragmentIndex !== -1) {
+					return document.location.search.substring(fragmentIndex + 1);
 				}
-
-				return "";
 			}
+
+			// Return probably an empty string
+			return searchAndHash;
 		}
 
 		/**


carldebilly (commented):
What will happend for browsers supporting `URLSearchParams`?

XiaotianNetlift:
I don't think you will find the fragment of the url inside `document.location.search`.
> [`Location.search`](https://developer.mozilla.org/en-US/docs/Web/API/Location/search)
Is a DOMString containing a '?' followed by the parameters or "querystring" of the URL. Modern browsers provide URLSearchParams and URL.searchParams to make it easy to parse out the parameters from the querystring.
> [`Location.hash`](https://developer.mozilla.org/en-US/docs/Web/API/Location/hash)
> Is a DOMString containing a '#' followed by the fragment identifier of the URL.
> -- https://developer.mozilla.org/en-US/docs/Web/API/Location

We could probably just concatenate the result with `document.location.hash`, but this is a breaking change that should be noted.

Redth:
Ok, admittedly I'm not a web person, but I've read up a bit now on `URLSearchParams` and `location.search`.  

If we just concatenate `location.search` and `location.hash`, if we feed that into `URLSearchParams` it escapes the `#` with a `%23`, which in my opinion is not helpful.

I propose we remove any usage of `URLSearchParams` and instead use `loctxt = location.search + location.hash`, and look for `?` (falling back to `#` since that should always proceed the `?`) to find the querystring and/or hash of the url, and pass it to the launch event args that way.

Thoughts?

XiaotianNetlift:
You can concatenate after the `toString`:
```
new URLSearchParams(document.location.search).toString() + document.location.hash
```

But, I agree the `URLSearchParams` is unnecessary here.

Redth:
Made the changes I proposed here.  The one thing I'm not sure I'm happy with is that the `#` won't show up in the args if there wasn't a `?` before it.  Maybe it should always return `#` even if that means the args are `#somefragment` since the `#` will always be in the args in the presence of `?` (eg: `?query#fragment` will show up as args of `query#fragment`).  I think _my_ preference would be to always have `?` in the args as well, but that would be even more of a breaking change.

MartinZikmund:
@Redth are you able to merge the latest master and sign the CLA please, so we can finalize review and merge this?

MartinZikmund:
@Redth I see it will be needed to build Uno.UI.Wasm and to commit the updated `Uno.UI.js` which will be generated (this happens during the build).

MartinZikmund:
@Redth Also please reword your commit message to be compatible with conventional commits (for example "feat: Check launch URL for fragments if no querystring exists"). If you don't know how to do that, let me know :-) .

src/Uno.UI/ts/WindowManager.ts
```diff
@@ -204,18 +204,34 @@ namespace Uno.UI {
...
+				return document.location.search.substring(queryIndex + 1);
+			} else {
+				/*
+				* Fall back to trying a url fragment if no querystring args were found
+				* since there may still be a url fragment we want to include
+				*/
+				const fragmentIndex = document.location.search.indexOf('#');
+					
+				if (fragmentIndex !== -1) {
+					return document.location.search.substring(fragmentIndex + 1);
```
MartinZikmund, line 229: I would suggest keeping the `#` character in the result, as it would make it more aligned with the general case of querystring + fragment, eg.:

- http://example.com?something#fragment => `something#fragment`
- http://example.com#fragment => `#fragment`

This would make it possible to distinguish "fragment" from plain query string.

MartinZikmund (changes requested):

src/Uno.UI/ts/WindowManager.ts
MartinZikmund: @Redth do you have time to make this change in near future please? Thanks 👍

MartinZikmund: @Redth Pinging regarding the PR again, is it still something you plan to work on, or should I take over and get it ready to merge?

Issue #1323, opened 2019-07-30
Support Dark and HighContrast in SystemResources.xaml


MartinZikmund:
## What would you like to be added:

Uno currently has support for the light theme resources only. Now that `ThemeResource` support exists, Dark and HighContrast could be added, too.

## Why is this needed:

To quote Lego Batman:

> I only work in black. And sometimes, very very dark grey.

Also, `HighContrast` is very important for accessibility

## For which Platform:

- [x] iOS
- [x] Android
- [x] WebAssembly
- [ ] Windows

MartinZikmund:
@davidjohnoliver Is the `SystemResources.xaml` autogenerated somehow or is it enough to bring the same set of `x:Key` resources for the new themes in manually?

davidjohnoliver:
@MartinZikmund I don't think it's autogenerated, it's fine to copy missing values manually. 

tbh I don't know where dark resources are defined by UWP. I see `<ResourceDictionary x:Key="Light">` in the system `generic.xaml` but don't see the precise equivalent for dark/high contrast. Do you know?

MartinZikmund:
@davidjohnoliver interestingly enough dark resources are in the "Default" section, so there is only Default, Light and HighContrast.

MartinZikmund:
Tried this, but I am not sure which resources are Uno specific and which can be reused from Windows, so leaving this for now, if someone else may want to take this.

MartinZikmund:
I was looking into this a bit more. Copy-pasting matching resources from UWP will produce mostly working app, except for a two issues:

- `TextBlock` foreground stays black even for `RequestedTheme=Dark`  (the default `Foreground` is hardcoded in the dependency property) - not sure how to fix this
- There are some missing resources even in the light theme, but also many resources which are present in different files outside of `SystemResources.xaml`, is there a reason to have them in separate XAML resource dictionaries, @jeromelaban ? I understand it is useful to have the control templates/styles separate, but colors and brushes could all be in `SystemResources.xaml` and in such case we could practically take the whole UWP `ThemeDictionaries` element and paste it as-is to `SystemResources.xaml`

jeromelaban:
[...]

PR #1628, opened 2019-09-29
Dark & HighContrast theme, system theme detection

MartinZikmund:
GitHub Issue (If applicable): #1323 

## PR Type

- Feature

Uno comes with `Light` theme resources only

## What is the new behavior?

`Dark` and `HighContrast`  theme resources included in Uno. I copied the resources side-by-side by comparing UWP and Light theme resources which were already in Uno, so there should hopefully be no differences.

PR also contains system theme detection on Android, iOS and macOS, so if user does not specify `RequestedTheme`, the system theme at the time of app startup will be used.

## PR Checklist

## Breaking change

This potentially is a "breaking change", as previously, when user removed `RequestedTheme="Light"` present in `App.xaml` from default project template, the app stayed in `Light` theme. Now such a change will display the app in `Dark` theme. This should however not affect too many users and is in line with UWP.

**No longer the case as per discussion with @davidjohnoliver - app default theme is Light on all targets except UWP to match the system default theme which is Light**


MartinZikmund:
@jeromelaban I am getting packagediff errors due to the fact that `HighContrast` theme uses `StaticResource` elements to reference existing brushes, instead of declaring `SolidColorBrush` directly as in the case of `Light` and `Dark` theme. This is however, not "actually" a breaking change, as the `StaticResource` still points to a `SolidColorBrush`. How to best resolve this? Should I replace those `StaticResource` declarations one by one?

MartinZikmund:
Threw some regexes on it and turned the `StaticResource` elements into proper `SolidColorBrushes` again. However, if we want full copy-paste compatibility with UWP sometime later, this will have to be addressed and the differences will probably need to go to the package diff XML.

jeromelaban:
Indeed. Thanks for trying out! Yes, we'll certainly have to straighten this up with style updates.

MartinZikmund:
@jeromelaban  not sure how to solve this error in VSTest though:

image

src/SolutionTemplate/UnoSolutionTemplate/Shared/App.xaml
```diff
@@ -2,7 +2,6 @@
     x:Class="$ext_safeprojectname$.App"
     xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-    xmlns:local="using:$ext_safeprojectname$"
-    RequestedTheme="Light">
+    xmlns:local="using:$ext_safeprojectname$">
```
davidjohnoliver: 👍

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.cs
```diff
@@ -387,12 +387,35 @@ private void BuildApplicationInitializerBody(IndentedStringBuilder writer, XamlO
...
+				blockWriter.AppendLineInvariant(
+					"// Source {0} (Line {1}:{2})",
+					_fileDefinition.FilePath,
+					topLevelControl.LineNumber,
+					topLevelControl.LinePosition
+				);
+				BuildLiteralProperties(blockWriter, topLevelControl, closure);
+			}
+
+			if (IsFrameworkElement(topLevelControl.Type))
```
davidjohnoliver: this doesn't really make sense here - `App.xaml` is never going to be a `FrameworkElement`.

src/Uno.UI/UI/Xaml/Application.Android.cs
```diff
@@ -32,6 +34,19 @@ public Application()
...
+		{		
+			if ((int)Build.VERSION.SdkInt >= 28)
+			{
+				var uiModeFlags = Android.App.Application.Context.Resources.Configuration.UiMode & UiMode.NightMask;
+				if (uiModeFlags == UiMode.NightNo)
+				{
+					return ApplicationTheme.Light;
+				}				
+			}
+			return ApplicationTheme.Dark;
```
davidjohnoliver: The default should be `Light`, if the OS level doesn't support dark mode.

src/Uno.UI/UI/Xaml/Application.iOS.cs
```diff
@@ -83,6 +83,19 @@ public override UIInterfaceOrientationMask GetSupportedInterfaceOrientations(UIA
...
+		{
+			//Ensure the current device is running 12.0 or higher, because `TraitCollection.UserInterfaceStyle` was introduced in iOS 12.0
+			if (UIDevice.CurrentDevice.CheckSystemVersion(12, 0))
+			{
+				if (UIScreen.MainScreen.TraitCollection.UserInterfaceStyle == UIUserInterfaceStyle.Light)
+				{
+					return ApplicationTheme.Light;
+				}
+			}
+			return ApplicationTheme.Dark;
```
davidjohnoliver: ditto here, it should default to light

src/Uno.UI/UI/Xaml/Application.net.cs
```diff
@@ -10,5 +10,7 @@ public Application()
 		{
 			Current = this;
 		}
+
+		private ApplicationTheme GetDefaultSystemTheme() => ApplicationTheme.Dark;
```
davidjohnoliver:
```suggestion
		private ApplicationTheme GetDefaultSystemTheme() => ApplicationTheme.Light;
```

src/Uno.UI/UI/Xaml/Application.wasm.cs
```diff
@@ -68,6 +68,8 @@ private void Initialize()
 				OnLaunched(new LaunchActivatedEventArgs(ActivationKind.Launch, arguments));
 			}
 		}
+
+		private ApplicationTheme GetDefaultSystemTheme() => ApplicationTheme.Dark;
```
davidjohnoliver:
```suggestion
		private ApplicationTheme GetDefaultSystemTheme() => ApplicationTheme.Light;
```

src/Uno.UI/UI/Xaml/Application.Android.cs
MartinZikmund: Even though the UWP default is `Dark`?

MartinZikmund: The reason why I chose `Dark` as the default is that UWP has dark theme resources in the `Default` theme dictionary, not light, but probably makes sense to match the target OS instead

MartinZikmund: Done :-)

MartinZikmund:
> Nice work Martin!
> 
> My branch will bring the ability to reapply `ThemeResource` assignations if the theme changes at runtime.
> 
> Also it'll be a major breaking change anyway, so that'll be a good occasion to use the correct `<StaticResource` syntax.

What is the branch you are working on? I am curious to see if it will be possible to make the app react to OS theme change at runtime automatically :-)

src/Uno.UI/UI/Xaml/Application.cs
```diff
@@ -29,7 +30,11 @@ public static class TraceProvider
 
 		public DebugSettings DebugSettings { get; } = new DebugSettings();
 
-		public ApplicationTheme RequestedTheme { get; set; }
+		public ApplicationTheme RequestedTheme
+		{
+			get => _requestedTheme ?? (_requestedTheme = GetDefaultSystemTheme()).Value;
+			set => _requestedTheme = value;
```
carldebilly: The _getter_ is incorrect here: it should be initialized during the initialization of the class.

Also the setter should fail - as UWP - if set after the application has been activated.

src/Uno.UI/UI/Xaml/Application.wasm.cs
```diff
@@ -68,6 +68,8 @@ private void Initialize()
 				OnLaunched(new LaunchActivatedEventArgs(ActivationKind.Launch, arguments));
 			}
 		}
+
+		private ApplicationTheme GetDefaultSystemTheme() => ApplicationTheme.Light;
```
carldebilly: Maybe we can use that?
<https://drafts.csswg.org/mediaqueries-5/#prefers-color-scheme>

src/Uno.UI/UI/Xaml/Application.cs
MartinZikmund: @carldebilly Actually the getter was the only way to get it work - on Android when I tried to initialize it in the constructor, it crashed with stack overflow, as resolving the application theme caused the android Application context to initialize (which it was not yet, as its initialization requires the `App` constructor to finish) and it got into a infinite loop. Solution would be to pass in a `App` factory to the `base` constructor of `Windows.UI.Xaml.NativeApplication` in Android, so that the Application Context could initialize before creating the Uno App but that would be a breaking change to the template's `Main.cs` file...

src/Uno.UI/UI/Xaml/Application.wasm.cs
MartinZikmund: Good find! I will look into it

src/Uno.UI/UI/Xaml/Application.cs
MartinZikmund: I also guess this change will need to be done sometime, but as this workaround was equivalent, I went with it

MartinZikmund: I will fix the Setter though!

src/Uno.UI/UI/Xaml/Application.wasm.cs
carldebilly: https://caniuse.com/#feat=prefers-color-scheme

MartinZikmund: Done :-) !!!

src/Uno.UI/UI/Xaml/Application.cs
MartinZikmund: Done :-)

src/Uno.UI/UI/Xaml/Application.wasm.cs
MartinZikmund: I had to add a boolean flage _initializationCompleted to check if the setter should throw, as behavior in UWP allows to set the ApplicationTheme any time during the constructor (even at the end), so the first place it should throw is during `OnLaunched`

src/Uno.UI/UI/Xaml/Application.cs
MartinZikmund: I had to add a boolean flag `_initializationCompleted`, as UWP allows setting `ApplicationTheme` anytime during the `App` constructor (even after Application.Current is set), but it should throw then during `OnLaunched`

src/Uno.UI.Tests/Windows_UI_Xaml/Given_ThemeResource.cs
```diff
@@ -14,14 +14,27 @@ namespace Uno.UI.Tests.Windows_UI_Xaml
...
+		public void When_System_ThemeResource_Dark()
+		{
+			var app = UnitTestsApp.App.EnsureApplication();
+			app.ForceSetRequestedTheme(Windows.UI.Xaml.ApplicationTheme.Dark);
+
+			var control = new Test_Control();
+			app.HostView.Children.Add(control);
+
+			Assert.AreEqual(Color.FromArgb(0xDE, 0xFF, 0xFF, 0xFF), (control.TestTextBlock.Foreground as SolidColorBrush).Color);
+		}
```
davidjohnoliver: Set the theme back to light at the end of this test, in case other tests are assuming that

MartinZikmund: Oh right! Similar issue happened in other test, I could not wrap my head around why it sometimes randomly crashes until I realized this :-D .

MartinZikmund: Done 👌

src/Uno.UI.Wasm/ts/Windows/UI/Xaml/Application.ts
```diff
@@ -0,0 +1,15 @@
...
+    export class Application {
+        public static getDefaultSystemTheme() : string {
+            if (window.matchMedia("(prefers-color-scheme: dark)").matches) {
+                return ApplicationTheme.Dark;
+            }
+            if (window.matchMedia("(prefers-color-scheme: light)").matches) {
+                return ApplicationTheme.Light;
+            }
+			//if none matches or not supported
+            return ApplicationTheme.Light;
```
carldebilly: I think it should return null here and make that decision on managed code instead.

MartinZikmund: Done :-)

doc/ReleaseNotes/_ReleaseNotes.md
```diff
@@ -2,6 +2,7 @@
 
 ## Next version
 ### Features
+* Added Dark and HighContrast theme resources, reacts to Dark/Light theme on iOS, Android and WASM automatically during startup of the app if `RequestedTheme` is not set in `App.xaml`
```
```suggestion
* Added Dark and HighContrast theme resources, reacts to Dark/Light theme on iOS, Android and WASM automatically during the startup of the app if `RequestedTheme` is not set in `App.xaml`
```

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.cs
```diff
@@ -387,12 +387,30 @@ private void BuildApplicationInitializerBody(IndentedStringBuilder writer, XamlO
...
 
 			RegisterResources(topLevelControl);
 			BuildProperties(writer, topLevelControl, isInline: false, returnsContent: false);
+
+			writer.AppendLineInvariant("");
+			writer.AppendLineInvariant("this");
+
+			string closure;
+
+			using (var blockWriter = CreateApplyBlock(writer, null, out closure))
```
jeromelaban:
```suggestion
			using (var blockWriter = CreateApplyBlock(writer, null, out var closure))
```

src/Uno.UI.Tests/Windows_UI_Xaml/Given_ThemeResource.cs
```diff
@@ -14,14 +14,35 @@ namespace Uno.UI.Tests.Windows_UI_Xaml
 	public class Given_ThemeResource
 	{
 		[TestMethod]
-		public void When_System_ThemeResource()
+		public void When_System_ThemeResource_Light()
 		{
 			var app = UnitTestsApp.App.EnsureApplication();
+			app.ForceSetRequestedTheme(Windows.UI.Xaml.ApplicationTheme.Light);
```
jeromelaban: Should there be a reset of the theme as well here ? (even if it's the default)

src/Uno.UI.Wasm/ts/Windows/UI/Xaml/ApplicationTheme.ts
```diff
@@ -0,0 +1,7 @@
+﻿namespace Windows.UI.Xaml {
+
+    export enum ApplicationTheme {
+        Light = "Light",
+		Dark = "Dark"
```
jeromelaban:
```suggestion
	Dark = "Dark"
```

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.cs
MartinZikmund: Fixed :-)

src/Uno.UI.Tests/Windows_UI_Xaml/Given_ThemeResource.cs
MartinZikmund: Fixed :-)

MartinZikmund:
@carldebilly @jeromelaban @davidjohnoliver Can this be merged? So that the new SystemResources.xaml is in master to avoid significant conflicts/forgotten resource keys 😀

davidjohnoliver:
> @carldebilly @jeromelaban @davidjohnoliver Can this be merged? So that the new SystemResources.xaml is in master to avoid significant conflicts/forgotten resource keys 😀

Done :)

Btw this is the styles+resources branch: https://github.com/unoplatform/uno/pull/1766

It's not wired up to the native events, but it has the binding-engine hooks to update the theme dynamically (`Application.OnRequestedThemeChanged()`)

MartinZikmund:
> > @carldebilly @jeromelaban @davidjohnoliver Can this be merged? So that the new SystemResources.xaml is in master to avoid significant conflicts/forgotten resource keys 😀
> 
> Done :)
> 
> Btw this is the styles+resources branch: #1766
> 
> It's not wired up to the native events, but it has the binding-engine hooks to update the theme dynamically (`Application.OnRequestedThemeChanged()`)

Epic! I will check to see how to react to theme changes automatically and then we roll!

jeromelaban:
We should change the sample app so that it changes its style via the URL, so we could test it in that mode and have screenshots testing that way.

PR #1701, opened 2019-10-03
updated gitversion.yml

SanyuktaSaha:
added few features.


jeromelaban:
Thank for your contribution @SanyuktaSaha. At this point, we're not considering changing the gitversion configuration.

PR #1740, opened 2019-10-07
Fix TYPO

riya-17:
I don't think there must be full stop after "XAML" and "Today". If there is any specific reason please specify.


jeromelaban:
Thank you for your contribution, but this is a marketing thing :) It will stay that way.

Issue #1834, opened 2019-10-14
Inconsistent use of #if __WASM__


ghuntley:
## Observation:

Some WebAssembly implementations use `#if __WASM__` but the majority does not.

![image](https://user-images.githubusercontent.com/127353/66728449-c3408e80-ee90-11e9-8054-33898b763da1.png)

![image](https://user-images.githubusercontent.com/127353/66728529-3ba74f80-ee91-11e9-86b5-5dd7a9ad8ac0.png)

## For which Platform:

- [ ] iOS
- [ ] Android
- [x] WebAssembly
- [ ] WebAssembly renderers for Xamarin.Forms
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

## Anything else we need to know?

What is the rule of thumb for refactoring? If `#if __WASM__` is missing - should it be added?

jeromelaban:
It should not be used for whole files, as there is a globbing pattern in place to avoid having to place it everywhere. See here:
https://github.com/unoplatform/uno/blob/6cc46e6268ef60eb9a214719da16faa1dce3a297/src/PlatformItemGroups.props#L48

ghuntley:
> 
> 
> It should not be used for whole files, as there is a globbing pattern in place to avoid having to place it everywhere. See here:
> 
> https://github.com/unoplatform/uno/blob/6cc46e6268ef60eb9a214719da16faa1dce3a297/src/PlatformItemGroups.props#L48

Documented over at https://github.com/unoplatform/uno/issues/1834#issuecomment-542161088

PR #1866, opened 2019-10-16
docs: add more insights about Uno's code style/conventions

ghuntley:
## PR Type

## Other information

Added insights from https://github.com/unoplatform/uno/issues/1834#issuecomment-542161088

### DIFF
--- a/doc/articles/contributing/guidelines/code-style.md
+++ b/doc/articles/contributing/guidelines/code-style.md
@@ -30,6 +30,19 @@ Uno uses EditorConfig ([here's our configuration](https://github.com/unoplatform
 - They should not be used for whole files, as there is [a globbing pattern](https://github.com/unoplatform/uno/blob/master/src/PlatformItemGroups.props) in place to avoid having to place it everywhere. See here:
 
 
+### Braces
+
+```csharp
+if (condition)
+{
+    // do something
+}
+else
+{
+    // use braces even for single line conditions
+}
+```            
+
 ### Filenames
 
 As defined over at https://github.com/unoplatform/uno/blob/master/src/PlatformItemGroups.props
@@ -43,19 +56,6 @@ As defined over at https://github.com/unoplatform/uno/blob/master/src/PlatformIt
 - ClassName.UWP.cs
 - ClassName.net.cs
 - ClassName.netstd.cs
-
-### Braces
-
-```csharp
-if (condition)
-{
-    // do something
-}
-else
-{
-    // use braces even for single line conditions
-}
-```            
 
 ### Strings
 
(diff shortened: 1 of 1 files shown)


ghuntley:
@jeromelaban can you please ship some guidance into this branch (or reply back) about `SampleControlInfo` naming conventions? How should `SampleControlInfoAttribute` be used and what is a category?

There seems to be multiple conventions in use within the samples app:

image

# example

Assuming we are creating samples for the `Windows.UI.Xaml.Controls.ContentDialog` control which is right?

```
[SampleControlInfo("ContentDialog", "ContentDialog_Closing", description: "Tests for ContentDialog.Closing event")]
```

Or

```
[SampleControlInfo("Windows.UI.Xaml.Controls", "ContentDialog_Closing", description: "Tests for ContentDialog.Closing event")]
``

Or ... something else

# remarks

The signature for `SampleControlInfoAttribute` currently is as follows:

```csharp
public SampleControlInfoAttribute(string category, string controlName, Type viewModelType = null, bool ignoreInAutomatedTests = false, string description = null)
{
   ...
}
```

Changing the signature for `SampleControlInfoAttribute` to be this would be an improvement:


```csharp
public SampleControlInfoAttribute(string namespace, string controlName, string testScenario, Type viewModelType = null, bool ignoreInAutomatedTests = false, string description = null)
{
   ...
}
```

ghuntley:
@jeromelaban additionally can you please do the same with naming of folders and tests

image

ghuntley:
There is some content at https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md#creating-ui-tests which can be used as a baseline.

davidjohnoliver:
- Using namespaces as a category mainly makes sense for non-UI features. Sticking every control in the same category would be a bad idea. 
 - That said we should standardise on one convention - eg 'ContentDialog' over 'ContentDialogTests'. (It's a mess for mainly historical reasons.)
 - for the *folder name* however I prefer `/FooTests` over just `Foo`, simply because if the namespace is just `...Foo`, then trying to create `new Foo()` in code-behind (in *any* sample) gives rise to compiler complaints.

jeromelaban:
`ContentDialogTests` must be used to avoid automatic naming, which conflicts with the original control's name.

doc/articles/contributing/guidelines/code-style.md
```diff
@@ -32,19 +39,45 @@ if (condition)
...
 ## Integration Tests
 
 ```csharp
-[ActivePlatforms(Platform.Android, Platform.Browser, Platform.iOS)]
 [TestFixture]
 public class LocalSettings_Tests : SampleControlUITestBase
 {
     [Test]
     [AutoRetry]
+    [ActivePlatforms(Platform.Android, Platform.Browser, Platform.iOS)]
```
jeromelaban: This must not be the default. Tests should run on all platforms by default, even those not listed yet.

carldebilly:
Since the merging of the "clipping" PR, there's a new feature where the sample name is not required on the attribute anymore: it will use the class name instead.

Same for sample category: it will use the namespace name (local name, not full name)

jeromelaban:
Many of these items have been documented in more recent PRs.

PR #1939, opened 2019-10-23
Handles nullables types in XAML file generator

### LINKED ISSUE #1896: XamlCodeGenerator - Attached property of nullable int type
Hello,

In my Uno class library (using the last prerelease of Uno.UI), I have a custom attached property like this :
```csharp
public partial class NullableIntHelper
{
    public static readonly DependencyProperty NullableIntProperty = DependencyProperty.RegisterAttached(
         "NullableInt", typeof(int?), typeof(NullableIntHelper), new PropertyMetadata(0));
    public static int? GetNullableInt(DependencyObject d)
    {
        return (int?)d.GetValue(NullableIntProperty);
    }
    public static void SetNullableInt(DependencyObject d, int? value)
    {
        d.SetValue(NullableIntProperty, value);
    }
}
```

If I use it in a XAML file (`local:NullableIntHelper.NullableInt='19'`), I have an exception with the XamlCodeGenerator :
```
System.Exception: Unable to convert 19 for NullableInt with type int?
MSBUILD : error :    at Uno.UI.SourceGenerators.XamlGenerator.XamlFileGenerator.BuildLiteralValue(INamedTypeSymbol propertyType, String memberValue, XamlMemberDefinition owner, String memberName, String objectUid)
```

If I use an int value instead of int?, there is no problem.

Cordially,
Stephane.

kakone:
GitHub Issue : fixes #1896

## PR Type

- Bugfix

Nullables types in attached properties causes an exception in XAML file generation.

## What is the new behavior?

Handles nullables types in XAML file generator by getting the underlying type for nullable types.

## PR Checklist

### DIFF
--- a/src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.cs
+++ b/src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.cs
@@ -2818,7 +2818,7 @@ namespace Uno.UI.SourceGenerators.XamlGenerator
             return "";
         }
         
-        private INamedTypeSymbol GetUnderlyingType(INamedTypeSymbol propertyType)
+        private INamedTypeSymbol FindUnderlyingType(INamedTypeSymbol propertyType)
         {
             return (propertyType.IsNullable(out var underlyingType) && underlyingType is INamedTypeSymbol underlyingNamedType) ? underlyingNamedType : propertyType;
         }
@@ -2835,7 +2835,7 @@ namespace Uno.UI.SourceGenerators.XamlGenerator
                 }
             }
             
-            propertyType = GetUnderlyingType(propertyType);
+            propertyType = FindUnderlyingType(propertyType);
             switch (propertyType.ToDisplayString())
             {
                 case "int":
@@ -4050,12 +4050,12 @@ namespace Uno.UI.SourceGenerators.XamlGenerator
 
             if (propertyType != null)
             {
+                propertyType = FindUnderlyingType(propertyType);
                 if (propertyType.TypeKind == TypeKind.Enum)
                 {
                     return true;
                 }
-
-                propertyType = GetUnderlyingType(propertyType);
+               
                 switch (propertyType.ToDisplayString())
                 {
                     case "int":
--- a/src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.cs
+++ b/src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.cs
@@ -1,400 +1,396 @@
-﻿using Uno.Extensions;
-using Uno.MsBuildTasks.Utils;
-using Uno.MsBuildTasks.Utils.XamlPathParser;
-using Uno.UI.SourceGenerators.XamlGenerator.Utils;
-using System;
+﻿using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Globalization;
 using System.IO;
 using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
 using System.Text.RegularExpressions;
 using Microsoft.CodeAnalysis;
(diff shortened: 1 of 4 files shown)


src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.cs
```diff
@@ -4049,6 +4055,7 @@ private bool HasCustomInitializer(XamlObjectDefinition definition)
                     return true;
                 }
 
+                propertyType = GetUnderlyingType(propertyType);
```
jeromelaban: Should it be placed immediately above the test? That way, we'd support nullable enums as well. If so, could you also add a test for nullable enums ?

src/SourceGenerators/XamlGenerationTests.Core/TestAttachedPropertyOwner.cs
```diff
@@ -1,13 +1,10 @@
...
-using System.Text;
-
+﻿
 using Windows.UI.Xaml;
 
 namespace XamlGenerationTests.Core
 {
-    public static class TestAttachedPropertyOwner
-    {
+	public static class TestAttachedPropertyOwner
```
jeromelaban: Please preserve the same whitespaces (tabs)

src/SourceGenerators/XamlGenerationTests.Core/TestAttachedPropertyOwner.cs
```diff
@@ -60,7 +57,16 @@ public static void SetMoreContent(DependencyObject obj, object value)
 		public static readonly DependencyProperty MoreContentProperty =
 			DependencyProperty.RegisterAttached("MoreContent", typeof(object), typeof(TestAttachedPropertyOwner), new PropertyMetadata(null));
 
+		public static int? GetNullableType(DependencyObject obj)
```
jeromelaban: Use expression body where possible.
```suggestion
		public static int? GetNullableType(DependencyObject obj) => (int?)obj.GetValue(NullableTypeProperty);
```

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.cs
```diff
@@ -2817,6 +2817,11 @@ private string GetCastString(INamedTypeSymbol targetProperty, XamlObjectDefiniti
             }
             return "";
         }
+        
+        private INamedTypeSymbol GetUnderlyingType(INamedTypeSymbol propertyType)
```
jeromelaban: `FindUnderlyingType` or `TryFindUnderlyingType` would be more appropriate here, as it's not a general case.

kakone: OK, I chose `FindUnderlyingType` (I prefer than `TryFindUnderlyingType` because TryFind makes think that it returns a bool).

src/SourceGenerators/XamlGenerationTests.Core/TestAttachedPropertyOwner.cs
kakone: I also prefer but I respected the syntax in the file.
OK, I replaced my new line only by an expression body to keep the PR readable.

src/SourceGenerators/XamlGenerationTests.Core/TestAttachedPropertyOwner.cs
kakone: Done

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.cs
kakone: You are right, I made these changes.

kakone:
Changes done and I added a line in the release notes.

PR #2010, opened 2019-10-31
Punctuation errors

trishitapingolia:
Fixed punctuation errors


jeromelaban:
Thank you for you contribution, but this change is not required.

PR #1961, opened 2019-10-26
StorageFile:DateCreated support

pkar70:
GitHub Issue (If applicable): #1187 (partly)

## PR Type
What kind of change does this PR introduce?
- Feature

## What is the current behavior?
StorageFile:DateCreated is not supported 

## What is the new behavior?
StorageFile:DateCreated is supported

## PR Checklist
Please check if your PR fulfills the following requirements:
- [X] Tested code with current [supported SDKs](https://github.com/unoplatform/uno/blob/master/README.md#uno-features)
- [ ] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)
- [ ] [Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [ ] [Wasm UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md#running-the-webassembly-ui-tests-snapshots) are not showing unexpected any differences. Validate PR `Screenshots Compare Test Run` results.
- [X] Contains **NO** breaking changes
- [X] Updated the [Release Notes](https://github.com/unoplatform/uno/tree/master/doc/ReleaseNotes)
- [ ] Associated with an issue (GitHub or internal)

## Other information

Internal Issue (If applicable):

Issue #1187, opened 2019-07-04
Add support for Windows.Storage.FileIO and Windows.Storage.StorageFile


ghuntley:
## What would you like to be added:

- https://docs.microsoft.com/en-us/uwp/api/windows.storage.fileio
- https://docs.microsoft.com/en-us/uwp/api/windows.storage.storagefile

- https://github.com/unoplatform/uno/blob/master/src/Uno.UWP/Generated/3.0.0.0/Windows.Storage/FileIO.cs
- https://github.com/unoplatform/uno/blob/master/src/Uno.UWP/Generated/3.0.0.0/Windows.Storage/StorageFile.cs

## Why is this needed:

https://gitter.im/uno-platform/Lobby?at=5d1dc6f326206b667c88f970

## For which Platform:

- [x] iOS
- [x] Android
- [ ] WebAssembly
- [ ] Windows

## Anything else we need to know?

The main thing to keep in mind is how iOS and Android behave differently related to file-system sandboxing aka "special folders". The Storage APIs that return Paths are implemented, it's only the APIs that do access the files themselves that are not, mainly because System.IO does it properly cross platform. The only part that is currently implemented is the ability to open a file, then transform it from an IRandomAccessStream to a System.IO.Stream

MartinZikmund:
I'd love to take a look at this, although after 19th July, after I hand out my master thesis :-D .

MartinZikmund:
This should be fairly complete now, feel free to open additional issues for specific features as needed :-) .

PR #2036, opened 2019-11-06
StorageFile.DateCreated support

pkar70:
GitHub Issue (If applicable): #1187 (partly)

## PR Type

- Feature

StorageFile:DateCreated is not supported.

## What is the new behavior?
StorageFile:DateCreated is supported.

## PR Checklist

## Other information
Re-created #1961, with added test.

### DIFF
--- a/doc/ReleaseNotes/_ReleaseNotes.md
+++ b/doc/ReleaseNotes/_ReleaseNotes.md
@@ -3,7 +3,7 @@
 ## Next version
 
 ### Features
-* StorageFile:DateCreated is supported
+* Add support for `StorageFile.DateCreated`
 * [#1970](https://github.com/unoplatform/uno/pull/1970) Added support for `AnalyticsInfo` properties on iOS, Android and WASM
 * [#1207] Implemented some `PackageId` properties
 * [#1919](https://github.com/unoplatform/uno/pull/1919) Support for `PathGeometry` on WASM.
--- a/doc/ReleaseNotes/_ReleaseNotes.md
+++ b/doc/ReleaseNotes/_ReleaseNotes.md
@@ -3,7 +3,7 @@
 ## Next version
 
 ### Features
-
+* StorageFile:DateCreated is supported
 * [#1970](https://github.com/unoplatform/uno/pull/1970) Added support for `AnalyticsInfo` properties on iOS, Android and WASM
 * [#1207] Implemented some `PackageId` properties
 * [#1919](https://github.com/unoplatform/uno/pull/1919) Support for `PathGeometry` on WASM.
--- a/src/SamplesApp/SamplesApp.UITests/Windows_Storage/StorageFileTests/UnoSamples_Test_StorageFile_DateCreated.cs
+++ b/src/SamplesApp/SamplesApp.UITests/Windows_Storage/StorageFileTests/UnoSamples_Test_StorageFile_DateCreated.cs
@@ -1,0 +1,49 @@
+﻿using NUnit.Framework;
+using SamplesApp.UITests.TestFramework;
+using System;
+using System.Collections.Generic;
+using System.Drawing;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Uno.UITest.Helpers;
+using Uno.UITest.Helpers.Queries;
+
+namespace SamplesApp.UITests.Windows_Storage.StorageFileTests
+{
+	[TestFixture]
+	class UnoSamples_Test_StorageFile_DateCreated : SampleControlUITestBase
+	{
+		[Test]
+		[AutoRetry]
+		[ActivePlatforms(Platform.Android, Platform.iOS)]
+		public void StorageFile_DateCreated()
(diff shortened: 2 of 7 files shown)


src/SamplesApp/SamplesApp.UITests/Windows_Storage/StorageFileTests/UnoSamples_Test_StorageFile_DateCreated.cs
```diff
@@ -0,0 +1,49 @@
...
+
+			_app.WaitForText(resultMessage, "created");  // timeout here means: something impossible?
+
+			// step 2: test date just after create
+			_app.Tap(_app.Marked("sfdButtonTest"));
+			_app.WaitForText(resultMessage, "success");   // timeout here means: date is wrong
+			TakeScreenshot("AfterFirstCheck");
+
+			// step 3: test date several seconds later
+			_app.Wait(TimeSpan.FromSeconds(30));
```
carldebilly: This is a big problem...  a test cannot take 30 seconds just for itself on the build server!

Is there a way to reduce this to max few seconds?

Specially when the result is not validated by the test itself...

pkar70: I don't know how many seconds would be enough to be sure that returned date is correct.
As if returned date is something like 'Date.Now', test should wait sufficiently long before getting date second time. But what means sufficiently long, I don't know :)
Maybe 10 seconds would be enough for test, and not locking build server for too long?

carldebilly: We currently have more than 600 tests... imagine if all of them took 30 seconds... That would mean a 5 hours build!

You already have a test in the sample itself... it's not enough?

carldebilly:
Why did you create a new PR? The same change should stay on the same PR or it will be quickly very confusing.

src/SamplesApp/SamplesApp.UITests/Windows_Storage/StorageFileTests/UnoSamples_Test_StorageFile_DateCreated.cs
pkar70: What do you mean: "You already have a test in the sample itself"?

Test after some seconds is for checking that we get not current time, but time from creating file.
Of course, as this PR is only "a macro" for already implemented API, it probably could be skipped. But, you stated question: "Did you ensure it works on all supported platforms?", and this test is some way to get answer for this.

pkar70: tests are run sequentially?

carldebilly: Yes, because they are run on an emulator, so it's not possible to have many tests run at the same time.

pkar70:
I would change this PR... but after learning about nonUI tests, and converting test from using UI to not using UI.

doc/ReleaseNotes/_ReleaseNotes.md
```diff
@@ -3,7 +3,7 @@
 ## Next version
 
 ### Features
-
+* StorageFile:DateCreated is supported
```
```suggestion
* Add support for `StorageFile.DateCreated`
```

src/SamplesApp/SamplesApp.UITests/Windows_Storage/StorageFileTests/UnoSamples_Test_StorageFile_DateCreated.cs
jeromelaban: For the wait time, you do not need to wait long. A few milliseconds should be enough, as long as it's different, and a greater date.

pkar70: I would convert test, and add DeleteAsync, DateCreated, and also StorageFolder.CreateAsync(string).

Few milliseconds is not good enough, as:
1. filetime resolution on underlying file system;
2. time spent in OS creating file could be longer than few milliseconds;

If time change (DST on/off) would occur between two reads, test would fail, but I hope that this would not be a frequent case :)

carldebilly: I think build servers are running on UTC, a TZ without DST ;-)
Anyway, we can manage a failed build for those ;-)

pkar70:
Would be recreated as bigger PR (merge of two PR), StorageFile:DeleteAsync, StorageFile:DateCreated, with also StorageFolder:CreateFile, and with single test for all three methods.

PR #2037, opened 2019-11-06
Update Xamarin.Android to 10.0 in CI

maxim-lobanov:
To build for Android v10 we need to set `6_4_0` bundle in ADO.
It contains Xamarin.Android 10.0 and Mono 6.4.0.
Also need to set .NET Core 3.0.100 to support latest features.
Source: https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops#mono

### DIFF
--- a/.azure-devops-android-tests.yml
+++ b/.azure-devops-android-tests.yml
@@ -20,7 +20,7 @@ jobs:
       path: $(NUGET_PACKAGES)
     displayName: Cache NuGet packages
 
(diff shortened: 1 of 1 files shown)


jeromelaban:
@vincentcastagna in the future, if you see PRs that do not have validations running, please let me know. For this one we did not see if the build actually passed.

vincentcastagna:
@jeromelaban Its on a feature branch ... that's why :D

Issue #748, opened 2019-04-08
Support for special URIs


MartinZikmund:
## I'm submitting a...
- Feature request

## Current behavior
Special URIs like `ms-settings` etc. are not supported by the `Launcher.LaunchUriAsync` API.

## Expected behavior
When special URIs are passed into `Launcher`, the appropriate system app is launched on Android and iOS.

## Environment
Nuget Package: Latest

Package Version(s): Latest

Affected platform(s):
- [x] iOS
- [x] Android
- [ ] WebAssembly
- [ ] Windows
- [ ] Build tasks

Visual Studio
- [x] 2017 (version: )
- [ ] 2017 Preview (version: )
- [ ] for Mac (version: )

Relevant plugins
- [ ] Resharper (version: )
```

MartinZikmund:
Would you agree this would be useful addition?

TopperDEL:
I think it would - as this would help with the "just do it like on UWP an it works"-flow.

jeromelaban:
That would be interesting indeed. Which ones are you using ?

MartinZikmund:
I hit it with `ms-settings`, but I guess there are other which could be implemented as well like `ms-sms` or `ms-call`

MartinZikmund:
I have started working on this as well.

carldebilly:
We should dress a list of interesting urls to implement.

MartinZikmund:
@carldebilly I already had `ms-settings` implemented, made #2040 for this

MartinZikmund:
Leaving this unassigned for now, maybe more special URIs can be implemented in the future.

MartinZikmund:
Closing this as no other interesting special URIs are waiting to be implemented, separate issues can opened for specific use cases

PR #2040, opened 2019-11-06
Launcher ms-settings: special URIs, UWP alignment

MartinZikmund:
GitHub Issue (If applicable): #748 

## PR Type

- Feature

- [ ] Launcher does not support special `ms-settings` URI.
- [ ] Launcher may launch URI from a non-UI thread
- [ ] Launcher does not throw exceptions consistent with UWP

## What is the new behavior?

- [x] Launcher does not support special `ms-settings` URI on Android and partially on iOS
- [x] Launcher throws if URI launch is attempted from non-UI thread, in line with UWP
- [x] Launcher throws the same exceptions as UWP counterpart

## PR Checklist


src/Uno.UWP/System/Launcher.cs
```diff
@@ -1,105 +1,67 @@
...
-#endif
-
 namespace Windows.System
 {
 	public static partial class Launcher
 	{
-		public static async Task<bool> LaunchUriAsync(Uri uri)
+		private const string MicrosoftUriPrefix = "ms-";
+
+		private static bool IsSpecialUri(Uri uri) => uri.Scheme.StartsWith(MicrosoftUriPrefix, StringComparison.InvariantCultureIgnoreCase);
```
carldebilly: I'm unsure that an uri beginning with `ms-` should be considered as a _Microsoft Uri_. It's a stretch, but you should fully qualify the scheme.

src/Uno.UWP/System/Launcher.iOS.SpecialUris.cs
```diff
@@ -0,0 +1,36 @@
+﻿#if __IOS__
```
carldebilly: You seems to have an implementation for iOS, but the documentation is only about Android.

MartinZikmund: I mention there that "In case of iOS, any such API opens the main page of system settings.", but I will add it to separate paragraph to make it more visible

src/Uno.UWP/System/Launcher.cs
MartinZikmund: The reasoning here is that all the special URIs in UWP start with `ms-`, so this is just the very "raw" filter. In the implementation I then check if the URI is actually supported (by comparing to the known URIs, currently `ms-settings` and if not, the handling proceeds in normal manner, so it should not cause any issues

src/Uno.UWP/System/Launcher.iOS.SpecialUris.cs
MartinZikmund: Done :-)

src/Uno.UI.RuntimeTests/Tests/Windows_System/Given_Launcher.cs
```diff
@@ -1,49 +1,94 @@
...
 {
 	[TestClass]
 	public class Given_Launcher
 	{
-		private async Task Dispatch(DispatchedHandler p)
+		[TestMethod]
+		public async Task When_Null_Uri_Is_Launched()
 		{
-			await CoreApplication.GetCurrentView().Dispatcher.RunAsync(Windows.UI.Core.CoreDispatcherPriority.Normal, p);
+			await Assert.ThrowsExceptionAsync<AccessViolationException>(() => Launcher.LaunchUriAsync(null).AsTask());
```
jeromelaban: it's most probably better to not throw this exception, even if it's what WinRT does. Returning an argument null exception makes more sense :)

src/Uno.UWP/System/Launcher.Android.SpecialUris.cs
```diff
@@ -0,0 +1,99 @@
...
+using System;
+using Android.Provider;
+
+namespace Windows.System
+{
+	public static partial class Launcher
+	{
+		private const string MicrosoftSettingsUri = "ms-settings";
+
+		private static readonly (string uriPrefix, string intent)[] _settingsHandlers = new (string uriPrefix, string intent)[]
```
jeromelaban: Why not a Dictionary<string, string> ?

src/Uno.UWP/System/Launcher.cs
```diff
@@ -1,105 +1,70 @@
...
-							new AppleUrl(uri.OriginalString));
-#elif __ANDROID__
-						var androidUri = Android.Net.Uri.Parse(uri.OriginalString);
-						var intent = new Intent(Intent.ActionView, androidUri);
+#if __IOS__ || __ANDROID__ || __WASM__
+
+			if (uri == null)
+			{
+				// counterintuitively, UWP throws AccessViolationException when passed in Uri is null
+				throw new AccessViolationException("Attempted to read or write protected memory. This is often an indication that other memory is corrupt.");
```
jeromelaban: That seems more like a WinRT bug than anything else :)

src/Uno.UI.RuntimeTests/Tests/Windows_System/Given_Launcher.cs
MartinZikmund: I'm thinking only if this would not mean that existing code which, for some reason, handles `AccessViolationException` here, as the developer didn't know the reason for it happening; would behave differently in Uno than in UWP? But agree that `ArgumentNullException` would make much more sense, and if you see it as a better solution, I will replace it there 😀

src/Uno.UWP/System/Launcher.Android.SpecialUris.cs
MartinZikmund: That's a veeeeery good question 😂 .

src/Uno.UWP/System/Launcher.cs
MartinZikmund: Especially when one method throws `AccessViolationException` and another throws `Exception`. Kind of confusing :-D .

MartinZikmund: Fixed :-)

src/Uno.UWP/System/Launcher.Android.SpecialUris.cs
MartinZikmund: Fixed :-)

src/Uno.UI.RuntimeTests/Tests/Windows_System/Given_Launcher.cs
MartinZikmund: Fixed :-)

src/Uno.UWP/System/Launcher.Android.SpecialUris.cs
```diff
@@ -0,0 +1,95 @@
...
+			{"defaultapps", Settings.ActionManageDefaultAppsSettings},
+			{"appsforwebsites", Settings.ActionManageDefaultAppsSettings},
+			{"cortana", Settings.ActionVoiceInputSettings},
+			{"bluetooth", Settings.ActionBluetoothSettings},
+			{"printers", Settings.ActionPrintSettings},
+			{"typing", Settings.ActionHardKeyboardSettings},
+			{"easeofaccess", Settings.ActionAccessibilitySettings},
+			{"network-airplanemode", Settings.ActionAirplaneModeSettings},
+			{"network-celluar", Settings.ActionNetworkOperatorSettings},
+#if __ANDROID_28__
```
MatFillion: this define is when the app is built for android 9 but do we handle the app not running on android28 ? What happens in that case ?

src/Uno.UWP/System/Launcher.cs
```diff
@@ -1,105 +1,70 @@
...
-						{
-							throw new InvalidOperationException(
-								"LaunchUriAsync was called too early in application lifetime. " +
-								"App context needs to be initialized");
-						}
-						((Android.App.Activity)Uno.UI.ContextHelper.Current).StartActivity(intent);
+			if (!CoreDispatcher.Main.HasThreadAccess)
+			{
+				// LaunchUriAsync throws the following exception if used on UI thread on UWP
+				throw new InvalidOperationException("A method was called at an unexpected time.");
```
MatFillion: I would suggest adding a error log here saying LaunchUriAsync should be call on the CoreDispatcher/MainThread.
It would help developers because the exception might not be obvious to everyone.

src/Uno.UWP/System/Launcher.Android.SpecialUris.cs
MartinZikmund: Fixed, made this `Lazy`, so that this particular link is added only if running on a supported OS. Previous versions will just fallback to `network`, which gives the networking settings main page 😉

src/Uno.UWP/System/Launcher.cs
MartinZikmund: Done 🙂

MartinZikmund:
@MatFillion The build ended with a weird error, can you please requeue it?

MartinZikmund:
@MatFillion Ah, bad luck, build failed as mono SDK could not be downloaded. Can you please re-queue it once again?

jeromelaban:
This PR introduced a breaking change that was not caught by the API diffing tool: https://github.com/unoplatform/uno/issues/2239

PR #1726, opened 2019-10-04
Clipping refactoring

carldebilly:
# Bugfix / Refactoring / Enhancement
Refactored the way the clipping is calculated on Uno

All controls were clippings to its bounds when it's not how it's working on WinUI:
* A control is clipped to its bound when its `desiredSize` is close or greater and its `arrangeSize`
* A canvas will never got clipped
* A scrollviewer (`ScrollViewerPresenter`) will always clip to its bounds
* A `ContentControl` won't clip if there's a `RenderTransform` on its content
* Any control capable of `CornerRadius` will clip when its value is `> 0.0`

## PR Checklist

Please check if your PR fulfills the following requirements:

- [X] Tested code with current [supported SDKs](https://github.com/unoplatform/uno/blob/master/README.md#uno-features)
- ~~[ ] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)~~
- [X] [Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [X] [Wasm UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md#running-the-webassembly-ui-tests-snapshots) are not showing unexpected any differences. Validate PR `Screenshots Compare Test Run` results.
- [X] Contains **NO** breaking changes (Canaries has been setup to reduce breaking changes)
- [ ] Updated the [Release Notes](https://github.com/unoplatform/uno/tree/master/doc/ReleaseNotes)
- [X] Associated with an issue (GitHub or internal)

## Related Issues
* https://github.com/unoplatform/uno/issues/652 Clipping does not works on Android
* https://github.com/unoplatform/uno/issues/1599 [Android] DoubleAnimation cause clipping issue
* https://github.com/unoplatform/uno/issues/1074 CornerRadius should clip its children
* https://github.com/unoplatform/uno/issues/2006 [Android] Animating a WebView removes its natural clipping
* https://github.com/unoplatform/uno/issues/649 [Android] Touch events are not dispatched properly with clipping and transforms
* https://github.com/unoplatform/uno/issues/1509 [Droid][WASM] CommandBar isn't visible
* https://nventive.visualstudio.com/Umbrella/_workitems/edit/165472 Clipping
[...]

PR #2049, opened 2019-11-08
Fixed clipping regression on Android

carldebilly:
## Bugfix
A bug was remaining in previous PR #1726 where clipping on Android were not applied properly.

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/Layouter/Layouter.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Layouter/Layouter.cs
@@ -196,7 +196,7 @@ namespace Windows.UI.Xaml.Controls
 						needsClipToSlot = true;
 					}
 
-					if (IsLessThanAndNotCloseTo(arrangeSize.Height, _unclippedDesiredSize.Height))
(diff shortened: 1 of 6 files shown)


src/Uno.UI/UI/Xaml/Controls/Layouter/Layouter.cs
```diff
@@ -532,108 +535,55 @@ private Rect ApplyMarginAndAlignments(View view, Rect frame)
...
 		{
 			//Default value
 			//childVerticalAlignment != VerticalAlignment.Stretch
-			var actualHeight = Min(height, desiredSize.Height);
+			var actualHeight = Min(size, desiredSize);
 
 			if (hasChildHeight)
 			{
-				actualHeight = Min(childHeight + childMargin.Top + childMargin.Bottom, height);
+				actualHeight = Min(childHeight + childMargins, size);
```
davidjohnoliver: All references to Height should be replaced with Size

carldebilly: Yep you're right!  It's already merged, so a new PR is required for that...

Issue #1928, opened 2019-10-22
ResourceLoader.GetForCurrentView().GetString() does not support /[file]/[name] format


pkar70:
## Current behavior
ResourceLoader.GetForCurrentView().GetString("/file/name") return empty string
As including here png file doesnt work ("something went really wrong, and we can't process that file"), check these screenshots:
[android](http://beskid.geo.uj.edu.pl/p/ssc/android.png)
[wasm](http://beskid.geo.uj.edu.pl/p/ssc/wasm.png)

## Expected behavior
return of value "name" from "file.resw" - as in UWP.
[uwp](http://beskid.geo.uj.edu.pl/p/ssc/uwp.png)

## How to reproduce it (as minimally and precisely as possible)
I don't know how to include .zip file here...
[repro](http://beskid.geo.uj.edu.pl/p/ssc/TestUnoResw.zip)

Shared\Strings\en\Resources.resw (only not default lines)
`  <data name="resUniq" xml:space="preserve">`
`    <value>Uniq text from Resources</value>`
`  </data>`
`  <data name="t1t" xml:space="preserve">`
`    <value>text from Resources</value>`
`  </data>`

Shared\Strings\en\tezy.resw (only not default lines)
`  <data name="t1t" xml:space="preserve">`
`    <value>text fom Tezy</value>`
`  </data>`
`  <data name="tezyUniq" xml:space="preserve">`
`    <value>Uniq text from tezy</value>`
`  </data>`

Shared\MainPage.xaml

`<Page
    x:Class="TestUnoResw.MainPage"
    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
    xmlns:local="using:TestUnoResw"
    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
    mc:Ignorable="d" Loaded="Page_Loaded">`
`    <StackPanel Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">`
` <TextBlock Name="uiText" Text="Should be resources.resw" Margin="20" FontSize="30" />`
`  <TextBlock Name="uiTextUniq" Text="Should be resources.resw (not in tezy)" Margin="20" FontSize="30" /> `
[...]

PR #2050, opened 2019-11-08
support for /[file]/[name] format in ResourceLoader.GetForCurrentView…

pkar70:
GitHub Issue (If applicable): #1928

## PR Type

- Feature

ResourceLoader.GetForCurrentView().GetString("/file/name") returns empty string

## What is the new behavior?
ResourceLoader.GetForCurrentView().GetString("/file/name") returns string from selected resources file.

## PR Checklist

- [X] Tested code with current [supported SDKs]  (tested as patch in 2.0.311-dev.3092 build).
- [ ] Docs have been added/updated which fit 
- [ ] [Wasm UI Tests]

## Other information
 Made possible only because of Jerome Laban comment in Issue.
Re-created PR#1938, with added unit tests, and with changes suggested in 1938. But it cannot be done as commits in 1938, as I deleted my previous fork of Uno.


src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.cs
```diff
@@ -48,6 +48,15 @@ public ResourceLoader()
 
 		public string GetString(string resource)
 		{
+			// "/[file]/[name]" format support
+			if (resource.ElementAtOrDefault(0) == '/')
+			{
+				var separatorIndex = resource.IndexOf("/", 1);
```
jeromelaban: Thanks for the modifications!

There should be a validation of the format though (and a test that validates that as well).

For instance, if the provided string is `/myfile` the substring will fail with an exception. With UWP, such a lookup returnns an empty string, so this will do:
```suggestion
				var separatorIndex = resource.IndexOf("/", 1);
				
				if(separatorIndex != null)
				{
```

With an additional test that validates that `/this_is_not_a_valid_file` returns `""`

pkar70: Very good point...
But why null (`!= null`)? It cannot be null (=0), 'not found' is -1 :)

pkar70: File not found - UWP returns "",
This test (for 'file exist') should be in GetForCurrentView(resourceFile) method, so it is something outside my PR (and I'm lost in code flow).

src/Uno.UI.RuntimeTests/Tests/Windows_ApplicationModel_Resources/Given_ResourceLoader.cs
```diff
@@ -131,5 +131,17 @@ public void When_MissingLocalizedResource_FallbackOnDefault()
...
+
+		[TestMethod]
+		public void When_FileAndStringNameFormat()
+		{
+			var SUT = ResourceLoader.GetForViewIndependentUse();
+
+			Assert.AreEqual(@"Uniq text from Resources", SUT.GetString("pkResUniqResources"));
+			Assert.AreEqual(@"Shared text from Resources", SUT.GetString("pkResShared"));
+			Assert.AreEqual(@"Shared text from Test01", SUT.GetString("/Test01/pkResShared"));
+			Assert.AreEqual(@"Uniq text from Test01", SUT.GetString("/Test01/pkResUniqTest01"));
```
jeromelaban: Thanks for the update!

Some tests to validate that behavior:
```suggestion
			Assert.AreEqual(@"Uniq text from Test01", SUT.GetString("/Test01/pkResUniqTest01"));
			Assert.AreEqual(@"", SUT.GetString("/this-does-not-exist"));
			Assert.AreEqual(@"", SUT.GetString("//this-does-not-exist"));
```

PR #2251, opened 2019-12-03
fixed Shapes pointer events hit-box for wasm

XiaotianNetlift:

## PR Type
- Bugfix

On wasm, pointer events for `Shape` classes trigger on the entire area occupied by the shape, instead of just the visible area. Additionally, on wasm, the `PointerRoutedEventArgs.OriginalSource` is incorrectly set to the shadow `SvgElement` instead of the actual `Shape`.

## What is the new behavior?
Aligned wasm behaviors to be like on uwp.

## PR Checklist

## Other information
In order for pointer events to be trigger only on the visible area in html, the `<svg>` needs to have the style `pointer-events: none`  and the visible elements (`<path>`, `<circle>`, etc.)  should the style `pointer-events: auto`. This is done by adjusting `IsViewHit` property.
By doing so, the `OriginalSource` (`event.target`) is now the visible elements, and we need to change it for the first `<svg>` parent which is the correct original source.


jeromelaban:
We probably should be removing all the SVG shadow elements from the public API.

XiaotianNetlift:
todo:
- [ ] remove shadow svg elements from public api
  - [ ] update `PackageDiffIgnore.xml`

XiaotianNetlift:
@jeromelaban I changed the shadow elements access modifier to `internal`. As a side effect of that, the `Shape` class is derivable outside of `uno.ui`, but its concrete derived classes can still be derived. (All due to `GetMainSvgElement()` being `internal protected abstract`.)

Not sure if this is desired.

src/Uno.UI/UI/Xaml/Media/LinearGradientBrush.wasm.cs
```diff
@@ -2,7 +2,7 @@
 using System.Linq;
 using Windows.Foundation;
 using Uno.Extensions;
-using Windows.UI.Xaml.Wasm;
+using static Windows.UI.Xaml.Shapes.Shape;
```
davidjohnoliver: why this static using?

src/Uno.UI/UI/Xaml/Shapes/DefsSvgElement.wasm.cs
```diff
@@ -2,9 +2,9 @@
 using System.Collections.Generic;
 using System.Text;
 using Windows.UI.Xaml.Controls;
-using Windows.UI.Xaml.Wasm;
+using static Windows.UI.Xaml.Shapes.Shape;
```
davidjohnoliver: why this static using?

davidjohnoliver:
> @jeromelaban I changed the shadow elements access modifier to `internal`. As a side effect of that, the `Shape` class is derivable outside of `uno.ui`, but its concrete derived classes can still be derived. (All due to `GetMainSvgElement()` being `internal protected abstract`.)
> 
> Not sure if this is desired.

If the `SVGElement` type is still accessible to inheriting classes outside the assembly, then it's still part of the public API.

src/Uno.UI/UI/Xaml/Media/LinearGradientBrush.wasm.cs
XiaotianNetlift: `Windows.UI.Xaml.Wasm.SvgElement` is now
`Windows.UI.Xaml.Shapes.Shape+SvgElement` (a nested class of SvgElement).

It was moved in an _attempt_ to hide it. It is awkward because we need for:
`abstract SvgElement Shape::GetMainSvgElement`
But, at the same time, this doesn't align with UWP api, so it needs to be hidden.

src/Uno.UI.Wasm/ts/WindowManager.ts
```diff
@@ -889,16 +889,30 @@
...
+						srcHandle = handle;
+						break;
+					}
 				}
 
 				src = src.parentElement;
 			}
 
-			return `${evt.pointerId};${evt.clientX};${evt.clientY};${(evt.ctrlKey ? "1" : "0")};${(evt.shiftKey ? "1" : "0")};${evt.button};${evt.pointerType};${srcHandle};${evt.timeStamp}`;
+			return [
```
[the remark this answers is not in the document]
XiaotianNetlift: Nothing, but a cosmetic change for readability. I could drop this if it is not desired.

davidjohnoliver:
If you try to derive from `Shape` on UWP:
image

src/Uno.UI/UI/Xaml/Shapes/DefsSvgElement.wasm.cs
jeromelaban: Indeed, don't use static usings for this, this gets confusing, we're not at that level of character economy ;)

src/Uno.UI/UI/Xaml/Media/LinearGradientBrush.wasm.cs
jeromelaban: There's no need to make it nested, just internal. Is there a problem with hiding it ?

src/Uno.UI/UI/Xaml/Shapes/Line.iOSmacOS.cs
```diff
@@ -11,6 +11,13 @@ namespace Windows.UI.Xaml.Shapes
 {
 	public partial class Line
 	{
+		public Line()
+		{
+#if __IOS__
+			ClipsToBounds = false;
```
carldebilly: If you need that it means there's a bug un the layouter.  Since the layouter is responsible to handler this property, you should not set it here.

Please implement `ICustomClippingElement` instead.

(I know you just moved the code... but it's still problematic)

src/Uno.UI/UI/Xaml/Media/LinearGradientBrush.wasm.cs
carldebilly: I don't see the point in hiddin it.  It could have value for wasm applications even if it's not in the UWP namespace.  I suggest to move it to a `Uno.UI` namespace instead of hidding it.

davidjohnoliver: It turns out it's impossible in practice to inherit from `Shape` in UWP (see my comment below), so there's not much value in exposing it.

XiaotianNetlift: @jeromelaban There was one:
```
internal protected abstract SvgElement Shape.GetMainSvgElement()
```
And, by making `SvgElement` internal, the compiler will complain that:
> CS0050 Inconsistent accessibility: return type 'SvgElement' is less accessible than method 'Shape.GetMainSvgElement()'

The nested internal protected was what I found out to work around this limitation. I guess @carldebilly's suggestion will be a better approach, essentially reverting my last commit.

XiaotianNetlift: Would `Uno.UI.Xaml.Shapes.SvgElement` be a good place for this?

XiaotianNetlift:
> id

XiaotianNetlift:
> @jeromelaban I changed the shadow elements access modifier to `internal`. As a side effect of that, the `Shape` class is not derivable outside of `uno.ui`, but its concrete derived classes can still be derived. (All due to `GetMainSvgElement()` being `internal protected abstract`.)
> 
> Not sure if this is desired.

Just an update this change has been dropped.

src/Uno.UI/UI/Xaml/Shapes/Line.iOSmacOS.cs
```diff
@@ -11,6 +11,13 @@ namespace Windows.UI.Xaml.Shapes
 {
 	public partial class Line
 	{
+		public Line()
```
carldebilly: Codacy Issue found: [Remove this redundant constructor.](https://app.codacy.com/gh/unoplatform/uno/pullRequest?prid=4826527)

carldebilly:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
+ Solved 3
- Added 1
           

Complexity increasing per file
==============================
- src/Uno.UI/UI/Xaml/Shapes/SvgElement.wasm.cs  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/unoplatform/uno/pullRequest?prid=4826527&bid=16039848)

XiaotianNetlift:
Shadowed by #2692.

PR #2300, opened 2019-12-11
[Wasm] Fix LaunchUriAsync should not validate dispatcher

jeromelaban:
## PR Type

- Bugfix

## What is the new behavior?

Fixed LaunchUriAsync fails on Wasm.

This kind of code cannot yet be validated because of the opening of an external tab. Will need to be adjusted once threading support is enabled (https://github.com/unoplatform/uno/issues/2302).

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UWP/System/Launcher.cs
+++ b/src/Uno.UWP/System/Launcher.cs
@@ -25,6 +25,7 @@ namespace Windows.System
 				throw new ArgumentNullException(nameof(uri));
 			}
 
+#if !__WASM__
 			if (!CoreDispatcher.Main.HasThreadAccess)
 			{
 				if (typeof(Launcher).Log().IsEnabled(LogLevel.Error))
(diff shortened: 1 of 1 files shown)


src/Uno.UWP/System/Launcher.cs
```diff
@@ -25,6 +25,7 @@ public static Task<bool> LaunchUriAsync(Uri uri)
 				throw new ArgumentNullException(nameof(uri));
 			}
 
+#if !__WASM__
```
davidjohnoliver: For this type of case there's also `CoreDispatcher.CheckThreadAccess()` which is already disabled for WASM. (The advantage being having less places in code that need to be changed when threading becomes available)

src/Uno.UWP/System/Launcher.cs
```diff
@@ -58,7 +60,11 @@ public static Task<bool> LaunchUriAsync(Uri uri)
 			}
 
 			// this method may run on the background thread on UWP
+#if !__WASM__
```
davidjohnoliver: Maybe we should have a `HasThreadAccessInternal` property (aka `HasThreadAccessButReally`), again thinking about centralization

jeromelaban: We're going to have to look up for other places when https://github.com/unoplatform/uno/issues/2302 will be enabled and remove #if blocks in any cases (even if we make another property for that). Let make sure we scan through properly.

Issue #2287, opened 2019-12-09
[Android] ScrollViewer inside ListView bugs the horizontal scrolling


jcantin-nventive:
## Current behavior

When a horizontal scrolling gesture includes vertical scrolling, then the horizontal scrolling stops being recognized, even if we resume horizontally scrolling in the same gesture.

## Expected behavior

Horizontal scrolling is recognized.

## How to reproduce it (as minimally and precisely as possible)

Launch the app from [this repo](https://github.com/jcantin-nventive/button-horizontal-scroll).
Try to scroll the items horizontally with a gesture that includes a bit of diagonal movement (not perfectly horizontal).
This works fine on iOS but not on Android.

## Environment

Nuget Package: Uno.UI

Package Version(s): 1.45.0 and 2.1.0-dev.189

Affected platform(s):

- [ ] iOS
- [x] Android
- [ ] WebAssembly
- [ ] WebAssembly renderers for Xamarin.Forms
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

Visual Studio:

- [x] 2017 (version: 15.9.11)
- [x] 2019 (version: 16.4.0 preview 6.0)
- [ ] for Mac (version: )

Relevant plugins:

- [ ] Resharper (version: )

jcantin-nventive:
I'm no longer able to repro this issue in my samples. I'll keep looking into it but will close this for the moment

jcantin-nventive:
After futher investigations, I've reproduced the issue in the sample and I'm reopening the issue.

davidjohnoliver:
Brief investigation results: this seems to initially work as expected, then something falls into a bad state and the inner ScrollView no longer captures the gesture properly.

davidjohnoliver:
Updated the sample to latest Uno and haven't been able to repro.

@jcantin-nventive wasn't this fixed by #2338?

jcantin-nventive:
> Updated the sample to latest Uno and haven't been able to repro.
> 
> @jcantin-nventive wasn't this fixed by #2338?

If I remember correctly, my PR helped but didn't entirely solve the issue.

If it's now fixed, that's good.

davidjohnoliver:
The attached repro definitely seems to be fixed. I'll close this issue, feel free to reopen it if you notice the bug occurring in a different context. :)

PR #2338, opened 2019-12-16
[Android] Fix horizontal scrolling when inside vertical ListView

jcantin-nventive:
GitHub Issue (If applicable): #[2287](https://github.com/unoplatform/uno/issues/2287)

## PR Type

- Bugfix

When a ListView contains a horizontal scrollViewer, only the first horizontal scroll gesture works correctly. After that, when the gesture contains any amount of vertical motion, the horizontal scrolling stops and does not resume until the gesture is stopped and restarted.

## What is the new behavior?

When a ListView contains a horizontal scrollViewer, horizontal scrolling works correctly.

## PR Checklist

## Other information


jeromelaban:
@jcantin-nventive could you add a UI Test ? Similar to this one: 

https://github.com/unoplatform/uno/blob/4e330b3862626d316fb0657194f1226d3d1d2057/src/SamplesApp/SamplesApp.UITests/Windows_UI_Xaml_Input/Manipulation_Tests.cs#L17-L25

The end screenshot can validate the actual scrolling happened, not using a straight line.

jcantin-nventive:
> @jcantin-nventive could you add a UI Test ? Similar to this one:
> 
> https://github.com/unoplatform/uno/blob/4e330b3862626d316fb0657194f1226d3d1d2057/src/SamplesApp/SamplesApp.UITests/Windows_UI_Xaml_Input/Manipulation_Tests.cs#L17-L25
> 
> The end screenshot can validate the actual scrolling happened, not using a straight line.

The automatic UI test framework does not currently support the type of complex gestures that can reproduce this particular bug. Therefore I added explanations to the sample so that it can be tested manually.

src/Uno.UI.BindingHelper.Android/Uno/UI/UnoMotionHelper.java
```diff
@@ -181,6 +181,14 @@ private boolean dispatchMotionEventCore(Uno.UI.MotionTargetAdapter adapter, Moti
 		if (adapter.getCustomDispatchIsActive()) {
 			// Log.i(LOGTAG, _indent + "CUSTOM dispatch (" + target.getChildrenRenderTransformCount() + " of " + view.getChildCount() + " children are transformed )");
 
+			// Because we do not call dispatchToSuper(), we must manually call requestDisallowInterceptTouchEvent in the same circumstances
```
davidjohnoliver: A+ comment 👍

src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/ListView/ListView_WithScrollViewer.xaml
```diff
@@ -0,0 +1,128 @@
...
+			 xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+			 mc:Ignorable="d"
+			 d:DesignHeight="300"
+			 d:DesignWidth="400">
+
+	<Grid>
+
+		<ListView ItemsSource="{Binding SingleItemList}"
+				  Header="{Binding}"
+				  Grid.Row="2">
```
agneszitte-nventive: Not needed I think in this case no ?

src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/ListView/ListView_WithScrollViewer.xaml
```diff
@@ -0,0 +1,128 @@
...
+				<DataTemplate>
+					<StackPanel>
+						<TextBlock Text="{Binding HorizontalScrollDescription}"
+								   TextWrapping="Wrap"/>
+
+						<Image Source="ms-appx:/Assets/ScrollGesture.png"
+							   Stretch="Uniform"
+							   HorizontalAlignment="Left"
+							   Height="50"
+							   Grid.Row="1" />
```
agneszitte-nventive: Same here, not needed normally

jcantin-nventive: done

src/Uno.UI.BindingHelper.Android/Uno/UI/UnoMotionHelper.java
dr1rrb: Actually I don't understand what the purpose of this change ...

First I think that the `view.requestDisallowInterceptTouchEvent` should have been handled at [line 148 - 151](https://github.com/unoplatform/uno/pull/2338/files#diff-365e56e68abf23c7d1195c0de11ab3a9R148) in order to be consistent

But the main issue for me is that from what I understand here is that as soon as we have an element with a `RenderTransform` it will prevent any parent `ScrollViewer` to do it job. It seems a bit restrictive ...

@jcantin-nventive

dr1rrb: Actually as it set it to `false` it's the opposite, so it means that it breaks the manipulation events when in a scroll viewer.

davidjohnoliver: @dr1rrb It's just replicating the behaviour of the `super` method, specifically this line: https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/view/ViewGroup.java;l=2839?q=resettouchstate

dr1rrb: Ok that gives more context. 

So in my point of view, for the _down_ ([line 2600 of the ViewGroup](https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/view/ViewGroup.java;bpv=1;bpt=1;l=2600)) it means that our [lines 148-151](https://github.com/unoplatform/uno/blob/master/src/Uno.UI.BindingHelper.Android/Uno/UI/UnoMotionHelper.java#L148) are not at the right place and should be moved **after** the super dispatch as `super` will erase what just did! (Basically I would do the exact opposite of this change: make sure that the `view.requestDisallowInterceptTouchEvent(true)` is properly set.) 

I would also agree to force a reset for the _up_, _cancel_ and _move over_ (even if I would argue about the utility for the last one) after a "custom dispatch".

And reading this, I realized that there is a **major** difference between the `resetTouchState` and the `view.requestDisallowInterceptTouchEvent`: the first one is only for safety and reset a **local** flag, while the second will bubble up in the whole visual tree. I'm pretty sure the real issue is here: we request a disallow on a control on line 151, then invoke the super which reset only the local flag (ViewGroup@2600) which drives us into an invalid state which cannot be recovered properly.

dr1rrb: And actually if we look at the implementation of the [`requestDisallowInterceptTouchEvent` in the `ViewGroup`](https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/view/ViewGroup.java;l=3168), it does not propagates the request if the **local flag** is not set ... it really looks like the described issue.

dr1rrb: FYI I fixed this in #2445 I've validated your test, and it seems to be OK, but let me know if you have any issue with the updated fix.

Issue #2257, opened 2019-12-04
Pointer events has wrong location when in a ViewBox on Android


dr1rrb:
## Current behavior
Pointer location are not valid

## Expected behavior
As UWP

## How to reproduce it (as minimally and precisely as possible)
```xaml
<Grid Background="#FF583793">
        <Viewbox  Stretch="Uniform" StretchDirection="Both">
            <Canvas x:Name="mainCanvas" Height="800" Width="800" Background="#FFA0ACA5" Margin="50,50,50,50" PointerPressed="UIElement_OnPointerPressed">
                <TextBlock x:Name="tblInfos" Height="64" Canvas.Left="10" Canvas.Top="10" Text="TextBlock" TextWrapping="Wrap" Width="780" FontSize="20" FontWeight="Bold"/>
                <TextBlock x:Name="tblInfos2" Height="64" Canvas.Left="10" Canvas.Top="63" Text="TextBlock" TextWrapping="Wrap" Width="780" FontSize="20" FontWeight="Bold"/>
                <Border x:Name="letter" BorderThickness="1" BorderBrush="Black" Height="200" Canvas.Left="50" Canvas.Top="300" Width="500" Background="#FF8D7E3B" HorizontalAlignment="Stretch"/>
            </Canvas>
        </Viewbox>

    </Grid>
```
```cs
        private void UIElement_OnPointerPressed(object sender, PointerRoutedEventArgs e)
        {
            var posCanvas = e.GetCurrentPoint(mainCanvas).Position;
            var posLetter = e.GetCurrentPoint(letter).Position;
            tblInfos.Text = $"canvas x:{posCanvas.X} y:{posCanvas.Y}";
            tblInfos2.Text = $"letter x:{posLetter.X} y:{posLetter.Y}";

        }
```

## Environment
Nuget Package: **Uno.UI**
Package Version(s): **2.0**
Affected platform(s):

- [?] iOS
- [x] Android
- [ ] WebAssembly
- [ ] WebAssembly renderers for Xamarin.Forms
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

Visual Studio: **_irrelevant_**
Relevant plugins: **_none_**

## Anything else we need to know?
https://stackoverflow.com/questions/59027861/content-not-centered-in-viewbox-and-incorect-mouse-coord-in-viewbox

agneszitte-nventive:
@jeromelaban / @dr1rrb 
[...]

Issue #2256, opened 2019-12-04
ViewBox default alignment


dr1rrb:
## Current behavior
The default alignment of the `ViewBox` is top-left

## Expected behavior
The default alignment of the `ViewBox` should be center-center

## How to reproduce it (as minimally and precisely as possible)
```xaml
<Grid>
	<Grid.RowDefinitions>
		<RowDefinition />
		<RowDefinition />
		<RowDefinition />
		<RowDefinition />
		<RowDefinition />
	</Grid.RowDefinitions>
	<Grid.ColumnDefinitions>
		<ColumnDefinition />
		<ColumnDefinition />
	</Grid.ColumnDefinitions>

	<Viewbox Grid.ColumnSpan="2" Stretch="None">
		<Border Width="100" Height="100" Background="DeepSkyBlue">
			<TextBlock Text="Default" />
		</Border>
	</Viewbox>

	<Viewbox Grid.Row="1" Grid.Column="0" VerticalAlignment="Top" HorizontalAlignment="Left" Stretch="None">
		<Border Width="100" Height="100" Background="#FF69B4">
			<TextBlock Text="ViewBox top-left" TextWrapping="Wrap" />
		</Border>
	</Viewbox>
	<Viewbox Grid.Row="1" Grid.Column="1" VerticalAlignment="Top" HorizontalAlignment="Right" Stretch="None">
		<Border Width="100" Height="100" Background="#FF0000" Margin="0,0,1,1">
			<TextBlock Text="ViewBox top-right" TextWrapping="Wrap" />
		</Border>
	</Viewbox>
	<Viewbox Grid.Row="2" Grid.Column="0" VerticalAlignment="Bottom" HorizontalAlignment="Left" Stretch="None">
		<Border Width="100" Height="100" Background="#FF8E00" Margin="0,0,1,1">
			<TextBlock Text="ViewBox bottom-left" TextWrapping="Wrap" />
		</Border>
	</Viewbox>
	<Viewbox Grid.Row="2" Grid.Column="1" VerticalAlignment="Bottom" HorizontalAlignment="Right" Stretch="None">
		<Border Width="100" Height="100" Background="#FFFF00" Margin="0,0,1,1">
			<TextBlock Text="ViewBox bottom-right" TextWrapping="Wrap" />
		</Border>
	</Viewbox>

	<Viewbox Grid.Row="3" Grid.Column="0"  Stretch="None">
[...]

Issue #2135, opened 2019-11-21
ViewBox is altering the RenderTransform of its child


paulovila:
## Current behavior
When a UI element in a `ViewBox` is applied a rotation transform, the transform is overwritten by the `ViewBox`
~~When a UI element is applied a rotation transform, for example over a border, the transform moves it across the X and Y , but doesn't rotate it~~
## Expected behavior
It should rotate  the control as shown in the image:
![image](https://user-images.githubusercontent.com/888391/69341829-37c1d680-0c62-11ea-8dc8-d945eb937d97.png)
## How to reproduce it (as minimally and precisely as possible)
clone https://github.com/SuperJMN/UnoSimpleCalc/tree/Rotation
checkout the **Rotation** branch
and see the difference between UWP and WASM

## Environment

Affected platform(s):

- [?] iOS
- [?] Android
- [X] WebAssembly
- [ ] WebAssembly renderers for Xamarin.Forms
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

Visual Studio:

- [ ] 2017 (version: )
- [X] 2019 (version: )
- [ ] for Mac (version: )

dr1rrb:
When I was investigating the pointer issue with the `ViewBox`, I notice that our current implementation is [modifying the `RenderTransform` of it child](https://github.com/unoplatform/uno/blob/master/src/Uno.UI/UI/Xaml/Controls/ViewBox/Viewbox.cs#L101), which is probably not valid!

In your sample, I would assume that if you add an extra border between the [`ViewBox` and the `Border` with the rotate transform](https://github.com/SuperJMN/UnoSimpleCalc/blob/Rotation/UnoSimpleCalc/UnoSimpleCalc.Shared/MainPage.xaml#L12), the rotation applies:

![image](https://user-images.githubusercontent.com/8635919/69347852-36cd7c80-0c43-11ea-8128-eaf761828db5.png)

Note: I also had to set the `RenderTransformOrigin=".5,.5"`, but it was like UWP.

I update the bug details accordingly.

paulovila:
Also note that when you rotate a button the clicking area does not rotate
```xml
<ItemsControl.ItemTemplate>
    <DataTemplate x:DataType="viewModels:ButtonViewModelBase">
        <Button VerticalAlignment="Stretch" HorizontalAlignment="Stretch"
                Command="{Binding }" Content="{Binding Text}" >
            <Button.RenderTransform>
[...]

Issue #2266, opened 2019-12-05
[iOS/Android/WASM] Content is not correctly aligned when using <Viewbox />


agneszitte-nventive:
## Current behavior 

Try to use the `<ViewBox />` control and specify alignments for the content inside.
Resize the Application and notice the alignments are not matching the UWP version.

**UWP**
![uwp_large](https://user-images.githubusercontent.com/16295702/70256570-b0429000-1756-11ea-96d2-3aef6ac4df51.PNG)
![uwp_small](https://user-images.githubusercontent.com/16295702/70256530-9b65fc80-1756-11ea-867d-03de51967ee5.PNG)

**WASM**
![wasm_large](https://user-images.githubusercontent.com/16295702/70256521-9903a280-1756-11ea-910f-9f4d726e5009.PNG)
![wasm_small](https://user-images.githubusercontent.com/16295702/70256589-b7699e00-1756-11ea-9427-aaed20182a37.PNG)

## Expected behavior

Content alignments inside the` <Viewbox />` on iOS/Android/WASM should match the UWP version.

## How to reproduce it (as minimally and precisely as possible)

Sample repro available with this issue

## Environment

Nuget Package: Uno.UI

Package Version(s): 2.1.0-dev.18

Affected platform(s):

- [x] iOS
- [x] Android
- [x] WebAssembly
- [ ] WebAssembly renderers for Xamarin.Forms
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

Visual Studio:

- [ ] 2017 (version: )
- [x] 2019 (version: 16.3.10)
- [ ] for Mac (version: )

## Anything else we need to know?

Same thing for the Cybertruck position when resizing the application.

agneszitte-nventive:
[App1.Shared.zip](https://github.com/unoplatform/uno/files/3928158/App1.Shared.zip)

agneszitte-nventive:
@jeromelaban & @dr1rrb I'm closing this issue since I cannot reproduce it with latest Uno dev 👍

PR #2392, opened 2019-12-23
Improve ViewBox and support of pointers for transformed views

dr1rrb:
GitHub Issue: 
fixes https://github.com/unoplatform/uno/issues/2257
fixes https://github.com/unoplatform/uno/issues/2256
fixes https://github.com/unoplatform/uno/issues/2135
https://github.com/unoplatform/uno/issues/2266

## Bugfix / Feature
Improve the ViewBox to better support new pointer events

1. On Android, the over events are taking in consideration the "static transform" (a.k.a. `RenderTransform`)
2. On Android, the _ACTION_HOVER_MOVE_ and the _ACTION_HOVER_EXIT_ might not be dispatched  to the right control (so they are muted) if 2 siblings are transformed.
3. The `ViewBox` alters the `RenderTransform` of its `Child`
4. The default alignments of the `ViewBox` are not the same as UWP (centered / centered)
5. Each platform has its own implementation of `UIElement.TransformToVisual`. On iOS and Android the implementation are only applying a _translate_  of the bounds, but does not apply the full transformation.

## What is the new behavior?
1. The opposite...
2. The opposite...
3. The opposite...
4. The opposite...
5. The WASM implementation of the `UIElement.TransformToVisual` has been multi-targeted and pointer events are now using it on each platform to compute the `Location` of the pointer relative to another control.

## PR Checklist
- [x] [Wasm UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md#running-the-webassembly-ui-tests-snapshots) are not showing unexpected any differences. Validate PR `Screenshots Compare Test Run` results.


dr1rrb:
Note: This PR currently includes the changes of https://github.com/unoplatform/uno/pull/2249 which is expected to be merged before merging this one!

src/SamplesApp/UITests.Shared/Windows_UI_Xaml/UIElementTests/TransformToVisual_Transform.xaml.cs
```diff
@@ -40,7 +41,7 @@ public TransformToVisual_Transform()
 
 		private async void TransformToVisual_Transform_Loaded(object sender, RoutedEventArgs e)
 		{
-			await Task.Yield();
+			await Task.Delay(1000);
```
jeromelaban: Why ?

src/SamplesApp/UITests.Shared/Windows_UI_Xaml/UIElementTests/TransformToVisual_Transform.xaml.cs
```diff
@@ -190,32 +194,32 @@ public void When_TransformToAnotherBranch_With_InheritedCompositeTransform_And_M
...
 
-		private class RectCloseComparer : IEqualityComparer<Rect>
-		{
-			private readonly double _epsilon;
+	internal class RectCloseComparer : IEqualityComparer<Rect>
+	{
+		private readonly double _epsilon;
 
-			public static RectCloseComparer Default { get; } = new RectCloseComparer(double.Epsilon);
+		public static RectCloseComparer Default { get; } = new RectCloseComparer(Double.Epsilon);
```
jeromelaban:
```suggestion
		public static RectCloseComparer Default { get; } = new RectCloseComparer(double.Epsilon);
```

src/Uno.UI.BindingHelper.Android/Uno/UI/UnoMotionHelper.java
```diff
@@ -25,16 +25,20 @@ protected MotionTargetAdapter(Uno.UI.UnoMotionTarget target) {
...
-		if (getIsTargetCachingSupported()) {
-			_customDispatchTarget = child;
-		}
-	}
+	// The "cached target" is used by the custom dispatch logic to track the last child to which an event was sent
+	// This is required in order to make sure to follow the full events sequences: down->move->up OR enter->move->exit
+	// (For instance on a HOVER_MOVE, if the pointer goes out of the current "cached target" we still make sure to
+	// send it to this "cached target" the event in order to raise the HOVER_EXIT)
+	private View _cachedTarget;
+	final View getCachedTarget() { return _cachedTarget; }
```
jeromelaban: Cached means "current" ?

src/Uno.UI/UI/Xaml/Controls/Layouter/Layouter.cs
```diff
@@ -443,6 +443,10 @@ internal void ArrangeChild(View view, Rect frame, bool raiseLayoutUpdated)
 				return;
 			}
 			frame = ApplyMarginAndAlignments(view, frame);
+			if (view is UIElement elt)
+			{
+				elt.LayoutSlotWithMarginsAndAlignments = frame;
```
jeromelaban: Why ?

src/SamplesApp/UITests.Shared/Windows_UI_Xaml/UIElementTests/TransformToVisual_Transform.xaml.cs
dr1rrb: To make sure the window is fully resized (especially in automated test on Android, where the StatusBar is animated when closed by the test engine. I could however probably reduce the delay, but FYI this test class is loaded only once.

src/Uno.UI.BindingHelper.Android/Uno/UI/UnoMotionHelper.java
dr1rrb: Kind of, as this "current" is relative to a given sequence (down / move / up **or** hover enter / move / exit) and it might be cleared depending of various factor (for instance when moving pointer over a sibling). I preferred the name "cached" as it really just to avoid re-calculation in that case ... but meh 🤷‍♂️ ...

src/Uno.UI/UI/Xaml/Controls/Layouter/Layouter.cs
dr1rrb: I use the `LayoutSlotWithMarginsAndAlignments` in the `TransformToVisual` to get the location of the element in the parent coordinates system. The issue is that the `LayoutSlot` from UWP does not contains the margins nor alignments applied by the parent element. I tried to reverse-compute it, but it's implies to down cast element (for instance for the margins I had to convert down cast to `FrameworkElement`, the scroll to `ScrollViewer` etc.), it's less performant and much more error prompt ... I found that the easiest / safest solution was to store the "native layout slot" (a.k.a. `LayoutSlotWithMarginsAndAlignments`) in the arrange.

src/Uno.UI.BindingHelper.Android/Uno/UI/UnoMotionHelper.java
dr1rrb: I will rename it in my next PR (multi touch) as I also change this file.

src/SamplesApp/UITests.Shared/Windows_UI_Xaml/UIElementTests/TransformToVisual_Transform.xaml.cs
dr1rrb: Reduced delay and added comment to explain why

src/Uno.UI.BindingHelper.Android/Uno/UI/UnoMotionHelper.java
```diff
@@ -133,14 +137,15 @@ private boolean dispatchMotionEventCore(Uno.UI.MotionTargetAdapter adapter, Moti
 		_currentMotionOriginalSource = null;
 
 		final boolean isDown = event.getAction() == MotionEvent.ACTION_DOWN;
-		if (isDown) {
+		final boolean isBeginOfSequence = isDown || event.getAction() == MotionEvent.ACTION_HOVER_ENTER;
```
davidjohnoliver:
```suggestion
		final boolean isBeginningOfSequence = isDown || event.getAction() == MotionEvent.ACTION_HOVER_ENTER;
```

src/Uno.UI.BindingHelper.Android/Uno/UI/UnoMotionHelper.java
```diff
@@ -133,14 +137,15 @@ private boolean dispatchMotionEventCore(Uno.UI.MotionTargetAdapter adapter, Moti
...
+		if (isBeginOfSequence) {
```
davidjohnoliver:
```suggestion
		if (isBeginningOfSequence) {
```

src/Uno.UI.BindingHelper.Android/Uno/UI/UnoMotionHelper.java
```diff
@@ -159,12 +164,12 @@ private boolean dispatchMotionEventCore(Uno.UI.MotionTargetAdapter adapter, Moti
 			return false;
 		}
 
-		if (isDown && !isMotionInView(event, view)) {
+		if (isBeginOfSequence && !isMotionInView(event, view)) {
```
davidjohnoliver:
```suggestion
		if (isBeginningOfSequence && !isMotionInView(event, view)) {
```

src/Uno.UI.BindingHelper.Android/Uno/UI/UnoMotionHelper.java
```diff
@@ -159,12 +164,12 @@ private boolean dispatchMotionEventCore(Uno.UI.MotionTargetAdapter adapter, Moti
...
 			// When using the "super" dispatch path, it's possible that for visual constraints (i.e. clipping),
 			// the view must not handle the touch. If that's the case, the touch event must be dispatched
 			// to other controls.
-			// Note: We do this check only when starting a new manipulation (isDown), so the whole down->move->up sequence is ignored;
-			// Warning: We should also do this check for events that does not have strong sequence concept (e.g. Hover)
+			// Note: We do this check only when starting a new manipulation (isBeginOfSequence), so the whole down->move->up sequence is ignored;
```
davidjohnoliver:
```suggestion
			// Note: We do this check only when starting a new manipulation (isBeginningOfSequence), so the whole down->move->up sequence is ignored;
```

src/SamplesApp/UITests.Shared/Helpers/RectCloseComparer.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.Linq;
+using Windows.Foundation;
+
+namespace Uno.UI.Samples.UITests.Helpers
+{
+	internal class RectCloseComparer : IEqualityComparer<Rect>
+	{
+		private readonly double _epsilon;
+
+		public static RectCloseComparer Default { get; } = new RectCloseComparer(double.Epsilon);
```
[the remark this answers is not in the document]
dr1rrb: Yeah ... I learned it the hard way since a while ;)

src/Uno.UI.BindingHelper.Android/Uno/UI/UnoMotionHelper.java
dr1rrb: Yes makes sense! 

But like the comment from Jérôme, I'll fix it in my next PR as I already did some changes in that file (and especially I changed this variable usage).

dr1rrb:
This is also expected to fix https://stackoverflow.com/questions/59027861/content-not-centered-in-viewbox-and-incorect-mouse-coord-in-viewbox/59180616#59180616

PR #2461, opened 2020-01-10
[Android] Always clip non-UIElements

davidjohnoliver:
Native Android views typically expect to be clipped, and display incorrectly otherwise. Ensure they're always clipped.

This allows native controls defined outside the framework to display correctly without additional adjustments.

## PR Type

- Bugfix

## PR Checklist

## Other information

Internal Issue (If applicable): https://github.com/unoplatform/private/issues/117

### DIFF
--- a/src/Uno.UI/UI/Xaml/UIElement.Android.cs
+++ b/src/Uno.UI/UI/Xaml/UIElement.Android.cs
@@ -56,7 +56,7 @@ namespace Windows.UI.Xaml
 			// This won't work when UIElements and non-UIElements are mixed in the same Panel,
 			// but it should cover most cases in practice, and anyway should be superceded when
 			// IFrameworkElement will be removed.
-			SetClipChildren(AreChildrenNativeViewsOnly);
+			SetClipChildren(FeatureConfiguration.UIElement.AlwaysClipNativeChildren ? AreChildrenNativeViewsOnly : false);
 
 			if (rect.IsEmpty)
 			{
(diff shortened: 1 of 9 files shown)


src/Uno.UI/UI/Xaml/UIElement.Android.cs
```diff
@@ -26,6 +52,12 @@ public UIElement()
 
 		partial void ApplyNativeClip(Rect rect)
 		{
+			// Non-UIElements typically expect to be clipped, and display incorrectly otherwise
+			// This won't work when UIElements and non-UIElements are mixed in the same Panel,
+			// but it should cover most cases in practice, and anyway should be superceded when
+			// IFrameworkElement will be removed.
+			SetClipChildren(FeatureConfiguration.UIElement.AlwaysClipNativeChildren ? AreChildrenNativeViewsOnly : false);
```
jeromelaban: Could it be a perf issue ?

davidjohnoliver: In what sense - changing the property more often?

Looking at the Android code, it will essentially be a no-op if the `ClipChildren` value doesn't change (which will be the usual case).

jeromelaban: Calling `AreChildrenNativeViewsOnly` enumerates the children, so this may be a perf issue, I'm wondering if it could be cached somehow, or computed on children changes.

davidjohnoliver: Ah, I see.

Yes, that value can be cached if it's a perf concern. Most of the time the first element will be a `UIElement` and it'll return `false` straight away, but maybe even getting the enumerator is a perf issue?

jeromelaban: That's fair, the common case won't be impacted much indeed. Let's keep it that way then.

Issue #2465, opened 2020-01-13
Button.Click is not fired on macOS


MartinZikmund:
## Current behavior

On macOS, the `Click` event of a button is never raised.

## Expected behavior

Click event should be raised.

## How to reproduce it (as minimally and precisely as possible)

Create a button on a page, set its `Click` event, run and click.

## Environment

Nuget Package:

Package Version(s):

Affected platform(s):

- [ ] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] WebAssembly renderers for Xamarin.Forms
- [x] macOS
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

Visual Studio:

- [ ] 2017 (version: )
- [ ] 2019 (version: )
- [x] for Mac (version: )

MartinZikmund:
Done

PR #2466, opened 2020-01-13
[macOS] Button.Click support, PointerEntered, PointerExited, PointerMoved

MartinZikmund:
GitHub Issue (If applicable): #2465 

## PR Type

- Bugfix
- Feature

- Button.Click is not raised on macOS
- PointerEntered, PointerExited and PointerMoved events are not supported on macOS

## What is the new behavior?

- Click is raised
- Pointer events supported

## PR Checklist


src/Uno.UI/UI/Xaml/UIElement.macOS.cs
```diff
@@ -254,17 +272,112 @@ public override void MouseDown(NSEvent evt)
...
-					pointerEventIsHandledInManaged = RaiseEvent(PointerPressedEvent, args) || pointerEventIsHandledInManaged;
+				if (evt.IsTouchInView(this))
+				{
+					IsPointerPressed = true;
+					IsPointerOver = true;
+
+					// evt.AllTouches raises a invalid selector exception
+					var args = new PointerRoutedEventArgs(null, evt)
+					{
+						CanBubbleNatively = true,
```
dr1rrb: You can remove this as it set by the ctor of `PointerRoutedEventArgs` (same for all `MouseXXX` overrides :) )

src/Uno.UI/UI/Xaml/UIElement.macOS.cs
```diff
@@ -254,17 +272,112 @@ public override void MouseDown(NSEvent evt)
...
+						OriginalSource = this
```
dr1rrb: I would prefer to give this in the ctor of the `PointerRoutedEventArgs` named `receiver` to be closer to other platforms (same for all `MouseXXX` overrides :) )

src/Uno.UI/UI/Xaml/UIElement.macOS.cs
```diff
@@ -254,17 +272,112 @@ public override void MouseDown(NSEvent evt)
 						OriginalSource = this
 					};
 
-					pointerEventIsHandledInManaged = RaiseEvent(PointerEnteredEvent, args);
+					pointerEventIsHandledInManaged = RaisePointerEvent(PointerPressedEvent, args);
```
dr1rrb: This PR is overall a great improvement of pointer handling on Mac OS, but I think that we could go a bit further by using the shared code of pointer events that is used on all other platforms. Doing so we will be able to get all gestures (especially the `Tapped` event) and manipulations events.

I'm not able to test it right now, but I think it should works pretty easily with your other changes 😃.

To do that, you should only have to remove all `IsPointerPressed = true|false;`, `IsPointerOver = true|false;` and here instead of directly raising the event, you should invoke [`OnNativePointerEnter(args)`](https://github.com/unoplatform/uno/blob/master/src/Uno.UI/UI/Xaml/UIElement.Pointers.cs#L399) and equivalent for other `MouseXXX` overrides.

src/Uno.UI/UI/Xaml/UIElement.macOS.cs
```diff
@@ -292,23 +405,23 @@ public override void MouseUp(NSEvent evt)
 
 				if (!wasPointerOver && IsPointerOver)
 				{
-					pointerEventIsHandledInManaged = RaiseEvent(PointerEnteredEvent, args);
+					pointerEventIsHandledInManaged = RaisePointerEvent(PointerEnteredEvent, args);
```
dr1rrb: In the hypothesis that you use the [`OnNativePointerUp`](https://github.com/unoplatform/uno/blob/master/src/Uno.UI/UI/Xaml/UIElement.Pointers.cs#L503), you can safely remove all the `<was|is>PointerOver` checks.

src/Uno.UI/UI/Xaml/UIElement.macOS.cs
```diff
@@ -41,9 +43,25 @@ public UIElement()
 			InitializePointers();
 		}
 
-
+		public override void UpdateTrackingAreas()
```
dr1rrb: What's the purpose of this ? Would be great to have a small comment to explain this :)

MartinZikmund: Will add it :-) . It is required to get the entered/exited/moved events to fire.

src/Uno.UI/UI/Xaml/Input/PointerRoutedEventArgs.macOS.cs
```diff
@@ -23,22 +22,83 @@ internal PointerRoutedEventArgs(NSSet touches, NSEvent nativeEvent) : this()
 			_nativeEvent = nativeEvent;
 			_nativeTouches = touches;
 
-			FrameId = _pseudoFrameId;
-			Pointer = new Pointer(nativeEvent);
+			var pointerId = (uint)0;
```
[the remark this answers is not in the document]
MartinZikmund: Unfortunately `Handle` does not work, it changes betweent pressed/released :-( .

src/Uno.UI/UI/Xaml/Input/PointerRoutedEventArgs.macOS.cs
```diff
@@ -23,22 +22,83 @@ internal PointerRoutedEventArgs(NSSet touches, NSEvent nativeEvent) : this()
...
+			var rawPosition = _nativeEvent.LocationInWindow;
+			var position = relativeTo != null ?
+				relativeTo.ConvertPointFromView(_nativeEvent.LocationInWindow, null) :
+				rawPosition;
+			
+			var properties = new PointerPointProperties()
+			{
+				IsInRange = true,
+				IsPrimary = true,
+				IsLeftButtonPressed = ((int)NSEvent.CurrentPressedMouseButtons & LeftMouseButtonMask) == LeftMouseButtonMask,
```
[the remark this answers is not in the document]
MartinZikmund: @dr1rrb Regarding the touch input, I think there is no support for touch in macOS at all, so it would be only guesswork currently. Maybe I am just not aware of something though :-D .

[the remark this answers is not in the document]
dr1rrb: I would have agree, until I noticed that there is also [a `TouchesBegan` on macOS ](https://developer.apple.com/documentation/appkit/nsresponder/1531151-touchesbegan)

Maybe for the touch pad ?

Anyway, we as didn't override those methods, I would assume that we will get only mouse pointers, so you can disregard my comments about pen and fingers (maybe just add a comment to remind us that we should probably have to updates those code once we have overridden all the input methods)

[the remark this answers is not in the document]
MartinZikmund: Hmm, one thing which could it work with is when you use iPad as second screen. I will check that 🤔

src/Uno.UI/UI/Xaml/UIElement.macOS.cs
MartinZikmund: I shall try it tomorrow :-)

dr1rrb: Great 😃 Not sure if you saw that I did a commit that should helps you to understand what I mean https://github.com/dr1rrb/Uno/commit/2c426732893cbb035c9cda3f193313d43b4fd7be , I only forgot to remove at line 138:
```cs
IsPointerPressed = false;
IsPointerOver = false;
```

MartinZikmund: Definitely keeping it in mind 👍👍

src/Uno.UI/UI/Xaml/UIElement.macOS.cs
MartinZikmund: Fixed :-)

src/Uno.UI/UI/Xaml/UIElement.macOS.cs
MartinZikmund: Done

src/Uno.UI/UI/Xaml/UIElement.macOS.cs
MartinZikmund: Commented :-)

src/Uno.UI/UI/Xaml/Input/PointerRoutedEventArgs.macOS.cs
```diff
@@ -23,22 +22,83 @@ internal PointerRoutedEventArgs(NSSet touches, NSEvent nativeEvent) : this()
...
-			Pointer = new Pointer(nativeEvent);
+			var pointerId = (uint)0;
+			var pointerDeviceType = GetPointerDeviceType(nativeEvent.Type);
+
+			var isInContact = GetIsInContact(nativeEvent);
+
+			FrameId = ToFrameId(_nativeEvent.Timestamp);
+			Pointer = new Pointer(pointerId, pointerDeviceType, isInContact, isInRange: true);
+			KeyModifiers = System.VirtualKeyModifiers.None; //TODO: Properly set virtual key modifiers
+			OriginalSource = null; //TODO: get original source
```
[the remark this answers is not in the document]
MartinZikmund: Replaced

src/Uno.UI/UI/Xaml/UIElement.macOS.cs
MartinZikmund: Done

src/Uno.UI/UI/Xaml/UIElement.macOS.cs
MartinZikmund: Done :-)

src/Uno.UI/UI/Xaml/Input/PointerRoutedEventArgs.macOS.cs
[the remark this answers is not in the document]
MartinZikmund: Ok, so I managed to improve support for pen with the iPad Sidecar for some of the events to return pressure and tilt, however it seems like it is all in a "war state" - for example pen events are in fact reported as mouse events, for some the NSEvent.Subtype, which should work for mouse events throws for MouseEntered and MouseExited (so it is impossible to detect whether the mouse event is in fact a pen event) and so on. Finally, the Sidecar does not support touch input, so I was unable to implement that, yet. However, the mouse events are now working nicely, button click and button pressed state too, after applying your suggestions :-)

src/Uno.UI/UI/Xaml/Input/PointerRoutedEventArgs.macOS.cs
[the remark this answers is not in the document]
MartinZikmund: I am using GetPointerDeviceId() for Pen events, but it returns 0 on my device anyway. Oh well :-D ...

src/Uno.UI/UI/Xaml/Input/PointerRoutedEventArgs.macOS.cs
```diff
@@ -23,22 +22,83 @@ internal PointerRoutedEventArgs(NSSet touches, NSEvent nativeEvent) : this()
...
+				rawPosition,
+				position,
+				Pointer.IsInContact,
+				properties);
 		}
 
 		#region Misc static helpers
+		private static long? _bootTime;
+
+		private static bool GetIsInContact(NSEvent nativeEvent)
```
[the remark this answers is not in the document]
MartinZikmund: Pen is handled, as it is reported as a "mouse event" with pen subtype. I am unable to replicate touch events...

src/Uno.UI/UI/Xaml/Input/PointerRoutedEventArgs.macOS.cs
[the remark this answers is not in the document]
dr1rrb: At some point I'm happy to not be the only one which had to deal with this nightmare that is unifying the mouse/pen/touch handling on all platforms ... but trust me, I understand your feeling right now ;)

src/Uno.UI/UI/Xaml/UIElement.Pointers.macOS.cs
```diff
@@ -0,0 +1,192 @@
...
+				IsPointerOver = evt.IsTouchInView(this);
+
+				var pointerEventIsHandledInManaged = false;
+
+				// evt.AllTouches raises a invalid selector exception
+				var args = new PointerRoutedEventArgs(null, evt, this);
+
+				if (IsPointerCaptured || IsPointerOver)
+				{
+					pointerEventIsHandledInManaged = OnNativePointerMove(args);
```
dr1rrb: Here instead of validating the `IsPointerCaptured` and `IsPointerOver` you should be able to use the `OnNativePointerMoveWithOverCheck(args, evt.IsTouchInView(this))` which already does that. 

BTW here we should update the over state (i.e. enter / exit) before raising the move since the coordinates of the `args` are already reflecting this updated state.

MartinZikmund: Will do :-)

MartinZikmund: Is the current version what you meant? The `OnNativePointerMoveWithOverCheck` triggers the Enter/Exit events automatically in case it is necessary, so it is ok to remove the other ones?

dr1rrb: Yes it does, that the purpose of the [`SetOver`](https://github.com/unoplatform/uno/blob/master/src/Uno.UI/UI/Xaml/UIElement.Pointers.cs#L663) 😀

MartinZikmund: Thanks 👍👍

PR #2449, opened 2020-01-07
Support of StorageFile.DateCreated

pkar70:
GitHub Issue (If applicable): #1187 (partly)

## PR Type

What kind of change does this PR introduce?
- Feature

## What is the current behavior?
StorageFile.DateCreated is not supported.

## What is the new behavior?
StorageFile.DateCreated is supported.

## PR Checklist

Please check if your PR fulfills the following requirements:
- [X] Tested code with current [supported SDKs](https://github.com/unoplatform/uno/blob/master/README.md#uno-features)
- [ ] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)
- [X] [Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [ ] [Wasm UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md#running-the-webassembly-ui-tests-snapshots) are not showing unexpected any differences. Validate PR `Screenshots Compare Test Run` results.
- [X] Contains **NO** breaking changes
- [X] Updated the [Release Notes](https://github.com/unoplatform/uno/tree/master/doc/ReleaseNotes)
- [X] Associated with an issue (GitHub or internal)

## Other information

Internal Issue (If applicable):

PR #2407, opened 2019-12-28
File.delete async

pkar70:
GitHub Issue (If applicable): #1187 (partly)

## PR Type

What kind of change does this PR introduce?
- Feature

## What is the current behavior?
StorageFile:DeleteAsync not supported (workaround exist)

## What is the new behavior?
StorageFile:DeleteAsync is supported.

## PR Checklist

Please check if your PR fulfills the following requirements:

- [X] Tested code with current [supported SDKs](https://github.com/unoplatform/uno/blob/master/README.md#uno-features)
- [ ] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)
- [X] [Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [ ] [Wasm UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md#running-the-webassembly-ui-tests-snapshots) are not showing unexpected any differences. Validate PR `Screenshots Compare Test Run` results.
- [X] Contains **NO** breaking changes
- [X] Updated the [Release Notes](https://github.com/unoplatform/uno/tree/master/doc/ReleaseNotes)
- [X] Associated with an issue (GitHub or internal)

## Other information

Internal Issue (If applicable):

PR #2484, opened 2020-01-15
Windows.Storage.FileIO.(Append|Write|Read)TextAsync implementation

pkar70:
GitHub Issue (If applicable): #1187 (partly)

## PR Type

- Feature

Windows.Storage.FileIO.(Append|Write|Read)TextAsync methods are not implemented.

## What is the new behavior?
Windows.Storage.FileIO.(Append|Write|Read)TextAsync methods are implemented.

## PR Checklist

## Other information


src/Uno.UWP/Storage/FileIO.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace Windows.Storage
+{
+	public partial class FileIO
+
+		public static global::Windows.Foundation.IAsyncAction WriteTextAsync( global::Windows.Storage.IStorageFile file,  string contents)
+		{
+        Stream fileStream = await file.OpenStreamForWriteAsync();
+        StreamWriter streamWriter = new StreamWriter(fileStream);
+        streamWriter.Write(contents);
```
MartinZikmund: Could the code use asynchronous version of the API instead here? From docs - https://docs.microsoft.com/en-us/dotnet/api/system.io.streamwriter.writeasync?view=netframework-4.8

src/Uno.UWP/Storage/FileIO.cs
```diff
@@ -0,0 +1,32 @@
...
+        streamWriter.Dispose();
+        fileStream.Dispose();
+		}
+
+		public static global::Windows.Foundation.IAsyncAction AppendTextAsync( global::Windows.Storage.IStorageFile file,  string contents)
+		{
+        Stream fileStream = await file.OpenStreamForWriteAsync();
+        fileStream.Seek(0, SeekOrigin.End);
+        StreamWriter streamWriter = new StreamWriter(fileStream);
+        streamWriter.Write(contents);
```
MartinZikmund: Same as in previous comment - can we use `WriteAsync` instead?

src/Uno.UWP/Storage/FileIO.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+	public partial class FileIO
+
+		public static global::Windows.Foundation.IAsyncAction WriteTextAsync( global::Windows.Storage.IStorageFile file,  string contents)
+		{
+        Stream fileStream = await file.OpenStreamForWriteAsync();
+        StreamWriter streamWriter = new StreamWriter(fileStream);
+        streamWriter.Write(contents); 
+        streamWriter.Dispose();
+        fileStream.Dispose();
```
MartinZikmund: I would suggest using `using` blocks for the `StreamWriter` and `FileStream` instead. In case an exception happens, in the current implementation the file could stay in open state.

src/Uno.UWP/Storage/FileIO.cs
```diff
@@ -0,0 +1,32 @@
...
+		}
+
+		public static global::Windows.Foundation.IAsyncAction AppendTextAsync( global::Windows.Storage.IStorageFile file,  string contents)
+		{
+        Stream fileStream = await file.OpenStreamForWriteAsync();
+        fileStream.Seek(0, SeekOrigin.End);
+        StreamWriter streamWriter = new StreamWriter(fileStream);
+        streamWriter.Write(contents); 
+        streamWriter.Dispose();
+        fileStream.Dispose();
```
MartinZikmund: Also please use `using` blocks instead.

src/Uno.UWP/Storage/FileIO.cs
pkar70: Yes, but why "await  WriteAsync" should be better than "Write"?

src/Uno.UWP/Storage/FileIO.cs
pkar70: I don't know architecture of tests. E.g. I don't know if every test has own filesystem. Or, can file from one test be seen by other tests?

davidjohnoliver: You should assume that a test _might_ see files from other tests (you can't know for sure because you don't know which order tests will run in). So you should clean up files after each test (probably in a tear-down method).

You should probably be writing runtime unit tests for this: https://platform.uno/docs/articles/contributing/guidelines/creating-tests.html#platform-runtime-unit-tests-unouiruntimetests

src/Uno.UWP/Storage/FileIO.cs
MartinZikmund: It is asynchronous, so it frees up the thread to do other work instead. Async versions of IO operations are always preferable. In this case especially as if you call this method as it is currently from the UI thread, the first `await` doesn't have a `ConfigureAwait(false)` which means it will return back to the UI thread and then will perform the whole write synchronously (which could take several seconds for large files) and that will block the UI completely.

src/Uno.UWP/Storage/FileIO.cs
MartinZikmund: Ideally you should perform cleanup after the test is finished - so wrapping the unit test in a `try...finally` and cleaning up in finally. Each build has its own separate file store, but still, better safe than sorry :-D . As a sample, you can check [this unit test](https://github.com/unoplatform/uno/blob/024eebddd33ac0dfa6b6d8ea0871d5c6effc9f12/src/Uno.UI.Tests/Windows_UI_Xaml/Given_ThemeResource.cs#L31 ) for resources - I am using finally there to reset back to defaults. Tests for this API would probably fit best in the `Uno.UI.RuntimeTests` library.

pkar70:
But test should be 'one level down', in StreamWriter implementation...

MartinZikmund:
> But test should be 'one level down', in StreamWriter implementation...

Sure, `StreamWriter` is tested in .NET Core, but in this case we would probably like to have the method itself tested as well. For example, what happens when the passed in `StorageFile` is `null`? The method should probably throw `ArgumentNullException` to guard against it, if that is the same behavior as in UWP, and so on :-) . The test should make sure, that from the outside, the method behaves correctly, without knowing the actual implementation uses `StreamWriter` specifically. This way, when the method implementation potentially changes in the future, we will know, that the "new" implementation behaves the same way as the original.

pkar70:
"what happens when the passed in StorageFile is null" - good point. Documentation doesn't specify this, so Uno could do anything we want? Or, should it be same as under Windows?

MartinZikmund:
> "what happens when the passed in StorageFile is null" - good point. Documentation doesn't specify this, so Uno could do anything we want? Or, should it be same as under Windows?

Ideally check what happens when you do this on UWP. I think it will throw `ArgumentNullException` as well.Then just ensure the implementation in Uno throws the exception as well and the unit test can then do `await Assert.ThrowsExceptionAsync<ArgumentNullException>(...)`

doc/ReleaseNotes/_ReleaseNotes.md
```diff
@@ -1,9 +1,7 @@
 ﻿# Release notes
 
 ### Features
-- Support for `EmailManager.ShowComposeNewEmailAsync`
-- Add support for `StorageFolder.CreateFileAsync(string path)`
-- Add support for ApplicationViewTitleBar.BackgroundColor on WASM
+- Add support of Windows.Storage.FileIO.AppendTextAsync(IStorageFile, String) and WriteTextAsync(IStorageFile, String)
```
jeromelaban: Don't remove release notes

src/Uno.UWP/Storage/FileIO.cs
```diff
@@ -0,0 +1,91 @@
...
+using Windows.Storage.Streams;
+
+namespace Windows.Storage
+{
+	public partial class FileIO
+	{
+
+		private static async Task AppendWriteTextAsync(global::Windows.Storage.IStorageFile file, string contents, global::Windows.Storage.Streams.UnicodeEncoding encoding, bool append)
+		{
+			if (file is null)
```
jeromelaban:
```suggestion
			`file = file ?? throw new NullReferenceException("StorageFile cannot be null");`
```

src/Uno.UWP/Storage/FileIO.cs
```diff
@@ -0,0 +1,91 @@
...
+			{
+				throw new NullReferenceException("StorageFile cannot be null");
+			}
+
+			global::System.Text.Encoding encodingForWriter = global::System.Text.Encoding.UTF8;
```
jeromelaban: Please use namespaces to avoid fully qualified names in non-generated files.

src/Uno.UWP/Storage/FileIO.cs
```diff
@@ -0,0 +1,91 @@
...
+					break;
+				case global::Windows.Storage.Streams.UnicodeEncoding.Utf16LE:
+					encodingForWriter = global::System.Text.Encoding.Unicode;
+					break;
+				case global::Windows.Storage.Streams.UnicodeEncoding.Utf16BE:
+					encodingForWriter = global::System.Text.Encoding.BigEndianUnicode;
+					break;
+			}
+
+			using (Stream fileStream = await file.OpenStreamForWriteAsync())
```
jeromelaban: [`File.Open`](https://docs.microsoft.com/en-us/dotnet/api/system.io.file.open?view=netframework-4.8#System_IO_File_Open_System_String_System_IO_FileMode_System_IO_FileAccess_) has an append mode, no need to seek, same comment for the other method below.

doc/ReleaseNotes/_ReleaseNotes.md
pkar70: Seems like my error in merging...

src/Uno.UWP/Storage/FileIO.cs
pkar70: is this more readable?

src/Uno.UWP/Storage/FileIO.cs
pkar70: Done. But I left one level up in Encoding, as we have here two different encodings.

src/Uno.UWP/Storage/FileIO.cs
pkar70: OpenStreamForWriteAsync has no mode (append/rewrite) parameter, also StreamWriter with Stream parameter has no such parametr. Or, at least, I cannot find it.

src/Uno.UWP/Storage/FileIO.cs
MartinZikmund: I would suggest throwing a `ArgumentNullException` here instead, as it is clearer to the user, that it is not a bug, but in fact a usage error.

pkar70: But UWP sends NullReferenceException, at least on my PC :)

MartinZikmund: Oh, in that case their mistake 😀 . Thanks for clarification 👍

jeromelaban: My apologies I missed that one, we should never throw a NullReferenceException. This is most probably an invalid validation on the WinRT side, let's throw a ArgumentNullException instead. Thanks @MartinZikmund .

src/Uno.UWP/Storage/FileIO.cs
jeromelaban: You can [create a `StreamWriter` from a `File`](https://docs.microsoft.com/en-us/dotnet/api/system.io.streamwriter.-ctor?view=netframework-4.8#System_IO_StreamWriter__ctor_System_IO_Stream_System_Text_Encoding_).

src/Uno.UWP/Storage/Streams/UnicodeEncoding.cs
```diff
@@ -0,0 +1,9 @@
+namespace Windows.Storage.Streams
+{
+	public   enum UnicodeEncoding
```
jeromelaban: Please adjust the formatting (multiple spaces).

src/Uno.UWP/Storage/FileIO.cs
pkar70: I checked and used same exception as UWP - to be even "bug compatible".
When someone wrote some app, with try/catch and check which exception is thrown, he would expect and handle NullReferenceException, not ArgumentNullException. When Uno would throw another exception, catch{} would be something like `if( (Uno && ArgumentNullException ) || (!Uno && NullReferenceException)) { handleBadFile() }` ...
Is this what we want?

src/Uno.UWP/Storage/FileIO.cs
pkar70: But I have Windows.Storage.StorageFile , not File ?

src/Uno.UWP/Storage/FileIO.cs
jeromelaban: In either case, it's a bad practice to handle such errors that way. Throwing a `NullReferenceException` is misleading because it's not an actual NRE. Let's throw an `ArgumentNullException` as it will show in the debugger more interesting details.

src/Uno.UWP/Storage/FileIO.cs
```diff
@@ -0,0 +1,204 @@
...
+			Streams.UnicodeEncoding unicodeEncoding = Streams.UnicodeEncoding.Utf8;
+
+			using (Stream fileReadStream = await file.OpenStreamForReadAsync())
+			{
+				using (BinaryReader binReader = new BinaryReader(fileReadStream))
+				{
+					try
+					{
+						byte firstByte = binReader.ReadByte();
+						switch (firstByte)
```
carldebilly: Codacy Issue found: [Replace this 'switch' statement with 'if' statements to increase readability.](https://app.codacy.com/gh/unoplatform/uno/pullRequest?prid=4827590)

src/Uno.UWP/Storage/FileIO.cs
```diff
@@ -0,0 +1,204 @@
...
+		{
+			if (file is null)
+			{
+				// UWP throws NullReferenceException
+				// but it was choosen to throw ArgumentNullException , as it is more appriopriate in this caase
+				throw new ArgumentNullException("StorageFile cannot be null");
+			}
+
+			string output = "";
+			using (Stream fileStream = await file.OpenStreamForReadAsync())
```
carldebilly: Codacy Issue found: [Refactor this code to make sure 'fileStream' is disposed only once.](https://app.codacy.com/gh/unoplatform/uno/pullRequest?prid=4827590)

src/Uno.UI.RuntimeTests/Tests/Windows_Storage/Given_FileIO.cs
```diff
@@ -0,0 +1,56 @@
...
+		{
+		}
+
+		[TestCleanup]
+		public void Cleanup()
+		{
+		}
+
+		[TestMethod]
+		public async void When_WriteReadAppendRead()
```
carldebilly: Codacy Issue found: [Return 'Task' instead.](https://app.codacy.com/gh/unoplatform/uno/pullRequest?prid=4827590)

src/Uno.UWP/Storage/FileIO.cs
```diff
@@ -0,0 +1,204 @@
...
+		public static IAsyncAction AppendTextAsync(IStorageFile file, string contents, Streams.UnicodeEncoding encoding) => AppendWriteTextAsync(file, contents, encoding, true).AsAsyncAction();
+
+		public static IAsyncAction WriteTextAsync(IStorageFile file, string contents, Streams.UnicodeEncoding encoding) => AppendWriteTextAsync(file, contents, encoding, false).AsAsyncAction();
+
+		private static string ConvertLinesToString(IEnumerable<string> lines)
+		{
+			string output = "";
+			foreach (string line in lines)
+			{
+				output = output + line + Environment.NewLine;
```
carldebilly: Codacy Issue found: [Use a StringBuilder instead.](https://app.codacy.com/gh/unoplatform/uno/pullRequest?prid=4827590)

src/Uno.UI.RuntimeTests/Tests/Windows_Storage/Given_FileIO.cs
carldebilly: Codacy is totally right here: using `void` will prevent the test engine from capturing any error in the test.

src/Uno.UWP/Storage/FileIO.cs
pkar70: Can it be left as it is?

src/Uno.UWP/Storage/FileIO.cs
pkar70: Can it be left as it is?

jeromelaban: No, this rule is correct, a stringbuilder is more appropriate here.

carldebilly: I think you should not try to create a big string with all the content of the file in it... instead you should write each line separately to the file.

This is a technique called "streaming" which require less memory for the same operation.

src/Uno.UI.RuntimeTests/Tests/Windows_Storage/Given_FileIO.cs
pkar70: So, #2407 and #2449 have invalid tests? :(
Can you add `Task` as return value from TestMethod in these two tests, or should I create PRs for this?

src/Uno.UWP/Storage/FileIO.cs
pkar70: Ok, but what is preferred? Time or memory? My version takes less CPU (less switches between user space and system space), and could be also much better from disk usage aspect (one write to file and file system tables, one transaction in file system transaction log, etc.) versus many writes. Depending of disk cache usage policy, writing whole file in one step could vastly extend SSD disk life. Also one write is better on encrypted file systems.

src/Uno.UWP/Storage/FileIO.cs
```diff
@@ -0,0 +1,213 @@
...
+		{
+			if (file is null)
+			{
+				// UWP throws NullReferenceException
+				// but it was choosen to throw ArgumentNullException , as it is more appriopriate in this caase
+				throw new ArgumentNullException("StorageFile cannot be null");
+			}
+
+			string output = "";
+			using (Stream fileStream = await file.OpenStreamForReadAsync())
```
carldebilly: Codacy Issue found: [Refactor this code to make sure 'fileStream' is disposed only once.](https://app.codacy.com/gh/unoplatform/uno/pullRequest?prid=4827590)

pkar70: And if I change code as Codacy wants, then it signal that `stream.Dispose();` should be placed inside `{}` :)
Ok, changing.

src/Uno.UWP/Storage/FileIO.cs
```diff
@@ -0,0 +1,222 @@
...
+				using (StreamReader streamReader = new StreamReader(fileStream))
+				{
+					output = streamReader.ReadToEnd();
+				}
+			}
+			finally
+			{
+				if(fileStream != null)
+                {
+					fileStream.Dispose();
```
carldebilly: Codacy Issue found: [Refactor this code to make sure 'fileStream' is disposed only once.](https://app.codacy.com/gh/unoplatform/uno/pullRequest?prid=4827590)

pkar70: But this is already implemented... conversion from using/using to try/using/finally.

pkar70:
Strange.
When I compile it, on my desktop, `return output.Split(Environment.NewLine);` works - as `string [] string.Split(string separator, [StringSplitOptions options = StringSplitOptions.None])`.
But when compiled in CI, it fails: `Argument 1: cannot convert from 'string' to 'char'`.
So, I'm trying to make it:

			var separators = new String[] { Environment.NewLine };
			return output.Split(separators, StringSplitOptions.None);

src/Uno.UWP/Storage/FileIO.cs
```diff
@@ -0,0 +1,224 @@
...
```
carldebilly: Codacy Issue found: [Refactor this code to make sure 'fileStream' is disposed only once.](https://app.codacy.com/gh/unoplatform/uno/pullRequest?prid=4827590)

pkar70: This is already done, changed from using/using to try/using/finally. I don't understand why codacy still says that it is error.

doc/ReleaseNotes/_ReleaseNotes.md
```diff
@@ -1,7 +1,7 @@
 # Release notes
 
 ### Features
-
+- Add support of Windows.Storage.FileIO.AppendTextAsync(IStorageFile, String) and WriteTextAsync(IStorageFile, String)
```
```suggestion
- Add support of `Windows.Storage.FileIO.AppendTextAsync(IStorageFile, String)` and `WriteTextAsync(IStorageFile, String)`
```

src/Uno.UWP/Storage/FileIO.cs
```diff
@@ -0,0 +1,224 @@
...
+{
+	public partial class FileIO
+	{
+
+		private static async Task AppendWriteTextAsync(IStorageFile file, string contents, Streams.UnicodeEncoding encoding, bool append)
+		{
+			if (file is null)
+			{
+				// UWP throws NullReferenceException
+				// but it was choosen to throw ArgumentNullException , as it is more appriopriate in this caase
```
jeromelaban:
```suggestion
				// but it was chosen to throw ArgumentNullException , as it is more appropriate in this case

```

src/Uno.UWP/Storage/FileIO.cs
```diff
@@ -0,0 +1,224 @@
...
+					break;
+				case Streams.UnicodeEncoding.Utf16BE:
+					encodingForWriter = Encoding.BigEndianUnicode;
+					break;
+			}
+
+			Stream fileStream = null;
+			try
+			{
+				fileStream = await file.OpenStreamForWriteAsync();
```
jeromelaban: Why did you implement the using pattern manually ?

This should work better:
```suggestion
				using(var fileStream = await file.OpenStreamForWriteAsync())

```

src/Uno.UWP/Storage/FileIO.cs
```diff
@@ -0,0 +1,224 @@
...
+			using (Stream fileReadStream = await file.OpenStreamForReadAsync())
+			{
+				using (BinaryReader binReader = new BinaryReader(fileReadStream))
+				{
+					try
+					{
+						byte firstByte = binReader.ReadByte();
+						switch (firstByte)
+						{
+							case 0xfe:
```
jeromelaban: Can you validate that behavior in a test ?

src/Uno.UWP/Storage/FileIO.cs
```diff
@@ -0,0 +1,224 @@
...
+				case Streams.UnicodeEncoding.Utf8:
+					encodingForReader = Encoding.UTF8;
+					break;
+				case Streams.UnicodeEncoding.Utf16LE:
+					encodingForReader = Encoding.Unicode;
+					break;
+				case Streams.UnicodeEncoding.Utf16BE:
+					encodingForReader = Encoding.BigEndianUnicode;
+					break;
+			}
```
jeromelaban: This should be a method, such as `ConvertEncoding`, and use it in the rest of the file.

src/Uno.UWP/Storage/FileIO.cs
```diff
@@ -0,0 +1,224 @@
...
+
+			string output = "";
+			using (Stream fileStream = await file.OpenStreamForReadAsync())
+			{
+				using (StreamReader streamReader = new StreamReader(fileStream, encodingForReader))
+				{
+					output = streamReader.ReadToEnd();
```
jeromelaban: No need for a variable here

```suggestion
					return streamReader.ReadToEnd();

```

src/Uno.UWP/Storage/FileIO.cs
```diff
@@ -0,0 +1,224 @@
...
+
+		public static IAsyncOperation<string> ReadTextAsync(IStorageFile file, Streams.UnicodeEncoding encoding)
+			=> ReadTextTaskAsync(file, encoding).AsAsyncOperation<string>();
+
+		private static async Task<string> ReadTextTaskAsync(IStorageFile file)
+		{
+			if (file is null)
+			{
+				// UWP throws NullReferenceException
+				// but it was choosen to throw ArgumentNullException , as it is more appriopriate in this caase
```
jeromelaban:
```suggestion
				// but it was chosen to throw ArgumentNullException , as it is more appropriate in this case

```

src/Uno.UWP/Storage/FileIO.cs
```diff
@@ -0,0 +1,224 @@
...
+				throw new ArgumentNullException("StorageFile cannot be null");
+			}
+
+			string output = "";
+			Stream fileStream = null;
```
jeromelaban: Same comments as above (`using` and `return`)

src/Uno.UWP/Storage/FileIO.cs
```diff
@@ -0,0 +1,224 @@
...
+			}
+			return output;
+		}
+
+		public static IAsyncOperation<string> ReadTextAsync(IStorageFile file)
+			=> ReadTextTaskAsync(file).AsAsyncOperation<string>();
+
+		private static async Task<IList<string>> ReadLinesTaskAsync(IStorageFile file, Streams.UnicodeEncoding encoding)
+		{
+			string output = await ReadTextTaskAsync(file, encoding);
```
jeromelaban: Use `System.IO.File.ReadAllLines()` instead.

src/Uno.UWP/Storage/FileIO.cs
```diff
@@ -0,0 +1,224 @@
...
+			var separators = new String[] { Environment.NewLine };
+			return output.Split(separators, StringSplitOptions.None);
+		}
+
+		public static IAsyncOperation<IList<string>> ReadLinesAsync(IStorageFile file, Streams.UnicodeEncoding encoding)
+			=> ReadLinesTaskAsync(file, encoding).AsAsyncOperation<IList<string>>();
+		private static async Task<IList<string>> ReadLinesTaskAsync(IStorageFile file)
+		{
+			string output = await ReadTextTaskAsync(file);
+			var separators = new String[] { Environment.NewLine };
```
jeromelaban: Use `System.IO.File.ReadAllLines()` instead.

src/Uno.UWP/Storage/FileIO.cs
pkar70: Because Codacy shows error when `using` is inside `using`.
https://app.codacy.com/gh/unoplatform/uno/pullRequest?prid=4827590

src/Uno.UWP/Storage/FileIO.cs
pkar70: In a test for CI build, or once at my side?

src/Uno.UWP/Storage/FileIO.cs
pkar70: Please, choose either
a) `using` inside `using`, or
b) `using` inside `try/finally`,
and if you choose (a), please block this codacy rule.

src/Uno.UWP/Storage/FileIO.cs
carldebilly: It's already disposed by the `using` on the `StreamReader`.

https://docs.microsoft.com/en-us/dotnet/api/system.io.streamreader.-ctor?view=netframework-4.8#System_IO_StreamReader__ctor_System_IO_Stream_System_Text_Encoding_System_Boolean_System_Int32_System_Boolean_
> Unless you set the leaveOpen parameter to true, the StreamReader object calls Dispose() on the provided Stream object when StreamReader.Dispose is called.

pkar70: I used `using(fileStream.. { using(streamWriter... } }` originally.
Codacy suggested same thing as now - change code to use `try { ... using(streamWriter...} finally` (see link from codacy, `Compliant Solution` section). So, I changed code.

And even if .Net API says that StreamReader disposes Stream, it cost nothing to try to stream.Dispose (), and it would be protection for bad implementation of StreamReader.Dispose(). I think calling dispose twice is better than memory/resources leak.

carldebilly: I agree. Codacy is _smart enough_ to know the `StreamWriter` will dispose the stream, but as you said in this case it's not a problem.

But it's common for components to throw an `ObjectDisposedException` when trying to dispose twice the same object. That's why this rule exists.

src/Uno.UWP/Storage/FileIO.cs
pkar70: If we are to use System.IO.File, all methods could be shortened. But as far as I know, StorageFile is broader than System.IO.File (the latter is only for real files).

sonnemaf:
Are there any plans to merge this PR? I would love to have it.

pkar70:
If you want/are able to compile Uno at your side, you can always use this PR in your compilation, before it will be merged. I'm using this PR in several my apps.
I didn't expect merging this PR this year.

MartinZikmund:
Commits merged as part of #3628 🎉

PR #2561, opened 2020-01-30
T4generator line endings independence

pkar70:

## PR Type

- Build or CI related changes

Sometimes we can get errors like
`Generating mixins .\Mixins\Android\FrameworkElementMixins.tt(1,1): error CS1519: Compiling transformation: Invalid token 'this' in class, struct, or interface member declaration`

## What is the new behavior?
No such errors.

## PR Checklist

## Other information
When I get new clone of Uno, my ability to build is vanished. I complain on Lobby, and Jérôme Laban says, "In most cases it's related to the .tt files having incorrect line endings", and "this is related to your default git configuration".
So, this small commit makes build independent of line endings (and one's default git configuration).


jeromelaban:
Thanks! Unfortunately, this solution does not work when building on macOS. There's probably a need to use the proper line ending based on the current platform.

pkar70:
Seems strange that Microsoft.VisualStudio.TextTemplating is so 'fragile', than line ending could cause compile errors. I don't have macOS, so I cannot test it on Mac...
Are _ANDROID__ etc. flags accessible while t4generator is compiled? And what is proper line ending on mac? Classic mac, 0x0D, or newer mac, 0x0A?

And what means build error in this simple PR: `The nuget command failed with exit code(1) and error(Response status code does not indicate success: 500 (Internal Server Error - VS800075: The project with id 'vstfs:///Classification/TeamProject/1dd81cbd-cb35-41de-a570-b0df3571a196' does not exist, or you do not have permission to access it. (DevOps Activity ID: 55C7C3FE-6EBE-4015-9B34-2D6870CE734F)).)`

jeromelaban:
For the line ending, using `Environment.NewLine` seems the most appropriate, let's try with that.

As for the not authorized error, this is a fork based PR error, I'll be fixing that in a separate PR.

pkar70:
So, now it uses Environment.NewLine :)

src/T4Generator/Program.cs
```diff
@@ -359,4 +362,26 @@ static void ProcessTemplate(string[] args)
             }
         }
     }
+
+	static class Tools
```
jeromelaban: That will do! Can you move the class to another file, and call it `StringHelper` maybe ?

pkar70: Which namespace? Or, class Tools, without namespace?

jeromelaban: Same namespace, another file.

pkar70:
So, file StringHelper.cs, with one function ConvertLineEndings, in namespace CustomHost and class Tools?

jeromelaban:
The file name and namespace does not matter much, it just has to not be in the same file.

jeromelaban:
If you can, can you squash your history ? There are commits that are only merges.

pkar70:
Sorry, I don't want to use squash - I tried to, but in effect one file was treated as new (before squash, it was only changed).

PR #2563, opened 2020-01-30
ToggleSwitch.Header visibility

pkar70:
GitHub Issue (If applicable): #3437

## PR Type

- Bugfix

ToggleSwitch.Header is not visible.

## What is the new behavior?
ToggleSwitch.Header is visible.

## PR Checklist

## Other information

### DIFF
--- a/doc/ReleaseNotes/_ReleaseNotes.md
+++ b/doc/ReleaseNotes/_ReleaseNotes.md
@@ -25,7 +25,7 @@
 - [Android] Enable support for non-native `Popup` by default. Can be deactivated through `FeatureConfiguration.Popup.UseNativePopup` set to true (See #2533 for more details)
 
 ### Bug fixes
-
+- ToggleSwitch.Header is now displayed.
 - [#2465] Raising macOS Button Click event
 - [#2506] `DesignMode.DesignMode2Enabled` no longer throws (is always `false` on non-UWP platforms)
 - [#915] FontFamily values are now properly parsed on WebAssembly, updated docs with new info
--- a/src/Uno.UI/UI/Xaml/Style/Generic/Generic.xaml
+++ b/src/Uno.UI/UI/Xaml/Style/Generic/Generic.xaml
@@ -5792,7 +5792,7 @@
 						</Grid.ColumnDefinitions>
 						<ContentPresenter x:Name="HeaderContentPresenter"
 										  x:DeferLoadStrategy="Lazy"
-										  Visibility="Collapsed"
+										  Visibility="Visible"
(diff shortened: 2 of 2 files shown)


jeromelaban:
Thank you for your contribution. Is the visibility set this way in the original UWP template? If not, the visibility is most likely changed by the control's code.

pkar70:
I don't know how it is in "original UWP template", but I know that this works on Android :)
Maybe it should be made visible somewhere in the code, not in template; maybe it is quick and dirty workaround... But it works with ToggleButton created in .cs code, and with ToggleButton defined in XAML.

jeromelaban:
After digging a little bit, the `HeaderContentPresenter` is kept collapsed to benefit from the `x:Load` feature. To enable the control, you'll need to add a test that validates that either the `Header` or `HeaderTemplate` property is set and change the visibility accordingly.

pkar70:
HeaderContentPresenter  is the only element in row 0 of ToggleSwitch, and height of this row is set to Auto. So, if no Header is set, then height = 0, and visibility=hidden is not necessary... my way is just simpler, and within my abilities :)
I don't know why Header works in other (at least, some other) controls, and doesn't work for ToggleSwitch.

jeromelaban:
`x:Load` is removing the element from the visual tree, so keeping it Collapsed is important.

pkar70:
Ok, where should be added "test that validates that either the Header or HeaderTemplate property is set and change the visibility accordingly", and why it actually works for e.g. TextBox or Slider, and not work for ToggleSwitch?

jeromelaban:
The other controls may not have the appropriate property management, because we did not know about that `x:Load` behavior. So yes, a test that validates that it's possible to set the header here is appropriate.

pkar70:
But WHERE such test should be? I can copy such test from TextBox, as it has correct Header handling, but I don't know where it is coded. Which file (or which class)? Maybe it is somewhere deep hidden in .tt files?

jeromelaban:
There is no test that validates that behavior for the other controls at this point. Create a new one that validates that behavior.

pkar70:
What is difference (and where) between Slider and ToggleSwitch? Both have (in generic.xaml) defined Grid, with first (0) row with Height=Auto, and HeaderContentPresenter with Visibility="Collapsed" in this row. But some code (I didn't find where this is in code) properly sets visibility in Slider, and Header is displayed, and such code is missing from ToggleSwitch.

jeromelaban:
As I mentioned, the original template has an `x:Load` defined, this is the reason it's done this way.

pkar70:
Resolved by another way in #3494

Issue #2598, opened 2020-02-04
[Android] Support HingeAngleSensor


MartinZikmund:
## What would you like to be added:

`HingeAngleSensor` on Android

## Why is this needed:

Surface Uno, Duo, Neo!

## For which Platform:

- [ ] iOS
- [x] Android
- [ ] WebAssembly
- [ ] WebAssembly renderers for Xamarin.Forms
- [ ] macOS
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

jeromelaban:
@MartinZikmund you should be able to use `ApiExtension` to enable that scenario. Let me know how this goes :)

jeromelaban:
(docs: https://github.com/unoplatform/uno/blob/e61a1da0df49d2d93e32d71e2801fd84689bb007/doc/articles/uno-development/api-extensions.md)

MartinZikmund:
Merged, closing

PR #2599, opened 2020-02-04
[Android] Implemented HingeAngleSensor

MartinZikmund:
GitHub Issue (If applicable): #2598 

## PR Type

- Feature

Surface Duo's hinge angle sensor is not supported.

## What is the new behavior?

Surface Duo's hinge angle sensor is supported.

## PR Checklist


MartinZikmund:
Unfortunately I haven't been able to find a way to simulate hinge events in the current emulator, but it should work. If you know how to do it, please let me know :-D  .

src/Uno.UWP/Devices/Sensors/HingeAngleSensor.Android.cs
```diff
@@ -0,0 +1,103 @@
...
+		private static HingeAngleSensor _instance;
+		private static INativeHingeAngleSensor _hingeAngleSensor = null;
+
+		private TypedEventHandler<HingeAngleSensor, HingeAngleSensorReadingChangedEventArgs> _readingChanged;
+
+		private HingeAngleSensor()
+		{
+		}
+
+		public static IAsyncOperation<HingeAngleSensor> GetDefaultAsync()
```
jeromelaban: Does GetDefaultAsync return a dummy HingeAngleSensor if there is none ?

src/Uno.UWP/Devices/Sensors/Helpers/INativeHingeAngleSensor.cs
```diff
@@ -0,0 +1,18 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Uno.Devices.Sensors.Helpers;
+using Windows.Devices.Sensors;
+using Windows.Foundation;
+
+namespace Uno.Devices.Sensors
+{
+    public interface INativeHingeAngleSensor
```
jeromelaban: All the WinRT apis are backed by an internal interface. Those are not yet exported by the Sync generator, but will soon be. For example, this one has this:

```csharp
namespace Windows.Devices.Sensors
{
	[ContractVersion(typeof(UniversalApiContract), 458752)]
	[ExclusiveTo(typeof(HingeAngleSensor))]
	[GuidAttribute(3922968066, 49119, 17279, 140, 41, 136, 199, 115, 147, 211, 9)]
	internal interface IHingeAngleSensor
	{
		[RemoteAsync]
		IAsyncOperation<HingeAngleReading> GetCurrentReadingAsync();

		string DeviceId { get; }
		double MinReportThresholdInDegrees { get; }
		double ReportThresholdInDegrees { get; set; }

		event TypedEventHandler<HingeAngleSensor, HingeAngleSensorReadingChangedEventArgs> ReadingChanged;
	}
}
```

src/Uno.UWP/Devices/Sensors/HingeAngleSensor.Android.cs
MartinZikmund: If there is no hinge on the device, the GetDefaultAsync method returns null, which is in line with other sensors when not available on the device. Also in case the SDK is not referenced, there is logging to inform the developer.

src/Uno.UWP/Devices/Sensors/Helpers/INativeHingeAngleSensor.cs
MartinZikmund: I was thinking that it still might be worth keeping as a separate interface to make it clear that it is a Uno extension. Plus I would have to add public constructor for HingeAngleReading (so the implementation can trigger the event)

jeromelaban: Fair points. The case for the public constructors will most probably have to be dealt with in the context of a more general extension mechanism. Let's keep it that way for now then.

Issue #626, opened 2019-02-25
[Epic] macOS support


jeromelaban:
## I'm submitting a...
- Feature request

## Description
Known tasks :

- [x] Publish macOS nuget target
- [x] Document macOS build target
- [x] Package the Samples app
- [ ] Update the vsix templates to include a macOS target
- [x] Validate measure/arrange phases
- [x] Validate overall initial behavior
- [x] Uno Calc: add support for macOS (https://github.com/unoplatform/uno/issues/3103)
- [ ] Update the vsix templates to include a main Storyboard (https://github.com/unoplatform/uno/issues/3124)
- [x] Add support for Localization (https://github.com/unoplatform/uno/issues/3129)

## Structure
- [ ] Text rendering
- [x] Font size (too small) https://github.com/unoplatform/uno/pull/2926
- [ ] TextRendering performance
- [ ] XAML events WeakReference is collected too early
- [ ] RenderTransforms https://github.com/unoplatform/uno/issues/2971
- [ ] UI Testing support
- [ ] Key events (https://github.com/unoplatform/uno/issues/3167)

## Runtime Tests (non exhaustive list)
https://github.com/unoplatform/uno/issues/9282

## Controls
- [x] TextBox https://github.com/unoplatform/uno/pull/2926
- [ ] TextBox support for multiline and Enter key
- [ ] List Virtualization 
- [ ] ScrollViewer (unable to scroll UnitTests)
- [ ] GridView support
- [x] Shapes (https://github.com/unoplatform/uno/issues/2961)
- [x] Image support
- [x] PasswordBox support (https://github.com/unoplatform/uno/issues/3005)
- [x] CheckBox (https://github.com/unoplatform/uno/issues/3035)
- [ ] Popups / Flyouts
- [x] ComboBox (https://github.com/unoplatform/uno/issues/2972)
- [x] Slider
- [ ] Canvas
- [ ] ViewBox
- [ ] Native Button
- [ ] Native ToggleSwitch
- [ ] Native ProgressBar
- [ ] Native CheckBox
- [ ] Native ProgressRing
- [ ] CameraCaptureUI
- [ ] NavigationView
- [ ] Animations
- [x] DisplayInformation (https://github.com/unoplatform/uno/issues/2907)
- [x] MediaPlayer (https://github.com/unoplatform/uno/issues/3020)
- [x] Clipboard (https://github.com/unoplatform/uno/issues/2915)
[...]

PR #2564, opened 2020-01-30
Enable animation of TransformGroup

dr1rrb:
GitHub Issue: _private_

## Feature
Add basic support of animation of `TransformGroup`

## What is the current behavior?
Trying to animate a property of a child transform of a `TransfomrGroup` either crashes the app or does nothing.

## What is the new behavior?
Basic animations are supported (basic means like for `CompositeTransfrom`, we are able to animate only one property at a time)

## PR Checklist
- [x] Tested code with current [supported SDKs](https://github.com/unoplatform/uno/blob/master/README.md#uno-features)
- [ ] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)
- [ ] [Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [ ] [Wasm UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md#running-the-webassembly-ui-tests-snapshots) are not showing unexpected any differences. Validate PR `Screenshots Compare Test Run` results.
- [x] Contains **NO** breaking changes
- [ ] Updated the [Release Notes](https://github.com/unoplatform/uno/tree/master/doc/ReleaseNotes)
- [x] Associated with an issue (GitHub or internal) (cf. comments below)

PR #2622, opened 2020-02-08
Support for DoubleAnimation, DoubleAnimationUsingKeyFrames on macOS

MartinZikmund:
GitHub Issue (If applicable): #626 

## PR Type

- Feature

`DoubleAnimation` and `DoubleAnimationUsingKeyFrames` are not supported on macOS.

## What is the new behavior?

Supported.

## PR Checklist


jeromelaban:
Amazing! Which samples does this PR light up ?

MartinZikmund:
@jeromelaban I tested it on DoubleAnimation_BeginTime and DoubleAnimationUsingKeyFrames_TranslateX, but should work on more, I will test that tomorrow :-)

src/Uno.UI/UI/Xaml/Media/Animation/Animators/FloatValueAnimator.iOSmacOS.cs
```diff
@@ -23,7 +23,12 @@ internal class FloatValueAnimator : IValueAnimator
 		private const int FrameRate = 50;
 
 		private long _numberOfFrames;
+#if __IOS__
 		private CADisplayLink _displayLink;
+#else
+		private NSTimer _timer;
+#endif
+
```
dotMorten: Curious if you could use an approach as described here instead of an NSTimer: https://stackoverflow.com/a/14165842 ?

MartinZikmund: I tried that, but I couldn't get the `CVDisplayLink` to run on the UI thread. It seems it is quite limited as compared to the `CADisplayLink` unfortunately

dotMorten: Isn't that fairly common though to have a render loop runs on the render thread, not the UI thread? (similar to how UWP has a separate composition thread).
I can however see how that could cause an issue with sync, but you probably want animations be synced to render thread instead of the UI Thread?

jeromelaban: The common reason the timer is needed on the UI Thread here is for dependent animations mainly, to update affected dependency properties. Independent animations are generally handled differently (@dr1rrb will most probably have some feedback there).

dotMorten: Aaah makes sense 👍

MartinZikmund: The `GPUFloatAnimator` is there too, so it should work correctly 🙂

MartinZikmund:
I have tested with the `DoubleAbimationTestsControl` and most samples do work properly, only have issues with Pause/Resume

src/Uno.UI/UI/Xaml/Media/Animation/Animators/FloatValueAnimator.iOSmacOS.cs
dr1rrb: Actually on iOS for independent animations we have 2 animators running at once:
1. The first which effectively animates the UI (the `GPUFloatAnimator`) 
1. The second which "fakely" animates the dependency property, and which obviously has to run on the UI thread (this `FloatValueAnimator`)

So it's indeed expected for the timer to tick on the UI thread here.

However I would argue about the need to always have those 2 animators instead of updating the DP only once at the end of the animation (or when paused). In my opinion we should add a feature flag for that, and disable it by default for perf considerations.

And about dependent animations, I think that they are not allowed on Uno yet, but they should indeed use this animator ... or actually a shared implementation which use the `DispatcherTimer`!

Note: On WASM animations are always running on UI thread, and on Android there is still no way to make them fully independent 😶

src/Uno.UI/UI/Xaml/Media/Animation/Animators/GPUFloatValueAnimator.iOSmacOS.cs
```diff
@@ -5,7 +5,13 @@
 using System.Text;
 using Windows.Foundation;
 using Uno.UI.DataBinding;
+#if __IOS__
 using UIKit;
+using NativeView = UIKit.UIView;
```
dr1rrb: Personally, when I use type aliases, I prefix the alias by `_` as it's a _private_ type. I found that much easier to understand when reading the code, especially when multi targeting code like here, it makes clear that we are dealing with a native object.

For instance here I would have defined:
```csharp
using _View = UIKit.UIView;
```

But FYI this is not an enforced rule, so it's your choice.

dr1rrb:
@MartinZikmund I'm curious, if you have the opportunity to check how it behaves with https://github.com/unoplatform/uno/pull/2564#issuecomment-580499573 

BTW, I'm about to automate this UI test, so it could also serve as test of your PR.

src/Uno.UI/UI/Xaml/Media/Animation/DoubleAnimation.cs
```diff
@@ -342,7 +342,7 @@ private void OnEnd()
 
 			if (FillBehavior == FillBehavior.HoldEnd)//Two types of fill behaviors : HoldEnd - Keep displaying the last frame
 			{
-#if __IOS__
+#if __IOS__ || __MACOS__
 				// iOS: Here we make sure that the final frame is applied properly (it may have been skipped by animator)
```
carldebilly: You should adjust the comments to include macOS too when appropriate.

src/Uno.UI/UI/Xaml/Media/Animation/Animators/FloatValueAnimator.iOSmacOS.cs
dr1rrb: Well, I just saw that we already have a [`DispatcherFloatAnimator`](https://github.com/unoplatform/uno/blob/master/src/Uno.UI/UI/Xaml/Media/Animation/Animators/DispatcherFloatAnimator.cs) @MartinZikmund did you tired to use it ? (Note: I didn't checked how complex it is to change the dependencies to use it instead of the `FloatValueAnimator`).

src/Uno.UI/UI/Xaml/Media/Animation/Animators/GPUFloatValueAnimator.iOSmacOS.cs
MartinZikmund: Fixed :-)

src/Uno.UI/UI/Xaml/Media/Animation/DoubleAnimation.cs
MartinZikmund: Fixed

Issue #2377, opened 2019-12-21
BitmapImage.SetSource is inconsistent between platforms


SuperJMN:
Giving this converter that converts `bytes[]` to a `BitmapImage()`

```
    public class ByteArrayToImageConverter : IValueConverter
    {
        public object Convert(object value, Type targetType, object parameter, string language)
        {
            if (!(value is byte[] bytes))
                return null;

            var image = new BitmapImage();

#if WINDOWS_UWP
            using (var stream = new InMemoryRandomAccessStream())
            {
              image.SetSource(stream);
              var s = stream.AsStreamForWrite();
              s.Write(bytes, 0, bytes.Length);
            }
#else
            using (Stream stream = new MemoryStream(bytes))
            {
                image.SetSource(stream);
            }
#endif
            return image;
        }

        public object ConvertBack(object value, Type targetType, object parameter, string language)
        {
            throw new NotImplementedException();
        }
    }
```

The WINDOWS_UWP part and the part for other platforms is **symmetric**. However, when trying in, for instance, Android, this code produces the following exception:

> **System.ObjectDisposedException:** : 'Can not access a closed Stream.'

As a workaround, the converter should be written as:

```
    public class ByteArrayToImageConverter : IValueConverter
    {
        public object Convert(object value, Type targetType, object parameter, string language)
        {
            if (!(value is byte[] bytes))
                return null;

            var image = new BitmapImage();

#if WINDOWS_UWP
            using (var stream = new InMemoryRandomAccessStream())
            {
              image.SetSource(stream);
              var s = stream.AsStreamForWrite();
              s.Write(bytes, 0, bytes.Length);
            }
#else
            Stream stream = new MemoryStream(bytes);
            image.SetSource(stream);
#endif
            return image;
        }

        public object ConvertBack(object value, Type targetType, object parameter, string language)
        {
            throw new NotImplementedException();
        }
    }
```

Please, notice that the #else part no longer encloses the stream in a `using` block, so the stream isn't disposed.

I don't know what are the implications of this, but I would like to point out that the stream should be disposed eventually. We need to ensure that the resources are freed properly.

I would like to put into consideration that this converter raises the concert of SetSource not behaving the same between platforms, not only in the abstractions used, but also regarding the life cycle of the underlying stream.

I personally think this should be improved.

jeromelaban:
There are two issues here:
[...]

PR #2698, opened 2020-02-26
Improve BitmapImage.SetSource()

TopperDEL:
GitHub Issue (If applicable): #2377 

## PR Type

- Bugfix

BitmapImage.SetSource Needs Access to the provided stream, Which makes it difficult to handle. It should be in line with SetSourceAsync.

## What is the new behavior?
BitmapImage.SetSource makes a local copy of the provided stream and therefore manges it in ist own.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Media/Imaging/BitmapSource.cs
+++ b/src/Uno.UI/UI/Xaml/Media/Imaging/BitmapSource.cs
@@ -60,7 +60,9 @@ namespace Windows.UI.Xaml.Media.Imaging
 			PixelWidth = 0;
 			PixelHeight = 0;
 
-			Stream = streamSource;
+			MemoryStream copy = new MemoryStream();
+			streamSource.CopyTo(copy);
+			Stream = copy;
 		}
 
 		public async Task SetSourceAsync(Stream streamSource)
--- a/src/Uno.UI/UI/Xaml/Media/Imaging/BitmapSource.cs
+++ b/src/Uno.UI/UI/Xaml/Media/Imaging/BitmapSource.cs
@@ -59,10 +59,18 @@ namespace Windows.UI.Xaml.Media.Imaging
 		{
 			PixelWidth = 0;
 			PixelHeight = 0;
-
-			MemoryStream copy = new MemoryStream();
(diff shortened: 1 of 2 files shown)


TopperDEL:
I don't know what's wrong with the checks ("Pull Request Labeler" failed). Did I name the PR wrong?

jeromelaban:
Thanks for your submission!

You'll need to add a UI Test to validate the behavior. You can add a control here:

jeromelaban:
Thank you for your contribution!

You'll need to add a UI test control here: https://github.com/unoplatform/uno/tree/master/src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/ImageTests

In the constructor or a loaded event, your can add reference an image and call `SetSource` with a stream that comes from an embedded resource image, and that sets that stream position to the end (so that it would fail if the image was read later.

Once there, you can add a UI Tests here: https://github.com/unoplatform/uno/blob/63065423805d82e275c7dfa927940fcf1313d25d/src/SamplesApp/SamplesApp.UITests/Windows_UI_Xaml_Controls/ImageTests/UnoSamples_Tests.BitmapSource.cs#L18

That will validate that the size of the image is the appropriate one. Note that the UI Test will need to be disabled for Wasm (`SetSource` is not supported yet).

You can get more information on [how to create uitests here](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md#creating-ui-tests).

src/Uno.UI/UI/Xaml/Media/Imaging/BitmapSource.cs
```diff
@@ -59,8 +59,18 @@ public void SetSource(Stream streamSource)
 		{
 			PixelWidth = 0;
 			PixelHeight = 0;
-
-			Stream = streamSource;
+			
+			if (streamSource != null)
```
carldebilly, line 63: Code convention here: when you have a `else` block, please try to avoid a _not equal_ `!=` had write the _equal_ `==` first.

jeromelaban:
The content of this PR has been addressed with another PR. Closing this one, thanks for your PR !

Issue #2839, opened 2020-03-19
[Android] SelectedItem state may not be synchronized properly


jeromelaban:
## Current behavior

SelectedItem state in ListViewItemStyle may not synchronized properly.

## How to reproduce it (as minimally and precisely as possible)

TBD

## Environment

Nuget Package:

Package Version(s):

Affected platform(s):

- [ ] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] WebAssembly renderers for Xamarin.Forms
- [ ] macOS
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

Visual Studio:

- [ ] 2017 (version: )
- [ ] 2019 (version: )
- [ ] for Mac (version: )

Relevant plugins:

- [ ] Resharper (version: )

## Anything else we need to know?

https://dev.azure.com/nventive/Umbrella/_workitems/edit/176403

PR #2847, opened 2020-03-20
Fix ListView selection mode changed

dr1rrb:
GitHub Issue: fixes https://github.com/unoplatform/uno/issues/2839

## Bugfix
Invalid visual state when changing a `ListView`'s selectuion mode while one or more items are selected.

When the `SelectionMode` is changed to `None` while at least on item is selected, the `ListView` tries to un-select them, but doing this it alter the collection that it is enumerating.

## What is the new behavior?
We create a clone of the selected items collection.

## PR Checklist

### DIFF
--- a/src/Uno.UI.Tests/Windows_UI_XAML_Controls/ListViewBaseTests/Given_ListViewBase.cs
+++ b/src/Uno.UI.Tests/Windows_UI_XAML_Controls/ListViewBaseTests/Given_ListViewBase.cs
@@ -19,6 +19,9 @@ namespace Uno.UI.Tests.ListViewBaseTests
 	public class Given_ListViewBase
 	{
 #if !NETFX_CORE
+		// Make sure to have a valid custom theme set, so it won't try to read it from the Application.Current(<<== null).RequestedTheme
+		[TestInitialize] public void Init() => global::Uno.UI.ApplicationHelper.RequestedCustomTheme = "HighContrast";
+
 		[TestMethod]
 		public void When_MultiSelectedItem()
 		{
@@ -233,7 +236,7 @@ namespace Uno.UI.Tests.ListViewBaseTests
 			{
 				Assert.Fail("Container should be a ListViewItem");
 			}
-	}
+		}
 
 		[TestMethod]
 		public void When_Single_SelectionChanged_And_SelectorItem_IsSelected_Changed()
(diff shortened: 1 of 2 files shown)


src/Uno.UI.Tests/Windows_UI_XAML_Controls/ListViewBaseTests/Given_ListViewBase.cs
```diff
@@ -19,6 +19,9 @@ namespace Uno.UI.Tests.ListViewBaseTests
 	public class Given_ListViewBase
 	{
 #if !NETFX_CORE
+		// Make sure to have a valid custom theme set, so it won't try to read it from the Application.Current(<<== null).RequestedTheme
+		[TestInitialize] public void Init() => global::Uno.UI.ApplicationHelper.RequestedCustomTheme = "HighContrast";
```
jeromelaban: We probably should set that at assembly level: https://github.com/unoplatform/uno/blob/d804b940617548f579c2826b7ba9e53c55eb554e/src/Uno.UI.Tests/Global.cs

davidjohnoliver: Not sure it's a good idea to set this, some tests IIRC rely on the default theme being set. 

Where is it causing a problem if RequestedCustomTheme is not set?

dr1rrb: It's in the generated code, while resolving the `ContentControlThemeFontFamily`:
```cs
public static FontFamily ContentControlThemeFontFamily
{
	get
	{
		// Custom themes defined for this resource: checking custom theme.
		var currentCustomTheme = global::Uno.UI.ApplicationHelper.RequestedCustomTheme;
		switch(currentCustomTheme)
		{
			case "HighContrast": return ContentControlThemeFontFamily___HighContrast;
		}

		// Element's RequestedTheme not supported yet. Fallback on Application's RequestedTheme.
		var currentTheme = global::Windows.UI.Xaml.Application.Current.RequestedTheme; <<<<==== Application.Current is null
		switch(currentTheme)
		{
			default: return ContentControlThemeFontFamily___Default;
			case global::Windows.UI.Xaml.ApplicationTheme.Light: return ContentControlThemeFontFamily___Light;
		}
	}
}
```

And I'm not convince it's good idea to put it at the assembly level, as it's only a patch to fix tests when run locally. I assume that on the CI the tests sequence make it work properly, so changing this value up-front might fail some other tests.

The real fix would be to generate a `null` proof code, and I actually started to to do that until I realized I was able to workaround the bug, without any risk of destabilization. Another solution would also be to have a non `null` `Application.Current`.

I've created another issue for this : https://github.com/unoplatform/uno/issues/2852

davidjohnoliver: Aha I didn't read the comment properly, the real problem is that `Application.Current` is null. 

That's an easy fix:
```csharp
UnitTestsApp.App.EnsureApplication();
```

jeromelaban:

Issue #2856, opened 2020-03-20
[WASM] Cannot load image from bytes


SuperJMN:
I'm processing downloading and image that comes from a remote service as a byte[]. Currently, it's impossible to make the Image load the image from bytes using Image.Source.

I'm using a converter like this that works for Android and UWP:

```
    public class ByteArrayToImageSourceConverter : IValueConverter
    {
        public object Convert(object value, Type targetType, object parameter, string language)
        {
            if (!(value is byte[] bytes))
                return null;

            var image = new BitmapImage();

#if WINDOWS_UWP
            using (var ms = new InMemoryRandomAccessStream())
            {
                using (var writer = new DataWriter(ms.GetOutputStreamAt(0)))
                {
                    writer.WriteBytes((byte[])value);
                    writer.StoreAsync().GetResults();
                }

                image.SetSource(ms);
            }
#else
            Stream stream = new MemoryStream(bytes);
            image.SetSource(stream);
#endif
            return image;
        }

        public object ConvertBack(object value, Type targetType, object parameter, string language) => throw new NotSupportedException();
    }
```
@jeromelaban said that adding support for this scenario isn't hard using this:
https://css-tricks.com/data-uris/

And modifying this:
https://github.com/unoplatform/uno/blob/318398d7045e437938b42bfd3155bf05b2915190/src/Uno.UI/UI/Xaml/Controls/Image/Image.wasm.cs#L140

I've taken a slight look. If I understand correctly, it would be a matter of modifying Uno Platform itself, concretely, the below above, so support "Data Uris".

But since I don't have enough knowledge, I can't go much far. 

If somebody gives more hints I may be able to help.

PR #2860, opened 2020-03-22
Image from bytes

SuperJMN:
GitHub Issue: https://github.com/unoplatform/uno/issues/2856

# Adds the ability to load images from bytes in WASM heads.

This is my first PR, so please, be patient. I don't expect this to be merged soon, because it surely will need refinement and maybe important changes

Basically, I've edited src/Uno.UI/UI/Xaml/Controls/Image/Image.wasm.cs to handle the use case when ImageSource.Stream isn't null. I don't know if this condition is enough to guarantee that the ImageSource should be loaded from the Stream, but it worked for me.

In case we are loading the image from the ImageSource.Stream, I read it using ImageSharp.
I know this introduces a 3rd party library, but I found no better option to load the pixel data, since I don't have enough knowledge about how this scenarios are handled inside Uno. Feel free to suggest/edit/criticize and I will refine the PR until it meets the conditions to be merged.

Thanks for your understanding :)

### DIFF
--- a/src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/ImageTests/LoadFromBytes.xaml.cs
+++ b/src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/ImageTests/LoadFromBytes.xaml.cs
@@ -1,0 +1,71 @@
+using System;
+using System.IO;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using Uno.UI.Samples.Controls;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Controls;
+using Windows.UI.Xaml.Media.Imaging;
+
+namespace Uno.UI.Samples.UITests.ImageTestsControl
+{
+	[SampleControlInfo("Image", "LoadFromBytes")]
+	public sealed partial class LoadFromBytes : UserControl
+	{
+		private readonly string _imageUrl = "https://www.spriters-resource.com/resources/sheet_icons/30/32521.png";
+
+		public LoadFromBytes()
+		{
+			InitializeComponent();
+			Loaded += OnLoaded;
+		}
+
+		private async void OnLoaded(object sender, RoutedEventArgs e)
+		{
+			var client = new HttpClient(new CorsByPassHandler());
+
+			var result = await client.GetAsync(_imageUrl);
+			var stream = await result.Content.ReadAsStreamAsync();
+
+
+			var bitmap = new BitmapImage();
+#if !NETFX_CORE
+			await bitmap.SetSourceAsync(stream);
+#else
+				var memStream = new MemoryStream();
+				await stream.CopyToAsync(memStream);
+				memStream.Position = 0;
+				using (memStream)
+				{
+					await bitmap.SetSourceAsync(memStream.AsRandomAccessStream());
+				}
+#endif
+			MyImage.Source = bitmap;
+
+		}
+
+		public class CorsByPassHandler : DelegatingHandler
+		{
+			public CorsByPassHandler()
+			{
+				InnerHandler = new HttpClientHandler();
+			}
+
+			protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+			{
+				var builder = new UriBuilder(request.RequestUri)
+				{
+					Host = "cors-anywhere.herokuapp.com",
+				};
+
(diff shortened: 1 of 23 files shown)


src/Uno.UI/UI/Xaml/Controls/Image/Image.wasm.cs
```diff
@@ -102,7 +105,28 @@ private void OnSourceChanged(DependencyPropertyChangedEventArgs e)
 
 			_lastMeasuredSize = _zeroSize;
 
-			if (source is WriteableBitmap wb)
+			var stream = source?.Stream;
+			if (stream != null)
+			{
+				stream.Seek(0, SeekOrigin.Begin);
+				var image = SixLabors.ImageSharp.Image.Load<Bgra32>(stream);
```
jeromelaban: Thanks for the PR! In general we try to avoid adding mandatory dependencies to Uno, and in this case it's probably not necessary.

For this instance, you'll need to build a data uri, and call `SetImageUrl(dataUri);`, and let the browser do the decoding. You probably won't be able to determine the format of the image, so maybe use `application/octet-stream` mime type will work.

SuperJMN: Done! Thanks for the info!

src/Uno.UI.sln
```diff
@@ -155,8 +155,13 @@ EndProject
 Global
 	GlobalSection(SharedMSBuildProjectFiles) = preSolution
 		SamplesApp\SamplesApp.UnitTests.Shared\SamplesApp.UnitTests.Shared.projitems*{04b1b5eb-d42e-47de-ada0-eb863e5574fd}*SharedItemsImports = 13
+		SamplesApp\Benchmarks.Shared\Benchmarks.Shared.projitems*{21206058-d5ef-466d-8e4d-ac62357fb479}*SharedItemsImports = 5
```
jeromelaban: Those changes should not be needed.

src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/ImageTests/LoadFromBytes.xaml.cs
```diff
@@ -0,0 +1,71 @@
...
+
+		public LoadFromBytes()
+		{
+			InitializeComponent();
+			Loaded += OnLoaded;
+		}
+
+		private async void OnLoaded(object sender, RoutedEventArgs e)
+		{
+			var client = new HttpClient(new CorsByPassHandler());
```
jeromelaban: You do not need to download the image from the network. You can simply get the file locally in the embedded resources, and pass it as a stream to the ImageSource.

src/Uno.UI/UI/Xaml/Controls/Image/Image.wasm.cs
```diff
@@ -102,7 +106,15 @@ private void OnSourceChanged(DependencyPropertyChangedEventArgs e)
 
 			_lastMeasuredSize = _zeroSize;
 
-			if (source is WriteableBitmap wb)
+			var stream = source?.Stream;
+			if (stream != null)
+			{
+				stream.Seek(0, SeekOrigin.Begin);
+				var encodedBytes = Convert.ToBase64String(stream.ReadBytes());
+				var url = $"data:application/octet-stream;base64,{encodedBytes}";
```
jeromelaban: No need to use interpolation here:
```suggestion
				var url = "data:application/octet-stream;base64," + encodedBytes;
```

src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/ImageTests/LoadFromBytes.xaml.cs
jeromelaban: Also, you can show the same image as an example using normal loading mechanisms, so that the person reviewing the sample can determine that the behavior is the right one.

SuperJMN: > You do not need to download the image from the network. You can simply get the file locally in the embedded resources, and pass it as a stream to the ImageSource.

Is there a sample on how to do that? I don't know the exact mechanism. Do I need to copy the image to the Samples project and set the build action to Embedded Resource and load it using the GetResourceManifestStream to get the stream to the image?

SuperJMN:
Sorry, I did a mess. I will have to repeat the PR 😅

Issue #2836, opened 2020-03-19
[iOS] Button content alignment may be incorrect


jeromelaban:
## Current behavior

Button content alignment may be incorrect, given specific margins and alignments.

## How to reproduce it (as minimally and precisely as possible)

TBD

## Environment

Nuget Package: 2.2

Package Version(s):

Affected platform(s):

- [x] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] WebAssembly renderers for Xamarin.Forms
- [ ] macOS
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

Visual Studio:

- [ ] 2017 (version: )
- [ ] 2019 (version: )
- [ ] for Mac (version: )

Relevant plugins:

- [ ] Resharper (version: )

## Anything else we need to know?

https://dev.azure.com/nventive/Umbrella/_workitems/edit/177339

carldebilly:
PR https://github.com/unoplatform/uno/pull/2828 has been merged. Will produce Uno.UI v2.2.0-dev.194 which should be available on Nuget soon there: https://www.nuget.org/packages/Uno.UI/2.2.0-dev.194

PR #2861, opened 2020-03-22
fix: VisibleBoundsPadding causing text alignment problem on iOS.

carldebilly:
# Bugfix

Looks like a concurrency issue in iOS itself with `NSTextContainer`: at a very specific measurement, when the text is multi-line (max-lines=2), center-aligned and there's a notch (iPhone-X), it's causing a very weird alignment problem of the text where it's centered on the right-hand bound of the container, cutting the text in half.

## What is the new behavior?
The solution consists to requeue the `.SetPadding()` in the `VisibleBoundsPadding` (for iOS only). This could cause flickering issue which need to be tested, but should solve the reported problem.

## PR Checklist

- ~~[ ] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)~~
- ~~[ ] [Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)~~
- [X] Associated with an issue (GitHub or internal) and uses the [automatic close keywords]

## Other information
The situation is hard to reproduce elsewhere than in the reported app. No repro has been successfully done.

* fix https://github.com/unoplatform/uno/issues/2836
* fix https://dev.azure.com/nventive/Umbrella/_workitems/edit/177339


davidjohnoliver:
> The solution consists to requeue the .SetPadding() in the VisibleBoundsPadding (for iOS only). This could cause flickering issue which need to be tested, but should solve the reported problem.

It absolutely will cause flickering issues. 

It sounds like something that should be fixed in `TextBlock` itself.

Can we get more insight into this bug? Which change introduced it, etc

jeromelaban:
This seems extreme... Won't it create layout updates that will ripple through the whole UI very quickly ?

carldebilly:
New version should minimize impact on existing applications.

src/Uno.UI/Behaviors/VisibleBoundsPadding.cs
```diff
@@ -144,8 +145,13 @@ internal VisibleBoundsDetails(FrameworkElement owner)
 
 				_visibleBoundsChanged = (s2, e2) => UpdatePadding();
 
-				owner.LayoutUpdated += (s, e) => UpdatePadding();
-				owner.Loaded += (s, e) => ApplicationView.GetForCurrentView().VisibleBoundsChanged += _visibleBoundsChanged;
+				owner.SizeChanged += (s, e) => UpdatePadding();
```
davidjohnoliver: Unfortunately `SizeChanged` won't work here. The VisibleBoundsPadding depends on the view's absolute position in the window, which may be affected by changes anywhere in the view's hierarchy, without its own size necessarily changing. We absolutely need to hook onto `LayoutUpdated` here.

carldebilly: Ok...  if we add both it works...

@davidjohnoliver

Thanks for you greate reviews, as always.

jeromelaban:
@mergifyio rebase

jeromelaban:

Issue #2843, opened 2020-03-19
[WASM] Tapped event isn't called on nested child


davidjohnoliver:
## How to reproduce it (as minimally and precisely as possible)

1. Run the Windows Community Toolkit samples app. 
2. Navigate to the DataGrid sample.
3. Try to select one of the cells.

## Current behavior
Selection always ends up on the first cell.

The `DataGridCell.DataGridCell_PointerTapped` handler should be called, but isn't. Looking at the logs, `PointerDown` is raised for 4 views in the hierarchy (`DataGrid` handles pointer events at numerous levels), but `PointerUp` is only raised for 2 of them (`DataGridCell` not included). 

```
dotnet.js:1       DataGridRowsPresenter-919369"RowsPresenter": Dispatching event pointermove
dotnet.js:1 dbug: Windows.UI.Input.GestureRecognizer[0]
dotnet.js:1       Received a 'Move' for a pointer which was not considered as down. Ignoring event.
dotnet.js:1 dbug: Windows.UI.Xaml.UIElement.EventRegistration[0]
dotnet.js:1       DataGrid-917777"dataGrid": Dispatching event pointermove
dotnet.js:1 dbug: Windows.UI.Xaml.UIElement.EventRegistration[0]
dotnet.js:1       ScrollContentPresenter-8041"ScrollContentPresenter": Dispatching event pointermove
dotnet.js:1 dbug: Windows.UI.Xaml.UIElement.EventRegistration[0]
dotnet.js:1       DataGridCell-995873: Dispatching event pointerdown
dotnet.js:1 dbug: Windows.UI.Xaml.UIElement.EventRegistration[0]
dotnet.js:1       DataGridRow-995673: Dispatching event pointerdown
dotnet.js:1 dbug: Windows.UI.Xaml.UIElement.EventRegistration[0]
dotnet.js:1       DataGridRowsPresenter-919369"RowsPresenter": Dispatching event pointerdown
dotnet.js:1 info: Windows.UI.Xaml.UIElement.PointerCapture[0]
dotnet.js:1       DataGridRowsPresenter-919369"RowsPresenter": Capturing (Implicit) pointer Mouse/1
dotnet.js:1 dbug: Windows.UI.Xaml.UIElement.EventRegistration[0]
dotnet.js:1       DataGrid-917777"dataGrid": Dispatching event pointerdown
dotnet.js:1 dbug: Windows.UI.Xaml.UIElement.EventRegistration[0]
[...]

PR #2874, opened 2020-03-24
[WASM] Fix pointers for elt nested in manip aware content

dr1rrb:
GitHub Issue: fixes#https://github.com/unoplatform/uno/issues/2843

## Bugfix
Fix pointers event for elements that are nested in an element that has enabled manipulations.

When a manipulation stars on an element, it will implicitly capture the pointer to make sure to receive all updates even when pointer moves out of the range of this element. But this will prevent any nested child to receive pointer events (instead they will receive a 'pointerleave' on capture and a 'pointerenter' when releasing capture ... for mouse).

## What is the new behavior?
For implicit captures we use the `OriginalSource` (if possible) instead of the element that is actually capturing the pointer, so event will still bubble up from the lowest layer.

## PR Checklist

### DIFF
--- a/src/Uno.UI/UI/Xaml/UIElement.PointerCapture.cs
+++ b/src/Uno.UI/UI/Xaml/UIElement.PointerCapture.cs
@@ -102,10 +102,15 @@ namespace Windows.UI.Xaml
 					{
 						return false;
 					}
+					else
+					{
+						// Add the new king to the target
+						target.Kind |= kind;
+					}
 				}
 				else
 				{
-					target = new PointerCaptureTarget(element);
+					target = new PointerCaptureTarget(element, kind);
 					_targets.Add(element, target);
 
 					// If the capture is made while raising an event (usually captures are made in PointerPressed handlers)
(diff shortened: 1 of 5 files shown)


src/Uno.UI/UI/Xaml/UIElement.PointerCapture.cs
```diff
@@ -225,6 +243,12 @@ public bool ValidateAndUpdate(UIElement element, PointerRoutedEventArgs args, bo
 
 					return true;
 				}
+				else if (IsImplicitOnly)
+				{
+					// If the capture is implicit, we should not filter out events for child elements.
```
jeromelaban:
```suggestion
					// If the capture is implicit, we should not filter out events for children elements.
```

src/Uno.UI/UI/Xaml/UIElement.PointerCapture.cs
```diff
@@ -268,7 +292,7 @@ private void EnsureEffectiveCaptureState()
 
 					if (_nativeCaptureElement == null)
 					{
-						_nativeCaptureElement = _targets.Single().Key;
+						_nativeCaptureElement = _targets.Single().Value.NativeCaptureElement;
```
jeromelaban: Guaranteed to have a value ?

src/Uno.UI/UI/Xaml/UIElement.PointerCapture.cs
```diff
@@ -102,10 +108,15 @@ public bool TryAddTarget(UIElement element, PointerCaptureKind kind, PointerRout
 					{
 						return false;
 					}
+					else
+					{
+						// Add the new king to the target
```
davidjohnoliver:
```suggestion
						// Add the new kind to the target
```

src/Uno.UI/UI/Xaml/UIElement.PointerCapture.cs
dr1rrb: Yes, otherwise I fallback to the `element` (so my fix doesn't apply)

davidjohnoliver:
Seems like some tests are failing

image

PR #2876, opened 2020-03-24
[WASM] Ability to load image from bytes

SuperJMN:
GitHub Issue: https://github.com/unoplatform/uno/issues/2856

# Adds the ability to load images from bytes in WASM heads.

This is my first PR, so please, be patient. I don't expect this to be merged soon, because it surely will need refinement and maybe important changes

Basically, I've edited src/Uno.UI/UI/Xaml/Controls/Image/Image.wasm.cs to handle the use case when ImageSource.Stream isn't null. I don't know if this condition is enough to guarantee that the ImageSource should be loaded from the Stream, but it worked for me.

In case we are loading the image from the ImageSource.Stream, I read it using ImageSharp.
I know this introduces a 3rd party library, but I found no better option to load the pixel data, since I don't have enough knowledge about how this scenarios are handled inside Uno. Feel free to suggest/edit/criticize and I will refine the PR until it meets the conditions to be merged.

Thanks for your understanding :)


src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/ImageTests/LoadFromBytes.xaml.cs
```diff
@@ -0,0 +1,71 @@
...
+			InitializeComponent();
+			Loaded += OnLoaded;
+		}
+
+		private async void OnLoaded(object sender, RoutedEventArgs e)
+		{
+			var client = new HttpClient(new CorsByPassHandler());
+
+			var result = await client.GetAsync(_imageUrl);
+			var stream = await result.Content.ReadAsStreamAsync();
```
jeromelaban: Intead of doing this, add an embedded image (https://github.com/unoplatform/uno/tree/867d997852b05202530221494b68cdd41c40281c/src/SamplesApp/UITests.Shared/Assets) set its build action to EmbeddedREsource and load it like this : https://github.com/unoplatform/uno/blob/867d997852b05202530221494b68cdd41c40281c/src/Uno.UI.Tests/Windows_UI_Xaml_Markup/XamlReaderTests/Given_XamlReader.cs#L742

src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/ImageTests/LoadFromBytes.xaml.cs
```diff
@@ -0,0 +1,54 @@
...
+						where name.EndsWith("mslug.png")
+						select assembly.GetManifestResourceStream(name);
+
+			using (var stream = query.First())
+			{
+				MyImage.Source = await CreateImageSource(stream);
+			}
+		}
+
+		private async Task<ImageSource> CreateImageSource(Stream stream)
```
SuperJMN: Since it's quite common to create an ImageSource from a stream, would it be OK to make this reusable somewhere? Like an utility helper method?

src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/ImageTests/LoadFromBytes.xaml.cs
```diff
@@ -0,0 +1,54 @@
...
+			{
+				throw new ArgumentNullException(nameof(stream));
+			}
+
+			var image = new BitmapImage();
+
+#if WINDOWS_UWP
+			await image.SetSourceAsync(stream.AsRandomAccessStream()).AsTask();
+#else
+			var copy = stream.ToMemoryStream();
```
SuperJMN: I'm creating a copy here to avoid the disposal of the original input stream

src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/ImageTests/LoadFromBytes.xaml.cs
```diff
@@ -0,0 +1,54 @@
...
+		{
+			if (stream == null)
+			{
+				throw new ArgumentNullException(nameof(stream));
+			}
+
+			var image = new BitmapImage();
+
+#if WINDOWS_UWP
+			await image.SetSourceAsync(stream.AsRandomAccessStream()).AsTask();
```
SuperJMN: I'm not sure if this is the correct way to assign the stream in UWP

src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/ImageTests/LoadFromBytes.xaml.cs
jeromelaban: This is the sort of helpers we generally place in [Uno.Core](https://github.com/unoplatform/Uno.Core). The problem here is mostly about how to find the proper asset name.

That could be an extension like `var s = assembly.GetResourceStream(n => n.EndsWith("mslug.png")`.

src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/ImageTests/LoadFromBytes.xaml.cs
jeromelaban: This is the correct way for now, until we support AsRandomAccessStream correctly

src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/ImageTests/LoadFromBytes.xaml.cs
SuperJMN: That's OK! but I was referring to the method
```
private async Task<ImageSource> CreateImageSource(Stream stream)
```
I think it could be a good addition to centralize loading a `BitmapSource` from a `Stream`. But that could be in another PR, if you think it's OK.

jeromelaban: That'll be fine for now.

jeromelaban:
Can you also adjust your commit text to include [Wasm] in it ?

SuperJMN:
> Can you also adjust your commit text to include [Wasm] in it ?

Of course!

SuperJMN:
Is there any problem with the code? I've seen a check failing, but I don't understand what's wrong with it.

jeromelaban:
Looks like it works properly :) Nice work! https://dev.azure.com/uno-platform/Uno%20Platform/_build/results?buildId=13378&view=ms.vss-test-web.build-test-results-tab&runId=1056400&resultId=100692&paneView=attachments

image

PR #2892, opened 2020-03-27
fix(android): Fix for bug #2879

### LINKED ISSUE #2879: [Android] TextBlocks in a StackPanel are cropped when dimensions are changed
## Current behavior

TextBlocks in a StackPanel are cropped horizontally or vertically after dimensions are changed

## Expected behavior

TextBlocks in a StackPanel are not cropped when the dimensions are changed

Press the "Click Me" button multiple times to display the red stackpanel on the right (invisible at start)

Sample: [TextCropped.zip](https://github.com/unoplatform/uno/files/4382466/TextCropped.zip)

## Environment

Nuget Package: Uno.UI

Package Version(s): 2.2.0.dev-256

Affected platform(s):

- [ ] iOS
- [x] Android
- [ ] WebAssembly
- [ ] macOS
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

Visual Studio:

- [ ] 2017 (version: )
- [ ] 2019 (version: )
- [ ] for Mac (version: )

Relevant plugins:

- [ ] Resharper (version: )

Internal issue: 
https://nventive.visualstudio.com/Umbrella/_workitems/edit/177488

carldebilly:
# Bugfix

The problem was caused by a cache in Android layout system. Was appearing when the `DesiredSize` is required to layout an element, like in the `<StackPanel>`. In this case, the `Measure()` were not properly called at the right time, causing the panel (`<StackPanel>` here) to use the previous (wrong) measured/desired size.

The fix is to call `ForceLayout()`, causing Android to bypass the cache during the `Measure()` phase.

## PR Checklist

* Fix #2879 
* Fix https://nventive.visualstudio.com/Umbrella/_workitems/edit/177488

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/Layouter/Layouter.Android.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Layouter/Layouter.Android.cs
@@ -58,6 +58,8 @@ namespace Windows.UI.Xaml.Controls
 			var widthSpec = ViewHelper.SpecFromLogicalSize(slotSize.Width);
 			var heightSpec = ViewHelper.SpecFromLogicalSize(slotSize.Height);
 
+			view.ForceLayout(); // Bypass Android cache, to ensure the Child's Measure() is actually invoked.
+
 			MeasureChild(view, widthSpec, heightSpec);
 			
 			var ret = Uno.UI.Controls.BindableView.GetNativeMeasuredDimensionsFast(view)
--- a/src/Uno.UI/UI/Xaml/Controls/Layouter/Layouter.Android.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Layouter/Layouter.Android.cs
@@ -58,7 +58,20 @@ namespace Windows.UI.Xaml.Controls
 			var widthSpec = ViewHelper.SpecFromLogicalSize(slotSize.Width);
 			var heightSpec = ViewHelper.SpecFromLogicalSize(slotSize.Height);
 
-			view.ForceLayout(); // Bypass Android cache, to ensure the Child's Measure() is actually invoked.
+			var previousDesiredSize = DesiredChildSize(view);
+
+			if (previousDesiredSize.Width > slotSize.Width || previousDesiredSize.Height > slotSize.Height)
+			{
+				// Bypass Android cache, to ensure the Child's Measure() is actually invoked.
+				view.ForceLayout();
+
+				// We must do this here because we're using the MeasureSpecMode.AtMost mode to force Android to
+				// behave like the UWP's measure phase.
+
+				// We can't use MeasureSpecMode.Exactly because native controls would take all available space.
+
+				// Issue: https://github.com/unoplatform/uno/issues/2879
+			}
 
 			MeasureChild(view, widthSpec, heightSpec);
 			
--- a/src/Uno.UI/UI/Xaml/Controls/Layouter/Layouter.Android.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Layouter/Layouter.Android.cs
@@ -58,17 +58,13 @@ namespace Windows.UI.Xaml.Controls
 			var widthSpec = ViewHelper.SpecFromLogicalSize(slotSize.Width);
 			var heightSpec = ViewHelper.SpecFromLogicalSize(slotSize.Height);
 
-			var previousDesiredSize = DesiredChildSize(view);
-
-			if (previousDesiredSize.Width > slotSize.Width || previousDesiredSize.Height > slotSize.Height)
+			if (double.IsPositiveInfinity(slotSize.Width) || double.IsPositiveInfinity(slotSize.Height))
 			{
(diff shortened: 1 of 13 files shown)


src/Uno.UI/UI/Xaml/Controls/Layouter/Layouter.Android.cs
```diff
@@ -58,6 +58,8 @@ protected Size MeasureChildOverride(View view, Size slotSize)
 			var widthSpec = ViewHelper.SpecFromLogicalSize(slotSize.Width);
 			var heightSpec = ViewHelper.SpecFromLogicalSize(slotSize.Height);
 
+			view.ForceLayout(); // Bypass Android cache, to ensure the Child's Measure() is actually invoked.
```
davidjohnoliver: This is a pretty aggressive approach, which will negatively impact performance, potentially severely. It basically makes it so that the entire visual tree will remeasure whenever any view is invalidated.

Normally, the native Android cache should be invalidated whenever the available dimensions change, or when the child itself calls `RequestLayout()`. Can you say why that's not working or not sufficient here? Is there a way to apply this 'brute force' approach only in a more specific case?

carldebilly: That's the main difference between `.ForceLayout()` and `.RequestLayout()`.  The first one will _only_ bypass the cache of the control to force the `.Measure()` to be actually invoked. The `.RequestLayout()` will invalidate all the tree.

davidjohnoliver: But you're calling `ForceLayout()` recursively!

davidjohnoliver: Consider the following tree:

```
    A
  | | |
  v v v
  B C D
    |
    v
    E
```

1. Call `RequestLayout()` on E. This will invalidate E, C, and A, and prompt the window to schedule a layout pass.

2. A will be measured. (Refer to [this diagram](https://platform.uno/docs/articles/uno-development/Uno-UI-Layouting-Android.html) for more details.) It will measure its children, B, C, and D. Normally it should see that B and D are unchanged and skip them. This is implicitly handled by the native caching.

_But_ now you're calling `ForceLayout()` on B and D, so they'll be remeasured. And recursively, in the same way, the entire subtrees under them. So a huge chunk of the visual tree will be remeasured that doesn't need to be. This is why this is a significant performance concern.

jeromelaban: There's measure caching integrated in the layouter, but I'm wondering if the issue here is more around the fact that the TextBlock is not invalidating properly based on changes of the arrange values.

davidjohnoliver: IIRC for iOS Uno does some measure caching, for Android we mostly rely on the native caching.

carldebilly: @davidjohnoliver A better fix has been found. The invalidation will be done only when we know it's required.

davidjohnoliver: It's less extreme, but it could still be a performance hit in the right (wrong) circumstances, namely a view at the root of a large subtree that consistently returns a larger desired size than its available bounds.

And I still wonder if it's really necessary or if it's masking a deeper problem that could/should be fixed.

carldebilly: @davidjohnoliver @jeromelaban 
Ok, here's my final fix ;-)

src/Uno.UI/UI/Xaml/Controls/Layouter/Layouter.Android.cs
```diff
@@ -58,6 +58,17 @@ protected Size MeasureChildOverride(View view, Size slotSize)
 			var widthSpec = ViewHelper.SpecFromLogicalSize(slotSize.Width);
 			var heightSpec = ViewHelper.SpecFromLogicalSize(slotSize.Height);
 
+			if (double.IsPositiveInfinity(slotSize.Width) || double.IsPositiveInfinity(slotSize.Height))
+			{
+				// Bypass Android cache, to ensure the Child's Measure() is actually invoked.
+				view.ForceLayout();
+
+				// This could occurs when one of the dimension is _Infinite_: Android will cache the
```
jeromelaban:
```suggestion
				// This could occur when one of the dimension is _Infinite_: Android will cache the
```

jeromelaban:
Is this expected ? https://dev.azure.com/uno-platform/Uno%20Platform/_build/results?buildId=13431&view=ms.vss-test-web.build-test-results-tab&runId=1057572&resultId=100132&paneView=attachments

jeromelaban:

PR #2917, opened 2020-04-02
[macOS] Update measure/arrange passes

jeromelaban:
## PR Type

- Bugfix

Slow layout performance.

## What is the new behavior?

This change removes the reliance on NSView.NeedsLayout to invalidate the measure/arrange phases for self and parents. NeedsLayout is set to true when NSView.Frame is changed, causing a chain of multiple unneeded updates for the element and its parents.

## PR Checklist

## Other information

Related to #626

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.macOS.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.macOS.cs
@@ -128,7 +128,7 @@ namespace Windows.UI.Xaml.Controls
 				{
 					// This measures the height correctly, even if the Text is null or empty
 					// This matches Windows where empty TextBlocks still have a height (especially useful when measuring ListView items with no DataContext)
-					var font = NSFontHelper.TryGetFont((float)FontSize, FontWeight, FontStyle, FontFamily);
+					var font = NSFontHelper.TryGetFont((float)FontSize*2, FontWeight, FontStyle, FontFamily);
 
 					var str = new NSAttributedString(Text, font);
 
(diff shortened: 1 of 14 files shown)


src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.macOS.cs
```diff
@@ -128,7 +128,7 @@ protected override Size MeasureOverride(Size size)
 				{
 					// This measures the height correctly, even if the Text is null or empty
 					// This matches Windows where empty TextBlocks still have a height (especially useful when measuring ListView items with no DataContext)
-					var font = NSFontHelper.TryGetFont((float)FontSize, FontWeight, FontStyle, FontFamily);
+					var font = NSFontHelper.TryGetFont((float)FontSize*2, FontWeight, FontStyle, FontFamily);
```
davidjohnoliver, line 131: Hmm... I get why, but is this the good fix?

Either way, needs a bit more green.

src/Uno.UI/UI/Xaml/FrameworkElement.macOS.cs
```diff
@@ -33,15 +33,38 @@ public override bool NeedsLayout
...
-				RequiresArrange = true;
-
 				if (ShouldInterceptInvalidate)
 				{
 					return;
 				}
+			}
+		}
+
+		protected internal override void OnInvalidateMeasure()
```
davidjohnoliver, line 43: Does this logic need to be in the `IFrameworkElementHelper` method? Otherwise it seems like `IFrameworkElements` won't invalidate their parent correctly.

src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.macOS.cs
jeromelaban: No green for now. It'll be addressed in another PR.

src/Uno.UI/UI/Xaml/FrameworkElement.macOS.cs
jeromelaban: Yes, it'll most likely fail. Will be addressed in a later PR as well.

PR #1766, opened 2019-10-08
Styles and Xaml resources compatibility

davidjohnoliver:
GitHub Issue (If applicable): fixes #119 , fixes #1335, fixes #55, fixes #49, fixes #55, fixes #96, fixes #371, fixes #833, fixes #1427, fixes #3216, fixes #1420, fixes #2404

Related: https://github.com/unoplatform/uno/issues/2744

## PR Type

- Bugfix
- Feature

What kind of change does this PR introduce?

## What is the current behavior?

 - Uno only supports global-level resources, including implicit styles

## What is the new behavior?

Numerous compatibility fixes:

 - Resources can be defined at any level of the visual tree
 - Resources are resolved by scope, including implicit styles
 - default (framework) styles are additive with implicit/explicit styles
 - Resources property is now correctly populated, resources can be accessed/modified from code-behind
 - theme resources can be dynamically updated when theme changes
 - `DefaultStyleKey` is now respected
 - custom subclasses will now use the default style of the parent framework control unless `DefaultStyleKey` is overridden

## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Tested code with current [supported SDKs](https://github.com/unoplatform/uno/blob/master/README.md#uno-features)
- [ ] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)
- [x] [Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [ ] [Wasm UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md#running-the-webassembly-ui-tests-snapshots) are not showing unexpected any differences. Validate PR `Screenshots Compare Test Run` results.
- [ ] ~~Contains **NO** breaking changes~~
- [ ] Updated the [Release Notes](https://github.com/unoplatform/uno/tree/master/doc/ReleaseNotes)
- [x] Associated with an issue (GitHub or internal)

## Other information

Internal Issue (If applicable):

PR #2932, opened 2020-04-04
[WIP] feat: Theme change notification

### LINKED ISSUE #1785: React to theme changes at runtime

If user does not specify default theme using `RequestedTheme` in App.xaml, app should change its theme even in runtime when OS theme is changed.

## Why is this needed:

To make the app look at home during the day and during the night. ☀🌛

## For which Platform:

- [x] iOS
- [x] Android
- [x] WebAssembly
- [ ] Windows

Functionality depends on @davidjohnoliver's awesome PR #1766

MartinZikmund:
GitHub Issue (If applicable): fixes #1785

## PR Type

- Feature

Support for theme change at runtime is missing.

## What is the new behavior?

Theme change at runtime is available on all platforms.

## PR Checklist


MartinZikmund:
This branch is based on `feature/styles-compat`

src/SamplesApp/SamplesApp.Droid/MainActivity.cs
```diff
@@ -9,7 +9,7 @@ namespace SamplesApp.Droid
 {
 	[Activity(
 			MainLauncher = true,
-			ConfigurationChanges = ConfigChanges.Orientation | ConfigChanges.ScreenSize,
+			ConfigurationChanges = ConfigChanges.Orientation | ConfigChanges.ScreenSize | ConfigChanges.UiMode,
```
davidjohnoliver, line 12: Should this be changed in the solution template? Should also be documented somewhere for existing apps wanting to implement dark mode.

src/Uno.UI/UI/Xaml/Application.cs
```diff
@@ -75,6 +86,16 @@ public ApplicationTheme RequestedTheme
...
 		public event UnhandledExceptionEventHandler UnhandledException;
 
+		public void OnSystemThemeChanged()
+		{
+			// if user overrides theme, don't apply system theme
+			if (!_themeSetExplicitly)
+			{
+				var theme = GetDefaultSystemTheme();
+				SetRequestedTheme(theme);
+			}
```
davidjohnoliver, line 96: Should `_themeSetExplicitly` be reset to `false` here?

src/Uno.UI/UI/Xaml/Application.cs
```diff
@@ -45,21 +46,31 @@ public static class TraceProvider
...
 
 		public ApplicationTheme RequestedTheme
 		{
-			get => _requestedTheme ?? (_requestedTheme = GetDefaultSystemTheme()).Value;
+			get
+			{
+				if (_requestedTheme != null)
+				{
+					_requestedTheme = GetDefaultSystemTheme();
+					ObserveSystemThemeChanges();
```
davidjohnoliver, line 61: Is this the right place? I think we want to observe system theme changes unconditionally, and it also seems like this would add a new listener every time `RequestedTheme` is accessed.

src/SamplesApp/SamplesApp.Droid/MainActivity.cs
MartinZikmund: That is true. I will update the template and write a doc for it

src/Uno.UI/UI/Xaml/Application.cs
MartinZikmund: The `_themeSetExplicitly` variable is set to `true` only if user sets the theme in `Application.xaml`, and after it becomes `true`, it cannot become `false` anymore (the app either reacts to system theme automatically, or the developer is fully in control)

src/Uno.UI/UI/Xaml/Application.cs
MartinZikmund: The `_requestedTheme` will be `null` only initially, and after that the `if` block will never be executed again. If the user sets the theme explicitly, we no longer need to observe system theme changes, as they will never be applied (once developer sets the `RequestedTheme` once, it is forever in their hands, and will never again react to system)

davidjohnoliver: Ok, that makes sense.

I think the condition is wrong then - shouldn't it be `if (_requestedTheme == null)`?

src/SamplesApp/SamplesApp.Droid/MainActivity.cs
MartinZikmund: Done

MartinZikmund:
@davidjohnoliver Due to the force-pushes in your branch this one is now a bit messy, can you cherry pick the changes manually? I have squashed all my changes into one commit: `2f670f0`

davidjohnoliver:
> @davidjohnoliver Due to the force-pushes in your branch this one is now a bit messy, can you cherry pick the changes manually? I have squashed all my changes into one commit: `2f670f0`

Thanks Martin! I'll cherry-pick it to my branch.

MartinZikmund:
Closing, will be part of  #1766

Issue #2240, opened 2019-12-02
Breaking changes rollup


jeromelaban:
This issue is a meta-issue for tracking breaking changes that need to be made as a batch to avoid making too many breaking changes for consumers of Uno.

This issue does not list the larger changes needed to be made to align for the removal of `IFrameworkElement`.

- [ ] https://github.com/unoplatform/uno/issues/2239
- [ ] https://github.com/unoplatform/uno/blob/38d3c0ded22601d79b5a1a006b63b37915912e0a/src/Uno.UI/UI/Xaml/Controls/UserControl/UserControl.iOSmacOS.cs#L29
- [ ] ToggleButton OnIsCheckedChanged

MartinZikmund:
Many `Geolocator` methods currently use `Task` as return type instead of `IAsyncOperation` (see for example `Task<Geoposition> GetGeopositionAsync()` vs `IAsyncOperation<Geoposition> GetGeopositionAsync()`

Youssef1313:
>  Color default constructor should allow writing `Color a;` and not `Color a = new Color()`

@jeromelaban @MartinZikmund Any more info on that? Both forms works on UWP and in Uno.

jeromelaban:
> > Color default constructor should allow writing `Color a;` and not `Color a = new Color()`
> 
> @jeromelaban @MartinZikmund Any more info on that? Both forms works on UWP and in Uno.

It's a change that was needed somewhere in WCT, where the `Color a;` syntax was not working. It could be that we've changed something since this issue was logged.

Youssef1313:
> * [ ]  `ContentPresenter.CornerRadiusProperty` is a field
> *

Also Control.FocusStateProperty, Control.IsTabStopProperty, StackLayout.SpacingProperty, StackLayout.DisableVirtualizationProperty

PR #2936, opened 2020-04-05
feat: OpenStreamForReadAsync without token

pkar70:

## PR Type

- Feature

There is method `OpenStreamForReadAsync(CancellationToken ct)`, but no method `OpenStreamForReadAsync()` .

## What is the new behavior?

Method `OpenStreamForReadAsync()` is implemented.

## PR Checklist

## Other information
I hope I would not be required to create test for this method.

### DIFF
--- a/src/Uno.UWP/Storage/StorageFile.cs
+++ b/src/Uno.UWP/Storage/StorageFile.cs
@@ -130,8 +130,9 @@ namespace Windows.Storage
             public IAsyncOperationWithProgress<uint, uint> WriteAsync(IBuffer buffer) => throw new NotImplementedException();
         }
 
+		public IAsyncOperation<Stream> OpenStreamForReadAsync() => OpenStreamForReadAsync(CancellationToken.None).AsAsyncOperation<Stream>();
 
-        public async Task<Stream> OpenStreamForReadAsync(CancellationToken ct)
+		public async Task<Stream> OpenStreamForReadAsync(CancellationToken ct)
 		{
 			switch (Scheme)
 			{


jeromelaban:
Thanks for your PR! 

This is unfortunately not a valid change, as the method does not exist on the `StorageFile` type, and the original one that takes a token should not have been there either (I'll reference it in #2240 so we can remove it).

Note that this method already exists here as an extension:
https://github.com/unoplatform/uno/blob/6927750e398683a01db98d97061ff0413467c1ed/src/Uno.UWP/System.IO/WindowsRuntimeStorageExtensions.cs#L10

pkar70:
If this method exist, why Visual Studio underline OpenStreamForReadAsync with info that method is not implemented in Uno?

jeromelaban:
You added a new method, not a not-implemented method. 

The one you're seeing is probably another method. If you go the definition of that not-implemented method you're seeing, where does it lead you to ?

pkar70:
When I use not my Uno compilation but 2.2.0-dev380, Visual Studio wants cancelationtoken parameter:
image
it leads to:
image

jeromelaban:
Yes, that's what I meant in my original comment. The method you're adding does not exist in the API, if you want to use that method without a `CancellationToken`, you need to add a using to use this method:

https://github.com/unoplatform/uno/blob/6927750e398683a01db98d97061ff0413467c1ed/src/Uno.UWP/System.IO/WindowsRuntimeStorageExtensions.cs#L10

pkar70:
But I want my app code build in standard UWP, and in Android. Without changing code.
I don't care how method is implemented, as extension to IStorageFile, or inside StorageFile. I just want to use this method.
I don't know why 'your method' is not visible by Visual Studio, and why it want me to add cancellation token parameter.

jeromelaban:
I understand that you want to use a method that has this signature, yet the change you're proposing is not the one that will fix this. The method is located somewhere else.

Can you provide a repro case that works on UWP, and that demonstrates the issue in Uno ?

jeromelaban:
Can you create a sample that I can validate ? A screenshot is not enough.

pkar70:
I got it.
When I add `using System.IO;`, OpenStreamForReadAsync() is not underlined. 
And when I removed `OpenStreamForReadAsync(CancellationToken ct)` from StorageFile.cs , Visual Studio doesnt complain about missing CancellationToken, and suggest adding `using System.IO;`.

PR #3010, opened 2020-04-17
fix: [Wasm] Fixes #2834 -Symbol icon Home not shown

### LINKED ISSUE #2834: Symbol Icon Home not showed in WASM NavigationViewItem
## Current behavior

Home Symbol Icon in NavigationItem is showed in UWP but not WASM. Other Symbol Icons (not Home) are showed in both platforms: UWP and Wasm.

## Expected behavior

Home Symbol Icon showed in WASM Platform.

Show a NavigationView control as Content of Page with several NavigationItem and Home SymbolIcon in one Item.

    <NavigationView>
        <NavigationView.MenuItems>
            <NavigationViewItem Icon="Home">
                <NavigationViewItem Icon="Accept"/>
            </NavigationViewItem>
        </NavigationView.MenuItems>
    </NavigationView>

UWP local machine start will show Home and Accept SymbolIcons. WASM IIS Express start will show Accept SymbolIncon but blank instead of Home SymbolIcon.

Affected platform(s):

(iOS and Android not tested)

- [ ] iOS
- [ ] Android
- [* ] WebAssembly
- [ ] macOS
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

Visual Studio:

- [ *] 2017 (15.9.17 )

kakone:
GitHub Issue : #2834

## PR Type
- Bugfix

The symbol icon **Home** is not shown in WASM (**Home** is set as the default symbol in the **Symbol** dependency property but the inner **FontIcon.Glyph** property is not set).

## What is the new behavior?

The symbol icon **Home** is now shown in WASM, and the default symbol is now **Emoji** (like in UWP **SymbolIcon** class).

## PR Checklist

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/SymbolIcon/SymbolIcon.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/SymbolIcon/SymbolIcon.cs
@@ -12,15 +12,14 @@ namespace Windows.UI.Xaml.Controls
 	{
 		private FontIcon _icon;
 
-		public SymbolIcon()
+		public SymbolIcon() : this(Symbol.Emoji)
 		{
-			_icon = new FontIcon();
-
-			AddIconElementView(_icon);
 		}
 
-		public SymbolIcon(Symbol symbol) : this()
+		public SymbolIcon(Symbol symbol)
 		{
+			_icon = new FontIcon();
+			AddIconElementView(_icon);
 			_icon.Glyph = new string((char)symbol, 1);
 		}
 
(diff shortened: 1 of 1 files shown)


jeromelaban:
Thank you for your contribution!

It's not the proper change, though. The problem here is the default font we're using which does not have the proper glyph. One possible solution would be to map individual characters to their proper representations, but that would be font-specific.

The best way to fix this would be to use the Segoe Symbols font from windows, but it's not licensed properly, another way would be to create a similar font just for Uno.

Let's track it with this issue: https://github.com/unoplatform/uno/issues/3011

kakone:
> The problem here is the default font we're using which does not have the proper glyph

No, it's not the problem here. The used font has the good Home glyph with the same code as Segoe MDL 2. It's really a bug in the SymbolIcon class.

jeromelaban:
I see! Indeed, the default value is not taken into account properly. Apologies for the misreading :)

Two things, we're not using the ReleaseNotes file anymore, we're using conventional commits, as stated in the default PR template. You can view the format here: https://github.com/unoplatform/uno/blob/master/.github/pull_request_template.md

kakone:
The bug is as follows : **Symbol.Home** is set as the defaut value of the **Symbol** dependency property :
```csharp
public static DependencyProperty SymbolProperty { get; } = DependencyProperty.Register("Symbol", typeof(Symbol), typeof(SymbolIcon), new PropertyMetadata(Symbol.Home, OnSymbolChanged));
```
But the underlying FontIcon.Glyph is not set with **Symbol.Home** by default, that's why the icon is not shown in WASM.
When the user set the **Symbol** property to **Symbol.Home**, the **OnSymbolChanged** method is not called because **Symbol.Home** is the default value of the dependency property.
So, this PR fixes this issue and also set the default value to **Symbol.Emoji** (same default value as the **SymbolIcon** UWP class).

kakone:
> we're not using the ReleaseNotes file anymore, we're using conventional commit

OK, I reverted the ReleaseNotes file.

jeromelaban:
You'll need to follow the conventional commits specification as well: https://www.conventionalcommits.org/en/v1.0.0-beta.2/#summary

jeromelaban:
Thanks for your contribution!

PR #2898, opened 2020-03-29
internal: Android working permission helper

pkar70:
GitHub Issue (If applicable): #

## PR Type

What kind of change does this PR introduce?
- Bugfix
- Feature
- Refactoring (no functional changes, no api changes)
All of above, and - at same time - none of above :)

## What is the current behavior?
Current implementation of Android permission helper functions doesn't work.
Currently there is no possibility to invoke system dialog asking for permission from Uno code (e.g. from RequestAccessAsync methods).
So, PR#2024 (geolocator) is labelled as 'help needed' (to split code into levels - as was said in Lobby or in some comments), also PR#2544 (contacts), PR#2757 (calllog), and soon-to-be PR (SMS access) cannot be fully implemented (some code have to be included in app code, not inside Uno).

## What is the new behavior?
New helper methods works inside Uno, and can invoke system dialog asking for permission.

## PR Checklist

Please check if your PR fulfills the following requirements:
- [ ] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)
- [ ] [Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [ ] Validated PR `Screenshots Compare Test Run` results.
- [X] Contains **NO** breaking changes
- [ ] Associated with an issue (GitHub or internal) and uses the [automatic close keywords](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue).
- Commits must be following the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) specification.

## Other information
 I made it as separate PR, to be used in all four PRs mentioned above, as code is common (and I don't want contacts-PR be dependend on e.g. geolocator PR). Also, it can be used from other Uno PRs and from user code.

 I don't edit/modify current Android methods in permissionhelper.cs, as they can be somewhere used (structure of these methods are cumbersome, too complicated for me). 

Internal Issue (If applicable):

Issue #2757, opened 2020-03-08
MIDI in/out


MartinZikmund:
## What would you like to be added:

Support for MIDI Device input/output

## Why is this needed:

Music apps

## For which Platform:

- [x] iOS
- [x] Android
- [x] WebAssembly
- [ ] WebAssembly renderers for Xamarin.Forms
- [x] macOS
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

MartinZikmund:
I have this in a draft phase with first working examples on Android and WASM. Hope to finish during March/April.

PR #2544, opened 2020-01-28
Contacts and android reader

pkar70:
GitHub Issue (If applicable): unoplatform/uno-private#1439

## PR Type

What kind of change does this PR introduce?
- Feature

## What is the current behavior?
Nothing in Windows.ApplicationModel.Contacts is implemented.

## What is the new behavior?
Parts of this namespace is implemented. See details in "Other information" section.

## PR Checklist

Please check if your PR fulfills the following requirements:

- [X] Tested code with current [supported SDKs](https://github.com/unoplatform/uno/blob/master/README.md#uno-features)
- [ ] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)
- [X] [Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [ ] [Wasm UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md#running-the-webassembly-ui-tests-snapshots) are not showing unexpected any differences. Validate PR `Screenshots Compare Test Run` results.
- [X] Contains **NO** breaking changes
- [X] Updated the [Release Notes](https://github.com/unoplatform/uno/tree/master/doc/ReleaseNotes)
- [ ] Associated with an issue (GitHub or internal)

## Other information

Code, for Android only:
- ContactBatch
- ContactManager
- ContactReader
- ContactStore

Part of my apps:
* https://github.com/pkar70/CallStat
* https://github.com/pkar70/BackupSMS

Internal Issue (If applicable):

PR #2024, opened 2019-11-05
more functional Geolocator.RequestAccessAsync()

pkar70:
GitHub Issue (If applicable): #1967

## PR Type
What kind of change does this PR introduce?
- Bugfix

## What is the current behavior?
RequestAccessAsync() => GeolocationAccessStatus.Allowed;

## What is the new behavior?
Above is correct only on API <23. 
Now, it RequestAccessAsync() returns real GeolocationAccessStatus.

## PR Checklist
- [X] Tested code with current [supported SDKs](https://github.com/unoplatform/uno/blob/master/README.md#uno-features)
- [ ] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)
- [ ] [Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [ ] [Wasm UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md#running-the-webassembly-ui-tests-snapshots) are not showing unexpected any differences. Validate PR `Screenshots Compare Test Run` results.
- [X] Contains **NO** breaking changes
- [X] Updated the [Release Notes](https://github.com/unoplatform/uno/tree/master/doc/ReleaseNotes)
- [X] Associated with an issue (GitHub or internal)

## Other information
Recreated PR#1998 to have one commit.

This should be only first step in rewriting RequestAccessAsync() - it should behave as on Windows, i.e. it should display system dialog asking for permission. But this is beyond my Android knowledge.

I don't know how to make test for this PR... Not from programmers point of view, but no idea how to test this change.

Internal Issue (If applicable):

PR #3025, opened 2020-04-20
feat: Add PermissionHelperAndroid to support additional permissions validation

pkar70:

## PR Type

- Bugfix
- Feature
All of above, and - at same time - none of above :)

Current implementation of Android permission helper functions doesn't work.
Currently there is no possibility to invoke system dialog asking for permission from Uno code (e.g. from RequestAccessAsync methods).
So, PR #2024 (geolocator) is labelled as 'help needed' (to split code into levels - as was said in Lobby or in some comments), also PR #2544 (contacts), PR #2757 (calllog), and soon-to-be PR (SMS access) cannot be fully implemented (some code have to be included in app code, not inside Uno).

## What is the new behavior?
New helper methods works inside Uno, and can invoke system dialog asking for permission.

## PR Checklist

## Other information
 I made it as separate PR, to be used in all four PRs mentioned above, as code is common (and I don't want contacts-PR be dependend on e.g. geolocator PR). Also, it can be used from other Uno PRs and from user code.

 I don't edit/modify current Android methods in permissionhelper.cs, as they can be somewhere used (structure of these methods are cumbersome, too complicated for me). 

 Reopened #2898/#2964


pkar70:
I created new branch from unoplatform/master, reapply all changes from previous PR, push it.... and got (on web) conflict in `src/SamplesApp/UITests.Shared/UITests.Shared.projitems`. So, I edited this file on web interface.

jeromelaban (commented):
Thanks. I've tested your PR, and this is what the sample gives:

image

Creating an activity to validate the permissions is not the right way to do this, as it pops up a new full screen activity to ask for permissions.

Also, I noticed that the existing `PermissionsHelper` already provides the infrastructure to validate permissions, so I wonder why you needed another one to do so. What's missing in here ?

https://github.com/unoplatform/uno/blob/6e8a9273acbd6fcab385fc36a3d29cb48295f824/src/Uno.UWP/Extensions/PermissionsHelper.cs#L20

Finally, you need to apply the code formatting rules of the Uno project. The best way to do so is to use the "cleanup" tool from VS, at the bottom of the code editor:

image

pkar70:
Seems like test environment is different than normal app environment (as this code is working when I call it from my uno-app). Android.App.Application.Context is null?

Yes, there is some permission helper implementation, but it doesn't work. I tried to use it when I tried to use geolocation. As effects of my work on geolocation, I created PR for geolocation (#1998, Oct 31, 2019) - simple bug correction in Geolocator.RequestAccessAsync, than #2024 (Nov 5, 2019) - much more complete Geolocator implementation, and full code I'm using in my app (KrakTram, public transport in Kraków) - "// this code would not compile here - but it compile in your own app.". Than you add 'help needed' tag for it, to split this into levels (I don't know what you exactly mean by this levels), than using your suggestion I open another PR with all code that not require splitting into levels to #2536 (Jan 25). Many comments/code changes from peer review was done in this process :)

Searching how to implement OpenFilePicker, I encounter example with defining Activity (not in manifest, but as [Activity flag). I use this general idea to make permission subroutines to my build of Uno, reworked geolocator, SMS messages, call log, known folders, etc. to use this subroutines instead of implementing it in all of them (as you commented in #2773, on 23rd March).

jeromelaban:
What does not 

> Seems like test environment is different than normal app environment (as this code is working when I call it from my uno-app). Android.App.Application.Context is null?

No the environment is not null, but the activity is not created with the proper flags. In any case, this is not a good solution to this problem.

> Yes, there is some permission helper implementation, but it doesn't work. I tried to use it when I tried to use geolocation. 

Can you remind me what does not work ? The wiring up should be enough for the permissions request to be raised.

pkar70:
Ok, environment is not null, but maybe standard Uno-Android-App is wrapped in Activity, and test is not wrapped? I only guessing, because when I call AndroidPermissionAsync or MissingPermissions from my app, it works. But in test, it throws exception (screenshot you included shows it) about StartActivity outside Activity, so in test it is outside Activity, and in app - it is inside Activity.

Previous implementation doesn't work, i.e. it doesn't display any dialog.

jeromelaban:
The point of the existing implementation is to forward the calls from the Uno layer to the Uno.UI layer. If calls are missing, let's add them to make the missing permissions work.

pkar70:
The point of existing implementation is - it doesn't work. And never did.
If I remember correctly, to show system dialog, code has to know Activity. But this is inaccessible in Uno.UWP code. All my previous implementation of permission-related code (calllog, SMS, etc.) uses RequestAsync inside app (where I have Uno.UI.BaseActivity.Current accessible), and now (#3025) this requirement is worked-around.
As I wrote in opening PR note: "I don't edit/modify current Android methods in permissionhelper.cs, as they can be somewhere used (structure of these methods are cumbersome, too complicated for me)."

If you don't want to accept this PR, you can change current permission helper methods, and mark it as 'breaking change'. And I will simply wait till this would be done, and then create PR for accessing SMS log (read/write).
But I need methods with required and optional permissions, both as list of permissions and not simply one permission (as e.g. getting call log requires both contact and callog permission, and it should be done in one system call and not two).

CalendarDatePicker PR was ready in November, now support for it 'is planned' (six months later), so also callog, contacts, SMS, knownfolders can wait months till someone else (re)implement it.

jeromelaban:
Thank you for the explanation. The current implementation you're proposing is not adequate as it starts a new activity, and should not. The current implementation of the permission helper does have access to the current activity through this method:

https://github.com/unoplatform/uno/blob/6e8a9273acbd6fcab385fc36a3d29cb48295f824/src/Uno.UWP/Extensions/PermissionsHelper.cs#L25

And are currently used by existing parts of Uno (camera capture being one). Are those not working for you as well ?

As for the other PRs you're mentioning, many of those PRs are missing the samples we requested.

pkar70:
I didn't try camera, I tried to use current implementation for GPS. As far as I remember (it was in november), current implementation doesn't show system dialog, so I was forced to make working implementation by myself. At first, code works only in app, not in Uno, as it has not access to activity (or something like that, my memory is not so good to remember such details after half a year).
But after I discover way to show FileOpenPicker, with own activity for it, I re-implemented RequestAccess/PermissionHelper with own activity, and it works also when placed within Uno.

Why making Activity is wrong?

pkar70:
I created empty Uno app. Shared mainpage.xaml has only TextBlock and Button, button click handler exactly as in test file in this PR. As Uno.UI, I used my build (with this PR, #3025, included). Build/deploy to real device, with Android 8.1 Go.

When I click button without adding permission to Manifest, it correctly shows "MissingPermissions returned NULL, check if app manifest includes AccessFineLocation". And when I click it with declaring permission in Manifest, I got system dialog asking grant/deny permission.

So, there is some difference between running code as UserControl/inside test and as App. And this is very serious problem! Maybe some other tests also fail, both false positive and false negative...

jeromelaban:
As I mentioned above, making a activity pops up an opaque activity above the application. This is not an acceptable behavior for an application.

As for the existing permissions helper, it works properly. I've adjusted the GeoLocator in https://github.com/unoplatform/uno/pull/3064 so that it requests the permissions properly.

Please take a look at the change to use it in your PR. If you need more features in the permissions helper, add it to the existing structure.

pkar70:
Geolocator for Android should be rewritten - see my PRs for this.  Not only asking for permission.
And when I tried to use current implementation of PermissionsHelper.TryGetFineLocationPermission, it doesn't work. Maybe I make some mistake in calling it, I don't remember after half a year how it failed.
Also, at least in my apps ported to Android, I need many permissions, not one of them. So, methods in this PR have permission list as parameter.
To emulate call history, I need call log and contacts. Same for chat messages.

jeromelaban:
For the GeoLocation, try again, this is what the PR I mentioned does.

pkar70:
Correction of false implementation of geolocator RequestAccessAsync was in PR#1998, Oct 31, 2019. At this time, no one comments that current implementation works...

Before I change any of my PRs, I have to check again if current implementation of permissionhelper works. Maybe extend it to methods as in this PR (with list of permission, and not single permission on input). I don't expect that your PR #3064 would have any tests... Besides, it probably should be in permissionhelper.android.cs, not permissionhelper.cs file, as it is only for Android.

jeromelaban:
The PR #3064 already has a test, and it was failing on android: https://github.com/unoplatform/uno/blob/master/src/SamplesApp/UITests.Shared/Windows_Devices/GeolocatorTests.xaml

pkar70:
Ok, I tested your version. And yes, it shows system permission dialog. This dialog is same as in my version (same look of whole screen, same ability to interact - either answer in this dialog, or open task manager and kill app). 
Your version uses ContextCompat.CheckSelfPermission , without tests if permission is declared in Manifest (my version distinguish these two situations: no permission declaration, and permission denied).
Also, CheckPermission requires CancellationToken  - what is function of this parameter?

jeromelaban:
That what I mean by modifying the PermissionsHelper to add the missing parts. If there is a need for multi-permissions check, they you can add support for it.

The CancellationToken is generally provided (or should be if not) by the AsyncAction builder, but in https://github.com/unoplatform/uno/pull/3064 it cannot be provided because the API signatures are not the right ones (and will need to be adjusted).

pkar70:
Can testing permission be in Uno.UWP, without call to Uno.UI? This call is unncesessary.
Your version of check permission have two possible outputs: yes/no. My version has three outputs: yes, no, error (i.e.: not defined in Manifest).

jeromelaban:
No it cannot, it has to be delegated to Uno.UI, but that's not an issue.

It's also possible to change the output to be an enum instead of a boolean. We can add a `Task<PermissionsCheckResult> CheckPermissionsExtended()` and adjust the delegation method to return the appropriate values.

pkar70:
I mean permission check, not system dialog asking for permission.
CheckPermission method in your way (check if permission is granted), or MissingPermission in my way (check Manifest, and check  if permission is granted), can work inside Uno.UWP.

Why "it has to be delegated to Uno.UI"?

jeromelaban:
It has to be delegated to Uno.UI because that's where the `BaseActivity` is, the one that avoids popping up a new activity to check permissions. See here :

https://github.com/unoplatform/uno/blob/6e8a9273acbd6fcab385fc36a3d29cb48295f824/src/Uno.UI/Extensions/PermissionsHelper.Android.cs#L22

pkar70:
BaseActivity is not used in permission checking at all.
To check Manifest, and check if permission is granted, you need only Android.App.Application.Context - accessible in Uno.UWP.
See this PR (3025) - method MissingPermissions (splitted into many submethods, as you requested).

jeromelaban:
BaseActivity is used in the permissions checking, please take a look at this line:

https://github.com/unoplatform/uno/blob/6e8a9273acbd6fcab385fc36a3d29cb48295f824/src/Uno.UI/Extensions/PermissionsHelper.Android.cs#L52

jeromelaban:
And the permission request:

https://github.com/unoplatform/uno/blob/6e8a9273acbd6fcab385fc36a3d29cb48295f824/src/Uno.UI/Extensions/PermissionsHelper.Android.cs#L80

pkar70:
Yes, you use BaseActivity. But my version uses Android.Support.V4.Content.ContextCompat.CheckSelfPermission(Android.App.Application.Context, permission). See code in this PR.

pkar70:
My question "Can testing permission be in Uno.UWP, without call to Uno.UI?" is not about technical possibility (I know that it can be done), but "Would you accept such code?".

jeromelaban:
The relevant code should be at the same place, whether it's checking or requesting, even if Application.Context is available in the Uno assembly. If you need to makes changes, change the delegates so that it fits your extended scenario.

pkar70:
Ok. So, if required methods has to be hard to create (and not as simple as possible), I would not have enough time for this in near future... Pity.

jeromelaban:
It's not hard to create, it's about moving code from one layer to another. Otherwise, use the methods as is in your existing PRs even if it's asking one permissions at a time.

pkar70:
Would it be ok? With body of main methods in Uno.UI, and called from several methods in Uno.UWP?

src/Uno.UI/Extensions/PermissionsHelper.Android.cs
```diff
@@ -19,11 +19,185 @@ public static void Initialize()
...
+						allPermissions.Add(permission);
+					}
+				}
+			}
+
+			return allPermissions;
+		}
+
+		private static List<string> PermissionToAsk(IList<string> allPermissions)
+		{// return list of permission, with all permission from input list that are not already granted
```
jeromelaban, line 79: This is a comment that goes into XML comments above the method

src/Uno.UI/Extensions/PermissionsHelper.Android.cs
```diff
@@ -19,11 +19,185 @@ public static void Initialize()
 					.CreateAsync<string, bool>(TryGetPermissionCore)
 					.LockInvocation(InvocationLockingMode.Share);
 
-			Windows.Extensions.PermissionsHelper.Initialize(getPermission, CheckPermission);
-		}
+			Windows.Extensions.PermissionsHelper.Initialize(getPermission, CheckPermission, MissingPermissions, TryAskPermissionAsync);
+		}
+
+		private static int _permissionRequest;
+		#region "submethods for MissingPermissions"
```
jeromelaban, line 26: There is no need for regions

src/Uno.UI/Extensions/PermissionsHelper.Android.cs
```diff
@@ -19,11 +19,185 @@ public static void Initialize()
...
-			Windows.Extensions.PermissionsHelper.Initialize(getPermission, CheckPermission);
-		}
+			Windows.Extensions.PermissionsHelper.Initialize(getPermission, CheckPermission, MissingPermissions, TryAskPermissionAsync);
+		}
+
+		private static int _permissionRequest;
+		#region "submethods for MissingPermissions"
+		private static IList<string> GetManifestPermissions()
+		{
+			// get all permission declared in Manifest (can be null)
```
jeromelaban, line 29: This goes into the xml comment

src/Uno.UI/Extensions/PermissionsHelper.Android.cs
```diff
@@ -19,11 +19,185 @@ public static void Initialize()
...
+		}
+
+		private static List<string> PermissionWeWant(IList<string> manifestPermissions, string[] requiredPermissions, string[] optionalPermissions)
+		{
+			// prepare list of all permissions
+			var allPermissions = new List<string>();
+			if (requiredPermissions != null)
+				allPermissions.AddRange(requiredPermissions.ToList());
+
+			// add all optional permission, found in Manifest
```
jeromelaban, line 63:
```suggestion
			// add all optional permissions, found in Manifest

```

src/Uno.UI/Extensions/PermissionsHelper.Android.cs
```diff
@@ -19,11 +19,185 @@ public static void Initialize()
...
+				{
+					return false;
+				}
+			}
+
+			return true;
+
+		}
+
+		private static List<string> PermissionWeWant(IList<string> manifestPermissions, string[] requiredPermissions, string[] optionalPermissions)
```
jeromelaban, line 56:
```suggestion
		private static List<string> GetRequestedPermissions(IList<string> manifestPermissions, string[] requiredPermissions, string[] optionalPermissions)

```

src/Uno.UI/Extensions/PermissionsHelper.Android.cs
```diff
@@ -19,11 +19,185 @@ public static void Initialize()
...
+					{
+						allPermissions.Add(permission);
+					}
+				}
+			}
+
+			return allPermissions;
+		}
+
+		private static List<string> PermissionToAsk(IList<string> allPermissions)
```
jeromelaban, line 78:
```suggestion
		private static List<string> GetPermissionsToRequest(IList<string> allPermissions)

```

src/Uno.UI/Extensions/PermissionsHelper.Android.cs
```diff
@@ -19,11 +19,185 @@ public static void Initialize()
...
+			// check if permission is granted
+			foreach (var permission in allPermissions)
+			{
+				if (Android.Support.V4.Content.ContextCompat.CheckSelfPermission(context, permission)
+						!= Android.Content.PM.Permission.Granted)
+				{
+					askForPermission.Add(permission);
+				}
+			}
+			return askForPermission;
```
jeromelaban, line 92: Use Linq:

```suggestion
			return allPermissions
			   .Where(p => ContextCompat.CheckSelfPermission(context, p)
						!= Content.PM.Permission.Granted))
			   .ToList();

```

src/Uno.UI/Extensions/PermissionsHelper.Android.cs
```diff
@@ -19,11 +19,185 @@ public static void Initialize()
...
+
+		#endregion
+
+		/// <summary>
+		/// Return null on error, or array of permission to be asked for (not granted at this time). Both parameters can be null
+		/// </summary>
+		/// <param name="requiredPermissions">Permissions that are required. Can be null.</param>
+		/// <param name="optionalPermissions">Permissions that are required only if declared in Manifest. Can be null.</param>
+		/// <returns>Array of all permission that is not granted at this moment (can be empty!), or null if some error occured</returns>
+		public static string[] MissingPermissions(string[] requiredPermissions, string[] optionalPermissions)
```
jeromelaban, line 104:
```suggestion
		public static string[] GetMissingPermissions(string[] requiredPermissions, string[] optionalPermissions)

```

src/Uno.UI/Extensions/PermissionsHelper.Android.cs
```diff
@@ -19,11 +19,185 @@ public static void Initialize()
...
+				return null;
+			}
+
+			// prepare list of all permissions (all required, and all optional defined in Manifest
+			var allPermissions = PermissionWeWant(manifestPermissions, requiredPermissions, optionalPermissions);
+
+			// prepare list of permission to ask for
+			var askForPermission = PermissionToAsk(allPermissions);
+
+			return askForPermission.ToArray();
```
jeromelaban, line 129:
```suggestion
			return PermissionToAsk(allPermissions);

```

src/Uno.UI/Extensions/PermissionsHelper.Android.cs
```diff
@@ -19,11 +19,185 @@ public static void Initialize()
...
+
+		/// <summary>
+		/// Return true if granted are: all requiredPermissions, and all optionalPermissions mentioned in Manifest. Show dialog asking for missing permissions. Both parameters can be null.
+		/// </summary>
+		/// <param name="requiredPermissions">Permissions that are required. Can be null.</param>
+		/// <param name="optionalPermissions">Permissions that are required only if declared in Manifest. Can be null.</param>
+		/// <param name="ignoreErrors">if 'true', then do not throw exception on error, and simply retur false</param>
+		/// <returns>Bool if all permissions are granted, false if any of them is not granted</returns>
+
+		private static async Task<bool> TryAskPermissionAsync(CancellationToken ct, string[] requiredPermissions, string[] optionalPermissions, bool ignoreErrors)
```
jeromelaban, line 141: Let's use the same naming android uses.

```suggestion
		private static async Task<bool> TryRequestPermissionsAsync(CancellationToken ct, string[] requiredPermissions, string[] optionalPermissions, bool ignoreErrors)
```

src/Uno.UI/Extensions/PermissionsHelper.Android.cs
```diff
@@ -19,11 +19,185 @@ public static void Initialize()
 					.CreateAsync<string, bool>(TryGetPermissionCore)
 					.LockInvocation(InvocationLockingMode.Share);
 
-			Windows.Extensions.PermissionsHelper.Initialize(getPermission, CheckPermission);
-		}
+			Windows.Extensions.PermissionsHelper.Initialize(getPermission, CheckPermission, MissingPermissions, TryAskPermissionAsync);
+		}
+
+		private static int _permissionRequest;
```
jeromelaban, line 25: Move all the added code at the end of the file

src/Uno.UI/Extensions/PermissionsHelper.Android.cs
```diff
@@ -19,11 +19,185 @@ public static void Initialize()
...
+					ActivityCompat.RequestPermissions(BaseActivity.Current, askForPermission, code);
 
+					var result = await tcs.Task;
+
+					if (result.GrantResults.Length > 0)
+					{
+						bool allGranted = result.GrantResults.All(r => r == Android.Content.PM.Permission.Granted);
+						return allGranted;
+					}
+					return false;   // some error?
```
jeromelaban, line 191:
```suggestion
					return false;

```

src/Uno.UI/Extensions/PermissionsHelper.Android.cs
```diff
@@ -35,9 +209,9 @@ public static void Initialize()
 		/// <param name="ct">Cancellation Token</param>
 		/// <param name="permissionIdentifier">A permission identifier defined in Manifest.Permission.</param>
 		/// <returns>A <see cref="Task"/> representing the asynchronous operation</returns>
-		public static Task<bool> TryGetPermission(CancellationToken ct, string permissionIdentifier)
-			=> Windows.Extensions.PermissionsHelper.TryGetPermission(ct, permissionIdentifier);
-
+		public static Task<bool> TryGetPermission(CancellationToken ct, string permissionIdentifier)
+					=> Windows.Extensions.PermissionsHelper.TryGetPermission(ct, permissionIdentifier);
```
jeromelaban, line 213: Not needed

src/Uno.UWP/Extensions/PermissionsHelper.cs
```diff
@@ -13,32 +13,82 @@
...
 	/// <summary>
 	/// A service that determines if a permission is granted and if not, request it to the user.
 	/// </summary>
-	public class PermissionsHelper
+	public partial class PermissionsHelper
 	{
 		private static FuncAsync<string, bool> _tryGetPermission;
-		private static FuncAsync<string, bool> _checkPermission;
+		private static FuncAsync<string, bool> _checkPermission;
+		private static Func<string[], string[], string[]> _missingPermissions;
```
jeromelaban, line 24:
```suggestion
		private static Func<string[], string[], string[]> _getMissingPermissions;
```

src/Uno.UWP/Extensions/PermissionsHelper.cs
```diff
@@ -13,32 +13,82 @@
...
+		private static FuncAsync<string[], string[], bool, bool> _tryGetpermissionsAsync;
 
-		internal static void Initialize(FuncAsync<string, bool> getter, FuncAsync<string, bool> checkPermission)
+		internal static void Initialize(FuncAsync<string, bool> getter, FuncAsync<string, bool> checkPermission, Func<string[], string[], string[]> missingPermissions, FuncAsync<string[], string[], bool, bool> tryGetpermissionsAsync)
```
jeromelaban, line 27:
```suggestion
		internal static void Initialize(FuncAsync<string, bool> getter, FuncAsync<string, bool> checkPermission, Func<string[], string[], string[]> missingPermissions, FuncAsync<string[], string[], bool, bool> tryRequestPermissionsAsync)
```

src/Uno.UWP/Extensions/PermissionsHelper.cs
```diff
@@ -13,32 +13,82 @@
...
+		/// <returns>Array with permission if it is not granted at this moment (can be empty!), or null if some error occured</returns>
+		public static string[] MissingPermissions(string requiredPermission)
+			=> _missingPermissions(new string[] {requiredPermission}, null);
+
+		/// <summary>
+		/// Return true if requiredPermission is granted. Show dialog asking for missing permission.
+		/// </summary>
+		/// <param name="requiredPermission">Permission that are required</param>
+		/// <returns>Bool if permission is granted, false if not</returns>
+		public static Task<bool> TryAskPermissionAsync(string requiredPermission)
```
jeromelaban, line 57:
```suggestion
		public static Task<bool> TryRequestPermissionAsync(string requiredPermission)

```

src/Uno.UWP/Extensions/PermissionsHelper.cs
```diff
@@ -13,32 +13,82 @@
...
+			=> TryAskPermissionAsync(new string[] { requiredPermission }, null);
+
+		/// <summary>
+		/// Return true if granted are: requiredPermission, and optionalPermissions (if mentioned in Manifest). Show dialog asking for missing permissions..
+		/// </summary>
+		/// <param name="requiredPermissions">Permission that are required</param>
+		/// <param name="optionalPermissions">Permission that are required only if declared in Manifest</param>
+		/// <returns>Bool if all permissions are granted, false if any of them is not granted</returns>
+		public static Task<bool> TryAskPermissionAsync(string requiredPermission, string optionalPermission)
```
jeromelaban, line 66:
```suggestion
		public static Task<bool> TryRequestPermissionAsync(string requiredPermission, string optionalPermission)

```

src/Uno.UWP/Extensions/PermissionsHelper.cs
```diff
@@ -13,32 +13,82 @@
...
+
+
+		/// <summary>
+		/// Return true if granted are: all requiredPermissions, and all optionalPermissions mentioned in Manifest. Show dialog asking for missing permissions. Both parameters can be null.
+		/// </summary>
+		/// <param name="requiredPermissions">Permissions that are required. Can be null.</param>
+		/// <param name="optionalPermissions">Permissions that are required only if declared in Manifest. Can be null.</param>
+		/// <param name="ignoreErrors">if 'true', then do not throw exception on error, and simply retur false</param>
+		/// <returns>Bool if all permissions are granted, false if any of them is not granted</returns>
+		public static Task<bool> TryAskPermissionAsync(string[] requiredPermissions, string[] optionalPermissions, bool ignoreErrors = false)
```
jeromelaban, line 77:
```suggestion
		public static Task<bool> TryRequestPermissionAsync(string[] requiredPermissions, string[] optionalPermissions, bool ignoreErrors = false)

```

src/Uno.UWP/Extensions/PermissionsHelper.cs
```diff
@@ -13,32 +13,82 @@
...
+		/// <returns>Array of all permission that is not granted at this moment (can be empty!), or null if some error occured</returns>
+		public static string[] MissingPermissions(string[] requiredPermissions, string[] optionalPermissions)
+			=> _missingPermissions(requiredPermissions, optionalPermissions);
+
+		/// <summary>
+		/// Return null on error, or array of permission to be asked for (not granted at this time).
+		/// </summary>
+		/// <param name="requiredPermission">Permission that is </param>
+		/// <returns>Array with permission if it is not granted at this moment (can be empty!), or null if some error occured</returns>
+		public static string[] MissingPermissions(string requiredPermission)
```
jeromelaban, line 49:
```suggestion
		public static string[] GetMissingPermissions(string requiredPermission)

```

src/Uno.UWP/Extensions/PermissionsHelper.cs
```diff
@@ -13,32 +13,82 @@
...
+			_tryGetpermissionsAsync = tryGetpermissionsAsync;
+		}
+
+		/// <summary>
+		/// Return null on error, or array of permission to be asked for (not granted at this time). Both parameters can be null
+		/// </summary>
+		/// <param name="requiredPermissions">Permissions that are required. Can be null.</param>
+		/// <param name="optionalPermissions">Permissions that are required only if declared in Manifest. Can be null.</param>
+		/// <returns>Array of all permission that is not granted at this moment (can be empty!), or null if some error occured</returns>
+		public static string[] MissingPermissions(string[] requiredPermissions, string[] optionalPermissions)
```
jeromelaban, line 41:
```suggestion
		public static string[] GetMissingPermissions(string[] requiredPermissions, string[] optionalPermissions)

```

src/Uno.UWP/Extensions/PermissionsHelper.cs
```diff
@@ -13,32 +13,82 @@
...
 	/// A service that determines if a permission is granted and if not, request it to the user.
 	/// </summary>
-	public class PermissionsHelper
+	public partial class PermissionsHelper
 	{
 		private static FuncAsync<string, bool> _tryGetPermission;
-		private static FuncAsync<string, bool> _checkPermission;
+		private static FuncAsync<string, bool> _checkPermission;
+		private static Func<string[], string[], string[]> _missingPermissions;
+		private static FuncAsync<string[], string[], bool, bool> _tryGetpermissionsAsync;
```
jeromelaban, line 25:
```suggestion
		private static FuncAsync<string[], string[], bool, bool> _tryRequestPermissionsAsync;
```

src/Uno.UI/Extensions/PermissionsHelper.Android.cs
```diff
@@ -19,11 +19,185 @@ public static void Initialize()
...
+			}
+
+			// system dialog asking for permission
+
+			var code = Interlocked.Increment(ref _permissionRequest);
+			var tcs = new TaskCompletionSource<BaseActivity.RequestPermissionsResultWithResultsEventArgs>();
+
+			void handler(object sender, BaseActivity.RequestPermissionsResultWithResultsEventArgs e)
+			{
+
```
jeromelaban, line 167:
```suggestion
```

jeromelaban (commented):
Thanks for the update, this is much better! This will allow you to update your other PRs.

src/Uno.UI/Extensions/PermissionsHelper.Android.cs
pkar70: a) how can I prevent it?
b) this method is `private`, so any comments could be displayed only within this file (class), so I guess it is not important

src/Uno.UI/Extensions/PermissionsHelper.Android.cs
pkar70: but it can help in code understanding...

src/Uno.UI/Extensions/PermissionsHelper.Android.cs
pkar70: "Requested" is misleading - requested by whom from whom, and when?
Second parameter is list of permissions that "maybe", so not requested.
And returned list is list of permission we want to ask for, so "ToBeRequested" :)

src/Uno.UI/Extensions/PermissionsHelper.Android.cs
pkar70: "some error" means: this is some error condition, as we should never get into this `return` (`result.GrantResults.Length` should always be > 0), but we don't throw exception here. Maybe we should...

src/Uno.UI/Extensions/PermissionsHelper.Android.cs
pkar70: General comment about "Ask" and "Request" (in all methods) - we doesn't request permission, we kindly ask user for granting us permission...

pkar70:
Why there are conflicting files? I used rebase unoplatform/master?

src/Uno.UI/Extensions/PermissionsHelper.Android.cs
jeromelaban: Sure, but that's this project's standard way to document methods. If the comment is required to explain what the method does, it has its place there.

src/Uno.UI/Extensions/PermissionsHelper.Android.cs
jeromelaban: That's fair. Lets call it `GetPermissionsToRequest` then.

src/Uno.UI/Extensions/PermissionsHelper.Android.cs
jeromelaban: `Request` is what the android API is called, if I'm not mistaken, this is only aligning.

src/Uno.UI/Extensions/PermissionsHelper.Android.cs
pkar70: So, where I should place this comment?

src/Uno.UI/Extensions/PermissionsHelper.Android.cs
pkar70: It is not permissions to request, because some of them are already granted and would be filtered-out from this list.

jeromelaban: If the list returned the items to be requested, it's a list of items to request.

src/Uno.UI/Extensions/PermissionsHelper.Android.cs
pkar70: So, where I should place this comment?

src/Uno.UI/Extensions/PermissionsHelper.Android.cs
pkar70: No, we doesn't request this list of permissions - before asking for these permissions, this list is filtered.

src/Uno.UI/Extensions/PermissionsHelper.Android.cs
jeromelaban: Place it as an XML comment, as I mentioned above, that's the way to document methods.

pkar70: Oh, misunderstanding...
I thought that this comment, as it is, is placed into comments visible in editor as tooltip, and that you want it to not be placed into tooltip :)

This method is private, so it should not be called from any other place than main method, GetMissingPermissions()... and in current version these method is converted to one-line LINQ expression in GetMissingPermissions method :)

src/Uno.UWP/Extensions/PermissionsHelper.cs
```diff
@@ -17,17 +17,67 @@ namespace Windows.Extensions
...
+		/// <param name="optionalPermissions">Permissions that are required only if declared in Manifest. Can be null.</param>
+		/// <returns>Array of all permission that is not granted at this moment (can be empty!), or null if some error occured</returns>
+		public static string[] MissingPermissions(string[] requiredPermissions, string[] optionalPermissions)
+			=> _getMissingPermissions(requiredPermissions, optionalPermissions);
+
+		/// <summary>
+		/// Return null on error, or array of permission to be asked for (not granted at this time).
+		/// </summary>
+		/// <param name="requiredPermission">Permission that is </param>
+		/// <returns>Array with permission if it is not granted at this moment (can be empty!), or null if some error occured</returns>
```
agneszitte-nventive, line 48:
```suggestion
		/// <returns>Array with permission if it is not granted at this moment (can be empty!), or null if some error occurred </returns>
```

src/Uno.UWP/Extensions/PermissionsHelper.cs
```diff
@@ -17,17 +17,67 @@ namespace Windows.Extensions
...
+		public static Task<bool> TryAskPermissionAsync(string requiredPermission, string optionalPermission)
+			=> TryAskPermissionAsync(new string[] { requiredPermission }, new string[] { optionalPermission });
+
+
+		/// <summary>
+		/// Return true if granted are: all requiredPermissions, and all optionalPermissions mentioned in Manifest. Show dialog asking for missing permissions. Both parameters can be null.
+		/// </summary>
+		/// <param name="requiredPermissions">Permissions that are required. Can be null.</param>
+		/// <param name="optionalPermissions">Permissions that are required only if declared in Manifest. Can be null.</param>
+		/// <param name="ignoreErrors">if 'true', then do not throw exception on error, and simply retur false</param>
```
agneszitte-nventive, line 75:
```suggestion
		/// <param name="ignoreErrors">if 'true', then do not throw exception on error, and simply return false</param>
```

agneszitte-nventive:
@pkar70 some conflicts need to be resolved for this PR ;)

Issue #2813, opened 2020-03-13
[IOS] DoubleTap event does not fire


peternary:
The DoubleTap event does not fire in iOS builds.  Appears to work on all other platforms.  To test, in attached sample project double-tap the green grid - text is updated on all platforms but iOS.

[App30.zip](https://github.com/unoplatform/uno/files/4330567/App30.zip)

Using Uno.UI version 2.2.0-dev.131

dr1rrb:
FYI, this has been broken when I added the support for multi-touch. The pointer ID was the same for all pointers of the same type (finger vs. pen vs. mouse), now they are different for each, but are also different for each press/release sequence (i.e. each time you release the pointer it changes it ID) (cf. https://github.com/unoplatform/uno/blob/master/src/Uno.UI/UI/Xaml/Input/PointerRoutedEventArgs.iOS.cs#L34). 

I need to figure a way to track them properly (which does seems to be supported on iOS), and if not possible, I'll have to adjust the `GestureRecognizer` specifically for iOS to track only the type of the pointer.

I'll also update the UI test engine to allow tests of double tap to prevent those kind of issue in future updates.

PR #3033, opened 2020-04-21
fix(pointers): [iOS] Fix the DoubleTap

dr1rrb:
fixes https://github.com/unoplatform/uno/issues/2813

## Bugfix
The `UIElement.DoubleTapped` event is not fired

On iOS, each time a pointer is pressed down, it has a new ID. The issue is that ID is needed to track and detect double tap.

## What is the new behavior?
We now generates a managed pointer ID which is reset only once all pointers are release. So a pointer will always have the same ID for single touch, and we don't have conflicting pointer ID (like we used to have before using the `UITouch.Handle`)

## PR Checklist

### DIFF
--- a/src/Uno.UI/UI/Xaml/UIElement.Pointers.wasm.cs
+++ b/src/Uno.UI/UI/Xaml/UIElement.Pointers.wasm.cs
@@ -138,7 +138,7 @@ namespace Windows.UI.Xaml
 				}
 				if (args.wheelDeltaY != 0)
 				{
-					// Note: Vertical scrolling is inverted between web browser and WinUI!
+					// Note: Web browser vertical scrolling is the opposite compared to WinUI!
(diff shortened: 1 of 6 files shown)


src/Uno.UI/UI/Xaml/UIElement.Pointers.wasm.cs
```diff
@@ -138,7 +138,7 @@ private static bool DispatchNativePointerWheel(UIElement target, string eventPay
 				}
 				if (args.wheelDeltaY != 0)
 				{
-					// Note: Vertical scrolling is inverted between web browser and WinUI!
+					// Note: Web browser vertical scrolling is the opposite compared to WinUI!
```
jeromelaban, line 141:
```suggestion
					// Note: Web browser vertical scrolling is the opposite compared of WinUI!
```

jeromelaban:
Build says:  `error CS1517: Invalid preprocessor expression`

PR #3048, opened 2020-04-24
Update documentation to mention you need VS2019

MatFillion:
Update documentation to mention you need VS2019

## PR Type


jeromelaban:
It's not true for all targets, but that's better to be safe at that level :) Thanks!

PR #3078, opened 2020-04-28
Improve NativeFramePresenter.iOS to fix concurrency issues.

jeanplevesque:
## PR Type

- Bugfix

The application crashes you press the back button on a native `CommandBar` then immediately call `Frame.Navigate()`.

## What is the new behavior?

The `Frame.Navigating` gets cancelled when you press the back button on a native `CommandBar` just before calling `Frame.Navigate()`.

## PR Checklist

180990

### DIFF
--- a/src/Uno.UI/Controls/NativeFramePresenter.iOS.cs
+++ b/src/Uno.UI/Controls/NativeFramePresenter.iOS.cs
@@ -14,15 +14,68 @@ using Windows.UI.Xaml.Media.Animation;
 using System.Collections.ObjectModel;
 using System.Collections.Specialized;
 using System.Threading.Tasks;
+using ObjCRuntime;
 
 namespace Uno.UI.Controls
 {
 	public partial class NativeFramePresenter : FrameworkElement
 	{
+		/* Architecture
+		 *
+		 * This class deals with the navigations events from its Frame and UINavigationController.
+		 * There are 2 main event routes: events from the Frame, and events from the UINavigationController.
+		 * 
+		 * 1. From the Frame
+		 * 
+		 *    - Frame.Navigating --> Frame.Navigated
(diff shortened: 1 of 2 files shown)


src/Uno.UI/Controls/NativeFramePresenter.iOS.cs
```diff
@@ -139,40 +381,290 @@ private bool GetIsAnimated(PageStackEntry entry)
 			// - CommonNavigationTransitionInfo
 		}
 
+		private void WillShowViewController(UINavigationController navigationController, [Transient] UIViewController viewController, bool animated)
+		{
+			TraceViewControllers(nameof(WillShowViewController), viewController);
+
+			if (viewController is PageViewController pageViewController)
```
davidjohnoliver, line 388: There's a huge amount of nesting here. To the extent that you can rewrite this on an 'early exit' pattern (ex: `if (!(viewController is PageViewController pageViewController)) {return;}`) it's going to be much easier to read.

jeanplevesque: My idea was to make sure all cases were though of. That's why there are *empty* `else` blocks that contain only comments.

jeanplevesque: I've integrated you suggestions

jeromelaban:
Indeed, having a native frame navigation sample would help. You may not be able to validate the behaviors like the backswipe, but testing programmatically would be enough. You can build this test by making it for Android, using native frame styles, and the CI will validate it on iOS.

jeanplevesque:
@jeromelaban I created a new Issue for the mentioned UI tests: https://github.com/unoplatform/uno/issues/3108

PR #3373, opened 2020-06-15
Improve DepencencyProperty R/W performance

jeromelaban:
## PR Type

-- Refactoring (no functional changes, no api changes)

## What is the new behavior?

- Move DependencyObjectStore to C# 8.0 nullable
- Change `DependencyPropertyDetailsCollection` to use an O(1) lookup
- Other minor performance updates

## PR Checklist

## Other information

### DIFF
--- a/src/SamplesApp/Benchmarks.Shared/Suite/Windows_UI_Xaml_Controls/DependencyPropertyBench/SimpleDPBenchmark.cs
+++ b/src/SamplesApp/Benchmarks.Shared/Suite/Windows_UI_Xaml_Controls/DependencyPropertyBench/SimpleDPBenchmark.cs
@@ -1,0 +1,38 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using BenchmarkDotNet.Attributes;
+using Windows.Foundation;
+using Windows.UI.Xaml.Controls;
+
+namespace SamplesApp.Benchmarks.Suite.Windows_UI_Xaml_Controls.GridBench
+{
+	public class SimpleDPBenchmark
+	{
+		private Grid SUT;
+
+		[GlobalSetup]
+		public void Setup()
+		{
+			SUT = new Grid();
+		}
+
+		[Benchmark()]
+		public void DP_Write()
+		{
+			for (int i = 0; i < 100; i++)
+			{
+				SUT.Width = i;
+			}
+		}
+
+		[Benchmark()]
+		public void DP_Read()
+		{
+			for (int i = 0; i < 100; i++)
+			{
+				var r = SUT.Width;
+			}
+		}
+	}
+}
--- a/src/Uno.UI/UI/Xaml/DependencyObjectStore.Binder.cs
+++ b/src/Uno.UI/UI/Xaml/DependencyObjectStore.Binder.cs
@@ -1,3 +1,5 @@
+#nullable enable
+
 using System;
 using System.Linq;
 using Uno.Disposables;
@@ -32,13 +34,13 @@ namespace Windows.UI.Xaml
 		private readonly object _gate = new object();
 
 		private readonly Dictionary<DependencyProperty, int> _childrenBindableMap = new Dictionary<DependencyProperty, int>(0, DependencyPropertyComparer.Default);
-		private readonly List<object> _childrenBindable = new List<object>();
+		private readonly List<object?> _childrenBindable = new List<object?>();
 
 		private bool _isApplyingTemplateBindings;
 		private bool _isApplyingDataContextBindings;
 		private bool _bindingsSuspended;
-		private DependencyProperty _dataContextProperty;
-		private DependencyProperty _templatedParentProperty;
+		private DependencyProperty _dataContextProperty = UIElement.DataContextProperty;
+		private DependencyProperty _templatedParentProperty = UIElement.TemplatedParentProperty;
 
 		/// <summary>
 		/// Sets the templated parent, with the ability to control the propagation of the templated parent.
@@ -48,7 +50,7 @@ namespace Windows.UI.Xaml
 		/// Applies the templated parent to children if true. False is generally used when a control is template-able
 		/// to avoid propagating its own templated parent to its children.
 		/// </param>
-		public void SetTemplatedParent(FrameworkElement templatedParent)
+		public void SetTemplatedParent(FrameworkElement? templatedParent)
 		{
 #if !HAS_EXPENSIVE_TRYFINALLY
 			// The try/finally incurs a very large performance hit in mono-wasm, and SetValue is in a very hot execution path.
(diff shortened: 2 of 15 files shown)


src/SamplesApp/Benchmarks.Shared/Suite/Windows_UI_Xaml_Controls/DependencyPropertyBench/SimpleDPBenchmark.cs
```diff
@@ -0,0 +1,38 @@
...
+		[GlobalSetup]
+		public void Setup()
+		{
+			SUT = new Grid();
+		}
+
+		[Benchmark()]
+		public void DP_Write()
+		{
+			for (int i = 0; i < 100; i++)
```
ebariche, line 23: Does unrolling give you nicer numbers here ?

ebariche: Also is a 100 times enough ?

src/SamplesApp/Benchmarks.Shared/Suite/Windows_UI_Xaml_Controls/DependencyPropertyBench/SimpleDPBenchmark.cs
```diff
@@ -0,0 +1,38 @@
...
+				SUT.Width = i;
+			}
+		}
+
+		[Benchmark()]
+		public void DP_Read()
+		{
+			for (int i = 0; i < 100; i++)
+			{
+				var r = SUT.Width;
```
ebariche, line 34: Are we sure the read doesn't get discarded in release ?

jeromelaban: It's considered a method invocation, so the compiler won't remove them because it can have a side effect.

src/SamplesApp/Benchmarks.Shared/Suite/Windows_UI_Xaml_Controls/DependencyPropertyBench/SimpleDPBenchmark.cs
jeromelaban: Probably. I'm not sure if it's the best way to test this tbh. I'll adjust the test with the next PR to validate more.

src/SamplesApp/Benchmarks.Shared/Suite/Windows_UI_Xaml_Controls/DependencyPropertyBench/SimpleDPBenchmark.cs
ebariche: Good to know :)

src/Uno.UI/UI/Xaml/DependencyObjectStore.Binder.cs
```diff
@@ -32,13 +34,13 @@ public partial class DependencyObjectStore
...
 		private readonly Dictionary<DependencyProperty, int> _childrenBindableMap = new Dictionary<DependencyProperty, int>(0, DependencyPropertyComparer.Default);
-		private readonly List<object> _childrenBindable = new List<object>();
+		private readonly List<object?> _childrenBindable = new List<object?>();
 
 		private bool _isApplyingTemplateBindings;
 		private bool _isApplyingDataContextBindings;
 		private bool _bindingsSuspended;
-		private DependencyProperty _dataContextProperty;
-		private DependencyProperty _templatedParentProperty;
+		private DependencyProperty _dataContextProperty = UIElement.DataContextProperty;
```
davidjohnoliver, line 42:
```suggestion
		// Pre-initialize to the most likely value, since in practice most DependencyObjects will be UIElements. Note this is not readonly since there is not a single DataContextProperty (since DependencyObject is an interface in Uno)
		private DependencyProperty _dataContextProperty = UIElement.DataContextProperty;
```

I propose a comment, since it took me a while to figure out why this shouldn't be `static readonly` 😛

jeromelaban: That's fair :) Will add!

jeromelaban: In theory, it may not need to be assigned, but since it's a non-nullable reference, adding a default value is the workaround. No sure I should keep this that way, though...

davidjohnoliver: It's not sufficient to ensure it's initialized to a non-null value in the ctor? I haven't played around with the nullability checking much.

src/Uno.UI/UI/Xaml/DependencyPropertyDetailsCollection.cs
```diff
@@ -37,184 +50,129 @@ public DependencyPropertyDetailsCollection(Type ownerType, ManagedWeakReference
...
-				entry.Id = propertiesForType[i].UniqueId;
-				entry.Details = null;
-			}
+				_minId = propertiesForType[0].UniqueId;
+				_maxId = propertiesForType[propertiesForType.Length - 1].UniqueId;
 
-			// Entries are pre-sorted by the DependencyProperty.GetPropertiesForType method
-			AssignEntries(entries, propertiesForType.Length, sort: false);
+				var entriesLength = _maxId - _minId + 1;
+				var entries = _pool.Rent(entriesLength);
```
dr1rrb, line 59: Wondering if it would be interesting to align the `entriesLength` on 16 or 32 to increase the possibility to have a match in the `_pool` and so increase array re-use ?

src/Uno.UI/UI/Xaml/FrameworkElement.Layout.wasm.cs
```diff
@@ -270,23 +270,7 @@ private void InnerArrangeCore(Rect finalRect)
...
-					return g.BorderThickness;
-
-				case Controls.ContentPresenter p:
-					return p.BorderThickness;
-
-				default:
-					return Thickness.Empty;
-			}
-		}
+		private protected virtual Thickness GetThicknessAdjust() => Thickness.Empty;
```
dr1rrb, line 273: I would change the name of this method if it becomes overridable. In the "layout.wasm.cs" file I can (barely) understand its purpose, but while in `Border` for instance I just didn't understand while it was not just named `GetBorderThickness`.

src/Uno.UI/UI/Xaml/DependencyPropertyDetailsCollection.cs
jeromelaban: The pool provides arrays based on the closest higher size, and this collection class is already using the larger size if provided :)

src/Uno.UI/UI/Xaml/DependencyObjectStore.Binder.cs
jeromelaban: Looking at it again, it's not needed. The `InitializeBinder` method is not needed anymore and those variables can be initialized from the ctor.

PR #3387, opened 2020-06-17
[WIP] Element RequestedTheme, ActualTheme

### LINKED ISSUE #3302: Element-level Theme

Support for:

- `FrameworkElement.RequestedTheme`
- `FrameworkElement.ActualTheme
- `FrameworkElement.ActualThemeChanged`

## Why is this needed:

For full theming fidelity.

## For which Platform:

- [x] iOS
- [x] Android
- [x] WebAssembly
- [x] macOS
- [x] Windows
- [ ] Build tasks
- [ ] Solution Templates

MartinZikmund:
GitHub Issue (If applicable): closes #3302 

## PR Type

- Feature

Element-level theming is not supported.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Application.cs
+++ b/src/Uno.UI/UI/Xaml/Application.cs
@@ -323,58 +323,9 @@ namespace Windows.UI.Xaml
 
 		private void OnRequestedThemeChanged()
 		{
-			if (GetTreeRoot() is { } root)
-			{
-				// Update theme bindings in application resources
-				Resources?.UpdateThemeBindings();
-
-				// Update theme bindings in system resources
-				ResourceResolver.UpdateSystemThemeBindings();
-
-				PropagateThemeChanged(root);
-			}
-
-			// Start from the real root, which may not be a FrameworkElement on some platforms
-			View GetTreeRoot()
-			{
-				View current = Windows.UI.Xaml.Window.Current.Content;
-				var parent = current?.GetVisualTreeParent();
-				while (parent != null)
-				{
-					current = parent;
-					parent = current?.GetVisualTreeParent();
-				}
-				return current;
-			}
-		}
-
-		/// <summary>
-		/// Propagate theme changed to <paramref name="instance"/> and its descendants, to have them update any theme bindings.
-		/// </summary>
-		internal static void PropagateThemeChanged(object instance)
-		{
-
-			// Update ThemeResource references that have changed
-			if (instance is FrameworkElement fe)
-			{
-				fe.UpdateThemeBindings();
-			}
-
-			//Try Panel.Children before ViewGroup.GetChildren - this results in fewer allocations
-			if (instance is Controls.Panel p)
-			{
-				foreach (object o in p.Children)
-				{
-					PropagateThemeChanged(o);
-				}
-			}
-			else if (instance is ViewGroup g)
-			{
-				foreach (object o in g.GetChildren())
-				{
-					PropagateThemeChanged(o);
-				}
-			}
+			FrameworkElement.PropagateThemeScope(
+				Windows.UI.Xaml.Window.Current.Content,
+				ElementTheme.Default);
 		}
 	}
 }
--- a/src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.cs
@@ -417,6 +417,9 @@ namespace Windows.UI.Xaml.Controls
 				candidate = candidate?.GetParent();
 			}			
 
+			// We set values defined by the nearest explicit provider of Foreground
+			// as ImplicitStyle to ensure that the value overrides the inherited value,
+			// as RequestedTheme must "break" the inheritance
 			if (candidate != null)
 			{
 				var foregroundSource = (FrameworkElement)candidate;
--- a/src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.cs
@@ -31,6 +31,7 @@ namespace Windows.UI.Xaml.Controls
 	[ContentProperty(Name = "Text")]
 	public partial class TextBlock : DependencyObject
 	{
+		private bool _requestedThemeForegroundOverride = false;
 		private InlineCollection _inlines;
 		private string _inlinesText; // Text derived from the content of Inlines
 		private readonly SerialDisposable _foregroundChanged = new SerialDisposable();
@@ -394,7 +395,7 @@ namespace Windows.UI.Xaml.Controls
 			);
 
 
-		private void SetDefaultForeground()
+		private void SetImplicitForeground()
 		{
 			// override the default value from dependency property based on actual theme
 			if (ActualTheme == ElementTheme.Default)
@@ -415,7 +416,17 @@ namespace Windows.UI.Xaml.Controls
 					}
 				}
 				candidate = candidate?.GetParent();
-			}			
+			}
+
+			if (_requestedThemeForegroundOverride)
+			{
+				// Unset the implicit value previously applied by requested theme
+				this.SetValue(
+					ForegroundProperty,
+					DependencyProperty.UnsetValue,
+					DependencyPropertyValuePrecedences.ImplicitStyle);
+				_requestedThemeForegroundOverride = false;
+			}
 
 			// We set values defined by the nearest explicit provider of Foreground
 			// as ImplicitStyle to ensure that the value overrides the inherited value,
@@ -425,28 +436,18 @@ namespace Windows.UI.Xaml.Controls
 				var foregroundSource = (FrameworkElement)candidate;
 				if (foregroundSource.RequestedTheme != ElementTheme.Default)
 				{
+					// set foreground on implicit style level and flag, so we can later
+					// know if it can be unset
 					this.SetValue(
 						ForegroundProperty,
 						foregroundSource.RequestedTheme == ElementTheme.Light ?
 							SolidColorBrushHelper.Black : SolidColorBrushHelper.White,
 						DependencyPropertyValuePrecedences.ImplicitStyle);
-				}
-				else
-				{
-					this.SetValue(
-						ForegroundProperty,
-						foregroundSource.GetValue(Control.ForegroundProperty),
-						DependencyPropertyValuePrecedences.ImplicitStyle);
+					_requestedThemeForegroundOverride = true;
 				}
 			}
 			else
 			{
-				// Unset the "implicit" value to ensure proper fallback 
-				this.SetValue(
-					ForegroundProperty,
-					DependencyProperty.UnsetValue,
-					DependencyPropertyValuePrecedences.ImplicitStyle);
-
 				this.SetValue(
 					ForegroundProperty,
 					ActualTheme == ElementTheme.Light ?
@@ -458,7 +459,7 @@ namespace Windows.UI.Xaml.Controls
 		internal override void UpdateThemeBindings()
 		{
 			base.UpdateThemeBindings();
-			SetDefaultForeground();
+			SetImplicitForeground();
 		}
 
 		private void OnForegroundChanged()
--- a/src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.cs
@@ -396,32 +396,53 @@ namespace Windows.UI.Xaml.Controls
 
 		private void SetDefaultForeground()
 		{
-			//override the default value from dependency property based on application theme
-			//and requested element theme
-			if (RequestedTheme == ElementTheme.Default)
+			// override the default value from dependency property based on actual theme
+			if (ActualTheme == ElementTheme.Default)
 			{
-				// Ensure the implicit style is unset before setting application default
-				this.SetValue(
-					ForegroundProperty,
-					DependencyProperty.UnsetValue,
-					DependencyPropertyValuePrecedences.ImplicitStyle);
+				throw new InvalidOperationException("Actual theme may not be default");
+			}
 
-				this.SetValue(
-					ForegroundProperty,
-					Application.Current == null || Application.Current.RequestedTheme == ApplicationTheme.Light
-						? SolidColorBrushHelper.Black
-						: SolidColorBrushHelper.White,
-					DependencyPropertyValuePrecedences.DefaultStyle);
+			// find nearest parent with set Foreground or RequestedTheme
+			object candidate = this;
+			while (candidate != null)
+			{
+				if (candidate is FrameworkElement candidateFe)
+				{
+					if (candidateFe.RequestedTheme != ElementTheme.Default ||
+						candidateFe.GetCurrentHighestValuePrecedence(Control.ForegroundProperty) < DependencyPropertyValuePrecedences.Inheritance)
+					{
+						break;
+					}
+				}
+				candidate = candidate?.GetParent();
+			}
+
+			if (candidate != null)
+			{
+				var foregroundSource = (FrameworkElement)candidate;
+				if (foregroundSource.RequestedTheme != ElementTheme.Default)
+				{
+					this.SetValue(
+						ForegroundProperty,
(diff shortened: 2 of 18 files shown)


src/Uno.UI/UI/Xaml/FrameworkElement.cs
```diff
@@ -357,6 +365,90 @@ private void UpdateActiveStyle()
...
+		public ElementTheme ActualTheme
+		{
+			get => (ElementTheme)GetValue(ActualThemeProperty);
+		}
+
+		public static DependencyProperty ActualThemeProperty { get; } =
+			DependencyProperty.Register(
+				nameof(ActualTheme), typeof(ElementTheme),
+				typeof(FrameworkElement),
+				new PropertyMetadata(ElementTheme.Default, ActualThemePropertyChanged));
```
jeromelaban, line 421: Should this property be property inherited ?

src/Uno.UI/UI/Xaml/FrameworkElement.cs
```diff
@@ -357,6 +365,90 @@ private void UpdateActiveStyle()
...
+			}
+			else
+			{
+				this.SetValue(ActualThemeProperty, RequestedTheme, DependencyPropertyValuePrecedences.DefaultValue);
+			}
+		}
+
+		private static void ActualThemePropertyChanged(DependencyObject dependencyObject, DependencyPropertyChangedEventArgs args)
+		{
+			if (args.NewPrecedence < DependencyPropertyValuePrecedences.DefaultValue)
```
dr1rrb, line 442: That's smart! Like it!

src/Uno.UI/UI/Xaml/ResourceDictionary.cs
```diff
@@ -159,12 +159,12 @@ private void TryMaterializeLazy(object key, ref object value)
 			}
 		}
 
-		private bool GetFromMerged(object key, out object value)
+		private bool GetFromMerged(object key, out object value, ElementTheme elementTheme = ElementTheme.Default)
```
dr1rrb, line 162: I would assume that this parameter should not be optional, or at least, it should default to the `RequestedTheme` (same thing for all other methods that now accepts a theme as parameter).

src/Uno.UI/UI/Xaml/FrameworkElement.cs
```diff
@@ -357,6 +365,90 @@ private void UpdateActiveStyle()
...
+			set => SetValue(RequestedThemeProperty, value);
+		}
+
+		public static DependencyProperty RequestedThemeProperty { get; } =
+			DependencyProperty.Register(
+				nameof(RequestedTheme), typeof(ElementTheme),
+				typeof(FrameworkElement),
+				new FrameworkPropertyMetadata(ElementTheme.Default, FrameworkPropertyMetadataOptions.Inherits, RequestedThemePropertyChanged));
+
+		private static void RequestedThemePropertyChanged(DependencyObject dependencyObject, DependencyPropertyChangedEventArgs args)
```
davidjohnoliver, line 380: Should be `OnRequestedThemePropertyChanged` per normal convention

src/Uno.UI/UI/Xaml/FrameworkElement.cs
```diff
@@ -357,6 +365,90 @@ private void UpdateActiveStyle()
...
+				}
+				else if (instance is ViewGroup g)
+				{
+					foreach (object o in g.GetChildren())
+					{
+						PropagateThemeChanged(o);
+					}
+				}
+			}
+			PropagateThemeChanged(frameworkElement);
```
davidjohnoliver, line 409: I'm not sure why you need this - isn't the point of making the DP inherited that you _don't_ have to manually propagate the change?

src/Uno.UI/UI/Xaml/FrameworkElement.cs
```diff
@@ -357,6 +365,90 @@ private void UpdateActiveStyle()
...
+						ElementTheme.Light : ElementTheme.Dark,
+					DependencyPropertyValuePrecedences.DefaultValue);
+			}
+			else
+			{
+				this.SetValue(ActualThemeProperty, RequestedTheme, DependencyPropertyValuePrecedences.DefaultValue);
+			}
+		}
+
+		private static void ActualThemePropertyChanged(DependencyObject dependencyObject, DependencyPropertyChangedEventArgs args)
```
davidjohnoliver, line 440: should be `OnActualThemePropertyChanged`

src/Uno.UI/UI/Xaml/ResourceDictionary.cs
davidjohnoliver: +1, I prefer not to have default parameter values on private or internal methods, unless it'd really introduce a lot of noise otherwise.

src/Uno.UI/UI/Xaml/ResourceDictionary.cs
```diff
@@ -188,7 +188,15 @@ private bool ContainsKeyMerged(object key)
 			return false;
 		}
 
-		private ResourceDictionary GetThemeDictionary() => GetThemeDictionary(Themes.Active) ?? GetThemeDictionary(Themes.Default);
+		private ResourceDictionary GetThemeDictionary(ElementTheme elementTheme = ElementTheme.Default)
+		{
+			var theme = Themes.Active;
+			if (elementTheme != ElementTheme.Default)
+			{
+				theme = elementTheme.ToString("g");
```
davidjohnoliver, line 196: Not sure if calling enum.ToString() has potential perf issues - a switch block might be safer.

MartinZikmund:
@davidjohnoliver Have you been able to get to #3388 yet, please? I have noticed it behaves unusually even for some non-visual state items. I will update the issue with a comment :-) .

src/Uno.UI/UI/Xaml/FrameworkElement.cs
MartinZikmund: Adjusted :-)

src/Uno.UI/UI/Xaml/ResourceDictionary.cs
MartinZikmund: Adjusted :-)

src/Uno.UI/UI/Xaml/FrameworkElement.cs
MartinZikmund: Adjusted :-)

MartinZikmund:
@davidjohnoliver @jeromelaban @dr1rrb Pushed new version, made several changes to the logic

- both dependency properties now use inheritance
- theme dictionary search does fallback to `Default` if the requested theme dictionary exists, but the searched key does not
- added element theme support to some code paths which I previously didn't notice
- removed default parameter values to make `ElementTheme` explicit

There is still a bit more work required, as I noticed `BasicThemeResources` sample does not yet work the same way as in UWP, so I will continue with this

MartinZikmund:
@davidjohnoliver Maybe I found why the `TextBlock` color didn't update in my Messages sample - it seems that when `TextBlock` color is not explicitly set, it does not inherit its value from the container control's foreground. This causes an issue as the color stays on its default value. You can see this currently on the `BasicThemeResources` page when you switch to **Local Dark** theme - this top `TextBlock` which stays always black:
image

davidjohnoliver:
> @davidjohnoliver
> 
> ## Update
> Maybe this will be a bit more complicated :-D . Consider the following:

Indeed it may not be straightforward :D It seems to touch on this issue - https://github.com/unoplatform/uno/issues/3443 . Inheritance predecence is tricky.

MartinZikmund:
@davidjohnoliver Made another round of significant changes:

- `RequestedTheme` no longer inherits - I noticed that setting it on a scope in UWP does not affect the children. Hence now changing performs a "tunneling" similar to when Application theme is set - both are now refactored to use a common `FrameworkElement` logic
- I had to add precedence to `ResourceBinding`, as default style theme resources were overwriting the explicit style resources (as theme resource setting had previously used local precedence). I hope I didn't miss something, please double check this portion especially.
- I had to add logic to update theme resources when parent of an element is changed. This is needed because during parsing, the parent is not yet set, so its resources cannot be found.
- I have updated the `ResourceDictionary` indexer to throw when key is not found, to match UWP. I have found a commented out test for this, however so I am not sure if this was done for some specific reason? Maybe because we expect some resources may not yet exist like reveal brushes?

What is missing is to adjust the `Foreground` property which is super tricky. The current inheritance is not sufficient, as setting `RequestedTheme` anywhere in between the source and target needs to "reset" this to default. I have a solution in, but it does not seem to work completely (switching to default theme from dark leaves font white), so I will need to work on that some more.

davidjohnoliver:
@MartinZikmund Nice stuff, I'll take a look. :)

For the commented-out test, I suspect the behaviour might've been different for the older UWP 1709, but I can't remember for sure. It's fine to restore it if it matches the behaviour of the UWP version we're currently targeting.

MartinZikmund:
> @MartinZikmund Nice stuff, I'll take a look. :)
> 
> For the commented-out test, I suspect the behaviour might've been different for the older UWP 1709, but I can't remember for sure. It's fine to restore it if it matches the behaviour of the UWP version we're currently targeting.

@davidjohnoliver Tested with the two latest SDKs and it seems both throws, so the exception should be ok.

src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.cs
```diff
@@ -393,6 +393,71 @@ Brush Foreground
...
+			if (candidate != null)
+			{
+				var foregroundSource = (FrameworkElement)candidate;
+				if (foregroundSource.RequestedTheme != ElementTheme.Default)
+				{
+					this.SetValue(
+						ForegroundProperty,
+						foregroundSource.RequestedTheme == ElementTheme.Light ?
+							SolidColorBrushHelper.Black : SolidColorBrushHelper.White,
+						DependencyPropertyValuePrecedences.ImplicitStyle);
```
davidjohnoliver, line 429: I'm not sure I understand the reasoning with the precedences. Why are you using `ImplicitStyle` here?

src/Uno.UI/UI/Xaml/ResourceDictionary.cs
```diff
@@ -89,32 +89,34 @@ public void Add(object key, object value)
...
-		public bool ContainsKey(object key, bool shouldCheckSystem) => _values.ContainsKey(key) || ContainsKeyMerged(key) || ContainsKeyTheme(key)
+		public bool ContainsKey(object key) => ContainsKey(key, shouldCheckSystem: true, ElementTheme.Default);
+		public bool ContainsKey(object key, bool shouldCheckSystem, ElementTheme elementTheme) => _values.ContainsKey(key) || ContainsKeyMerged(key) || ContainsKeyTheme(key, elementTheme)
 			|| (shouldCheckSystem && !IsSystemDictionary && ResourceResolver.ContainsKeySystem(key));
 
-		public bool TryGetValue(object key, out object value) => TryGetValue(key, out value, shouldCheckSystem: true);
-		public bool TryGetValue(object key, out object value, bool shouldCheckSystem)
+		public bool TryGetValue(object key, out object value) => TryGetValue(key, out value, shouldCheckSystem: true, ElementTheme.Default);
+
+		/// <remarks>Should become internal in the future?</remarks>
```
davidjohnoliver, line 98: oops! Yes this should be internal, same for the `ContainsKey()` overload

src/Uno.UI/UI/Xaml/DependencyObjectStore.Binder.cs
```diff
@@ -317,6 +323,12 @@ public void SetBinding(object target, string dependencyProperty, BindingBase bin
 			}
 		}
 
+		public void SetResourceBinding(DependencyProperty dependencyProperty, object resourceKey, bool isTheme, object context)
+		{
+			var binding = new ResourceBinding(resourceKey, isTheme, context, _precedenceOverride ?? DependencyPropertyValuePrecedences.Local);
```
davidjohnoliver, line 328: ah, nice

src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.cs
MartinZikmund: My reasoning here is that I need to make sure that RequestedTheme has precedence over the Inherited value - which ImplicitStyle does. The reason is that setting RequestedTheme somewhere in the tree "stops" the Foreground property inheritance

davidjohnoliver: Can you add a comment here explaining this?

I'm also curious how this would interact with issue of style vs inheritance precedence - https://github.com/unoplatform/uno/issues/3443

src/Uno.UI/UI/Xaml/ResourceDictionary.cs
MartinZikmund: Should I change it within this PR?

src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.cs
MartinZikmund: Hmm, good question 🤔 . I am now realizing, I will probably also need a similar "trick" added to the `Control` foreground logic

src/Uno.UI/UI/Xaml/ResourceDictionary.cs
davidjohnoliver: Yes, go ahead

MartinZikmund: Done :-)

src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.cs
MartinZikmund: I was wrong, this is `TextBlock`-specific, it behaves a bit differently. I have updated the code to be less invasive about the implicit setting and all test I have done seem ok now. Let's see what UI test suite says. Regarding #3443, these changes, unfortunately, don't fix it, but it still behaves the same at least :-D . There is still the outstanding issue with visual states however (that theme is not properly applied when a button is pressed, etc.). If the tests pass, we could probably merge this and let that as a separate issue, as for existing scenarios there is no change.

src/Uno.UI/UI/Xaml/Application.cs
```diff
@@ -185,35 +185,9 @@ internal void SetRequestedTheme(ApplicationTheme requestedTheme)
...
-				else if (instance is ViewGroup g)
-				{
-					foreach (object o in g.GetChildren())
-					{
-						PropagateThemeChanged(o);
-					}
-				}
-			}
+			FrameworkElement.PropagateThemeScope(
+				Windows.UI.Xaml.Window.Current.Content,
```
davidjohnoliver, line 189: pretty edge-case scenario, but what happens if by some chance RequestedTheme has been set on `Windows.UI.Xaml.Window.Current.Content`?

davidjohnoliver (approved):

src/Uno.UI/UI/Xaml/Application.cs
MartinZikmund: That is actually covered - the second argument for the `PropagateThemeScope` method just says what is the requested theme of the outside scope, if any. Inside the `UpdateActualTheme` method is called, which checks the processed element's `RequestedTheme` and gives it precedence.

MartinZikmund:
Oh ok, maybe there are still issues 😂😂😂
image

MartinZikmund:
@davidjohnoliver Will we be able to merge this in some form so the element theme support can be worked on further?

davidjohnoliver:
@MartinZikmund We should be able to merge this, providing it doesn't introduce any regressions.

'No regressions' probably means preserving the current special-casing for `Window.Current.Root`, for which changing the `RequestedTheme` just changes the whole app theme behind the scenes, at least until we're sure that the current element-level logic fully matches it.

jeromelaban:
Is there something that changed about this PR, particularly the last comment ?

MartinZikmund:
> Is there something that changed about this PR, particularly the last comment ?

@jeromelaban Ah, sorry, no. I didn't get to it since TreeView/TabView. Once I finish prepping TabView for merging, I will continue with this one to rebase on master and make sure to have it as a "alternative" for now, not to break the global setting.

Issue #3409, opened 2020-06-19
Numeric element IDs in WASM are not compatible with CSS


MartinZikmund:
## What would you like to be added:

Uno currently generates numeric IDs for elements, which however cannot be referenced by CSS rules normally:

For example this does not work:

    #423506::selection { ... } 

While this does:

    [id='423506']::selection { ... }

## Why is this needed:

To be able to style pseudo-selectors for elements.

## For which Platform:

- [ ] iOS
- [ ] Android
- [x] WebAssembly
- [ ] WebAssembly renderers for Xamarin.Forms
- [ ] macOS
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

carldebilly:
Hi @MartinZikmund,

  I agree they are not CSS compliant.  But the real question is: why would you want to target those IDs? They are memory offset for their conterparts in managed code (GC Handles). You should not hard-code CSS files on those IDs.

MartinZikmund:
The issue I am facing is I need to style `::selection` of a specific TextBox, which cannot be done with in line style. Is there a way to do this without directly using the ID?

carldebilly:
You can dynamically add a CSS class to the control and target it using this class.

francoistanguay:
Should this be closed or still valid?

MartinZikmund:
I would still suggest making the IDs start with a letter - could be something like `uno-423506` for example to ensure the result is a valid markup

Youssef1313:
This should be closed per https://github.com/unoplatform/uno/pull/6828#issuecomment-905432230.

PR #3410, opened 2020-06-19
Support for `PlaceholderForeground` and `SelectionHighlightColor`

### LINKED ISSUE #3406: Support PlaceholderForeground and SelectionHighlightColor in TextBox

Support for `PlaceholderForeground` and `SelectionHighlightColor` properties.

## Why is this needed:

UWP parity

## For which Platform:

- [x] iOS
- [x] Android
- [x] WebAssembly
- [x] macOS
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

MartinZikmund:
GitHub Issue (If applicable): closes #3406 

## PR Type

- Feature

Not supported.

## What is the new behavior?

- `PlaceholderForeground` supported everywhere.
- `SelectionHighlightColor` supported on iOS, Android and WASM.

## PR Checklist


MartinZikmund:
@carldebilly here is why I needed #3409 - see changes in `TextBox.wasm.cs` and `WindowManager.ts`. The solution is kinda messy now, I would prefer something cleaner, but there is no way to style `::selection` pseudoelement with inline style (or I didn't find it). If there are any alternatives, I will be happy to use them instead of the "hack" I tried so far 😀

src/Uno.UI.Wasm/ts/WindowManager.ts
```diff
@@ -584,6 +584,26 @@ namespace Uno.UI {
...
+			let styleElement = document.getElementById(styleElementId) as HTMLStyleElement;
+			if (!styleElement) {
+				styleElement = document.createElement("style");
+				styleElement.id = styleElementId;
+				styleElement.type = "text/css";
+				document.body.appendChild(styleElement);
+			}
+			styleElement.innerHTML = "";
+			var styleText = document.createTextNode(cssText);
+			styleElement.appendChild(styleText);
```
carldebilly, line 597: Try to add a class instead.

MartinZikmund: Will do. But in general, do you see some cleaner way to implement this than generating new `<style>` for each `TextBox` and deleting it afterwards? I can't find any alternative...

carldebilly: As I said... you can just add a CSS class to the element for that instead of generating new styles dynamically.

MartinZikmund: Im sorry, I'm not sure I understand 🤔. I need the `::selection` pseudoselector to target only the selected text of the input, which seems to require a full CSS rule, as far as I found. If I use a unique class, it makes it easier to target the specific element, but I will still need something like `.myclass::selection` and because each input can have potentially different highlight color, I may need a unique rule for each input. Am I missing something 😅?

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.wasm.cs
```diff
@@ -93,6 +95,29 @@ protected void SetIsPassword(bool isPassword)
...
 		}
 
+		partial void OnSelectionHighlightColorChangedPartial(Color color)
+		{
+			if (_textBoxView != null)
+			{
+				Color foregroundColor = Colors.White;
+
+				//check highlight color luminance to choose if black or white foreground is more appropriate
+				var luminance = (0.299 * color.R + 0.587 * color.G + 0.114 * color.B) / 255;
```
carldebilly, line 105: Put this in a helper somewhere...

carldebilly (approved):

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.cs
```diff
@@ -325,6 +328,51 @@ public string PlaceholderText
...
+			get => (SolidColorBrush)this.GetValue(SelectionHighlightColorProperty);
+			set => this.SetValue(SelectionHighlightColorProperty, value);
+		}
+
+		public static DependencyProperty SelectionHighlightColorProperty { get; } =
+			DependencyProperty.Register(
+				nameof(SelectionHighlightColor), typeof(SolidColorBrush),
+				typeof(TextBox),
+				new FrameworkPropertyMetadata(
+					default(SolidColorBrush),
```
Youssef1313, line 344: It looks like this is not the default on UWP.

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.cs
```diff
@@ -325,6 +328,51 @@ public string PlaceholderText
...
+					propertyChangedCallback: (s, e) => ((TextBox)s)?.OnSelectionHighlightColorChanged(e)));
```
Youssef1313, line 345: We may want to listen to color changes after the brush is set.

src/Uno.UI/UI/Xaml/WindowManagerInterop.wasm.cs
```diff
@@ -426,6 +426,21 @@ private struct WindowManagerSetStylesParams
...
+			WebAssemblyRuntime.InvokeJS(command);
+		}
+
+		internal static void RemoveGeneralStyle(string styleId)
+		{
+			var command = "Uno.UI.WindowManager.current.removeGeneralStyle(\"" + styleId + "\");";
+
+			WebAssemblyRuntime.InvokeJS(command);
+		}
+
```
Youssef1313, line 443: These are unused now I think?

```suggestion
```

src/Uno.UI/UI/Xaml/WindowManagerInterop.wasm.cs
```diff
@@ -426,6 +426,21 @@ private struct WindowManagerSetStylesParams
 			public string[] Pairs;
 		}
 
+		internal static void SetGeneralStyle(string styleId, string cssText)
```
jeromelaban, line 429: Would be best to use the newer interop mechanisms, same below.

src/Uno.UI/UI/Xaml/Style/Generic/Generic.xaml
```diff
@@ -232,7 +232,7 @@
 		<Setter Property="BorderBrush"
 				Value="{ThemeResource SystemControlForegroundChromeDisabledLowBrush}" />
 		<Setter Property="SelectionHighlightColor"
-				Value="{ThemeResource SystemControlHighlightAccentBrush}" />
+				Value="{ThemeResource TextControlSelectionHighlightColor}" />
```
carldebilly, line 235: The doc says it's `TextSelectionHighlightThemeColor`
https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.textbox.selectionhighlightcolor?view=winrt-22000#property-value

src/Uno.UI/UI/Xaml/WindowManagerInterop.wasm.cs
MartinZikmund: These are old parts of the code, no longer needed, so I am removing them 👍

src/Uno.UI/UI/Xaml/Style/Generic/Generic.xaml
MartinZikmund: Seems the actual default is different in Generic.xaml in Windows Kits 🤷‍♂️🤷‍♂️

src/SamplesApp/Benchmarks.Shared/Suite/Windows_UI/ColorBench/XamlControlsResourcesCreationBenchmark.cs
```diff
@@ -42,3 +43,4 @@ public void Create_Color_Equals()
 		}
 	}
 }
+#pragma warning restore CS0219
```
carldebilly, line 46: It's not really required to restore this at the end: it won't apply to other files like compilers in the old times ;-)

src/Uno.UI.Runtime.Skia.Gtk/UI/Xaml/Controls/TextBoxViewExtension.cs
```diff
@@ -425,14 +428,14 @@ public int GetSelectionLength()
 			}
 		}
 
-		public void SetForeground(Windows.UI.Xaml.Media.Brush brush)
+		public void SetForeground(Brush brush)
 		{
-			if (_currentInputWidget is not null && brush is SolidColorBrush scb)
+			if (_currentInputWidget is { } widget && brush is SolidColorBrush scb)
```
Youssef1313, line 433: They are equivalent, but I think `is not null` is more readable

```suggestion
			if (_currentInputWidget is not null && brush is SolidColorBrush scb)
```

PR #3489, opened 2020-07-05
fix: Overwrite ResourceDictionary key via indexer

### LINKED ISSUE #3486: Item in ResourceDictionary cannot be overwritten
## Current behavior

When an existing key in `ResourceDictionary` is overwritten using indexer, an exception is thrown.

## Expected behavior

Should not throw.

```
dictionary["test"] = "something";
dictionary["test"] = "something";
```

## Environment

Nuget Package: latest

Package Version(s):

Affected platform(s):

- [x] iOS
- [x] Android
- [x] WebAssembly
- [x] macOS
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

Visual Studio:

- [ ] 2017 (version: )
- [x] 2019 (version: )
- [ ] for Mac (version: )

Relevant plugins:

- [ ] Resharper (version: )

MartinZikmund:
GitHub Issue (If applicable): fixes #3486 

## PR Type

- Bugfix

`ResourceDictionary` throws when existing key is set.

## What is the new behavior?

- `ResourceDictionary` no longer throws when existing key is set.
- `Insert` method now returns `true` for existing items and `false` for non-existing items, in line with UWP.
- Lazy initialization support for `Insert`.

## PR Checklist


src/Uno.UI/UI/Xaml/ResourceDictionary.cs
```diff
@@ -64,11 +64,19 @@ public bool HasKey(object key)
...
+		/// <returns>True if calling the method replaced a value that already existed for the key;
+		/// false if calling the method defined a new key.</returns>
 		public bool Insert(object key, object value)
 		{
-			_values[key] = value;
-			return true;
-		}
+			var existed = _values.ContainsKey(key);
+			Set(key, value);
+			return existed;
```
jeromelaban, line 78: Could you add a test for that behavior as well ?

MartinZikmund: Done

MartinZikmund:
> Looking at the CI build error, it seems like this might actually be a C++-only method. What do you need it for?

Oh, you are right, it is not accessible in UWP. The method itself was already there before, I just updated the logic according to what I saw in the docs, but I didn't notice it doesn't exist in UWP at all 🤔 . We could then probably make it internal as breaking changes later. For now I can undo the changes and keep only the overwriting fix

davidjohnoliver:
@MartinZikmund According to the docs https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.resourcedictionary.insert?view=winrt-19041 the `Add()` method is the C# equivalent... does it have the same behaviour?

MartinZikmund:
> @MartinZikmund According to the docs https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.resourcedictionary.insert?view=winrt-19041 the `Add()` method is the C# equivalent... does it have the same behaviour?
> 
> > **[ResourceDictionary.Insert(Object, Object) Method (Windows.UI.Xaml) - Windows UWP applications](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.resourcedictionary.insert)**Adds a new entry to the ResourceDictionary.

@davidjohnoliver Actually no, it throws when a duplicate key is added, which matches what Uno does, so `Add` should be ok

MartinZikmund:
@davidjohnoliver Reverted the changes and added a comment that the `Insert` method can be removed as part of #2240

src/Uno.UI/UI/Xaml/ResourceDictionary.cs
```diff
@@ -129,7 +131,20 @@ public bool TryGetValue(object key, out object value, bool shouldCheckSystem)
 
 				return value;
 			}
-			set => Add(key, value);
+			set => Set(key, value);
+		}
+
+		private void Set(object key, object value)
```
davidjohnoliver, line 137: Would be good for `Add()` to call `Set()` as well to keep the `new LazyInitializer()` code together, adding a bool parameter to determine if it should throw if the key is already present. Not a big deal though. Otherwise looks good!

PR #3600, opened 2020-07-22
fix(macOS): crash when finishing animation

ajpinedam:

## PR Type

- Bugfix

macOS apps crash when rapidly clicking on a Button or Checkbox.

## What is the new behavior?
macOS apps will __not__ crash when rapidly clicking on a Button or Checkbox.

## PR Checklist

## Other information

https://github.com/unoplatform/Uno.Material/issues/229

### DIFF
--- a/src/Uno.UI/UI/Xaml/Media/Animation/Animators/GPUFloatValueAnimator.iOSmacOS.cs
+++ b/src/Uno.UI/UI/Xaml/Media/Animation/Animators/GPUFloatValueAnimator.iOSmacOS.cs
@@ -412,7 +412,7 @@ namespace Windows.UI.Xaml.Media.Animation
 				this.Log().DebugFormat("Finalizing animation for GPU Float value animator on property {0}.", _bindingPath.LastOrDefault().PropertyName);
 			}
 
-			if (_valueAnimator.IsRunning)
+			if (_valueAnimator?.IsRunning == true)
 			{
 				_valueAnimator.Cancel();
 			}
--- a/src/Uno.UI/UI/Xaml/Media/Animation/Animators/GPUFloatValueAnimator.iOSmacOS.cs
+++ b/src/Uno.UI/UI/Xaml/Media/Animation/Animators/GPUFloatValueAnimator.iOSmacOS.cs
@@ -412,7 +412,7 @@ namespace Windows.UI.Xaml.Media.Animation
 				this.Log().DebugFormat("Finalizing animation for GPU Float value animator on property {0}.", _bindingPath.LastOrDefault().PropertyName);
 			}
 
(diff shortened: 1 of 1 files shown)


src/Uno.UI/UI/Xaml/Media/Animation/Animators/GPUFloatValueAnimator.iOSmacOS.cs
```diff
@@ -412,7 +412,7 @@ private void FinalizeAnimation()
 				this.Log().DebugFormat("Finalizing animation for GPU Float value animator on property {0}.", _bindingPath.LastOrDefault().PropertyName);
 			}
 
-			if (_valueAnimator.IsRunning)
+			if (_valueAnimator?.IsRunning == true)
```
dr1rrb, line 415:
```suggestion
			if (_valueAnimator?.IsRunning ?? false)
```

I found this much easier to read :)

ajpinedam: There was a discussion about this earlier this week looking to find which one was easier to read and it's true that for the ones who know what "??" means/do this seems to be more readable but overall people reacted saying that "anyValue==true" is easier to be read as it explicitly indicates what is expected (the value to be true).

MatFillion: I agree with andres :) == true is clearer for me at least =D

hminaya: I'll add that it helps new folks without a lot of experience in C#, they feel more confident and less intimidated.

jeromelaban: The problem here is the fact that `_valueAnimator?.IsRunning` is a `int?` and that making this test with `== false` would lead to invalid results if `_valueAnimator` is null.

Even if null coalescing is more complex to understand at first, once the syntax is learnt, it avoids errors leading to understanding how default value for nullable expressions works, when generalizing this pattern.

Moreover, enabling nullable ref types in this type of scenario would avoid additional errors.

hminaya: Isn't IsRunning a boolean?

jeromelaban: Right ;) Typo, yet still the same point.

PR #3516, opened 2020-07-09
feat: Implemented Theme Qualifier for Android

pdecostervgls:
GitHub Issue (If applicable): 
#2197 (partially)

## PR Type

What kind of change does this PR introduce?
Feature

## What is the current behavior?
On Android, assets with a theme qualifier don't get retargeted to the corresponding theme folders on Android.
There was no use of theme qualifiers in the UI Samples

## What is the new behavior?
On Android, assets with a theme qualifier do get retargeted to a `drawable` folder with the `-night` qualifier added for dark theme assets.
Updated BasicThemeResources to be theme aware, and refer to an image that is specified with the theme qualifier.

## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)
- [x] [Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [ ] Validated PR `Screenshots Compare Test Run` results.
- [x] Contains **NO** breaking changes
- [ ] Associated with an issue (GitHub or internal) and uses the [automatic close keywords](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue).
- [x] Commits must be following the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) specification.

## Other information

Internal Issue (If applicable):

PR #3724, opened 2020-08-06
Revert "feat: Implemented Theme Qualifier for Android"

jeromelaban:
Reverts unoplatform/uno#3516 as this causes instabilities on iOS.


pdecostervgls:
Any idea what aspect causes the instabilities?

PR #3730, opened 2020-08-06
Documentation: sub navigation position in main.scss amended

SetTrend:
In the previous version of the documentation there was a position typo causing the sub navigation bar to be too low, so the scrolling content peeked above through the navigation header.

In this commit, I corrected that typo (`75px`) by the correct value (`57px`).

### What kind of change does this PR introduce?

- Documentation style bug fix

The sub navigation bar is positioned too low, so the scrolling content peeks through above the sub navigation bar:

Documentation styling error

## What is the new behavior?

I fixed the typo, so the sub navigation bar position matches the top navigation bar height:

Documentation styling fix

### DIFF
--- a/doc/templates/uno/styles/scss/main.scss
+++ b/doc/templates/uno/styles/scss/main.scss
@@ -178,7 +178,7 @@ h3 {
 
 .subnav {
     position: fixed;
-    top: 75px;
+    top: 57px;
     width: 100%;
     z-index: 999;
 }


SetTrend:
@MartinZikmund, @davidjohnoliver, @jeromelaban: 

Hi, this is such a minor, simple, easy documentation layout fix. Would you mind approving and merging it?

jeromelaban:
Thanks for the fix, and sorry about that we missed it! 

The checks will not pass if the commit style does not respect conventional commits (the `docs:` prefix in your case), could you rephrase it ? Thanks!

SetTrend:
Done ...

Again, let me thank you, guys, for your never-ending marvellous efforts! (Gee, I'm repeating myself 😎, but it's no less true.)

jeromelaban (approved):
Thanks ;) Glad you find this project useful!

PR #3743, opened 2020-08-07
[Wasm] Setting a Padding on TextBlock on Wasm were not working

carldebilly:
# Bugfix

Setting a `Padding` on a `<TextBlock>` on Wasm were not applied because of [an `!important` clause in the Uno.UI.css](https://github.com/unoplatform/uno/blob/a795eb779cd672c8d34c88ea65d921d0075889dc/src/Uno.UI/WasmCSS/Uno.UI.css#L29).

## What is the new behavior?
Just removed that `!important` clause.

## PR Checklist

https://nventive.visualstudio.com/Uno%20Gallery/_workitems/edit/191222

### DIFF
--- a/src/Uno.UI/WasmCSS/Uno.UI.css
+++ b/src/Uno.UI/WasmCSS/Uno.UI.css
@@ -26,7 +26,7 @@ body {
     margin & padding are calculated through Xaml elements.
   */
   margin: 0 !important;
-  padding: 0 !important;
+  padding: 0;
   line-height: normal !important;
   -webkit-box-sizing: border-box !important;
   -moz-box-sizing: border-box !important;


src/Uno.UI/WasmCSS/Uno.UI.css
```diff
@@ -26,7 +26,7 @@ body {
     margin & padding are calculated through Xaml elements.
   */
   margin: 0 !important;
-  padding: 0 !important;
+  padding: 0;
```
alextrepa, line 29: Si enlever le tag !important fait que le padding fonctionne, est-ce que ca vx dire que les marges, le line-height et les autres proprité css ici ne fonctionne pas aussi? Peut-être que c'est l'effet escompté aussi je demande par curiosité,

agneszitte-nventive: Bien vu @alextrepa, je me posais la meme question aussi ;)

davidjohnoliver: Curious +1, I was under the impression that Xaml Margin, Padding etc were handled entirely by Uno and the absolute position of the element was being set directly.

carldebilly: It was there because the HTML Padding were not supposed to be used at all in the first versions of Uno-WASM, since it's the responsibility of the layour engine.

But in the case of TextBlock, it's a legitimate reason.

PR #3759, opened 2020-08-09
fix(transform): [MacOS] Fix transform not applied

dr1rrb:
## BugFix
The `RenderTransform` is not applied on macOS views when they are set before rendering the elment.

When a `RenderTransform` is set on a `NSView` it's synchronously set on the `NSView.Layer.Transform`. But this cause a bug on Mac OS and the transform is never applied to the view.

## What is the new behavior?
We wait for the next dispatcher loop after the control has been loaded to set the `NSView.Layer.Transform`.

Note: We hide the control while deferring in order to avoid flickers.

## PR Checklist

## Other information
Internal Issue: https://nventive.visualstudio.com/_workitems/edit/191447

### DIFF
--- a/src/Uno.UI/Media/NativeRenderTransformAdapter.iOSmacOS.cs
+++ b/src/Uno.UI/Media/NativeRenderTransformAdapter.iOSmacOS.cs
@@ -2,6 +2,8 @@
 using System.Collections.Generic;
 using System.Numerics;
 using System.Text;
+using Windows.UI.Core;
+using Windows.UI.Xaml;
 using Windows.UI.Xaml.Media.Animation;
 using CoreAnimation;
 using CoreGraphics;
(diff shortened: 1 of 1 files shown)


src/Uno.UI/Media/NativeRenderTransformAdapter.iOSmacOS.cs
```diff
@@ -11,25 +13,66 @@ namespace Uno.UI.Media
...
+
+#if __MACOS__
+		private bool _isDeferring;
 
-		private CATransform3D _transform = CATransform3D.Identity;
-		private bool _wasAnimating = false;
+		private static void DeferredInitialize(object sender, RoutedEventArgs e)
+		{
+			var fwElt = (FrameworkElement)sender;
+			var adapter = fwElt._renderTransform;
```
jeromelaban, line 50: Could this be null ?

dr1rrb: Not here: the `_renderTransform` it actually itself

PR #2536, opened 2020-01-25
[Android] More functional Geolocator

pkar70:
GitHub Issue (If applicable): closes #1967, closes #5661, closes #4432

## PR Type

What kind of change does this PR introduce?
- Bugfix
- Feature

## What is the current behavior?
a) RequestAccessAsync() => GeolocationAccessStatus.Allowed;
b) GetPositionAsync waits very, very long for fix
c) some options / attributes are not used

## What is the new behavior?
a) current behaviour is correct only on API <23 - now, some checks are done before returning status.
b) Geolocator now can use any method of getting fix, in most scenario it means getting fix in seconds
c) more attributes are set in Geolocation

## PR Checklist

Please check if your PR fulfills the following requirements:
- [X] Tested code with current [supported SDKs](https://github.com/unoplatform/uno/blob/master/README.md#uno-features)
- [ ] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)
- [ ] [Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [ ] [Wasm UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md#running-the-webassembly-ui-tests-snapshots) are not showing unexpected any differences. Validate PR `Screenshots Compare Test Run` results.
- [X] Contains **NO** breaking changes
- [X] Updated the [Release Notes](https://github.com/unoplatform/uno/tree/master/doc/ReleaseNotes)
- [X] Associated with an issue (GitHub or internal)

## Other information
Part of my apps SmogMeter and KrakTram, published on Google Play and Amazon.

Internal Issue (If applicable):

PR #3891, opened 2020-08-26
fix: Correctly initialize value of GeoCoordinate.Point on Android

jcantin-nventive:

## PR Type

When using Geolocator to obtain a Geocoordinate position, Geocoordinate.Latitude and Geocoordinate.Longitude are correct but Geocoordinate.Point.Latitude and Geocoordinate.Point.Longitude (which are the correct way to obtain those values according to MSDN doc) are all set to 0, because we use a ctor which is not implemented.

## What is the new behavior?

Geocoordinate.Point is correctly set.

## PR Checklist

- [ ] ~~[Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)~~ (I have tested that the values are set correctly)

## Other information

### DIFF
--- a/src/Uno.UWP/Devices/Geolocation/Geopoint.cs
+++ b/src/Uno.UWP/Devices/Geolocation/Geopoint.cs
@@ -15,7 +15,7 @@ namespace Windows.Devices.Geolocation
 
 		public GeoshapeType GeoshapeType => GeoshapeType.Geopoint;
 
-#if __WASM__
+#if __WASM__ || __ANDROID__
 		public Geopoint(BasicGeoposition position, AltitudeReferenceSystem altitudeReferenceSystem) : this(position)
 		{
 			AltitudeReferenceSystem = altitudeReferenceSystem;
--- a/src/Uno.UWP/Generated/3.0.0.0/Windows.Devices.Geolocation/Geopoint.cs
+++ b/src/Uno.UWP/Generated/3.0.0.0/Windows.Devices.Geolocation/Geopoint.cs
@@ -8,7 +8,7 @@ namespace Windows.Devices.Geolocation
 	public  partial class Geopoint : global::Windows.Devices.Geolocation.IGeoshape
 	{
 		// Skipping already declared property Position
-		#if __ANDROID__ || __IOS__ || NET461 || false || __SKIA__ || __NETSTD_REFERENCE__ || __MACOS__
+		#if __IOS__ || NET461 || false || __SKIA__ || __NETSTD_REFERENCE__ || __MACOS__
 		[global::Uno.NotImplemented("__ANDROID__", "__IOS__", "NET461", "__SKIA__", "__NETSTD_REFERENCE__", "__MACOS__")]
 		public  global::Windows.Devices.Geolocation.AltitudeReferenceSystem AltitudeReferenceSystem
 		{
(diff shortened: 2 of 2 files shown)


src/Uno.UWP/Devices/Geolocation/Geolocator.Android.cs
```diff
@@ -125,9 +125,7 @@ public static Geoposition ToGeoPosition(this Location location)
 							Latitude = location.Latitude,
 							Longitude = location.Longitude,
 							Altitude = location.Altitude,
-						},
-						AltitudeReferenceSystem.Ellipsoid,
-						Wgs84SpatialReferenceId
+						}
```
MatFillion, line 128: constant Wgs84SpatialReferenceId was used for which reason? 

Also I suppose you can remove the private constant as it's not used anymore.

jcantin-nventive: I don't know why it was added, but all it did was break the code.

src/Uno.UWP/Devices/Geolocation/Geolocator.Android.cs
```diff
@@ -125,9 +123,7 @@ public static Geoposition ToGeoPosition(this Location location)
 							Latitude = location.Latitude,
 							Longitude = location.Longitude,
 							Altitude = location.Altitude,
-						},
-						AltitudeReferenceSystem.Ellipsoid,
-						Wgs84SpatialReferenceId
```
jeanplevesque, line 130: By removing this, are we losing some functionality (that we may not understand right now)?
Wouldn't it be better to fix the Geopoint ctor instead?

jcantin-nventive: We never had this functionality. It was added for wasm only by this commit. https://github.com/unoplatform/uno/commit/b7bba3d68e3ef597bedb8ebd892d2e4f14b059b5#diff-60dce75981e870b5855507ba793dbf87

These properties don't exist in UWP.

jeanplevesque: But if this was added for WASM, are we breaking something by removing it?

pkar70: AltitudeReferenceSystem.Ellipsoid and Wgs84SpatialReferenceId is also in #2536.

davidjohnoliver: From the history, it looks like some changes that are relevant to WASM, were mistakenly also applied to Android. Maybe @MartinZikmund can confirm but this fix looks good to me.

pkar70: Correct (full) ToGeoPosition() method SHOULD put AltitudeReferenceSystem.Ellipsoid and Wgs84SpatialReferenceId values. Current implementation (at least, Android version) should be corrected, as in my PR #2536. Also, some other changes should be done, but this PR (2536) has small chances to be merged.

davidjohnoliver: @pkar70 thanks for the clarification, on closer inspection it looks like you're right, the better fix is to make the correct ctor available for Android.

jcantin-nventive: @pkar70 @davidjohnoliver OK, the changes we discussed have been made!

MartinZikmund:
@davidjohnoliver @jcantin-nventive I added the hardcoded AltitudeReferenceSystem because the Android docs state that this is always used (see https://gis.stackexchange.com/a/75412)

PR #3893, opened 2020-08-26
feat: Throw appropriate exception when obtaining location without permission

jcantin-nventive:

## PR Type

When the permission is denied and GetGeopositionAsync is called, then a NullReferenceException is thrown.

## What is the new behavior?

When the permission is denied and GetGeopositionAsync is called, then the method throws a helpful exception, as in UWP and Android.

## PR Checklist

- [ ] ~~[Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)~~

## Other information

### DIFF
--- a/src/SamplesApp/SamplesApp.iOS/Info.plist
+++ b/src/SamplesApp/SamplesApp.iOS/Info.plist
@@ -64,5 +64,9 @@
 	</array>
 	<key>UIApplicationSupportsIndirectInputEvents</key>
 	<true/>
+	<key>NSLocationWhenInUseUsageDescription</key>
+	<string>TODO NSLocationWhenInUseUsageDescription</string>
+	<key>NSLocationUsageDescription</key>
+	<string>TODO NSLocationUsageDescription</string>
 </dict>
 </plist>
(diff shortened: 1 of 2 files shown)


src/Uno.UI/UI/Xaml/Input/PointerRoutedEventArgs.iOS.cs
```diff
@@ -61,7 +61,6 @@ public PointerPoint GetCurrentPoint(UIElement relativeTo)
 			touchType switch
 			{				
 				UITouchType.Stylus => PointerDeviceType.Pen,
-				UITouchType.IndirectPointer => PointerDeviceType.Mouse,
```
jeanplevesque, line 64: Should this be in this PR?
I think @dr1rrb worked on pointers?

jcantin-nventive: This line prevents the solution from building on iOS

davidjohnoliver: For stuff like this it's better to create an issue or contact someone on the core team directly, rather than including it in an unrelated PR where it might accidentally get committed and break functionality.

In this case it seems that `UITouchType.IndirectPointer` was added in Xamarin.iOS 13.16: https://docs.microsoft.com/en-us/xamarin/ios/release-notes/api-changes/ios-13-14-1-13-16-0, so it's probably a question of an older version of Visual Studio.

PR #3955, opened 2020-09-03
feat(ColorPicker): Add ColorPicker, ColorPickerSlider & ColorSpectrum

### LINKED ISSUE #3875: Add ColorPicker & ColorSpectrum Control
##  What would you like to be added:

Add the ColorPicker and ColorSpectrum controls from WinUI.

https://github.com/microsoft/microsoft-ui-xaml/tree/master/dev/ColorPicker
https://docs.microsoft.com/en-us/uwp/api/microsoft.ui.xaml.controls.colorpicker?view=winui-2.4

## Why is this needed:

 1. Feature parity
 2. Allows colors to be selected in Uno
 3. Opens the door for the inclusion of the ColorPickerButton in the community toolkit which depends on the ColorSpectrum

## For which Platform:

- [x] iOS
- [x] Android
- [x] WebAssembly
- [x] macOS
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

robloo:

Closes #3875 

## PR Type

- Feature

The Microsoft.UI.Xaml.Controls.ColorPicker and related primitives (ColorSpectrum, ColorPickerSlider) are not implemented.

## What is the new behavior?

The following controls are implemented:
Microsoft.UI.Xaml.Controls.ColorPicker
Microsoft.UI.Xaml.Controls.Primitives.ColorPickerSlider
Microsoft.UI.Xaml.Controls.Primitives.ColorSpectrum

## PR Checklist

## Other information

Submission containing materials of a third party: Microsoft, MIT License https://github.com/microsoft/microsoft-ui-xaml

## Notes

Architecture of the ColorPicker itself can be found here:
https://github.com/microsoft/microsoft-ui-xaml/blob/786129a37d36890d9d8ef1a61f945cc2f3be19a4/dev/ColorPicker/readme.md

* Rgb and Hsv were converted to structs and placed in their own files, take a look at ColorConversion.h to see what is in WinRT
* Rgb and Hsv structs in WinUI have lowercase fields. This did not align with C# convention so all fields were converted to uppercase properties -- which unnecessarily changed a lot of code. Since every line was re-written to C# it seems justifiable. Edit: The properties were changed back to fields so their ref could be used.
* New internal static classes ColorConversion and ColorHelpers were created. It is required in C# as methods cannot be directly in a namespace.
* ColorConversion goes in a new 'Common' directory outside of ColorPicker. This follows the organization of WinUI.
* ColorChangedEventArgs, as projected in C#, has get-only properties. In WinRT you can read/write to the properties and setting to the property directly is done in the RaiseColorChanged() method in ColorPicker. To keep this read-only externally, properties were set to 'internal set'
* The property changed callbacks in ColorPicker/ColorSpectrum were adjusted to to avoid OnColorPropertyChanged -> OnPropertyChanged -> OnColorChanged which makes little sense in C# and is only needed in C++ with the auto-generated properties. Instead OnPropertyChanged -> OnColorChanged is called (same ideas as in NumberBox). All of this is considered negligible as the properties are nearly 100% re-written for C#.
* DownlevelHelper.ToDisplayNameExists() was implemented as a stub and always returns false. It automatically disables using ColorHelper.ToDisplayName (which isn't implemented in Uno) and it provides a straightforward mechanism to enable this in the future without any code changes. The base ColorPicker code will still match WinUI.
* On RS2 or above the ColorPicker actually uses it's own SpectrumBrush to render the spectrum itself. This is disabled and I'm falling back to the byte array -> WritableBitmap -> ImageBrush methodology which should work just fine in Uno. If Uno ever supports Composition this could be re-enabled for added performance. This means SpectrumBrush.cs is not in this port.
* The threading in ColorSpectrum/ColorHelper was simply ported to C# it does not always use the standard async/await/Task patterns of C#.
* IColorChangedEventArgs may not be required as it isn't projected into C# normally.
* GetImpl() was not added to the AutomationPeers. It is not clear why this exists in NumberBox.
* String literals were removed from dependency property registration and nameof() is used instead. This is different from other controls like NumberBox but convention is so strong here other controls should probably be using nameof() as well.
* Resource lookup is not using string literals directly. It maintains the 'key' variables from WinUI. Again other ported controls should probably be updated to go back to this convention. New keys can be added to the ResourceAccessor.ResourceKeys.cs partial class.
* The code is 1-1 converted C++ -> C# in all places possible. Where it had to re-write in some places I commented out the C++ code nearby so you can compare in the future. This is especially important when porting in future changes from WinUI to know where the code goes as it was replaced with a C# implementation. Codacy doesn't like that commented out code.
* 'this.' was being used in code inconsistently in the C# port. In C++ it isn't used at all. Now 'this.' is only used for public properties and events. It may need to be removed entirely in the future.

## Questions

1. What to do about strings in string resources that have a different format than .net String.Format() accepts. Should these be converted? Is there a string formatter that is equivalent?
1. Is disabling the ContractXPresent/NotPresent code in XAML still required?
   * It shouldn't be necessary to comment it out with Uno 3.0, the ContractPresent/NotPresent markup should work correctly.
1. The existing NumberBox is disposing of the event handlers in a strange way. What is the proper method of detaching event handlers (for template part controls) for ported code?
   * https://github.com/unoplatform/uno/pull/3955#issuecomment-687320676


davidjohnoliver:
Very impressive! And the thorough PR documentation is a great help. 👍 

Your questions:
> What to do about strings in string resources that have a different format than .net String.Format() accepts. Should these be converted? Is there a string formatter that is equivalent?
Not sure off the top of my head. Do you have an example?

> Is disabling the ContractXPresent/NotPresent code in XAML still required?
It shouldn't be required any more.

PS: re the Codacy tool - for C++-converted code, take the warnings with a grain of salt; fidelity with the original code is preferable to following C# stylistic best practices.

src/Uno.UI/Microsoft/UI/Xaml/Controls/ColorPicker/ColorHelpers.cs
```diff
@@ -0,0 +1,641 @@
...
+						incrementAmount = amount == IncrementAmount.Small ? 1 : 10;
+						break;
+
+					case ColorPickerHsvChannel.Value:
+						valueToIncrement = newHsv.V;
+						incrementAmount = amount == IncrementAmount.Small ? 1 : 10;
+						break;
+
+					default:
+						throw new Exception("Invalid ColorPickerHsvChannel."); // Uno Doc: 'winrt::hresult_error(E_FAIL);'
```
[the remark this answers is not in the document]
davidjohnoliver: `InvalidOperationException` is fine here.

src/Uno.UI/Uno.UI.csproj
```diff
@@ -153,6 +153,8 @@
 		<Service Include="{508349b6-6b84-4df5-91f0-309beebad82d}" />
 	</ItemGroup>
 	<ItemGroup>
+		<Folder Include="Microsoft\UI\Xaml\Controls\Common\" />
```
davidjohnoliver, line 156: Remove the changes to Uno.UI.csproj from the PR (and you can safely revert them locally too) - they get added when you include files via Visual Studio, but really they're unnecessary, because files are picked up by globbing.

davidjohnoliver (commented):
The imported code looks good as best as I can judge - the thorough importation notes are much appreciated.

It looks like the build is failing on some (all?) platforms because Math.Clamp() isn't available in .NET Standard 2.0 - you can use the internal `MathEx.Clamp()` method instead.

src/Uno.UI/Uno.UI.csproj
robloo: Yes, I noticed that when I imported the Strings folder it just popped up automatically in the solution. I'll be sure to clean this up!

robloo: Concerning `InvalidOperationException`, I debated using that one or 
`InvalidEnumArgumentException`. Both don't fit 100% so I went with the generic base Exception with a note to take another look. I'll be happy to switch it, thanks for the recommendation.

robloo:
> Not sure off the top of my head. Do you have an example?

The string formatting methods can take arguments to automatically insert the specified arguments into the string. In C# this might look like "The value is {0:X2}" -> "The value is FF" in C++/WinRT the string formatting is different. Instead I have localized strings that look like:

https://github.com/unoplatform/uno/blob/938e88637a5b140b647eb688445eb255a4d786ef/src/Uno.UI/Microsoft/UI/Xaml/Controls/ColorPicker/Strings/en-us/Resources.resw#L216-L219

> PS: re the Codacy tool - for C++-converted code, take the warnings with a grain of salt; fidelity with the original code is preferable to following C# stylistic best practices.

Great! I was hoping you would say that. I ensured the code is 1-1 all places possible but where I had to re-write I commented out the C++ code nearby so you can compare in the future. Especially important when porting in future changes from WinUI to know where the code goes if it was replaced with a C# implementation. Codacy doesn't like that.

> Is disabling the ContractXPresent/NotPresent code in XAML still required?

I was actually referring to the XAML files where I noticed in NumberBox all the Contract7Present namespace usage was commented out: 
https://github.com/unoplatform/uno/blob/df53cf8ef5eeb4a14138e1e403c4633e5e036779/src/Uno.UI/Microsoft/UI/Xaml/Controls/NumberBox/NumberBox.xaml#L34

> It looks like the build is failing on some (all?) platforms because Math.Clamp() isn't available in .NET Standard 2.0 - you can use the internal MathEx.Clamp() method instead.

I'm feeling brave and will remove the clamping. I was hitting overflows in an earlier port but I think some fixes elsewhere will render these unnecessary now. It will then match the original C++ implementation as well.

I have another question:

 * The existing NumberBox is disposing of the event handlers in a strange way. What is the proper method of detaching event handlers (for template part controls) for ported code?

davidjohnoliver:
> > Is disabling the ContractXPresent/NotPresent code in XAML still required?
> 
> I was actually referring to the XAML files where I noticed in NumberBox all the Contract7Present namespace usage was commented out

I messed up the reply 😆. It shouldn't be necessary to comment it out with Uno 3.0, the ContractPresent/NotPresent markup should work correctly.

> I have another question:
> 
> The existing NumberBox is disposing of the event handlers in a strange way. What is the proper method of detaching event handlers (for template part controls) for ported code?

Let me address the broader question of events/lifetime concerns in Uno, then try to explain what's going on in `NumberBox` (as much as anything for a rough draft that I can add to the [contributor docs](https://platform.uno/docs/articles/contributing/guidelines/code-style.html#disposables)).

Event handling patterns in Uno revolve around the fact that on iOS, in particular, it's very easy to create memory leaks because [the underlying native runtime uses reference-counting for memory management](https://docs.microsoft.com/en-us/xamarin/ios/deploy-test/performance#avoid-strong-circular-references). That means that any reference cycle - like a child view holding a hard reference to a parent view - automatically creates a leak where that subtree will never get collected. This most commonly manifests when subscribing to events on child views.

The pattern to address this in Uno is typically the following:
 - subscribe to the events in OnApplyTemplate()
 - and also from Loaded
 - and unsubscribe from Unloaded

That way the reference cycle is guaranteed to be broken when the view is removed the visual tree, avoiding a leak.

It's not obligatory to use the 'disposable pattern' for this, you can subscribe+unsubscribe in a normal way if you prefer. The disposable pattern just makes it somewhat easier to avoid double subscriptions, forgotten unsubscriptions etc.

---

As for how the 'disposable pattern' in NumberBox works: It's making use of two fairly simple types, `SerialDisposable` and `CompositeDisposable`, that were pinched from [Reactive Extensions](https://github.com/dotnet/reactive). As the names imply they both implement `IDisposable` and have the following qualities:
 - whenever `SerialDisposable.Disposable` is set, the previous value of the disposable, if any, is disposed
 - whenever `CompositeDisposable` is disposed, all of its children are disposed
 - the `Add(Action action)` extension method for `CompositeDisposable` registers an action to be run when the `CompositeDisposable` is disposed

So to take a subsection of the relevant code:
```csharp
		private void InitializeTemplate()
		{
			_eventSubscriptions.Disposable = null;

			var registrations = new CompositeDisposable();

			if (this.GetTemplateChild(c_numberBoxDownButtonName) is RepeatButton spinDown)
			{
				spinDown.Click += OnSpinDownClick;
				registrations.Add(() => spinDown.Click -= OnSpinDownClick);
			}
			...

			_eventSubscriptions.Disposable = registrations;
		}
```

What that code is doing is:
1. Unsubscribing any previous event subscriptions (`_eventSubscriptions.Disposable = null`)
2. Subscribing to the `OnSpinDownClick` event
3. Creating a handler that will unsubscribe from the event when disposed (`registrations.Add(() => spinDown.Click -= OnSpinDownClick)`)
4. Registering that handler (and any others) for later disposal (`_eventSubscriptions.Disposable = registrations`)

Sorry for the long reply, but I'm probably going to turn this comment into a section in the docs 😄

robloo:
> It's not obligatory to use the 'disposable pattern' for this, you can subscribe+unsubscribe in a normal way if you prefer. The disposable pattern just makes it somewhat easier to avoid double subscriptions, forgotten unsubscriptions etc.

I normally clean up the event handlers and have a note to get that done. However, looking at NumberBox I realized there might be more to it. Thanks so much for your detailed response! I was planning on unsubscribing one-by-one in the Unloaded event. However, using the reactive pattern is definitely helpful so you can remove all at once. It should simplify a lot of code and help minimize chances of mistakes. I'll take a look at what makes the most sense here.

robloo:
@davidjohnoliver I have been testing this with Android and the color spectrum is not rendering at all and the control is zero size. The other components of the ColorPicker seem to appear just fine though. As the ColorSpectrum was working fine in C# this is either a problem with the Uno Platform or a simple mistake importing the code into Uno. Either way, I can't seem to find any errors being reported to discover what the problem might be. I think I need someone with in-depth knowledge of Uno to take a look at what the problem might be. If you have a chance to look at this please see ColorPicker_SimpleText in the sample app.

robloo:
@davidjohnoliver I wonder if the following is even supported? The workItemHandler code is never being called. There are no errors and this works fine in UWP.

```
			WorkItemHandler workItemHandler = 
			(IAsyncAction workItem) =>
			{
				
			};

			m_createImageBitmapAction = ThreadPool.RunAsync(workItemHandler);
			m_createImageBitmapAction.Completed = new AsyncActionCompletedHandler(
			(IAsyncAction asyncInfo, AsyncStatus asyncStatus) =>
			{

			});
```

Looking in the ThreadPool class it seems all RunAsyc() methods are not implemented. Can this be implemented? What is the suggested work-around?

davidjohnoliver:
@robloo Indeed seems that `ThreadPool` is not implemented on Uno.

It should probably be pretty easy to implement using standard `Task.Run()` - there's functionality already in Uno for [converting to a Task from an IAsyncAction](https://github.com/unoplatform/uno/blob/master/src/Uno.UWP/System.Runtime/WindowsRuntimeSystemExtensions.cs#L22) and [vice versa](https://github.com/unoplatform/uno/blob/master/src/Uno.Foundation/AsyncAction.cs#L19).

robloo:
@davidjohnoliver Implementing a ThreadPool -- not just doing a hack to get it working in this case -- seems complex by my understanding. I expect it would also take quite a bit of time to test and ensure this is done correctly for all cases that may use the ThreadPool in the future.

robloo:
I don't want to devote the time to implement a thread pool for this. As a work around for the ThreadPool I am using standard Task.Run() in the ColorPicker itself. This has downsides such as no cancellation -- which I could add but it becomes a larger re-write. 

I'm also noticing a few bugs in related controls. Should I file separate issues for these?

 * Slider thumb doesn't appear. There is no way to know selected value on both the third dimension and alpha sliders.
 * The ComboBox does not select the first item
 * The 'More' button is not styled correctly

example1

davidjohnoliver:
Fantastic!

> I don't want to devote the time to implement a thread pool for this. As a work around for the ThreadPool I am using standard Task.Run() in the ColorPicker itself. This has downsides such as no cancellation -- which I could add but it becomes a larger re-write.

Sure, that's fine for this PR.

>I'm also noticing a few bugs in related controls. Should I file separate issues for these?

Yes, you can file issues and they can be tackled separately. 

Again, great work!

robloo:
@davidjohnoliver The ColorPicker isn't functional on macOS. The failure occurs when applying the default, implicit style of the ColorPicker itself. Debugging this, the ControlTemplate is being applied and it gets to the point where a ScrollViewer is used in a control template -- of a TextBox (RedTextBox) it looks like. The ColorPickerItself doesn't use a ScrollViewer itself. Regardless, in ScrollViewer.ApplyScrollContentPresenterContent() null appears as the Content and eventually results in failure in the native ScrollViewer implementation. Who is the best person to take a look at this? Code indicates https://github.com/unoplatform/uno/issues/3762 might be the root problem.

I'm also getting some additional failures on the CI checks -- not sure what the issues might be there.

davidjohnoliver:
> @davidjohnoliver The ColorPicker isn't functional on macOS. The failure occurs when applying the default, implicit style of the ColorPicker itself. Debugging this, the ControlTemplate is being applied and it gets to the point where a ScrollViewer is used in a control template -- of a TextBox (RedTextBox) it looks like. The ColorPickerItself doesn't use a ScrollViewer itself. Regardless, in ScrollViewer.ApplyScrollContentPresenterContent() null appears as the Content and eventually results in failure in the native ScrollViewer implementation. Who is the best person to take a look at this? 

@ajpinedam any idea what might be going wrong with the macOS `ScrollViewer` here?

robloo:
@davidjohnoliver @ajpinedam Does TextBox work at all on macOS? I had a thought to test it out in the sample app by itself and nothing seems to work. I'm getting the same errors seen the ColorPicker template.

davidjohnoliver:
`TextBox` should be supported on macOS, yes. @ajpinedam could this be a regression related to https://github.com/unoplatform/uno/pull/3966 ?

@robloo For the remaining CI errors, I made a comment where there's Xaml that's causing the UWP SamplesApp to fail.

Apart from that, it seems to be failing on the transformation step that builds the Uno.WinUI package (with WinUI 3-compatible namespaces). I don't really understand why but it doesn't seem to find the ColorHelper.ToDisplayName() method.

Can you try cut-pasting the method here:
https://github.com/unoplatform/uno/blob/master/src/Uno.UWP/Generated/3.0.0.0/Windows.UI/ColorHelper.cs#L12

to the authored file here? https://github.com/unoplatform/uno/blob/master/src/Uno.UWP/UI/ColorHelper.cs

(I'm not asking you to implement the method, just to move the unimplemented method to the authored file, so that hopefully the CI can see it)

robloo:
@davidjohnoliver Thanks for taking a look at this:

>  I made a comment where there's Xaml that's causing the UWP SamplesApp to fail.

Yes, I at first got a XAML error that a UserControl cannot host content directly -- I forget the error exactly. So I looked at the NumberBox code and saw it was using [xamarin:Grid instead of Grid directly](https://github.com/unoplatform/uno/blob/a9aff5c4dd3ea9897ec644c77352040cd18d4b9e/src/SamplesApp/UITests.Shared/Microsoft_UI_Xaml_Controls/NumberBoxTests/MUX_Test.xaml#L14). Therefore, I did the same and the code worked fine. However, this same code appears to cause the issue when building for WinUI 3. I guess the question is: Can a grid now be used directly? And if that is causing the error, why does this code work for NumberBox? 

> Apart from that, it seems to be failing on the transformation step that builds the Uno.WinUI package (with WinUI 3-compatible namespaces). I don't really understand why but it doesn't seem to find the ColorHelper.ToDisplayName() method.

I just double checked the code paths. All usages of  ToDisplayName() are contingent on DownlevelHelper.ToDisplayNameExists() being true. This is [hard-coded](https://github.com/unoplatform/uno/blob/3dfc1aefdf15e4aec1250f7dfaab3bc3d9532f55/src/Uno.UI/Helpers/WinUI/DownlevelHelper.cs) to always return false until it is actually implemented by Uno (after WinUI 3 source code is publicly available I would presume). So how is the WinUI 3 build hitting this code path? It must have it's own implementation of ToDisplayNameExists() that is returning true? I don't believe I can track this down on my end. I can try copying over the method stub but even if that works the problem seems to be somewhere else.

davidjohnoliver:
@robloo 

1. Re `Grid`: With/without the prefix, you are setting the Grid 'directly' as the UserControl Content (which certainly works). Is it possible you were seeing a false positive error from the Intellisense?

The reason to put the `xamarin` prefix is to exclude the Grid and its contents from building on UWP - @jeromelaban might recall why that was necessary for the `NumberBox` sample. In any case, it works for the `NumberBox` sample because `xamarin` is included in the `mc:Ignorable` list.

In your case, remove the `xamarin` prefix and if the SamplesApp builds for UWP, all's well.

2. It's failing because it can't find `ToDisplayName()` at compile-time, not because it's hitting that code path at runtime. The WinUI build step sees a different set of generated `NotImplemented` stubs. I don't really understand why it's not finding that method, but copying it to the authored file should fix it.

robloo:
@davidjohnoliver Thanks, all makes sense now and I'll make the updates later today and test. Apologies for the confusion about the xamarin namespace -- I'll be up to speed with cross-platform xaml eventually.

robloo:
@davidjohnoliver I'm guessing the problem with ToDisplayName() was occurring because the Color was fully qualified. The WinUI conversion tool was then not converting it to the MUX namespace and no method was matching the expected signature. So I changed the method signature to `public static string ToDisplayName(Color color)` which uses the `using Color = global::Windows .UI.Color;` in the file https://github.com/unoplatform/uno/blob/master/src/Uno.UWP/UI/ColorHelper.cs

davidjohnoliver:
Looking good on iOS!

image

robloo:
@davidjohnoliver Thanks for testing this on iOS! That's not a setup I have so it's a huge help. Since I have looked at this control so much a few questions/issues do stand out to me :) 

 1) First, do fluent styles work? The slider thumb should be a black circle in light theme.
 2) Does the more button work or does it crash when adding TextBox to the visual tree? (guessing that's still a macOS-only bug unrelated to this control)
 3) The slider background gradient origins are somehow incorrect. There is a bug in iOS here it looks like as the gradient should not have a diagonal slant to it -- should be perfectly uniform horizontally.

Finally, any idea how to fix the last CI issues related to the ColorPicker? I have no idea how 'using System' or 'System.Array' is getting converted to 'using Microsoft.System' or 'Microsoft.System.Array'

src\Uno.UI\Microsoft\UI\Xaml\Controls\ColorPicker\ArrayList.cs#L20
src\Uno.UI\Microsoft\UI\Xaml\Controls\ColorPicker\ArrayList.cs(20,13): Error CS0234: The type or namespace name 'Array' does not exist in the namespace 'Microsoft.System' (are you missing an assembly reference?)

When this is closed what else is needed to merge this branch?

davidjohnoliver:
@robloo 1. 2. Yep, Fluent style works, and 'More' button seems to work:
image

3. I see that now in your Android screenshots. I'll create a bug for that one.

davidjohnoliver:
> Finally, any idea how to fix the last CI issues related to the ColorPicker? I have no idea how 'using System' or 'System.Array' is getting converted to 'using Microsoft.System' or 'Microsoft.System.Array'

I thought I'd made a comment regarding this but it looks like I somehow managed to edit one of your comments 🤦 

It's because the WinUI 3 build step adds types that are in the Microsoft.System namespace. That makes the namespace resolution try to look for System.Array in Microsoft.System (because the code is in the Microsoft.* namespace). The WinUI 3 build step is kinda tedious 😕 

The fix is to use `global::System.Array` everywhere instead of `System.Array` in the `ArrayList` file.

I think once the build passes it's good to merge - the key boxes are all ticked (clean code, samples are present, control is working well). There are known issues but it's fine to tackle them after this PR is merged.

robloo:
@davidjohnoliver

> It's because the WinUI 3 build step adds types that are in the Microsoft.System namespace. That makes the namespace resolution try to look for System.Array in Microsoft.System (because the code is in the Microsoft.* namespace). The WinUI 3 build step is kinda tedious

Ok, I'll fully qualify those types. I should get in the habit of trying that every time! 

I'm looking forward to WinUI 3 (or more likely 3.x) when we can all clean up our codebases and use consistent namespaces and .net 5/6. A lot of this headache will go away. Until then, I'm glad Uno support both WinUI 2/3 as it will be a while before WinUI 3 supports all the necessary features to transition over completely.

davidjohnoliver (approved):

src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
```diff
@@ -336,16 +336,27 @@ internal void EnsureTemplate()
...
 				?? FindName(childName) as DependencyObject;
 		}
 
+		/// <summary>
+		/// Finds a realized element in the control template of the specified type.
+		/// </summary>
+		/// <typeparam name="T">The type of the template part.</typeparam>
+		/// <param name="childName">The name of the template part.</param>
+		/// <returns>The first template part of the specified name; otherwise, null.</returns>
+		public T GetTemplateChild<T>(string childName) where T : class, DependencyObject
```
davidjohnoliver, line 355: This is a nice idea, but it should be an internal method. We want to avoid adding stuff that isn't part of the UWP contract, even helpful stuff like this, because it allows people to write code that compiles on Uno targets but not UWP.

```suggestion
		internal T GetTemplateChild<T>(string childName) where T : class, DependencyObject
```

robloo: Yes, you're absolutely right I missed that when it was copy-pasted from my private code base.

davidjohnoliver (approved):

robloo:
@davidjohnoliver I see you already made the internal change, thanks. I also updated to the latest WinUI version to get some important bug fixes. On my end I'm complete, so it's ready whenever the Uno Team is ready to merge.

davidjohnoliver:
It's in! Thanks again @robloo for this great contribution. :)

robloo:
Awesome! Thanks so much for all your guidance and help. I really appreciate all that you and the Uno Team has accomplished - it is no small feat. It's also a huge time saver for me going forward with cross-platform development.

jeromelaban:
Thanks for the contribution!!

MelbourneDeveloper:
@robloo Hi. Would we expect this to be in the current release 3.4? I'm not seeing the ColorPicker control on Android or other platforms...

robloo:
Hi @MelbourneDeveloper, yes this should be in current 3.4. Both iOS and Android were tested to work (however slow) when the code was added.

Have you added the WinUI library to your project? This control is only available under the new Microsoft.UI.Xaml namespaces NOT the Windows.UI.Xaml namespace of vanilla UWP.

If thats not it, it might be a regression so please file a new issue.

Issue #4009, opened 2020-09-09
[iOS] Geolocator.ctor requests location permission


jcantin-nventive:
## Current behavior

When the geolocator's constructor is invoked on iOS, the location permission is immediately requested from the user. This is inconsistent with UWP, where the permission is not asked in the constructor, but rather upon RequestAccessAsync.

## Expected behavior

When the geolocator's constructor is invoked on iOS, the location permission is not requested from the user.

## How to reproduce it (as minimally and precisely as possible)

var geolocator = new Geolocator();

## Environment

Nuget Package:

- [x] Uno.UI / Uno.UI.WebAssembly / Uno.UI.Skia
- [ ] Uno.WinUI / Uno.WinUI.WebAssembly / Uno.WinUI.Skia
- [ ] Uno.SourceGenerationTasks
- [ ] Uno.UI.RemoteControl / Uno.WinUI.RemoteControl
- [ ] Other: 

Nuget Package Version(s): 3.1.0-dev.231

Affected platform(s):

- [x] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] WebAssembly renderers for Xamarin.Forms
- [ ] macOS
- [ ] Skia
  - [ ] WPF
  - [ ] GTK (Linux)
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

IDE:

- [ ] Visual Studio 2017 (version: )
- [x] Visual Studio 2019 (version: 16.4.3)
- [ ] Visual Studio for Mac (version: )
- [ ] Rider Windows (version: )
- [ ] Rider macOS (version: )
- [ ] Visual Studio Code (version: )

Relevant plugins:

- [ ] Resharper (version: )

jcantin-nventive:
Fixed with PR 4010

PR #4010, opened 2020-09-09
fix: [iOS] Do not ask permission in Geolocator.ctor

jcantin-nventive:

https://github.com/unoplatform/uno/issues/4009

## PR Type

When the geolocator's constructor is invoked on iOS, the location permission is immediately requested from the user. This is inconsistent with UWP, where the permission is not asked in the constructor, but rather upon RequestAccessAsync.

## What is the new behavior?

When the geolocator's constructor is invoked on iOS, the location permission is not requested from the user.

## PR Checklist

- [ ] ~~[Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)~~

## Other information

### DIFF
--- a/src/Uno.UWP/Devices/Geolocation/Geolocator.iOSmacOS.cs
+++ b/src/Uno.UWP/Devices/Geolocation/Geolocator.iOSmacOS.cs
@@ -21,10 +21,6 @@ namespace Windows.Devices.Geolocation
 				DesiredAccuracy = DesiredAccuracy == PositionAccuracy.Default ? 10 : 1,
 			};
 
-#if __IOS__ //required only for iOS
-			_locationManager.RequestWhenInUseAuthorization();
-#endif
-
 			_locationManager.LocationsUpdated += _locationManager_LocationsUpdated;
 
 			_locationManager.StartUpdatingLocation();
--- a/src/SamplesApp/SamplesApp.iOS/Info.plist
+++ b/src/SamplesApp/SamplesApp.iOS/Info.plist
@@ -49,8 +49,12 @@
 		<key>NSExceptionMinimumTLSVersion</key>
 		<string>TLSv1.2</string>
 	</dict>
-    <key>NSMotionUsageDescription</key>
-    <string>This is a step counter, and we&apos;d like to track motion. </string>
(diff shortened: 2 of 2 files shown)


src/Uno.UWP/Devices/Geolocation/Geolocator.iOSmacOS.cs
```diff
@@ -21,10 +21,6 @@ public Geolocator()
 				DesiredAccuracy = DesiredAccuracy == PositionAccuracy.Default ? 10 : 1,
 			};
 
-#if __IOS__ //required only for iOS
-			_locationManager.RequestWhenInUseAuthorization();
```
MatFillion, line 25: You don't request it somewhere else?

jcantin-nventive: yes! it's already being asked in RequestAccessAsync, so we don't need to add it there.

src/Uno.UWP/Devices/Geolocation/Geolocator.iOSmacOS.cs
```diff
@@ -21,10 +21,6 @@ public Geolocator()
 				DesiredAccuracy = DesiredAccuracy == PositionAccuracy.Default ? 10 : 1,
 			};
 
-#if __IOS__ //required only for iOS
```
jeromelaban, line 24: @MartinZikmund do you remember why you had to add this here ?

jeanplevesque (approved):

src/Uno.UWP/Devices/Geolocation/Geolocator.iOSmacOS.cs
MartinZikmund: Originally added to support backgrounded scenarios, but we shouldn't force it on Geolocator users unless explicitly needed. Without this call, the Geolocator will still work properly when in the foreground. We can remove this for now and maybe we will have to add a platform-specific method to allow users to request background support later.

jeromelaban (approved):

PR #4147, opened 2020-09-25
feat(WASM): Enable A11y voice over capabilities

### LINKED ISSUE #1141: a11y: output aria attributes or preferably semantic elements
# What would you like to be added:

image

## Why is this needed:

- Using `divs` makes webasm+html invisible to screen readers.
- Enabling better keyboard interaction scenaros.
- Improving accessibility is the right thing to do.

## For which Platform:

- [ ] iOS
- [ ] Android
- [x] WebAssembly
- [ ] Windows

- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA

ajpinedam:
closes #1141 

## PR Type

- Feature

WASM target did not support screen reader capabilities.

## What is the new behavior?
WASM target does support A11y screen reader capabilities 

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Automation/AutomationProperties.cs
+++ b/src/Uno.UI/UI/Xaml/Automation/AutomationProperties.cs
@@ -146,9 +146,13 @@ namespace Windows.UI.Xaml.Automation
 				view.ContentDescription = (string)args.NewValue;
 			}
 #elif __WASM__
-			if (FrameworkElementHelper.IsUiAutomationMappingEnabled && dependencyObject is UIElement uiElement)
+			if (dependencyObject is UIElement uiElement)
 			{
-				uiElement.SetAttribute("xamlautomationid", (string)args.NewValue);
+				if (FrameworkElementHelper.IsUiAutomationMappingEnabled)
+				{
+					uiElement.SetAttribute("xamlautomationid", (string)args.NewValue);
+				}
+
 				uiElement.SetAttribute("aria-label", (string)args.NewValue);
 
 				var role = FindHtmlRole(uiElement);
--- a/src/Uno.UI/UI/Xaml/Automation/AutomationProperties.cs
+++ b/src/Uno.UI/UI/Xaml/Automation/AutomationProperties.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Text;
 using Uno.UI;
 using Windows.UI.Xaml.Automation.Peers;
+using Windows.UI.Xaml.Controls;
 
 namespace Windows.UI.Xaml.Automation
 {
@@ -148,6 +149,13 @@ namespace Windows.UI.Xaml.Automation
 			if (FrameworkElementHelper.IsUiAutomationMappingEnabled && dependencyObject is UIElement uiElement)
 			{
 				uiElement.SetAttribute("xamlautomationid", (string)args.NewValue);
+				uiElement.SetAttribute("aria-label", (string)args.NewValue);
+
+				var role = GetHtmlRole(uiElement);
+				if (!string.IsNullOrWhiteSpace(role))
+				{
+					uiElement.SetAttribute("role", role);
+				}
 			}
 #endif
 		}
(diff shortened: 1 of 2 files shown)


src/Uno.UI/UI/Xaml/Automation/AutomationProperties.cs
```diff
@@ -148,6 +149,13 @@ private static void OnAutomationIdChanged(DependencyObject dependencyObject, Dep
 			if (FrameworkElementHelper.IsUiAutomationMappingEnabled && dependencyObject is UIElement uiElement)
 			{
 				uiElement.SetAttribute("xamlautomationid", (string)args.NewValue);
+				uiElement.SetAttribute("aria-label", (string)args.NewValue);
+
+				var role = GetHtmlRole(uiElement);
+				if (!string.IsNullOrWhiteSpace(role))
```
dr1rrb, line 155: TIP: As it's only internal well-known values, the `GetHtmlRole` should return `null` (and actually should be renamed to `**Find**HtmlRole`) and here the comparison should only check for `null`, it would be much faster. You could also use `string.Empty` and only compare with it if you prefer, but probably a bit less performant. I think it's not a critical path here so it's does not really matter, but just a better practice IMHO.

dr1rrb (approved):

src/Uno.UI/UI/Xaml/Automation/AutomationProperties.cs
ajpinedam: Updated the PR with your suggestions. They were good ones.

src/Uno.UI/UI/Xaml/Automation/AutomationProperties.cs
```diff
@@ -148,6 +149,13 @@ private static void OnAutomationIdChanged(DependencyObject dependencyObject, Dep
 			if (FrameworkElementHelper.IsUiAutomationMappingEnabled && dependencyObject is UIElement uiElement)
 			{
 				uiElement.SetAttribute("xamlautomationid", (string)args.NewValue);
+				uiElement.SetAttribute("aria-label", (string)args.NewValue);
+
+				var role = FindHtmlRole(uiElement);
+				if (role != null)
+				{
+					uiElement.SetAttribute("role", role);
+				}
```
carldebilly, line 158: Please use the overload of `.SetAttribute()` to call it only once.  Interop with javascript are costly.

carldebilly (approved):

src/Uno.UI/UI/Xaml/Automation/AutomationProperties.cs
```diff
@@ -148,6 +149,13 @@ private static void OnAutomationIdChanged(DependencyObject dependencyObject, Dep
 			if (FrameworkElementHelper.IsUiAutomationMappingEnabled && dependencyObject is UIElement uiElement)
 			{
 				uiElement.SetAttribute("xamlautomationid", (string)args.NewValue);
+				uiElement.SetAttribute("aria-label", (string)args.NewValue);
```
jeromelaban, line 152: We should probably calls this anyways if `AutomationId` is set, regardless of IsUiAutomationMappingEnabled being true.

src/Uno.UI/UI/Xaml/Automation/AutomationProperties.cs
```diff
@@ -148,6 +149,13 @@ private static void OnAutomationIdChanged(DependencyObject dependencyObject, Dep
...
+
+				var role = FindHtmlRole(uiElement);
+				if (role != null)
+				{
+					uiElement.SetAttribute("role", role);
```
jeromelaban, line 157: Is it necessary to set `aria-label` if role is null ?

ajpinedam: In those cases, we might not have the full requirement (role + aria-label) but it would make it easy to find for errors in the future. Meaning: `arial-label` available but no `role` it's a matter of the casing.

src/Uno.UI/UI/Xaml/Automation/AutomationProperties.cs
ajpinedam: @jeromelaban  updated the PR with this suggestion

src/Uno.UI/UI/Xaml/Automation/AutomationProperties.cs
ajpinedam: @jeromelaban  updated the PR to set the `arial-label` only when a `role` is found. This way we can use the method overload suggested by @carldebilly. Two birds one shot. :)

jeromelaban (approved):

jeromelaban:
You may also want to add some documentation about what works and what does not, and how to use it.

jeromelaban (approved):

PR #4185, opened 2020-09-30
IsEnabled fixes, Routed Events fixes

### LINKED ISSUE #4183: Disabling control does not prevent events being triggered
## Current behavior

Even when a control has `IsEnabled=false`, it still triggers all manipulation, key and pointer events.

## Expected behavior

No such events should be triggered (as per UWP).

-

## Workaround

-

## Environment

Nuget Package:

- [ ] Other: 

Nuget Package Version(s):

Affected platform(s):

- [x] iOS
- [x] Android
- [x] WebAssembly
- [x] macOS
- [x] Skia
  - [x] WPF
  - [x] GTK (Linux)
  - [x] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

IDE:

- [ ] Rider macOS (version: )

Relevant plugins:

- [ ] Resharper (version: )

### LINKED ISSUE #4184: SelectorItem disabling does not disable pointer states
## Current behavior

When `SelectorItem.IsEnabled` is set to `false`, the pointer hover/pressed states are still active.

## Expected behavior

Only selection should be applicable to match UWP.

Set `ListViewItem.IsEnabled = false` and observe the behavior.

## Workaround

-

## Environment

Nuget Package:

- [ ] Other: 

Nuget Package Version(s):

Affected platform(s):

- [x] iOS
- [x] Android
- [x] WebAssembly
- [x] macOS
- [x] Skia
  - [x] WPF
  - [x] GTK (Linux)
  - [x] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

IDE:

- [ ] Visual Studio 2019 (version: )
- [ ] Rider macOS (version: )

Relevant plugins:

- [ ] Resharper (version: )

MartinZikmund:
GitHub Issue (If applicable): fixes #4184, fixes #4183 

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/Primitives/SelectorItem.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Primitives/SelectorItem.cs
@@ -134,7 +134,7 @@ namespace Windows.UI.Xaml.Controls.Primitives
 
 		private void UpdateCommonStates(ManipulationUpdateKind manipulationUpdate = ManipulationUpdateKind.None)
 		{
-			var state = GetState(IsSelected, IsPointerOver, HasPointerCapture);
+			var state = GetState(IsEnabled, IsSelected, IsPointerOver, HasPointerCapture);
 
 			// On Windows, the pressed state appears only after a few, and won't appear at all if you quickly start to scroll with the finger.
 			// So here we make sure to delay the beginning of a manipulation to match this behavior (and avoid flickering when scrolling)
@@ -150,7 +150,7 @@ namespace Windows.UI.Xaml.Controls.Primitives
 				// When clicked (i.e. pointer released), but not yet in pressed state, we force to go immediately in pressed state
 				// Then we let the standard go to state process (i.e. with delay handling) reach the final expected state.
 
-				var pressedState = GetState(IsSelected, IsPointerOver, isPressed: true);
+				var pressedState = GetState(IsEnabled, IsSelected, IsPointerOver, isPressed: true);
 				_currentState = pressedState;
 				VisualStateManager.GoToState(this, pressedState, true);
 
@@ -200,30 +200,42 @@ namespace Windows.UI.Xaml.Controls.Primitives
 			}
 		}
 
-		private string GetState(bool isSelected, bool isOver, bool isPressed)
-		{
+		private string GetState(bool isEnabled, bool isSelected, bool isOver, bool isPressed)
+		{			
 			var state = CommonStates.Normal;
-			if (isSelected && isPressed)
+
+			if (isEnabled)
 			{
-				state = CommonStates.PressedSelected;
+				if (isSelected && isPressed)
+				{
+					state = CommonStates.PressedSelected;
+				}
+				else if (FeatureConfiguration.SelectorItem.UseOverStates
+					&& isSelected && isOver)
+				{
+					state = CommonStates.OverSelected;
+				}
+				else if (isSelected)
+				{
+					state = CommonStates.Selected;
+				}
+				else if (isPressed)
+				{
+					state = CommonStates.Pressed;
+				}
+				else if (FeatureConfiguration.SelectorItem.UseOverStates
+					&& isOver)
+				{
+					state = CommonStates.Over;
+				}
 			}
-			else if (FeatureConfiguration.SelectorItem.UseOverStates
-				&& isSelected && isOver)
+			else
 			{
-				state = CommonStates.OverSelected;
-			}
-			else if (isSelected)
-			{
-				state = CommonStates.Selected;
-			}
-			else if (isPressed)
-			{
-				state = CommonStates.Pressed;
-			}
-			else if (FeatureConfiguration.SelectorItem.UseOverStates
-				&& isOver)
-			{
-				state = CommonStates.Over;
+				// If item is disabled, we only care about selection state
+				if (isSelected)
+				{
+					state = CommonStates.Selected;
+				}
 			}
 
 			return state;
--- a/src/Uno.UI/UI/Xaml/UIElement.RoutedEvents.cs
+++ b/src/Uno.UI/UI/Xaml/UIElement.RoutedEvents.cs
@@ -610,11 +610,8 @@ namespace Windows.UI.Xaml
 				{
 					if (!isHandled || handler.HandledEventsToo)
 					{
-						if (IsEnabledOverride())
-						{
-							InvokeHandler(handler.Handler, args);
-							isHandled = IsHandled(args);
-						}
+						InvokeHandler(handler.Handler, args);
+						isHandled = IsHandled(args);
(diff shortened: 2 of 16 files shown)


src/Uno.UI/UI/Xaml/Controls/Primitives/SelectorItem.cs
```diff
@@ -134,7 +134,7 @@ private void UpdateCommonStatesWithoutNeedsLayout(ManipulationUpdateKind manipul
 
 		private void UpdateCommonStates(ManipulationUpdateKind manipulationUpdate = ManipulationUpdateKind.None)
 		{
-			var state = GetState(IsSelected, IsPointerOver, HasPointerCapture);
+			var state = GetState(IsEnabled, IsSelected, IsPointerOver, HasPointerCapture);
```
jeromelaban, line 137: @dr1rrb does this rings a bell ?

src/Uno.UI/UI/Xaml/UIElement.RoutedEvents.cs
```diff
@@ -549,7 +549,10 @@ internal bool RaiseEvent(RoutedEvent routedEvent, RoutedEventArgs args)
 				{
 					if (!IsHandled(args) || handler.HandledEventsToo)
 					{
-						InvokeHandler(handler.Handler, args);
+						if (IsEnabledOverride())
```
dr1rrb, line 552: I think this is not valid and I'm really uncomfortable to this like that.

For instance when a `Control` gets disabled we raise some pointer events (depending of the Pressed/Over/Capture state - and soon also for dragging state) (https://github.com/unoplatform/uno/blob/master/src/Uno.UI/UI/Xaml/UIElement.Pointers.cs#L53 and https://github.com/unoplatform/uno/blob/master/src/Uno.UI/UI/Xaml/UIElement.Pointers.cs#L119

But this method is also used for all `RoutedEvent`, so I would suspect that some other events can still be triggered even if the control is disabled e.g. `LostFocus`.

For the pointers, the `IsEnabled` is checked here https://github.com/unoplatform/uno/blob/master/src/Uno.UI/UI/Xaml/UIElement.Pointers.cs#L182 which is used by the hit testing. Actually you not even receive the native events.

dr1rrb: I just validated, when a control gets disabled while the pointer is over we do receive a `PointerExited`, but no `PointerReleased`. Not tested but I would assume the same for `CaptureLost`.

BTW: I think that the behavior is already correct on Skia as we dispatch pointers by ourselves.

carldebilly: 100% agree with @dr1rrb here. The enability of a control should not be considered as a filter for events.

MartinZikmund: Hm, that is weird, when I tried various pointer events over disabled button in UWP, it didn't trigger any of the `OnPointerXXX` or `OnManipulationXXX` methods, nor events. Maybe I made some mistake somewhere 🤔

MartinZikmund: I actually noticed the issue on Skia, where a disabled button still raised all of the events, which caused issues in case of TabView

MartinZikmund: Please see the sample page `RoutedEvents_IsEnabled.xaml` - there I have a custom button with an override for all routed events and whatever user does, none of them is triggered on UWP. So probably the case which needs to be handled differently is when `IsEnabled` changes? Or maybe I am missing something 😅

src/Uno.UI/UI/Xaml/Controls/Primitives/SelectorItem.cs
```diff
@@ -200,30 +200,42 @@ private void UpdateCommonStates(ManipulationUpdateKind manipulationUpdate = Mani
...
-				&& isOver)
+			else
 			{
-				state = CommonStates.Over;
+				// If item is disabled, we only care about selection state
+				if (isSelected)
+				{
+					state = CommonStates.Selected;
+				}
 			}
```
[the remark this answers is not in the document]
MartinZikmund: In fact the disabled/enabled states are already there and handled, but the problem is that even when the item is disabled, its however/pressed states apply on Skia (maybe other platforms too). However, the `IsEnabled` "filtering" could probably be sufficient, as the pointer events would not arrive for the disabled item.

src/Uno.UI/UI/Xaml/UIElement.RoutedEvents.cs
MartinZikmund: Based on [docs](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.control.isenabled?view=winrt-19041) it seems that indeed the control with `IsEnabled=false` cannot on itself raise input events:

> A disabled control is still visible to hit testing. It can't handle any input events on itself. However, a disabled control can still source the input events, and input routed events can bubble to a parent where they can be handled.

In this case, the bubbling should still apply, as the filter just prevents the events on this particular control from being raised, but the remainder of the logic remains. What do you think?

dr1rrb: Yes a disabled element should not raise pointer events, and yes the bubbling will still apply, but the issue is really the moment where we disable the control, we have to raise some events and a filter like this would mute them.

But actually I think the issue is really more that we should not even consider the element when we hit test to find the element on which we should raise the event. It's important that to properly ignore them in order to find element under disabled one for instance.

If you notice the issue on Skia then it's probably because the `HitVisibilityProperty` has not bee updated properly (or maybe just a bit to late) when the `IsEnabled` has changed. This should have been done here https://github.com/unoplatform/uno/blob/master/src/Uno.UI/UI/Xaml/UIElement.Pointers.Skia.cs#L573

I'll try to have a look to you sample soon

dr1rrb: Well after investigation, it appears that the fix you did [in `Control`](https://github.com/unoplatform/uno/pull/4185/files?diff=split&w=1#diff-7e87918895f8b453fa65dab2ed76539d91c6f5c9f896ef2962aaadd7af2e121aR58) was sufficient. We were validating the `IsEnabledOverride` in the coerced hit testability property, however as it was not override on skia, all element were always considered as enabled.

I however had add an update request on change on Skia: https://github.com/unoplatform/uno/pull/4185/files#diff-5420c68867b462946606ca55b247c946582f2d0b74028d4c47c9bd02c8eacdfdR79

dr1rrb (approved):

PR #4210, opened 2020-10-02
docs: Update conditional code/Xaml to include Skia

robloo:

## PR Type

Skia is excluded from condition code and Xaml docs.

## What is the new behavior?

Skia is included.

## PR Checklist

- [ ] ~[Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)~
- [ ] ~Validated PR `Screenshots Compare Test Run` results.~
- [ ] ~Associated with an issue (GitHub or internal) and uses the [automatic close keywords]~(https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue).

## Other information

* Someone needs to double check the pre-defined prefixes table for conditional Xaml. I added Skia for all entries and made some assumptions (like `xamarin` should exclude Skia).

### DIFF
--- a/doc/articles/platform-specific-xaml.md
+++ b/doc/articles/platform-specific-xaml.md
@@ -99,7 +99,7 @@ The pre-defined prefixes are listed below:
 | Prefix        | Included platforms                 | Excluded platforms                 | Namespace                                                   | Put in `mc:Ignorable`? |
 |---------------|------------------------------------|------------------------------------|-------------------------------------------------------------|------------------------|
 | `win`         | Windows                            | Android, iOS, web, macOS, Skia     | `http://schemas.microsoft.com/winfx/2006/xaml/presentation` | no                     |
-| `xamarin`     | Android, iOS, web, macOS           | Windows, Skia                      | `http:/uno.ui/xamarin`                                      | yes                    |
+| `xamarin`     | Android, iOS, web, macOS, Skia     | Windows                            | `http:/uno.ui/xamarin`                                      | yes                    |
 | `not_win`     | Android, iOS, web, macOS, Skia     | Windows                            | `http:/uno.ui/not_win`                                      | yes                    |
 | `android`     | Android                            | Windows, iOS, web, macOS, Skia     | `http:/uno.ui/android`                                      | yes                    |
 | `ios`         | iOS                                | Windows, Android, web, macOS, Skia | `http:/uno.ui/ios`                                          | yes                    |
@@ -122,7 +122,7 @@ More visually, platform support for the pre-defined prefixes is shown in the bel
 | `wasm`        | ✖ | ✖ | ✖ | ✔ | ✖ | ✖ |
 | `macos`       | ✖ | ✖ | ✖ | ✖ | ✔ | ✖ |
 | `skia`        | ✖ | ✖ | ✖ | ✖ | ✖ | ✔ |
-| `xamarin`     | ✖ | ✔ | ✔ | ✔ | ✔ | ✖ |
+| `xamarin`     | ✖ | ✔ | ✔ | ✔ | ✔ | ✔ |
 | `not_win`     | ✖ | ✔ | ✔ | ✔ | ✔ | ✔ |
 | `not_android` | ✔ | ✖ | ✔ | ✔ | ✔ | ✔ |
 | `not_ios`     | ✔ | ✔ | ✖ | ✔ | ✔ | ✔ |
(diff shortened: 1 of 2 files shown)


jeromelaban (approved):

jeromelaban:
Thanks @robloo !

doc/articles/platform-specific-xaml.md
```diff
@@ -96,16 +96,40 @@ Consider the following XAML which is using the Windows Community Toolkit's [Blur
...
-| `not_wasm`    | Windows, Android, iOS, macOS | web                          | `http://schemas.microsoft.com/winfx/2006/xaml/presentation` | no                     |
-| `not_macos`   | Windows, Android, iOS, web   | macOS                        | `http://schemas.microsoft.com/winfx/2006/xaml/presentation` | no                     |
+| Prefix        | Included platforms                 | Excluded platforms                 | Namespace                                                   | Put in `mc:Ignorable`? |
+|---------------|------------------------------------|------------------------------------|-------------------------------------------------------------|------------------------|
+| `win`         | Windows                            | Android, iOS, web, macOS, Skia     | `http://schemas.microsoft.com/winfx/2006/xaml/presentation` | no                     |
+| `xamarin`     | Android, iOS, web, macOS           | Windows, Skia                      | `http:/uno.ui/xamarin`                                      | yes                    |
```
davidjohnoliver, line 102: Are you sure about this? The `xamarin` prefix includes Wasm, I would assume it also includes Skia. (Whether it *should* include them is another subject.)

doc/articles/platform-specific-xaml.md
```diff
@@ -96,16 +96,40 @@ Consider the following XAML which is using the Windows Community Toolkit's [Blur
...
+
+| Prefix        |  Win  | Droid |  iOS  |  Web  | macOS | Skia  | 
+|---------------|-------|-------|-------|-------|-------|-------|
+| `win`         | ✔ | ✖ | ✖ | ✖ | ✖ | ✖ |
+| `android`     | ✖ | ✔ | ✖ | ✖ | ✖ | ✖ |
+| `ios`         | ✖ | ✖ | ✔ | ✖ | ✖ | ✖ |
+| `wasm`        | ✖ | ✖ | ✖ | ✔ | ✖ | ✖ |
+| `macos`       | ✖ | ✖ | ✖ | ✖ | ✔ | ✖ |
+| `skia`        | ✖ | ✖ | ✖ | ✖ | ✖ | ✔ |
+| `xamarin`     | ✖ | ✔ | ✔ | ✔ | ✔ | ✖ |
```
davidjohnoliver, line 125: same question here

doc/articles/platform-specific-xaml.md
robloo: No, not sure about that and it was the main assumption I mentioned in other information of the PR. It doesnt make sense for Xamarin to include skia (wasm still uses mono), if that is how it is implemented I can change it though.

davidjohnoliver: Right, I didn't check the PR description properly.

I checked and the definition for `"xamarin"` hasn't changed, it's a blanket include that's synonymous with `"not_win"`:
https://github.com/unoplatform/uno/blob/master/src/SourceGenerators/Uno.UI.SourceGenerators/Content/Uno.UI.SourceGenerators.props#L41

You're right that it doesn't make sense that it includes Skia, and really IMO it shouldn't include Wasm either (Mono != Xamarin). It'd be a significant breaking change to adjust it, though. (And worse, a silent breaking change in many cases.)

robloo: Yea, you're right that this doesn't make sense but needs to stay for compatibility. I'll adjust so xamarin is the same as not_win.

robloo: I realize skia has just been released a few months ago and isn't documented yet. While we can't really fix Wasm, it is possible to fix Skia in this case to exclude it from Xamarin. If that broke any existing code it would be a very, very small amount using technically undocumented assumptions. It's kind-of now or never to fix this.

Otherwise, I made the updates so `xamarin` = `non_win`

davidjohnoliver: The problem is that 'existing code' includes Uno code that was not specifically written for Skia (or even Wasm), where `xamarin` was used as a synonym for 'not UWP'. 

In the medium term it's just as likely that we move toward a different mechanism for conditional Xaml, in any case, because this approach is quite limited in certain use cases (eg when you want to prefix a type that's not coming from one of the default namespaces; when you want to include >1 and exclude >1 targets, eg iOS && macOS but not Android or WASM).

davidjohnoliver (approved):

Issue #1480, opened 2019-09-04
Drag and Drop support


dr1rrb:
## What would you like to be added:
`UIElement.< DragEnter | DragLeave | DragOver | DragStarting | Drop | DropCompleted >` events, and the matching overridable methods `Control.< OnDragEnter(DragEventArgs) | OnDragLeave(DragEventArgs) | OnDragOver(DragEventArgs) | OnDrop(DragEventArgs) >`

## Why is this needed:
Cause I like to drag and drop stuff ... and it matches the UWP API.

## For which Platform:
- [x] iOS
- [x] Android
- [x] WebAssembly
- [ ] Windows

## Related
https://github.com/unoplatform/uno/issues/995 (same issue but specific to the ListView)

abrull:
Are there any updates on this?

robloo:
Funny, I just thought of checking up on this myself. It's a pretty major deficiency still in the platform.

jeromelaban:
We've been making progress on the support for pointers events, pointer capture and other related features, making the implementation of this feature closer.

Make sure to upvote this issue to make it visible!

robloo:
This seems more important now with the Skia backends for Linux/Win7. It seems it wouldn't be possible to use native drag-and-drop like Android/iOS in this case as Uno is rendering everything and managing the controls themselves.

robloo:
I need this for a macOS port as its critical to the user-experience in a desktop environment. It makes more sense at this point to just implement WinUI-style drag-and-drop rather than making platform-specific code for my use case.

I'm happy to make myself available as a resource on this. If someone from nventive that has more experience cross-platform led the development it probably makes more sense than if I tackle it only for macOS right now. But I would be happy to help adding all classes, enums, etc. where they are needed. Please just let me know how I can help!

francoistanguay:
If all goes well, @dr1rrb  is starting on it this week. Talk about timing!

robloo:
@dr1rrb I quickly outlined the events/classes used in drag/drop.

UIElement Events
 * public event DragEventHandler DragEnter;
 * public event DragEventHandler DragLeave;
 * public event DragEventHandler DragOver;
 * public event TypedEventHandler<UIElement,DragStartingEventArgs>   DragStarting;
 * public event DragEventHandler Drop;
 * public event TypedEventHandler<UIElement,DropCompletedEventArgs>   DropCompleted;

Classes

 * public delegate void DragEventHandler(object sender, DragEventArgs e);
[...]

PR #4216, opened 2020-10-04
WIP: feat(dragdrop): Update drag and drop macOS support

robloo:
Relates to #1480

### DIFF
--- a/src/Uno.UWP/ApplicationModel/DataTransfer/DataPackage.macOS.cs
+++ b/src/Uno.UWP/ApplicationModel/DataTransfer/DataPackage.macOS.cs
@@ -1,0 +1,187 @@
+﻿#if __MACOS__
+#nullable enable
+
+using System;
+using System.Collections.Generic;
+using AppKit;
+using Foundation;
+using Windows.UI.Core;
+
+namespace Windows.ApplicationModel.DataTransfer
+{
+	public partial class DataPackage
+	{
+		/* UWP has the StandardDataFormats class with properties to define certain types.
+		 * These types are mapped to macOS/iOS as shown in the below table.
+		 * The latest convention in macOS/iOS is to use Uniform Type Identifiers. 
+		 *   https://developer.apple.com/documentation/uniformtypeidentifiers
+		 * However, this is still in beta and only available in macOS 11+.
+		 * NSPasteboard types are therefore used in code until such time that UTI framework
+		 * is available everywhere.
+		 * 
+		 *  UWP                    Uniform           NSPasteboard types                      NSPasteboard types     
+		 *                         Type Identifiers  (currently used)                        (legacy, deprecated)
+		 *  ---------------------  ----------------  --------------------------------------  -------------------------
+		 *  ApplicationLink        UTType.URL        NSPasteboard.NSPasteboardTypeUrl        NSPasteboard.NSUrlType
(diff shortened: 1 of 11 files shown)


src/Uno.UWP/ApplicationModel/DataTransfer/DataPackage.macOS.cs
```diff
@@ -0,0 +1,187 @@
...
+		 */
+
+		// An Uno-internal abstraction is used to simplify native drag/drop & clipboard integration
+
+		/// <summary>
+		/// Sets the contents of the <see cref="DataPackage"/> to the native clipboard.
+		/// Any conversion between supported standard types is done automatically.
+		/// </summary>
+		/// <param name="content">The contents to set to the native cipboard.</param>
+		internal static void SetToNativeClipboard(DataPackage content)
```
dr1rrb, line 48: If in the partial class, I would have make this method non static

dr1rrb (approved):

src/Uno.UWP/Uno.csproj
```diff
@@ -74,4 +76,24 @@
 	
 	<Target Name="GetBuiltProjectOutputRecursive" Condition=" '$(TargetFramework)' == 'xamarinios10' " />
 
+	<Target Name="VS16Mac_RemoveSystemMemory" BeforeTargets="ResolveAssemblyReferences" Condition="'$(MSBuildVersion)' &gt;= '16.0' and $([MSBuild]::IsOsPlatform('OSX'))">
```
jeromelaban, line 79: This is only needed when building a head, this can be removed.

src/Uno.UWP/Uno.csproj
```diff
@@ -74,4 +76,24 @@
...
+				We remove the Reference added by the nuget targets so that ResolveAssemblyReferences
+				is properly adding the local System.Memory to the Reference item group.
+		-->
+		<ItemGroup>
+			<_ReferenceToRemove Include="@(Reference)" Condition="'%(Reference.Identity)'=='System.Memory'" />
+			<Reference Remove="@(_ReferenceToRemove)" />
+		</ItemGroup>
+	</Target>
+
+	<Target Name="VS16_RemoveSystemMemory" BeforeTargets="FindReferenceAssembliesForReferences" Condition="'$(MSBuildVersion)' &gt;= '16.0' and ($(IsMonoAndroid) or $(IsXamarinIOS) or $(IsXamarinMac))">
```
jeromelaban, line 92: Same for this.

src/Uno.UWP/Uno.csproj
dr1rrb: https://dev.azure.com/uno-platform/Uno%20Platform/_build/results?buildId=20386&view=logs&jobId=0857e763-0135-5534-86be-f14b16d15a21&j=a5e52b91-c83f-5429-4a68-c246fc63a4f7&t=e7472d04-b09e-59b7-6e25-156b34b46add yes the build fails without that :)

src/Uno.UWP/ApplicationModel/DataTransfer/DataPackage.macOS.cs
robloo: The only downside is GetFromNativeClipboard() should be static. As these methods are strongly related it might make sense for other developers to use them the same way. Let's see what the final API in this area looks like after drag/drop is supported, I can certainly change it if you think best.

PR #4229, opened 2020-10-07
First <Setter> on a <Style> were potentially applying values using wrong precedence

carldebilly:
# Bugfix
There's a mechanism in Uno to force all properties set during a _property changed_ callback to be set using the `Local` precedence. This is correct for all cases except one: when this property itself is the `Style`. In this case, the _property changed_ MUST apply the `ExplicitStyle` precedence.

Details: in the previous version, while manually setting the Style property (which uses Local precedence), it was setting both _precedenceOverride to ExplicitStyle and the _isSettingAProperty to true, preventing the setters from applying there values under correct explicit precedence.

## PR Checklist

- [X] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)
- ~~[ ] Validated PR `Screenshots Compare Test Run` results.~~
- [X] Contains **NO** breaking changes **I HOPE**
- ~~[ ] Associated with an issue (GitHub or internal) and uses the [automatic close keywords]~~(https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue).

### DIFF
--- a/src/Uno.UI/UI/Xaml/DependencyObjectStore.cs
+++ b/src/Uno.UI/UI/Xaml/DependencyObjectStore.cs
@@ -103,7 +103,7 @@ namespace Windows.UI.Xaml
 
 		private static bool _validatePropertyOwner = Debugger.IsAttached;
 
-		private bool _isSettingAProperty;
+		private uint _propertySettingDepth;
 
 		/// <summary>
 		/// Provides the parent Dependency Object of this dependency object
@@ -435,7 +435,7 @@ namespace Windows.UI.Xaml
 			if (actualInstanceAlias != null)
 			{
 				ApplyPrecedenceOverride(ref precedence);
-				_isSettingAProperty = true;
+				_propertySettingDepth++;
 
 #if !HAS_EXPENSIVE_TRYFINALLY // Try/finally incurs a very large performance hit in mono-wasm - https://github.com/mono/mono/issues/13653
 				try
@@ -486,7 +486,7 @@ namespace Windows.UI.Xaml
 				finally
 #endif
 				{
-					_isSettingAProperty = false;
+					_propertySettingDepth--;
 				}
 			}
 			else
@@ -599,24 +599,26 @@ namespace Windows.UI.Xaml
 		[MethodImpl(MethodImplOptions.AggressiveInlining)]
 		private void ApplyPrecedenceOverride(ref DependencyPropertyValuePrecedences precedence)
 		{
-			if (_isSettingAProperty)
+			if(_precedenceOverride == DependencyPropertyValuePrecedences.ExplicitStyle)
+			{
+				// **WHEN OVERRIDE IS SET ON EXPLICIT STYLE**
+				// It means we need to propagate the "ExplicitStyle" precedence for the styling system to work well:
+				// we don't want style's setters to overwrite local values.
+			}
+			else if (!_precedenceOverride.HasValue || _propertySettingDepth > 0)
 			{
 				// We only want to override the precedence of properties set directly from a style. Nested sets (within property changed callbacks, etc)
-				// should be applied with the normal precedence.
+				// should be applied with the normal (usually "local") precedence.
+
 				return;
 			}
 
-			if (_precedenceOverride != null)
+			if (this.Log().IsEnabled(Microsoft.Extensions.Logging.LogLevel.Debug))
 			{
-				if (this.Log().IsEnabled(Microsoft.Extensions.Logging.LogLevel.Debug))
-				{
-					this.Log().Debug(
-						$"Overriding {precedence} precedence with {_precedenceOverride}."
-					);
-				}
-
-				precedence = _precedenceOverride.Value;
+				this.Log().Debug($"Overriding {precedence} precedence with {_precedenceOverride}.");
 			}
+
+			precedence = _precedenceOverride.Value;
 		}
 
 		/// <summary>
--- a/src/Uno.UI/UI/Xaml/DependencyObjectStore.cs
+++ b/src/Uno.UI/UI/Xaml/DependencyObjectStore.cs
@@ -92,7 +92,7 @@ namespace Windows.UI.Xaml
 		private readonly SerialDisposable _inheritedProperties = new SerialDisposable();
 		private ManagedWeakReference? _parentRef;
 		private readonly Dictionary<DependencyProperty, ManagedWeakReference> _inheritedForwardedProperties = new Dictionary<DependencyProperty, ManagedWeakReference>(DependencyPropertyComparer.Default);
-		private Stack<DependencyPropertyValuePrecedences?>? _overridenPrecedences;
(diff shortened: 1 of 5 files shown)


jeromelaban (approved):

src/Uno.UI/UI/Xaml/DependencyObjectStore.cs
```diff
@@ -599,24 +599,26 @@ private void WritePropertyEventTrace(int eventId, DependencyProperty property, D
 		[MethodImpl(MethodImplOptions.AggressiveInlining)]
 		private void ApplyPrecedenceOverride(ref DependencyPropertyValuePrecedences precedence)
 		{
-			if (_isSettingAProperty)
+			if(_precedenceOverride == DependencyPropertyValuePrecedences.ExplicitStyle)
```
davidjohnoliver, line 602: I don't think that there should be a special case for `ExplicitStyle`. Properties set from within property-changed callbacks should still be applied at local precedence, whatever type of style is triggering the callback.

davidjohnoliver: You should probably add a variation of the `When_Set_Within_Style_Application()` test where the callback is triggered by an explicit style.

carldebilly: Nope... that's the problem: `<Setter>`s MUST apply their values using the `ExplicitStyle` precedence. That's the origin of this bug: the first setter were applying its value using the `Local` precedence (another bug were preventing other setters from applying the overriden precedence).

Explanation: in the previous version, while manually setting the `Style` property (which uses `Local` precedence), it was setting both `_precedenceOverride` to `ExplicitStyle` and the `_isSettingAProperty` to true, preventing the setters from applying there values under correct explicit precedence.

But I think there's a remaining bug: only the "first level" of setters should using the overriden precedence. I'll need to add a test fot it.

jeromelaban: I originally thought of animations, but yes it makes sense.

carldebilly: TL;DR: you're right, except when this property is the `Style` itself.  In that case, the override for `ExpliciteStyle` precedence should get prioritize over the `Local` precedence.

davidjohnoliver: > there's a remaining bug: only the "first level" of setters should using the overriden precedence.

Well yeah, that's what I'm getting at.

I thought after we talked about this, that it would be easiest if `FrameworkElement.OnStyleChanged(Style oldStyle, Style newStyle, DependencyPropertyValuePrecedences precedence)` were to reset `DependencyObjectStore._isSettingAProperty` to false, which should handle the case of setting the `Style` property, but I guess you identified problems with that approach?

davidjohnoliver: Please add the following test:

```csharp
		[TestMethod]
		public void When_Set_Style_Property_PrecedenceInContext_No_Local()
		{
			var style = new Style();
			style.Setters.Add(new Setter(Border.BorderThicknessProperty, 10d));

			var sut = new Border { /*Don't set local BorderThickness*/ };

			using var registration1 = sut.RegisterDisposablePropertyChangedCallback(
				Border.BorderThicknessProperty,
				(dependencyObject, args) =>
				{
					sut.Tag = args.NewPrecedence;
				});

			using var registration2 = sut.RegisterDisposablePropertyChangedCallback(
				FrameworkElement.TagProperty,
				(dependencyObject, args) =>
				{
					args.NewPrecedence.Should().Be(DependencyPropertyValuePrecedences.Local);
					args.NewValue.Should().Be(DependencyPropertyValuePrecedences.ExplicitStyle);
				});

			using var _ = new AssertionScope();

			sut.Tag.Should().BeNull("After applying style");

			sut.Style = style;
		}
```

carldebilly: @davidjohnoliver I just pushed a new version which should solve the problem you identified.  You test is also integrated in it.

src/Uno.UI/UI/Xaml/DependencyObjectStore.cs
```diff
@@ -90,20 +89,20 @@ public static class TraceProvider
 		private ManagedWeakReference? _thisWeakRef;
 
 		private readonly Type _originalObjectType;
-		private SerialDisposable _inheritedProperties = new SerialDisposable();
+		private readonly SerialDisposable _inheritedProperties = new SerialDisposable();
 		private ManagedWeakReference? _parentRef;
-		private Dictionary<DependencyProperty, ManagedWeakReference> _inheritedForwardedProperties = new Dictionary<DependencyProperty, ManagedWeakReference>(DependencyPropertyComparer.Default);
-		private DependencyPropertyValuePrecedences? _precedenceOverride;
+		private readonly Dictionary<DependencyProperty, ManagedWeakReference> _inheritedForwardedProperties = new Dictionary<DependencyProperty, ManagedWeakReference>(DependencyPropertyComparer.Default);
+		private Stack<DependencyPropertyValuePrecedences?>? _overridenPrecedences;
```
davidjohnoliver, line 95: nit-pick: should be `_overriddenPrecedences`

src/Uno.UI/UI/Xaml/DependencyObjectStore.cs
```diff
@@ -592,31 +589,26 @@ private void WritePropertyEventTrace(int eventId, DependencyProperty property, D
 		/// Applies the precedence override that has been set through <seealso cref="OverrideLocalPrecedence(object, DependencyPropertyValuePrecedences)" />
 		/// Used to ambiently change the precedence specified when using standard DependencyProperty accessors, particularly when applying styles.
 		/// </summary>
-		/// <param name="property">The property being set</param>
-		/// <param name="value"></param>
 		/// <param name="precedence"></param>
 		/// <returns></returns>
 		[MethodImpl(MethodImplOptions.AggressiveInlining)]
```
davidjohnoliver, line 594: not sure if this is still desired

src/Uno.UI/UI/Xaml/DependencyObjectStore.cs
```diff
@@ -592,31 +589,26 @@ private void WritePropertyEventTrace(int eventId, DependencyProperty property, D
...
-					this.Log().Debug(
-						$"Overriding {precedence} precedence with {_precedenceOverride}."
-					);
-				}
+				precedence = current;
 
-				precedence = _precedenceOverride.Value;
+				// The ambient precedence is also set to "Local" for affected properties
+				// (that means if properties are set in any callback, this override won't apply)
+				return OverrideLocalPrecedence(null);
```
davidjohnoliver, line 608: neat solution! :)

davidjohnoliver (approved):

PR #4235, opened 2020-10-08
Update debugging-uno-ui.md - add info about long paths

weitzhandler:
This PR adds a recommendation to use a short path when cloning Uno, to avoid build errors.


jeromelaban (approved):

jeromelaban:
Thanks for your contribution!

PR #4239, opened 2020-10-08
Encode Android images with path

### LINKED ISSUE #3754: Android issues with duplicate image file names
## Current behavior

Currently having two images with the same name in different subfolders on Android causes issues as incorrect image may be referenced. For example:

```
/Assets/Code/TitleBar.png
/Assets/TitleBar.png
```

When `Image.Source` is set to `/Assets/TitleBar.png`, the `/Assets/Code/TitleBar.png` is displayed instead.

## Expected behavior

The correct image should be displayed.

See above.

## Possible solutions

Encode the path to the image by including the path to the image - for example "/Assets/Code/TitleBar.png" would be copied as a "Assets_Code_TitleBar.png" drawable to Android package. Then reference the images accordingly when loading them.

## Workaround

Use unique file names :-) .

## Environment

Nuget Package:

Package Version(s):

Affected platform(s):

- [ ] iOS
- [x] Android
- [ ] WebAssembly
- [ ] macOS
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

kazo0:
closes #3754 

## PR Type
- Bugfix

Currently having two images with the same name in different subfolders on Android causes issues as incorrect image may be referenced. For example:
 
 ```
 /Assets/Code/TitleBar.png
 /Assets/TitleBar.png
 ```
 
 When `Image.Source` is set to `/Assets/TitleBar.png`, the `/Assets/Code/TitleBar.png` is displayed instead.

## What is the new behavior?

 ```
 /Assets/Code/TitleBar.png --> assets_code_titlebar.png
 /Assets/TitleBar.png --> assets_titlebar.png
 ```
## PR Checklist


kazo0:
Would a UI Test be helpful here? Two images, one referencing `Assets/image.png` and one referencing `Assets/Subfolder/image.png`? Just to make sure the right images are displayed?

src/Uno.UWP/Helpers/DrawableHelper.Android.cs
```diff
@@ -33,8 +33,8 @@ public static Type Drawables
 		/// <param name="imageName">Name of the image</param>
 		/// <returns>Resource's id</returns>
 		public static int? FindResourceId(string imageName)
```
jeromelaban, line 35: I've been looking at compatibility, and this method is unfortunately used in ways that require the name to be just the file name.

We should do two things, adjust this method so that it searches through all resources using an "EndsWith" on keys or maybe lazily create a second dictionary that only uses the last part of a `Split('_')` for performance) using just the file name without the path, and make a second method, internal this time, which takes a full path.

That method was public a long time ago (you probably remember it ;)) and obviously should not have been...

kazo0: How do we want to deal with collisions when we have two images with the same name but they were just in different folders in the case of the new dictionary? Just take the first one?

kazo0: There's also the issue of when an image already has a underscore in its name, this new drawables lookup dictionary will have the wrong key

kazo0: Unless we want to encode the folder hierarchy in some wacky way that normal files wouldn't be named? But I know there's a lot of restrictions around resource names for Android

jeromelaban: That's fair... For duplicates and duplicate paths, it was not working before in all cases and we can detail these as a restrictions in the documentation.

In all cases, this method should not be used and be marked as non-browsable.

src/Uno.UWP/Helpers/DrawableHelper.Android.cs
```diff
@@ -78,13 +75,43 @@ public static Drawable FromUri(Uri uri)
...
+				}
+
+				return null;
+			}
+
+			return id;
+		}
+
+		private static int LookupDrawableFromFileName(string fileName) =>
+			_drawablesLookup
```
kazo0, line 111: Not my favourite solution but we can't really build a second dictionary here since the filenames themselves can include underscores

jeromelaban: Indeed. And this may become a performance issue... I wonder if we should instead a source generator to build the dictionaries. After all, enumerating the entries using reflection is not a good thing to do. This way, we could provide both lookups in an efficient way.

The generator would create a class that would provide two lookup methods with a switch/case based on the same Resource class used here: 
https://github.com/unoplatform/uno/blob/0c9de5d418ab197928bda5a69c61670298a18652/src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.cs#L475-L477

kazo0: Not sure I understand fully, would we be creating a dictionary using the source generators or would the source generators be creating two switch case methods that would be "hardcoded' with all the drawable names?

jeromelaban: This would be pretty much the same, as switch/case statements big enough get converted to dictionaries by the compiler. Are you familiar with source generation ? You can find a simple generator here: https://github.com/unoplatform/uno/blob/master/src/SamplesApp/SamplesApp.UITests.Generator/SnapShotTestGenerator.cs

kazo0: Source generation is pretty new to me, I'll give it some research and come back to this

kazo0: Ok, bear with me here. So what I should do is create an implementation of a `SourceGenerator`, something like `DrawableLookupGenerator`. Within the `Execute` method, I am given a `SourceGeneratorContext` with which I can grab the `Compilation` of the app. I can then query that `Compilation` to find the global `Resource.Drawable` class and query for its public properties. Now that I have the public properties, I can build up a massive switch statement method that will match up a provided file name or file path to the proper property name? I'd create two lookup methods, one that accepts just a straight filename that would match to potentially more than one property name of the `Resource.Drawable` class, we could search based on if the property names end with the filename. And one that accepts a path which we can encode and then see which `Resource.Drawable` property name matches that encoded string? The part I'm struggling to wrap my head around is what exactly am I **generating** in these new lookup methods. A big switch of every single public static property in the `Resource.Drawable` class? Basically this new class should do exactly what the current `DrawableHelper` is doing but it just won't be using Reflection to build the dictionary/switch?

kazo0: Also, where is this being generated? This wouldn't be part of the Android head right? It would have to generate a class that lives somewhere in the Uno assembly? Unless we generate this class in the Android head and add a reference to it to the DrawableHelper.Android.cs partial class, in the same sort of way we do:
https://github.com/unoplatform/uno/blob/0c9de5d418ab197928bda5a69c61670298a18652/src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.cs#L475-L477

We could do something like:
```
writer.AppendLineInvariant($"#if __ANDROID__");
writer.AppendLineInvariant($"global::Uno.Helpers.DrawableHelper.DrawableLookup = new global::{_defaultNamespace}.DrawableLookup());";
writer.AppendLineInvariant($"#endif");
```

Where `.DrawableHelper.DrawableLookup` would be a `IDrawableLookup`, and that interface would live inside Uno and allow us to call the methods without having an implementation inside Uno itself.

Does any of this make sense?

jeromelaban:
Let's merge this for 3.5

MartinZikmund:
@kazo0 Hey, any chance for this to be in 3.4? Seems there are some test fails, but would be very useful 👍

PR #4270, opened 2020-10-13
feat(dragdrop): Add support of Drag and Drop

dr1rrb:
GitHub Issue: https://github.com/unoplatform/uno/issues/1480

## Feature
Add support `UIElement.Drag<Starting|Entered|Over|Exited>` and `UIElement.Drop[Completed]` events for inter-app and intra-app drag and drop.

Drag and drop is not supported at all.

## What is the new behavior?
🙃

## PR Checklist
- [ ] **TODO** [Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)


src/Uno.UWP/ApplicationModel/DataTransfer/DataPackage.cs
```diff
@@ -1,73 +1,97 @@
...
-			if (value == null)
+			SetData(formatId, new DataProviderHandler(SetDataCore));
+
+			async void SetDataCore(DataProviderRequest request)
 			{
-				throw new ArgumentNullException("Cannot set DataPackage.Uri to null");
+				var deferral = request.GetDeferral();
+				try
+				{
+					await delayRenderer(request.CancellationToken);
```
robloo, line 46: This will not work. The delayRenderer data is going nowhere. To fix this in my branch I added:

```csharp
var data = await delayRenderer(request.CancellationToken);
request.SetData(data);
```

robloo:
It's great to see this PR. Lots of good stuff in here, awesome job!

src/Uno.UI.Runtime.Skia.Wpf/WPF/WpfDragDropExtension.cs
```diff
@@ -0,0 +1,279 @@
...
+					encoder.Save(copy);
+					copy.Position = 0;
+
+					return copy.AsRandomAccessStream().TrySetContentType("image/bmp");
+				}));
+			}
+
+			if (src.GetData(DataFormats.FileDrop) is string[] files)
+			{
+				dst.SetStorageItems(files.Select(StorageFile.GetFileFromPath));
```
robloo, line 137: Doing this when the DataPackage is created means that the file will be read as soon as the cursor enters the Uno app. This **may** not be performance friendly (although the file isn't copied here) and it would be better to use `DataPackage.SetDataProvider()`. For the macOS platform I use the following code:

```csharp
dataPackage.SetDataProvider(
    StandardDataFormats.StorageItems,
    async cancellationToken =>
    {
        // Convert rom a temp Url (see above example) into an absolute file path
        var fileUrl = new NSUrl(tempFileUrl);
        var file = await StorageFile.GetFileFromPathAsync(fileUrl.FilePathUrl.AbsoluteString);

        var storageItems = new List<IStorageItem>();
        storageItems.Add(file);

        return storageItems.AsReadOnly();
        });
```

src/Uno.UI.Runtime.Skia.Wpf/WPF/WpfDragDropExtension.cs
```diff
@@ -0,0 +1,279 @@
...
+			}
+
+			if (src.GetData(DataFormats.Rtf) is string rtf)
+			{
+				dst.SetRtf(rtf);
+			}
+
+			if (src.GetData(DataFormats.Bitmap) is BitmapSource bitmap)
+			{
+				dst.SetBitmap(new RandomAccessStreamReference(async ct =>
```
robloo, line 123: For large images this copy may cause an unwanted performance hit as the pointer first drags over an Uno app. Instead it might be best to use a DataProvider as commented below for StorageItems.

src/Uno.UWP/ApplicationModel/DataTransfer/DataPackage.cs
dr1rrb: Whoups ... good point! I applied your fix.

dr1rrb:
> 
> 
> It's great to see this PR. Lots of good stuff in here, awesome job!

@robloo Thanks for your contribution to this PR 😃

src/Uno.UI.Runtime.Skia.Wpf/WPF/WpfDragDropExtension.cs
dr1rrb: With the [latest version of the `StorageFile`](https://github.com/unoplatform/uno/blob/master/src/Uno.UWP/Storage/StorageFile.cs#L24), this will actually only create some instances of `StorageFile` (and it does not create a `FileInfo` which would indeed have touched the file system). It's even using an internal sync overload which makes it the most lightweight possible I think.

Did you notice any perf issue with this one or it's just theoretical?

src/Uno.UI.Runtime.Skia.Wpf/WPF/WpfDragDropExtension.cs
dr1rrb: It's very good point, however here I only create a `RandomAccesStream**Reference**`, which is designed to load its content only when [the reference is open for read](https://github.com/unoplatform/uno/blob/master/src/Uno.UWP/Storage/Streams/RandomAccessStreamReference.cs#L44). So it's also deferred to the effective usage.

I will however add some unit test to make sure that this behavior is not altered.

robloo:
Not sure if you will see it here faster: The commits in PR https://github.com/unoplatform/uno/pull/4216 need to be added here to fix the macOS and iOS build issues.

doc/articles/features/pointers-keyboard-and-other-user-inputs.md
```diff
@@ -0,0 +1,183 @@
+# Uno support for user inputs
+
+## Supported user inputs
+
+User inputs are usually propagated to the using `RoutedEvent`, see. Uno's [routed events documentation](routed-events.md) to better understand their implementation on Uno.
+
+| Routed Event                  | Android | iOS     | Wasm    | MacOS | Skia WPF | Skia GtK | Tizen |    |
```
robloo, line 7: I think there is a mismatch between the table header and column count. The table doesn't render correctly to HTML.

src/Uno.UI.Runtime.Skia.Wpf/WPF/WpfDragDropExtension.cs
robloo: This one was just theoretical :) I did it differently on macOS just to be on the extra-safe side. Perhaps that wasn't necessary but I didn't know the internals of how StorageFile works on all platforms and could see it accessing the file system so wanted to be careful.

src/Uno.UI.Runtime.Skia.Wpf/WPF/WpfDragDropExtension.cs
robloo: Yes, I realized the RandomAccessStreamReference won't load the stream until it's needed. However, to create the stream reference you **are** making a copy which is what I was worried about:

```csharp
var copy = new MemoryStream();
var encoder = new BmpBitmapEncoder();
encoder.Frames.Add(BitmapFrame.Create(bitmap));
encoder.Save(copy);
copy.Position = 0;
```

robloo:
@dr1rrb I realize the DragDropManager may need another method similar to ProcessAborted()/ProcessDropped(). 

When an external drop is completed outside of Uno we should notify the source UIElement inside Uno to raise DropCompleted. This means the manager really should have a ProcessCompleted(). This could also go inside the other Complete() handling code.

Otherwise, maybe the missing functionality should be in ProcessDropped()? Or I'm completely missing something? Either way, DropCompleted is not getting raised inside Uno when an external drop is completed (on macOS at least). As soon as the pointer comes back into Uno the drag is still active even after dropping content externally.

dr1rrb:
> 
> 
> @dr1rrb I realize the DragDropManager may need another method similar to ProcessAborted()/ProcessDropped().
> 
> When an external drop is completed outside of Uno we should notify the source UIElement inside Uno to raise DropCompleted. This means the manager really should have a ProcessCompleted(). This could also go inside the other Complete() handling code.
> 
> Otherwise, maybe the missing functionality should be in ProcessDropped()? Or I'm completely missing something? Either way, DropCompleted is not getting raised inside Uno when an external drop is completed (on macOS at least). As soon as the pointer comes back into Uno the drag is still active even after dropping content externally.

@robloo On Skia I track them using the lost of the pointer capture ... but you made me realized that I probably don't propagate the actual drop result in that case. So you're right and I would agree that we should add those methods. I'll see if I'm able to fix this afternoon. (If not, I don't consider this as a blocker to merge this PR as anyway the external drag and drop is supported on skia, and this PR is already way to massive to wait more longer)

robloo:
@dr1rrb OK, I totally understand if this is all getting too big and we just need to merge as-is and fix more later. I'm good with whatever you decide :)

src/Uno.UI.Runtime.Skia.Wpf/WPF/WpfDragDropExtension.cs
```diff
@@ -0,0 +1,279 @@
...
+
+		public void StartNativeDrag(CoreDragInfo info)
+			=> WpfHost.Current.Dispatcher.InvokeAsync(async () =>
+			{
+				try
+				{
+					var data = await ToDataObject(info.Data, CancellationToken.None);
+					var effects = ToDropEffects(info.AllowedOperations);
+
+					DragDrop.DoDragDrop(WpfHost.Current, data, effects);
```
robloo, line 77: How is Uno notified when the host completes a drop? How is DropCompleted raised? Maybe that comes through the window events (OnHostDrop) again?

robloo: Edit: Agree to address this later as needed. (assuming I'm understanding correctly anyway).

carldebilly (approved):

doc/articles/features/pointers-keyboard-and-other-user-inputs.md
```diff
@@ -0,0 +1,183 @@
...
+| `DragOver`                    | Yes     | Yes     | Yes     | Yes   | Yes      | Yes      | Yes   | [Documentation](https://docs.microsoft.com/uwp/api/windows.ui.xaml.uielement.dragover) |
+| `DragLeave`                   | Yes     | Yes     | Yes     | Yes   | Yes      | Yes      | Yes   | [Documentation](https://docs.microsoft.com/uwp/api/windows.ui.xaml.uielement.dragleave) |
+| `Drop`                        | Yes     | Yes     | Yes     | Yes   | Yes      | Yes      | Yes   | [Documentation](https://docs.microsoft.com/uwp/api/windows.ui.xaml.uielement.drop) |
+| `DropCompleted`               | Yes     | Yes     | Yes     | Yes   | Yes      | Yes      | Yes   | [Documentation](https://docs.microsoft.com/uwp/api/windows.ui.xaml.uielement.dropcompleted) |
+
+
+Notes:
+
+1. **Focus** events:
+   * **iOS**: The concept of _focus_ is emulated because not supported by the platform, so this event is
```
davidjohnoliver, line 47:
```suggestion
   * **iOS**: The concept of _focus_ is emulated because it's not supported by the platform, so this event is
```

doc/articles/features/pointers-keyboard-and-other-user-inputs.md
```diff
@@ -0,0 +1,183 @@
...
+     those are translated as `KeyUp` with `KeyCode.Enter`). Some soft keyboard **MAY** generate those events, but your code shouldn't rely
+     on that. This is a limitation [in the Android platform](https://developer.android.com/training/keyboard-input/commands) (see note on this link content).
+     > Because of those limitations, _Key Events_ are not being implemented as _routed events_ on Android, so `AddHandler` & `RemoveHandler`
+     > won't work for keyboard events. **They won't bubble in managed code**.
+   * **iOS**: `KeyDown` & `KeyUp` routed events are generated from only a `TextBox`. Only character-related keyboard events are generated.
+     They are implemented as _Routed Events_ and they are **always bubbling in managed code**.
+
+## Pointer Events
+
+Those events are the base for all others pointing device related events (i.e. Manipulation, Gesture and drag and dop events).
```
davidjohnoliver, line 61:
```suggestion
These events are the base for all other pointing device related events (i.e. Manipulation, Gesture and drag and dop events).
```

doc/articles/features/pointers-keyboard-and-other-user-inputs.md
```diff
@@ -0,0 +1,183 @@
...
+They are directly linked to the native events of each platform:
+* `Touches[Began|Moved|Ended|Cancelled]` on iOS
+* `dispatchTouchEvent` and `dispatchGenericMotionEvent` on Android
+* `pointer[enter|leave|down|up|move|cancel]` on WebAssembly
+
+On skia however, they are fully managed events.
```
davidjohnoliver, line 67:
```suggestion
On Skia however, they are fully managed events.
```

doc/articles/features/pointers-keyboard-and-other-user-inputs.md
```diff
@@ -0,0 +1,183 @@
...
+
+### Pointers events and the ScrollViewer
+
+Like on WinUI as soon as the system detects that the user wants to scroll, a control gets a `PointerCancelled` and that control won't receive
+any other pointer event until the user release the pointer. That behavior can be prevented by setting the `ManipulationMode`
```
davidjohnoliver, line 72:
```suggestion
any other pointer event until the user releases the pointer. That behavior can be prevented by setting the `ManipulationMode` 
```

doc/articles/features/pointers-keyboard-and-other-user-inputs.md
```diff
@@ -0,0 +1,183 @@
...
+		A.PointerPressed
+		```
+	* If you handle the `PointerEnter` on **B**, the parent control **A** won't get the `PointerEnter` (as expected) nor the  `PointerPressed`.
+* On Android with a mouse or a pen, the `PointerEnter` and `PointerExit` are going to be raised without taking clipping in consideration.
+  This means that you will get the enter earlier and the exit later than on other platform.
+* On Android if you have an element with a `RenderTransform` which overlaps one of its sibling element, 
+  the element at the top will get the pointer events.
+* On WASM, iOS and Android, the `RoutedPointerEventArgs.FrameId` will be reset to 0 after 49 days of running time of the application.
+* Unlike on UWP, controls that are under a `Popup` won't receive the unhandled pointer events.
+* On non skia-based platforms, unlike UWP, it's impossible to receive a `PointerReleased` without getting a `PointerPressed` before.
```
davidjohnoliver, line 104:
```suggestion
* On non Skia-based platforms, unlike UWP, it's impossible to receive a `PointerReleased` without getting a `PointerPressed` before.
```

doc/articles/features/pointers-keyboard-and-other-user-inputs.md
```diff
@@ -0,0 +1,183 @@
...
+  (For instance if a child control handled the pressed event but not the released event).
+  > On WASM as `TextElement` are `UIElement`, it means that unlike UWP `TextBlock` won't raise the
```
davidjohnoliver, line 106:
```suggestion
  > On WASM as `TextElement` inherits from `UIElement`, it means that unlike UWP `TextBlock` won't raise the 
```

doc/articles/features/pointers-keyboard-and-other-user-inputs.md
```diff
@@ -0,0 +1,183 @@
...
+* On Android with a mouse or a pen, the `PointerEnter` and `PointerExit` are going to be raised without taking clipping in consideration.
+  This means that you will get the enter earlier and the exit later than on other platform.
+* On Android if you have an element with a `RenderTransform` which overlaps one of its sibling element, 
+  the element at the top will get the pointer events.
+* On WASM, iOS and Android, the `RoutedPointerEventArgs.FrameId` will be reset to 0 after 49 days of running time of the application.
+* Unlike on UWP, controls that are under a `Popup` won't receive the unhandled pointer events.
+* On non skia-based platforms, unlike UWP, it's impossible to receive a `PointerReleased` without getting a `PointerPressed` before.
+  (For instance if a child control handled the pressed event but not the released event).
+  > On WASM as `TextElement` are `UIElement`, it means that unlike UWP `TextBlock` won't raise the 
+  > a `PointerReleased` when clicking on an `Hyperlink`.
```
davidjohnoliver, line 107:
```suggestion
  > `PointerReleased` event when clicking on a `Hyperlink`.
```

doc/articles/features/pointers-keyboard-and-other-user-inputs.md
```diff
@@ -0,0 +1,183 @@
...
+<true/>
+```
+
+Without this key the current version of iPadOS reports mouse interaction as normal touch.
+
+## Manipulation Events
+
+They are generated from the PointerXXX events (using the `Windows.UI.Input.GestureRecognizer`) and are bubbling in managed only.
+
+Currently there is no intertia support, so the `IsInertial` will always be `false` and the `UIElement.ManipulationInertiaStarting` event
```
davidjohnoliver, line 143:
```suggestion
Currently there is no inertia support, so the `IsInertial` will always be `false` and the `UIElement.ManipulationInertiaStarting` event 
```

doc/articles/features/pointers-keyboard-and-other-user-inputs.md
```diff
@@ -0,0 +1,183 @@
...
+```
+
+Without this key the current version of iPadOS reports mouse interaction as normal touch.
+
+## Manipulation Events
+
+They are generated from the PointerXXX events (using the `Windows.UI.Input.GestureRecognizer`) and are bubbling in managed only.
+
+Currently there is no intertia support, so the `IsInertial` will always be `false` and the `UIElement.ManipulationInertiaStarting` event 
+will never be fired. The `Velocities` properties of event args are not implemented neither.
```
davidjohnoliver, line 144:
```suggestion
will never be fired. The `Velocities` properties of event args are not implemented either.
```

doc/articles/features/pointers-keyboard-and-other-user-inputs.md
```diff
@@ -0,0 +1,183 @@
...
+Be aware that on some browser (Firefox), user can still request to get the "Context menu" on right click.
+
+## Drag and drop
+
+Those events are also 100% managed events, built from the PointerXXX events (using the `Windows.UI.Input.GestureRecognizer`)
+
+### Intra-app drag and drop support
+
+A _drag and drop_ operation can be used to move content within an app, but it can also be used to **copy** / **move** / **link** between apps.
+While inter-app _drag and drop_ is supported on all platforms without limiations, intra-app _drag and drop_ requires platfrom specific support.
```
davidjohnoliver, line 163:
```suggestion
While inter-app _drag and drop_ is supported on all platforms without limitations, intra-app _drag and drop_ requires platform specific support.
```

I corrected the typos but... is it true? You write "supported on all platforms without limitations" but there's a lot of 'No' below :p

doc/articles/features/pointers-keyboard-and-other-user-inputs.md
```diff
@@ -0,0 +1,183 @@
+# Uno support for user inputs
```
davidjohnoliver, line 1: Great doc! Should be added to TOC https://github.com/unoplatform/uno/blob/master/doc/articles/toc.yml

src/Uno.UI/UI/Xaml/DragView.cs
```diff
@@ -0,0 +1,192 @@
...
+			Visibility = Visibility.Collapsed;
+		}
+
+		private static Visibility ToVisibility(bool isVisible)
+			=> isVisible ? Visibility.Visible : Visibility.Collapsed;
+
+		private static string ToGlyph(DataPackageOperation result)
+		{
+			// If multiple flags set (which should not!), the UWP precedence is Link > Copy > Move
+			// TODO: Real glyph
```
davidjohnoliver, line 151: This TODO and the one below warrant issues to track them

davidjohnoliver (approved):

doc/articles/features/pointers-keyboard-and-other-user-inputs.md
dr1rrb: Yeah it's true: the "inter-app" does not suffer real limitation, just a minor bug which won't have any functional impact.

src/Uno.UI/UI/Xaml/DragView.cs
dr1rrb: Good point, I'll try to fix it before the merge, otherwise I'll create a new PR for this one

doc/articles/features/pointers-keyboard-and-other-user-inputs.md
davidjohnoliver: Ok then are you mixing up 'inter-' and 'intra-'?

dr1rrb: Whoups rigth! ... Fixed

PR #4308, opened 2020-10-19
feat(dragdrop): Drag-and-drop integration with macOS

robloo:
GitHub Issue fixes https://github.com/unoplatform/uno/issues/4311

Relates to Issue #1480 and PR #4270 

## PR Type

Adds support 

- Feature

Drag and drop is not integrated with macOS and other apps.

## What is the new behavior?

Drag and drop is integrated with macOS supporting inter-app drag and drop:
 - Drop of external content on an Uno App is fully supported (images, files, text, etc.)
 - Drag of Uno App content to external macOS apps is supported but has some limitations/issues.

## PR Checklist

- [ ] ~[Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)~
- [ ] ~Validated PR `Screenshots Compare Test Run` results.~

## Other information

ToDo:
 - [x] Add back `internal static Task SetToNativeClipboardAsync(DataPackage content)` to DataPackage.macOS.cs to fix tests (temporarily removed to allow cherry-picking commits without conflicts).
 - [x] Update docs indicating macOS support for inter-app drag/drop
 - [ ] Drag and drop of images is fully supported but images are not redrawn when a stream is updated async. This was fixed for development with a temporary work-around not included in this PR. An additional PR is planned to fully fix this issue on supported platforms.

Other open questions are commented in code.

### DIFF
--- a/src/Uno.UI/Controls/Window.macOS.cs
+++ b/src/Uno.UI/Controls/Window.macOS.cs
@@ -164,11 +164,24 @@ namespace Uno.UI.Controls
 		[Export("draggingEnded:")] // Do not remove
 		public virtual void DraggingEnded(NSDraggingInfo draggingInfo)
 		{
-			// Not used in this context.
-			// These methods within NSWindow are only used for external drop into Uno apps.
-			// If Uno doesn't accept and handle the drop it doesn't need to be aware of when it ends.
+			try
+			{
+				DraggingEndedAction.Invoke(draggingInfo);
+			}
+			catch { }
+
 			return;
 		}
+
+		/// <summary>
+		/// Code to execute when the <see cref="DraggingEnded(NSDraggingInfo)"/> method is invoked.
+		/// This can be used in a similar way to <see cref="UIElement.DropCompleted"/>.
+		/// </summary>
+		public Action<NSDraggingInfo> DraggingEndedAction { get; set; } =
+			(NSDraggingInfo draggingInfo) =>
+			{
+				return;
+			};
 
 		/// <summary>
 		/// Method invoked when a drag operation exited the <see cref="NSWindow"/>.
--- a/src/Uno.UI/Controls/Window.macOS.cs
+++ b/src/Uno.UI/Controls/Window.macOS.cs
@@ -100,7 +100,7 @@ namespace Uno.UI.Controls
 		/// </remarks>
 		/// <param name="info">Information about the dragging session from the sender.</param>
 		/// <returns>The accepted drag operation(s).</returns>
-		[Export("draggingEntered:")]
+		[Export("draggingEntered:")] // Do not remove
 		public virtual NSDragOperation DraggingEntered(NSDraggingInfo draggingInfo)
 		{
 			try
@@ -131,7 +131,7 @@ namespace Uno.UI.Controls
 		/// See remarks in <see cref="DraggingEntered(NSDraggingInfo)"/>
 		/// </remarks>
 		/// <param name="draggingInfo">Information about the dragging session from the sender.</param>
-		[Export("draggingUpdated:")]
+		[Export("draggingUpdated:")] // Do not remove
 		public virtual NSDragOperation DraggingUpdated(NSDraggingInfo draggingInfo)
 		{
 			try
@@ -161,7 +161,7 @@ namespace Uno.UI.Controls
 		/// See remarks in <see cref="DraggingEntered(NSDraggingInfo)"/>
 		/// </remarks>
 		/// <param name="draggingInfo">Information about the dragging session from the sender.</param>
-		[Export("draggingEnded:")]
+		[Export("draggingEnded:")] // Do not remove
 		public virtual void DraggingEnded(NSDraggingInfo draggingInfo)
 		{
 			// Not used in this context.
@@ -177,7 +177,7 @@ namespace Uno.UI.Controls
 		/// See remarks in <see cref="DraggingEntered(NSDraggingInfo)"/>
 		/// </remarks>
 		/// <param name="draggingInfo">Information about the dragging session from the sender.</param>
-		[Export("draggingExited:")]
+		[Export("draggingExited:")] // Do not remove
 		public virtual void DraggingExited(NSDraggingInfo draggingInfo)
 		{
 			return;
@@ -199,7 +199,7 @@ namespace Uno.UI.Controls
 		/// </summary>
 		/// <param name="draggingInfo">Information about the dragging session from the sender.</param>
 		/// <returns>True if the destination accepts the drag operation; otherwise, false. </returns>
-		[Export("prepareForDragOperation:")]
+		[Export("prepareForDragOperation:")] // Do not remove
 		public virtual bool PrepareForDragOperation(AppKit.NSDraggingInfo draggingInfo)
 		{
 			// Always return true as UWP doesn't really have an equivalent step.
@@ -212,7 +212,7 @@ namespace Uno.UI.Controls
 		/// </summary>
 		/// <param name="draggingInfo">Information about the dragging session from the sender.</param>
 		/// <returns>True if the destination accepts the data; otherwise, false.</returns>
-		[Export("performDragOperation:")]
+		[Export("performDragOperation:")] // Do not remove
 		public virtual bool PerformDragOperation(NSDraggingInfo draggingInfo)
 		{
 			try
--- a/src/Uno.UI/Controls/Window.macOS.cs
+++ b/src/Uno.UI/Controls/Window.macOS.cs
@@ -94,7 +94,7 @@ namespace Uno.UI.Controls
 		///      give no examples.
 		/// 
 		/// The Export "method_name" attribute is fundamentally important to make this work, removing it will
-		/// break functionality and the method will never be called by macOS. Again, this seemly is because
+		/// break functionality and the method will never be called by macOS. Again, this seemingly is because
 		/// Xamarin.macOS is not aware of it.
 		/// 
 		/// </remarks>
@@ -170,7 +170,7 @@ namespace Uno.UI.Controls
 			}
 			catch
 			{
-				// Simply return if an error occured, it is unrecoverable
+				// Simply return if an error occurred, it is unrecoverable
 			}
 
 			return;
@@ -196,6 +196,15 @@ namespace Uno.UI.Controls
 		[Export("draggingExited:")] // Do not remove
 		public virtual void DraggingExited(NSDraggingInfo draggingInfo)
 		{
+			try
+			{
+				DraggingExitedAction.Invoke(draggingInfo);
+			}
+			catch
+			{
+				// Simply return if an error occurred, it is unrecoverable
+			}
+
 			return;
 		}
 
--- a/src/Uno.UI/Controls/Window.macOS.cs
+++ b/src/Uno.UI/Controls/Window.macOS.cs
@@ -44,7 +44,196 @@ namespace Uno.UI.Controls
 			Delegate = new WindowDelegate(this);
 			_inputPane = InputPane.GetForCurrentView();
 			_inputPane.Window = this;
+
+			// Register the types that can be dragging into the window (effectively 'AllowDrop').
+			// This must be done before dragging enters the window or dragging methods won't be invoked.
+			// We register for all possible types and then confirm whether they
+			// are actually supported within the draggingEntered and draggingUpdated methods.
+			// This has a minor side effect in the standard UI (zoom effect) but is considered acceptable.
+			RegisterForDraggedTypes(new string[]
+				{
+					NSPasteboard.NSPasteboardTypeUrl,
+					NSPasteboard.NSPasteboardTypeTIFF,
+					NSPasteboard.NSPasteboardTypePNG,
+					NSPasteboard.NSPasteboardTypeHTML,
+					NSPasteboard.NSPasteboardTypeRTF,
+					NSPasteboard.NSPasteboardTypeRTFD,
+					NSPasteboard.NSPasteboardTypeFileUrl,
+					NSPasteboard.NSPasteboardTypeString
+
+					/* For future use
+					UTType.URL,
+					UTType.Image,
+					UTType.HTML,
+					UTType.RTF,
+					UTType.FileURL,
+					UTType.PlainText,
+					*/
+				});
 		}
+
+		/// <summary>
+		/// Method invoked when a drag operation enters the <see cref="NSWindow"/>.
+		/// </summary>
+		/// <remarks>
+		/// 
+		/// While it is never documented directly, DraggingEntered can be added to NSWindow just like NSView.
+		/// Key docs telling this story include:
+		/// 
+		///  (1) NSWindow documentation does not include most NSView drag/drop methods
+		///      https://developer.apple.com/documentation/appkit/nswindow
+		///  (2) Since NSWindow documentation doesn't reference them, they also aren't included in Xamarin
+		///      https://docs.microsoft.com/en-us/dotnet/api/appkit.nswindow?view=xamarin-mac-sdk-14
+		///  (3) However, the 'draggingEntered' method docs reference 'Window' at the very end
+		///      "Invoked when the mouse pointer enters the destination’s bounds rectangle (if it is a view object)
+		///       or its frame rectangle (if it is a window object). "
+		///      https://developer.apple.com/documentation/appkit/nsdraggingdestination/1416019-draggingentered
+		///  (4) In legacy macOS docs this is a little more explitly stated:
+		///      https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/DragandDrop/Concepts/dragdestination.html#//apple_ref/doc/uid/20000977-BAJBJFBG
+		///  (5) Other macOS drag/drop articles and blogs refer to dragging/dropping directly from Window - although
+		///      give no examples.
+		/// 
+		/// The Export "method_name" attribute is fundamentally important to make this work, removing it will
+		/// break functionality and the method will never be called by macOS. Again, this seemly is because
+		/// Xamarin.macOS is not aware of it.
+		/// 
+		/// </remarks>
+		/// <param name="info">Information about the dragging session from the sender.</param>
+		/// <returns>The accepted drag operation(s).</returns>
+		[Export("draggingEntered:")]
+		public virtual NSDragOperation DraggingEntered(NSDraggingInfo draggingInfo)
+		{
+			try
+			{
+				return DraggingEnteredAction.Invoke(draggingInfo);
+			}
+			catch
+			{
+				return NSDragOperation.None;
+			}
+		}
+
+		/// <summary>
+		/// Code to execute when the <see cref="DraggingEntered(NSDraggingInfo)"/> method is invoked.
+		/// This can be used in a similar way to <see cref="UIElement.DragEnter"/>.
+		/// </summary>
+		public Func<NSDraggingInfo, NSDragOperation> DraggingEnteredAction { get; set; } =
+			(NSDraggingInfo draggingInfo) =>
+			{
+				return NSDragOperation.None;
+			};
+
+		/// <summary>
+		/// Method invoked when a drag operation updates over the <see cref="NSWindow"/>.
+		/// This is periodically called when the pointer is moved.
+		/// </summary>
+		/// <remarks>
+		/// See remarks in <see cref="DraggingEntered(NSDraggingInfo)"/>
+		/// </remarks>
+		/// <param name="draggingInfo">Information about the dragging session from the sender.</param>
+		[Export("draggingUpdated:")]
+		public virtual NSDragOperation DraggingUpdated(NSDraggingInfo draggingInfo)
+		{
+			try
+			{
+				return DraggingUpdatedAction.Invoke(draggingInfo);
+			}
+			catch
+			{
+				return NSDragOperation.None;
+			}
+		}
+
+		/// <summary>
+		/// Code to execute when the <see cref="DraggingUpdated(NSDraggingInfo)"/> method is invoked.
+		/// This can be used in a similar way to <see cref="UIElement.DragOver"/>.
+		/// </summary>
+		public Func<NSDraggingInfo, NSDragOperation> DraggingUpdatedAction { get; set; } =
+			(NSDraggingInfo draggingInfo) =>
+			{
+				return NSDragOperation.None;
+			};
+
+		/// <summary>
+		/// Method invoked when a drag operation ends.
+		/// </summary>
+		/// <remarks>
+		/// See remarks in <see cref="DraggingEntered(NSDraggingInfo)"/>
+		/// </remarks>
+		/// <param name="draggingInfo">Information about the dragging session from the sender.</param>
+		[Export("draggingEnded:")]
+		public virtual void DraggingEnded(NSDraggingInfo draggingInfo)
+		{
+			// Not used in this context.
+			// These methods within NSWindow are only used for external drop into Uno apps.
+			// If Uno doesn't accept and handle the drop it doesn't need to be aware of when it ends.
+			return;
+		}
+
+		/// <summary>
+		/// Method invoked when a drag operation exited the <see cref="NSWindow"/>.
+		/// </summary>
+		/// <remarks>
+		/// See remarks in <see cref="DraggingEntered(NSDraggingInfo)"/>
+		/// </remarks>
+		/// <param name="draggingInfo">Information about the dragging session from the sender.</param>
+		[Export("draggingExited:")]
+		public virtual void DraggingExited(NSDraggingInfo draggingInfo)
+		{
+			return;
+		}
+
+		/// <summary>
+		/// Code to execute when the <see cref="DraggingExited(NSDraggingInfo)"/> method is invoked.
+		/// This can be used in a similar way to <see cref="UIElement.DragLeave"/>.
+		/// </summary>
+		public Action<NSDraggingInfo> DraggingExitedAction { get; set; } =
+			(NSDraggingInfo draggingInfo) =>
+			{
+				return;
+			};
+
+		/// <summary>
+		/// Method invoked when the pointer is released over the <see cref="NSWindow"/> during a drag operation
+		/// This is the last chance to reject a drop.
+		/// </summary>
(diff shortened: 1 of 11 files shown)


src/Uno.UWP/UI/Core/CoreWindow.macOS.cs
```diff
@@ -13,7 +13,8 @@ namespace Windows.UI.Core
 {
 	public partial class CoreWindow 
 	{
-        private readonly NSWindow _window;
+		// HACK: TODO: Find a better way to get this reference
+		public readonly NSWindow _window;
```
robloo, line 17: I'm not sure what to do about this exactly. It should at least be made 'internal'; however, I need a global way to access the reference to the NSWindow.

src/Uno.UI/UI/Xaml/DragDropExtension.macOS.cs
```diff
@@ -0,0 +1,178 @@
...
+		}
+
+		public void StartNativeDrag(CoreDragInfo info)
+			=> CoreDispatcher.Main.RunAsync(CoreDispatcherPriority.High, async () =>
+			{
+				try
+				{
+					NSDraggingSource source = new NSDraggingSource();
+					NSEvent sourceEvent = new NSEvent();
+
```
robloo, line 82: The native dragging session is started from the ContentView insead of the NSWindow. [NSWindow.Drag](https://developer.apple.com/documentation/appkit/nswindow/1419224-drag) does not seem to be supported in Xamarin -- this method does not exist in Xamarin.mac. Attempts at invoking the objective-c code directly are so far unsuccessful.

robloo:
@dr1rrb FYI

src/Uno.UI/UI/Xaml/DragDropManager.cs
```diff
@@ -23,10 +23,16 @@ internal sealed class DragDropManager : CoreDragDropManager.IDragDropManager
 		public DragDropManager(Window window)
 		{
 			_window = window;
+
+#if __MACOS__
```
robloo, line 27: I'm not sure how difficult it would be to get dependency injection working on macOS. Regardless, this should change when/if it is ever supported.

src/Uno.UI/Controls/Window.macOS.cs
```diff
@@ -44,8 +44,210 @@ public Window(CGRect contentRect, NSWindowStyle aStyle, NSBackingStore buffering
...
+					UTType.PlainText,
+					*/
+				});
+		}
+
+		/// <summary>
+		/// Method invoked when a drag operation enters the <see cref="NSWindow"/>.
+		/// </summary>
+		/// <remarks>
+		///
```
robloo, line 79: All of these 'remarks' comments could be moved out into a separate document explaining how macOS drag/drop integration works. It might be useful to have this document for all platforms adding to it as new platforms are supported. Let me know if a separate document is desirable.

src/Uno.UI/Controls/Window.macOS.cs
```diff
@@ -44,8 +44,210 @@ public Window(CGRect contentRect, NSWindowStyle aStyle, NSBackingStore buffering
 			Delegate = new WindowDelegate(this);
 			_inputPane = InputPane.GetForCurrentView();
 			_inputPane.Window = this;
+
+			// Register the types that can be dragging into the window (effectively 'AllowDrop').
+			// This must be done before dragging enters the window or dragging methods won't be invoked.
+			// We register for all possible types and then confirm whether they
+			// are actually supported within the draggingEntered and draggingUpdated methods.
+			// This has a minor side effect in the standard UI (zoom effect) but is considered acceptable.
```
dr1rrb, line 52: When you say it "has a minor effect in the standard UI", I was wondering if we should maintain a list of loaded elements that allow drop. If so, we could at least invoke this registration only if we have an element that allows drop (and even maybe unregister if element gets unloaded).

We would still not be able to pre-filter supported types like macOS seems to ask for, but we could at least not appear as a possible drop target if nothing is supported.

Does this would significantly improve the user experience?

src/Uno.UI/Controls/Window.macOS.cs
```diff
@@ -44,8 +44,210 @@ public Window(CGRect contentRect, NSWindowStyle aStyle, NSBackingStore buffering
...
+		///  (2) Since NSWindow documentation doesn't reference them, they also aren't included in Xamarin
+		///      https://docs.microsoft.com/en-us/dotnet/api/appkit.nswindow?view=xamarin-mac-sdk-14
+		///  (3) However, the 'draggingEntered' method docs reference 'Window' at the very end
+		///      "Invoked when the mouse pointer enters the destination’s bounds rectangle (if it is a view object)
+		///       or its frame rectangle (if it is a window object). "
+		///      https://developer.apple.com/documentation/appkit/nsdraggingdestination/1416019-draggingentered
+		///  (4) In legacy macOS docs this is a little more explitly stated:
+		///      https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/DragandDrop/Concepts/dragdestination.html#//apple_ref/doc/uid/20000977-BAJBJFBG
+		///  (5) Other macOS drag/drop articles and blogs refer to dragging/dropping directly from Window - although
+		///      give no examples.
```
dr1rrb, line 94: 😮 Really nice documentation!

I'm wondering however if this is referenced only in old version of the doc, is it possible that it's deprecated?

robloo: Its possible, good luck finding documentation stating that though :) macOS documentation is quite poor but draggingEntered, current version, references this for NSWindow: https://developer.apple.com/documentation/appkit/nsdraggingdestination/1416019-draggingentered

src/Uno.UI/Controls/Window.macOS.cs
robloo: It is such a minor user facing issue. No plus icon appears next to the cursor at this point. Just the dragging image frame will zoom out slightly over an Uno App indicating it is a target. Without the plus appearing though it is clearly not supporting drop unless we accept it which is what users would expect. It is possible to maintain a global list of registered UIElement and their accepted drop types. However we would also need notification to every window when this global list changes. It seems like a lot of work that still wouldn't completely fix this. It might be better to wait and see if there are user complaints before attempting something here. I expect most will never notice.

src/Uno.UWP/UI/Core/CoreWindow.macOS.cs
dr1rrb: Should definitely be `internal` (and converted to a property like `NativeWindow`). 

Otherwise it's 100% legit to need access to the native window (in internal code), and you can either keep it here, either move it to the `Window.UI.Xaml.Window`. Personally I think it's better here, but it might be easier to discover in the other one.

src/Uno.UI/UI/Xaml/DragDropManager.cs
dr1rrb: Yes at some point it will be supported ... but for now this is good enough 👍

src/Uno.UWP/ApplicationModel/DataTransfer/DataPackage.macOS.cs
```diff
@@ -51,14 +81,6 @@ internal static void SetToNativeClipboard(DataPackage content)
...
 		}
 
-		/// <summary>
-		/// Asynchronously sets the contents of the <see cref="DataPackage"/> to the native clipboard.
-		/// Any conversion between supported standard types is done automatically.
-		/// </summary>
-		/// <param name="content">The contents to set to the native cipboard.</param>
-		internal static Task SetToNativeClipboardAsync(DataPackage content)
-			=> SetToNativeAsync(content, NSPasteboard.GeneralPasteboard);
-
```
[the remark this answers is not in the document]
robloo: Yes, I know. It was in the PR Other information section to circle back and add this :) I had to remove it so I could cherry-pick commits without conflicts.

src/Uno.UWP/ApplicationModel/DataTransfer/DataPackage.macOS.cs
```diff
@@ -69,22 +91,167 @@ internal static DataPackageView GetFromNativeClipboard()
...
+					using (MemoryStream ms = new MemoryStream())
+					{
+						await stream.CopyToAsync(ms);
+						ms.Flush();
+						ms.Position = 0;
+
+						image = NSImage.FromStream(ms);
+					}
+
+					stream.Close();
```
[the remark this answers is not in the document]
robloo: Close automatically calls Dispose(true) internally. However, I do normally put this in a using to be safe. Even the docs say to call Dispose... so ooops.
https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.close?view=netcore-3.1

src/Uno.UWP/ApplicationModel/DataTransfer/DataPackage.macOS.cs
```diff
@@ -69,22 +91,167 @@ internal static DataPackageView GetFromNativeClipboard()
...
+				}	
+
+				if (image != null)
+				{
+					draggingItem = new NSDraggingItem(image);
+
+					// For an NSImage, we will use the image itself as the dragging visual.
+					// The visual should be no larger than the max dimension setting and is therefore scaled.
```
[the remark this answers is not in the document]
robloo: I had the same idea but DragUI isn't implemented right now. See line 122 above in the comments for the plan here.

Right now it serves as a proof-of-concept for the drag image. If you are going to get the DragUI done soon, great! This can be added later.

src/Uno.UWP/ApplicationModel/DataTransfer/DataPackage.macOS.cs
```diff
@@ -69,22 +91,167 @@ internal static DataPackageView GetFromNativeClipboard()
...
+			}
+
+			if (data?.Contains(StandardDataFormats.Html) ?? false)
+			{
+				var html = await data.GetHtmlFormatAsync();
+
+				if (!string.IsNullOrEmpty(html))
+				{
+					var pasteboardItem = new NSPasteboardItem();
+					pasteboardItem.SetStringForType(html ?? string.Empty, NSPasteboard.NSPasteboardTypeHTML);
```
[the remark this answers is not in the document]
robloo: Yea, I'll admit that is probably a relic from earlier code. I am also a big fan of over-checking for null as code tends to move around but certainly I can remove this. `!` may not be required here so I'll check next time I'm in macOS.

src/Uno.UWP/ApplicationModel/DataTransfer/DataPackage.macOS.cs
```diff
@@ -69,22 +91,167 @@ internal static DataPackageView GetFromNativeClipboard()
...
+
+			if (data?.Contains(StandardDataFormats.Rtf) ?? false)
+			{
+				var rtf = await data.GetRtfAsync();
+
+				if (!string.IsNullOrEmpty(rtf))
+				{
+					// Use `NSPasteboardTypeRTF` instead of `NSPasteboardTypeRTFD` for max compatiblity
+					var pasteboardItem = new NSPasteboardItem();
+					pasteboardItem.SetStringForType(rtf ?? string.Empty, NSPasteboard.NSPasteboardTypeRTF);
```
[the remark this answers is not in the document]
robloo: Same as above

src/Uno.UWP/UI/Core/CoreWindow.macOS.cs
robloo: Ok, if there aren't any concerns adding a new internal property, I'll go ahead and add it!

src/Uno.UI/Controls/Window.macOS.cs
```diff
@@ -44,8 +44,210 @@ public Window(CGRect contentRect, NSWindowStyle aStyle, NSBackingStore buffering
...
+			{
+				return NSDragOperation.None;
+			}
+		}
+
+		/// <summary>
+		/// Code to execute when the <see cref="DraggingEntered(NSDraggingInfo)"/> method is invoked.
+		/// This can be used in a similar way to <see cref="UIElement.DragEnter"/>.
+		/// </summary>
+		public Func<NSDraggingInfo, NSDragOperation> DraggingEnteredAction { get; set; } =
```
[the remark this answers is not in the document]
robloo: I actually decided against using events because:
 - There is nothing stopping more than one object from subscribing to an event, that wouldn't make sense in this case.
 - I need returned information -- also very strange with an event although it could be done with a property in a new eventArgs class. 
 - This design more closely follows how macOS actually works. It's invoking a method to be handled internally -- not notifying any observers to take appropriate action. Fundamentally, it is a different concept from what we would normally see in c#/UWP so I implemented as such.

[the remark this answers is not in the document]
robloo: I can switch to using a good old-fashioned delegate if it's needed :) I wasn't aware of any performance concerns with Func or Action usage though. In fact, these are widely used now days and recommended over the original delegate in cases like this, easier to read and what not. Is there any information you can share about those performance concerns? Is that something that only is a concern in an AOT scenario? I also know Uno itself uses these all over the place so wonder what the criteria is for decided where to use them?

src/Uno.UWP/ApplicationModel/DataTransfer/DataPackage.macOS.cs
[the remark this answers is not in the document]
robloo: Just to refresh my memory on this: I confirmed my understanding is correct that Close() is all that is needed here assuming the stream classes are implemented following the docs. This has been a long point of confusion but Close() and Dispose(true) are -- and have to be for compatiblity -- equivalent. I'll still try to put this inside a using and remove Close() just to make it more readable and follow the latest conventions.

src/Uno.UWP/ApplicationModel/DataTransfer/DataPackage.macOS.cs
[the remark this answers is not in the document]
robloo: Made the change without `!` which wasn't required as the Xamarin API does not have nullable annotations.

src/Uno.UI/UI/Xaml/DragDropExtension.macOS.cs
robloo: If no one knows of a solution to this the 'hack' I can think of would be to abort the drag within Uno as soon as it is handed off to the system. This means either (1) Dragging back into Uno wouldn't be possible or (2) Dragging back into Uno would create a new dragging session as if it started externally. I think if (2) is possible it is the best that can be done and much better than leaving two dragging sessions active: one inside Uno, one externally.

src/Uno.UI/Controls/Window.macOS.cs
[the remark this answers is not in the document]
dr1rrb: Yeah I would agree that writing our own delegate is no longer the norm. It's something that we discovered while doing some performance passes and which is specific to Xamarin / mono in AOT mode. For more details we gonna have to ask to the master of the subject: @jeromelaban 🤓. But for now as the rule is to use "old-fashioned delegate", could you please change them?

I've no real concern of not using an event as they are indeed not intended to be listen by application, but in that case I would make those properties `internal`. But even then, as it's in the `Uno` namespace, it's just a suggestion!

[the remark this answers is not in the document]
robloo: Yea, even the Microsoft docs say not to create typed delegates for every signature (code bloat) and instead use Func and Action these days. If this is some AOT issue with generics that does make sense to me. Seems like a Xamarin issue should be filed as well though.

I'll change the code hopefully by end of today.

src/Uno.UI/UI/Xaml/DragDropExtension.macOS.cs
dr1rrb: Just had a look in the macOS contract, did you notice that the [`NSDraggingSource` is a protocol since macOS 10.7](https://developer.apple.com/documentation/appkit/nsdraggingsource) and it does have 3 methods `draggingSession` for "_endedAt_", "_movedTo_" and "_willBeginAt_" which looks like to match our needs.

Unfortunately the Xamarin's macOS binding does reflect the old contract (which is no longer present in apple doc but can be found in the [archived documentation](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/DragandDrop/Concepts/dragsource.html)) that supports only dragging images. 

I would assume that if you create your own source which inherits from the `NSDraggingSource` (you can also try to create a new one which inherits from `NSObject` and implements the [`INSDraggingSource`](https://docs.microsoft.com/en-us/dotnet/api/appkit.insdraggingsource?view=xamarin-mac-sdk-14) but sounds more risky), then you should be able to implements those methods by adding the `Export` attribute as you did for the Dragging* methods on the `Window`.

Otherwise I would try to go for option (2), the (1) wouldn't be really user friendly :/

robloo: I did look at DraggingSource at one point -- clearly not enough :) I missed those methods so thanks for bringing it to my attention! It would make sense if that solves this.

robloo: As usual what should have been a quick change turned into a few hours of trial and error. The [current macOS docs](https://developer.apple.com/documentation/appkit/nsdraggingsource) appear to be wrong or are somehow incompatible with Xamarin. The following code never works. I tried almost everything.

```csharp
private class BindableNSDraggingSource : NSDraggingSource
{
	[Export("draggingSession:willBeginAt:")]
	public void DraggingSessionWillBegin(NSDraggingSession session, CGPoint screenPoint)
	{
		return;
	}

	[Export("draggingSession:endedAt:operation:")]
	public void DraggingSessionEnded(NSDraggingSession session, CGPoint screenPoint, NSDragOperation operation)
	{
		return;
	}
}
 ```

However, what does work are the old methods -- still present for backwards compatibility. I found the original signatures [here](http://mirror.informatimago.com/next/developer.apple.com/documentation/Cocoa/Reference/ApplicationKit/ObjC_classic/Protocols/NSDraggingSource.html) which corresponds to the [legacy docs](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/DragandDrop/Concepts/dragsource.html).

```csharp
private class BindableNSDraggingSource : NSDraggingSource
{
	[Export("draggedImage:beganAt:")]
	public void DraggingSessionWillBegin(NSImage session, CGPoint screenPoint)
	{
	}

	[Export("draggedImage:movedTo:")]
	public void DraggingSessionMoved(NSImage session, CGPoint screenPoint)
	{
	}

	[Export("draggedImage:endedAt:operation:")]
	public void DraggingSessionEnded(NSImage session, CGPoint screenPoint, NSDragOperation operation)
	{
	}
}
```

It's a bit ugly but the DraggingSessionEnded method is invoked now when an external drop is completed... yikes.

robloo:
@dr1rrb This should be all set for a final review and merge. There are a few things that may need to be addressed in the future:
 - Use `Content` of the DragUI to create the native dragging frame image. Since macOS has dragging frame content (image) for each NSDraggingItem, we should probably only set the image for the first data type and skip it for the others.
 - The DragUI property of CoreDragInfo should probably be DragUI? type instead of object?
 - Performance could be improved - especially Uno->External. How is currently unknown.

dr1rrb:
Thanks a lot for your contribution @robloo , it's really nice 😍!

About your remarks:

> 
>     * Use `Content` of the DragUI to create the native dragging frame image. Since macOS has dragging frame content (image) for each NSDraggingItem, we should probably only set the image for the first data type and skip it for the others.

Actually I think that the `Content` should replicate the UWP behavior: a "screenshot" of the dragged element by default, and it can then be changed by the application using [`DragUIOverride`](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.draguioverride.setcontentfrombitmapimage). It would then be the choice of the application to render the content of one of the dragged item ... or anything else.

> 
>     * The DragUI property of CoreDragInfo should probably be DragUI? type instead of object?

I would have loved to, unfortunately the `CoreDragInfo` is in the `Uno` assembly while the `DragUI` is in the `Uno.UI`, which is dependent upon the `Uno` one.

> 
>     * Performance could be improved - especially Uno->External. How is currently unknown.

Yeah, the UWP API is async (for good reason) while native APIs aren't, so we have to pay the cost when we leave the application's window bounds. For now as a MVP I think it's acceptable, we will then have to monitor if it's a real concern in real world or just theoretical.

src/Uno.UI/Controls/Window.macOS.cs
robloo: Ok, of course in separate docs it would be expanded as an overview as well.

robloo:
This last comment was really more of a note for myself if I take a look at this again:

> Actually I think that the Content should replicate the UWP behavior: a "screenshot" of the dragged element by default, and it can then be changed by the application using DragUIOverride. It would then be the choice of the application to render the content of one of the dragged item ... or anything else.

I do understand this and I think we are in agreement on how to implement it. As soon as Uno generates a proper dragging image this code can be updated. My comments at the end "Since macOS has dragging frame content (image) for each NSDraggingItem..." are notes for macOS only. In macOS every single type of dragged data can have its own image set in the dragging frame. You can also have multiple of the same types of data -- images for example -- dragged at once. This allows stacks of images to appear during a drag. This is different from UWP which only supports one type of data (except for StorageItems) within the DataPackage and only one image separate from the DataPackage for the entire dragging operation. For macOS we will likely have to just set the DragUI image for the first data type only. 

> I would have loved to, unfortunately the CoreDragInfo is in the Uno assembly while the DragUI is in the Uno.UI, which is dependent upon the Uno one.

Ah, ok, makes sense.

> Yeah, the UWP API is async (for good reason) while native APIs aren't, so we have to pay the cost when we leave the application's window bounds. For now as a MVP I think it's acceptable, we will then have to monitor if it's a real concern in real world or just theoretical.

Congrats on the MVP :) Again, this is more macOS specific notes for myself in the future: 
 1. Starting a drag, even internally within an Uno App, is very slow. I haven't had a chance to investigate this following your ideas and probably won't for the foreseeable future. This is noticeably slower than other platforms such as Android and Skia running the same test page.
 2. Dragging externally does exactly as you say and it takes a while to build the data as soon as it leaves the app. However, this is largely worse-case as I intentionally added a large image in the drag/drop test page to validate this sort of worse-case scenario. Nevertheless, you will notice a marked delay when dragging a large image out of an Uno app and before the UI updates to preview it. This can probably be improved with internal image source handling changes (rendering only to a desired size) but for now it is quite noticeable -- it is not a theoretical concern.

src/Uno.UWP/ApplicationModel/DataTransfer/DataPackage.macOS.cs
```diff
@@ -36,7 +42,32 @@ public partial class DataPackage
...
+		 * only one may be used at a time in the above defined priority. 
+		 * WebLink is considered more specific than ApplicationLink.
+		 *
+		 * --- macOS to UWP ---
+		 *
+		 * A macOS URL must be specially mapped for UWP as UWP's direct equivalent 
+		 * standard data format 'Uri' is deprecated.
+		 *
+		 * 1. WebLink is used if the macOS URL has a scheme of http or https 
+		 * 2. ApplicationLink is used if not #1
```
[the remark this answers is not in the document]
robloo: As I was copying the code from the Android DataPackage updates I was also wondering how this could be put in one location to avoid duplication. Putting it as an extension or helper method some place would also allow us to quickly implement the logic on other platforms. Some ideas are below. Not sure where to locate these methods:

`bool IsWebLink(string url)` : would at least allow sharing the detection code, ideally we would like to share more logic.
`void SeparateUrl(string url, out string? webLink, out string? applicationLink)` : Slightly better as all logic could be shared now.

Going the other way:

`string CombineUrl(string? webLink, string? applicationLink, string? uri = null)`

If we get common methods some place I would be happy to update Skia. Generally speaking, I do have to be careful about spending much time on platforms that are not currently on the roadmap for my apps. As awesome as Skia is I must put that time elsewhere :)

src/Uno.UI/UI/Xaml/DragDropExtension.macOS.cs
```diff
@@ -0,0 +1,319 @@
...
+				CGPoint windowLocation = new CGPoint(0.0, 0.0);
+				if (_macOSDraggingInfo != null)
+				{
+					windowLocation = _window.ContentView.ConvertPointFromView(_macOSDraggingInfo.DraggingLocation, null);
+				}
+				var location = new Windows.Foundation.Point(windowLocation.X, windowLocation.Y);
+
+				// macOS requires access to NSEvent.ModifierFlags to determine key press.
+				// This isn't available from the dragging info.
+				var mods = DragDropModifiers.None;
```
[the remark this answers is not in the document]
robloo: Ok, I'll add a note for that.

src/Uno.UI/UI/Xaml/DragDropExtension.macOS.cs
robloo: Ah, I did not see these were in the Xamarin class. Wasn't really looking honestly since they weren't in the macOS docs.

src/Uno.UWP/ApplicationModel/DataTransfer/DataPackage.macOS.cs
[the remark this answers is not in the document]
robloo: @dr1rrb When you have a moment please take a look at the changes in:

 * src/Uno.UWP/ApplicationModel/DataTransfer/DataPackage.cs and
 * src/Uno.UWP/ApplicationModel/DataTransfer/Clipboard.Android.cs

Corresponding with the last two commits. I went ahead and separate out this logic and added it to the Android clipboard. I think it does nicely simplify some things. If you are good with it I'll apply these same changes to macOS/Skia.

dr1rrb (approved):

robloo:
I wasn't working in macOS today so didn't have a chance to make the updates to use Uri methods there. I also didn't make the doc changes you requested yet. I guess to avoid this in the future I'm going to be more clear with draft PRs. I dont know how Uno usually handles things but I was expecting two approvals and a bit of feedback with my last comments to know a merge was happening.

PR #4337, opened 2020-10-21
feat: ContactPicker

### LINKED ISSUE #4336: Add support for ContactPicker API

`ContactPicker` API

## Why is this needed:

UWP parity

## For which Platform:

- [x] iOS
- [x] Android
- [x] WebAssembly
- [ ] macOS
- [x] Skia
  - [ ] WPF
  - [ ] GTK (Linux)
  - [x] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

macOS does not have API to pick contacts directly but could be possible by writing a custom UI. For that, we will need `ContactManger` too.

MartinZikmund:
GitHub Issue (If applicable): closes #4336 

## PR Type

- Feature

Not supported.

## What is the new behavior?

Supported on iOS, Android, WASM and Tizen.

## PR Checklist


carldebilly:
@MartinZikmund Is this working? Why this is still a draft PR?

MartinZikmund:
@carldebilly yes, it should be working, but I wanted to add support to pick multiple contacts on some platforms which support it. I just got sidetracked with `TabView` and `NavigationView` at the end of 2020 😀 . I will try to finalize this for 3.5

src/Uno.UWP/ApplicationModel/Contacts/Contact.cs
```diff
@@ -0,0 +1,105 @@
...
+	public partial class Contact
+	{
+		/// <summary>
+		/// Gets the full name of the Contact.
+		/// </summary>
+		public string FullName
+		{
+			get
+			{
+				return string.Join(" ", new [] { HonorificNamePrefix, FirstName, MiddleName, LastName, HonorificNameSuffix }.Where(n => !string.IsNullOrEmpty(n)));
```
dr1rrb, line 21: Not sure is this should be customized per culture. For instance in France it's pretty common to put the `LastName` (a.k.a. family name) first.

Definitely not a blocker, if we are able to easily (i.e. copy it :P ) fix that, it would be really nice.

src/Uno.UWP/ApplicationModel/Contacts/ContactPicker.Android.cs
```diff
@@ -0,0 +1,384 @@
...
+		{
+			if (!await PermissionsHelper.CheckReadContactsPermission(default))
+			{
+				if (!await PermissionsHelper.TryGetReadContactsPermission(default))
+				{
+					throw new InvalidOperationException("android.permission.READ_CONTACTS permission is required");
+				}
+			}
+
+			if (ContextHelper.Current == null)
```
dr1rrb, line 30: Does the ApplicationContext is enough? (We try to avoid the usage of the `ContextHelper`)

src/Uno.UWP/ApplicationModel/Contacts/ContactPicker.cs
```diff
@@ -0,0 +1,50 @@
...
+	/// </summary>
+	public partial class ContactPicker
+	{
+		/// <summary>
+		/// Gets a Boolean value indicating if the contact picker is supported on the current platform.
+		/// </summary>
+		/// <returns>A Boolean value indicating if the contact picker is supported on the current platform.</returns>
+		/// <remarks>Unsupported platforms return false.</remarks>
+		public static IAsyncOperation<bool> IsSupportedAsync() =>
+			IsSupportedTaskAsync().AsAsyncOperation();
```
dr1rrb, line 23: Even if not used, it would be better if the `IsSupportedTaskAsync` accepts a `CancellationToken`

dr1rrb (approved):

src/Uno.UWP/ApplicationModel/Contacts/ContactPicker.Android.cs
davidjohnoliver: I missed that, when did we start trying to avoid `ContextHelper`? So far as I know it's fine

image

src/Uno.UWP/ApplicationModel/Contacts/Contact.cs
MartinZikmund: Hmm, you are right the order of the name parts should be culture dependent as per documentation, but not sure if there is some definitive guide on how to do it :-D .

src/Uno.UWP/ApplicationModel/Contacts/ContactPicker.Android.cs
dr1rrb: @davidjohnoliver well, for my understanding we try to avoid it to access to the current `Activity` as it's usually assumed to be a single instance while it could change overtime ... and it could even not be set if accessed too early. But it's also known that we have no other choices so far to use it sometimes, and actually here as it's just a check I would assume it's valid case.

Maybe @jeromelaban could give more details?

carldebilly (approved):

src/Uno.UWP/ApplicationModel/Contacts/ContactPicker.Android.cs
MartinZikmund: Turns out the ContextHelper is not used here anyway, so I removed it

src/Uno.UWP/ApplicationModel/Contacts/Contact.cs
pkar70: So, why implement this field? Erroneous implementation is much worse than no implementation.
Field should be left as unimplemented.
(I know, it is already merged...)

PR #4374, opened 2020-10-25
DO NOT MERGE fix!: Adjust Loading, DataContextChanged event types to match UWP

### LINKED ISSUE #2870: FrameworkElement.DataContextChanged has different signature from UWP
## Current behavior

`DataContextChanged` event has a different signature in Uno than in UWP. If it is not intentional, maybe it could be added to #2240 ?

UWP uses:
```
TypedEventHandler<FrameworkElement, DataContextChangedEventArgs>
```

While Uno has:
```
TypedEventHandler<DependencyObject, DataContextChangedEventArgs>
```

Is this difference intentional?

## Expected behavior

The event handler is the same.

Attach a DataContextChanged event to page in UWP project head. Notice that other platforms now fail to compile until changed to DependencyObject.

## Environment

Nuget Package: latest

Package Version(s): latest

Affected platform(s):

- [x] iOS
- [x] Android
- [x] WebAssembly
- [x] macOS
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

MartinZikmund:
GitHub Issue (If applicable): fixes #2870

## PR Type

- Bugfix

The event types don't match UWP.

## What is the new behavior?

Match UWP.

## PR Checklist

## Other information

**Although this is a breaking change, existing Uno apps must already use `DependencyObject` as sender for both changed events, which will continue to work even after the change (as `FrameworkElement` derives from `DependencyObject`).**

### DIFF
--- a/src/SourceGenerators/Uno.UI.SourceGenerators/DependencyObject/DependencyObjectGenerator.cs
+++ b/src/SourceGenerators/Uno.UI.SourceGenerators/DependencyObject/DependencyObjectGenerator.cs
@@ -542,7 +542,7 @@ namespace Uno.UI.SourceGenerators.DependencyObject
 					private readonly static IEventProvider _binderTrace = Tracing.Get(DependencyObjectStore.TraceProvider.Id);
 					private BinderReferenceHolder _refHolder;
 
-					public event Windows.Foundation.TypedEventHandler<DependencyObject, DataContextChangedEventArgs> DataContextChanged;
+					public event Windows.Foundation.TypedEventHandler<FrameworkElement, DataContextChangedEventArgs> DataContextChanged;
 
 					partial void InitializeBinder();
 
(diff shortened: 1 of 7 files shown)


jeromelaban:
This will make a binary breaking change that we can't take at this point (it may break dependents as well, e.g. Prism or WCT).

MartinZikmund:
I see, I was thinking in terms of app which is reckmpiled, because in such case it won't have effect. However, the sooner we can fix it, the better, as the end user experience is that VS generates the UWP handler, which then causes compile time error, which is very confusing for the developer

carldebilly:
@jeromelaban Should be put in the queue for the next "breaking change"

jeromelaban:
It is already.

MartinZikmund:
@davidjohnoliver I will close this PR, but feel free to cherry-pick the commit to breaking changes if useful

Issue #4426, opened 2020-10-30
WinUI InfoBar


MartinZikmund:
## What would you like to be added:

WinUI InfoBar control

## Why is this needed:

WinUI parity

## For which Platform:

- [ ] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] WebAssembly renderers for Xamarin.Forms
- [ ] macOS
- [ ] Skia
  - [ ] WPF
  - [ ] GTK (Linux)
  - [ ] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

MartinZikmund:
Done! 🎆

PR #4427, opened 2020-10-30
WinUI InfoBar

MartinZikmund:
GitHub Issue (If applicable): #4426 

## PR Type

- Feature

Not supported.

## What is the new behavior?

InfoBar supported.

## PR Checklist


src/Uno.UI/Microsoft/UI/Xaml/Controls/InfoBar/Strings/en-us/Resources.resw
```diff
@@ -0,0 +1,140 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<root>
```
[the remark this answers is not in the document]
robloo: Since I just browed this in WinUI I can answer. Yes, only en-US strings are available.  It's almost always a separate commit after the translation team is done to add the others. Sometimes this can take several weeks it seems and this control is brand-new. Martin is too quick!

https://github.com/microsoft/microsoft-ui-xaml/tree/master/dev/InfoBar

src/Uno.UI/Microsoft/UI/Xaml/Controls/InfoBar/InfoBar.Events.cs
```diff
@@ -0,0 +1,13 @@
+﻿using Windows.Foundation;
+
+namespace Microsoft.UI.Xaml.Controls
+{
+	public partial class InfoBar
```
robloo, line 5: Is this what we are doing now? Actually, I always like splitting out Events and Properties in separate files as that matches the documentation structure and is also latest convention with c# controls. However, this is not done in WinUI and I thought fidelity with original code was highest priority?

We should be fairly consistent with ported controls and add .Events.cs for the others if that's what we agree.

src/Uno.UI/Microsoft/UI/Xaml/Controls/InfoBar/Strings/en-us/Resources.resw
```diff
@@ -0,0 +1,140 @@
...
+  <data name="InfoBarClosedNotification" xml:space="preserve">
+    <value>InfoBar dismissed</value>
+    <comment>The string read by narrator when the InfoBar closes.</comment>
+  </data>
+  <data name="InfoBarCustomLandmarkName" xml:space="preserve">
+    <value>InfoBar</value>
+    <comment>This is the custom landmark used to denote an InfoBar to narrator.</comment>
+  </data>
+  <data name="InfoBarOpenedNotification" xml:space="preserve">
+    <value>%1!s! %2!s!</value>
```
robloo, line 137: Again we have format strings that won't work in c# I think. We need a plan for this. ColorPicker has the same unresolved problem. We probably need a special string.FormatWinUI() method that will convert from cpp-style to c#-style and then pass to string.Fromat().

MartinZikmund: Agreed, this needs to be adjusted. I didn't notice that :-D .

src/Uno.UI/Microsoft/UI/Xaml/Controls/InfoBar/InfoBar.Events.cs
MartinZikmund: I have been doing this for the controls I have ported so far. I think it is not a big problem as the events are defined in the header file/.idl anyway, so if we really wanted to be precise, we would have to split the structure to "cpp" and "h" or something like that. The WinUI porting story will definitely evolve over time, so I would just opt for an approach that works and we can refactor as needed over time. The public interface will remain the same anyway :-) .

src/Uno.UI/Microsoft/UI/Xaml/Controls/InfoBar/Strings/en-us/Resources.resw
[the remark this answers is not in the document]
MartinZikmund: Yes, so far there are no translations :-D . I think they will be added soon though.

src/Uno.UI/Microsoft/UI/Xaml/Controls/InfoBar/InfoBar.Events.cs
robloo: Actually, the events are also in the generated [.properties.cpp files](https://github.com/microsoft/microsoft-ui-xaml/blob/f5034668462987302f1e1c58d1e055e45e40c658/dev/Generated/InfoBar.properties.cpp#L437). Obviously we merge .cpp and .h files together into one .cs file for c#. The IDL is only used to generate the source code and set initial values. We shouldnt use it here other than to make sure default values match. In the future we could generate properties based on the IDL just like WinUI though. I'm getting off track, my point was everything is in .properties.cs in WinUI.

MartinZikmund: I see, I didn't notice that. I can move the events in Properties file, but need to adjust my previous controls the same way too

robloo: Well, I wouldn't redo anything in past controls. It can be changed when the control is updated to latest WinUI next. Or forgotten about since it works just fine :) Probably we should keep everything in the properties file for new contrils. Maybe change that when/if we generate from IDL. I will file an issue for some ideas today as to keep in sync long term we will need some automation.

jeromelaban (approved):

src/Uno.UI/Microsoft/UI/Xaml/Controls/InfoBar/Strings/en-us/Resources.resw
MartinZikmund: @robloo turns out this is in fact already handled in Uno :-D  - `Uno.UI.Helpers.WinUI.StringUtil.cs`

robloo: Great, I dont think many knew about that! I'll take a look then update ColorPicker as well.

Issue #4463, opened 2020-11-04
[WASM] Build fails if trying to use ThemeResource inside Flyout


peternary:
## Current behavior

If a control inside a Flyout makes use of a ThemeResource WASM will fail to build.

## Expected behavior

WASM should build and resource be applied appropriately.

## How to reproduce it (as minimally and precisely as possible)

Attempt to build the attached project in both UWP and Wasm.  Notice that UWP builds and performs as expected, but WASM refuses to build.

[WasmThemeResource.zip](https://github.com/unoplatform/uno/files/5490446/WasmThemeResource.zip)

## Environment

Nuget Package:
- [x] Uno.UI / Uno.UI.WebAssembly / Uno.UI.Skia
- [ ] Uno.WinUI / Uno.WinUI.WebAssembly / Uno.WinUI.Skia
- [ ] Uno.SourceGenerationTasks
- [ ] Uno.UI.RemoteControl / Uno.WinUI.RemoteControl
- [ ] Other: 

Nuget Package Version(s):
3.2.0-dev.434
Affected platform(s):

- [ ] iOS
- [ ] Android
- [x] WebAssembly
- [ ] WebAssembly renderers for Xamarin.Forms
- [ ] macOS
- [ ] Skia
  - [ ] WPF
  - [ ] GTK (Linux)
  - [ ] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

IDE:

- [ ] Visual Studio 2017 (version: )
- [ ] Visual Studio 2019 (version: )
- [ ] Visual Studio for Mac (version: )
- [ ] Rider Windows (version: )
- [ ] Rider macOS (version: )
- [ ] Visual Studio Code (version: )

Relevant plugins:

- [ ] Resharper (version: )

DierkDroth:
@jeromelaban this is pretty much WASM show stopper for us.

PR #4498, opened 2020-11-10
fix(Flyout): WASM - Fixed build failing when attempting to create a f…

LHuss:
…lyout containing a control using a themeresource

GitHub Issue (If applicable): #4463 

## PR Type

- Bugfix

If a flyout contains a control that uses a themeresource, the build fails.

## What is the new behavior?

If a flyout contains a control that uses a themeresource, the build succeeds.

## PR Checklist

## Other information

### DIFF
--- a/src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.Reflection.cs
+++ b/src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.Reflection.cs
@@ -129,26 +129,38 @@ namespace Uno.UI.SourceGenerators.XamlGenerator
 		{
 			var type = FindType(xamlType);
 
-			if (type != null)
+			return IsType(type, typeSymbol);
+		}
+
+		private bool IsType(INamedTypeSymbol namedTypeSymbol, ISymbol typeSymbol)
+		{
+			if (namedTypeSymbol != null)
 			{
 				do
 				{
-					if (Equals(type, typeSymbol))
+					if (Equals(namedTypeSymbol, typeSymbol))
 					{
 						return true;
 					}
 
-					type = type.BaseType;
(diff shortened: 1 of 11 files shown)


src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.Reflection.cs
```diff
@@ -129,28 +129,40 @@ private bool IsType(XamlType xamlType, ISymbol typeSymbol)
 		{
 			var type = FindType(xamlType);
 
-			if (type != null)
+			return IsType(type, typeSymbol);
+		}
+
+		private bool IsType(INamedTypeSymbol symbol, ISymbol typeSymbol)
```
davidjohnoliver, line 135: The parameter names here are a bit confusing (particularly compared to their respective types)

davidjohnoliver (approved):
This will unblock the use case. 

However I think that there's still a bug that these theme resource bindings (and probably ElementName bindings?) won't update correctly, because the update method is not getting hooked up. So we should probably keep the related issue open.

jeromelaban:
> This will unblock the use case.
> 
> However I think that there's still a bug that these theme resource bindings (and probably ElementName bindings?) won't update correctly, because the update method is not getting hooked up. So we should probably keep the related issue open.

@davidjohnoliver in this case, because it's a flyout, the theme updating will likely not be impacted much because the flyout is likely to be closed and reopened. If it's an issue, let's open another one.

jeromelaban:
UWP does not seem to like this:

```
src\SamplesApp\UITests.Shared\Windows_UI_Xaml_Controls\Flyout\Flyout_Local.xaml(9,5): Error WMC0011: Unknown member 'Background' on element 'Flyout'
```

jeromelaban (approved):

PR #4598, opened 2020-11-23
fix: ViewHelper.Android not considering IgnoreTextScaleFactor at runtime

dmitry-sv:

## PR Type

- Bugfix

Setting `IgnoreTextScaleFactor = true` doesn't prevent Android app from changing text scale when device's text size is changed while the app is running.

## What is the new behavior?

Setting `IgnoreTextScaleFactor = true` always prevents Android app from changing text scale.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/Extensions/ViewHelper.Android.cs
+++ b/src/Uno.UI/Extensions/ViewHelper.Android.cs
@@ -78,16 +78,11 @@ namespace Uno.UI
 		{
 			using (Android.Util.DisplayMetrics displayMetrics = Android.App.Application.Context.Resources.DisplayMetrics)
 			{
+				AdjustScaledDensity(displayMetrics);
+
 				// WARNING: The Density value is not completely based on the DPI of the device.
 				// On two 8" devices, the Density may not be consistent.
 				_cachedDensity = displayMetrics.Density;
-				if (FeatureConfiguration.Font.IgnoreTextScaleFactor)
-				{
-					// To disable text scaling, we put the Density value in ScaledDensity so that the ratio between them is 1.
-					// This ensures it's disabled for everything using ScaledDensity (e.g. TextBlock, TextBox, AppBarButton, etc.)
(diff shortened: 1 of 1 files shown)


src/Uno.UI/Extensions/ViewHelper.Android.cs
```diff
@@ -107,6 +107,12 @@ public static void RefreshFontScale()
 		{
 			using (Android.Util.DisplayMetrics displayMetrics = Android.App.Application.Context.Resources.DisplayMetrics)
 			{
+				if (FeatureConfiguration.Font.IgnoreTextScaleFactor)
+				{
+					// Keep disabling text scaling
+					displayMetrics.ScaledDensity = displayMetrics.Density;
+				}
```
jeromelaban, line 114: This is only doing parts of the work that is done here: 

https://github.com/unoplatform/uno/blob/7e80735642b9f0ed76f54452964afcce7ef1a0d1/src/Uno.UI/Extensions/ViewHelper.Android.cs#L79-L96

We should probably refresh the whole set instead of just those two (and keep the comment that is very insightful).

dmitry-sv: What specifically do you think I should add? 
because `_cachedScaledDensity` will be set right after my changes and I don't see why we should refresh the whole set

dmitry-sv: I mean the comment is insightful, it's worth adding. 
But in my opinion `displayMetrics.Density` and `displayMetrics.Xdpi` are not related to the `RefreshFontScale` method, are they?

jeromelaban: Agreed, the rest of the method does not seem impacted. In that case, let's extract the portion that is the same in both methods to avoid duplication.

jeromelaban (approved):

jeromelaban:
@mergifyio backport release/beta/3.1

PR #4668, opened 2020-12-05
fix(Button): Touch pressed do not update Visual State

microhobby:
Using a touch pointer, touch screen, with Skia GTK does not set the
IsPointerOver only the IsPointerPressed when touching a button. The
check for IsPointerPressed and IsPointerOver is never true for the
touch case, so the Visual State does not change the appearance when
pressed.

I believe that the normal thing is that if we have an IsPointerPressed
true automatically IsPointerOver is also true, I cannot think of any
case where this is not true, but for touch this is getting only the
pressed not the over, please correct me if I am wrong. This commit
removes the IsPointerOver from the check to set the visual state to
"Pressed".

## PR Type

- Bugfix

Touching a button using a touch screen does not change the visual state

## What is the new behavior?

Touching a button using a touch screen changes the visual state correctly

## PR Checklist

## Other information

As I wrote in the commit description I couldn't think of any case where `IsPointerPressed` is true without `IsPointerOver` not being true, so I think that using `IsPointerOver` at check is not necessary, but I may be wrong please let me know.

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/VisualStatesImplementation.tt
+++ b/src/Uno.UI/UI/Xaml/Controls/VisualStatesImplementation.tt
@@ -199,7 +199,7 @@ namespace <#= mixin.NamespaceName #>
 				return __disabledState;
 			}
 #if <#= mixin.HasCommonPressedState #>
-			else if (IsPointerPressed && IsPointerOver)
+			else if (IsPointerPressed)
 			{
 				return __pressedState;
 			}


jeromelaban:
Thanks Matheus! 

@dr1rrb It's likely that we need to simulate the "pointer over" when pressing in that case. Isn't this something we're doing on iOS already ?

dr1rrb:
Thanks @microhobby finally someone took a chance to look at that code 👍

Yeah on iOS (and other platforms) and like UWP we simulate the `Enter/Exit` when touching the screen https://github.com/unoplatform/uno/blob/master/src/Uno.UI/UI/Xaml/UIElement.Pointers.iOS.cs#L105

And actually yes we can have a pressed while not over: if you press an element, then move out of its bounds while holding finger on screen, you will remain `IsPointerPressed = true`, but the `IsPointerOver` flag will be `false`. `IsPointerPressed` is an internal flag which is not cleared until `PointerReleased` as, for instance, with a mouse the `Button` does allow "re-entrance" while pressed. 

Also even if we can choose to automatically clear this pressed flag when pointer is a finger, we still have to raise the `PointerEntered` and `PointerExited` events has some apps do rely on specific events sequence. 

(Welcome to the wonderful land of pointers 😒)

microhobby:
Thanks @dr1rrb , got it. I will take a look at this.

> (Welcome to the wonderful land of pointers 😒)
🙃

microhobby:
cbb4be1c8c349547e0f2d9346a1783b4cd6f81e3

PR #4692, opened 2020-12-08
Fix tapped event bubbling when handled

dr1rrb:
## Bugfix

`Tapped` event is bubbling even if the args are marked as `Handled = true`.

When a gesture and manipulation events bubbles to the parent, it will disable the local gesture recognizer of the parent in order to avoid double event. But if the event args is flagged as `Handled = true` the bubbling stops, so the parent's recognizer still detects a "tapped gesture" so it will fire the `Tapped` event again on it's own.

## What is the new behavior?

Manipulation and gesture events are flagged "always bubbling" so even if handled, they will bubble to the parent and disable the local `GestureRecognizer`

## PR Checklist


MartinZikmund (approved):
🎉🎉🎉

jeromelaban (approved):

MartinZikmund:
@dr1rrb Seems like the offset is applied to the wrong test according to the test assert - `When_Nested` expects 9-11 and gets 35, while `When_Nested_And_Handling` expects 34-36 and gets 10.

PR #4750, opened 2020-12-17
fix: Keep track of current view type even if content is set to null

kazo0:
Closes https://github.com/unoplatform/nventive-private/issues/137

## PR Type

- Bugfix

Setting `Frame.Content` to `null` and then navigating using `Frame.Navigate` causes the page we just navigated from to not be added to the `Frame.BackStack`. Also causes iOS to not navigate at all when using the Native Frame Presenter

## What is the new behavior?

`Frame`'s `BackStack` is intact after navigation and navigation is performed properly.

## PR Checklist

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/Frame/Frame.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Frame/Frame.cs
@@ -43,17 +43,6 @@ namespace Windows.UI.Xaml.Controls
 
 		internal PageStackEntry CurrentEntry { get; set; }
 
-		protected override void OnContentChanged(object oldValue, object newValue)
-		{
-			base.OnContentChanged(oldValue, newValue);
-
-			// Make sure we void CurrentEntry when someone sets Frame.Content = null;
-			if (newValue == null)
-			{
-				CurrentEntry = null;
-			}
-		}
-
 		#region BackStackDepth DependencyProperty
 
 		public int BackStackDepth
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/FramePages/FirstPage.xaml
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/FramePages/FirstPage.xaml
@@ -1,0 +1,14 @@
+﻿<Page
+    x:Class="Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls.FramePages.FirstPage"
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:local="using:Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls.FramePages"
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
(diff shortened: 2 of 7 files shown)


jeromelaban:
Looks like this is going against a change made by @jeanplevesque here: https://github.com/unoplatform/uno/blame/da8327e7c2f5a939e729ae8723829011b7154503/src/Uno.UI/UI/Xaml/Controls/Frame/Frame.cs

@jeanplevesque do you recall why you needed this ?

carldebilly:
The title of your PR should mention the context (the Frame navigation)

src/Uno.UI/UI/Xaml/Controls/Frame/Frame.cs
```diff
@@ -43,17 +43,6 @@ public Frame()
...
-		protected override void OnContentChanged(object oldValue, object newValue)
-		{
-			base.OnContentChanged(oldValue, newValue);
-
-			// Make sure we void CurrentEntry when someone sets Frame.Content = null;
-			if (newValue == null)
-			{
-				CurrentEntry = null;
-			}
-		}
```
carldebilly, line 55: So your only real change is to remove code? ;-)

carldebilly (approved):

jeanplevesque:
> Looks like this is going against a change made by @jeanplevesque here: https://github.com/unoplatform/uno/blame/da8327e7c2f5a939e729ae8723829011b7154503/src/Uno.UI/UI/Xaml/Controls/Frame/Frame.cs

It's probably fine. I did this change because it made sense to me and worked at that time, but I did not validate the UWP behavior. I find it weird that clearing the frame's content doesn't make it truly forget the page type. I definitely didn't have this in mind at that time.

jeanplevesque:
@kazo0 The part I'm not understanding right now is that the bug was only happening on iOS, and the modified code affects all platforms. I wondering were this change is truly impacting iOS specifically.

kazo0:
> @kazo0 The part I'm not understanding right now is that the bug was only happening on iOS, and the modified code affects all platforms. I wondering were this change is truly impacting iOS specifically.

The reason it affects iOS is because [this code](https://github.com/unoplatform/uno/blob/5c53c2f37df48c9d74bd2ef34afe58bc7e43760e/src/Uno.UI/Controls/NativeFramePresenter.iOS.cs#L350-L357) is called when the BackStack is cleared. Without the CurrentEntry, the causes an empty list of `UIViewController`s to be set on the `NavigationController` which iOS does not like.

jeanplevesque:
> > @kazo0 The part I'm not understanding right now is that the bug was only happening on iOS, and the modified code affects all platforms. I wondering were this change is truly impacting iOS specifically.
> 
> The reason it affects iOS is because [this code](https://github.com/unoplatform/uno/blob/5c53c2f37df48c9d74bd2ef34afe58bc7e43760e/src/Uno.UI/Controls/NativeFramePresenter.iOS.cs#L350-L357) is called when the BackStack is cleared. With a null `CurrentEntry`, an empty list of `UIViewController`s are set on the `NavigationController` which iOS does not like.

Ah! Might be a good idea to add some comments over there.
Also, can you make sure that setting `Frame.Content = null` actually clears the view?

jeanplevesque (approved):

jeromelaban:
@mergifyio backport release/beta/3.1

PR #4789, opened 2020-12-22
fix(ado-204739): fixed race condition

bchhun:
Fixed the race condition between the docfx menu and the wordpress menu.

## PR Type

- Bugfix

There's a double top menu that shows up in certain pages within the documentation section managed by `docfx`.

## What is the new behavior?

Fixed the race condition issue between the 2 XHR calls that is made to fetch the `docfx` & the `wordpress` menu.

Ultimately, we only want to keep the `wordpress` menu. 

The `docfx` menu will show up only if the `wordpress` api call fails.

## PR Checklist

## Other information

Internal Issue (If applicable): https://dev.azure.com/nventive/UNO%20Marketing/_workitems/edit/204739


doc/README.md
```diff
@@ -0,0 +1,40 @@
+#Documentation Uno
```
jeromelaban, line 1:
```suggestion
# Documentation Uno

```

jeromelaban (approved):
Thanks ! There's just a simple markdown typo at the top with the header, make sure to squash your commits.

doc/README.md
bchhun: on it

ajpinedam:
@bchhun The CI is complaining because of the commit message, specifically the scope. It seems it does not like this "ado-204739". I would try just with "ado" or without specifying any scope in case "ado" doesn't work.

ajpinedam (approved):

PR #4818, opened 2021-01-03
Fixed UWP manifest to use project executable as app automatically by setting substitution moniker

xperiandri:
The moniker is substituted by MSBuild on build

## PR Type

If you decide to rename UWP assembly name you will get a weird error without any clue how to fix it.
Unless you know the trick you will spend hours if ever find it.

## What is the new behavior?

The error will not occur if you rename UWP assembly

## PR Checklist

### DIFF
--- a/src/SolutionTemplate/UnoSolutionTemplate/UWP/Package.appxmanifest
+++ b/src/SolutionTemplate/UnoSolutionTemplate/UWP/Package.appxmanifest
@@ -29,7 +29,7 @@
 
   <Applications>
     <Application Id="App"
-      Executable="$ext_safeprojectname$.exe"
+      Executable="$targetnametoken$.exe"
       EntryPoint="$ext_safeprojectname$.App">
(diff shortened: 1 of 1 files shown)


davidjohnoliver (approved):

jeromelaban:
This change will break the creation of the UWP project through .NET new (it will fail at runtime).

You'll also need to replace that token here:
https://github.com/unoplatform/uno/blob/ae632a66747dad60998adbca0145bb12a79b3aba/src/SolutionTemplate/Uno.ProjectTemplates.Dotnet/Uno.ProjectTemplates.Dotnet.csproj#L195-L198

xperiandri:
@jeromelaban, the moniker must remain in the UWP manifest. It is not replaced during project creation.
How to achieve that?

jeromelaban:
Ah my bad, I did not notice this originally. It will do then. Thanks !

PR #4841, opened 2021-01-06
fix(passwordbox): Don't fail on setting null on PasswordBox.Password

jeromelaban:
GitHub Issue (If applicable): https://github.com/unoplatform/nventive-private/issues/142

## PR Type

- Bugfix

## What is the new behavior?

Databinding null to `PasswordBox.Password` won't raise an exception and rely on coercion to set an empty string.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI.Tests/Windows_UI_XAML_Controls/PasswordBoxTests/Given_PasswordBox.cs
+++ b/src/Uno.UI.Tests/Windows_UI_XAML_Controls/PasswordBoxTests/Given_PasswordBox.cs
@@ -1,0 +1,99 @@
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Controls;
+using Windows.UI.Xaml.Data;
+
+namespace Uno.UI.Tests.PasswordBoxTests
+{
+	[TestClass]
+	public class Given_PasswordBox
+	{
+		[TestMethod]
+		public void When_Binding_Set_Null()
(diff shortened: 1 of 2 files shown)


YGuerin (commented):
Just for my personal culture, what the difference between setting the property using the property itself vs using the dependency property ?

If I'm not mistaken, SetValue will throw an exception if the value parameter type is not an expected type according to [the documentation](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.dependencyobject.setvalue?view=winrt-19041#Windows_UI_Xaml_DependencyObject_SetValue_Windows_UI_Xaml_DependencyProperty_System_Object_) and the [null keyword](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/null) does not refer to any type.

I might miss something here to fully understand the fix.

davidjohnoliver (approved):

davidjohnoliver:
> Just for my personal culture, what the difference between setting the property using the property itself vs using the dependency property ?
> 
> If I'm not mistaken, SetValue will throw an exception if the value parameter type is not an expected type according to [the documentation](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.dependencyobject.setvalue?view=winrt-19041#Windows_UI_Xaml_DependencyObject_SetValue_Windows_UI_Xaml_DependencyProperty_System_Object_) and the [null keyword](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/null) does not refer to any type.
> 
> I might miss something here to fully understand the fix.

This is a very TextBox-specific thing. There's a null check directly on the `Text` property, and using `SetValue()` bypasses that check.

https://github.com/unoplatform/uno/blob/master/src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.cs#L151-L163

jeromelaban:
Also, there's value coercion to empty string on the Text property. Yet, you're right, this is an odd behavior to say the least :)

PR #4901, opened 2021-01-13
feat: UWP buildable and deployable under Any CPU

### LINKED ISSUE #4892: Build UWP app under AnyCPU

Set the UWP app to build and deploy under AnyCPU (using x86 configuration)

## Why is this needed:

When switching between projects in VS, AnyCPU is the default for most. When the user makes UWP default again, it stays under AnyCPU, which then "breaks" UWP build, as it is not built nor deployed. While seasoned developers will be able to fix that by switching target, it is confusing for new developers trying out Uno.

## For which Platform:

- [ ] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] macOS
- [ ] Skia
  - [ ] WPF
  - [ ] GTK (Linux)
  - [ ] Tizen
- [x] Windows
- [ ] Build tasks
- [x] Solution Templates

MartinZikmund:
GitHub Issue (If applicable): fixes #4892

## PR Type

- Feature

UWP project does not build nor deploy under AnyCPU.

## What is the new behavior?

UWP project builds and deploys its x86 version under AnyCPU.

## PR Checklist

### DIFF
--- a/src/SolutionTemplate/UnoSolutionTemplate.Wizard/UnoSolutionWizard.cs
+++ b/src/SolutionTemplate/UnoSolutionTemplate.Wizard/UnoSolutionWizard.cs
@@ -103,10 +103,6 @@ namespace UnoSolutionTemplate.Wizard
 				{
 				}
 			}
-			else
-			{
-				throw new InvalidOperationException();
-			}
 		}
 
 
--- a/src/SolutionTemplate/UnoSolutionTemplate.Wizard/UnoSolutionWizard.cs
+++ b/src/SolutionTemplate/UnoSolutionTemplate.Wizard/UnoSolutionWizard.cs
@@ -4,9 +4,6 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Security.Cryptography.Xml;
-using System.Text;
-using System.Threading.Tasks;
 using EnvDTE;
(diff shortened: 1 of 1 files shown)


src/SolutionTemplate/UnoSolutionTemplate.Wizard/UnoSolutionWizard.cs
```diff
@@ -79,6 +77,38 @@ private void SetStartupProject()
...
+						}
+					}
+				}
+				catch (Exception)
+				{
+				}
+			}
+			else
+			{
+				throw new InvalidOperationException();
```
jeromelaban, line 108: There's probably no reason to make this fail. If it happens, let's let it continue.

MartinZikmund: @jeromelaban the other methods `SetDefaultConfiguration` and `SetStartupProject` also throw in this case, should I remove the throws everywhere, so the solution creation doesn't fail?

jeromelaban: Right, let's keep it that way for now then. It's unlikely to fail in this case, and if it would, lots more would fail anyways.

jeromelaban (approved):

PR #4984, opened 2021-01-21
fix(iOS): CommandBar BackButtonTitle

ajpinedam:

## PR Type

- Bugfix

In iOS when setting a Global Style for the `CommandBar` to remove the Back Button Title this is not working correctly. Some pages still show the Back Button Text.

## What is the new behavior?
The Back Button Title for iOS will disappear from all Pages

## PR Checklist

## Other information

### DIFF
--- a/src/SamplesApp/SamplesApp.Shared/App.xaml
+++ b/src/SamplesApp/SamplesApp.Shared/App.xaml
@@ -2,8 +2,10 @@
 			 xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
 			 xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
 			 xmlns:xamarin="http://nventive.com/xamarin"
+			 xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
 			 xmlns:ios="http://nventive.com/ios"
 			 xmlns:toolkit="using:Uno.UI.Toolkit"
+			 mc:Ignorable="ios"
 			 RequestedTheme="Light">
 	<Application.Resources>
 		<ResourceDictionary>
@@ -113,15 +115,14 @@
 			</Style>
 
 			<Style x:Key="DefaultCommandBarStyle"
-					TargetType="CommandBar"
-					xamarin:BasedOn="{StaticResource NativeDefaultCommandBar}">
+					TargetType="CommandBar">
 
-				<!-- Remove the back button title (and only leave the back arrow) -->
+				 <!--Remove the back button title (and only leave the back arrow)--> 
 				<ios:Setter Property="(toolkit:CommandBarExtensions.BackButtonTitle)"
 							Value="" />
 			</Style>
 
-			<!-- Default Style -->
+			 <!--Default Style--> 
 			<Style TargetType="CommandBar"
 					BasedOn="{StaticResource DefaultCommandBarStyle}" />
 
--- a/src/SamplesApp/SamplesApp.Shared/App.xaml
+++ b/src/SamplesApp/SamplesApp.Shared/App.xaml
@@ -1,6 +1,9 @@
 ﻿<Application x:Class="SamplesApp.App"
 			 xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
 			 xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+			 xmlns:xamarin="http://nventive.com/xamarin"
+			 xmlns:ios="http://nventive.com/ios"
+			 xmlns:toolkit="using:Uno.UI.Toolkit"
 			 RequestedTheme="Light">
 	<Application.Resources>
 		<ResourceDictionary>
@@ -108,6 +111,19 @@
 				<Setter Property="Background"
 						Value="LightBlue" />
 			</Style>
+
+			<Style x:Key="DefaultCommandBarStyle"
+					TargetType="CommandBar"
+					xamarin:BasedOn="{StaticResource NativeDefaultCommandBar}">
+
+				<!-- Remove the back button title (and only leave the back arrow) -->
+				<ios:Setter Property="(toolkit:CommandBarExtensions.BackButtonTitle)"
+							Value="" />
+			</Style>
+
+			<!-- Default Style -->
+			<Style TargetType="CommandBar"
+					BasedOn="{StaticResource DefaultCommandBarStyle}" />
 
 		</ResourceDictionary>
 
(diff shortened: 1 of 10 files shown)


src/Uno.UI/Controls/CommandBarHelper.iOS.cs
```diff
@@ -33,7 +32,11 @@ public static void PageCreated(UIViewController pageController)
 				// For example the Uno.UI.Toolkit.CommandBarExtensions.BackButtonTitle attached property is often set globally to "" through
 				// a default CommandBar style in order to remove the back button text throughout an entire application.
 				// In order to leverage this information, we create a new CommandBar instance that only exists to "render" the NavigationItem.
-				topCommandBar = new CommandBar();
+				// Since Uno 3.0 objects which are not part of the Visualtree does not get the Global Styles applied. Hence the fact we are manually applying it here.
+				topCommandBar = new CommandBar
+				{
+					Style = Application.Current.Resources[typeof(CommandBar)] as Style
```
jeromelaban, line 38: Can we get the visual tree parent some other way ? (so that implicit styling can work and we can resolve the implicit style through the closest attached parent and not the app ?

/cc @davidjohnoliver

davidjohnoliver: We may be able to, but I still think it's not absolutely necessary. The app developer is not creating any CommandBar themselves, so why they would expect a locally-defined CommandBar style to be applied (to a page with no CommandBar, from the app's perspective)?

src/SamplesApp/SamplesApp.Shared/App.xaml
```diff
@@ -109,6 +112,19 @@
 						Value="LightBlue" />
 			</Style>
 
+			<Style x:Key="DefaultCommandBarStyle"
+					TargetType="CommandBar"
+					xamarin:BasedOn="{StaticResource NativeDefaultCommandBar}">
```
davidjohnoliver, line 117: Does the test work if you omit this `BasedOn` line? If so it's preferable, otherwise this will interfere with any tests, screenshot-diffing etc which was previously using the UWP CommandBar style.

src/SamplesApp/SamplesApp.UITests/Windows_UI_Xaml_Controls/CommandBarTests/UnoSamples_Tests.NativeCommandBar.cs
```diff
@@ -203,5 +203,25 @@ public void When_Navigated_CommandBarShouldKeepColor_Native_Frame()
...
+
+			_app.WaitForElement("NavigateInitialButton");
+			_app.FastTap("NavigateInitialButton");
+
+			_app.WaitForElement("NavigateToPage2Button");
+			_app.FastTap("NavigateToPage2Button");
+
+			_app.WaitForElement("Page2CommandBar");
+
+			var backButtonText = _app.Marked("Back").FirstResult();
```
davidjohnoliver, line 222: I don't recall, does .Marked("SomeText") return a TextBlock if its Text="SomeText"? In other words, would this really fail without the change?

src/Uno.UI/Controls/BindableUIView.iOS.cs
```diff
@@ -43,16 +43,34 @@ public override void SubviewAdded(UIView uiview)
 
 		internal List<UIView>.Enumerator GetChildrenEnumerator() => _shadowChildren.Materialized.GetEnumerator();
 
-		public override void WillRemoveSubview(UIView uiview)
+		public override void AddSubview(UIView view)
 		{
-			base.WillRemoveSubview(uiview);
+			// As iOS will invoke the MovedToWindow on the 'view' (which will invoke the Loaded event)
```
jeromelaban, line 48: Is this supposed to be there ?

src/Uno.UI/Controls/CommandBarHelper.iOS.cs
jeromelaban: I was wondering from a testing perspective, to avoid changing the CommandBar style for the whole sample application, and set it just in the tested sample.

src/Uno.UI/Controls/BindableUIView.iOS.cs
ajpinedam: not at all,  wrong commit (cherry-pick) in the wrong branch.

jeromelaban:
@ajpinedam looks like you need to ignore the `ios` namespace in `app.xaml`

src/Uno.UI/Controls/CommandBarHelper.iOS.cs
davidjohnoliver: Right - the hope (not sure if @ajpinedam has already confirmed) is that the app-level style would only need to set an attached property, which would just be ignored by the UWP-styled CommandBar.

davidjohnoliver: And it would actually be possible to set the app-level style for the duration of the test only, ie by having a button that injected the style into Application.Current.Resources, and then removed it again when the test was completed.

ajpinedam: @davidjohnoliver  that is a wonderful idea. I just updated the UITest following your suggestion. We are now adding the GLobal Style for the CommandBar right before the UItest starts and remove it before leaving the UITest. 

By doing this we have encapsulated the Style to the execution of this UITest and preventing future issues with other UITests that could have applied the Style.

src/SamplesApp/SamplesApp.UITests/Windows_UI_Xaml_Controls/CommandBarTests/UnoSamples_Tests.NativeCommandBar.cs
```diff
@@ -203,5 +203,36 @@ public void When_Navigated_CommandBarShouldKeepColor_Native_Frame()
...
+			_app.FastTap("NavigateInitialButton");
+
+			_app.WaitForElement("NavigateToPage2Button");
+			_app.FastTap("NavigateToPage2Button");
+
+			_app.WaitForElement("BackButtonTitleButton");
+
+			_app.FastTap("BackButtonTitleButton");
+
+			_app.Wait(TimeSpan.FromMilliseconds(800));
```
jeromelaban, line 228: You should be able to use `App.WaitForDependencyPropertyValue()` to avoid arbitrarily waiting some time (it may fail if the underlying machine is slow).

ajpinedam: `App.WaitForDependencyPropertyValue()`This one is new for me.. Will update. Thanks :)

ajpinedam: Updated!

dr1rrb (approved):

jeromelaban:
@mergifyio backport release/stable/3.1

Issue #4878, opened 2021-01-11
BitmapImage.SetSource does not clone its input


TopperDEL:
## Current behavior
When I update my app from Uno.UI 3.3.0 to Uno.UI 3.4.0 (and the pre-release 3.5.0*) my Image-Bindings in a list view do not show any image anymore. I did not change anything else.
The interesting part is: I have an issue with the SymbolIcons as mentioned [here](https://github.com/unoplatform/uno/issues/4831) and that gets resolved when updating to a newer Uno-UI.version than 3.3.0.

So either I have Symbols or I see me images - but not both...

The images are bound within a DataTemplate, as is the Symbol. I'll may try to use a "Binding" instead of "x:Bind" - but it works with 3.3.0 - so this is a regression.

## Expected behavior
The Images are shown again.

## How to reproduce it (as minimally and precisely as possible)
That's the tricky part - I'm not sure and the app is quite complex. [This](https://github.com/TopperDEL/NutzMich) is the app and it's currently easier for me to just link it. Clone the latest version and start it - after the welcome screen you should see a list of items - they have no images but the button above show a Refresh-Symbol. Now go back to Uno.UI on the iOs-Head and start again - the list now shows images but the Refresh-Icon changed to a smiley.

I searched around if there are any changes between bot Uno-Version that might point to this error, but I could not find any.

## Workaround
None yet

## Environment

Nuget Package:

- [X] Uno.UI 3.3.0 / 3.4.0
- [ ] Uno.WinUI / Uno.WinUI.WebAssembly / Uno.WinUI.Skia
- [ ] Uno.SourceGenerationTasks
- [ ] Uno.UI.RemoteControl / Uno.WinUI.RemoteControl
- [ ] Other: 

Nuget Package Version(s):

Affected platform(s):

- [X] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] WebAssembly renderers for Xamarin.Forms
- [ ] macOS
- [ ] Skia
  - [ ] WPF
  - [ ] GTK (Linux)
  - [ ] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

IDE:

- [ ] Visual Studio 2017 (version: )
- [X] Visual Studio 2019 (version: )
- [ ] Visual Studio for Mac (version: )
- [ ] Rider Windows (version: )
- [ ] Rider macOS (version: )
- [ ] Visual Studio Code (version: )

Relevant plugins:

- [ ] Resharper (version: )

[...]

PR #5027, opened 2021-01-25
 fix(imagesource): Fix the SetSource[Async] must clone the provided stream before returning

dr1rrb:
fixes https://github.com/unoplatform/uno/issues/4878

## Bugfix
Make sure to always clone the source stream before completing the `BitmapSource.SetSource[Async]` method.

It might be clone asynchronously, especially when using the `SetSourceAsync` (clone was dispatch on next dispatcher loop).

## What is the new behavior?
🙃

## PR Checklist

### DIFF
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Media_Imaging/Given_BitmapSource.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Media_Imaging/Given_BitmapSource.cs
@@ -57,6 +57,7 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Media_Imaging
 			Assert.IsTrue(success);
 		}
 
+#if !WINDOWS_UWP
 		[TestMethod]
 		[RunsOnUIThread]
 		public async Task When_SetSource_Stream_Then_StreamClonedSynchronously()
@@ -96,6 +97,7 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Media_Imaging
 
 			Assert.IsTrue(success);
 		}
+#endif
 
 		private class Given_BitmapSource_Exception : Exception
 		{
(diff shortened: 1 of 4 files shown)


carldebilly (commented):
What is `imagfesource` ?

jeromelaban (approved):

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Media_Imaging/Given_BitmapSource.cs
```diff
@@ -0,0 +1,125 @@
...
+
+			var success = false;
+			try
+			{
+				sut.SetSourceAsync(raStream);
+			}
+			catch (Given_BitmapSource_Exception ex) when (ex.Caller is nameof(Given_BitmapSource_Stream.Read))
+			{
+				success = true;
+			}
```
carldebilly, line 55: Very strange pattern to rely on exceptions to validate a method is called.  There's features for that in mocking frameworks...

src/Uno.UI/UI/Xaml/Media/Imaging/BitmapSource.cs
```diff
@@ -67,56 +67,63 @@ protected BitmapSource(string sourceString) : base(sourceString)
...
-#if __NETSTD__
-				_stream = streamSource.CloneStream();
+			// The source has to be cloned before leaving the "SetSource[Async]".
+			var clonedStreamSource = streamSource.CloneStream();
 
-				var tcs = new TaskCompletionSource<object>();
+#if __NETSTD__
+			_stream = clonedStreamSource;
+#else
+			Stream = clonedStreamSource.AsStream();
```
ajpinedam, line 107: Was this intended? the property instead of the field? `Stream` / `_stream`

ajpinedam (approved):

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Media_Imaging/Given_BitmapSource.cs
dr1rrb: Yes ... except it's a runtime test, there isn't any setup and I'm pretty sure that none exists for iOS which does not allow `Reflection.Emit` ;)

src/Uno.UI/UI/Xaml/Media/Imaging/BitmapSource.cs
dr1rrb: The `Stream` property is in the `ImageSource` base class and is present only on iOS and Android. On .net standard we have moved it back into a private field of the `BitmapSource` as it's this class which has the ability to load an image from a `Stream` (or actually an `IRandomAccessStream`)

PR #5056, opened 2021-01-27
fix(Bindings): Avoid propagating DataContext if you are not the child's parent

### LINKED ISSUE #4922: [Android] ElementName binding within a Flyout breaks other bindings in that Flyout
## Current behavior

On Android, using an ElementName binding in a flyout breaks other bindings.

## Expected behavior

All bindings work, regardless of platform or visual hierarchy.

Page demonstrating issue:
```xaml
<Page
    x:Class="FlyoutDataContextIssue.MainPage"
    Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">
    <Button DataContext="data context value">
        <TextBlock>
            <Run Text="DataContext: "/><Run Text="{Binding}"/>
            <LineBreak/>
            <Run Text="Length: "/><Run Text="{Binding Length}"/>
        </TextBlock>
        <Button.Flyout>
            <Flyout x:Name="FlyoutElement">
                <!--<Border>-->
                <StackPanel>
                    <TextBlock>
                        <Run Text="DataContext: "/><Run Text="{Binding}"/>
                        <LineBreak/>
                        <Run Text="Length: "/><Run Text="{Binding Length}"/>
                    </TextBlock>
                    <Button
                        Content="Button"
                        CommandParameter="{Binding ElementName=FlyoutElement}"/>
                </StackPanel>
                <!--</Border>-->
            </Flyout>
        </Button.Flyout>
    </Button>
</Page>
```
The top-level Button and the Flyout contain the same TextBlock content, but only the top-level bindings show content on Android due to the ElementName binding on the Button within the Flyout:
Screenshot_20210116-120258

## Workaround

Wrapping the Flyout content in a Border fixes the bindings (similar to #4626).

## Environment

Nuget Package:
- [ ] Other: 

Nuget Package Version(s):

- 3.4.0

Affected platform(s):

- [ ] iOS
- [x] Android
- [ ] WebAssembly
- [ ] macOS
- [ ] Skia
  - [ ] WPF
  - [ ] GTK (Linux)
  - [ ] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

IDE:

- [x] Visual Studio 2019 (version: 16.8.4)
- [ ] Rider macOS (version: )

Relevant plugins:

- [ ] Resharper (version: )

kazo0:
closes #4922

## PR Type

- Bugfix

Binding to Flyout from within the Flyout itself causes the wrong DataContext to be set on the Flyout

Example: 

- Button A contains a Flyout with Button B inside of it
- Button B has a binding to the Flyout itself
- We should not propagate Button B's DataContext to the Flyout since its real parent is actually Button A 

## What is the new behavior?

Binding value is not propagated to child unless you are the actual Parent of that child

## PR Checklist

### DIFF
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_Flyout.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_Flyout.cs
@@ -4,6 +4,7 @@ using System.Text;
 using System.Threading.Tasks;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Private.Infrastructure;
+using Uno.Extensions;
 using Uno.UI.Extensions;
 using Uno.UI.RuntimeTests.Extensions;
 using Uno.UI.RuntimeTests.Helpers;
@@ -11,6 +12,7 @@ using Windows.UI;
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls;
 using Windows.UI.Xaml.Controls.Primitives;
+using Windows.UI.Xaml.Data;
 using Windows.UI.Xaml.Media;
 
 namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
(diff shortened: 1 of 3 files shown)


carldebilly (approved):
Nice work...  + unit tests, that's cool.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_Flyout.cs
```diff
@@ -290,6 +318,31 @@ private static void VerifyRelativeContentPosition(HorizontalPosition horizontalP
 			return (flyout, content);
 		}
 
+		private (Flyout Flyout, FrameworkElement Content) CreateFlyoutWithBinding()
+		{
+
+			var flyout = new Flyout
+			{
+				//Content = content,
```
davidjohnoliver, line 326:
```suggestion
```

davidjohnoliver (approved):
My feeling is that `When_DataContext_Inherited_And_Child_Removed()` is wrong, but we'll see what @dr1rrb says. It looks like you should be able to add it as a RuntimeTest and see what it does on UWP.

src/Uno.UI/UI/Xaml/DependencyObjectStore.Binder.cs
```diff
@@ -105,10 +105,21 @@ void SetInherited(IDependencyObjectStoreProvider provider)
...
 				var child = _childrenBindable[i];
+				var parent = child?.GetParent();
+
+				//Do not propagate value if you are not this child's parent
+				//Covers case where a child may hold a binding to a view higher up the tree
+				//Example: Button A contains a Flyout with Button B inside of it
+				//	Button B has a binding to the Flyout itself
+				//	We should not propagate Button B's DataContext to the Flyout
+				//	since its real parent is actually Button A 
+				if (parent != null && parent != ActualInstance)
```
dr1rrb, line 119: Are we sure that we won't leak the `DataContext` when we remove a child from it's  parent and we pout it in a pool?

Even if I dislike it a lot, the way uno is avoiding datacontext leaks is by explicitly pushing a `null` as `DataContext` when an element is unloaded. If here we stop that propagation when the parent is not set, I would make sure that we still are allowing that `null` propagation (i.e. the parent is removed after the `null` has been pushed).

davidjohnoliver: If parent is not set, it'll be `null`, which is being checked for. It's only if parent is non-null **and** something else that we skip propagating the DataContext.

PR #5118, opened 2021-02-03
fix(navigationview): Prevent crash when selecting item too early

dr1rrb:
fixes https://github.com/unoplatform/uno/issues/5069

## Bugfix
`NavigationView` is crashing if `SelectedItem` is set before first layouting pass.

When an item is realized by the `ItemsRepeater` the navigation forces a `UpdateLayout` which causes the `ItemsRepeater` to re-measure itself ... which throws an exception.

## What is the new behavior?
We are now considering the "native" layouting passes in the `UpdateLayout` and ignore the requests if layouting is already in progress.

## PR Checklist

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/Layouter/Layouter.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Layouter/Layouter.cs
@@ -54,11 +54,14 @@ namespace Windows.UI.Xaml.Controls
 
 		internal Size _unclippedDesiredSize;
 
+		private UIElement _elementAsUIElement;
+
 		public IFrameworkElement Panel { get; }
 
 		protected Layouter(IFrameworkElement element)
 		{
 			Panel = element;
+			_elementAsUIElement = element as UIElement;
 
 			var log = this.Log();
 			if (log.IsEnabled(LogLevel.Debug))
@@ -90,10 +93,14 @@ namespace Windows.UI.Xaml.Controls
 				return default;
 			}
 
-			using (traceActivity)
+			try
 			{
-				var (minSize, maxSize) = Panel.GetMinMax();
+				if (_elementAsUIElement?.IsVisualTreeRoot ?? false)
+				{
+					UIElement.IsLayoutingVisualTreeRoot = true;
+				}
 
+				var (minSize, maxSize) = Panel.GetMinMax();
 				var marginSize = Panel.GetMarginSize();
 
 				// NaN values are accepted as input here, particularly when coming from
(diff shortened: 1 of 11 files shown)


src/Uno.UI/UI/Xaml/Controls/Layouter/Layouter.cs
```diff
@@ -54,11 +54,14 @@ internal abstract partial class Layouter : ILayouter
 
 		internal Size _unclippedDesiredSize;
 
+		private UIElement _elementAsUIElement;
```
davidjohnoliver, line 57: property here instead of field?

src/Uno.UI/UI/Xaml/Controls/Layouter/Layouter.cs
```diff
@@ -90,10 +93,14 @@ public Size Measure(Size availableSize)
 				return default;
```
davidjohnoliver, line 93: Not a bug you introduced, but I notice we don't dispose the trace activity if the element is collapsed

davidjohnoliver (approved):

davidjohnoliver:
This will probably have an impact on TabView (hopefully positive)... it was entering VirtualizingPanelLayout.MeasureOverride() a _lot_ because UpdateLayout() gets called every time a TabViewItem is added

src/Uno.UI/UI/Xaml/Controls/Layouter/Layouter.cs
carldebilly: We should change the pattern to use the `using var` syntax.

src/Uno.UI/UI/Xaml/Controls/Layouter/Layouter.cs
dr1rrb: It's `private` ... why use a property?

ajpinedam (approved):

src/Uno.UI/UI/Xaml/Controls/Layouter/Layouter.cs
davidjohnoliver: For legibility, since it will always point to the same instance as Panel. Why use a field?

jeromelaban: For performance, in that case :)

src/Uno.UI/UI/Xaml/Controls/Layouter/Layouter.cs
dr1rrb: Fixed

PR #5183, opened 2021-02-08
fix(FrameworkElement): [macOS] Ensure we set parent to a DependencyObject

kazo0:
closes https://github.com/unoplatform/Uno.Gallery/issues/129
closes https://github.com/unoplatform/Uno.Gallery/issues/130

## PR Type

- Bugfix

Bindings are not propagated properly when using a ScrollViewer since the FrameworkElement Implementation of `ViewDidMoveToWindow` tries to set the 

## What is the new behavior?

Bindings are properly propagated 

## PR Checklist

### DIFF
--- a/src/Uno.UI/UI/Xaml/IFrameworkElementImplementation.macOS.tt
+++ b/src/Uno.UI/UI/Xaml/IFrameworkElementImplementation.macOS.tt
@@ -25,9 +25,9 @@ using Windows.UI.Xaml.Automation.Peers;
 using Windows.UI.Xaml.Automation.Provider;
 using Windows.UI.Xaml.Automation;
 using Uno.UI.Xaml;
-
-<# 
-	foreach(var mixin in _mixins) { 
+using Uno.UI.Extensions;
+<#
+	foreach(var mixin in _mixins) {
 #>
 
 namespace <#= mixin.NamespaceName #>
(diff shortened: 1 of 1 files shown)


src/Uno.UI/UI/Xaml/IFrameworkElementImplementation.macOS.tt
```diff
@@ -69,17 +70,18 @@ namespace <#= mixin.NamespaceName #>
...
 					{
-						if(_superViewRef?.GetTarget() == null && superView != null)
+						var parent = GetDependencyObjectAncestor(Superview) as NSView;
+						parent ??= Superview;
+						if(_superViewRef?.GetTarget() == null && parent != null)
 						{
-							_superViewRef = new WeakReference<NSView>(superView);
-							SyncBinder(superView, newWindow);
-							((IDependencyObjectStoreProvider)this).Store.Parent = superView;
+							_superViewRef = new WeakReference<NSView>(parent);
```
kazo0, line 82: Still not sure about what should be set here, the "real" Superview or the ancestor that we may have found that is a `DependencyObject`?

davidjohnoliver: I'm not sure either - it only seems to be checked here, and TBH I'm not clear on what the intent of the check is.

Any thoughts @jeromelaban ?

kazo0: But even for the `_superViewRef`? We shouldn't set that to the actual Superview regardless of it being a DependencyObject?

davidjohnoliver: @kazo0 I would leave it as you have it, setting it to `parent` (the actual dependency object), but I'm still not sure what the check is even for.

davidjohnoliver (approved):

Issue #5262, opened 2021-02-17
Creating a `ContentDialog` before setting the `Windows.Current.Content` crashes


xperiandri:
Crashes with `IndexOutOfRangeException` because no Window is located here https://github.com/unoplatform/uno/blob/64f2d0a10dd73f48a88fe51212deb131241834aa/src/Uno.UI/UI/Xaml/Controls/Popup/Popup.iOS.cs#L26

## Current behavior

`IndexOutOfRangeException`

## Expected behavior

Creation delayed until UI is ready

## How to reproduce it (as minimally and precisely as possible)

Create ContentDialog before setting `Window.Current.Content`

## Workaround

Create ContentDialog after setting `Window.Current.Content`

## Environment

Nuget Package:
- [X] Uno.UI / Uno.UI.WebAssembly / Uno.UI.Skia
- [ ] Uno.WinUI / Uno.WinUI.WebAssembly / Uno.WinUI.Skia
- [ ] Uno.SourceGenerationTasks
- [ ] Uno.UI.RemoteControl / Uno.WinUI.RemoteControl
- [ ] Other: 

Nuget Package Version(s):
3.5.1

Affected platform(s):

- [X] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] WebAssembly renderers for Xamarin.Forms
- [ ] macOS
- [ ] Skia
  - [ ] WPF
  - [ ] GTK (Linux)
  - [ ] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

IDE:

- [ ] Visual Studio 2017 (version: )
- [X] Visual Studio 2019 (version: )
- [ ] Visual Studio for Mac (version: )
- [ ] Rider Windows (version: )
- [ ] Rider macOS (version: )
- [ ] Visual Studio Code (version: )

MartinZikmund:
No longer relevant for Skia renderer targets

PR #5264, opened 2021-02-17
[refactor] Added descriptive exception for the case when a popup is created before any app Windows is created

xperiandri:
Related to https://github.com/unoplatform/uno/issues/5262
Probably it is better to subscribe to some event that will occur when an app will have at least one Window

## PR Type

- Refactoring (no functional changes, no API changes)

App crashes with `IndexOutOfRangeException`

## What is the new behavior?

App crashes with `InvalidOperationException`

## PR Checklist

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/Popup/Popup.iOS.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Popup/Popup.iOS.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Text;
 using CoreGraphics;
 using Uno.Extensions;
@@ -23,7 +24,9 @@ namespace Windows.UI.Xaml.Controls
 			{
 				if (_mainWindow == null)
 				{
-					_mainWindow = UIApplication.SharedApplication.KeyWindow ?? UIApplication.SharedApplication.Windows[0];
+					_mainWindow = UIApplication.SharedApplication.KeyWindow
+							   ?? UIApplication.SharedApplication.Windows.FirstOrDefault()
+							   ?? throw new InvalidOperationException("Do not create popup until any app Window is created");
 				}
 
 				return _mainWindow;


src/Uno.UI/UI/Xaml/Controls/Popup/Popup.iOS.cs
```diff
@@ -23,7 +24,9 @@ public UIView MainWindow
 			{
 				if (_mainWindow == null)
 				{
-					_mainWindow = UIApplication.SharedApplication.KeyWindow ?? UIApplication.SharedApplication.Windows[0];
+					_mainWindow = UIApplication.SharedApplication.KeyWindow
+							   ?? UIApplication.SharedApplication.Windows.FirstOrDefault()
+							   ?? throw new InvalidOperationException("Do not create popup until any app Window is created");
```
davidjohnoliver, line 29: It's generally better not to throw an exception from a getter (particularly this one that is for some reason public). Can you put the exception one step up in the call stack?

davidjohnoliver (commented):
Thanks for the contribution!

You'll need to make sure all your commits are in Conventional Commits format: https://platform.uno/docs/articles/uno-development/git-conventional-commits.html

src/Uno.UI/UI/Xaml/Controls/Popup/Popup.iOS.cs
xperiandri: Maybe you know a better way? Maybe we can subscribe to some event if UI has not loaded at the time of a call?

jeromelaban: Can you provide the stacktrace you're getting when this exception happens ?

xperiandri: Create `ContentDialog` instance in App constructor. As I used `Host` I built DI in the constructor and put dialogs as dependencies. That caused them to be created before any window set.

MartinZikmund: I would say this fix is fine, as it will provide a better exception for the user, but should not happen in most cases (when the API is used after initialization). I think we use a similar approach for APIs which need access to KeyWindow

MartinZikmund (approved):

jeromelaban:
Thanks for your contribution. This code has significantly changed since this PR was introduced and will have to be revisited soon. Let's close for now.

PR #5689, opened 2021-04-06
fix(FileSavePicker.wasm): FileSavePicker dialog now shows all file type extensions.

### LINKED ISSUE #5685: FileSavePicker FileTypeChoices incorrect
## Current behavior

I have this code:

```
            var savePicker = new Windows.Storage.Pickers.FileSavePicker();
            savePicker.SuggestedStartLocation = Windows.Storage.Pickers.PickerLocationId.DocumentsLibrary;
            savePicker.FileTypeChoices.Add("C/C++ source files", new List<string>() { ".cpp", ".c", ".cc", ".cxx", ".c++", ".cp" });
            savePicker.SuggestedFileName = "Hacked";
            Windows.Storage.StorageFile file = await savePicker.PickSaveFileAsync();
```

However, instead of showing these file types, they show `.cp`, `.com`, `.exe`, `.bin`.

image

## Expected behavior

It shows the right file types, like in the my UWP-only project:

image

- Create a Uno project.
- Open a `FileSavePicker`, like in the code above.
- Build the WASM project.

## Workaround
Found none.

## Environment

Nuget Package:

- [ ] Other: 

Nuget Package Version(s):

Affected platform(s):

- [ ] iOS
- [ ] Android
- [x] WebAssembly
- [ ] macOS
- [ ] Skia
  - [ ] WPF
  - [ ] GTK (Linux)
  - [ ] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

IDE:

- [x] Visual Studio 2019 (version: 16.9.3)
- [ ] Rider macOS (version: )

Relevant plugins:

- [ ] Resharper (version: )

`FileOpenPicker` seems to work well. Only FileSavePicker fails.

trungnt2910:
GitHub Issue (If applicable): fixes #5685

## PR Type
- Bugfix

Some file type extensions, added by `FileTypeChoices.Add` are missing.

Code: 
```
picker.FileTypeChoices.Add("C/C++ source files", new List<string>() { ".cpp", ".c", ".cc", ".cxx", ".c++", ".cp" });
```

Result:
Only shows `.cp` extension, and also some other irrelevant `.com`, `.exe`, and `.bin`.
image

## What is the new behavior?

Shows all included extensions, as well as the irrelevant extensions `.com`, `.exe`, and `.bin`.
image

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UWP/Storage/Pickers/FileSavePicker.wasm.cs
+++ b/src/Uno.UWP/Storage/Pickers/FileSavePicker.wasm.cs
@@ -69,16 +69,9 @@ namespace Windows.Storage.Pickers
 				var acceptType = new NativeFilePickerAcceptType();
 				acceptType.Description = choice.Key;
 
-				var acceptItems = new List<NativeFilePickerAcceptTypeItem>();
-				foreach (var extension in choice.Value)
-				{
-					var acceptItem = new NativeFilePickerAcceptTypeItem()
-					{
-						MimeType = MimeTypeService.GetFromExtension(extension),
-						Extensions = new[] { extension }
-					};
-					acceptItems.Add(acceptItem);
-				}
+				var acceptItems = choice.Value
+					.GroupBy(ext => MimeTypeService.GetFromExtension(ext))
+					.Select(group => new NativeFilePickerAcceptTypeItem() { MimeType = group.Key, Extensions = group.ToArray() });
 
 				acceptType.Accept = acceptItems.ToArray();
 				acceptTypes.Add(acceptType);
--- a/src/Uno.UWP/Storage/Pickers/FileSavePicker.wasm.cs
+++ b/src/Uno.UWP/Storage/Pickers/FileSavePicker.wasm.cs
@@ -69,11 +69,9 @@ namespace Windows.Storage.Pickers
 				var acceptType = new NativeFilePickerAcceptType();
 				acceptType.Description = choice.Key;
 
-				var acceptItems = choice.Value
-					.GroupBy(ext => MimeTypeService.GetFromExtension(ext))
-					.Select(group => new NativeFilePickerAcceptTypeItem() { MimeType = group.Key, Extensions = group.ToArray() });
+				var acceptItem = new NativeFilePickerAcceptTypeItem() { MimeType = "*/*", Extensions = choice.Value.ToArray() };
 
-				acceptType.Accept = acceptItems.ToArray();
+				acceptType.Accept = new NativeFilePickerAcceptTypeItem[] { acceptItem };
 				acceptTypes.Add(acceptType);
 			}
 


trungnt2910:
I can remove the `.exe`, `.bin`, and `.com` files, by using every file's mime type as `*/*`, but I don't know if it brings in any security problem:
```C#
var acceptItem = new NativeFilePickerAcceptTypeItem() { MimeType = "*/*", Extensions = choice.Value.ToArray() };

acceptType.Accept = new NativeFilePickerAcceptTypeItem[] { acceptItem };
```

Result:

image

MartinZikmund:
@trungnt2910 I would not use `*/*` to avoid this issue, as that would mean the "wildcard" option would always be there. Instead it might be worth considering removing the `application/octet-stream` fallback from `MimeTypeService` and ignore extensions without appropriate mapping to force the user to provide a custom MIME time manually.

trungnt2910:
> force the user to provide a custom MIME time manually

You mean forcing the caller to provide MIME types whenever they use the API? Does this means breaking cross platform behavior?

Or do you mean making the end user of the app choose the types?

MartinZikmund:
@trungnt2910 I meant instead of mapping all "unknown" types to the `octet-steam` in `MimeTypeService`, these types would be ignored. So in case the developer wanted: ".jpg, .xyz, .gif". only ".jpg, .gif" would show up in the dialog. To get the ".xyz", they would have to provide custom mapping to Uno via `WinRTFeatureConfiguration` (which could even be `*/*` as a wildcard)

trungnt2910:
> @trungnt2910 I meant instead of mapping all "unknown" types to the `octet-steam` in `MimeTypeService`, these types would be ignored. So in case the developer wanted: ".jpg, .xyz, .gif". only ".jpg, .gif" would show up in the dialog. To get the ".xyz", they would have to provide custom mapping to Uno via `WinRTFeatureConfiguration` (which could even be `*/*` as a wildcard)

So this **does** force developers to write platform-specific code. It might make quite many people wonder why `.xyz` is shown on their UWP test app but not on their website.

MartinZikmund:
@trungnt2910 The current behavior is the same though, as you get the `.com`, `.exe`, etc. instead of the custom `.xyz` extension. That might be even weirder :-D . The `WinRTFeatureConfiguration` approach is documented as part of the file picker docs, so I wouldn't consider it a big problem. Unfortunately there is no way to make a fully platform-independent solution without these edge cases which require the developer's intervention... The `*/*` as a fallback might be a good workaround though, as that would be just a full wildcard, better than the `.com`, etc..

trungnt2910:
> @trungnt2910 The current behavior is the same though, as you get the `.com`, `.exe`, etc. instead of the custom `.xyz` extension. That might be even weirder :-D . The `WinRTFeatureConfiguration` approach is documented as part of the file picker docs, so I wouldn't consider it a big problem. Unfortunately there is no way to make a fully platform-independent solution without these edge cases which require the developer's intervention... The `*/*` as a fallback might be a good workaround though, as that would be just a full wildcard, better than the `.com`, etc..

Another reason why I thought about `*/*` might be a good solution:

Code:
```C#
picker.FileTypeChoices.Add("HTML file", new List<string>() { ".html" });
```

Result:
image

I only want the user to save as `.html`, but the browser added a ton of other extensions, because I specified `text/html`.
In this case, I had to register the type as `application/my-custom-html`, which somehow breaks the purpose of MIME types (No better than `*.*`, I guess).

trungnt2910:
image

I've already included an image in a comment above.

MartinZikmund:
@trungnt2910 Ah, missed that. Ok, let's use the `*/*` workaround here 👍

MartinZikmund (approved):

carldebilly (approved):

Issue #3093, opened 2020-05-02
Support Focus Visuals


MartinZikmund:
## What would you like to be added:

Currently the `UseSystemFocusVisuals` property does not have any effect and for keyboard navigation, the focused visuals are not visible.

## Why is this needed:

Improved keyboard and gamepad navigation support.

## Notes

- Border applies the same CornerRadius as the element
- Border scales and transforms with the element (same RenderTransform)
- Focus visual animates with the element
- Adheres to Viewbox
- Relevant properties
   - `FocusVisualPrimaryBrush` - outer border
   - `FocusVisualPrimaryThickness` - outer border thickness
   - `FocusVisualSecondaryBrush` - inner border
   - `FocusVisualSecondaryThickness` - inner border thickness
   - `FocusVisualMargin` - margin, negative value = visual is "bigger", positive makes it smaller (inside the element)
   - `UseSystemFocusVisuals` - allows turning the default visuals off
- Apply only to `Keyboard` focus, not to `Pointer`
   - In case the control template overrides `FocusState` group, the `Focused` state is applied on keyboard, `PointerFocused` on pointer interaction
   - `UseSystemFocusVisuals`

## To do

- [ ] Move with scroll view
- [ ] Move with animations
- [x] React to `UseSystemFocusVisuals` changes when already focused

## Resources

https://docs.microsoft.com/en-us/windows/uwp/design/input/guidelines-for-visualfeedback

## For which Platform:

- [x] iOS
- [x] Android
- [x] WebAssembly
- [ ] WebAssembly renderers for Xamarin.Forms
- [x] macOS
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

## Anything else we need to know?

Related: https://github.com/unoplatform/uno/issues/90

MartinZikmund:
Example of focus visual scaled with `ViewBox`:

![image](https://user-images.githubusercontent.com/1075116/114384880-948f9b00-9b8f-11eb-9dc1-503dec328ae4.png)

MartinZikmund:
Initial support is merged, additional features will be added as separate issues

PR #5746, opened 2021-04-14
Initial Focus Visuals support

MartinZikmund:
GitHub Issue (If applicable): #3093 

## PR Type

- Feature

Keyboard/Gamepad focus is not recognizable

## What is the new behavior?

Initial support for Focus visuals.

## PR Checklist

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/FocusVisual/SystemFocusVisual.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/FocusVisual/SystemFocusVisual.cs
@@ -1,11 +1,12 @@
-﻿using System;
+﻿#nullable enable
+
 using Uno.Disposables;
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls;
 
 namespace Uno.UI.Xaml.Controls
 {
-	public partial class SystemFocusVisual : Control
+	internal partial class SystemFocusVisual : Control
 	{
 		private SerialDisposable _focusedElementSubscriptions = new SerialDisposable();
 
@@ -13,8 +14,6 @@ namespace Uno.UI.Xaml.Controls
 		{
 			DefaultStyleKey = typeof(SystemFocusVisual);
 		}
-
-		protected override void OnApplyTemplate() => base.OnApplyTemplate();
 
 		public FrameworkElement FocusedElement
 		{
@@ -37,6 +36,7 @@ namespace Uno.UI.Xaml.Controls
 
 			if (args.NewValue is FrameworkElement element)
 			{
+				element.EnsureFocusVisualBrushDefaults();
 				element.SizeChanged += focusVisual.FocusedElementSizeChanged;
 				element.LayoutUpdated += focusVisual.FocusedElemenLayoutUpdated;
 
--- a/src/Uno.UI/UI/Xaml/Controls/FocusVisual/SystemFocusVisual.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/FocusVisual/SystemFocusVisual.cs
@@ -1,0 +1,24 @@
+﻿using Windows.UI.Xaml;
+using Windows.UI.Xaml.Controls;
+
+namespace Uno.UI.Xaml.Controls
+{
+	public partial class SystemFocusVisual : Control
+	{
+		public SystemFocusVisual()
+		{
+			DefaultStyleKey = typeof(SystemFocusVisual);
+		}
+
+		protected override void OnApplyTemplate() => base.OnApplyTemplate();
+
+		public FrameworkElement FocusedElement
+		{
+			get => (FrameworkElement)GetValue(FocusedElementProperty);
+			set => SetValue(FocusedElementProperty, value);
+		}
+
+		public static readonly DependencyProperty FocusedElementProperty =
+			DependencyProperty.Register(nameof(FocusedElement), typeof(FrameworkElement), typeof(SystemFocusVisual), new PropertyMetadata(default));
+	}
+}
--- a/src/Uno.UI/UI/Xaml/Controls/FocusVisual/SystemFocusVisual.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/FocusVisual/SystemFocusVisual.cs
@@ -1,10 +1,14 @@
-﻿using Windows.UI.Xaml;
+﻿using System;
+using Uno.Disposables;
+using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls;
 
 namespace Uno.UI.Xaml.Controls
 {
 	public partial class SystemFocusVisual : Control
 	{
+		private SerialDisposable _focusedElementSubscriptions = new SerialDisposable();
+
 		public SystemFocusVisual()
 		{
 			DefaultStyleKey = typeof(SystemFocusVisual);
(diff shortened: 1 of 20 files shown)


src/Uno.UI/UI/Xaml/Controls/FocusVisual/SystemFocusVisual.cs
```diff
@@ -0,0 +1,72 @@
...
+			focusVisual._focusedElementSubscriptions.Disposable = null;
+
+			if (args.NewValue is FrameworkElement element)
+			{
+				element.SizeChanged += focusVisual.FocusedElementSizeChanged;
+				element.LayoutUpdated += focusVisual.FocusedElemenLayoutUpdated;
+
+				focusVisual.SetLayoutProperties();
+
+				focusVisual._focusedElementSubscriptions.Disposable = Disposable.Create(() =>
```
jeromelaban, line 45: Make sure to set the disposable to null first, otherwise subscriptions may not get removed properly.

src/Uno.UI/UI/Xaml/FrameworkElement.cs
```diff
@@ -438,6 +441,77 @@ private void OnRequestedThemeChanged(ElementTheme oldValue, ElementTheme newValu
 			Application.Current?.ActualElementTheme ?? ElementTheme.Default
 			: ElementTheme.Default;
 
+		public Thickness FocusVisualSecondaryThickness
```
jeromelaban, line 444: Those may be used quite a bit, it'd be best if they were using generated DependencyProperty

jeromelaban (commented):
Very nice! Is there a way to automate the testing?

src/Uno.UI/UI/Xaml/Controls/FocusVisual/SystemFocusVisual.cs
MartinZikmund: I'm setting it null a bit above on line 36 👍

src/Uno.UI/UI/Xaml/FrameworkElement.cs
MartinZikmund: Will do 👍

MartinZikmund:
> Very nice! Is there a way to automate the testing?

Possibly yes. At minimum the screenshot testing will work, but maybe with higher thickness I could verify the color is there. Will investigate

carldebilly (approved):

ajpinedam (approved):

src/Uno.UI/UI/Xaml/Controls/FocusVisual/SystemFocusVisual.cs
```diff
@@ -0,0 +1,72 @@
+﻿using System;
+using Uno.Disposables;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Controls;
+
+namespace Uno.UI.Xaml.Controls
+{
+	public partial class SystemFocusVisual : Control
+	{
+		private SerialDisposable _focusedElementSubscriptions = new SerialDisposable();
```
davidjohnoliver, line 10: 🤖

src/Uno.UI/UI/Xaml/Controls/FocusVisual/SystemFocusVisual.cs
```diff
@@ -0,0 +1,72 @@
+﻿using System;
+using Uno.Disposables;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Controls;
+
+namespace Uno.UI.Xaml.Controls
+{
+	public partial class SystemFocusVisual : Control
```
davidjohnoliver, line 8: Can this be internal?

src/Uno.UI/UI/Xaml/Controls/FocusVisual/SystemFocusVisual.cs
```diff
@@ -0,0 +1,72 @@
...
+	{
+		private SerialDisposable _focusedElementSubscriptions = new SerialDisposable();
+
+		public SystemFocusVisual()
+		{
+			DefaultStyleKey = typeof(SystemFocusVisual);
+		}
+
+		protected override void OnApplyTemplate() => base.OnApplyTemplate();
```
davidjohnoliver, line 17: ?

src/Uno.UI/UI/Xaml/Controls/FocusVisual/SystemFocusVisual.cs
```diff
@@ -0,0 +1,72 @@
+﻿using System;
```
davidjohnoliver, line 1:
```suggestion
#nullable enable

﻿using System;
```

src/Uno.UI/UI/Xaml/FrameworkElement.cs
```diff
@@ -438,6 +441,63 @@ private void OnRequestedThemeChanged(ElementTheme oldValue, ElementTheme newValu
...
+		{
+			get => GetFocusVisualMarginValue();
+			set => SetFocusVisualMarginValue(value);
+		}
+
+		private static Thickness GetFocusVisualMarginDefaultValue() => Thickness.Empty;
+
+		private void InitializeFocusVisualProperties()
+		{
+			ResourceResolver.ApplyResource(this, FocusVisualPrimaryBrushProperty, new SpecializedResourceDictionary.ResourceKey("SystemControlFocusVisualPrimaryBrush"), true, null, DependencyPropertyValuePrecedences.DefaultValue);
```
davidjohnoliver, line 497: Hmm... this means that every single FrameworkElement in the visual tree is going to be listening for updates to a theme-bound value. There may be perf implications. I wonder if it would be possible to only subscribe for updates if we know that the control is actually being used as a focus visual?

cc @jeromelaban

src/Uno.UI/UI/Xaml/Controls/FocusVisual/SystemFocusVisual.cs
jeromelaban: It should indeed.

src/Uno.UI/UI/Xaml/FrameworkElement.cs
jeromelaban: There will be performance implication for sure. Maybe it should be initialized lazily and only if it can be focused ?

MartinZikmund: In that case I would need to be able to "plug-in" that lazy initialization when someone calls `element.GetValue(FocusVisualPrimaryBrushProperty)` - not sure if we have such capability currently 🤔 .

MartinZikmund: @davidjohnoliver @jeromelaban Maybe could use an approach similar to this and just resolve the static resource instead of `ThemeResource`? Static lookup for two resources per element could be quite cheap
https://github.com/unoplatform/uno/blob/56d3ba98ee6cab055dc8d7c4941ce4494e20e946/src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.cs#L931-L935?

davidjohnoliver: @MartinZikmund Sounds like a good way to go - static resource look-up isn't a problem, only the theme-update binding is a potential concern.

davidjohnoliver: Though the more I think about it, I wonder if we still want to avoid the static look-up for every single element - it still has to walk upward and check all available resources.

MartinZikmund: Yes, it is possible to override the value be declaring a resource with the same name unfortunately 🤔

MartinZikmund: @davidjohnoliver  @jeromelaban  Adjusted code to use static retrieval

MartinZikmund: If this would still be too performance intensive, I could make the default to return some hardcoded color (like transparent) and in the property getter initialize the DP lazily. The only disadvantage would then be that `GetValue(FocusVisualPrimaryBrushProperty)` will return wrong value until `FocusVisualPrimaryBrush` is accessed. To make sure it works for the focused element, the `SystemFocusVisual` control could force this initialization when the data-bound element changes

davidjohnoliver: It's definitely better, but I wonder if even the static walk up the tree is too costly if it's done for every single element.

Actually, since you're doing this from the constructor, it's never going to find elements in the visual tree (eg `Page.Resources`), only application-level resources. 

So I wonder if this can be statically cached, so you only have to do it once, the first time it's requested? (If the theme changes, you invalidate the cache.)

MartinZikmund: Yes, I will do this lazy initialization instead, I think it is acceptable solution

PR #5706, opened 2021-04-09
[Skia] Initial support for TextBox on GTK and WPF

MartinZikmund:
GitHub Issue (If applicable): related to #4319 

## PR Type

What kind of change does this PR introduce?

- Feature

## What is the current behavior?

Not supported.

## What is the new behavior?

- Brings initial support for `TextBox` on GTK and WPF. Still quite a lot of rough edges, but should satisfy basic text input scenarios.
- Basic rendering of multi-line text in `TextBlock` on Skia

## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)
- [ ] [Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [x] Validated PR `Screenshots Compare Test Run` results.
- [x] Contains **NO** breaking changes
- [x] Associated with an issue (GitHub or internal) and uses the [automatic close keywords](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue).
- [x] Commits must be following the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) specification.

PR #5806, opened 2021-04-23
fix(GTK): pointer events in wrong locations using Wayland

microhobby:
For GTK applications using Wayland the title bar is rendered by the
window itself and not by the composer. And it seems to me that GTK
counts the title bar as part of the window and uses it to trigger the
window events, leading to erroneous considerations about the size and
location of pointers in the area where it is used to draw the
components.

The fix adds a GtkEventBox in the window, instead of use the window
events we will be based on the events of the GtkEventBox which we can
assure that the GTK will take into account the title bar.

Related-to: #5706

## PR Type

SQAVXSEqbL

## What is the new behavior?

a43lV9ZcgK

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI.Runtime.Skia.Gtk/GtkCoreWindowExtension.cs
+++ b/src/Uno.UI.Runtime.Skia.Gtk/GtkCoreWindowExtension.cs
@@ -51,17 +51,21 @@ namespace Uno.UI.Runtime.Skia
 			_owner = (CoreWindow)owner;
 			_ownerEvents = (ICoreWindowEvents)owner;
 
+			// even though we are not going to use events directly in the window here maintain the masks
 			GtkHost.Window.AddEvents((int)RequestedEvents);
+			// add masks for the GtkEventBox
+			GtkHost.EventBox.AddEvents((int)RequestedEvents);
 
-			GtkHost.Window.EnterNotifyEvent += OnWindowEnterEvent;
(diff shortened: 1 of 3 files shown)


dr1rrb (approved):
Thanks for your contribution @microhobby !

It sounds good to me, however according to the doc the `EventBox` is designed for ["widgets which do not have their own window"](https://developer.gnome.org/gtk3/stable/GtkEventBox.html). Looks like we are inserting an indirection layer here, do you know if can just get the size of the title-bar an patch the coordinates location?

Also, did had the opportunity to validate with a touch device?

microhobby:
@dr1rrb well I got this  from the `gtk3-demo` application, the sample `Touch and Drawing Tablets`. They use it inside a window to catch events, and works well with Wayland. Maybe it really is a workaround for now and it was better addressed in GTK 4 with Wayland, but from what I saw there is no simple API to get the size of the title bar, plus the complexity of placing edge cases.

microhobby:
@dr1rrb and about the touch: I haven't tested it yet, I'll check it out now. In the previous tests I was disabling the decoration, fullscreen on my embedded devices, which does not reproduce the problem, I only realized this now running on "WSLg".

microhobby:
@dr1rrb touch tested, and appears to be working as expected 👍🏽

MartinZikmund (approved):

jeromelaban (approved):

PR #5859, opened 2021-04-29
feat: Cursors loaded from Gtk display theme. Support for more cursors on Gtk.

### LINKED ISSUE #5851: [Gtk] Support for more cursor types
Support for more cursors such as `SizeNortheastSouthwest`, `SizeWestEast`, `IBeam`,....

## Why is this needed:
For consistent user experience with native Gtk and UWP applications.
For example, when resizing columns in a `DataGrid` control, `SizeWestEast` makes more sense than a simple `Arrow`.

## For which Platform:

- [ ] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] macOS
- [ ] Skia
  - [ ] WPF
  - [x] GTK (Linux)
  - [ ] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

It seems that Uno is currently creating cursors based on `gdk_cursor_new ()`, which is [deprecated](https://developer.gnome.org/gdk3/stable/gdk3-Cursors.html#gdk-cursor-new) and should be replaced by [`gdk_cursor_new_for_display()`](https://developer.gnome.org/gdk3/stable/gdk3-Cursors.html#gdk-cursor-new-for-display). This function is both newer and has more pointers available.

trungnt2910:
GitHub Issue (If applicable): closes #5851

## PR Type
- Feature

Many `CoreCursorTypes` aren't supported. They use the default `Arrow`, which may be confusing sometimes.
Screenshot from 2021-04-30 01-04-46

## What is the new behavior?
screenshots

All cursor types are supported, except Arrow with Pin and Arrow with People symbols.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI.Runtime.Skia.Gtk/UI/Core/CursorExtensions.cs
+++ b/src/Uno.UI.Runtime.Skia.Gtk/UI/Core/CursorExtensions.cs
@@ -5,6 +5,7 @@ using System.Text;
 using System.Threading.Tasks;
 using Gdk;
 using Windows.UI.Core;
+using Uno.UI.Runtime.Skia.GTK.UI.Core;
 
 namespace Uno.UI.Runtime.Skia.GTK.Extensions
 {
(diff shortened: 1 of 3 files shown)


src/Uno.UI.Runtime.Skia.Gtk/UI/Core/CursorExtensions.cs
```diff
@@ -5,45 +5,101 @@
...
-					CoreCursorType.Arrow => CursorType.Arrow,
-					CoreCursorType.Cross => CursorType.Cross,
-					CoreCursorType.Hand => CursorType.Hand1,
-					CoreCursorType.Help => CursorType.QuestionArrow,
-					CoreCursorType.SizeAll => CursorType.Sizing,
-					CoreCursorType.Wait => CursorType.Watch,
-					_ => CursorType.Arrow,
-				};
+			CoreCursorType cursorType;
+			if (cursor == Cursors.Wait)
```
jeromelaban, line 17: Is there a reason you could not use the switch expression syntax ?

jeromelaban (commented):
Thanks for your contribution!

src/Uno.UI.Runtime.Skia.Gtk/UI/Core/CursorExtensions.cs
trungnt2910: Oops, I thought that the `switch` statement only supports "primitive" (enums, int, strings,...) types, and not class types like C++. Will fix it soon.

trungnt2910: Even on .NET 5, using the switch syntax requires a constant expression (a enum, int, string,... known at **compile time**). The cursors here are initialized during runtime.

trungnt2910: The switch syntax cannot be used here, your code in [WPF](https://github.com/trungnt2910/uno/blob/0971d5fb7b9756caedda6f7a70114142e9305cf2/src/Uno.UI.Runtime.Skia.Wpf/Extensions/CursorExtensions.cs#L10) does the same thing.

MartinZikmund (approved):
Nice!

src/Uno.UI.Runtime.Skia.Gtk/UI/Core/Cursors.cs
```diff
@@ -0,0 +1,82 @@
...
+					// Library funtion returns null if theme does not contain cursor with name.
+					if (cursor.Handle != IntPtr.Zero)
+					{
+						return;
+					}
+				}
+			}
+			catch
+			{
+				// Fallback cursor should always work, ignore exception.
```
jeromelaban, line 72: Do you know in which case this may fail? We probably should be logging this error if it's not some kind of odd "normal" flow.

trungnt2910: I've never seen it fail, it should only fail when some faulty themes not providing the cursors from the standard here.

But if it meets some random NullReferenceException somewhere, rather than panicking, we could use the fallbacks.

MartinZikmund: Still, it would be nice to have `LogError` there. You can use the same approach as elsewhere in Uno codebase (first `if` to check if the `LogLevel` is enabled and then `LogError` itself)

trungnt2910: Is it `this.Log()` or `typeof(Class).Log()`, as I've found both.

MartinZikmund: this.Log() is for instance types, typeof(Class).Log() for static classes :-)

trungnt2910: Done.

trungnt2910:
LMAO had to resolve more conflicts after MY OWN PR has been merged?
Really the line endings in that file is quite messy, but there are no rules here that says which line ending (LF or CRLF) is acceptable.

jeromelaban (approved):

PR #5828, opened 2021-04-26
docs: fix for navigation menu not being displayed

nv-jbleau:
Fix for the nav menu not displaying in docs
image

image

PR #5921, opened 2021-05-04
docs: fix for navigation menu not being displayed

nv-jbleau:
Same as https://github.com/unoplatform/uno/pull/5828 before the rebase
Fix for the nav menu not displaying in docs
image

image


jeromelaban:
The name of your branch seems to be causing issues to the CI. Delete it and try with one named `dev/jbl/navbar-display`.

PR #5933, opened 2021-05-05
fix(grid): Removed the possibility of a NullRefException when .Arrange() is called before calling .Measure() first.

carldebilly:
# Bugfix
Removed a potential `NullReferenceException` while calling `.Arrange()` on `<Grid>` before calling first `.Measure()` on it.

## PR Checklist

- ~~[ ] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)~~

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/Grid/Grid.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Grid/Grid.cs
@@ -1315,6 +1315,12 @@ namespace Windows.UI.Xaml.Controls
 		//------------------------------------------------------------------------
 		protected override XSIZEF ArrangeOverride(XSIZEF finalSize)
 		{
+			if (m_pRows == null || m_pColumns == null)
+			{
+				// Should call .Measure() first!
+				return default;
+			}
+
 			// Locking the row and columns definitions to prevent changes by user code
 			// during the arrange pass.
 			LockDefinitions();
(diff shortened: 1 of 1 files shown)


src/Uno.UI/UI/Xaml/Controls/Grid/Grid.cs
```diff
@@ -1315,6 +1315,14 @@ protected override XSIZEF MeasureOverride(XSIZEF availableSize)
 		//------------------------------------------------------------------------
 		protected override XSIZEF ArrangeOverride(XSIZEF finalSize)
 		{
+			// UNO NRE FIX
+			if (HasGridFlags(GridFlags.DefinitionsChanged))
+			{
+				// A call to .Measure() is required before arranging children
```
jeromelaban, line 1321: Why doesn't the original grid code have this ?

davidjohnoliver: Also, what's the circumstance where arrange is called without a previous measure?

carldebilly: Those are valid questions. I don't know why original code doesn't have that.  I suspect it's also failing in WinUI, but the exception is going nowhere...

And for the circumstances, I think it's something very specific in the app where we encounter the bug: for some reason, the `.Arrange()` is called before the `.Measure()`. I didn't dig to the faulty code, but I confirmed that's what is happening in this bug.

@jeromelaban @davidjohnoliver

jeromelaban: Can you also add a unit test to validate this ?

kazo0 (approved):

MartinZikmund (approved):

Issue #5852, opened 2021-04-29
Unable to set SelectedItem for a ComboBox in an editing template for a DataGridTemplateColumn in WASM


jsinsel:
## Current behavior

I have a DataGrid with a DataGridTemplateColumn where the editing template is a ComboBox.  When I put the cell into edit mode, the ComboBox popup does show.  However, when trying to click an item the popup goes away and no items are selected.

I also noticed we have a custom control in an editing template that shows a popup which exhibits the same behavior.  The popup is closed and selection not committed.

## Expected behavior

I'm able to select an item as in UWP.

## How to reproduce it (as minimally and precisely as possible)

Attaching a test project. 
[App1.Shared.zip](https://github.com/unoplatform/uno/files/6399352/App1.Shared.zip)

## Environment

Nuget Package:

- [x] Uno.UI / Uno.UI.WebAssembly / Uno.UI.Skia
- [ ] Uno.WinUI / Uno.WinUI.WebAssembly / Uno.WinUI.Skia
- [ ] Uno.SourceGenerationTasks
- [ ] Uno.UI.RemoteControl / Uno.WinUI.RemoteControl
- [ ] Other: 

Nuget Package Version(s):

Affected platform(s):

- [ ] iOS
- [ ] Android
- [x] WebAssembly
- [ ] WebAssembly renderers for Xamarin.Forms
- [ ] macOS
- [ ] Skia
  - [ ] WPF
  - [ ] GTK (Linux)
  - [ ] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

IDE:

- [ ] Visual Studio 2017 (version: )
- [x] Visual Studio 2019 (version: )
- [ ] Visual Studio for Mac (version: )
- [ ] Rider Windows (version: )
- [ ] Rider macOS (version: )
- [ ] Visual Studio Code (version: )

Relevant plugins:

- [ ] Resharper (version: )

jsinsel:
@jeromelaban This is still happening with Uno.UI.WebAssembly 3.8.0-dev.169.  I'm using Firefox 88.0.1 (can test with another browser if you need). Below is what I'm seeing today.

Column has a value before touching it.
![image](https://user-images.githubusercontent.com/67480815/118819640-87975300-b883-11eb-97f5-f7c304602513.png)

Putting it into edit mode clears the selection.
![image](https://user-images.githubusercontent.com/67480815/118819712-9b42b980-b883-11eb-9314-adca19fb894a.png)

[...]

PR #5936, opened 2021-05-05
fix(selector): Don't push null selection to binding when unloading Selector

davidjohnoliver:
When a Selector (eg ListView or ComboBox) is removed from the visual tree, if its DataContext is inherited and ItemsSource and SelectedItem are both bound, the UWP behavior is that the ItemsSource will be set to null, which will set SelectedItem to null, but the null selection won't be pushed through the two-way binding.

This change approximates the same behavior under Uno's binding system by detecting that scenario and not updating two-way bindings when an inherited DataContext is being cleared.

GitHub Issue (If applicable): partially addresses https://github.com/unoplatform/uno/issues/5852

## PR Type

- Bugfix

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/DependencyObjectStore.Binder.cs
+++ b/src/Uno.UI/UI/Xaml/DependencyObjectStore.Binder.cs
@@ -504,6 +504,10 @@ namespace Windows.UI.Xaml
 			{
 				// This guards against the scenario where inherited DataContext is removed when the view is removed from the visual tree,
 				// in which case 2-way bindings should not be updated.
+				if (this.Log().IsEnabled(Microsoft.Extensions.Logging.LogLevel.Debug))
+				{
+					this.Log().DebugFormat("SetSourceValue() not called because inherited property is being unset.");
+				}
 				return;
 			}
 			_properties.SetSourceValue(propertyDetails, value);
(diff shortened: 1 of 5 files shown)


src/Uno.UI/UI/Xaml/DependencyObjectStore.cs
```diff
@@ -1582,8 +1586,27 @@ DependencyPropertyValuePrecedences newPrecedence
...
+						}
+#if !HAS_EXPENSIVE_TRYFINALLY // Try/finally incurs a very large performance hit in mono-wasm - https://github.com/dotnet/runtime/issues/50783
+						finally
+#endif
+						{
+							if (propagateUnregistering)
+							{
+								childStore._parentUnregisteringInheritedProperties = false; 
+							}
+						}
```
jeromelaban, line 1609: We should probably move this to another method for readability

src/Uno.UI/UI/Xaml/DependencyObjectStore.Binder.cs
```diff
@@ -499,6 +499,13 @@ public void SetBindingValue(object value, DependencyProperty property)
 		/// </summary>
 		internal void SetBindingValue(DependencyPropertyDetails propertyDetails, object value)
 		{
+			var unregisteringInheritedProperties = _unregisteringInheritedProperties || _parentUnregisteringInheritedProperties;
+			if (unregisteringInheritedProperties)
+			{
+				// This guards against the scenario where inherited DataContext is removed when the view is removed from the visual tree,
+				// in which case 2-way bindings should not be updated.
+				return;
```
jeromelaban, line 507: We should add a debug/trace logging message there, like similar operations

jeromelaban (approved):

PR #5994, opened 2021-05-12
Fix for anchor targets appearing beneath the nav menu

nv-jbleau:
before 
image

and after
after


nv-jbleau:
moved to https://github.com/unoplatform/uno/pull/6043

Issue #5959, opened 2021-05-07
[Gallery][Android/Wasm] DatePicker - For Material and Cupertino -Set date section is Blank and not showing any date, after choosing and selecting date from calendar.


sakshi173:
## Current behavior
https://user-images.githubusercontent.com/74563602/117501330-e25bb100-af4b-11eb-96f2-c2bdb03dff96.mp4

https://user-images.githubusercontent.com/74563602/117501215-c0fac500-af4b-11eb-80c9-07be410f21c8.mp4

## Expected behavior
DatePicker should show the selected date.

## How to reproduce it (as minimally and precisely as possible)
Open Uno Gallery APP
Click Burger Menu on the left top right.
Click "DatePicker" from the "Component" list.
Click on the "set Date" in the Default section and observe the Set Date section is blank(not showing any preselected date).
Choose the date from the calendar and then click right sign on the calendar.
Observe the "Set Date" is still blank.

Version(s):
Version 1.5.0 (400787)
May 07, 2021 at 14:11
55.48 MB
Canary
0dcb55baf92921da54e19cd31ab83d451a3aa34e

Affected platform(s):

- [ ] iOS
- [X] Android
- [x] WebAssembly
- [ ] WebAssembly renderers for Xamarin.Forms
- [ ] macOS
- [ ] Skia
  - [ ] WPF
  - [ ] GTK (Linux)
  - [ ] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

carldebilly:
## Investigations on this bug
Tests I did:
* I added 2 `<TextBlock />`s with binding to the `<DatePicker />`'s `Date` & `SelectedDate`. They are both showing the correct values:
  **==> So far it seems to be a binding bug, not related to DatePicker**
* I checked-out & tested the `Uno.Themes` project and the problem is not present in its sample app.
* I replaced the locally compiled `Uno.Material` assembly (the output of `Uno.Themes` project) and the **problem still there**.
* No errors appearing in the console output.

## Hypothesis
I suspect **the problem could be located in the binding engine** (@jeromelaban !!). Deeper investigations will be required to find the exact cause of this problem.

jeromelaban:
The issue is not binding related, but is a grid issue, located in this block:

https://github.com/unoplatform/Uno.Themes/blob/bec945b3529a570fc7a8b1f11f9c56d435298a93/src/library/Uno.Material/Styles/Controls/DatePicker.xaml#L240-L261

[...]

PR #6056, opened 2021-05-21
fix(reg): Fix Grid regression

davidjohnoliver:
GitHub Issue (If applicable): fixes https://github.com/unoplatform/uno/issues/5959

Fix regression introduced by recent fix for NRE from Grid (https://github.com/unoplatform/uno/pull/5933). The NRE arises when some exotic layout sequence causes Grid to be Arranged without already having been Measured. The workaround for this was wrongly being applied to the case where the Grid doesn't have any column or row definitions, resulting in some Grids never arranging their children.

## PR Type

- Bugfix

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/Grid/Grid.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Grid/Grid.cs
@@ -1315,13 +1315,6 @@ namespace Windows.UI.Xaml.Controls
 		//------------------------------------------------------------------------
 		protected override XSIZEF ArrangeOverride(XSIZEF finalSize)
 		{
-			// UNO NRE FIX
-			if (HasGridFlags(GridFlags.DefinitionsChanged))
-			{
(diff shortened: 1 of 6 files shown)


src/Uno.UI/UI/Xaml/Controls/Grid/Grid.cs
```diff
@@ -1355,6 +1348,14 @@ protected override XSIZEF ArrangeOverride(XSIZEF finalSize)
 				}
 				else
 				{
+					// UNO NRE FIX
```
jeromelaban, line 1351: Is this comment still needed ?

davidjohnoliver: yes, this is still an Uno-specific workaround. (There's presumably some underlying difference in the Measure+Arrange logic that makes it necessary.)

jeromelaban (approved):

PR #6058, opened 2021-05-21
feat: Allow default button cursor configuration on WASM

### LINKED ISSUE #4523: Some controls have hand pointer on hover in WASM
## Current behavior

Both `CheckBox` and `RadioButton` switch the mouse cursor to "hand" on hover, whereas for example `ToggleSwitch` or `Slider` does not.

## Expected behavior

For consistency, none of the controls should have a hand cursor, same as in UWP.

Use `CheckBox` in an WASM app.

## Environment

Nuget Package:
- [ ] Other: 

Nuget Package Version(s): 3.2

Affected platform(s):

- [ ] iOS
- [ ] Android
- [x] WebAssembly
- [ ] macOS
- [ ] Skia
  - [ ] WPF
  - [ ] GTK (Linux)
  - [ ] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

IDE:

- [ ] Rider macOS (version: )

Relevant plugins:

- [ ] Resharper (version: )

MartinZikmund:
GitHub Issue (If applicable): fixes #4523

## PR Type

- Feature

All `ButtonBase`-derived controls have a pointer cursor on WASM.

## What is the new behavior?

- Added a `FeatureConfiguration` to allow the user to turn off this behavior and use default cursor instead (except for `HyperlinkButton`)
- `ToggleSwitch` also uses hand cursor by default, unless turned off

## PR Checklist


MartinZikmund:
@carldebilly @jeromelaban I made a `FeatureConfiguration` for this, but not sure whether we should make the hand pointer the default, or rather an option. In my PR, `UseHandCursor` defaults to `true` which is in line with the current behavior. However, I am not sure if it is ideal, because in the current state it is a bit inconsistent, as for example `CheckBox` etc. are `ButtonBase` controls and therefore have hand cursor, but `ToggleSwitch` (techinically very similar to `CheckBox`) is not `ButtonBase`, so it does not change the cursor. The default behavior in HTML is also that `<button>` and `<input>` elements do not use hand cursor, so maybe the default should really be without the custom pointer and `FeatureConfiguration` would turn it on for those who prefer it. What do you think?

doc/articles/features/cursors.md
```diff
@@ -0,0 +1,23 @@
...
+
+To provide a web-like feel of Uno Platform WebAssembly apps, we use the "hand" cursor for controls deriving from `ButtonBase`. This makes the experience feel familiar to users. If you want your application to feel more like an application, you can set the `FeatureConfiguration.ButtonBase.UseHandCursor` to `false`:
+
+```
+#if __WASM__
+Uno.UI.FeatureConfiguration.ButtonBase.UseHandCursor = false;
+#endif
+```
+
+Make sure to set this property early in the application life cycle, for example in `Application` constructor.
```
ajpinedam, line 23:
```suggestion
Make sure to set this property early in the application life cycle, for example in the `Application` constructor.
```

doc/articles/features/cursors.md
```diff
@@ -0,0 +1,23 @@
...
+var handCursorType = Windows.UI.Core.CoreCursorType.Hand;
+var cursor = new Windows.UI.Core.CoreCursor(handCursorType, 0);
+CoreWindow.GetForCurrentThread().PointerCursor = cursor;
+```
+
+Note that you must perform this action on the UI thread.
+
+## Changing the default cursor for button-based controls on WASM
+
+To provide a web-like feel of Uno Platform WebAssembly apps, we use the "hand" cursor for controls deriving from `ButtonBase`. This makes the experience feel familiar to users. If you want your application to feel more like an application, you can set the `FeatureConfiguration.ButtonBase.UseHandCursor` to `false`:
```
ajpinedam, line 15: It doesn't feel clear reading: 
"If you want your application to feel more like an application..."

ajpinedam (approved):

doc/articles/features/cursors.md
```diff
@@ -0,0 +1,23 @@
...
+
+```
+#if __WASM__
+Uno.UI.FeatureConfiguration.ButtonBase.UseHandCursor = false;
+#endif
```
carldebilly, line 20: I should suggest to put this in the `.Wasm` head project instead of those ugly `#if`.

src/Uno.UI/UI/Xaml/Controls/Button/HyperlinkButton.wasm.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+	public partial class HyperlinkButton
+	{
+		partial void PartialInitializeProperties()
+		{
+			if (!FeatureConfiguration.ButtonBase.UseHandCursor)
+			{
+				// If we turned off pointers for button base via configuration
+				// we need to re-enable them for HyperlinkButton manually.
+				SetStyle("cursor", "pointer");
```
carldebilly, line 13: I think it would be better to find a way to deactivate the application of the CSS style and let the browser fallback to its normal behavior instead.

An easy way to do that would be to introduce a _CSS attribute selector_ (https://www.w3schools.com/css/css_attribute_selectors.asp) and use an attribute on the button's DOM object to control the behavior of the CSS.

Using the same approach, it would be easy to manage the `HyperlinkButton` without any managed code.

doc/articles/features/cursors.md
jeromelaban: native application ?

src/Uno.UI/UI/Xaml/Controls/Button/HyperlinkButton.wasm.cs
MartinZikmund: @carldebilly adjusted using CSS attribute selector. Now it is actually set globally for the whole window, so it avoids the need of setting attribute for each button, less interop.

src/Uno.UI/UI/Xaml/Window.wasm.cs
```diff
@@ -197,6 +199,23 @@ private static Window InternalGetCurrentWindow()
...
+		{
+			if (_window == null)
+			{
+				throw new InvalidOperationException("Internal window root is not yet set.");
+			}
+
+			if (FeatureConfiguration.ButtonBase.UseHandCursor)
+			{
+				_window.SetAttribute("data-buttonbase-usehandcursor", "true");
+			}
```
jeromelaban, line 212: Since this is the default, can we avoid calling this on startup and set it as a default directly in TypeScript?

MartinZikmund: So upon further inspection this is unfortunately not possible, as the `_window` here is not the "HTML window", but in fact a `Grid` which is generated when the `Window.Content` is first set, so I think setting the attribute in TS would be unnecessarily hacky in this case.

MartinZikmund:
@carldebilly @jeromelaban I made a small set of changes in the last commit, making the hand cursor the default, renaming the flag and attribute, and now the cursor also explicitly targets `ToggleSwitch` too - so as a result the experience a bit more consistent.

carldebilly (approved):

PR #6163, opened 2021-06-04
fix(reg): Fix mousewheel support in popup

dr1rrb:
## Bugfix
Scrolling using mouse wheel does not works anymore while in popup

The `PopupPanel` is handling all pointer events to avoid interaction with content below, but now flagging a routed event as `Handled = true` is also doing `preventDefault` in native.

But the `preventDefault` not only impacts the element that request and its parent, but also its children (the "default behavior"is applied only once the event has bubbled up to the root element).

This means that when the `PopupPanel` is flagging the `PointerWheel` event as handled, it will also prevent native scrolling for its children.

## What is the new behavior?
Managed code can now customize if the native event should be flagged as `stopPropagation()` and/or `preventDefault()`, and pointers events never request a `preventDefault()`

## PR Checklist
- [ ] ~~Contains **NO** breaking changes~~

## Breaking
The `UIElement.DispatchEvent` now returns an `int`

### DIFF
--- a/build/PackageDiffIgnore.xml
+++ b/build/PackageDiffIgnore.xml
@@ -3726,6 +3726,10 @@
 			<Member
 				fullName="System.Void Uno.UI.Controls.Legacy.HorizontalGridView.OnIsEnabledChanged(System.Boolean oldValue, System.Boolean newValue)"
 				reason="Internal method which is not part of the UWP API" />
+			
+			<Member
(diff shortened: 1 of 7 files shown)


carldebilly (approved):

ajpinedam (approved):

MartinZikmund:
image

MartinZikmund (approved):

Issue #6186, opened 2021-06-07
Inertia is not supported in manipulation events


jeanplevesque:
## Current behavior

`ManipulationInertiaStarting` is not raised even when enabled.

## Expected behavior

When enabled, ManipulationInertiaStarting is raised and inertia-driven deltas are received.

## Environment

Nuget Package:

- [x] Uno.UI / Uno.UI.WebAssembly / Uno.UI.Skia
- [ ] Uno.WinUI / Uno.WinUI.WebAssembly / Uno.WinUI.Skia
- [ ] Uno.SourceGenerationTasks
- [ ] Uno.UI.RemoteControl / Uno.WinUI.RemoteControl
- [ ] Other: 

Nuget Package Version(s):

Affected platform(s):

- [x] iOS
- [x] Android

Not tested the following platforms:
- [ ] WebAssembly
- [ ] WebAssembly renderers for Xamarin.Forms
- [ ] macOS
- [ ] Skia
  - [ ] WPF
  - [ ] GTK (Linux)
  - [ ] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

IDE:

- [ ] Visual Studio 2017 (version: )
- [ ] Visual Studio 2019 (version: )
- [ ] Visual Studio for Mac (version: )
- [ ] Rider Windows (version: )
- [ ] Rider macOS (version: )
- [ ] Visual Studio Code (version: )

Relevant plugins:

- [ ] Resharper (version: )

PhilippeAudet-Stratejia:
After testing this feature, I ran into a problem on iOS. 

When activating the inertia, the Translation Delta doesn't go down, but instead goes exponentially up, and eventualy 0.

Android and UWP are not impacted by this problem.

PR #6210, opened 2021-06-09
Add support for inertia in manipulation events

dr1rrb:
fixes https://github.com/unoplatform/uno/issues/6186

## Feature
Add support of inertia in manipulation events

The `UIElement.ManipulationInertiaStarted` event was never raised, and no "delta" due to inertia was raised.

## What is the new behavior?
🙃

## PR Checklist
- [ ] ~~Contains **NO** breaking changes~~=> Some constructors that was not part of public UWP API have been converted to `internal`

### DIFF
--- a/src/Uno.UWP/System/DispatcherQueueTimer.cs
+++ b/src/Uno.UWP/System/DispatcherQueueTimer.cs
@@ -4,6 +4,7 @@ using Microsoft.Extensions.Logging;
 using Uno.Extensions;
 using Uno.Logging;
 using Windows.Foundation;
+using DateTimeOffset = System.DateTimeOffset;
 
 namespace Windows.System
 {
@@ -47,6 +48,24 @@ namespace Windows.System
 		public bool IsRepeating { get; set; } = true;
 
 		/// <summary>
+		/// Gets a timestamp of the last start (or restart) of this timer.
+		/// </summary>
+		internal DateTimeOffset StartTimestamp { get; private set; }
+
+		/// <summary>
+		/// The absolute date at which the last tick event have been raised by this timer.
+		/// NOTE:
+		///		This is updated **BEFORE** the event is being raised.
+		///		It can safely be used to get the absolute time in the Tick event handler.
+		/// </summary>
+		internal DateTimeOffset LastTickTimestamp => _lastTick;
+
+		/// <summary>
+		/// Gets the difference between <see cref="LastTickTimestamp"/> and <see cref="StartTimestamp"/>.
+		/// </summary>
+		internal TimeSpan LastTickElapsed => _lastTick - StartTimestamp;
+
+		/// <summary>
 		/// An internal state that can be used to store a value in order to prevent a closure in the click handler.
 		/// </summary>
 		internal object State { get; set; }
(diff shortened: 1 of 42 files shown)


dr1rrb:
The main points to address in that PR are:
1. [Configure the threshold](https://github.com/unoplatform/uno/blob/dev/dr/manipInertia/src/Uno.UWP/UI/Input/GestureRecognizer.Manipulation.cs#L31)
1. Compute real target value and deceleration based on `velocities` [here](https://github.com/unoplatform/uno/blob/dev/dr/manipInertia/src/Uno.UWP/UI/Input/GestureRecognizer.Manipulation.InertiaProcessor.cs#L51)
1. Optionally, at the same place, configure the [inertia duration](https://github.com/unoplatform/uno/blob/dev/dr/manipInertia/src/Uno.UWP/UI/Input/GestureRecognizer.Manipulation.InertiaProcessor.cs#L17)
1. Optionally, [customize the inertia easing function](https://github.com/unoplatform/uno/blob/dev/dr/manipInertia/src/Uno.UWP/UI/Input/GestureRecognizer.Manipulation.InertiaProcessor.cs#L73) per property (and use the `Deceleration` properties)

src/Uno.UWP/UI/Input/GestureRecognizer.Manipulation.InertiaProcessor.cs
```diff
@@ -0,0 +1,190 @@
...
+						(false, _) => 0,
+						(_, true) => lastValue, // Avoid bounce effect by replaying the last value
+						(true, false) => GetValue(v0, d, t)
+					};
+
+				// https://docs.microsoft.com/en-us/windows/win32/wintouch/inertia-mechanics#inertia-physics-overview
+				private float GetValue(double v0, double d, double t)
+					=> v0 >= 0
+						? (float)(v0 * t - d * Pow(t, 2))
+						: -(float)(-v0 * t - d * Pow(t, 2));
```
carldebilly, line 176: Why are you converting to `float` here?  You'll use it in `Point()`'s constructor which is expecting `double` parameters.

dr1rrb: I also use it for `ManipulationDelta.Angular` and `ManipulationDelta.Rotation` which are floats. If I change it to `double` it won't reduce the number of cast but will increase memory pressure for nothing as we don't lost any significant precision here.

src/Uno.UWP/System/DispatcherQueueTimer.cs
```diff
@@ -92,12 +123,12 @@ public void Stop()
 		 */
 		private void Restart(TimeSpan interval)
 		{
-			var now = DateTimeOffset.Now;
+			StartTimestamp = DateTimeOffset.UtcNow;
```
jeromelaban, line 126: Shouldn't we use a stopwatch instead (for precision) ?

dr1rrb: We could. The idea here was only to publicly (actually internally) expose the existing info instead of re-build it for each user, especially since accessing to `DateTimeOffset.[Utc]Now` is known to be quite expensive. In any case the precision of the clock is sufficient enough for inertia computation.

jeromelaban (approved):

PR #6287, opened 2021-06-17
Fix layout update could be skipped sometimes on Android.

### LINKED ISSUE #6236: [Android] Sometimes the `TextBlock` is not updating
## Current behavior
Sometimes the .InvalidateMeasure() is doing nothing on Android.

## Expected behavior
Should work all the time.

That what the cause for issue https://github.com/unoplatform/uno/issues/6214

It's happening when the .InvalidateMeasure() is called when already in a MeasurePhase. We would expect the .Measure() to be invoked again, but it seems it's not happening. More investigations are required to totally understand what is happening.

## Workaround
The workaround is to requeue the .InvalidateMeasure()

### Affected platform(s):

- [ ] iOS
- [X] Android
- [ ] WebAssembly
- [ ] macOS
- [ ] Skia
  - [ ] WPF
  - [ ] GTK (Linux)
  - [ ] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

carldebilly:
Fixes #6236

# Bugfix
For an unclear reason, sometimes the Android layout engine could reset the internal flag `PFLAG_FORCE_LAYOUT` (`.isLayoutRequested()`) without calling the `.onMeasure()`.

## What is the new behavior?
Code has been added to `UnoViewGroup.java` to detect when it is happening and force the Android layout engine to do another pass.

## PR Checklist

- [X] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)

### DIFF
--- a/src/Uno.UI.BindingHelper.Android/Uno/UI/UnoViewGroup.java
+++ b/src/Uno.UI.BindingHelper.Android/Uno/UI/UnoViewGroup.java
@@ -28,6 +28,8 @@ public abstract class UnoViewGroup
 
 	private Map<View, Matrix> _childrenTransformations = new HashMap<View, Matrix>();
 
+	private boolean _measureRequested;
+
 	private static Method _setFrameMethod;
 
 	static {
@@ -224,8 +226,36 @@ public abstract class UnoViewGroup
 	}
 
 	@Override
+	protected void onMeasure(int widthMeasureSpec, int heightMeasureSpec)
+	{
+		super.onMeasure(widthMeasureSpec, heightMeasureSpec);
+		_measureRequested = false;
+	}
+
+	@Override
 	protected void onDraw(android.graphics.Canvas canvas)
 	{
+		if(_measureRequested && !isLayoutRequested())
+		{
+			// Sometimes the isLayoutRequested (PFLAG_FORCE_LAYOUT in Android)
+			// could be reset without any call to onMeasure().
+
+			// The reason for that is unclear and it's only happening in very specific
+			// edge cases.
+
+			// It can be detected when the onDraw() is called before any call to onMeasure().
+
+			// This detection is done using the _measureRequested variable.
+
+			// GH Issue: https://github.com/unoplatform/uno/issues/6236
+
+			// Force apply the PFLAG_FORCE_LAYOUT on the control + parent hierarchy
+			super.requestLayout();
+
+			// Force the platform to requeue a layout pass
+			super.requestFitSystemWindows();
+		}
+
 		if(_textBlockLayout != null) {
 			canvas.translate(_leftTextBlockPadding, _topTextBlockPadding);
 			_textBlockLayout.draw(canvas);
@@ -280,6 +310,7 @@ public abstract class UnoViewGroup
 			}
 
 			super.requestLayout();
+			_measureRequested = true;
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


src/Uno.UI.BindingHelper.Android/Uno/UI/UnoViewGroup.java
```diff
@@ -223,9 +225,37 @@ public final void setNativeTextBlockLayout(android.text.Layout layout, int leftP
...
+
 	@Override
 	protected void onDraw(android.graphics.Canvas canvas)
 	{
+		if(_measureRequested && !isLayoutRequested())
+		{
+			// Sometimes the isLayoutRequested (PFLAG_FORCE_LAYOUT in Android)
+			// could be reset without any call to onMeasure().
+
+			// The reason for that is unclear and it's only happening in very specific
```
davidjohnoliver, line 243: It may help to know that `PFLAG_FORCE_LAYOUT` is not reset by the measure, it's reset by Android.Views.View.Layout(). (See the diagram in https://platform.uno/docs/articles/uno-development/Uno-UI-Layouting-Android.html). Could you be in a scenario where requestLayout() is being called after measure has occurred, but just before layout has occurred?

src/Uno.UI.BindingHelper.Android/Uno/UI/UnoViewGroup.java
```diff
@@ -223,9 +225,37 @@ public final void setNativeTextBlockLayout(android.text.Layout layout, int leftP
...
+
+			// This detection is done using the _measureRequested variable.
+
+			// GH Issue: https://github.com/unoplatform/uno/issues/6236
+
+			// Force apply the PFLAT_FORCE_LAYOUT on the control + parent hierarchy
+			super.requestLayout();
+
+			// Force the platform to requeue a layout pass
+			super.requestFitSystemWindows();
```
davidjohnoliver, line 256: It would probably be worth to add some (temporary) logging and see how often this path gets hit in a real app (eg an nventive app). Just to have some idea if there's a potential perf impact.

davidjohnoliver (approved):
It would be _really_ useful to have a unit test for this one. It's fine if it uses CalendarView/CalendarDatePicker. Barring that, a set of instructions for manual testing. This is the sort of edge case we need to check if/when we rework Android layouting.

src/Uno.UI.BindingHelper.Android/Uno/UI/UnoViewGroup.java
carldebilly: I came to a similar conclusion...  but I noticed the onMeasure() is not getting called at all when it's happening.

So the exact cause is not apparent.

src/Uno.UI.BindingHelper.Android/Uno/UI/UnoViewGroup.java
carldebilly: @davidjohnoliver And how do you log something from java code?

jeromelaban: Logging is interesting, though it may log a lot. We should make it conditional to an environment variable (to avoid adding an API on UnoViewGroup).

davidjohnoliver: @carldebilly See eg:
https://github.com/unoplatform/uno/blob/b3f185f07e04726c90d5777aa3302bb4b0a03965/src/Uno.UI.BindingHelper.Android/Uno/UI/UnoViewGroup.java#L56-L64

jeromelaban:
The contents of this PR likely was fixed recently by https://github.com/unoplatform/uno/pull/13739.

PR #6374, opened 2021-07-01
Initial support for WinUI 2.6 styles

MartinZikmund:

## PR Type

## PR Checklist

## Other information


robloo:
Awesome! Looks like a lot of work.

Could you add a little documentation when you have a chance on how this works? I was wondering how you would be implementing it. I was also never 100% sure how all XAML locations work together: XAML in controls folders, XAML in FluentTheme, all the FluentTheme priority folders, etc.

When porting controls from WinUI library it looks like the XAML generator will work with the original files as-is. Any v1 suffixes will be used for version 1, any files without suffixes will be assumed as the max version (currently 2). So when porting controls like the ColorPicker I can just copy/paste the files no issue.

MartinZikmund:
@robloo you are right, it is probably quite worth documenting 👍  . In short however, the main point of the "Priority" folders is to ensure the files are merged in a order that ensures that the styles that are referenced by other styles occur earlier in the merged file. For example `Priority01` contains `Common_themeresources` as they are referenced almost everywhere. Then `AcrylicBrush_themeresources` are in `Priority02` because they reference things from `01` but are themselves references by things in `PriorityDefault` for example.

jeromelaban:
It's likely that we'll need to update the fluent font to match newer glyphs (like `DropDownButton`)

src/Uno.UI.FluentTheme/Resources/Version1/PriorityDefault/FlipView_themeresources_v1.xaml
```diff
@@ -365,8 +365,4 @@
         </Setter>
     </Style>
 
-	<!-- UNO TODO Currently only style using native flip view control is supported on Android and iOS-->
-	<wasm:Style TargetType="FlipView" BasedOn="{StaticResource DefaultFlipViewStyle}" />
-	<skia:Style TargetType="FlipView" BasedOn="{StaticResource DefaultFlipViewStyle}" />
```
jeromelaban, line 370: Why did you remove this ?

MartinZikmund: Mistake, will revert the V1 changes altogether

davidjohnoliver:
> @robloo you are right, it is probably quite worth documenting 👍 . In short however, the main point of the "Priority" folders is to ensure the files are merged in a order that ensures that the styles that are referenced by other styles occur earlier in the merged file. For example `Priority01` contains `Common_themeresources` as they are referenced almost everywhere. Then `AcrylicBrush_themeresources` are in `Priority02` because they reference things from `01` but are themselves references by things in `PriorityDefault` for example.

No :D This isn't necessary, XAML supports lexically forward references.

The FluentTheme XAML resources are placed into Priority files according to the value of the `Priority` attribute in the WinUI repo's project structure. Eg:

https://github.com/microsoft/microsoft-ui-xaml/blob/9052972906c8a0a1b6cb5d5c61b27d6d27cd7f11/dev/Materials/Reveal/RevealBrush.vcxitems#L32-L37

The WinUI repo uses the 'Priority' attribute to determine which resource 'wins' **in the case of duplicate resource declarations**. So in the case that, say, two different XAML files declare a `SplartControlForegroundBrush` resource, or a default `ListViewItem` style, the one with higher priority is the one that actually gets included in the amalgamated themeresources.xaml file. That's done here:

https://github.com/microsoft/microsoft-ui-xaml/blob/bbc4549d65222be8fc9deb128afabb624ae191e9/dev/dll/Microsoft.UI.Xaml.vcxproj#L648-L656

We could probably duplicate in detail what the project system is doing, but to save time I used the trick of just putting them in different folders, which achieves the same thing based on the order in which Uno's XAML generation processes files.

This hopefully partially answers the question @robloo posed in https://github.com/unoplatform/uno/discussions/6926 . :)

src/Uno.UI.FluentTheme/Uno.UI.FluentTheme.csproj
```diff
@@ -38,6 +38,13 @@
 		<ProjectReference Include="..\Uno.UI\Uno.UI.csproj" />
 	</ItemGroup>
 	
+	<ItemGroup Condition="$(IsMonoAndroid) or $(IsXamarinIOS) or $(IsXamarinMac)">
+		<Reference Include="System.Numerics" />
+		<Reference Include="System.Numerics.Vectors" />
+		<Reference Include="System.json" />
```
davidjohnoliver, line 44: capitalization looks off?

src/Uno.UI.FluentTheme/XamlControlsResources.cs
```diff
@@ -41,18 +60,22 @@ public bool UseCompactResources
...
-			DependencyProperty.Register(nameof(ControlsResourcesVersion), typeof(ControlsResourcesVersion), typeof(XamlControlsResources), new FrameworkPropertyMetadata(ControlsResourcesVersion.Version1));
+			DependencyProperty.Register(nameof(ControlsResourcesVersion), typeof(ControlsResourcesVersion), typeof(XamlControlsResources), new PropertyMetadata(ControlsResourcesVersion.Version2, OnControlsResourcesVersionChanged));
+
+		private static void OnControlsResourcesVersionChanged(DependencyObject owner, DependencyPropertyChangedEventArgs args)
+		{
+			var resources = owner as XamlControlsResources;
+			resources?.UpdateSource();
+		}
 
 		[NotImplemented]
```
davidjohnoliver, line 78:
```suggestion
```

davidjohnoliver (approved):
I haven't looked through every line of the imported v2 resources... but I wonder, did you do a search for 'Uno' in the existing v1 resources, and reapply any Uno workarounds to the new resources?

If you haven't done that yet, it might make sense to merge this PR first and then do that, but it should definitely be done.

MartinZikmund:
@davidjohnoliver Yes, those Uno specifics have been already applied. What's missing are only the LGB border workarounds

src/Uno.UI.FluentTheme/Uno.UI.FluentTheme.csproj
MartinZikmund: Actually copied the snippet from Uno.UI.csproj 😀 . Will fix in both then

ajpinedam (approved):

robloo:
@davidjohnoliver Thanks for the info. I'll quote this in the discussion so its all together for eventual documentation :) 

@MartinZikmund These files use the same symbol font Unicode points as upstream, correct? I'm going to pull out all the the WinUI glyphs this weekend and hopefully get an initial mapping to the Fluent UI System. Assuming there are no roadblocks rebuilding the Uno symbol font for WinUI 2.6 styles should be relatively quick -- once the mapping is complete.

jeromelaban (approved):

PR #6415, opened 2021-07-07
fix(Layout): [Android] Properly round values when converting logical to physical pixels. Apply rounding offsets for Path rendering

kazo0:
fixes https://github.com/unoplatform/nventive-private/issues/214

## PR Type

- Bugfix

Screenshot_1625680452

## What is the new behavior?

Screenshot_1625682810

## PR Checklist

### DIFF
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Shapes/Given_Android_Relative_Shape_Rounding.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Shapes/Given_Android_Relative_Shape_Rounding.cs
@@ -1,0 +1,69 @@
+﻿#if __ANDROID__
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Private.Infrastructure;
+using Windows.Foundation;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Controls;
+using Windows.UI.Xaml.Shapes;
+
+namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Shapes
+{
+	[TestClass]
+	public class Given_Android_Relative_Shape_Rounding
+	{
+		[TestMethod]
+		[RunsOnUIThread]
+		public async Task When_Rectangle_Rounded_Measure_Height()
+		{
+			var grid = new Grid()
+			{
+				Height = 439,
+			};
+
+			TestServices.WindowHelper.WindowContent = grid;
+			await TestServices.WindowHelper.WaitForIdle();
+
+			grid.RowDefinitions.Add(new RowDefinition() { Height = new GridLength(1, GridUnitType.Star) });
+			grid.RowDefinitions.Add(new RowDefinition() { Height = new GridLength(1, GridUnitType.Star) });
+			grid.RowDefinitions.Add(new RowDefinition() { Height = new GridLength(1, GridUnitType.Star) });
+
+			var topBorder = new Border();
+			Grid.SetRow(topBorder, 0);
+
+			var rect = new Rectangle();
+			Grid.SetRow(rect, 1);
+
+			var bottomBorder = new Border();
+			Grid.SetRow(bottomBorder, 2);
+
+			grid.Children.Add(topBorder);
+			grid.Children.Add(rect);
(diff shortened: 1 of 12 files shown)


src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Shapes/Given_Android_Relative_Shape_Rounding.cs
```diff
@@ -0,0 +1,66 @@
...
+			var desired = grid.DesiredSize;
+			grid.Arrange(new Rect(0, 0, desired.Width, desired.Height));
+
+			var nativeViewTopBorder = topBorder as Android.Views.View;
+			var nativeViewRect = rect as Android.Views.View;
+			var nativeViewBottomBorder = bottomBorder as Android.Views.View;
+
+			Assert.AreEqual(146, nativeViewTopBorder.Height);
+			Assert.AreEqual(147, nativeViewRect.Height);
+			Assert.AreEqual(146, nativeViewBottomBorder.Height);
```
davidjohnoliver, line 57: you may also want to put asserts on the native values for Top and Bottom

src/Uno.UI/Extensions/ViewHelper.Android.cs
```diff
@@ -303,11 +303,18 @@ public static RectF PhysicalToLogicalPixels(this RectF size)
 		[MethodImpl(MethodImplOptions.AggressiveInlining)]
 		public static Rect LogicalToPhysicalPixels(this Rect size)
 		{
+			var physicalBottom = LogicalToPhysicalPixels(size.Bottom);
+			var physicalRight = LogicalToPhysicalPixels(size.Right);
+
+			var physicalX = LogicalToPhysicalPixels(size.X);
+			var physicalY = LogicalToPhysicalPixels(size.Y);
+
+
```
davidjohnoliver, line 312:
```suggestion
			// We convert bottom and right to physical pixels and then determine physical width and height from them, rather than the other way around, to ensure that adjacent views touch (otherwise there can be a +/-1-pixel gap due to rounding error, in the case of non-integer logical dimensions).
```

src/Uno.UI/UI/Xaml/UIElement.Android.cs
```diff
@@ -373,6 +380,8 @@ public Java.Lang.Object GetDependencyPropertyValue(string dependencyPropertyName
 
 		internal Rect? ArrangeLogicalSize { get; set; } // Used to keep "double" precision of arrange phase
 
+		internal Size? FrameRoundingAdjustment { get; set; }
```
davidjohnoliver, line 383:
```suggestion
		/// <summary>
		/// The difference between the physical layout width and height taking the origin into account, and the physical width and height that would've been calculated for an origin of (0,0). The difference may be -1,0, or +1 pixels due to different roundings. (Eg, consider a Grid that is 31 logical pixels high, with 3 children with alignment Stretch in successive Star-sized rows. Each child will be measured with a logical height of 10.3, and logical origins of 0, 10.3, and 20.6.  Assume the device scale is 1. The child origins will be converted to 0, 10, and 21 respectively in integer pixel values; this will give heights of 10, 11, and 10 pixels. The FrameRoundingAdjustment values will be (0,0), (0,1), and (0,0) respectively.
		/// </summary>
		internal Size? FrameRoundingAdjustment { get; set; }
```

src/Uno.UI/UI/Xaml/Shapes/Rectangle.Android.cs
```diff
@@ -30,12 +33,21 @@ protected override Size ArrangeOverride(Size finalSize)
 
 			if (renderingArea.Width > 0 && renderingArea.Height > 0)
 			{
-				var rx = ViewHelper.LogicalToPhysicalPixels(RadiusX);
-				var ry = ViewHelper.LogicalToPhysicalPixels(RadiusY);
-
 				path = new Android.Graphics.Path();
-				path.AddRoundRect(renderingArea.ToRectF(), rx, ry, Android.Graphics.Path.Direction.Cw);
 
+			//Android's path rendering logic rounds values down to the nearest int, make sure we round up here instead using the ViewHelper scaling logic
```
davidjohnoliver, line 38:
```suggestion
			//Android's path rendering logic rounds values down to the nearest int, make sure we round up here instead using the ViewHelper scaling logic. However we only want to round the height and width, not the frame offsets.
```

davidjohnoliver (approved):

PR #6506, opened 2021-07-19
fix(iOS): RadioButton Command when VoiceOver is On

ajpinedam:
fixes https://github.com/unoplatform/nventive-private/issues/231

## PR Type

- Bugfix
- 
On iOS When using RadioButton with VoiceOver on, the Command is not executed on a RadioButton button when the CheckedChanged.

## PR Checklist

- [] Validated PR `Screenshots Compare Test Run` results.

## Other information
**Note:** Please use the sample app attached to the Ticket to manually test this fix. Not possible to test using Unit Test/ UITest / 

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/Primitives/ToggleButton.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Primitives/ToggleButton.cs
@@ -124,7 +124,14 @@ namespace Windows.UI.Xaml.Controls.Primitives
 
 		internal void AutomationPeerToggle()
 		{
+			var oldValue = IsChecked;
 			OnToggle();
+			var newValue = IsChecked;
+
+			if ((!oldValue.HasValue || !oldValue.Value) && newValue == true)
+			{
+				InvokeCommand();
+			}
 		}
 
 		protected override AutomationPeer OnCreateAutomationPeer()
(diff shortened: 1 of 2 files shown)


jeromelaban (approved):
Only the manual test to add, and good to go!

src/Uno.UI/UI/Xaml/Controls/Primitives/ToggleButton.cs
```diff
@@ -124,7 +124,14 @@ protected virtual void OnToggle()
 
 		internal void AutomationPeerToggle()
 		{
+			var oldValue = IsChecked;
 			OnToggle();
+			var newValue = IsChecked;
+
+			if ((!oldValue.HasValue || !oldValue.Value) && newValue == true)
+			{
+				InvokeCommand();
```
davidjohnoliver, line 133: I'm not too familiar with the AutomationPeer mechanics, but is there any risk of the command being called twice? @ajpinedam I think you mentioned that without the fix, the command was still being raised for a particular sub-area of the touch target of the toggle button - would it get raised once or twice in that case?

ajpinedam: @davidjohnoliver Tested it and indeed we could see cases when the Command is executed twice. We will be merging this code as it and created a follow-up issue https://github.com/unoplatform/uno/issues/6535 . 

Also, @MartinZikmund  is working on this control and might be fixing this issue

PR #6618, opened 2021-08-01
feat: Add support for DataContext inheritance for non-UIElement

jeromelaban:
## PR Type

- Bugfix

## What is the new behavior?

The `DataContext` now propagates properly to non-`UIElement` `DependencyObject`-typed values in dependency properties. This behavior is generally used to data bind dependency objects in XAML for advanced configurations.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/DependencyObjectStore.Binder.cs
+++ b/src/Uno.UI/UI/Xaml/DependencyObjectStore.Binder.cs
@@ -96,7 +96,7 @@ namespace Windows.UI.Xaml
 
 		private void ApplyChildrenBindable(object? inheritedValue, bool isTemplatedParent)
 		{
-			void SetInherited(IDependencyObjectStoreProvider provider)
+			static void SetInherited(IDependencyObjectStoreProvider provider, object? inheritedValue, bool isTemplatedParent)
 			{
 				if (isTemplatedParent)
 				{
@@ -107,6 +107,7 @@ namespace Windows.UI.Xaml
 					provider.Store.SetInheritedDataContext(inheritedValue);
 				}
 			}
+
 			for (int i = 0; i < _childrenBindable.Count; i++)
 			{
 				var child = _childrenBindable[i];
@@ -130,7 +131,7 @@ namespace Windows.UI.Xaml
 
 				if (childAsStoreProvider != null)
 				{
-					SetInherited(childAsStoreProvider);
+					SetInherited(childAsStoreProvider, inheritedValue, isTemplatedParent);
 				}
 				else
 				{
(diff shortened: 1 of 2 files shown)


src/Uno.UI/UI/Xaml/DependencyObjectStore.Binder.cs
```diff
@@ -561,43 +562,69 @@ private void OnDependencyPropertyChanged(DependencyPropertyDetails propertyDetai
...
-		{
-			if (
-				propertyDetails.Property == _templatedParentProperty
-				|| propertyDetails.Property == _dataContextProperty
+		void GetPropertyInheritanceConfiguration(
+			DependencyPropertyDetails propertyDetails,
+			DependencyPropertyChangedEventArgs args,
+			out bool hasValueInherits,
+			out bool hasValueDoesNotInherit,
+			out IDependencyObjectStoreProvider? valueAsProvider
```
davidjohnoliver, line 604: I'm not really sure why this `out` parameter is needed? Seems like the `as` could be done just as easily in the calling method.

src/Uno.UI/UI/Xaml/DependencyObjectStore.Binder.cs
```diff
@@ -561,43 +562,69 @@ private void OnDependencyPropertyChanged(DependencyPropertyDetails propertyDetai
...
+			GetPropertyInheritanceConfiguration(
+				propertyDetails,
+				args,
+				hasValueInherits: out var hasValueInherits,
+				hasValueDoesNotInherit: out var hasValueDoesNotInherits,
+				valueAsProvider: out var newValueAsProvider
+			);
 
-			if (!hasValueDoesNotInherit && hasValueInherits)
+			if (!hasValueDoesNotInherits)
```
davidjohnoliver, line 573: This block of conditions starts to get quite hard to follow... I wonder if it'd be more readable with a nested method like:

```csharp
SetChildrenBindable(object? obj) => _childrenBindable[GetOrCreateChildBindablePropertyIndex(propertyDetails.Property)] = obj;
```

davidjohnoliver (approved):

src/Uno.UI/UI/Xaml/DependencyObjectStore.Binder.cs
jeromelaban: The reason for the out is to avoid casting twice, once inside the method, and another one outside if either of the two boolean not true, but also because it removes the use of a generic ValueTuple.

jeromelaban: Reconsidering it, multiple alterations actually don't need this anymore. Will remove.

PR #6630, opened 2021-08-02
Ensure that all uno controls defined dependency properties are using `FrameworkPropertyMetadata`

### LINKED ISSUE #5984: Ensure that all uno controls defined dependency properties are using `FrameworkPropertyMetadata`
Many of the controls defined inside of Uno need to have dependency properties are declared with `PropertyMetadata` instead of `FrameworkPropertyMetadata`. This generally causes issues like this https://github.com/unoplatform/uno/pull/5976.

## Note for implementers
- Changing this implies the creation of a runtime test or an analyzer that scans all dependency properties defined and checks if the metadata is of the right type
- Make changes to all properties metadata that are not of the right type

Youssef1313:
GitHub Issue (If applicable): closes #5984

## PR Type

- Bugfix
- Build or CI related changes

Usage of `FrameworkPropertyMetadata` isn't enforced over `PropertyMetadata` in CI, and some controls are not using `FrameworkPropertyMetadata`.

## What is the new behavior?

Enforce usage of `FrameworkPropertyMetadata` by introducing a new internal `Register` overload and ban the public overload.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/Microsoft/UI/Xaml/Controls/Expander/Expander.Properties.cs
+++ b/src/Uno.UI/Microsoft/UI/Xaml/Controls/Expander/Expander.Properties.cs
@@ -45,7 +45,7 @@ namespace Microsoft.UI.Xaml.Controls
 				nameof(HeaderTemplate),
 				typeof(DataTemplate),
 				typeof(Expander),
-				new FrameworkPropertyMetadata(null));
+				new FrameworkPropertyMetadata(null, FrameworkPropertyMetadataOptions.ValueDoesNotInheritDataContext));
 
 		public DataTemplateSelector HeaderTemplateSelector
 		{
--- a/src/Uno.UI/Microsoft/UI/Xaml/Controls/Expander/Expander.Properties.cs
+++ b/src/Uno.UI/Microsoft/UI/Xaml/Controls/Expander/Expander.Properties.cs
@@ -17,7 +17,7 @@ namespace Microsoft.UI.Xaml.Controls
 				nameof(ExpandDirection),
 				typeof(ExpandDirection),
 				typeof(Expander),
-				new PropertyMetadata(
+				new FrameworkPropertyMetadata(
 					ExpandDirection.Down,
 					(s, e) => (s as Expander)?.OnExpandDirectionPropertyChanged(e)));
 
@@ -32,7 +32,7 @@ namespace Microsoft.UI.Xaml.Controls
 				nameof(Header),
 				typeof(object),
 				typeof(Expander),
-				new PropertyMetadata(null));
+				new FrameworkPropertyMetadata(null));
 
 		public DataTemplate HeaderTemplate
 		{
@@ -45,7 +45,7 @@ namespace Microsoft.UI.Xaml.Controls
 				nameof(HeaderTemplate),
 				typeof(DataTemplate),
 				typeof(Expander),
-				new PropertyMetadata(null));
+				new FrameworkPropertyMetadata(null));
 
 		public DataTemplateSelector HeaderTemplateSelector
 		{
@@ -58,7 +58,7 @@ namespace Microsoft.UI.Xaml.Controls
 				nameof(HeaderTemplateSelector),
 				typeof(DataTemplateSelector),
 				typeof(Expander),
-				new PropertyMetadata(null));
+				new FrameworkPropertyMetadata(null));
 
 		public bool IsExpanded
 		{
@@ -71,7 +71,7 @@ namespace Microsoft.UI.Xaml.Controls
 				nameof(IsExpanded),
 				typeof(bool),
 				typeof(Expander),
-				new PropertyMetadata(
+				new FrameworkPropertyMetadata(
 					true,
 					(s, e) => (s as Expander)?.OnIsExpandedPropertyChanged(e)));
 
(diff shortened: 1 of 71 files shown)


Youssef1313 (commented):
For now, build should fail since I didn't fix all usage of `PropertyMetadata`. Before I move forward with fixing all usages (which is a bit tedious), I want an initial review on the approach I've taken and its correctness.

dr1rrb (commented):
I love the idea 😍, wondering though if we have an option to forcefully use a standard `PropertyMetadata` if needed?

Youssef1313:
> I love the idea 😍, wondering though if we have an option to forcefully use a standard `PropertyMetadata` if needed?

The diagnostic can be suppressed when using `PropertyMetadata` is intentional. I expect this should be rare, right?

Youssef1313:
@jeromelaban @dr1rrb I fixed the large portion of it. The remaining usages are in runtime tests and samples app. Should these remain the same and disable the diagnostic for them?

davidjohnoliver:
> @jeromelaban @dr1rrb I fixed the large portion of it. The remaining usages are in runtime tests and samples app. Should these remain the same and disable the diagnostic for them?
> 
> There is also RegisterAttached, we should do the same for it too right? If so, do you prefer it in the same PR or separate PR?

For the runtime tests and SamplesApp, those ones are also compiled on Windows so they should (and must) remain the same.

davidjohnoliver:
> Thanks @davidjohnoliver. Could someone look at the failure in CI? Is it a known flakiness or something related to my changes?

This one is not just run-of-the-mill flakiness:

**Uno.UI - CI (Unit Tests CSharp9)**
> RunMessage : The active test run was aborted. Reason: Test host process crashed : Process is terminated due to StackOverflowException.

Sounds related to this warning from the code analysis:

image

Youssef1313:
@jeromelaban I think this is ready for review.

MartinZikmund:
@Youssef1313 just a tip - I see you use "fix:" for commits fixing bugs in the PR scope, but it is better to just use "chore:" for these. "fix:" should be used for bug fixes for issues that are already in master. It is useful to avoid using "fix:" for these incremental adjustments, as we then use the prefixes to generate changelog, and this way the "fix" commits will show up under bug fixes.

Youssef1313:
Thanks for the tip @MartinZikmund. I squashed and changed the commit message prefix to "chore:"

MartinZikmund:
@Youssef1313 sorry, maybe I wrote it wrong - the initial commit would be "fix:" - the metadata is wrong in master and it fixes this. But the subsequent commits like fixing the StackOverflow would then be "chore" as they just fix an issue introduced by a previous commit in the PR 👍

Youssef1313:
Thanks again for the clarification!

src/Uno.UI/Microsoft/UI/Xaml/Controls/NavigationView/NavigationView.Properties.cs
```diff
@@ -185,7 +185,7 @@ public DataTemplate HeaderTemplate
 		/// Identifies the HeaderTemplate dependency property.
 		/// </summary>
 		public static DependencyProperty HeaderTemplateProperty { get; } =
-			DependencyProperty.Register(nameof(HeaderTemplate), typeof(DataTemplate), typeof(NavigationView), new PropertyMetadata(null, OnPropertyChanged));
+			DependencyProperty.Register(nameof(HeaderTemplate), typeof(DataTemplate), typeof(NavigationView), new FrameworkPropertyMetadata(null, OnPropertyChanged));
```
davidjohnoliver, line 188: Be careful, DPs of type `DataTemplate` or `Style` should be marked `FrameworkPropertyMetadataOptions.ValueDoesNotInheritDataContext`.

Youssef1313: Thanks for noting, I'll fix that.

Youssef1313: Fixed. Note that some existing properties in master were already using `FrameworkPropertyMetadata` but not specifying the option. Example from master:

https://github.com/unoplatform/uno/blob/890207ec33affe032e5251ddc2276704520ba63b/src/Uno.UI/UI/Xaml/Controls/DatePicker/DatePicker.Properties.cs#L175-L180

Youssef1313: Also note that this PR only handles `DependencyProperty.Register`, but not `RegisterAttached`. If both should be handled, please let me know whether you want it in the same PR or in a separate PR. Thanks!

davidjohnoliver: I think it should be the same for `RegisterAttached()`, but it probably makes sense to put it in a separate PR and close this one off.

davidjohnoliver (approved):

src/Uno.UI/Microsoft/UI/Xaml/Controls/Expander/Expander.Properties.cs
```diff
@@ -45,7 +45,7 @@ public DataTemplate HeaderTemplate
 				nameof(HeaderTemplate),
 				typeof(DataTemplate),
 				typeof(Expander),
-				new PropertyMetadata(null));
+				new FrameworkPropertyMetadata(null, FrameworkPropertyMetadataOptions.ValueDoesNotInheritDataContext));
```
jeromelaban, line 48: Maybe I missed some of the conversation, but why did you add this option?

Youssef1313: @jeromelaban This was requested in https://github.com/unoplatform/uno/pull/6630#discussion_r685865886

jeromelaban: Thanks!

PR #6664, opened 2021-08-04
fix: Cleanup few warnings

Youssef1313:
The change doesn't affect the product. Just few compile-time warnings.


src/SamplesApp/SamplesApp.UITests/TestFramework/ImageAssert.ExpectedPixels.cs
```diff
@@ -83,7 +83,7 @@ public ExpectedPixels Pixels(Bitmap source, Rectangle rect)
 
 				return new ExpectedPixels(Name, Location, rect.Location, colors, Tolerance);
 			}
-			catch (Exception ex)
```
jeromelaban, line 86: Let's pass the exception as the inner parameter

jeromelaban (approved):

davidjohnoliver (approved):

PR #6571, opened 2021-07-27
fix(menuflyout): Position MenuFlyout correctly with native AppBarButton

davidjohnoliver:
GitHub Issue (If applicable): fixes https://github.com/unoplatform/nventive-private/issues/232

When native AppBarButton style is being used in conjunction with XAML MenuFlyout style, ensure that the native app bar button view is used as anchor when positioning the flyout, because the managed AppBarButton is not actually in the visual tree.

Note: if MenuFlyout.UseNativePopup is set to "True" (native flyout style), the positioning is already correctly handled.

## PR Type

What kind of change does this PR introduce?

- Bugfix

## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)
- [x] [Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [ ] Validated PR `Screenshots Compare Test Run` results.
- [x] Contains **NO** breaking changes
- [x] Associated with an issue (GitHub or internal) and uses the [automatic close keywords](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue).
- [x] Commits must be following the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) specification.

## Other information

Internal Issue (If applicable):

PR #6673, opened 2021-08-05
fix(appbarbutton): [iOS] Fix AppBarButton icon sometimes disappearing

davidjohnoliver:
Fixes bug which occurs in narrow circumstances* where AppBarButton icon would disappear after navigating away from page then back.

*The bug occurs when properties of the AppBarButton are modified, causing Render() to be called, after navigating away; setting CustomView to null triggers a native exception similar to this one: https://stackoverflow.com/questions/53219042/self-navigationitem-rightbarbuttonitem-customview-nil-crash

The bug surfaced after recent changes to MenuFlyout (https://github.com/unoplatform/uno/pull/6571), but seems more a timing issue rather than directly related to those changes. The bug didn't occur in a standalone repro mimicking the configuration of the app for which it was reported.

GitHub Issue (If applicable): closes https://github.com/unoplatform/nventive-private/issues/243

## PR Type

- Bugfix

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/Extensions/UIBarButtonItemExtensions.iOS.cs
+++ b/src/Uno.UI/Extensions/UIBarButtonItemExtensions.iOS.cs
@@ -1,0 +1,23 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Uno.Extensions;
+using UIKit;
+
+namespace Uno.UI.Extensions
+{
+	internal static class UIBarButtonItemExtensions
+	{
+		/// <summary>
+		/// In some narrow circumstances (eg when the page is unloaded, but not always) setting CustomView to null can trigger a native exception,
+		/// even if it's already null. To avoid this, skip redundantly setting it to null.
+		/// </summary>
+		public static void ClearCustomView(this UIBarButtonItem barButtonItem)
+		{
+			if (barButtonItem?.CustomView != null)
+			{
+				barButtonItem.CustomView = null;
+			}
+		}
+	}
+}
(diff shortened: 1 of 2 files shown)


src/Uno.UI/Extensions/UIBarButtonItemExtensions.iOS.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Text;
+using Uno.Extensions;
+using UIKit;
+
+namespace Uno.UI.Extensions
+{
+	internal static class UIBarButtonItemExtensions
+	{
+		/// <summary>
+		///
```
ajpinedam, line 12: Missing the comments?

ajpinedam (approved):

src/Uno.UI/Extensions/UIBarButtonItemExtensions.iOS.cs
jeromelaban: Also, why did you choose to make it an extension instead of a method in the renderer ?

davidjohnoliver: @ajpinedam Good catch, I added the comment.

@jeromelaban No reason other than it made sense to me, I can move it to the renderer if it bothers you.

jeromelaban: That's fine, let's keep it there.

jeromelaban (approved):

PR #6681, opened 2021-08-06
feat: Implement ScrollViewer.ViewChanging

### LINKED ISSUE #78: ScrollViewer.ViewChanging doesn't fire on android
<ScrollViewer VerticalScrollMode="Enabled" HorizontalScrollMode="Enabled"
        ViewChanged="ScrollViewer_ViewChanged" ViewChanging="ScrollViewer_ViewChanging">
        <Border Height="1000" Width="1000" Background="Red"/>
    </ScrollViewer>

The ScrollViewer_ViewChanged fires, the ScrollViewer_ViewChanging doesn't. 
Tried with 1.31.0-dev.91
[Repo inside](https://github.com/Henk7/Uno.QuickStart/commits/scrollviewer)

Youssef1313:
*PR is still early to be ready. Opening as a draft for now*

GitHub Issue (If applicable): closes #78

## PR Type

- Feature

ViewChanging is not implemented.

## What is the new behavior?

It's now implemented.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/Generated/3.0.0.0/Windows.UI.Xaml.Controls/ScrollViewer.cs
+++ b/src/Uno.UI/Generated/3.0.0.0/Windows.UI.Xaml.Controls/ScrollViewer.cs
@@ -769,22 +769,7 @@ namespace Windows.UI.Xaml.Controls
 		// Skipping already declared method Windows.UI.Xaml.Controls.ScrollViewer.GetBringIntoViewOnFocusChange(Windows.UI.Xaml.DependencyObject)
 		// Skipping already declared method Windows.UI.Xaml.Controls.ScrollViewer.SetBringIntoViewOnFocusChange(Windows.UI.Xaml.DependencyObject, bool)
 		// Skipping already declared event Windows.UI.Xaml.Controls.ScrollViewer.ViewChanged
-		#if __ANDROID__ || __IOS__ || NET461 || __WASM__ || __SKIA__ || __NETSTD_REFERENCE__ || __MACOS__
-		[global::Uno.NotImplemented("__ANDROID__", "__IOS__", "NET461", "__WASM__", "__SKIA__", "__NETSTD_REFERENCE__", "__MACOS__")]
-		public  event global::System.EventHandler<global::Windows.UI.Xaml.Controls.ScrollViewerViewChangingEventArgs> ViewChanging
-		{
-			[global::Uno.NotImplemented("__ANDROID__", "__IOS__", "NET461", "__WASM__", "__SKIA__", "__NETSTD_REFERENCE__", "__MACOS__")]
-			add
-			{
-				global::Windows.Foundation.Metadata.ApiInformation.TryRaiseNotImplemented("Windows.UI.Xaml.Controls.ScrollViewer", "event EventHandler<ScrollViewerViewChangingEventArgs> ScrollViewer.ViewChanging");
-			}
-			[global::Uno.NotImplemented("__ANDROID__", "__IOS__", "NET461", "__WASM__", "__SKIA__", "__NETSTD_REFERENCE__", "__MACOS__")]
-			remove
-			{
-				global::Windows.Foundation.Metadata.ApiInformation.TryRaiseNotImplemented("Windows.UI.Xaml.Controls.ScrollViewer", "event EventHandler<ScrollViewerViewChangingEventArgs> ScrollViewer.ViewChanging");
-			}
-		}
-		#endif
+		// Skipping already declared event Windows.UI.Xaml.Controls.ScrollViewer.ViewChanging
 		#if __ANDROID__ || __IOS__ || NET461 || __WASM__ || __SKIA__ || __NETSTD_REFERENCE__ || __MACOS__
 		[global::Uno.NotImplemented("__ANDROID__", "__IOS__", "NET461", "__WASM__", "__SKIA__", "__NETSTD_REFERENCE__", "__MACOS__")]
 		public  event global::System.EventHandler<object> DirectManipulationCompleted
--- a/src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.cs
@@ -96,6 +96,11 @@ namespace Windows.UI.Xaml.Controls
 		private static bool IsAnimationEnabled => Uno.UI.Helpers.WinUI.SharedHelpers.IsAnimationsEnabled();
 
 		/// <summary>
+		/// Occurs when manipulations such as scrolling and zooming cause the view to change.
+		/// </summary>
+		public event EventHandler<ScrollViewerViewChangingEventArgs>? ViewChanging;
+
+		/// <summary>
 		/// Occurs when manipulations such as scrolling and zooming have caused the view to change.
 		/// </summary>
 		public event EventHandler<ScrollViewerViewChangedEventArgs>? ViewChanged;
@@ -1298,6 +1303,7 @@ namespace Windows.UI.Xaml.Controls
 
 		private void Update(bool isIntermediate)
 		{
+			ViewChanging?.Invoke(this, new ScrollViewerViewChangingEventArgs(true, null, null)); // TODO: Fix values.
 			_hasPendingUpdate = false;
 
 			HorizontalOffset = _pendingHorizontalOffset;
@@ -1310,7 +1316,6 @@ namespace Windows.UI.Xaml.Controls
 			ScrollOffsets = new Point(_pendingHorizontalOffset, _pendingVerticalOffset);
 			InvalidateViewport();
 #endif
-
 			ViewChanged?.Invoke(this, new ScrollViewerViewChangedEventArgs { IsIntermediate = isIntermediate });
 		}
 
--- a/src/Uno.UI/Uno.UI.csproj
+++ b/src/Uno.UI/Uno.UI.csproj
@@ -175,6 +175,10 @@
 	</ItemGroup>
 
 	<ItemGroup>
+	  <UpToDateCheckInput Remove="UI\Xaml\Controls\ScrollViewer\ScrollViewerViewChangingEventArgs.cs" />
+	</ItemGroup>
+
+	<ItemGroup>
 		<Compile Include="Mixins\Android\BaseActivity.Callbacks.g.cs" />
 		<Compile Include="Mixins\Android\FrameworkElementMixins.g.cs" />
 		<Compile Include="Mixins\Android\FrameworkElementMixins.g.cs" />
--- a/src/Uno.UI/Generated/3.0.0.0/Windows.UI.Xaml.Controls/ScrollViewerViewChangingEventArgs.cs
+++ b/src/Uno.UI/Generated/3.0.0.0/Windows.UI.Xaml.Controls/ScrollViewerViewChangingEventArgs.cs
@@ -2,43 +2,10 @@
 #pragma warning disable 114 // new keyword hiding
 namespace Windows.UI.Xaml.Controls
 {
-	#if __ANDROID__ || __IOS__ || NET461 || __WASM__ || __SKIA__ || __NETSTD_REFERENCE__ || __MACOS__
+	#if false || false || false || false || false || false || false
 	[global::Uno.NotImplemented]
 	#endif
 	public  partial class ScrollViewerViewChangingEventArgs 
 	{
-		#if __ANDROID__ || __IOS__ || NET461 || __WASM__ || __SKIA__ || __NETSTD_REFERENCE__ || __MACOS__
-		[global::Uno.NotImplemented("__ANDROID__", "__IOS__", "NET461", "__WASM__", "__SKIA__", "__NETSTD_REFERENCE__", "__MACOS__")]
-		public  global::Windows.UI.Xaml.Controls.ScrollViewerView FinalView
-		{
-			get
-			{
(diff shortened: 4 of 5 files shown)


src/Uno.UI/Generated/3.0.0.0/Windows.UI.Xaml.Controls/ScrollViewer.cs
```diff
@@ -769,22 +769,7 @@ public static void SetIsDeferredScrollingEnabled( global::Windows.UI.Xaml.Depend
...
-				global::Windows.Foundation.Metadata.ApiInformation.TryRaiseNotImplemented("Windows.UI.Xaml.Controls.ScrollViewer", "event EventHandler<ScrollViewerViewChangingEventArgs> ScrollViewer.ViewChanging");
-			}
-			[global::Uno.NotImplemented("__ANDROID__", "__IOS__", "NET461", "__WASM__", "__SKIA__", "__NETSTD_REFERENCE__", "__MACOS__")]
-			remove
-			{
-				global::Windows.Foundation.Metadata.ApiInformation.TryRaiseNotImplemented("Windows.UI.Xaml.Controls.ScrollViewer", "event EventHandler<ScrollViewerViewChangingEventArgs> ScrollViewer.ViewChanging");
-			}
-		}
-		#endif
+		// Skipping already declared event Windows.UI.Xaml.Controls.ScrollViewer.ViewChanging
```
Youssef1313, line 772: Change here is made manually. I don't know how to run the generator that generates this file? Building any of the solution filters doesn't seem to do that.

src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.cs
```diff
@@ -1298,6 +1303,7 @@ private void RequestUpdate()
 
 		private void Update(bool isIntermediate)
 		{
+			ViewChanging?.Invoke(this, new ScrollViewerViewChangingEventArgs(true, null, null)); // TODO: Fix values.
```
Youssef1313, line 1306: This is the part where I'm struggling.

From UWP docs:

> `IsInertial` is **true** for manipulations that occur:
>
> - using a mouse wheel.
> - using key strokes on `ScrollViewer` and other controls that support direct manipulation.
> - by calling `ChangeView` with animation enabled.
>
> `IsInertial` is **false** for manipulations that occur:
>
> - by dragging the Thumb on a scrollbar.
> - using key strokes on controls that don't support direct manipulation.
> - by calling ChangeView with animation disabled.

What's the equivalent of `mouse wheel` vs `dragging` in mobile apps? How to access this information.

---

I don't get what `FinalView` and `NextView` are from the docs:

https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.scrollviewerviewchangingeventargs.finalview?view=winrt-20348

https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.scrollviewerviewchangingeventargs.nextview?view=winrt-20348

src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.cs
```diff
@@ -1310,7 +1316,6 @@ private void Update(bool isIntermediate)
 			ScrollOffsets = new Point(_pendingHorizontalOffset, _pendingVerticalOffset);
 			InvalidateViewport();
 #endif
-
 			ViewChanged?.Invoke(this, new ScrollViewerViewChangedEventArgs { IsIntermediate = isIntermediate });
```
Youssef1313, line 1319: fwiw, this doesn't match the UWP API. The property shouldn't have a public setter. This is unrelated to the PR. Let me know if you want a separate tracking issue for that.

src/Uno.UI/Uno.UI.csproj
```diff
@@ -174,6 +174,10 @@
 		</ProjectReference>
 	</ItemGroup>
 
+	<ItemGroup>
+	  <UpToDateCheckInput Remove="UI\Xaml\Controls\ScrollViewer\ScrollViewerViewChangingEventArgs.cs" />
+	</ItemGroup>
```
Youssef1313, line 179: This was auto-generated by Visual Studio. I'm not sure if this is needed.

Youssef1313 (commented):
cc @jeromelaban @MartinZikmund for help here. Thanks in advance!

src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.cs
MartinZikmund: @dr1rrb will probably know the answer 🙂

src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.cs
MartinZikmund: Which property please?

src/Uno.UI/Uno.UI.csproj
MartinZikmund: Yes, VS often generates these, you can undo this change

src/Uno.UI/Generated/3.0.0.0/Windows.UI.Xaml.Controls/ScrollViewer.cs
MartinZikmund: There is UWPSyncGenerator which can do this, but we usually don't run it for single changes like this, it is usually enough to adjust the #if and `[NotImplemented]` accordingly

src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.cs
Youssef1313: IsIntermediate

MartinZikmund: I see, you can make the setter `internal` then, as it will match UWP. It is a breaking change though, so you will need to update `PackageDiffIgnore.xml` accordingly. You can for now just make it internal and commit the change, and the build will tell you specifically how to update that file.

Youssef1313: I think this might be suitable for a separate PR since this is not part of `ViewChanging`. Going to open another issue

src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.cs
davidjohnoliver: btw if you want to test UWP touch behavior, deploying the UWP head to 'Simulator' instead of 'Local Machine' is useful

dr1rrb: > What's the equivalent of `mouse wheel` vs `dragging` in mobile apps? How to access this information.

There is no real equivalent to _mouse wheel_ on mobile apps, unless you connect a mouse to your device (on Android I mean ... mouses are poorly handled like a touch pointer on iOS, at OS level).

I would put the `IsInertial` in parallel of the [`ViewChanged.IsIntermediate`](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.scrollviewerviewchangedeventargs.isintermediate?view=winrt-20348#Windows_UI_Xaml_Controls_ScrollViewerViewChangedEventArgs_IsIntermediate) which only indicates if the event is the final one or not, so we can avoid to do useless work and preserve resources for a more performant scrolling experience.

From my understanding, the `IsInertial` means that the user interaction is already over (key up / wheel events are already over), so we already know where the scrolling is going to stop, but the `ScrollViewer` is going to animate that change. On the other side, when dragging (the content -touch/pen- or the scrollbars -mouse/pen-), we don't know yet where the scrolling is going to stop. The application can then adjust its logic to get a more responsive scrolling (like triggering the loading of the element that is going to be visible). I would however confirm that with touch we do have `IsInertial==false` when dragging and when flicking, it's only an assumption.

And about where to get the info: Well you are not at the right place here, the view has already changed, and the info has already been lost. You have to raise that event at each place which can trigger the scrolling.
* For programmatic scrolling, you can put it right before the [`ChangeViewNative`](https://github.com/unoplatform/uno/blob/master/src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.cs#L1385). The `disableAnimation` flag is already accessible here.
* For **MANAGED** `ScrollBar` (WASM, Skia, macOS) I would assume it will be already correct as the end in the `ChangeViewCore` (FYI cf. [`On<Vertical|Horizontal>ScrollBarScrolled`](https://github.com/unoplatform/uno/blob/4029e71b5da907ca53fe7e85d0590b5ca533b786/src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.cs#L1158)) 
* For **MANAGED** `ScrollContentPresenter` (Skia, macOS) you will have to look at the [`ScrollContentPresenter_PointerWheelChanged`](https://github.com/unoplatform/uno/blob/4029e71b5da907ca53fe7e85d0590b5ca533b786/src/Uno.UI/UI/Xaml/Controls/ScrollContentPresenter/ScrollContentPresenter.Managed.cs#L212) and [`BeginTouchScroll`](https://github.com/unoplatform/uno/blob/4029e71b5da907ca53fe7e85d0590b5ca533b786/src/Uno.UI/UI/Xaml/Controls/ScrollContentPresenter/ScrollContentPresenter.Managed.cs#L252)
* And the trickiest part, for **NATIVE** `ScrollContentPresenter` (WASM, iOS, Android) you will have to register to native event of each platform:
    * iOS: https://developer.apple.com/documentation/uikit/uiscrollviewdelegate/1619394-scrollviewwillbegindragging but I assume that you won't be able to make a distinction between input kinds.
    * Android: 
        * keyboard: https://github.com/unoplatform/uno/blob/4029e71b5da907ca53fe7e85d0590b5ca533b786/src/Uno.UI.BindingHelper.Android/Uno/UI/UnoTwoDScrollView.java#L318
        * touch: https://github.com/unoplatform/uno/blob/4029e71b5da907ca53fe7e85d0590b5ca533b786/src/Uno.UI.BindingHelper.Android/Uno/UI/UnoTwoDScrollView.java#L437
        * wheel: ???
     * WASM: ??? (I see only the `scroll` event, but I think that it's raised only after the scroll has began, and browsers have started to enable async scrolling so it will become less and less relevant)

Even if not equivalent to UWP, I think that we could accept to raise the `ViewChanging` right before the `ViewChanged`, no matter id the UI has already been updated, but I would try to do that only where we are not able to get it at the right time, and in that case I think we won't be able to provide the right info.

> 
> I don't get what `FinalView` and `NextView` are from the docs:
> 

From my understanding the `NextView` is the view which will be visible (at the top/center/bottom of the viewport?) once the scroll has occured and the `ViewChangedEvent` corresponding to that `ViewChanging` event has been raised, no matter if the scrolling is still pending (i.e. while dragging) or animating (i.e. inertia). The `FinalView` is the view that is going to be visible once scrolling has stopped, including inertia. So I assume that this will stay `null` while dragging as we cannot yet know where the scroll will end, but it will be full-filled while animating as we already know where the scroll will stop.

Youssef1313: Thanks a lot @dr1rrb. This is very useful. Couple of points:


- In UWP, Looks like `FinalView` is not null when `IsInertial == true`

    image

    Seems it's always going to be equals to `NextView` (at least based on what I've seen so far).

- Regarding the value of `IsInertial`, it seems unrelated to `IsIntermediate` (`IsIntermediate` can be true for both mouse wheel and scrollbar - while `IsInertial` is true for mousewheel but false for scrollbar)

Youssef1313: Also from my experiments with UWP, it looks like ViewChanging can be fired multiple times (while animating) before ViewChanged is fired. ie, Not every ViewChanging is followed by ViewChanged, only the final one. The code path in `ChangeViewCore` and `Update` doesn't seem to agree with that behavior (tested on Wasm).

How far can we diverge from UWP behavior?

dr1rrb: > 
> 
> Thanks a lot @dr1rrb. This is very useful. Couple of points:
> 
>     * In UWP, Looks like `FinalView` is not null when `IsInertial == true`
>       Seems it's always going to be equals to `NextView` (at least based on what I've seen so far).

You should try to put a list of elements that are just a bit slower than the viewport, then request programmatically a `ChangeView` to the last one with animations enabled. I'm pretty sure that the `NextView` won't be directly the `FinalView`. 
> 
>     * Regarding the value of `IsInertial`, it seems unrelated to `IsIntermediate` (`IsIntermediate` can be true for both mouse wheel and scrollbar - while `IsInertial` is true for mousewheel but false for scrollbar)

Yeah by "comparable", I mean in term of usage: both are used to indicate into which step the `ScrollViewer` is, so applications can enhance their behavior.

> Also from my experiments with UWP, it looks like ViewChanging can be fired multiple times (while animating) before ViewChanged is fired. ie, Not every ViewChanging is followed by ViewChanged, only the final one. 

The `ScrollViewer` is a really weird control, and is highly optimized. The `ViewChanged` (and most probably the `ViewChanging`) is raised in a "best effort" way: it does not wait for your handler to continue / complete scrolling. It will raise as much event as it can, and as far as I know it does raise multiple `ViewChanged` on UWP also. You should use logging, and not breakpoints (i would not even recommend tracepoint as they are pretty slow).

PR #6726, opened 2021-08-11
fix: Fix image flickering

### LINKED ISSUE #4473: [WASM] BitmapIcon blinking invalid image on hover
## Current behavior

blinking2

## Expected behavior

Should not blink.

```
<Button
    Grid.RowSpan="2"
    Grid.ColumnSpan="2"
    HorizontalAlignment="Right"
    VerticalAlignment="Bottom"
    Background="Transparent"
    BorderBrush="Transparent"
    Command="{x:Bind ViewModel.OpenSettingsCommand}">
    <BitmapIcon
        Width="48"
        Height="48"
        ShowAsMonochrome="false"
        UriSource="/Assets/Settings.png" />
</Button>
```

## Workaround

N/A

## Environment

Nuget Package:

- [ ] Other: 

Nuget Package Version(s): 3.2.0-dev.401

Affected platform(s):

- [ ] iOS
- [ ] Android
- [x] WebAssembly
- [ ] macOS
- [ ] Skia
  - [ ] WPF
  - [ ] GTK (Linux)
  - [ ] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

IDE:

- [ ] Rider macOS (version: )

Youssef1313:
GitHub Issue (If applicable): closes #4473

## PR Type

- Bugfix

## PR Checklist

## Other information

### DIFF
--- a/src/SamplesApp/SamplesApp.UITests/Windows_UI_Xaml_Controls/BitmapIconTests/UnoSamples_Tests.BitmapIcon.cs
+++ b/src/SamplesApp/SamplesApp.UITests/Windows_UI_Xaml_Controls/BitmapIconTests/UnoSamples_Tests.BitmapIcon.cs
@@ -15,23 +15,6 @@ namespace SamplesApp.UITests.Windows_UI_Xaml_Controls.BitmapIconTests
 	[TestFixture]
 	public partial class BitmapIcon_Tests : SampleControlUITestBase
 	{
-		private bool ContainsColor(Bitmap bitmap, Rectangle rect, Color color)
-		{
-			for (var x = rect.Left; x < rect.Right; x++)
-			{
-				for (var y = rect.Top; y < rect.Bottom; y++)
-				{
-					var pixel = bitmap.GetPixel(x, y);
-					if ((pixel.A, pixel.R, pixel.G, pixel.B) == (color.A, color.R, color.G, color.B))
-					{
-						return true;
-					}
-				}
-			}
-
-			return false;
-		}
-
 		[Test]
 		[AutoRetry]
 		public void When_BitmapIcon_Generic()
@@ -45,16 +28,15 @@ namespace SamplesApp.UITests.Windows_UI_Xaml_Controls.BitmapIconTests
 
 			using var initial = TakeScreenshot("Initial");
 			
-			var bitmap = initial.GetBitmap();
-			Assert.IsTrue(ContainsColor(bitmap, icon1.Rect.ToRectangle(), Color.Red));
-			Assert.IsTrue(ContainsColor(bitmap, icon2.Rect.ToRectangle(), Color.Blue));
+			ImageAssert.HasColorInRectangle(initial, icon1.Rect.ToRectangle(), Color.Red);
+			ImageAssert.HasColorInRectangle(initial, icon2.Rect.ToRectangle(), Color.Blue);
 
 			_app.FastTap(colorChange);
 
 			using var afterColorChange = TakeScreenshot("Changed");
-			bitmap = afterColorChange.GetBitmap();
-			Assert.IsTrue(ContainsColor(bitmap, icon1.Rect.ToRectangle(), Color.Yellow));
-			Assert.IsTrue(ContainsColor(bitmap, icon2.Rect.ToRectangle(), Color.Green));
+
+			ImageAssert.HasColorInRectangle(afterColorChange, icon1.Rect.ToRectangle(), Color.Yellow);
+			ImageAssert.HasColorInRectangle(afterColorChange, icon2.Rect.ToRectangle(), Color.Green);
 
 			_app.WaitForDependencyPropertyValue(_app.Marked("icon1"), "Foreground", "[SolidColorBrush #FFFFFF00]");
 			_app.WaitForDependencyPropertyValue(_app.Marked("icon2"), "Foreground", "[SolidColorBrush #FF008000]");
--- a/src/SamplesApp/SamplesApp.UITests/TestFramework/ImageAssert.cs
+++ b/src/SamplesApp/SamplesApp.UITests/TestFramework/ImageAssert.cs
@@ -222,6 +222,27 @@ namespace SamplesApp.UITests.TestFramework
 		public static void HasColorAt(ScreenshotInfo screenshot, float x, float y, Color expectedColor, byte tolerance = 0, [CallerLineNumber] int line = 0)
 			=> HasColorAtImpl(screenshot, (int)x, (int)y, expectedColor, tolerance, line);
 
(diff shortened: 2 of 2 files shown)


src/Uno.UI/UI/Xaml/Controls/BitmapIcon.cs
```diff
@@ -67,14 +67,6 @@ private void RefreshImage()
...
 #if !NET461
 			UpdateImageMonochromeColor();
-
-			if (UriSource != null)
-			{
-				// Force a reload
-				var tmp = UriSource;
-				UriSource = null;
-				UriSource = tmp;
-			}
```
jeromelaban, line 77: This part is needed for reloading on other targets. We may need to adjust this specifically for webassembly.

jeromelaban (changes requested):

Youssef1313:
@jeromelaban Is anything still missing here?

jeromelaban:
@Youssef1313 This is likely to not be enough. In this case, if we have a change in "ShowAsMonochrome" the image won't be refreshed properly. It's likely that the image refreshes itself because the foreground color has changed, something we do not need to do if we don't use the `ShowAsMonochrome` property.

Changing this would still provoque a flicker when using the `ShowAsMonochrome`. We should add additional UI tests to validate that changes to the `Foreground` and `ShowAsMonochrome` are actually updating the image accordingly.

Youssef1313:
There are already tests for monochrome images: https://github.com/unoplatform/uno/blob/master/src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/BitmapIconTests/BitmapIcon_Generic.xaml

Wondering how they're passing if my change is supposed to break them.

Youssef1313:
I also just validated the behavior of [the existing test](https://github.com/unoplatform/uno/blob/master/src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/BitmapIconTests/BitmapIcon_Generic.xaml) with and without my change. Changes to Foreground update the image correctly in both cases.

jeromelaban:
Good, then I wonder what's the source of the refresh then, if it's not the foreground change. Do you know which part does this?

Youssef1313:
I don't :(

btw, just a clarification to my previous comment: the flickering I'm referring to is when clicking the "change color" button in the monochrome test, while the one in the issue worked on for this PR happens when hovering.

I wonder if it's okay to keep the current change as is until we get any feedback that it broke something? or if someone knows a particular issue can be caused by this change?

Youssef1313:
Would someone be able to help here? cc @jeromelaban @carldebilly

MartinZikmund:
As long as the monochrome foreground color change still updates properly, it should be ok.

MartinZikmund (approved):

jeromelaban:
I've been looking again at this PR, and I don't see how this can work. `_image.MonochromeColor` is only read when the Uri changes, which means that removing cycling through null should also break the Foreground color change. Let's restart a build to take a look at the sample mentioned above, it's likely that we're missing something.

jeromelaban:
I just confirmed, the color [of this test](https://github.com/unoplatform/uno/blob/master/src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/BitmapIconTests/BitmapIcon_Generic.xaml.cs) does not change with the reload commented out.

The original issue was about the fact that the icon was reloaded. We probably should be avoiding the reload if the foreground changes, but icon is not monochrome.

Youssef1313 (commented):
Thanks @jeromelaban. Can you take a look at the new approach? I'm now forcing a reload without temporarily setting `UriSource` to null.

jeromelaban:
> Thanks @jeromelaban. Can you take a look at the new approach? I'm now forcing a reload without temporarily setting `UriSource` to null.

This looks a lot better, thanks :) This PR has shown something though, which is that the validation of the color changing is not done properly in the UI tests. This test:

https://github.com/unoplatform/uno/blob/ece6481aec938528b3420fe6e9322584436787c3/src/SamplesApp/SamplesApp.UITests/Windows_UI_Xaml_Controls/BitmapIconTests/UnoSamples_Tests.BitmapIcon.cs

Should also validate that the color is actually changed visually (searching for the new color in the new `Image` should do).

Youssef1313 (commented):
I'm seeing the following exception when running UI Tests:

```csharp
    OneTimeSetUp: System.TypeInitializationException : The type initializer for 'SamplesApp.UITests.SampleControlUITestBase' threw an exception.
      ----> System.Exception : Failed to execute: C:\Program Files (x86)\Android\android-sdk\platform-tools\adb.exe -s emulator-5554 forward tcp:56863 tcp:56863 - exit code: 1
    adb.exe: error: cannot bind listener: cannot bind to 127.0.0.1:56863: An attempt was made to access a socket in a way forbidden by its access permissions. (10013)

```

Have been encountering that since yesterday. @jeromelaban Any idea about that?

MartinZikmund:
@Youssef1313 did you try rebooting PC 🙂 ?

Youssef1313:
Thanks @jeromelaban @MartinZikmund! Fixed :)

src/SamplesApp/SamplesApp.UITests/Windows_UI_Xaml_Controls/BitmapIconTests/UnoSamples_Tests.BitmapIcon.cs
```diff
@@ -14,28 +15,49 @@ namespace SamplesApp.UITests.Windows_UI_Xaml_Controls.BitmapIconTests
...
+					var pixel = bitmap.GetPixel(x, y);
+					if ((pixel.A, pixel.R, pixel.G, pixel.B) == (color.A, color.R, color.G, color.B))
+					{
+						return true;
+					}
+				}
+			}
+
+			return false;
+		}
```
jeromelaban, line 33: ImageHelper likely contains this method.

Youssef1313: @jeromelaban I couldn't find it anywhere.

jeromelaban: Then let's add it in there then, if nothing matches the behavior you're adding: https://github.com/unoplatform/uno/blob/master/src/SamplesApp/SamplesApp.UITests/TestFramework/ImageAssert.cs

jeromelaban (approved):

Youssef1313:
@jeromelaban Anything left here?

PR #6772, opened 2021-08-14
perf: Avoid multiple JS calls when updating styles

### LINKED ISSUE #2196: [Wasm] Improve `TextBlock` fast path performance with style batching

Following https://github.com/unoplatform/uno/pull/2195, the updates made are still making many JS calls to update a TextBlock properties to the corresponding HTML element. 

Style updates should be batched as a single JS optimized call.

## For which Platform:

- [ ] iOS
- [ ] Android
- [x] WebAssembly
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

Youssef1313:
GitHub Issue (If applicable): closes #2196

## PR Type

- Performance

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.wasm.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.wasm.cs
@@ -68,21 +68,24 @@ namespace Windows.UI.Xaml.Controls
 
 		private void SynchronizeHtmlParagraphAttributes()
 		{
-			ConditionalUpdate(ref _fontStyleChanged, () => this.SetFontStyle(FontStyle));
-			ConditionalUpdate(ref _fontWeightChanged, () => this.SetFontWeight(FontWeight));
-			ConditionalUpdate(ref _fontFamilyChanged, () => this.SetFontFamily(FontFamily));
-			ConditionalUpdate(ref _fontSizeChanged, () => this.SetFontSize(FontSize));
-			ConditionalUpdate(ref _maxLinesChanged, () => this.SetMaxLines(MaxLines));
-			ConditionalUpdate(ref _textAlignmentChanged, () => this.SetTextAlignment(TextAlignment));
-			ConditionalUpdate(ref _lineHeightChanged, () => this.SetLineHeight(LineHeight));
-			ConditionalUpdate(ref _characterSpacingChanged, () => this.SetCharacterSpacing(CharacterSpacing));
-			ConditionalUpdate(ref _textDecorationsChanged, () => this.SetTextDecorations(TextDecorations));
-			ConditionalUpdate(ref _paddingChangedChanged, () => this.SetTextPadding(Padding));
-
-			if(_textTrimmingChanged || _textWrappingChanged)
+			using (SuspendCssStyleUpdates())
 			{
-				_textTrimmingChanged = _textWrappingChanged = false;
-				this.SetTextWrappingAndTrimming(textTrimming: TextTrimming, textWrapping: TextWrapping);
+				ConditionalUpdate(ref _fontStyleChanged, () => this.SetFontStyle(FontStyle));
+				ConditionalUpdate(ref _fontWeightChanged, () => this.SetFontWeight(FontWeight));
+				ConditionalUpdate(ref _fontFamilyChanged, () => this.SetFontFamily(FontFamily));
+				ConditionalUpdate(ref _fontSizeChanged, () => this.SetFontSize(FontSize));
+				ConditionalUpdate(ref _maxLinesChanged, () => this.SetMaxLines(MaxLines));
+				ConditionalUpdate(ref _textAlignmentChanged, () => this.SetTextAlignment(TextAlignment));
+				ConditionalUpdate(ref _lineHeightChanged, () => this.SetLineHeight(LineHeight));
+				ConditionalUpdate(ref _characterSpacingChanged, () => this.SetCharacterSpacing(CharacterSpacing));
+				ConditionalUpdate(ref _textDecorationsChanged, () => this.SetTextDecorations(TextDecorations));
+				ConditionalUpdate(ref _paddingChangedChanged, () => this.SetTextPadding(Padding));
+
+				if (_textTrimmingChanged || _textWrappingChanged)
+				{
+					_textTrimmingChanged = _textWrappingChanged = false;
+					this.SetTextWrappingAndTrimming(textTrimming: TextTrimming, textWrapping: TextWrapping);
+				}
 			}
 
 			if (_textChanged)
--- a/src/Uno.UI/UI/Xaml/UIElement.wasm.cs
+++ b/src/Uno.UI/UI/Xaml/UIElement.wasm.cs
@@ -27,6 +27,75 @@ namespace Windows.UI.Xaml
 		internal const string DefaultHtmlTag = "div";
 
 		private readonly GCHandle _gcHandle;
+		private bool _isCssStyleUpdatesSuspended = false;
+
+		/// <summary>
+		/// Since there is no way to apply all set and reset updates in bulk, we create them of groups and execute
+		/// each group separately to maintain the desired order.
+		/// </summary>
+		private readonly Queue<IQueuedUpdatesGroup> _queuedUpdates = new Queue<IQueuedUpdatesGroup>();
+
+		internal interface IQueuedUpdatesGroup
+		{
+			void Execute();
+		}
+
+		internal class SetStylesQueuedUpdates : IQueuedUpdatesGroup
+		{
+			private readonly UIElement _uiElement;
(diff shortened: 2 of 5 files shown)


src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.wasm.cs
```diff
@@ -68,21 +68,35 @@ private void ConditionalUpdate(ref bool condition, Action action)
...
+			ConditionalUpdate(ref _fontStyleChanged, () => this.SetFontStyle(FontStyle, setStyles, resetStyles));
+			ConditionalUpdate(ref _fontWeightChanged, () => this.SetFontWeight(FontWeight, setStyles, resetStyles));
+			ConditionalUpdate(ref _fontFamilyChanged, () => this.SetFontFamily(FontFamily, setStyles, resetStyles));
+			ConditionalUpdate(ref _fontSizeChanged, () => this.SetFontSize(FontSize, setStyles, resetStyles));
+			ConditionalUpdate(ref _maxLinesChanged, () => this.SetMaxLines(MaxLines, setStyles, resetStyles));
+			ConditionalUpdate(ref _textAlignmentChanged, () => this.SetTextAlignment(TextAlignment, setStyles, resetStyles));
+			ConditionalUpdate(ref _lineHeightChanged, () => this.SetLineHeight(LineHeight, setStyles, resetStyles));
+			ConditionalUpdate(ref _characterSpacingChanged, () => this.SetCharacterSpacing(CharacterSpacing, setStyles, resetStyles));
+			ConditionalUpdate(ref _textDecorationsChanged, () => this.SetTextDecorations(TextDecorations, setStyles, resetStyles));
+			ConditionalUpdate(ref _paddingChangedChanged, () => this.SetTextPadding(Padding, setStyles, resetStyles));
```
jeromelaban, line 84: This is looking interesting! The only part that I think can be optimized is the closure on `setStyles` and `resetStyles`, which forces the instanciation of a display class. I think those two should be class members instead.

/cc @carldebilly

Youssef1313: > I think those two should be class members instead.

@jeromelaban Are you referring to `setStyles` and `resetStyles` lists? They're not needed anywhere else. What would be the benefit of having them as class members? Reduce allocations of new lists by clearing the existing ones probably?

jeromelaban: It'll avoid the compiler to create a closure over those two members (and a display class as result). Per your last change, making those class members is not needed anymore for that specific reason (as you removed the lambdas as well). Otherwise, it's looking good.

carldebilly:
I would have done another approach for this problem: create a kind of "transactional update" where we can suspend all updates on an element and apply the differential by _commiting_ all updates at once.  Something like:

```
using(this.SuspendCssUpdates())
{
  this.SetStyle("Foreground", "UglyColor");
  this.SetSyle("Background", "BeautifulColor");
  [...]
}
```

src/Uno.UI/UI/Xaml/UIElement.wasm.cs
```diff
@@ -27,6 +27,75 @@ public partial class UIElement : DependencyObject
 		internal const string DefaultHtmlTag = "div";
 
 		private readonly GCHandle _gcHandle;
+		private bool _isCssStyleUpdatesSuspended = false;
+
+		/// <summary>
+		/// Since there is no way to apply all set and reset updates in bulk, we create them of groups and execute
+		/// each group separately to maintain the desired order.
+		/// </summary>
+		private readonly Queue<IQueuedUpdatesGroup> _queuedUpdates = new Queue<IQueuedUpdatesGroup>();
```
Youssef1313, line 36: To maintain the order, this may still lead to multiple JS calls when SetStyles are mixed with ResetStyles, but the number of calls should be less than before.

In the future, we may try to add a mechanism that allows us to know whether the order is important or not and optimize this further.

src/Uno.UI/UI/Xaml/UIElement.wasm.cs
```diff
@@ -27,6 +27,75 @@ public partial class UIElement : DependencyObject
...
+
+		internal interface IQueuedUpdatesGroup
```
Youssef1313, line 38: Not sure the name is good enough

src/Uno.UI/UI/Xaml/UIElement.wasm.cs
```diff
@@ -27,6 +27,75 @@ public partial class UIElement : DependencyObject
...
+		{
+			void Execute();
+		}
+
+		internal class SetStylesQueuedUpdates : IQueuedUpdatesGroup
+		{
+			private readonly UIElement _uiElement;
+			private readonly List<(string name, string value)> _styles = new List<(string name, string value)>();
```
jeromelaban, line 46: Try to avoid tuples on wasm, specifically. This uses generics and it's not efficient in current builds of the runtime.

jeromelaban (commented):
[There are benchmarks](https://github.com/unoplatform/uno/blob/master/src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/TextBlockControl/TextBlock_MeasurePeformance.xaml) for TextBlock updates, can you post the numbers before and after ?

You can  use the `site-net5.0` in the build artifacts:

image

and run it with `dotnet serve`.

src/Uno.UI/UI/Xaml/UIElement.wasm.cs
Youssef1313: @jeromelaban The existing methods `SetStyle` and `ResetStyle` works with tuples. I can introduce a small struct containing the name and value, but this will probably require lots of updates over the codebase for every caller to these methods. 

Alternatively, I could introduce new overloads and leave the existing ones, but that will keep the inefficiency problem in some places.

What do you think?

Youssef1313:
@jeromelaban I'm not seeing differences using [this test](https://github.com/unoplatform/uno/blob/master/src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/TextBlockControl/TextBlock_MeasurePeformance.xaml). Does it actually test the code path I'm modifying?

jeromelaban:
> @jeromelaban I'm not seeing differences using [this test](https://github.com/unoplatform/uno/blob/master/src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/TextBlockControl/TextBlock_MeasurePeformance.xaml). Does it actually test the code path I'm modifying?

Yes it's supposed to. This would mean that the change is not particularly interesting. Did you run the test using the artifacts or a local build ?

Youssef1313:
> > @jeromelaban I'm not seeing differences using [this test](https://github.com/unoplatform/uno/blob/master/src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/TextBlockControl/TextBlock_MeasurePeformance.xaml). Does it actually test the code path I'm modifying?
> 
> Yes it's supposed to. This would mean that the change is not particularly interesting. Did you run the test using the artifacts or a local build ?

Artifacts from Azure. One from this PR and one from master.

jeromelaban:
Thanks. That would mean that either the code impact the performance, or maybe we should be adding a different test that changes the color, for instance.

Youssef1313:
@jeromelaban I added another test in the last commit. It looks like the results are almost the same, except for few runs without the fix it takes noticeable higher time, but that's not the usual case, so I'm afraid these are just outliers.

src/Uno.UI/UI/Xaml/UIElement.wasm.cs
```diff
@@ -169,11 +238,53 @@ private string GetHtmlTag(string htmlTag)
...
+				Uno.UI.Xaml.WindowManagerInterop.SetStyles(HtmlId, new[] { (name, value) });
+			}
 		}
 
+		/// <summary>
+		/// TODO: This doesn't respect _isCssUpdatesSuspended.
+		/// </summary>
+		/// <param name="name"></param>
+		/// <param name="value"></param>
 		protected internal void SetStyle(string name, double value)
```
Youssef1313, line 288: Oh probably that's it. I haven't updated this overload yet.

jeromelaban:
This would mean that the interop is not particularly expensive in this case (the measure is) and that it's not relevant to adjust this piece of the code to make it faster (and potentially introduce issues). It could be interesting to do the same test in a separate app and use Wasm AOT instead.

Youssef1313:
@jeromelaban Any instructions on how to test with AOT?

jeromelaban:
Some instructions are provided here: https://github.com/unoplatform/Uno.Wasm.Bootstrap#runtime-execution-modes.

jeromelaban:
@Youssef1313 can you take a look at the build errors ? Also, did you get to try building with AOT?

jeromelaban:
@Youssef1313 closing this PR as interop cost has gone down with the use of JSImport.

Issue #4440, opened 2020-11-02
Ways to Automate WinUI Control Updates


robloo:
## What would you like to be added:

Currently WinUI controls are imported manually. Right now that is unavoidable. However, long-term we need to automate as much as possible. Areas of automation that are possible right now are:

 1. Change detection
     * We should be recording in a ledger some place, for each control directory, which commit it was last manually updated for. This would allow a script to check if a new WinUI commit modified a directory and flag code that needs to be updated. It should be possible over time to gradually make this more granular than simply by directory.
       `ColorPicker : SHA' 
 2. Strings / translation updates should automatically be copied over. We don't change these at all.
 3. XAML should automatically be copied over. There might be some special cases but for the most part XAML is copy pasted.
 4. Generate .Properties.cs (and maybe an .Events.cs) file directly from the IDL in WinUI.
 5. WinUI API tests and control test pages may be copied over automatically with minor modification.

Doing the above will actually catch a significant number of issues before they happen and ensure code is synced with WinUI. This is especially important as long-term Uno will essentially just be a c++ -> c# conversion of controls with no special code.

## Why is this needed:

 * Improve developer experience and speed
 * Reduce the potential for missed updates or bugs fixed upstream never applied to Uno

## For which Platform:

- [ ] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] WebAssembly renderers for Xamarin.Forms
- [ ] macOS
- [ ] Skia
  - [ ] WPF
  - [ ] GTK (Linux)
  - [ ] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

## Anything else we need to know?

This is really to open the discussion. I don't plan on working on this myself in the foreseeable future.

jeromelaban:
Thanks for the report. Yes, this should be something needed mid-term.

MartinZikmund:
# Repro generator

This issue was classified as not reproducible with a code sample or test.

**Reason:** This is a tooling/automation improvement request for the WinUI control porting process. There is no user-facing behavior to reproduce — it is a developer process/automation enhancement.

PR #6790, opened 2021-08-17
feat: Automate syncing resources from mux repository

Youssef1313:
GitHub Issue (If applicable): Addresses point number 2 of https://github.com/unoplatform/uno/issues/4440. Can work on the whole issue in a series of PRs if you liked that approach I took

## PR Type

- Feature

## What is the new behavior?

Automated PRs for translation file changes.

## PR Checklist

## Other information

### DIFF
--- a/.github/workflows/sync.yml
+++ b/.github/workflows/sync.yml
@@ -2,7 +2,7 @@ name: 'Sync with MUX'
 
 on:
   schedule:
-  - cron: '0 0 1 * *'
+  - cron: '0 0 * * 0' # Runs weekly on Sunday.
   workflow_dispatch:
     inputs:
       reason:
@@ -33,7 +33,7 @@ jobs:
       uses: peter-evans/create-pull-request@v3.10.0
       with:
         title: 'Sync with MUX'
-        commit-message: 'fix: Sync with MUX'
+        commit-message: 'chore: Update WinUI RESW files'
         base: 'master'
         path: uno
         branch: 'sync-with-mux'
--- a/.github/workflows/sync.yml
+++ b/.github/workflows/sync.yml
@@ -1,0 +1,40 @@
+name: 'Sync with MUX'
+
+on:
+  schedule:
+  - cron: '0 0 1 * *'
+  workflow_dispatch:
+    inputs:
+      reason:
+        description: 'The reason for running the workflow'
+        required: true
+        default: 'Manual run'
+
+jobs:
+  sync-mux:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        path: uno
+    - uses: actions/checkout@v2
+      with:
+        repository: microsoft/microsoft-ui-xaml
+        path: microsoft-ui-xaml
+    - name: 'Print manual run reason'
+      if: ${{ github.event_name == 'workflow_dispatch' }}
+      run: |
+        echo 'Reason: ${{ github.event.inputs.reason }}'
+    - name: Sync with MUX
+      id: sync-mux
+      uses: ./uno//src/Tools/Automation
+    - name: create-pull-request
+      uses: peter-evans/create-pull-request@v3.10.0
+      with:
+        title: 'Sync with MUX'
+        commit-message: 'fix: Sync with MUX'
+        base: 'master'
+        path: uno
+        branch: 'sync-with-mux'
+        branch-suffix: 'timestamp'
--- a/src/Tools/Automation/ActionRunner/Program.cs
+++ b/src/Tools/Automation/ActionRunner/Program.cs
@@ -36,7 +36,7 @@ namespace ActionRunner
 				var unoContent = File.ReadAllText(unoPath);
 				if (muxContent != unoContent)
 				{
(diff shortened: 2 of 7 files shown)


Youssef1313:
Example automated PR: https://github.com/Youssef1313/uno/pull/4

.github/workflows/sync.yml
```diff
@@ -0,0 +1,40 @@
+name: 'Sync with MUX'
+
+on:
+  schedule:
+  - cron: '0 0 1 * *'
```
Youssef1313, line 5: This runs monthly, with ability to manual run from "Actions" tab on GitHub. Let me know if you want it to be more frequent.

src/Tools/Automation/ActionRunner/Program.cs
```diff
@@ -0,0 +1,60 @@
...
+				{
+					GitHubLogger.LogWarning($"Path '{muxPath}' was not found.");
+					continue;
+				}
+
+				var muxContent = File.ReadAllText(muxPath);
+				var unoContent = File.ReadAllText(unoPath);
+				if (muxContent != unoContent)
+				{
+					GitHubLogger.LogError($"Files '{muxPath}' and '{unoPath}' are not identical!!!");
```
MartinZikmund, line 39: I would not log this as error - as this is actually the success path for the tool :-)

.github/workflows/sync.yml
```diff
@@ -0,0 +1,40 @@
...
+      run: |
+        echo 'Reason: ${{ github.event.inputs.reason }}'
+    - name: Sync with MUX
+      id: sync-mux
+      uses: ./uno//src/Tools/Automation
+    - name: create-pull-request
+      uses: peter-evans/create-pull-request@v3.10.0
+      with:
+        title: 'Sync with MUX'
+        commit-message: 'fix: Sync with MUX'
```
MartinZikmund, line 36: In this case I would just call it a "chore: Update WinUI RESW files"

.github/workflows/sync.yml
MartinZikmund: I think this will run quite quickly, so it could even be weekly

src/Tools/Automation/ActionRunner/Program.cs
```diff
@@ -0,0 +1,60 @@
...
+			}
+			else if (!Directory.Exists(UnoPathPrefix))
+			{
+				GitHubLogger.LogError($"Can't find '{UnoPathPrefix}' directory.");
+			}
+
+			string[] unoResourceFiles = CollectResourcesFromUno();
+			GitHubLogger.LogInformation($"Found {unoResourceFiles.Length} resources files.");
+
+			foreach (string unoPath in unoResourceFiles)
```
Youssef1313, line 26: Note: the workflow currently will only update existing files, but won't create new ones. Probably it should also create new ones as well?

MartinZikmund: For now copying XAML styles is tricky, as many of them need Uno specific adjustments and we also are in process of adding support for WinUI 2.6. I would suggest postponing this after we figure out what will be the structuring/how to deal with the specifics

MartinZikmund: Regarding new resources - we could probably include them, as it will not impact the assembly size at all, just it could be a bit confusing, that the folders for the controls are there while the controls are not. What do you think?

Youssef1313: > For now copying XAML styles is tricky, as many of them need Uno specific adjustments and we also are in process of adding support for WinUI 2.6. I would suggest postponing this after we figure out what will be the structuring/how to deal with the specifics

What I was planning to do here isn't really a "copy/paste". But get the diff and try to apply it. If there are conflicts, the bot will open an issue, otherwise, it will open a PR. I'm not sure if that would be easy enough, but worth trying. Let's look into that after adding support for WinUI 2.6.

For the new resources, I won't include ones from controls that don't exist. Only for existing controls (ie, if a new language was added in WinUI repo for an existing control). Let me know if you prefer that to be in a separate PR or this one.

Youssef1313:
@jeromelaban Can you take a look please?

jeromelaban:
> @jeromelaban Can you take a look please?

I like the idea of an automated PR for this, I don't think updating the resources manually makes sense without pulling the sources as well. The resources content may be desynchronized with the resources.

The tool itself makes sense, though and we should add it to a document that explain how to port controls from WinUI to Uno.

Youssef1313:
> we should add it to a document that explain how to port controls from WinUI to Uno.

Does such a document already exist somewhere?

Youssef1313:
Pinging @jeromelaban

Youssef1313:
Tagging @jeromelaban for review

jeromelaban:
I don't think it makes sense to synchronize resources standalone without the code associated to it. As I mentioned previously, automating is not particularly useful, but the tool itself is, as part of a document which explains how to updates the resources.

jeromelaban:
Let's review this again once winui sources are released.

MartinZikmund:
Superseeded by https://github.com/unoplatform/uno/issues/14223

Issue #6522, opened 2021-07-21
[iOS/Android] SwipeControl, Releasing swipe on a Button / ListItem, triggers the click


MaximeDion-Work:
## Current behavior

When releasing a swipe the click action of the item underneath is triggered.

## Expected behavior

Releasing a swipe should not trigger the click action.

## How to reproduce it (as minimally and precisely as possible)

Apply a SwipeControl around a LisViewItem or Button.
Start swiping then release
Notice command of of the control is triggered 

## Workaround

None

## Works on UWP/WinUI

Yes

## Environment

Nuget Package:

Uno.UI: 3.9.0-dev.307
Uno.SourceGenerationTasks 3.1.0
Uno.UI.RemoteControl: 3.9.0-dev.233

Nuget Package Version(s):

Affected platform(s):

- [x] iOS
- [x] Android

IDE:

- [ ] Visual Studio 2017 (version: )
- [x] Visual Studio 2019 (Version 16.9.4)
- [ ] Visual Studio for Mac (version: )
- [ ] Rider Windows (version: )
- [ ] Rider macOS (version: )
- [ ] Visual Studio Code (version: )

jeromelaban:
@dr1rrb related to or fixed by your recent fixes ?

dr1rrb:
Same UI, but not the same bug. Validated on Windows, indeed the button clicked is not triggered ... which is against the common rule (a `Button` clicked is raised as soon as you press and release a pointer over it, no matter how far you moved).

dr1rrb:
A workaround would be to use the `Tapped` event on the button (so you still get the visual states of the `Button`) instead of the `Clicked/Command`. The down side is if you disable the button using the `Command.CanExecute`, you will have either:
- Use 2 commands (one databound to the `Button.Command` and used only for the `CanExecute` and another one which has to be triggered by the `Tapped` using a custom behavior and which does the effective work)
- Handle the `Disable` by your own

dr1rrb:
BTW, I would argue about the user experience of having a `Button` inside a `SwipeControl`. Either the action should probably be a one handled by the `SwipeControl`; either it's used for non touch platform (where `SwipeControl` does nothing) and then you should either use a context menu, either show the buttons only on mouse over (like the window's mail application)

jeanplevesque:
[...]

PR #6796, opened 2021-08-17
Fix ListView scrolling with manipulation

dr1rrb:
fixes https://github.com/unoplatform/uno/issues/6522#issuecomment-899600535

## Bugfix
We cannot scroll a `ListView` as soon as we initiate scrolling from an element which has a `ManipulationMode != System`.

* On iOS we are using the default `TouchesManager` which does not take in consideration the scroll direction
* On Android, the scroll direction is wrongly reported by the `NativeListViewBase` (reports `Hidden` instead of `Disabled`)

## What is the new behavior?
🙃

## PR Checklist


src/Uno.UI/UI/Xaml/Controls/ListViewBase/NativeListViewBase.Android.cs
```diff
@@ -127,7 +127,7 @@ public ScrollBarVisibility HorizontalScrollBarVisibility
 		{
 			get
 			{
-				return HorizontalScrollBarEnabled ? ScrollBarVisibility.Visible : ScrollBarVisibility.Hidden;
+				return HorizontalScrollBarEnabled ? ScrollBarVisibility.Visible : ScrollBarVisibility.Disabled;
```
dr1rrb, line 130: @davidjohnoliver I definitively need your input on this one :)

davidjohnoliver: tbh I struggle with the nuances of scroll bar enablovisibleness... I don't have a problem with this change, but should we be instead preserving (and returning) the value sent to `set`? Currently we're throwing that information away because the underlying native property doesn't accommodate the distinction.

davidjohnoliver (commented):
I think you should be able to add a runtime test for this bug.

src/Uno.UI/UI/Xaml/Controls/ListViewBase/NativeListViewBase.Android.cs
dr1rrb: I found that there is a good explanation in the MS docs actually https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.scrollbarvisibility?view=winrt-20348, but the mess starts when you also consider the `ScrollMode` ... each time I've to spend way to much time to remind all possible values, including the weird case where you can have a `ScrollViewer` which can be scrolled only programmatically but displaying scrollbars.

Anyway you're right, I was a bit lazy, but keeping the "real" value in a backing field is much better.

davidjohnoliver (approved):

PR #6818, opened 2021-08-20
fix: Hide open flyouts when MessageDialog is shown

### LINKED ISSUE #1929: Flyout does not light dismiss in the same way as UWP on Android.
On UWP a flyout will light dismiss when a button on the flyout is pressed, this does not happen on Android.

The flyout only dismisses when the screen outside the flyout is tapped

A sample that reproduces the issue can be found here:
https://github.com/BrianDT/UnoFlyoutIssue

Package Version(s):
Uno.UI: 2.0.512-dev.3530

Affected platform(s):

- [ ] iOS
- [x] Android
- [ ] WebAssembly
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

Have not  been unable to determine if the issue also exists on iOS because on iOS button tap on a flyout does not execute the button command. This will be raised as a separate issue.

Visual Studio:

- [ ] 2017 (version: )
- [x] 2019 (version: )
- [ ] for Mac (version: )

Relevant plugins:

- [ ] Resharper (version: )

Youssef1313:
GitHub Issue (If applicable): closes #1929

## PR Type

- Bugfix

flyouts not closed when a MessageDialog is shown.

## What is the new behavior?

They are now closed, matching UWP behavior.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UWP/Helpers/VisualTreeHelperProxy.cs
+++ b/src/Uno.UWP/Helpers/VisualTreeHelperProxy.cs
@@ -1,0 +1,15 @@
+﻿#nullable enable
+
+using System;
+
+namespace Uno.Helpers
+{
+	internal static class VisualTreeHelperProxy
+	{
+		private static Action? _closeAllPopups;
+
+		public static void CloseAllPopups() => _closeAllPopups?.Invoke();
+
+		public static void SetCloseAllPopupsAction(Action closeAllPopups) => _closeAllPopups = closeAllPopups;
+	}
+}
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Popups/Given_MessageDialog.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Popups/Given_MessageDialog.cs
@@ -2,7 +2,6 @@
 using System.Collections.Generic;
 using System.Text;
 using System.Threading.Tasks;
-using Android.Widget;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Windows.UI.Popups;
 using Windows.UI.Xaml;
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Popups/Given_MessageDialog.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Popups/Given_MessageDialog.cs
@@ -1,0 +1,36 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+using Android.Widget;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Windows.UI.Popups;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Controls;
+using Windows.UI.Xaml.Controls.Primitives;
+using Windows.UI.Xaml.Media;
+using static Private.Infrastructure.TestServices;
+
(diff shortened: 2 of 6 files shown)


src/Uno.UWP/VisualTreeHelperHelper.cs
```diff
@@ -0,0 +1,15 @@
+﻿#nullable enable
+
+using System;
+
+namespace Uno
+{
+	internal static class VisualTreeHelperHelper
```
Youssef1313, line 7: The duplication may feel weird. But two reasons:

1. Following the same naming done for `DispatcherTimerHelper` (ie, type name existing in one layer but not the otherr + "Helper"). Here, the type name in itself is "VisualTreeHelper".
2. Avoid name clashes between the two types

MartinZikmund:
@jeromelaban could we use the extension facility for this purpose? Not sure if the setup works on non-Skia targets currently

src/Uno.UWP/VisualTreeHelperHelper.cs
MartinZikmund: Please move this type to `Helpers` folder. As for the naming, maybe `VisualTreeHelperProxy` would be nicer. We could also rename the other to `DispatcherTimerProxy` - but let's see if we could even use the extension facility instead, that would be even cleaner.

Youssef1313: @MartinZikmund I moved it to `Helpers`. (Note that `DispatcherTimeHelper` is in `Uno` folder not `Helpers` folder)

MartinZikmund: While you are at it, you can move the `DispatcherTimeHelper` too, for cleanup purposes :-)

src/Uno.UWP/Helpers/VisualTreeHelperProxy.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Uno.Helpers
+{
+	internal static class VisualTreeHelperProxy
+	{
+		private static Action? _closeAllPopups;
+
+		public static void CloseAllPopups() => _closeAllPopups?.Invoke();
+
+		public static void SetPopupsCloser(Action closeAllPopups) => _closeAllPopups = closeAllPopups;
```
jeromelaban, line 13:
```suggestion
		public static void SetCallAllPopupsAction(Action closeAllPopups) => _closeAllPopups = closeAllPopups;
```

jeromelaban:
> @jeromelaban could we use the extension facility for this purpose? Not sure if the setup works on non-Skia targets currently

I don't think we need this type of registration for now, as it'll add a dictionary entry, that is not really needed. It's internal, so we can refactor as needed in the future.

jeromelaban:
@Youssef1313  Could you rebase this PR? I also wonder if we can test that. Likely by trying find multiple flyouts if any.

jeromelaban (approved):

Youssef1313:
@jeromelaban The test is failing on Wasm. I found that I was missing `CloseAllPopups` call for Wasm, but even after adding it the test is still failing, I disabled the test on Wasm for now.

Youssef1313:
> > @jeromelaban The test is failing on Wasm. I found that I was missing `CloseAllPopups` call for Wasm, but even after adding it the test is still failing, I disabled the test on Wasm for now. Want me to open an issue tracking that?
> 
> Is the test working when executed manually ?

No, fails locally too. I have already disabled the test on Wasm.

PR #6828, opened 2021-08-22
fix: Use ids compatible with CSS

### LINKED ISSUE #3409: Numeric element IDs in WASM are not compatible with CSS

Youssef1313:
GitHub Issue (If applicable): closes #3409

## PR Type

- Bugfix

Element IDs aren't compatible with CSS

## What is the new behavior?

Prefix the IDs with "uno-"

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI.Runtime.WebAssembly/Xaml/UIElementWasmExtensions.cs
+++ b/src/Uno.UI.Runtime.WebAssembly/Xaml/UIElementWasmExtensions.cs
@@ -121,7 +121,7 @@ namespace Windows.UI.Xaml
 			var js = @$"
 (function(element) {{
 {jsCode}
-}})(Uno.UI.WindowManager.current.getView({element.HtmlId}));
+}})(Uno.UI.WindowManager.current.getView(""{element.HtmlId}""));
 ";
 			return WebAssemblyRuntime.InvokeJS(js);
 		}
@@ -140,7 +140,7 @@ namespace Windows.UI.Xaml
 (function(element) {{
 const __f = () => {asyncJsCode};
 return __f(element);
-}})(Uno.UI.WindowManager.current.getView({element.HtmlId}));
+}})(Uno.UI.WindowManager.current.getView(""{element.HtmlId}""));
 ";
 			return WebAssemblyRuntime.InvokeAsync(js);
 		}
--- a/src/Uno.UI/UI/Xaml/UIElement.wasm.cs
+++ b/src/Uno.UI/UI/Xaml/UIElement.wasm.cs
@@ -69,7 +69,7 @@ namespace Windows.UI.Xaml
 
 		private Rect GetBoundingClientRect()
 		{
-			var sizeString = WebAssemblyRuntime.InvokeJS("Uno.UI.WindowManager.current.getBoundingClientRect(" + HtmlId + ");");
+			var sizeString = WebAssemblyRuntime.InvokeJS("Uno.UI.WindowManager.current.getBoundingClientRect(\"" + HtmlId + "\");");
 			var sizeParts = sizeString.Split(';');
 			return new Rect(double.Parse(sizeParts[0]), double.Parse(sizeParts[1]), double.Parse(sizeParts[2]), double.Parse(sizeParts[3]));
 		}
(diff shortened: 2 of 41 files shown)


src/Uno.UI/UI/Xaml/UIElement.wasm.cs
```diff
@@ -94,7 +94,7 @@ public UIElement(string htmlTag, bool isSvg)
 			var type = GetType();
 
 			Handle = GCHandle.ToIntPtr(_gcHandle);
-			HtmlId = Handle;
+			HtmlId = $"uno-{Handle}";
```
Youssef1313, line 97: I'm really wondering why there is both `Handle` and `HtmlId` if they are the same? The properties are readonly and can't be changed after constructor execution, so they should always remain the same. What was the point of having both of them?

src/Uno.UI.Runtime.WebAssembly/Xaml/UIElementWasmExtensions.cs
```diff
@@ -121,7 +121,7 @@ public static string ExecuteJavascript(this UIElement element, string jsCode)
 			var js = @$"
 (function(element) {{
 {jsCode}
-}})(Uno.UI.WindowManager.current.getView({element.HtmlId}));
+}})(Uno.UI.WindowManager.current.getView(""{element.HtmlId}""));
```
Youssef1313, line 124: My concern about this PR is if I missed any places where quotes are needed. I hope I didn't miss, and hope coverage is enough to fail the CI if anything is missing.

Youssef1313:
My best guess about the current CI failure is:

https://github.com/unoplatform/Uno.UITest/blob/9a9c6d22aebbc566cffcf1e930270531c8c2c795/src/Uno.UITest.Puppeteer/SeleniumApp.cs#L616

@MartinZikmund @davidjohnoliver How do we approach updates in both repositories and testing them against each other?

jeromelaban:
This change will incur a performance cost because the ID is now a string that needs to be split and converted. We could investigate instead for using a different attribute on html element and see if the lookup performance is the same.

Youssef1313:
> This change will incur a performance cost because the ID is now a string that needs to be split and converted. We could investigate instead for using a different attribute on html element and see if the lookup performance is the same.

I'm not following. I haven't done any splitting anywhere, so I don't get when/why does this happen. There is already "xamlhandle" attribute which is still numeric.

jeromelaban:
> > This change will incur a performance cost because the ID is now a string that needs to be split and converted. We could investigate instead for using a different attribute on html element and see if the lookup performance is the same.
> 
> I'm not following. I haven't done any splitting anywhere, so I don't get when/why does this happen. There is already "xamlhandle" attribute which is still numeric.

Right, I read the code wrong for the splitting, but changing the id to a string is adding some performance cost. It needs to stay as a integer during the interop.

Youssef1313:
> > > This change will incur a performance cost because the ID is now a string that needs to be split and converted. We could investigate instead for using a different attribute on html element and see if the lookup performance is the same.
> > 
> > 
> > I'm not following. I haven't done any splitting anywhere, so I don't get when/why does this happen. There is already "xamlhandle" attribute which is still numeric.
> 
> Right, I read the code wrong for the splitting, but changing the id to a string is adding some performance cost. It needs to stay as a integer during the interop.

So what should the next step be? Or does this suggest that the linked issue shouldn't be fixed?

jeromelaban:
> So what should the next step be? Or does this suggest that the linked issue shouldn't be fixed?
Adding an `uno-` is interesting, but it should should be not be marshaled as is. It can be added when setting the `id` property on the javascript side, but based on the original `Handle` value.

Also, for all interop calls, the value to be passed needs to be the handle value, and for the element lookups, let's try to see if lookups on another property that is not "id" is as fast as the current lookup.

In all cases, this change is breaking so we can't merge it on master.

carldebilly:
I'm still convinced this change is not required.

jeromelaban:
I think we should not do this modification, and the original issue should be closed as well. Manipulating uno-generated elements using CSS is not a scenario we want to support, all things considered.

Youssef1313:
Let's close the PR and the linked issue then.

Issue #6882, opened 2021-08-26
Enable performance analyzers from NetAnalyzers package


Youssef1313:
## What would you like to be added:

Enable [performance analyzers](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/performance-warnings). It may or may not catch violations. But should be good to have.

## Why is this needed:

Performance.

## For which Platform:

- [ ] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] WebAssembly renderers for Xamarin.Forms
- [ ] macOS
- [ ] Skia
  - [ ] WPF
  - [ ] GTK (Linux)
  - [ ] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

Youssef1313:
https://github.com/unoplatform/uno/blob/e74305bc905f53f1acc46fb80e39a836ad061fde/.globalconfig#L7

PR #6892, opened 2021-08-27
perf: Enable CA1805 analyzer and fix violations

Youssef1313:
[Analyzer documentation](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1805).

GitHub Issue (If applicable): part of #6882

## PR Type

- Performance
- Build or CI related changes

## PR Checklist

## Other information


src/Uno.UI/Microsoft/UI/Xaml/Controls/Repeater/UniformGridLayout/UniformGridLayoutState.h.cs
```diff
@@ -38,8 +38,8 @@ public FlowLayoutAlgorithm FlowAlgorithm()
 		// private
 		FlowLayoutAlgorithm m_flowAlgorithm = new FlowLayoutAlgorithm();
 		
-		double m_effectiveItemWidth = 0.0;
-		double m_effectiveItemHeight = 0.0;
+		double m_effectiveItemWidth;
+		double m_effectiveItemHeight;
```
carldebilly, line 42: This is code imported from Microsoft.  We should not change it.

src/Uno.UI/Microsoft/UI/Xaml/Controls/RatingControl/RatingControl.cs
```diff
@@ -54,13 +54,13 @@ public partial class RatingControl : Control
...
-		private double m_mousePercentage = 0.0;
+		private bool m_isPointerOver;
+		private bool m_isPointerDown;
+		private double m_mousePercentage;
 
 		private RatingInfoType m_infoType = RatingInfoType.Font;
-		private double m_preEngagementValue = 0.0;
-		private bool m_disengagedWithA = false;
+		private double m_preEngagementValue;
+		private bool m_disengagedWithA;
```
carldebilly, line 63: This is code imported from Microsoft. We should not change it.

src/Uno.UI/Microsoft/UI/Xaml/Controls/TabView/TabView.cs
```diff
@@ -54,8 +54,8 @@ public partial class TabView : Control
 
 		private readonly DispatcherHelper m_dispatcherHelper;
 
-		private ContentPresenter m_tabContentPresenter = null;
-		private ContentPresenter m_rightContentPresenter = null;
+		private ContentPresenter m_tabContentPresenter;
+		private ContentPresenter m_rightContentPresenter;
```
carldebilly, line 58: This is code imported from Microsoft. We should not change it.

src/Uno.UI/Microsoft/UI/Xaml/Controls/TabView/TabViewItem.cs
```diff
@@ -24,10 +24,10 @@ public partial class TabViewItem : ListViewItem
...
 
 		private bool m_firstTimeSettingToolTip = true;
-		private bool m_hasPointerCapture = false;
-		private bool m_isMiddlePointerButtonPressed = false;
-		private bool m_isDragging = false;
-		private bool m_isPointerOver = false;
+		private bool m_hasPointerCapture;
+		private bool m_isMiddlePointerButtonPressed;
+		private bool m_isDragging;
+		private bool m_isPointerOver;
```
carldebilly, line 30: This is code imported from Microsoft. We should not change it.

src/Uno.UI/Microsoft/UI/Xaml/Controls/TreeView/TreeViewList.cs
```diff
@@ -22,8 +22,8 @@ namespace Microsoft.UI.Xaml.Controls
 	/// </summary>
 	public partial class TreeViewList : ListView
 	{
-		private bool m_isMultiselectEnabled = false;
-		private bool m_itemsSourceAttached = false;
+		private bool m_isMultiselectEnabled;
+		private bool m_itemsSourceAttached;
```
carldebilly, line 26: This is code imported from Microsoft. We should not change it.

src/Uno.UI/Microsoft/UI/Xaml/Controls/TreeView/TreeViewNode.cs
```diff
@@ -17,8 +17,8 @@ public partial class TreeViewNode : DependencyObject, ICustomPropertyProvider, I
 	{
 		private WeakReference<TreeViewNode> m_parentNode;
 		private bool m_HasUnrealizedChildren;
-		private object m_itemsSource = null;
-		private ItemsSourceView m_itemsDataSource = null;
+		private object m_itemsSource;
+		private ItemsSourceView m_itemsDataSource;
```
carldebilly, line 21: This is code imported from Microsoft. We should not change it.

src/Uno.UI/Microsoft/UI/Xaml/Controls/TreeView/TreeViewViewModel.cs
```diff
@@ -24,7 +24,7 @@ internal partial class TreeViewViewModel : ObservableVector<object>
 
 		private bool m_isContentMode;
 
-		private int m_selectionTrackingCounter = 0;
+		private int m_selectionTrackingCounter;
```
carldebilly, line 27: This is code imported from Microsoft. We should not change it.

src/Uno.UI/Microsoft/UI/Xaml/Controls/TwoPaneView/TwoPaneView.cs
```diff
@@ -25,7 +25,7 @@ public partial class TwoPaneView : Windows.UI.Xaml.Controls.Control
 
 		ViewMode m_currentMode = ViewMode.None;
 
-		bool m_loaded = false;
+		bool m_loaded;
```
carldebilly, line 28: This is code imported from Microsoft. We should not change it.

src/Uno.UI/UI/Xaml/Controls/CalendarView/LayoutStrategyBase.h.cs
```diff
@@ -77,8 +77,8 @@ public virtual ILayoutDataInfoProvider GetLayoutDataInfoProviderNoRef
 		//    static int GetRemainingGroups( int referenceGroupIndex,  int totalGroups,  RelativePosition positionOfReference);
 		//    static int GetRemainingItems( int referenceItemIndex,  int totalItems,  RelativePosition positionOfReference);
 
-		protected static int c_specialGroupIndex = 0;
-		protected static int c_specialItemIndex = 0;
+		protected static int c_specialGroupIndex;
+		protected static int c_specialItemIndex;
```
carldebilly, line 81: This is code imported from Microsoft. We should not change it.

src/Uno.UI/UI/Xaml/Controls/ContentDialog/ContentDialog.Visuals.cs
```diff
@@ -25,12 +25,15 @@ enum PlacementMode
...
 
 		PlacementMode m_placementMode = PlacementMode.Undetermined;
+#pragma warning disable CA1805 // Do not initialize unnecessarily
+		// TODO: This field is ALWAYS false. Either remove it or assign when appropriate.
 		bool m_hideInProgress = false;
-		bool m_hasPreparedContent = false;
-		bool m_isShowing = false;
+#pragma warning restore CA1805 // Do not initialize unnecessarily
+		bool m_hasPreparedContent;
+		bool m_isShowing;
```
carldebilly, line 33: This is code imported from Microsoft. We should not change it.

src/Uno.UI/UI/Xaml/Controls/DatePicker/DatePicker.cs
```diff
@@ -151,7 +151,7 @@ public partial class DatePicker : Control
 		// for cancellation.
 		IAsyncInfo m_tpAsyncSelectionInfo;
 
-		bool m_isPropagatingDate = false;
+		bool m_isPropagatingDate;
```
carldebilly, line 154: This is code imported from Microsoft. We should not change it.

src/Uno.UI/UI/Xaml/Controls/Grid/Grid.h.cs
```diff
@@ -268,17 +268,17 @@ internal void InvalidateDefinitions()
...
-		CColumnDefinitionCollection m_pColumns = null; // Effective column collection.
+		CRowDefinitionCollection m_pRows; // Effective row collection.
+		CColumnDefinitionCollection m_pColumns; // Effective column collection.
 
 		// This is a temporary storage that is released after arrange.
 		// Note the ScopeExit in ArrangeOveride
-		DefinitionBase[] m_ppTempDefinitions = null; // Temporary definitions storage.
-		Xuint m_cTempDefinitions = 0; // Size in elements of temporary definitions storage
+		DefinitionBase[] m_ppTempDefinitions; // Temporary definitions storage.
+		Xuint m_cTempDefinitions; // Size in elements of temporary definitions storage
```
carldebilly, line 281: This is code imported from Microsoft. We should not change it.

src/Uno.UI/UI/Xaml/Controls/MenuFlyout/MenuFlyoutItem.cs
```diff
@@ -43,10 +43,10 @@ public partial class MenuFlyoutItem : MenuFlyoutItemBase
 		// UNO TODO
 		// IDisposable m_epMenuFlyoutItemClickEventCallback;
 
-		double m_maxKeyboardAcceleratorTextWidth = 0;
+		double m_maxKeyboardAcceleratorTextWidth;
 		TextBlock m_tpKeyboardAcceleratorTextBlock;
 
-		bool m_isTemplateApplied = false;
+		bool m_isTemplateApplied;
```
carldebilly, line 49: This is code imported from Microsoft. We should not change it.

src/Uno.UI/UI/Xaml/Controls/MenuFlyout/MenuFlyoutPresenter.cs
```diff
@@ -52,7 +52,7 @@ public partial class MenuFlyoutPresenter : ItemsControl, IMenuPresenter
 
 		private bool m_isSubPresenter;
 
-		private int m_depth = 0;
+		private int m_depth;
```
carldebilly, line 55: This is code imported from Microsoft. We should not change it.

src/Uno.UI/UI/Xaml/Controls/NavigationView/NavigationView.Members.cs
```diff
@@ -100,48 +103,48 @@ enum NavigationRecommendedTransitionDirection
...
-		bool m_shouldInvalidateMeasureOnNextLayoutUpdate = false;
+		bool m_shouldInvalidateMeasureOnNextLayoutUpdate;
 
 		// during measuring, we should ignore SelectChange in overflow, otherwise it enters deadloop.
-		bool m_shouldIgnoreOverflowItemSelectionChange = false;
+		bool m_shouldIgnoreOverflowItemSelectionChange;
 
 		// when exchanging items between overflow and primary, it cause selectionchange. and then item invoked, and may cause MeasureOverride like customer changed something.
-		bool m_shouldIgnoreNextMeasureOverride = false;
+		bool m_shouldIgnoreNextMeasureOverride;
```
carldebilly, line 130: This is code imported from Microsoft. We should not change it.

src/Uno.UI/UI/Xaml/Controls/NavigationView/NavigationViewItem.cs
```diff
@@ -36,11 +36,11 @@ public partial class NavigationViewItem : NavigationViewItemBase
...
 
-		bool m_isClosedCompact = false;
+		bool m_isClosedCompact;
 
-		bool m_appliedTemplate = false;
-		bool m_hasKeyboardFocus = false;
-		bool m_isContentChangeHandlingDelayedForTopNav = false;
+		bool m_appliedTemplate;
+		bool m_hasKeyboardFocus;
+		bool m_isContentChangeHandlingDelayedForTopNav;
```
carldebilly, line 43: This is code imported from Microsoft. We should not change it.

src/Uno.UI/UI/Xaml/Controls/NavigationView/NavigationViewList.cs
```diff
@@ -21,7 +21,7 @@ public  partial class NavigationViewList : ListView
 	{
 		NavigationViewListPosition m_navigationViewListPosition = NavigationViewListPosition.LeftNav;
 		bool m_showFocusVisual = true;
-		WeakReference<NavigationView> m_navigationView = null;
+		WeakReference<NavigationView> m_navigationView;
```
carldebilly, line 24: This is code imported from Microsoft. We should not change it.

src/Uno.UI/UI/Xaml/Controls/SwipeControl/SwipeControl.h.cs
```diff
@@ -196,20 +196,20 @@ private enum CreatedContent { Left, Top, Bottom, Right, None };
...
 
-		private bool m_thresholdReached = false;
+		private bool m_thresholdReached;
 
 		//Near content = left or top
 		//Far content = right or bottom
-		private bool m_blockNearContent = false;
-		private bool m_blockFarContent = false;
+		private bool m_blockNearContent;
+		private bool m_blockFarContent;
```
carldebilly, line 212: This is code imported from Microsoft. We should not change it.

Youssef1313:
Thanks for review @carldebilly. However, I don't see why we can't make the change for files imported from microsoft-ui-xaml since the change isn't big? Just removing explicitly default value per Microsoft code-analysis rule CA1805.

The original files are written in C++, and it's possible that uninitialized class members will be undefined in C++ (I haven't confirmed). So it's a matter of how C++ is different than C# regarding whether an explicit default value is needed or not.

I know the performance benefit here is very little, but I see zero harm in the change. Please correct me if I'm wrong. Thanks again!

davidjohnoliver:
> Thanks for review @carldebilly. However, I don't see why we can't make the change for files imported from microsoft-ui-xaml since the change isn't big? Just removing explicitly default value per Microsoft code-analysis rule CA1805.
> 
> The original files are written in C++, and it's possible that uninitialized class members will be undefined in C++ (I haven't confirmed). So it's a matter of how C++ is different than C# regarding whether an explicit default value is needed or not.
> 
> I know the performance benefit here is very little, but I see zero harm in the change. Please correct me if I'm wrong. Thanks again!

The point is to keep it as easy as impossible to manage upstream changes, that's the reason we avoid making changes like this that don't have a functional reason in imported code.

(Personally I'm dubious about the value of this change even elsewhere. Explicitly assigning the default value communicates intent: 'this value is intended to default to false', ie the author didn't just forget to initialize it.)

Youssef1313:
> (Personally I'm dubious about the value of this change even elsewhere. Explicitly assigning the default value communicates intent: 'this value is intended to default to false', ie the author didn't just forget to initialize it.)

@davidjohnoliver I agree with that. But at the same time there could be a benefit performance-wise and assembly size-wise.

```csharp
public class C1 {
    private int x = 0;
} 

public class C2 {
    private int x;
} 
```

The above code generates the following IL:

```cil
.class public auto ansi beforefieldinit C1
    extends [System.Private.CoreLib]System.Object
{
    // Fields
    .field private int32 x

    // Methods
    .method public hidebysig specialname rtspecialname 
        instance void .ctor () cil managed 
    {
        // Method begins at RVA 0x2050
        // Code size 15 (0xf)
        .maxstack 8

        IL_0000: ldarg.0
        IL_0001: ldc.i4.0
        IL_0002: stfld int32 C1::x
        IL_0007: ldarg.0
        IL_0008: call instance void [System.Private.CoreLib]System.Object::.ctor()
        IL_000d: nop
        IL_000e: ret
    } // end of method C1::.ctor

} // end of class C1

.class public auto ansi beforefieldinit C2
    extends [System.Private.CoreLib]System.Object
{
    // Fields
    .field private int32 x

    // Methods
    .method public hidebysig specialname rtspecialname 
        instance void .ctor () cil managed 
    {
        // Method begins at RVA 0x2060
        // Code size 8 (0x8)
        .maxstack 8

        IL_0000: ldarg.0
        IL_0001: call instance void [System.Private.CoreLib]System.Object::.ctor()
        IL_0006: nop
        IL_0007: ret
    } // end of method C2::.ctor

} // end of class C2
```

So it's a trade-off where I really can't say which side should win 😄 .
If you think the assembly-size reduction and performance improvement don't worth it here (ie, they're too small), you can close the PR.

Also please have a look at the [performance analyzers](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/performance-warnings) and let me know if there are other rules you are not interested in to save some time :)

Thanks again!

Youssef1313:
Pinging @davidjohnoliver @carldebilly on whether this should be closed and to check the last question in the last comment :)

davidjohnoliver:
As for the trade-off, @jeromelaban is probably the best person to say. For me it's weighing up a trivial amount of IL against an nearly-trivial difference in readability; I would err on the side of the latter but I don't feel strongly about it.

For the rest of the rules, the one I think we'd most likely want to ignore is 'CA1823: Avoid unused private fields', because it's fairly common to end up with unused fields both with imported WinUI code and with shared code where a field may be used on some platforms but not others, and the maintainability hit is probably not worth the slight size difference. 

Just from casting a quick eye, some of those rules that might be interesting to prioritize:

 * CA1810: Initialize reference type static fields inline
 * CA1815: Override equals and operator equals on value types
 * CA1821: Remove empty finalizers _(though I wouldn't expect to find many)_
 * CA1826: Use property instead of Linq Enumerable method _(and similar 'unnecessary Linq' rules)_
 * maybe the Stream-related ones? I would be wary of breaking difficult-to-test code paths, though

jeromelaban:
@Youssef1313 while this PR is making valid changes, it's still too big to enable in one shot, particularly when working with regressions that inadvertently be introduced. Breaking it down per project would make it easier to deal with.

Youssef1313:
@jeromelaban Will break it more. But just want to confirm you want to enable this specific analyzer? and whether we want to disable it for Microsoft-imported code or not?

jeromelaban:
Yes this analyzer makes sense to enable, but let's skip it for Microsoft imported code (completely for the files). As for the others David mentioned, same, let's enable them one by one. Thanks!

src/Uno.UI/Helpers/WinUI/SharedHelpers.cs
```diff
@@ -31,13 +31,13 @@ namespace Uno.UI.Helpers.WinUI
 {
 	internal class SharedHelpers
 	{
-#pragma warning disable CS0414
-		private static bool s_isOnXboxInitialized = false;
-		private static bool s_isOnXbox = false;
-#pragma warning restore CS0414
+#if HAS_UNO && !(NET461 || __NETSTD_REFERENCE__)
```
jeromelaban, line 34: Why is this needed ?

src/Uno.UI/UI/Xaml/Input/FocusManager.mux.cs
```diff
@@ -85,32 +85,32 @@ public partial class FocusManager : IFocusManager
 		/// "first" tab.  This is deemed better than getting stuck at the end.
 		/// (Reverse "first" and "last" for Shift-Tab.)
 		/// </summary>
-		private bool _canTabOutOfPlugin =
+		private const bool CanTabOutOfPlugin =
```
jeromelaban, line 88: Let's keep the original name, even if not changing, as it matches code from winui.

jeromelaban (approved):

src/Uno.UI/Helpers/WinUI/SharedHelpers.cs
Youssef1313: All usages are under the same condition (see method `IsOnXbox` in the same file).
If this is not guarded with this condition, we'll get CS0169 warning. I can add `#pragma warning disable CS0169`, but thought it would be better to be error-free and make it more clear for human reading the code that this fields are used under specific conditions.

src/Uno.UI/UI/Xaml/Input/FocusManager.mux.cs
Youssef1313: Sure. I'll rename it back. I expect the current build to fail anyway due to a small [ca1841](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1841) violation which I'm fixing in https://github.com/unoplatform/uno/pull/9631.

So, I'll fix this after https://github.com/unoplatform/uno/pull/9631 is merged hoping for a green CI.

src/Uno.UI.Runtime.Skia.Wpf/WpfHost.cs
```diff
@@ -55,8 +55,6 @@ public class WpfHost : WpfControl, WinUI.ISkiaHost, IWpfHost
 		private readonly Func<UnoApplication> _appBuilder;
 		private CompositeDisposable _registrations = new();
 
-		private bool _appStarted = false;
-
```
Youssef1313, line 59: @MartinZikmund This was never set to true.

I kept the existing behavior and just removed the always-false field.
If this is a bug and there should be `_appStarted = true` in `StartApp`, let me know.

src/Uno.UI.XamlHost.Skia.Wpf/UnoXamlHostBase.Focus.cs
```diff
@@ -142,7 +142,7 @@ private static WF.Rect BoundsRelativeTo(FrameworkElement sibling1, System.Window
             return origin;
         }
 
-        private bool _onTakeFocusRequested = false;
+        //private bool _onTakeFocusRequested;
```
Youssef1313, line 145: Commented this out instead of completely deleting because the code that reads the field is commented as well.

https://github.com/unoplatform/uno/blob/1198d47de79caff733d94d5c01b7df677d85168c/src/Uno.UI.XamlHost.Skia.Wpf/UnoXamlHostBase.Focus.cs#L83

src/Uno.UI/Helpers/WinUI/SharedHelpers.cs
jeromelaban: Makes sense.

src/Uno.UI.XamlHost.Skia.Wpf/UnoXamlHostBase.Focus.cs
jeromelaban: Looks good.

src/Uno.UI.Runtime.Skia.Wpf/WpfHost.cs
jeromelaban: That's fine, let's keep the field removed.

jeromelaban:
It may be caused by the analyzers, but yesterday's builds were particularly slow so it may be unrelated.

Youssef1313:
@jeromelaban 

Security analyzers are a bit slow because they do heavy data-flow analysis, and CA1041 is a bit slow too https://github.com/dotnet/roslyn-analyzers/issues/6143

I'll disable security analyzers and keep CA1041 enabled. If it remains slow, I'll disable CA1041 too.

build/ci/.azure-devops-package-net6-win.yml
```diff
@@ -4,7 +4,7 @@ parameters:
 jobs:
 - job: net6_win_build
   displayName: 'Build net6 Binaries'
-  timeoutInMinutes: 90
+  timeoutInMinutes: 120
```
Youssef1313, line 7:
```suggestion
  timeoutInMinutes: 90
```

build/ci/.azure-devops-package-generic.yml
```diff
@@ -4,7 +4,7 @@ parameters:
 jobs:
 - job: generic_win_build
   displayName: 'Build Generic Binaries'
-  timeoutInMinutes: 90
+  timeoutInMinutes: 120
```
Youssef1313, line 7:
```suggestion
  timeoutInMinutes: 90
```

PR #6906, opened 2021-08-30
fix: Align PrimaryLanguageOverride behavior with UWP

### LINKED ISSUE #6880: Globalization: Uno doesn't save application language for the next application startup
### Current behavior

After setting the new language via `Windows.Globalization.ApplicationLanguages.PrimaryLanguageOverride = "fa` application just does it in the runtime and doesn't persist when app is restarted. 

### Expected behavior

As it is in UWP, application should persist the language after application does a restart.

Even if there's a workaround to fix the issue, still Uno team needs to address this issue to match the UWP behavior.

Take a look at [this code](https://github.com/Mohsens22/OpenPasat/blob/master/Pasat/Pasat/Pasat.Shared/Extentions/LanguageHelper.cs).

Also you can see the actual behavior [in the browser here](https://github.com/Mohsens22/OpenPasat/blob/master/Pasat/Pasat/Pasat.Shared/Extentions/LanguageHelper.cs).

### Workaround

_No response_

### Works on UWP/WinUI

Yes

### Environment

### NuGet package version(s)

_No response_

### Affected platforms

Android, WebAssembly

### IDE

Visual Studio 2022, Visual Studio 2019

### IDE version

latest

### Relevant plugins

_No response_

_No response_

Youssef1313:
GitHub Issue (If applicable): closes #6880, partial improvement to #11257

## PR Type

- Bugfix

- `PrimaryLanguageOverride` default value is null
- `PrimaryLanguageOverride` setter allows null
- `PrimaryLanguageOverride` not persisted.

## What is the new behavior?

- Default value is now string.Empty matching UWP
- Null isn't allowed.
- Value is persisted (not sure if UWP does it the same way)

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_Globalization/Given_ApplicationLanguages.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_Globalization/Given_ApplicationLanguages.cs
@@ -1,38 +1,31 @@
-﻿using System.Globalization;
-using System.Linq;
-using Windows.Globalization;
+﻿using Windows.Globalization;
 using FluentAssertions;
-using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Windows.Storage;
 
-namespace Uno.UI.RuntimeTests.Tests.Windows_Globalization
+namespace Uno.UI.RuntimeTests.Tests.Windows_Globalization;
+
+[TestClass]
+public class Given_ApplicationLanguages
 {
-	[TestClass]
-	public class Given_ApplicationLanguages
+	[TestCleanup]
+	public void CleanUp()
 	{
-		[TestCleanup]
-		public void CleanUp()
-		{
-			ApplicationLanguages.PrimaryLanguageOverride = null;
-		}
+		ApplicationLanguages.PrimaryLanguageOverride = string.Empty;
+	}
 
-		[TestMethod]
-		public void Test_Chinese_With_Script_Subtag()
-		{
-			ApplicationLanguages.PrimaryLanguageOverride = "zh-Hans-CN";
+	[TestMethod]
+	public void Test_Chinese_With_Script_Subtag()
+	{
+		ApplicationLanguages.PrimaryLanguageOverride = "zh-Hans-CN";
+		ApplicationLanguages.Languages[0].Should().Be("zh-Hans-CN");
+		ApplicationData.Current.LocalSettings.Values["__Uno.PrimaryLanguageOverride"].Should().Be("zh-Hans-CN");
+	}
 
-			ApplicationLanguages.Languages[0].Should().Be("zh-Hans-CN");
-			CultureInfo.CurrentCulture.Name.Should().BeOneOf("zh-CN", "zh-Hans-CN");
-			CultureInfo.CurrentUICulture.Name.Should().BeOneOf("zh-CN", "zh-Hans-CN");
-		}
-
-		[TestMethod]
-		public void Test_French_With_Script_Subtag()
-		{
-			ApplicationLanguages.PrimaryLanguageOverride = "fr-Latn-CA";
-
-			ApplicationLanguages.Languages[0].Should().Be("fr-Latn-CA");
-			CultureInfo.CurrentCulture.Name.Should().BeOneOf("fr-CA", "fr-Latn-CA");
-			CultureInfo.CurrentUICulture.Name.Should().BeOneOf("fr-CA", "fr-Latn-CA");
-		}
+	[TestMethod]
+	public void Test_French_With_Script_Subtag()
+	{
+		ApplicationLanguages.PrimaryLanguageOverride = "fr-Latn-CA";
+		ApplicationLanguages.Languages[0].Should().Be("fr-Latn-CA");
+		ApplicationData.Current.LocalSettings.Values["__Uno.PrimaryLanguageOverride"].Should().Be("fr-Latn-CA");
 	}
 }
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_DatePicker.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_DatePicker.cs
@@ -123,8 +123,19 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 		private static IDisposable SetAmbiantLanguage(string language)
 		{
 			var previousLanguage = ApplicationLanguages.PrimaryLanguageOverride;
+			var currentCulture = CultureInfo.CurrentCulture;
+			var currentUICulture = CultureInfo.CurrentUICulture;
+			var defaultThreadCurrentCulture = CultureInfo.DefaultThreadCurrentCulture;
+			var defaultThreadCurrentUICulture = CultureInfo.DefaultThreadCurrentUICulture;
 			ApplicationLanguages.PrimaryLanguageOverride = language;
-			return Disposable.Create(() => ApplicationLanguages.PrimaryLanguageOverride = previousLanguage);
+			return Disposable.Create(() =>
+			{
+				ApplicationLanguages.PrimaryLanguageOverride = previousLanguage;
+				CultureInfo.CurrentCulture = currentCulture;
+				CultureInfo.CurrentUICulture = currentUICulture;
+				CultureInfo.DefaultThreadCurrentCulture = defaultThreadCurrentCulture;
+				CultureInfo.DefaultThreadCurrentUICulture = defaultThreadCurrentUICulture;
+			});
 		}
 
 		private static void CheckDateTimeTextBlockPartPosition(DatePicker datePicker, string id, int expectedColumn)
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_DatePicker.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_DatePicker.cs
@@ -1,4 +1,5 @@
-﻿using System;
+﻿#if !WINDOWS_UWP
+using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
@@ -151,3 +152,4 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 		}
 	}
 }
+#endif
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_DatePicker.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_DatePicker.cs
@@ -128,6 +128,7 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 			var defaultThreadCurrentCulture = CultureInfo.DefaultThreadCurrentCulture;
 			var defaultThreadCurrentUICulture = CultureInfo.DefaultThreadCurrentUICulture;
 			ApplicationLanguages.PrimaryLanguageOverride = language;
+			ApplicationLanguages.Initialize();
 			return Disposable.Create(() =>
 			{
 				ApplicationLanguages.PrimaryLanguageOverride = previousLanguage;
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_DatePicker.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_DatePicker.cs
@@ -129,7 +129,7 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 			var defaultThreadCurrentCulture = CultureInfo.DefaultThreadCurrentCulture;
 			var defaultThreadCurrentUICulture = CultureInfo.DefaultThreadCurrentUICulture;
 			ApplicationLanguages.PrimaryLanguageOverride = language;
-			ApplicationLanguages.Initialize();
+			ApplicationLanguages.ApplyCulture();
 			return Disposable.Create(() =>
 			{
 				ApplicationLanguages.PrimaryLanguageOverride = previousLanguage;
--- a/src/Uno.UI/UI/Xaml/Application.cs
+++ b/src/Uno.UI/UI/Xaml/Application.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Globalization;
 using Uno;
 using Uno.UI;
 using Uno.Diagnostics.Eventing;
@@ -8,6 +9,7 @@ using Windows.Foundation.Metadata;
 using Windows.UI.Xaml.Controls.Primitives;
 using Windows.ApplicationModel.Core;
 using Windows.ApplicationModel;
+using Windows.Globalization;
 using Uno.Helpers.Theming;
 using Windows.UI.ViewManagement;
 using Uno.Extensions;
@@ -247,6 +249,7 @@ namespace Windows.UI.Xaml
 
 		public static void Start(global::Windows.UI.Xaml.ApplicationInitializationCallback callback)
 		{
+			ApplicationLanguages.Initialize();
 			StartPartial(callback);
 		}
 
--- a/src/Uno.UI/UI/Xaml/Application.cs
+++ b/src/Uno.UI/UI/Xaml/Application.cs
@@ -249,7 +249,7 @@ namespace Windows.UI.Xaml
 
 		public static void Start(global::Windows.UI.Xaml.ApplicationInitializationCallback callback)
 		{
-			ApplicationLanguages.Initialize();
+			ApplicationLanguages.ApplyCulture();
 			StartPartial(callback);
 		}
 
--- a/src/Uno.UWP/Globalization/ApplicationLanguages.cs
+++ b/src/Uno.UWP/Globalization/ApplicationLanguages.cs
@@ -14,6 +14,14 @@ namespace Windows.Globalization
 
 		static ApplicationLanguages()
 		{
+#if !NET461
+			if (ApplicationData.Current.LocalSettings.Values.TryGetValue(PrimaryLanguageOverrideSettingKey, out var savedValue)
+				&& savedValue is string stringSavedValue)
+			{
+				_primaryLanguageOverride = stringSavedValue;
+			}
+#endif
+
 			ApplyLanguages();
 		}
 
@@ -24,7 +32,9 @@ namespace Windows.Globalization
 			{
 				var culture = CreateCulture(primaryLanguageOverride);
 				CultureInfo.CurrentCulture = culture;
+				CultureInfo.DefaultThreadCurrentCulture = culture;
 				CultureInfo.CurrentUICulture = culture;
+				CultureInfo.DefaultThreadCurrentUICulture = culture;
 			}
 		}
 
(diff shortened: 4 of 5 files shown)


src/Uno.UI.RuntimeTests/Tests/Windows_Globalization/Given_ApplicationLanguages.cs
```diff
@@ -12,17 +13,20 @@ public class Given_ApplicationLanguages
...
 		public void Test_Chinese_With_Script_Subtag()
 		{
+			
 			ApplicationLanguages.PrimaryLanguageOverride = "zh-Hans-CN";
 
 			ApplicationLanguages.Languages.First().Should().Be("zh-Hans-CN");
 			CultureInfo.CurrentCulture.Name.Should().BeOneOf("zh-CN", "zh-Hans-CN");
 			CultureInfo.CurrentUICulture.Name.Should().BeOneOf("zh-CN", "zh-Hans-CN");
+
+			ApplicationData.Current.LocalSettings.Values["Uno.PrimaryLanguageOverride"].Should().Be("ch-Hans-CN");
```
Youssef1313, line 29: I'm not sure this is the appropriate kind of test.

src/Uno.UWP/Globalization/ApplicationLanguages.cs
```diff
@@ -1,15 +1,16 @@
...
 using Uno.UI;
+using Windows.Storage;
 
 namespace Windows.Globalization
 {
 	public static partial class ApplicationLanguages
 	{
-		private static string _primaryLanguageOverride;
+		private static string _primaryLanguageOverride = string.Empty;
+		private const string PrimaryLanguageOverrideSettingKey = "Uno.PrimaryLanguageOverride";
```
jeromelaban, line 13: We're using `__` as a prefix for other related local settings added by uno:

```suggestion
		private const string PrimaryLanguageOverrideSettingKey = "__Uno.PrimaryLanguageOverride";
```

src/Uno.UWP/Globalization/ApplicationLanguages.cs
```diff
@@ -18,11 +19,22 @@ static ApplicationLanguages()
 
 		public static string PrimaryLanguageOverride
 		{
-			get => _primaryLanguageOverride;
+			get
+			{
+				if (_primaryLanguageOverride.Length == 0 &&
+					ApplicationData.Current.LocalSettings.Values[PrimaryLanguageOverrideSettingKey] is string savedValue)
```
jeromelaban, line 25: We should use TryGetValue instead.

jeromelaban (commented):
Nice !

src/Uno.UWP/Globalization/ApplicationLanguages.cs
```diff
@@ -22,9 +22,9 @@ public static string PrimaryLanguageOverride
 			get
 			{
 				if (_primaryLanguageOverride.Length == 0 &&
-					ApplicationData.Current.LocalSettings.Values[PrimaryLanguageOverrideSettingKey] is string savedValue)
+					ApplicationData.Current.LocalSettings.Values.TryGetValue(PrimaryLanguageOverrideSettingKey, out var savedValue))
 				{
-					_primaryLanguageOverride = savedValue;
+					_primaryLanguageOverride = (string)savedValue;
```
jeromelaban, line 27: We should still be validating is the setting value is actually a string, in case it gets corrupted.

src/Uno.UWP/Globalization/ApplicationLanguages.cs
```diff
@@ -18,11 +19,22 @@ static ApplicationLanguages()
 
 		public static string PrimaryLanguageOverride
 		{
-			get => _primaryLanguageOverride;
+			get
+			{
+				if (_primaryLanguageOverride.Length == 0 &&
+					ApplicationData.Current.LocalSettings.Values.TryGetValue(PrimaryLanguageOverrideSettingKey, out var savedValue))
```
Youssef1313, line 25:
```suggestion
					ApplicationData.Current.LocalSettings.Values.TryGetValue(PrimaryLanguageOverrideSettingKey, out var savedValue) &&
					savedValue is string stringSavedValue)
```

src/Uno.UWP/Globalization/ApplicationLanguages.cs
Youssef1313:
```suggestion
					_primaryLanguageOverride = stringSavedValue;
```

carldebilly (commented):
The problem with this feature is the async access to files during startup.  On Windows, the OS/Runtime can fetch the settings before starting the application.  On Uno, we're required to wait for a file to load before continuing the loading and we're often in a sync context.

This can cost a lot of precious starting time - specially on Wasm were the file access must wait for the IndexDB synchronization to happend.

jeromelaban:
> The problem with this feature is the async access to files during startup. On Windows, the OS/Runtime can fetch the settings before starting the application. On Uno, we're required to wait for a file to load before continuing the loading and we're often in a sync context.
> 
> This can cost a lot of precious starting time - specially on Wasm were the file access must wait for the IndexDB synchronization to happend.

The loading of settings is not async on wasm, it got changed to rely on the [localStorage synchronous API](https://github.com/unoplatform/uno/blob/f9fa20de665c0c5d6e6ffe8c3b5177b0a615f191/src/Uno.UI/ts/Windows/Storage/ApplicationDataContainer.ts#L166), so this is fine to keep it like this.

carldebilly:
We may need to validate the behavior of the [Uno.Sample app demoing this feature](https://github.com/unoplatform/Uno.Samples/blob/master/UI/LocalizationSamples/RuntimeCultureSwitching/UnoLocalization.Shared/App.xaml.cs).

Youssef1313:
@carldebilly I couldn't get it working in this sample from Uno.Samples.

I built the sample, changed nuget override version to 3.7.4, built Uno.UI-Android-only.slnf, then ran the sample.

But since I'm still getting the not implemented warnings: 
image

I think somehow it's not using the newly built packages.

jeromelaban:
This was fixed in a later build of Uno, which means that the reference assemblies are not overriden by the uno nuget override package. Try updating the app to 3.10-dev, then override.

Youssef1313:
> This was fixed in a later build of Uno, which means that the reference assemblies are not overriden by the uno nuget override package. Try updating the app to 3.10-dev, then override.

I'm seeing the same after updating and overriding. :(

jeromelaban:
Do you see the warnings on latest 3.10-dev ?

Youssef1313:
Yes 
image

Youssef1313:
Ok the warning is expected here since I only updated Droid head, the warnings are for the other platforms. But the behavior isn't correct. Nothing is persisted after restarting the app 😕

src/Uno.UI.RuntimeTests/Tests/Windows_Globalization/Given_ApplicationLanguages.cs
carldebilly: It's definitely not "black box testing", but in this case, I think it's acceptable since we won't restart the app just for that.

But a another test would be to set the local settings and find a way to force the ApplicationLanguages to rerun its initialization.

jeromelaban:
Have you taken a look at what's actually persisted ? (e.g. dumping the values on start and write)

Youssef1313:
@jeromelaban I think I understand what's going on now.

The persisted value is simply NEVER read at application startup. So it doesn't have effect. I'm not 100% sure when UWP reads it and what it does with it.

For Uno, I'm thinking we could benefit from the generator to simply add `CultureInfo.Current[UI]Culture = ....` around this line:

https://github.com/unoplatform/uno/blob/d97a9b606a201857bd514fdacd1bc95fb70604a7/src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.cs#L591

However, I'm not sure about `ResourceLoader.DefaultLanguage`. It doesn't exist in UWP and looks like it's Uno-specific. Should this property be adjusted too? It looks like setting CurrentCulture and CurrentUICulture are sufficient alone.

src/Uno.UWP/Globalization/ApplicationLanguages.cs
```diff
@@ -18,11 +19,41 @@ static ApplicationLanguages()
...
+				}
+
+				try
+				{
+					// This doesn't match UWP for edge cases. In UWP:
+					// Windows.Globalization.ApplicationLanguages.PrimaryLanguageOverride = "fr-abcdefgh"  // This succeeds.
+					// Windows.Globalization.ApplicationLanguages.PrimaryLanguageOverride = "fr-abcdefghe" // This fails.
+					// However, new CultureInfo(..) fails in both cases.
+					// These are edge cases anyway.
+					_ = new CultureInfo(value);
```
Youssef1313, line 47: I don't know if you usually care to this kind of edge cases which are unlikely to be in a production application. I made my best effort to mostly align with UWP. But not 100% aligning.

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.cs
```diff
@@ -589,6 +589,18 @@ private void GenerateResourceLoader(IndentedStringBuilder writer)
...
+			{
+				writer.AppendLineInvariant("var culture = new global::System.Globalization.CultureInfo(primaryLanguageOverride);");
+				// Undocumented implementation detail of .NET. Non-existing cultures have ThreeLetterWindowsLanguageName = ZZZ.
+				// UWP doesn't set such cultures as current cultures.
+				using (writer.BlockInvariant("if (culture.ThreeLetterWindowsLanguageName != \"ZZZ\")"))
+				{
+					writer.AppendLineInvariant("global::System.Globalization.CultureInfo.CurrentCulture = culture;");
+					writer.AppendLineInvariant("global::System.Globalization.CultureInfo.CurrentUICulture = culture;");
+				}
+			}
```
Youssef1313, line 603: This "as a whole" doesn't match UWP. In UWP, the CurrentCulture and CurrentUICulture are set BEFORE the first line of `Main` (`global::Windows.UI.Xaml.Application.Start((p) => new App());)` executes. I don't really know how they do it.

Curious to know how they could be doing that if someone knows? (even if we don't need to match UWP - just curious :smile:)

src/Uno.UWP/Globalization/ApplicationLanguages.cs
jeromelaban: It's interesting to add the validation anyways. We should be validating this in a runtime or unit test.

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.cs
Youssef1313: Yeah it's much better in Application.Start. Fixed.

src/Uno.UWP/Globalization/ApplicationLanguages.cs
Youssef1313: @jeromelaban This will likely involve restarting the app. I don't know how this is done. Is there an API for that somewhere?

jeromelaban: Yes, there is but it's not available for all platforms. Let's keep it this way for now.

src/Uno.UI/UI/Xaml/Application.cs
```diff
@@ -223,6 +225,19 @@ public void Exit()
...
+				var culture = new CultureInfo(primaryLanguageOverride);
+				// Undocumented implementation detail of .NET. Non-existing cultures have ThreeLetterWindowsLanguageName = ZZZ.
+				// UWP doesn't set such cultures as current cultures.
+				if (culture.ThreeLetterWindowsLanguageName != "ZZZ")
+				{
+					CultureInfo.CurrentCulture = culture;
+					CultureInfo.CurrentUICulture = culture;
+				}
+			}
+
```
jeromelaban, line 240: Let's move this to a method named `Initialize` in `ApplicationLanguages`.

src/Uno.UWP/Globalization/ApplicationLanguages.cs
```diff
@@ -18,11 +19,42 @@ static ApplicationLanguages()
...
+				}
+				catch (CultureNotFoundException)
+				{
+					throw new ArgumentException("Value does not fall within the expected range.");
+				}
```
Youssef1313, line 52: @jeromelaban In UWP app `new CultureInfo("fr-Latn-CA")` doesn't throw, but when setting LanguageOverride to this value, it magically turns into `"fr-CA"` (I don't know what's logic behind this).

However things are even worse since Mono runtime throws an exception when "fr-Latn-CA" is passed to CultureInfo ctor. I'm completely unsure how should we move forward here.

jeromelaban: We may not want for the language property override to fail, that would make an app unusable if an user's current culture matches this throwing construct.

We may need to re-create a simpler culture through string splitting (e.g. if there are two dashes, we take first and last parts, then first only so it does not fail). I know @MartinZikmund had a problem with his own culture which has three parts.

Youssef1313: @jeromelaban Taking another look, I think the logic in UWP is more complex than this. In UWP, `ApplicationLanguages.PrimaryLanguageOverride = "az-Latn-AZ";` sets it to `az-Latn` (third part ignored), however, for `fr-Latn-CA` the value is set to `fr-CA` (second part ignored).

My best guess is for the three-part cultures case:

- Check if `FirstPart-ThirdPart` exists in `CultureInfo.GetCultures(CultureTypes.SpecificCultures)`, use that. (This is like the `fr-Latn-CA` case).
- If not, use `FirstPart-SecondPart` (This is like the `az-Latn-AZ` part).
- In all cases, throw an exception if any part is greater than 4 characters

For two-part cultures, rely on `new CultureInfo(...)` as it's now.

@jeromelaban Does that sounds good and/or correct?

jeromelaban: @Youssef1313 lets try this :)

Youssef1313: There is even more magic than that. `zh-Hans-CN` is set as-is. So we have three different 3-part cultures, which 3 different behaviors on UWP.

I'm not really sure what to do here. Ignoring the validation completely will cause problems when we try to set this value as `CurrentCulture` and `CurrentUICulture`. Any suggestions?

jeromelaban: Let's handle the scenarios we know of, and we'll adjust as we find out more.

Youssef1313: @jeromelaban Actually I think there is no need for custom logic here. It seems like `PrimaryLanguageOverride` follows whatever `Language` class do in UWP.

The `Language` class implementation looks far away from being implemented correctly in Uno, and relying on it is certainly the correct way to go here. But it really needs to get fixed first. Currently relying on `Language` API will really fail in many scenarios that we even have tests for.

src/Uno.UWP/Globalization/ApplicationLanguages.cs
```diff
@@ -1,29 +1,71 @@
...
-		private static string _primaryLanguageOverride;
+		private static string _primaryLanguageOverride = string.Empty;
+		private const string PrimaryLanguageOverrideSettingKey = "__Uno.PrimaryLanguageOverride";
 
 		static ApplicationLanguages()
 		{
 			ApplyLanguages();
 		}
 
+		internal static void Initialize()
```
nickrandolph, line 20: Is this called from somewhere?

Youssef1313: @nickrandolph Yes. In `Application.cs`

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_DatePicker.cs
```diff
@@ -123,8 +123,19 @@ public async Task When_Hungarian_Culture_Column_Order()
 		private static IDisposable SetAmbiantLanguage(string language)
 		{
 			var previousLanguage = ApplicationLanguages.PrimaryLanguageOverride;
+			var currentCulture = CultureInfo.CurrentCulture;
+			var currentUICulture = CultureInfo.CurrentUICulture;
+			var defaultThreadCurrentCulture = CultureInfo.DefaultThreadCurrentCulture;
+			var defaultThreadCurrentUICulture = CultureInfo.DefaultThreadCurrentUICulture;
 			ApplicationLanguages.PrimaryLanguageOverride = language;
```
Youssef1313, line 130: TODO (to make tests pass): change cultures here.

src/Uno.UWP/Globalization/ApplicationLanguages.cs
```diff
@@ -79,29 +110,6 @@ private static void ApplyLanguages()
 
 				Languages = languages.Distinct().ToArray();
 			}
```
nickrandolph, line 112: Can we add back logging for when primarylanguage changes, and perhaps on startup when culture is being set

src/Uno.UWP/Globalization/ApplicationLanguages.cs
```diff
@@ -79,29 +110,6 @@ private static void ApplyLanguages()
...
-			{
-				if (typeof(ApplicationLanguages).Log().IsEnabled(LogLevel.Debug))
-				{
-					typeof(ApplicationLanguages).Log().Debug($"Using {primaryLanguage} as primary language");
-				}
-
-				var primaryCulture = CreateCulture(primaryLanguage);
-
-				CultureInfo.CurrentCulture = primaryCulture;
-				CultureInfo.CurrentUICulture = primaryCulture;
```
nickrandolph, line 95: Concerned that we're no longer setting currentculture when primaryoverridelanguage changes - this would be a breaking change for uno apps that rely on this behaviour (perhaps??)

Youssef1313: Added a feature flag for that.

src/Uno.UWP/Globalization/ApplicationLanguages.cs
Youssef1313: Sure. While adding back the logging, I found out that I no longer fallback to `Languages[0]` when `PrimaryLanguageOverride` is empty string.

I'm not sure what is WinUI behavior around that, but I preserved back the original Uno behavior regarding this specific point.

src/Uno.UWP/Globalization/ApplicationLanguages.cs
```diff
@@ -1,29 +1,93 @@
...
 {
 	public static partial class ApplicationLanguages
 	{
-		private static string _primaryLanguageOverride;
+		private static string _primaryLanguageOverride = string.Empty;
+		private const string PrimaryLanguageOverrideSettingKey = "__Uno.PrimaryLanguageOverride";
 
 		static ApplicationLanguages()
 		{
+#if !NET461
```
nickrandolph, line 19: If setting localsettings not supported for net461, then we should have the same condition for where the PrimaryLanguageOverride is persisted in the static PrimaryLanguageOverride property.

nickrandolph (approved):

jeromelaban (approved):
@Youssef1313 could you also take a look at https://github.com/unoplatform/Uno.Samples/tree/master/UI/LocalizationSamples/RuntimeCultureSwitching to see if needs changes after this new feature?

carldebilly:
@jeromelaban @Youssef1313 The merge of this PR causes a regression in the Figma plugin project. See the link right before (if you have access to repo).

PR #6909, opened 2021-08-30
fix: Fallback to specific regional culture if neutral culture isn't found

### LINKED ISSUE #6469: Resource Localization fallback lookup is too strict
**Describe the bug**
Android app is not taking the device system language when in Spanish and the app stays in English.

Tested with other languages like French and it's working fine.

| Spanish       |  French     |
| :------------- |  -----------: |
|  photo_2021-07-10_13-43-36 |  photo_2021-07-10_13-43-29 |

**Steps To Reproduce**

Install app
Change System Language to Spanish (in case you don't have this ready)
Open App
Check on the side menu that everything is still English

**Expected behavior**

App should translate into Spanish when System Language is set to Spanish

**Screenshots**

**Device and Application Information**
 - OS Build: Android 10 (Samsung S9) and Android 11 (Pixel 3 XL)
 - Architecture: Android
 - Application Version: 1.2.4-uno.764
 - Region: 
 - Dev Version Installed: 
 

**Additional context**

Youssef1313:
GitHub Issue (If applicable): closes #6469

## PR Type

- Bugfix

Specific cultures are not used if neutral culture isn't found.

## What is the new behavior?

Fallback to specific cultures.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.cs
+++ b/src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.cs
@@ -188,7 +188,7 @@ namespace Windows.ApplicationModel.Resources
 				// For example, if we have es and it's not found, we want to match es-MX
 				if (culture.IsNeutralCulture)
 				{
-					var specificCultures = CultureInfo.GetCultures(CultureTypes.SpecificCultures).Where(c => c.Parent.Equals(culture));
+					var specificCultures = CultureInfo.GetCultures(CultureTypes.SpecificCultures).Where(c => c.Parent.Equals(culture)).OrderByDescending(x => x.Name);
 					foreach (var specificCulture in specificCultures)
 					{
 						yield return specificCulture.IetfLanguageTag.ToLowerInvariant();
--- a/src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.cs
+++ b/src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.cs
@@ -184,6 +184,17 @@ namespace Windows.ApplicationModel.Resources
 			{
 				yield return culture.IetfLanguageTag.ToLowerInvariant();
 
+				// If we have a culture that doesn't specify country/region, we want to match its specific cultures if it's not found.
+				// For example, if we have es and it's not found, we want to match es-MX
+				if (culture.IsNeutralCulture)
+				{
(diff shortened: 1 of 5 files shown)


src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.cs
```diff
@@ -184,6 +184,17 @@ private static IEnumerable<string> GetCulturesHierarchy(CultureInfo culture)
 			{
 				yield return culture.IetfLanguageTag.ToLowerInvariant();
 
+				// If we have a culture that doesn't specify country/region, we want to match its specific cultures if it's not found.
+				// For example, if we have es and it's not found, we want to match es-MX
+				if (culture.IsNeutralCulture)
+				{
+					var specificCultures = CultureInfo.GetCultures(CultureTypes.SpecificCultures).Where(c => c.Parent.Equals(culture));
```
jeromelaban, line 191: Very nice! WinUI seems to be ordering the cultures alphabetically; we should do the same so that es-ES is used instead of es-MX. Make sure to revalidate this is really the case on Windows, though.

Youssef1313: @jeromelaban In UWP, `es-MX` is used, not `es-ES`. It matches the last culture sorted alphabetically.

`GetCultures` already returns cultures sorted. The thing is that I was expecting my code to pick `es-ES` (which is wrong). But actually the code did the correct thing and picked `es-MX` matching UWP.

I don't know why, but it seems to work just like UWP 😄

jeromelaban: Ok then, let's keep it this way then! Thanks :)

jeromelaban (approved):

jeromelaban:
The test run is failing on WebAssembly:

image

https://dev.azure.com/uno-platform/Uno%20Platform/_TestManagement/Runs?runId=1567912&_a=resultSummary&resultId=100000

Youssef1313:
Only Wasm fails? 😨 That looks weird. I'll take a look

Youssef1313:
It looks like `GetCultures` isn't consistent regarding the ordering across different .NET runtimes. Even different runtimes can have different number of cultures. But the commonly used cultures should always exist.

jeromelaban:
In this case, we should re-order the results in all cases so it fits our scenario.

Youssef1313:
> In this case, we should re-order the results in all cases so it fits our scenario.

Yep that's what I did in e5386e8

PR #6937, opened 2021-09-02
perf: ImplementedRoutedEvents evaluated at compile-time via a source generator

### LINKED ISSUE #4288: We should definitely generate code instead of doing a check at runtime...
We should definitely generate code instead of doing a check at runtime...

_Originally posted by @carldebilly in https://github.com/unoplatform/uno/pull/4270#discussion_r506385978_

### LINKED ISSUE #6879: [Perf] GetImplementedRoutedEvents should use a generated implementation
### Current behavior

The following method should be generated for types in Uno:

https://github.com/unoplatform/uno/blob/b7f19b08fd276155868e9a95c29714af80402945/src/Uno.UI/UI/Xaml/Controls/Control/Control.cs#L995

Types outside of Uno will likely still need to use reflection for types defined outside of Uno

### Expected behavior

_No response_

_No response_

### Workaround

_No response_

### Works on UWP/WinUI

_No response_

### Environment

_No response_

### NuGet package version(s)

_No response_

### Affected platforms

_No response_

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

Youssef1313:
GitHub Issue (If applicable): closes #6879, closes #4288

## PR Type

- Performance

## PR Checklist

## Other information


src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
```diff
@@ -59,6 +59,157 @@ private void InitializeControl()
...
+				if (_cachedImplementedRoutedEvent != null)
+				{
+					return _cachedImplementedRoutedEvent.Value;
+				}
+
+				var result = RoutedEventFlag.None;
+				var type = this.GetType();
+
+				var baseClass = type.BaseType;
+				if (baseClass == null || type == typeof(Control))
```
Youssef1313, line 77: ⚠️ ⚠️ I removed `typeof(UIElement)` check since it seems unreachable. `Control` inherits from `UIElement`. `GetType` should return the more specific type, so it should never return `UIElement` based on my understanding. Let me know if I'm wrong here.

cc @carldebilly

Youssef1313 (commented):
I'll add tests soon.

src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
```diff
@@ -990,152 +1138,6 @@ protected virtual void OnBorderBrushChanged(Brush oldValue, Brush newValue)
...
-
-			if (GetIsEventOverrideImplemented(type, nameof(OnGotFocus), _routedArgsType))
-			{
-				result |= RoutedEventFlag.GotFocus;
-			}
-
-			return ImplementedRoutedEvents[type] = result;
-		}
-
 		private static bool GetIsEventOverrideImplemented(Type type, string name, Type[] args)
```
Youssef1313, line 1141: The existing reflection-based solution will likely return false for:

https://github.com/unoplatform/uno/blob/cd183d610dd800ee5a48af8008d72d8f8a16c40f/src/Uno.UI/UI/Xaml/Controls/ToggleSwitch.cs#L81

@carldebilly Can you confirm if this is the intended behavior?

src/SourceGenerators/Uno.UI.SourceGenerators/ImplementedRoutedEvents/ImplementedRoutedEventsGenerator.cs
```diff
@@ -0,0 +1,183 @@
...
+						}
+					}
+
+					GenerateCode(type, list);
+				}
+			}
+
+			private bool IsEventOverrideImplemented(INamedTypeSymbol type, string name, INamedTypeSymbol arg)
+			{
+				var method = type.GetMembers(name).SingleOrDefault(
```
Youssef1313, line 162: ⚠️ ⚠️ Likely I'll need to retrieve members from the inheritance hierarchy until reaching `Control` (not including it)

Youssef1313:
@jeromelaban This is ready for review now.

src/SourceGenerators/Uno.UI.SourceGenerators/ImplementedRoutedEvents/ImplementedRoutedEventsGenerator.cs
```diff
@@ -0,0 +1,208 @@
...
+				var implementedRoutedEventsSymbol = controlSymbol.GetMembers(ImplementedRoutedEventsPropertyName).SingleOrDefault(m => m.Kind == SymbolKind.Property) as IPropertySymbol;
+				if (implementedRoutedEventsSymbol is null)
+				{
+					return;
+				}
+
+				if (!implementedRoutedEventsSymbol.ContainingAssembly.IsSameAssemblyOrHasFriendAccessTo(context.Compilation.Assembly))
+				{
+					return;
+				}
```
Youssef1313, line 48: This prevents assemblies that can't access Uno.UI internals from generating code because the `ImplementedRoutedEvents` property is internal.

I'm thinking of one of two enhancements here:

- Make the property public (Uno-specific API) to let user-custom controls benefit from the performance improvement; **or**
- If we can't do the first suggestion, give internals visible to Uno.UI.Maps so that `MapPresenter` can benefit from this (probably there are other assemblies that can benefit too?)

carldebilly (commented):
Would be nice to find a way to test this automatically. Bugs or regressions in this code could potentially become very expensive and difficult to diagnose.

src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
carldebilly: RoutedEvents are not exclusive to `Control`.  Ok some events are only on this type, but not all of them.

Youssef1313: @carldebilly Can you elaborate more on what you mean here?

src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
```diff
@@ -990,152 +1140,6 @@ protected virtual void OnBorderBrushChanged(Brush oldValue, Brush newValue)
 		private static readonly Type[] _manipInertiaArgsType = new[] { typeof(ManipulationInertiaStartingRoutedEventArgs) };
 		private static readonly Type[] _manipCompletedArgsType = new[] { typeof(ManipulationCompletedRoutedEventArgs) };
 
-		protected static RoutedEventFlag GetImplementedRoutedEvents(Type type)
```
Youssef1313, line 993: This is going to be a breaking change :/
Even if the method is kept, it will have to be an instance virtual method.

src/SourceGenerators/Uno.UI.SourceGenerators/ImplementedRoutedEvents/ImplementedRoutedEventsGenerator.cs
carldebilly: In my opinion, it can be made public. There's no reason to restrict performance improvements to Uno only.

Complex libraries like Telerik, Syncfusion, WinToolKit should benefit from this as well.

src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
carldebilly: It was designed to be used for generated code eventually...  We just followed another path.

Youssef1313: @carldebilly Do you mean a breaking change would be okay here?

src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
carldebilly: I mean any FrameworkElement can generate RoutedEvent...  but I just realized only `Control` is providing virtual methods for them.  So you can ignore my comment :-) :-).

src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
carldebilly: I can't imagine any case where client code could implement or call those...  So I don't see any problem removing this method.

Youssef1313:
@carldebilly Can you check the way I'm testing this? Obviously the current tests are not enough, I'll expand them, just asking about the how you like the approach :)

Note to self: I need to handle nested classes case. Wondering if generators like native ctor supports this? @carldebilly do you know?

Youssef1313:
Pinging @carldebilly @jeromelaban for review.

carldebilly (approved):

jeromelaban:
I'm a bit bother by the introduction of an override specific to Uno on controls. If we ever need to change the signature, it'll be troublesome and a breaking change.

An intermediate change would be to use the generator for types inside of Uno only, and make the virtual method `private protected`.

As for making it available to external types, maybe with a registry similar to the one defined with bindable metadata, or by using a syntax like this in the generated type:

```csharp
private static readonly RoutedEventFlag __internal_implemented_routed_event = 
   ControlHelper.RegisterImplementedRoutedEvents(...RoutedEventFlag...);
```
The value would serve nothing but allow the execution in the `.cctor` without having to define one, and the control helper would avoid making the Control API surface wider.

Also, we're already caching the result in a dictionary, that would repurpose it.

Youssef1313:
> An intermediate change would be to use the generator for types inside of Uno only, and make the virtual method private protected.

That would be a breaking change too since there is already an existing method `protected static GetImplementedRoutedEvents` on Control. Do you mean you're okay with this as "one-time" breaking change, but want to avoid potential breaking changes later? or there shouldn't be breaking changes here at all?

Youssef1313:
- I made it `private protected`, not sure if that's enough @jeromelaban 
- CI is failing with: `Removed method Uno.UI.Xaml.RoutedEventFlag Windows.UI.Xaml.Controls.Control.GetImplementedRoutedEvents(System.Type type) not found in ignore set.`, but I added that to `PackageDiffIgnore` 😕

jeromelaban:
@Youssef1313 You added it to the version block that is not the latest available (3.10.7). It's going to change soon, so it'll have to be updated again anyways :)

For the private protected, that's not enough. The generator will try to be executed in apps, which won't have access to the base for the generated method. I think we should go the helper route, so all types can benefit from it, not just the internal types.

Youssef1313:
> The generator will try to be executed in apps, which won't have access to the base for the generated method.

This shouldn't happen since I have:

```csharp
				if (!getImplementedRoutedEventsSymbol.ContainingAssembly.IsSameAssemblyOrHasFriendAccessTo(context.Compilation.Assembly))
				{
					return;
				}
```

---

> I think we should go the helper route, so all types can benefit from it, not just the internal types.

I need some more info on that. The only `ControlHelper` I found is used in RuntimeTests only.

jeromelaban:
> > The generator will try to be executed in apps, which won't have access to the base for the generated method.
> 
> This shouldn't happen since I have:
> 
> ```cs
> 				if (!getImplementedRoutedEventsSymbol.ContainingAssembly.IsSameAssemblyOrHasFriendAccessTo(context.Compilation.Assembly))
> 				{
> 					return;
> 				}
> ```
> 
> > I think we should go the helper route, so all types can benefit from it, not just the internal types.
> 
> I need some more info on that. The only `ControlHelper` I found is used in RuntimeTests only.

Ok got it for the friends assembly validation. For the control helper, you'll need to create it. It's just a simple class that forwards the access to the internal method that provides access to the `GetImplementedRoutedEvents` implementation.

MartinZikmund:
@Youssef1313 sorry, missed this one and made a duplicate issue (#8411). Is there a chance you will get to this soon or should I pick it up? Important note - as part of my implementation of `BringIntoView` support I noticed the `OnBringIntoView`  virtual method is not on `Control` but on `UIElement` (along with a few other "handler" methods) - so this will need to be taken into account.

src/Uno.UI/UI/Xaml/UIElementGeneratedProxy.cs
```diff
@@ -0,0 +1,21 @@
...
+	{
+		lock (_implementedRoutedEvents)
+		{
+			_implementedRoutedEvents[type] = routedEventFlags;
+			return routedEventFlags;
+		}
+	}
+
+	internal static bool TryGetImplementedRoutedEvents(Type type, out RoutedEventFlag routedEventFlags) =>
+		_implementedRoutedEvents.TryGetValue(type, out routedEventFlags);
```
Youssef1313, line 20: Is using https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2?view=net-6.0 better?

src/Uno.UI/UI/Xaml/UIElementGeneratedProxy.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Uno.UI.Xaml;
+
+public static class UIElementGeneratedProxy
```
Youssef1313, line 6: @MartinZikmund Do we expect this to be used by consumers, or only generators?
If only generators, we should add `[EditorBrowsable(EditorBrowsableState.Never)]`

src/SourceGenerators/Uno.UI.SourceGenerators/ImplementedRoutedEvents/ImplementedRoutedEventsGenerator.cs
```diff
@@ -0,0 +1,251 @@
...
+
+			// Keep containing namespace here to avoid controls defined in both WUXC and MUXC from one overwriting the other.
+			var generatedSource = builder.ToString();
+			var generatedFileName = $"{GetFullMetadataNameForFileName(type)}_ImplementedRoutedEvents.g.cs";
+			_context.AddSource(generatedFileName, generatedSource);
+		}
+
+		private static void WriteClass(IIndentedStringBuilder builder, INamedTypeSymbol type, string routedEventFlags)
+		{
+			// TODO MZ: Handle generics
```
Youssef1313, line 233: @MartinZikmund Is `Given_Generic_Control_In_Global_Namespace` test enough to allow us to delete this? Or is there something else missing?

src/Uno.UI/UI/Xaml/UIElement.RoutedEvents.cs
```diff
@@ -79,7 +79,20 @@ namespace Windows.UI.Xaml
...
 
-	Note: this class is handling all this flow except [1] and [2]. */
+	Note: this class is handling all this flow except [1] and [2].
+	
+
+	Steps when enabling a new RoutedEvent:
+
+	1. Add/uncomment the routed event flag in RoutedEventFlag class.
+	2. Add the event in the proper group within RoutedEventFlagExtensions.
+	3. Move the RoutedEvent property and the event itself from generated partial class to here and initialize accodringly.
```
agneszitte-nventive, line 89:
```suggestion
	3. Move the RoutedEvent property and the event itself from generated partial class to here and initialize accordingly.
```

src/Uno.UI/UI/Xaml/UIElement.RoutedEvents.cs
```diff
@@ -79,7 +79,20 @@ namespace Windows.UI.Xaml
...
+	4. Add the exact event type handling to UIElement.InvokeHandler method.
+	5. If the event does not have a OnXYZ method on UIElement, FrameworkElement or Control, you are done. Otherwise:
+	6. Update the code in ImplementedRoutedEventsGenerator to generate the flag correctly when one of the aforementioned classes has a override.
```
agneszitte-nventive, line 92:
```suggestion
	6. Update the code in ImplementedRoutedEventsGenerator to generate the flag correctly when one of the aforementioned classes has an override.
```

src/Uno.UI/UI/Xaml/UIElement.RoutedEvents.cs
```diff
@@ -79,7 +79,20 @@ namespace Windows.UI.Xaml
...
+	7. Add the handler attachment code in SubscribeToOverridenRoutedEvents in UIElement, FrameworkElement or Control (depending on which level the method is declared).
```
agneszitte-nventive, line 93:
```suggestion
	7. Add the handler attachment code in SubscribeToOverridenRoutedEvents in UIElement, FrameworkElement, or Control (depending on which level the method is declared).
```

src/Uno.UI/UI/Xaml/UIElement.RoutedEvents.cs
```diff
@@ -79,7 +79,20 @@ namespace Windows.UI.Xaml
...
+	Note: this class is handling all this flow except [1] and [2].
+	
+
+	Steps when enabling a new RoutedEvent:
+
+	1. Add/uncomment the routed event flag in RoutedEventFlag class.
+	2. Add the event in the proper group within RoutedEventFlagExtensions.
+	3. Move the RoutedEvent property and the event itself from generated partial class to here and initialize accodringly.
+	4. Add the exact event type handling to UIElement.InvokeHandler method.
+	5. If the event does not have a OnXYZ method on UIElement, FrameworkElement or Control, you are done. Otherwise:
```
agneszitte-nventive, line 91:
```suggestion
	5. If the event does not have an OnXYZ method on UIElement, FrameworkElement, or Control, you are done. Otherwise:
```

src/Uno.UI.RuntimeTests/Tests/ImplementedRoutedEventArgsGeneratorTest/ImplementedRoutedEventArgsGeneratorTests.cs
```diff
@@ -0,0 +1,34 @@
...
+		public void Given_Control()
+		{
+			AssertRoutedEvent(typeof(Control), RoutedEventFlag.None);
+		}
+
+		[TestMethod]
+		public void Given_Button()
+		{
+			// It's okay to adjust the following assert if more events are needed and implemented in the future.
+			// This test is focused more of the source generator rather than Button functionality.
```
agneszitte-nventive, line 29:
```suggestion
			// This test is focused more on the source generator rather than the Button functionality.
```

src/SourceGenerators/Uno.UI.SourceGenerators/ImplementedRoutedEvents/ImplementedRoutedEventsGenerator.cs
```diff
@@ -0,0 +1,251 @@
...
+		}
+
+		/// <summary>
+		/// Checks whether the given TypeDeclarationSyntax is partial, and all parents are also partial
+		/// </summary>
+		private static bool IsPartialChainOfTypes(TypeDeclarationSyntax typeDeclaration)
+		{
+			if (typeDeclaration.Modifiers.Any(SyntaxKind.PartialKeyword))
+			{
+				// We have partial modifier, now we check the parent.
```
agneszitte-nventive, line 152:
```suggestion
				// We have a partial modifier, now we check the parent.
```

Youssef1313:
@jeromelaban @MartinZikmund Can you take a look please? Thanks!

src/Uno.UI/UI/Xaml/UIElementGeneratedProxy.cs
MartinZikmund: I think @jeromelaban preferred the locking here

src/Uno.UI/UI/Xaml/UIElementGeneratedProxy.cs
MartinZikmund: Indeed, good idea

src/SourceGenerators/Uno.UI.SourceGenerators/ImplementedRoutedEvents/ImplementedRoutedEventsGenerator.cs
MartinZikmund: I guess it should be 🤔

src/Uno.UI/UI/Xaml/UIElementGeneratedProxy.cs
Youssef1313, line 6:
```suggestion
[EditorBrowsable(EditorBrowsableState.Never)]
public static class UIElementGeneratedProxy
```

src/Uno.UI/UI/Xaml/UIElementGeneratedProxy.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using System.Collections.Generic;
```
Youssef1313, line 2:
```suggestion
using System.Collections.Generic;
using System.ComponentModel;
```

src/SourceGenerators/Uno.UI.SourceGenerators.net6.Tests/ImplementedRoutedEventsGeneratorTests/Given_ImplementedRoutedEventsGenerator.cs
```diff
@@ -0,0 +1,146 @@
...
+{
+	using Verify = CSharpSourceGeneratorVerifier<ImplementedRoutedEventsGenerator>;
+
+	[TestClass]
+	public class Given_ImplementedRoutedEventsGenerator
+	{
+		private static readonly ReferenceAssemblies s_defaultWithUno = ReferenceAssemblies.Default.AddPackages(
+			ImmutableArray.Create(new PackageIdentity("Uno.UI", "4.4.20"))); // UPDATE with a version that has a UIElementGeneratedProxy available.
+
+		// TODO: Remove and use s_defaultWithUno when we have a version with UIElementGeneratedProxy available.
```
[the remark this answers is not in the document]
Youssef1313: @jeromelaban Yes. Just to get rid of the fake `UIElementGeneratedProxy` we define in the test. There is no way to address that in this PR other than trying to consume local Uno.UI packages instead of from NuGet, but I think it's not worth doing that. Following up should be okay for now.

src/Uno.UI/UI/Xaml/UIElementGeneratedProxy.cs
Youssef1313: Let's keep the lock for now until we validate.

carldebilly:
@Youssef1313 I NEVER thought this feature would be that complicated to implement. Anyway, it seems ready to merge :-)

MartinZikmund (approved):

MartinZikmund:
Letssss goooooo! 🚀🚀🚀

MartinZikmund:
@Youssef1313 can you please try rebasing this? It seems there are some conflicts and build errors 👀

carldebilly:
Finally merged...  that was a long road. :-)

PR #6940, opened 2021-09-02
Implement CommandBar and AppBar

### LINKED ISSUE #6195: WinUI CommandBar

CommandBar code from WinUI should be imported

## Why is this needed:

Share CommandBar logic across all platforms

## For which Platform:

- [ ] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] macOS
- [ ] Skia
  - [ ] WPF
  - [ ] GTK (Linux)
  - [ ] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

kazo0:
GitHub Issue (If applicable): closes #6195

## PR Type

- Feature

Current `Windows.UI.Xaml.Controls.CommandBar` lacks many of the features from WinUI

## What is the new behavior?

There is now a `Microsoft.UI.Xaml.Controls.CommandBar` and `Microsoft.UI.Xaml.Controls.AppBar` that is a port of the WinUI code. 

## PR Checklist


robloo:
It's great to see this come over from WinUI!

src/Uno.UI.FluentTheme/Resources/PriorityDefault/CommandBar_themeresources.xaml
```diff
@@ -1,1716 +1,2668 @@
...
-        </ResourceDictionary>
-    </ResourceDictionary.ThemeDictionaries>
+					xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">
+	<ResourceDictionary.ThemeDictionaries>
+		<ResourceDictionary x:Key="Default">
+			<x:Double x:Key="CommandBarOverflowMinWidth">160</x:Double>
+			<x:Double x:Key="CommandBarOverflowTouchMinWidth">240</x:Double>
+			<x:Double x:Key="CommandBarOverflowMaxWidth">480</x:Double>
+			<x:Double x:Key="CommandBarOverflowMaxHeight">198</x:Double>
+			<StaticResource x:Key="CommandBarBackground"
```
davidjohnoliver, line 10: try not to change the white spacing in Uno.UI.FluentTheme resources, since it makes it harder to manage upstream changes from the MUX repo.

It also makes it impossible to see what actual functional changes you've made here... whatever they are, prefix them with a big `<!-- UNO WORKAROUND: [Explanation] -->` comment, or else they'll get lost the first time somebody copies in the latest version from upstream.

src/Uno.UI/Microsoft/UI/Xaml/Controls/CommandBar/CommandBar_Partial.cs
```diff
@@ -0,0 +1,3332 @@
...
+using static Microsoft.UI.Xaml.Controls._Tracing;
+using Uno.UI.Xaml.Input;
+using Windows.UI.Xaml.Controls;
+using Windows.UI.Xaml;
+using Popup = Windows.UI.Xaml.Controls.Popup;
+using CommandBarTemplateSettings = Microsoft.UI.Xaml.Controls.Primitives.CommandBarTemplateSettings;
+
+namespace Microsoft.UI.Xaml.Controls
+{
+	partial class CommandBar : IMenu
```
davidjohnoliver, line 29: I'm not sure we've settled on a firm convention, but it seems more common to use the C# naming convention of having the name of the class before the first dot, then indicating the original source file after the first dot and before the file extension. So in this case, something like 'CommandBar.partial.cs'.

davidjohnoliver (commented):
So if I understand correctly, this leaves the existing WUX.CommandBar for backwards compatibility on Android and iOS, and adds MUX.CommandBar (and AppBar) using WinUI code, for proper support on WASM, Skia etc?

I have a few questions/concerns:

1 - Is that correct?

2 - What's going to happen with the following XAML:

```xml
<Page x:Class="UnoTestbed56.MainPage"
	  xmlns:local="using:UnoTestbed56"
	  xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
	  mc:Ignorable="d">

	<Grid Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">
		<CommandBar x:Name="MyCommandBar" Content="Main page" />
	</Grid>
</Page>
```

Will `MyCommandBar` be a WUX.CommandBar or a MUX.CommandBar? If the latter, it's a breaking change for apps targeting mobile and using native frame navigation. If the former, it's going to not work properly out of the box on WASM in a confusing-to-diagnose way. We may even want special-casing for this in the XAML code generation.

3 - What's going to happen with Uno.WinUI, where the WUX namespace doesn't exist?

kazo0:
> So if I understand correctly, this leaves the existing WUX.CommandBar for backwards compatibility on Android and iOS, and adds MUX.CommandBar (and AppBar) using WinUI code, for proper support on WASM, Skia etc?
> 
> I have a few questions/concerns:
> 
> 1 - Is that correct?
> 
> 2 - What's going to happen with the following XAML:
> 
> ```
> <Page x:Class="UnoTestbed56.MainPage"
> 	  xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
> 	  xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
> 	  xmlns:local="using:UnoTestbed56"
> 	  xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
> 	  xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
> 	  mc:Ignorable="d">
> 
> 	<Grid Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">
> 		<CommandBar x:Name="MyCommandBar" Content="Main page" />
> 	</Grid>
> </Page>
> ```
> 
> Will `MyCommandBar` be a WUX.CommandBar or a MUX.CommandBar? If the latter, it's a breaking change for apps targeting mobile that use native frame navigation. If the former, it's going to not work properly out of the box on WASM in a confusing-to-diagnose way. We may even want special-casing for this in the XAML code generation.
> 
> 3 - What's going to happen with Uno.WinUI, where the WUX namespace doesn't exist?

1 - Correct
2 - The former, wouldn't it work the same way that it does now if you put a WUX CommandBar in a page and ran the WASM app. But yes I agree that keeping both causes problems.
3 - In Uno.WinUI, only the MUX CommandBar will exist I believe, right @jeromelaban?

src/Uno.UWP/Devices/Bluetooth/BluetoothAddressType.cs
```diff
@@ -1,10 +1,10 @@
...
-{
-	public   enum BluetoothAddressType 
-	{
-		Public,
-		Random,
-		Unspecified,
-	}
-}
+﻿
+namespace Windows.Devices.Bluetooth
```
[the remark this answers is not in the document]
MartinZikmund: Rebase on latest master could help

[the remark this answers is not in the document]
kazo0: Just did, still here. Strange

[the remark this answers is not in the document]
kazo0: Weird, I can't seem to revert these changes, they just keep showing up as changes in the PR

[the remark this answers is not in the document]
davidjohnoliver: even if they show up as local changes, you can just not stage them to be committed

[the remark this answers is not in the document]
kazo0: I guess I don't know what I am doing but, for the life of me, I can't seem to remove these changes. They just keep showing up on here

src/Uno.UI.FluentTheme.v1/Resources/Version1/Priority06/RevealBrush_themeresources.xaml
```diff
@@ -1,4 +1,4 @@
-<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
+﻿<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
```
MartinZikmund, line 1: Why did this file need to change? Is similar change needed for V2 too?

src/Uno.UI.FluentTheme.v1/Resources/Version1/PriorityDefault/CommandBar_themeresources.xaml
```diff
@@ -955,7 +1050,7 @@
         </Setter>
     </Style>
 
-    <Style x:Key="CommandBarWithoutRevealStyle" TargetType="CommandBar">
+    <Style x:Key="CommandBarWithoutRevealStyle" TargetType="muxc:CommandBar">
```
MartinZikmund, line 1053: I probably forgot the reason for this but - why does `CommandBar` need to be in MUX namespace? The problem I see is that this way it will not work in Uno.UI, only in Uno.WinUI. If I am not wrong, it would mean that this control would never work in UWP (because trying to use `muxc:CommandBar` in XAML would fail during compilation). Wouldn't it be possible to completely replace the fake `CommandBar` we have now with this one and then possibly move the existing `CommandBar` implementation to something like `UnoCommandBar`? I understand this would be a bigger breaking change, but we will have 4.0, so it would probably be worth it. The ported code is not technically a MUX `CommandBar` (akin to `TreeView` or `TabView`), but rather the WUX `CommandBar` - the control did not change in MUX and is not specific to it. @jeromelaban

MartinZikmund: The only way this could then be used would be something like:

```
<not_win:Grid>
    <muxc:CommandBar x:Name="MyCommandBar" ... />
</not_win:Grid>
<win:Grid>
    <CommandBar x:Name="MyCommandBar" ... />
</win:Grid>
```

Which would work, but it would be quite ugly, especially as `CommandBar` is one of the controls which are used quite often...

src/Uno.UI.FluentTheme.v2/Resources/Version2/PriorityDefault/AppBarButton_themeresources.xaml
```diff
@@ -107,15 +108,15 @@
     <Thickness x:Key="AppBarButtonInnerBorderCompactMargin">2,6,2,22</Thickness>
     <Thickness x:Key="AppBarButtonInnerBorderOverflowMargin">4,0,4,0</Thickness>
 
-    <Style TargetType="AppBarButton" BasedOn="{StaticResource DefaultAppBarButtonStyle}" />
+    <Style TargetType="local:AppBarButton" BasedOn="{StaticResource DefaultAppBarButtonStyle}" />
```
MartinZikmund, line 111: Similar question here for `AppBar`, `AppBarButton`, etc. These are not recommended for use in WinUI, so having them in `Windows.UI.Xaml.Controls` in "UWP mode" would make more sense to me.

src/Uno.UI.RuntimeTests/IntegrationTests/common/TestServices.KeyboardHelper.cs
```diff
@@ -1,49 +1,283 @@
...
+				{"GamePadRightThumbStickUp",    VirtualKey.GamepadRightThumbstickUp},
+				{"GamePadRightThumbStickLeft",  VirtualKey.GamepadRightThumbstickLeft},
+				{"GamePadMenu",                 VirtualKey.GamepadMenu},
+			};
+
+
+			public static void PressKeySequence(string keys, UIElement element = null)
+			{
+#if !NETFX_CORE
+				if (string.IsNullOrEmpty(keys) || element == null)
```
MartinZikmund, line 123: Cool!

src/Uno.UI/Microsoft/UI/Xaml/Controls/AppBar/AppBar.Partial.cs
```diff
@@ -0,0 +1,1701 @@
+﻿#nullable enable
```
MartinZikmund, line 1: Add the copyright comment and source file here and in other ported files (see for example `\src\Uno.UI\UI\Xaml\Controls\Primitives\ButtonBase\ButtonBase.mux.cs`)

src/Uno.UI/Microsoft/UI/Xaml/Controls/AppBar/AppBar.Properties.cs
```diff
@@ -0,0 +1,96 @@
...
+				typeof(AppBar),
+				new FrameworkPropertyMetadata(default(LightDismissOverlayMode))
+			);
+
+		#endregion
+
+		#region TemplateSettings
+		public AppBarTemplateSettings TemplateSettings
+		{
+			get { return (AppBarTemplateSettings)this.GetValue(TemplateSettingsProperty); }
```
MartinZikmund, line 89: Use `=>`

src/Uno.UI/Microsoft/UI/Xaml/Controls/AppBar/AppBarAutomationPeer.cs
```diff
@@ -0,0 +1,236 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.UI.Xaml.Controls;
+using Windows.UI.Xaml.Automation;
+using Windows.UI.Xaml.Automation.Peers;
+
+namespace Microsoft.UI.Xaml.Automation.Peers
+{
+	public partial class AppBarAutomationPeer : global::Windows.UI.Xaml.Automation.Peers.FrameworkElementAutomationPeer, global::Windows.UI.Xaml.Automation.Provider.IToggleProvider, global::Windows.UI.Xaml.Automation.Provider.IExpandCollapseProvider, global::Windows.UI.Xaml.Automation.Provider.IWindowProvider
```
MartinZikmund, line 10: The `global::` are probably not necessary, add `using` statements for those namespaces to make it cleaner 🙂

src/Uno.UI/Microsoft/UI/Xaml/Controls/AppBar/AppBarButtonTemplateSettings.cs
```diff
@@ -0,0 +1,23 @@
...
+		}
+
+		public double KeyboardAcceleratorTextMinWidth
+		{
+			get => (double)GetValue(KeyboardAcceleratorTextMinWidthProperty);
+			internal set => SetValue(KeyboardAcceleratorTextMinWidthProperty, value);
+		}
+
+		internal static DependencyProperty KeyboardAcceleratorTextMinWidthProperty { get; } =
+			DependencyProperty.Register("KeyboardAcceleratorTextMinWidth", typeof(double), typeof(AppBarButtonTemplateSettings), new FrameworkPropertyMetadata(0.0));
```
MartinZikmund, line 21: You can use `nameof(KeyboardAcceleratorTextMinWidth)` here for the first parameter

src/Uno.UI.FluentTheme.v1/Resources/Version1/PriorityDefault/CommandBar_themeresources.xaml
MartinZikmund: Maybe both the Uno-legacy `CommandBar` and WinUI/UWP CommandBar could live in one class too (similarly as we have the "native" `Button`, "native` `ToggleSwitch`, etc. supported)

MartinZikmund: Ah, I notice now that @davidjohnoliver had similar questions 😀

kazo0: Yeah I agree it's going to be ugly to use with Windows :( But we didn't want to make a breaking change by moving the current WUX CommandBar since it is used for native UINavigationBar/Toolbar rendering on iOS/Android.

@jeromelaban may be able to weigh in here. But I am still open to having the current CommandBar be moved to a Legacy namespace so the one true WinUI CommandBar can live in WUX

MartinZikmund (approved):

PR #684, opened 2019-03-14
Align TextBox events handling behavior with Windows

dr1rrb:
## PR Type
What kind of change does this PR introduce?
- Feature

## What is the current behavior?
Touch and key events are not handled as windows does

## What is the new behavior?
* `PointerRelease` events are now marked as handled by the `TextBox`
* `KeyDown` events that are changing the cursor position (left/right/top/bottom/home/end) are now marked as handled by the `TextBox`

## PR Checklist

Please check if your PR fulfills the following requirements:

- [x] Tested code with current [supported SDKs](../README.md#supported)
- [ ] Docs have been added/updated which fit [documentation template](https://github.com/nventive/Uno/blob/master/doc/.feature-template.md). (for bug fixes / features)
- [ ] [Unit Tests and/or UI Tests](doc/articles/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [x] Contains **NO** breaking changes
- [x] Updated the [Release Notes](https://github.com/nventive/Uno/tree/master/doc/ReleaseNotes)
- [x] Associated with an issue (GitHub or internal)

Internal Issue (If applicable):
https://nventive.visualstudio.com/Umbrella/_workitems/edit/149028

PR #6943, opened 2021-09-02
fix: Propagate key events on Wasm to fix TextBox keyboard navigation

### LINKED ISSUE #6933: [Wasm] The keyboard navigation buttons do not work in a TextBox
### Current behavior

The Home, End and arrow keyboard keys don't have an effect in a `TextBox`.

See https://playgroundcanary.z19.web.core.windows.net/#textbox in Uno 3.10-dev.

### Expected behavior

Home and End, or other navigation keys move the cursor,

Go to https://playgroundcanary.z19.web.core.windows.net/#textbox, enter some text then press the home button. Notice that the cursor stays at the end.

### Workaround

_No response_

### Works on UWP/WinUI

Yes

### Environment

### NuGet package version(s)

_No response_

### Affected platforms

WebAssembly

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

Youssef1313:
GitHub Issue (If applicable): closes #6933

## PR Type

- Bugfix

## PR Checklist

## Other information


src/Uno.UI/UI/Xaml/UIElement.wasm.cs
```diff
@@ -622,8 +622,8 @@ internal void MoveChildTo(int oldIndex, int newIndex)
 
 			RegisterEventHandler(
 				domEventName,
-				handler: new RoutedEventHandlerWithHandled((snd, args) => RaiseEvent(routedEvent, args)),
-				invoker: GenericEventHandlers.RaiseRoutedEventHandlerWithHandled,
+				handler: new RoutedEventHandler((snd, args) => RaiseEvent(routedEvent, args)),
+				invoker: GenericEventHandlers.RaiseRoutedEventHandler,
```
dr1rrb, line 626: This is not valid: you are only ignoring the [`Handled`](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.input.keyroutedeventargs.handled?view=winrt-20348) flag here. It means that a key event flag as handled would still bubble ...

dr1rrb (changes requested):

src/Uno.UI/UI/Xaml/UIElement.wasm.cs
Youssef1313: @dr1rrb Thanks! However, I don't feel I'm fully understanding the situation here (sorry). Is the suggestion is to modify this:

https://github.com/unoplatform/uno/blob/bc8d7fa884184c0aad3af1e4763d1f929757d503/src/Uno.UI/UI/Xaml/UIElement.GenericEventHandlers.wasm.cs#L56-L57

to check for `RoutedEventArgs.Handled` property, then decide whether to `return null` or `return handler(....)`?

Or get rid of `RaiseRoutedEventHandlerWithHandled` and use only `RaiseRoutedEventHandler` which then should behave differently based on `RoutedEventArgs.Handled` (ie, decide whether to `return null` or `return handler(....)`)?

Or something else?

dr1rrb: @Youssef1313 There is something that is really important about "routed" events, is that they are _bubbling_ from the top-most element (a.k.a. the leaf of the visual tree) up to the root of the visual tree. It means that for instance when you press the button of your mouse, the element right under the cursor will fire the `PointerPressed` event, then its parent will do the same, then its parent etc. until we reach the root of the visual tree. That's the case for almost all input events (pointers, keyboard), and a lot of so called "routed events" like `GotFocus` (with a noticeable exception of the [`PreviewKeyDown` event](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.uielement.previewkeydown?view=winrt-20348) which is the single _tunneling_ event on UWP, which means that the event is raised on the root first, then on children until we reach the leaf of the tree).

The args of those "routed" event have a `Handled` flag that allow any handler to stop that _bubbling_ at any time. This is really useful when, for instance, you have a button in a button: the "top most" button will handle the pointer events, so the `Click` will be fired only on one button.

In browsers we also have a second layer of "handling":
* `stopPropagation` is really the same thing as flagging an event `Handled` on UWP: it just stops the _bubbling_
* `preventDefault` which request to the browser to not apply the default behavior that it would normally do (scrolling, navigation, focus, etc.). 

Be aware that "stopping propagation" will **NOT** prevent the default behavior. Both flags are not linked in any way.
And the difference is really important. A good example is for the touch pointers: if an application decides to handle a `PointerMoved` event, we must not prevent the default behavior since for touch it will be to scroll the content (if possible).

Unfortunately we don't have a such distinction in UWP API (which is normal since there is no "default behavior" like that on UWP), so publicly we are mapping both to the single UWP `Handled` flag, which is enough so far. That's why we have [`HtmlEventDispatchResult.StopPropagation | HtmlEventDispatchResult.PreventDefault` here](https://github.com/unoplatform/uno/blob/9efab96fd59e28c68fa1430f7cbc74353ebc22bf/src/Uno.UI/UI/Xaml/UIElement.EventRegistration.wasm.cs#L155). And internally we are able to specify the `HtmlEventDispatchResult`, but only using the `RawEventHandler` for now.

In the https://github.com/unoplatform/uno/blob/bc8d7fa884184c0aad3af1e4763d1f929757d503/src/Uno.UI/UI/Xaml/UIElement.GenericEventHandlers.wasm.cs basically we just have an enumeration of supported strongly typed event handlers. Each handler was designed for some specific usage:
* The `EventHandler` is for non routed events
* The `RawEventHandler` is the more powerful since you are able to get the raw arguments without parsing, and explicitly specify the `HtmlEventDispatchResult`.
* The `RoutedEventHandler` is for routed event args that does not have the `Handled` flag (which should be really rare ... maybe it's  even for legacy support)
* The `ExceptionRoutedEventHandler` don't even know why it exists neither why it's named "routed"
* The `RoutedEventHandlerWithHandled` is the common one for routed events: it just returns a `bool` which is actually the `Handled` flag which is then converted in JS as "stop bubbling" AND "prevent default behavior" as describe above.

The main "fix" is not really in those files: It's to find which handler is flagging the key events as `Handled` and change that in order to make a distinction between "stop propagation" and "prevent default", but **only for that specific event handler**. If it's in the focus management like I suppose, I would assume that we just don't want to "prevent default" when the focused element is a `TextBox`.

Youssef1313: @dr1rrb Thank you a LOT for the detailed explanation! Really much appreciated.

If I got this right, the problem is in:

https://github.com/unoplatform/uno/blob/0789a59d33760ba6997640deeb50c03e20d1e676/src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.cs#L665-L688

However, the behavior seems like it was intentional, so I assume it's not as simple as removing the override. Any idea why this code used to exist?

Youssef1313: Looks like this was aligning the behavior with UWP: https://github.com/unoplatform/uno/pull/684

dr1rrb: Ok then from what I see here, we want to flag the events as "handled" to avoid bubbling in managed code (like in UWP does apparently). And in the PR we also made sure to **not** `preventDefault` when event was flagged as handled ... which is obviously no longer the case. So it means that in that event handler of the `TextBox` we should have a way to specify that on wasm, we don't want to `preventDefault`.

The easiest way would be to add the ability to set the `HtmlEventDispatchResult` on the [internal interface `IHandleableRoutedEventArgs`](https://github.com/unoplatform/uno/blob/master/src/Uno.UI/UI/Xaml/Input/IHandleableRoutedEventArgs.cs) (for `__WASM__` only) and then make sure to send the result back in event handlers to reach [that case](https://github.com/unoplatform/uno/blob/9efab96fd59e28c68fa1430f7cbc74353ebc22bf/src/Uno.UI/UI/Xaml/UIElement.EventRegistration.wasm.cs#L157). Setting `IHandleableRoutedEventArgs.Handled = true` would then just set `HtmlEventDispatchResult.StopPropagation | HtmlEventDispatchResult.PreventDefault` ... the issue is that there are a lot of classes which are implementing that interface, maybe the default interface implementation might help you here.

Youssef1313: @dr1rrb Thanks again! I'll take a look and see what I can do.

btw, the code above isn't as what you originally wrote. `AcceptsReturn` check didn't use to exist. I assume this could be a bug introduced later? Can you confirm?

Also, is there anyway to debug on Wasm? I was never able to get that working. `Console.WriteLine`s going to be very very tedious here.

Youssef1313: I wasn't able to add the ability to set `HtmlEventDispatchResult`. The problem I faced is that `RoutedEventHandlerWithHandled` delegate returns a `bool`. I tried to take a different (but relatively similar) approach. But it doesn't seem to work. Trying to figure that out.

src/Uno.UI/UI/Xaml/Input/IHandleableRoutedEventArgs.cs
```diff
@@ -11,5 +11,7 @@ internal interface IHandleableRoutedEventArgs
 		/// route from handling the same event again.
 		/// </summary>
 		bool Handled { get; set; }
+
+		bool ShouldPreventDefaultIfHandled { get; set; }
```
dr1rrb, line 15: The advantage to expose an enum like the `HtmlEventDispatchResult` instead of a `bool` is that we are future-proof if we need later to add another distinction between "managed handled" and "native handled" ... but your approach is faster and easier and as this interface is internal so that good 👍

I would however revert the `bool` so we won't forget to set it by default to `true` for new event args:
```suggestion
		/// <summary>
		/// [WASM ONLY] 
		/// By default an event flagged as <see cref="Handled"/> will prevent the "default behavior" of the browser.
		/// Setting this flag to `true` will make sure that this "default behavior" will not be suppressed.
		/// cf. https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault
		/// </summary>
		bool DoNotPreventDefaultIfHandled { get; set; }
```

src/Uno.UI/UI/Xaml/Controls/FocusEngagedEventArgs.cs
```diff
@@ -20,6 +20,8 @@ bool IHandleableRoutedEventArgs.Handled
 		{
 			get => Handled;
 			set => Handled = value;
-		}		
+		}
+
+		bool IHandleableRoutedEventArgs.ShouldPreventDefaultIfHandled { get; set; } = true;
```
dr1rrb, line 25:
```suggestion
		bool IHandleableRoutedEventArgs.DoNotPreventDefaultIfHandled { get; set; }
```

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.cs
```diff
@@ -666,6 +667,7 @@ protected override void OnPointerReleased(PointerRoutedEventArgs args)
 		protected override void OnKeyDown(KeyRoutedEventArgs args)
 		{
 			base.OnKeyDown(args);
+			((IHandleableRoutedEventArgs)args).ShouldPreventDefaultIfHandled = false;
```
dr1rrb, line 670:
```suggestion
		bool IHandleableRoutedEventArgs.DoNotPreventDefaultIfHandled { get; set; }
```

src/Uno.UI/UI/Xaml/DragDrop/DragEventArgs.cs
```diff
@@ -56,5 +56,7 @@ public Point GetPosition(UIElement relativeTo)
 
 		public DragOperationDeferral GetDeferral()
 			=> Deferral ??= new DragOperationDeferral();
+
+		bool IHandleableRoutedEventArgs.ShouldPreventDefaultIfHandled { get; set; } = true;
```
dr1rrb, line 60:
```suggestion
		bool IHandleableRoutedEventArgs.DoNotPreventDefaultIfHandled { get; set; }
```

src/Uno.UI/UI/Xaml/Input/CharacterReceivedRoutedEventArgs.cs
```diff
@@ -27,6 +27,8 @@ bool IHandleableRoutedEventArgs.Handled
 			set => Handled = value;
 		}
 
+		bool IHandleableRoutedEventArgs.ShouldPreventDefaultIfHandled { get; set; } = true;
```
dr1rrb, line 30:
```suggestion
		bool IHandleableRoutedEventArgs.DoNotPreventDefaultIfHandled { get; set; }
```

src/Uno.UI/UI/Xaml/Input/DoubleTappedRoutedEventArgs.cs
```diff
@@ -39,5 +39,7 @@ public Point GetPosition(UIElement relativeTo)
 				return _originalSource.GetPosition(_position, relativeTo);
 			}
 		}
+
+		bool IHandleableRoutedEventArgs.ShouldPreventDefaultIfHandled { get; set; } = true;
```
dr1rrb, line 43:
```suggestion
		bool IHandleableRoutedEventArgs.DoNotPreventDefaultIfHandled { get; set; }
```

src/Uno.UI/UI/Xaml/Input/GettingFocusEventArgs.cs
```diff
@@ -93,6 +93,8 @@ bool IChangingFocusEventArgs.Cancel
 			get => Cancel;
 		}
 
+		bool IHandleableRoutedEventArgs.ShouldPreventDefaultIfHandled { get; set; } = true;
```
dr1rrb, line 96:
```suggestion
		bool IHandleableRoutedEventArgs.DoNotPreventDefaultIfHandled { get; set; }
```

src/Uno.UI/UI/Xaml/Input/HoldingRoutedEventArgs.cs
```diff
@@ -33,7 +33,9 @@ bool IHandleableRoutedEventArgs.Handled
 		{
 			get => Handled;
 			set => Handled = value;
-		}		
+		}
+
+		bool IHandleableRoutedEventArgs.ShouldPreventDefaultIfHandled { get; set; } = true;
```
dr1rrb, line 38:
```suggestion
		bool IHandleableRoutedEventArgs.DoNotPreventDefaultIfHandled { get; set; }
```

src/Uno.UI/UI/Xaml/Input/KeyRoutedEventArgs.cs
```diff
@@ -18,6 +18,9 @@ internal KeyRoutedEventArgs(object originalSource, VirtualKey key, CorePhysicalK
 		}
 
 		public bool Handled { get; set; }
+
+		bool IHandleableRoutedEventArgs.ShouldPreventDefaultIfHandled { get; set; } = true;
```
dr1rrb, line 22:
```suggestion
		bool IHandleableRoutedEventArgs.DoNotPreventDefaultIfHandled { get; set; }
```

src/Uno.UI/UI/Xaml/Input/LosingFocusEventArgs.cs
```diff
@@ -93,6 +93,8 @@ bool IChangingFocusEventArgs.Cancel
 			get => Cancel;
 		}
 
+		bool IHandleableRoutedEventArgs.ShouldPreventDefaultIfHandled { get; set; } = true;
```
dr1rrb, line 96:
```suggestion
		bool IHandleableRoutedEventArgs.DoNotPreventDefaultIfHandled { get; set; }
```

src/Uno.UI/UI/Xaml/Input/ManipulationCompletedRoutedEventArgs.cs
```diff
@@ -31,6 +31,8 @@ internal ManipulationCompletedRoutedEventArgs(UIElement container, ManipulationC
 
 		public bool Handled { get; set; }
 
+		bool IHandleableRoutedEventArgs.ShouldPreventDefaultIfHandled { get; set; } = true;
```
dr1rrb, line 34:
```suggestion
		bool IHandleableRoutedEventArgs.DoNotPreventDefaultIfHandled { get; set; }
```

src/Uno.UI/UI/Xaml/Input/ManipulationDeltaRoutedEventArgs.cs
```diff
@@ -36,6 +36,8 @@ internal ManipulationDeltaRoutedEventArgs(UIElement container, GestureRecognizer
 
 		public bool Handled { get; set; }
 
+		bool IHandleableRoutedEventArgs.ShouldPreventDefaultIfHandled { get; set; } = true;
```
dr1rrb, line 39:
```suggestion
		bool IHandleableRoutedEventArgs.DoNotPreventDefaultIfHandled { get; set; }
```

src/Uno.UI/UI/Xaml/Input/ManipulationInertiaStartingRoutedEventArgs.cs
```diff
@@ -33,6 +33,8 @@ internal ManipulationInertiaStartingRoutedEventArgs(UIElement container, Manipul
 
 		public bool Handled { get; set; }
 
+		bool IHandleableRoutedEventArgs.ShouldPreventDefaultIfHandled { get; set; } = true;
```
dr1rrb, line 36:
```suggestion
		bool IHandleableRoutedEventArgs.DoNotPreventDefaultIfHandled { get; set; }
```

src/Uno.UI/UI/Xaml/Input/ManipulationStartedRoutedEventArgs.cs
```diff
@@ -32,6 +32,8 @@ internal ManipulationStartedRoutedEventArgs(UIElement container, GestureRecogniz
 
 		public bool Handled { get; set; }
 
+		bool IHandleableRoutedEventArgs.ShouldPreventDefaultIfHandled { get; set; } = true;
```
dr1rrb, line 35:
```suggestion
		bool IHandleableRoutedEventArgs.DoNotPreventDefaultIfHandled { get; set; }
```

src/Uno.UI/UI/Xaml/Input/ManipulationStartingRoutedEventArgs.cs
```diff
@@ -32,6 +32,8 @@ internal ManipulationStartingRoutedEventArgs(UIElement container, ManipulationSt
 
 		public bool Handled { get; set; }
 
+		bool IHandleableRoutedEventArgs.ShouldPreventDefaultIfHandled { get; set; } = true;
```
dr1rrb, line 35:
```suggestion
		bool IHandleableRoutedEventArgs.DoNotPreventDefaultIfHandled { get; set; }
```

src/Uno.UI/UI/Xaml/Input/NoFocusCandidateFoundEventArgs.cs
```diff
@@ -34,6 +34,8 @@ bool IHandleableRoutedEventArgs.Handled
 		{
 			get => Handled;
 			set => Handled = value;
-		}		
+		}
+
+		bool IHandleableRoutedEventArgs.ShouldPreventDefaultIfHandled { get; set; } = true;
```
dr1rrb, line 39:
```suggestion
		bool IHandleableRoutedEventArgs.DoNotPreventDefaultIfHandled { get; set; }
```

src/Uno.UI/UI/Xaml/Input/PointerRoutedEventArgs.cs
```diff
@@ -44,6 +44,8 @@ public IList<PointerPoint> GetIntermediatePoints(UIElement relativeTo)
 
 		public bool Handled { get; set; }
 
+		bool IHandleableRoutedEventArgs.ShouldPreventDefaultIfHandled { get; set; } = true;
```
dr1rrb, line 47:
```suggestion
		bool IHandleableRoutedEventArgs.DoNotPreventDefaultIfHandled { get; set; }
```

src/Uno.UI/UI/Xaml/Input/RightTappedRoutedEventArgs.cs
```diff
@@ -21,6 +21,8 @@ internal RightTappedRoutedEventArgs(UIElement originalSource, RightTappedEventAr
 
 		public bool Handled { get; set; }
 
+		bool IHandleableRoutedEventArgs.ShouldPreventDefaultIfHandled { get; set; } = true;
```
dr1rrb, line 24:
```suggestion
		bool IHandleableRoutedEventArgs.DoNotPreventDefaultIfHandled { get; set; }
```

src/Uno.UI/UI/Xaml/Input/TappedRoutedEventArgs.cs
```diff
@@ -25,6 +25,8 @@ internal TappedRoutedEventArgs(UIElement originalSource, TappedEventArgs args)
 
 		public bool Handled { get; set; }
 
+		bool IHandleableRoutedEventArgs.ShouldPreventDefaultIfHandled { get; set; } = true;
```
dr1rrb, line 28:
```suggestion
		bool IHandleableRoutedEventArgs.DoNotPreventDefaultIfHandled { get; set; }
```

src/Uno.UI/UI/Xaml/UIElement.EventRegistration.wasm.cs
```diff
@@ -185,6 +186,10 @@ private HtmlEventDispatchResult InnerDispatch(EventArgs eventArgs, string native
 						switch (handlerResult)
 						{
 							case bool isHandedInManaged when isHandedInManaged:
+								if (args is IHandleableRoutedEventArgs handleable && !handleable.ShouldPreventDefaultIfHandled)
```
dr1rrb, line 189:
```suggestion
								if (args is IHandleableRoutedEventArgs handleable && handleable.DoNotPreventDefaultIfHandled)
```

dr1rrb (approved):

src/Uno.UI/UI/Xaml/UIElement.wasm.cs
dr1rrb: > 
> Also, is there anyway to debug on Wasm? I was never able to get that working. `Console.WriteLine`s going to be very very tedious here.

You should be able just to press `F5` : https://platform.uno/blog/debugging-uno-platform-webassembly-apps-in-visual-studio-2019/ ... but the debugger is still not so nice and you will most probably end have to add some `Console.WriteLine` at some point.

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.cs
Youssef1313: Looks like this suggestion is a copy/paste error?

Should be:

```suggestion
			((IHandleableRoutedEventArgs)args).DoNotPreventDefaultIfHandled = true;
```

dr1rrb: Hey hey yes your right :)

jeromelaban:
@dr1rrb  will those changes be impacted by https://github.com/unoplatform/uno/pull/7088 ?

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.cs
```diff
@@ -666,6 +667,7 @@ protected override void OnPointerReleased(PointerRoutedEventArgs args)
 		protected override void OnKeyDown(KeyRoutedEventArgs args)
 		{
 			base.OnKeyDown(args);
+			((IHandleableRoutedEventArgs)args).DoNotPreventDefaultIfHandled = true;
```
MartinZikmund, line 670: I can't say for sure from memory, but I think this will break the `Tab` and `Shift + Tab` behavior. I think the preventDefault was needed specifically to prevent the browser from handling the same key combination again when cycling focus. Can you verify that when you go to `Focus_FocusCycle` sample page, click the "Set focus on B1" button and then use `Tab` and `Shift+Tab`, the focus does not skip over two buttons at once?

Youssef1313: @MartinZikmund Seems to be working:

*The video shows clicking the button → press few tabs → press few SHIFT+TABs → clicking the button → press few SHIFT+TABs*

https://user-images.githubusercontent.com/31348972/133755979-3279261f-3d04-4524-a690-647be43e9435.mp4

dr1rrb:
> 
> 
> @dr1rrb will those changes be impacted by #7088 ?

I don't see how it could conflict with this one.

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.cs
MartinZikmund: Interesting, I was quite sure that `preventDefault` was needed there 👀 . Can you please check also `FocusManager_FocusDirection`, click on the first `TextBox` and try `Tab` and `Shift+Tab`? If that works too, then I'm ok 😀

Youssef1313: @MartinZikmund Will check. It would be nice if we could write a UI test for these to prevent regressions anyway :)

src/Uno.UI/UI/Xaml/UIElement.wasm.cs
Youssef1313: F5 doesn't seem to work for me.

image

jeromelaban: The debugger is not updated yet in the sample app. Try testing your PR in a separate app, with the Uno.Wasm.Bootstrap package updated to 3.0.0.

MartinZikmund:
@Youssef1313 @dr1rrb I made some small adjustments:

- Created a separate interface `IPreventDefaultHandling` so that the property is only available on __WASM__ and only the events args that need it (now only on `KeyRoutedEventArgs`)
- Renamed the property to `DoNotPreventDefault` (as it is only used if `Handled` is `true` anyway)

I have tested the focus scenarios and it is safe, so this should be ready to go 🙂

MartinZikmund (approved):

Youssef1313:
Thanks for the cleanup @MartinZikmund!

PR #6946, opened 2021-09-02
feat: implements AppMemoryUsage and AppMemoryUsageLimit for Android

### LINKED ISSUE #5891: Please add MemoryManager
MemoryManager, at least .AppMemoryUsage and .AppMemoryUsageLimit

## Why is this needed:
I'm using it in my app (and I'm porting app to Uno).

## For which Platform:

- [ ] iOS
- [X] Android
- [ ] WebAssembly
- [ ] macOS
- [ ] Skia
  - [ ] WPF
  - [ ] GTK (Linux)
  - [ ] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

Hudney:
GitHub Issue: closes #5891

## PR Type
- Feature

Not implemented, throws an exception. #5891

## What is the new behavior?
Returns the correct values.

## PR Checklist

### DIFF
--- a/src/Uno.UWP/System/MemoryManager.Android.cs
+++ b/src/Uno.UWP/System/MemoryManager.Android.cs
@@ -1,0 +1,32 @@
+﻿using System;
+using Android.OS;
+using Android.App;
+using Uno.UI;
+
+namespace Windows.System
+{
+	partial class MemoryManager
+	{
+		public static ulong AppMemoryUsage
+		{
+			get
+			{
+				Debug.MemoryInfo mi = new Debug.MemoryInfo();
+				Debug.GetMemoryInfo(mi);
+				var totalMemory = mi.NativePrivateDirty + mi.OtherSharedDirty;
+				return (ulong)totalMemory;
+			}
+		}
+
+		public static ulong AppMemoryUsageLimit
+		{
+			get
+			{
+				ActivityManager.MemoryInfo memoryInfo = new ActivityManager.MemoryInfo();
+				ActivityManager.FromContext(ContextHelper.Current)?.GetMemoryInfo(memoryInfo);
+
+				return AppMemoryUsage + (ulong)memoryInfo.AvailMem - (ulong)memoryInfo.Threshold;
+			}
+		}
+	}
+}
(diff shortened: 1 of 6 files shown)


src/Uno.UWP/Uno.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="MSBuild.Sdk.Extras" ToolsVersion="15.0">
+﻿<Project Sdk="MSBuild.Sdk.Extras" ToolsVersion="15.0">
```
jeromelaban, line 1: You can rollback all the modifications made to this file. Visual Studio makes them, but they are not useful.

src/Uno.UWP/System/MemoryManager.Android.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+	partial class MemoryManager
+	{
+		public static ulong AppMemoryUsage
+		{
+			get
+			{
+				Debug.MemoryInfo mi = new Debug.MemoryInfo();
+				Debug.GetMemoryInfo(mi);
+				var totalMemory = mi.NativePrivateDirty + mi.OtherSharedDirty;
```
pkar70, line 16: a) Why not add also dalvikPrivateDirty?
b) OtherSharedDirty is, according to doc, in 1 KiB pages, not in bytes. So, multiply it by 1024
c) data from Debug.MemoryInfo are for process. Does Uno Android app has only one process? Or should we iterate processes per app?

teras: a+b) I think the correct way to do it is using totalPss:  https://developer.android.com/reference/android/os/Debug.MemoryInfo#getTotalPss() and multiply by 1024. To have an exact memory consumption for a Unix process (taking into consideration shared memory) is something there's no direct answer to it.

c) There's a concept of multi-process applications in Android, but you need specific code to support this and involves multiple & "remote" activities. Maybe @jeromelaban is more proficient on this, but I doubt that Uno has even remote interest on supporting any of this. And if I remember correctly after Android M (?) the OS provides you with a single PID when you ask for your application PIDs. So, IMHO I think this method method is the best bet to provide the information of the memory consumption of an Application.

jeromelaban: That's correct, multi-process is not of interest at this point, particularly when matching this particular WinRT API. If an application needs fined grained information, it'll need to go through the native api directly.

jeromelaban:
Merged in https://github.com/unoplatform/uno/pull/7048

PR #6966, opened 2021-09-04
fix: Workaround absolute position issue

Youssef1313:
GitHub Issue (If applicable): WORKAROUND #6964 (not sure if it's good enough). Would love to fix the root issue if someone can provide some help :)

## PR Type

- Bugfix

Manipulation event args position is always absolute (relative to screen)

## What is the new behavior?

It's now relative to the control as UWP, except for when the app isn't at the same position as screen.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Input/Internal/ManipulationEventArgsHelpers.cs
+++ b/src/Uno.UI/UI/Xaml/Input/Internal/ManipulationEventArgsHelpers.cs
@@ -1,0 +1,13 @@
+﻿using Windows.Foundation;
+using Windows.UI.Xaml;
+
+namespace Uno.UI.UI.Xaml.Input.Internal
+{
+	internal static class ManipulationEventArgsHelpers
+	{
+		internal static Point MapPointRelativeTo(UIElement element, Point point)
+			=> element is null
+				? point
+				: element.TransformToVisual(null).Inverse.TransformPoint(point);
+	}
+}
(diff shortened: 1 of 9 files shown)


src/Uno.UI/UI/Xaml/Input/Internal/ManipulationEventArgsHelpers.cs
```diff
@@ -0,0 +1,13 @@
...
+using Windows.UI.Xaml;
+
+namespace Uno.UI.UI.Xaml.Input.Internal
+{
+	internal static class ManipulationEventArgsHelpers
+	{
+		internal static Point MapPointRelativeTo(UIElement element, Point point)
+			=> element is null
+				? point
+				: element.TransformToVisual(null).Inverse.TransformPoint(point);
```
Youssef1313, line 11: Stolen from @Xiaoy312 sample :) Thanks for the workaround!

dr1rrb: I'm really not a fan of this one: `TransformToVisual(null)` is really intense as it required to walk the whole tree. Depending of the platform we can have an optimized way to get relative coordinates from the `PointerRoutedEventArgs` that are not so far from here ...

Youssef1313: Let me close the PR for now. I wasn't able to get it, and the workaround can easily be integrated into apps that need it. So no need for such a workaround in Uno itself.

PR #6970, opened 2021-09-05
feat: Implement CreateDefaultValueCallback

### LINKED ISSUE #4347: Add Windows.UI.Xaml.CreateDefaultValueCallback Support

Support for [Windows.UI.Xaml.CreateDefaultValueCallback](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.createdefaultvaluecallback)

## Why is this needed:

When a property is not value-type, suppling a default value should be done using a CreateDefaultValueCallback() delegate to ensure a new instance is created for every DependencyObject.

This is used to implement SelectedItems properties in custom controls (a partial measure since read-only dependency properties are not supported in WinUI).

## For which Platform:

- [x] iOS
- [x] Android
- [x] WebAssembly
- [x] macOS
- [x] Skia
  - [x] WPF
  - [x] GTK (Linux)
  - [x] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

Source location:
 * [CreateDefaultValueCallback.cs](https://github.com/unoplatform/uno/blob/master/src/Uno.UI/Generated/3.0.0.0/Windows.UI.Xaml/CreateDefaultValueCallback.cs)
 * [PropertyMetadata.cs](https://github.com/unoplatform/uno/blob/master/src/Uno.UI/Generated/3.0.0.0/Windows.UI.Xaml/PropertyMetadata.cs)

Youssef1313:
GitHub Issue (If applicable): closes #4347

## PR Type

- Feature

`CreateDefaultValueCallback` not implemented.

## What is the new behavior?

`CreateDefaultValueCallback` is now implemented.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/PropertyMetadata.cs
+++ b/src/Uno.UI/UI/Xaml/PropertyMetadata.cs
@@ -80,6 +80,11 @@ namespace Windows.UI.Xaml
 			BackingFieldUpdateCallback = backingFieldUpdateCallback;
 		}
 
+		internal PropertyMetadata(CreateDefaultValueCallback createDefaultValueCallback)
+		{
+			CreateDefaultValueCallback = createDefaultValueCallback;
+		}
+
 		public PropertyMetadata(
 			PropertyChangedCallback propertyChangedCallback
 		)
@@ -122,11 +127,20 @@ namespace Windows.UI.Xaml
 
 		public object DefaultValue
 		{
-			get => _defaultValue;
+			get
+			{
+				if (CreateDefaultValueCallback != null)
+				{
+					return CreateDefaultValueCallback.Invoke();
+				}
+
+				return _defaultValue;
+			}
 			internal set
 			{
 				_defaultValue = value;
 				_isDefaultValueSet = true;
+				CreateDefaultValueCallback = null;
 			}
 		}
 
@@ -143,6 +157,14 @@ namespace Windows.UI.Xaml
 		}
 
 		internal BackingFieldUpdateCallback BackingFieldUpdateCallback { get; set; }
+
+		internal CreateDefaultValueCallback CreateDefaultValueCallback { get; private set; }
+
+		public static PropertyMetadata Create(CreateDefaultValueCallback createDefaultValueCallback)
+			=> new PropertyMetadata(createDefaultValueCallback: createDefaultValueCallback);
+
+		public static PropertyMetadata Create(CreateDefaultValueCallback createDefaultValueCallback, PropertyChangedCallback propertyChangedCallback)
+			=> new PropertyMetadata(createDefaultValueCallback: createDefaultValueCallback) { PropertyChangedCallback = propertyChangedCallback };
 
 		internal protected virtual void Merge(PropertyMetadata baseMetadata, DependencyProperty dp)
 		{
--- a/src/Uno.UI.Tests/DependencyProperty/Given_DependencyProperty.cs
+++ b/src/Uno.UI.Tests/DependencyProperty/Given_DependencyProperty.cs
@@ -1753,6 +1753,17 @@ namespace Uno.UI.Tests.BinderTests
 			Assert.AreEqual(42, SUT.GetValue(MyDependencyObjectWithDefaultValueOverride.MyPropertyProperty));
 		}
 
+		[TestMethod]
+		public void When_CreateDefaultValueCallback()
+		{
+			var SUT = new MockDependencyObject();
+			var testProperty = DependencyProperty.Register(nameof(When_CreateDefaultValueCallback), typeof(string), typeof(MockDependencyObject), PropertyMetadata.Create(CreateTestString));
+
+			Assert.AreEqual("TestString!", SUT.GetValue(testProperty));
+
+			string CreateTestString() => "TestString!";
+		}
+
 		private class MyDependencyObject : FrameworkElement
 		{
 			internal static readonly DependencyProperty PropAProperty = DependencyProperty.Register(
--- a/src/Uno.UI/Generated/3.0.0.0/Windows.UI.Xaml/CreateDefaultValueCallback.cs
+++ b/src/Uno.UI/Generated/3.0.0.0/Windows.UI.Xaml/CreateDefaultValueCallback.cs
@@ -2,7 +2,7 @@
 #pragma warning disable 114 // new keyword hiding
 namespace Windows.UI.Xaml
 {
-	#if __ANDROID__ || __IOS__ || NET461 || __WASM__ || __SKIA__ || __NETSTD_REFERENCE__ || __MACOS__
+	#if false || false || false || false || false || false || false
 	public delegate object CreateDefaultValueCallback();
 	#endif
 }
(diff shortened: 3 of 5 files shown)


robloo:
This matches with what I was originally thinking. However, it was pointed out this isn't so simple. According to the docs the default value callback is only called once per (UI) thread -- so this is going to have to keep track of threads as well. It's usage was designed for multi-threaded scenarios where everything isn't happening on a single UI thread (rare but happens especially with multiple windows).

Honestly though, this might be good enough.

Youssef1313:
Thanks for review @robloo 
I'm open to exactly match UWP, but will need initial help as to how to implement that the correct way.

src/Uno.UI/UI/Xaml/PropertyMetadata.cs
```diff
@@ -122,11 +127,20 @@ BackingFieldUpdateCallback backingFieldUpdateCallback
 
 		public object DefaultValue
 		{
-			get => _defaultValue;
+			get
+			{
+				if (CreateDefaultValueCallback != null)
+				{
+					return CreateDefaultValueCallback.Invoke();
+				}
```
jeromelaban, line 135: Make sure to [validate the behavior](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.createdefaultvaluecallback?view=winrt-20348#remarks) of `CreateDefaultValueCallback`. We would definitely want to be behave like you implemented it, but that's unfortunately not the case. 

If I remember correctly, it's only called once per UI thread.

/cc @dr1rrb

Youssef1313: @dr1rrb Would you be able to provide help on how to do that?

jeromelaban: @Youssef1313 Can you validate the behavior first on UWP to see if the invocations are done as per what the documentation explain.

dr1rrb: @Youssef1313 the entry point of you work will require usage of the [`[ThreadStatic]` attribute](https://docs.microsoft.com/en-us/dotnet/api/system.threadstaticattribute?view=net-5.0) or it's more recent alternative a [`ThreadLocal`](https://docs.microsoft.com/en-us/dotnet/api/system.threading.threadlocal-1?view=net-5.0)

Youssef1313: I started looking at this again, and I think I need to more understand the behavior on UWP before jumping into how to implement.

---

> If I remember correctly, it's only called once per UI thread.

I tried the following in UWP:

```csharp
    public class MyCustomButton : Button
    {

        public int MyCustomProperty
        {
            get { return (int)GetValue(MyCustomPropertyProperty); }
            set { SetValue(MyCustomPropertyProperty, value); }
        }

        // Using a DependencyProperty as the backing store for MyCustomProperty.  This enables animation, styling, binding, etc...
        public static readonly DependencyProperty MyCustomPropertyProperty =
            DependencyProperty.Register("MyCustomProperty", typeof(int), typeof(MyCustomButton), PropertyMetadata.Create(createDefaultValueCallback: GetValue));

        private static object GetValue()
        {
            return 50;
        }
    }
```

And in a single view app, I added a normal button to MainPage.xaml, and on click, I do:

```csharp
            var myCustomBtn = new MyCustomButton();
            _ = myCustomBtn.MyCustomProperty;

```

It seems that every click executes `GetValue`. ie, it's not once per thread.

---

I'll try to dig more and see if I can fully understand how this works and how to implement it.

Youssef1313:
> According to the docs the default value callback is only called once per (UI) thread

@robloo I couldn't find this part in the docs nor I could reproduce this behavior in a UWP app. Is there anything I'm missing?

What I understand from the docs is that the callback is invoked from whatever thread is accessing the property, regardless which thread has created the property.

src/Uno.UI/UI/Xaml/PropertyMetadata.cs
Youssef1313: Oh I was creating new button each click, so it's very expected for it to execute every time 😄

robloo:
> > According to the docs the default value callback is only called once per (UI) thread
> 
> @robloo I couldn't find this part in the docs nor I could reproduce this behavior in a UWP app. Is there anything I'm missing?
> 
> What I understand from the docs is that the callback is invoked from whatever thread is accessing the property, regardless which thread has created the property.

I was referred to the comment here: https://github.com/unoplatform/uno/issues/4347#issuecomment-715584107

I wouldn't say I understand how this works. My initial understanding was the same as yours and I was then corrected. Testing in UWP as you are doing is likely the best way to determine the correct implementation.

Youssef1313:
I couldn't find the "interesting" test case where I can see a behavior not matching the current implementation. Probably @dr1rrb can help?

jeromelaban:
@Youssef1313 at this point, the usefulness of this feature (even in UWP) seems very niche. In UWP, it seems that the handler is invoked one every get of a DP until it does not have its default value changed through a set operation.

Let's put this PR aside for now, we'll come back to it if its behavior evolves in WinUI.

dr1rrb:
BTW I've validated the behavior using a blank UWP app ... it seems actually completely broken: It re-invokes the `CreateDefaultValueCallback` each time you access the value for an object, until you explicitly set the DP.

```cs
{
	public MainPage()
	{
		this.InitializeComponent();

		Loaded += (snd, args) =>
		{
			var o = new MyDO();

			Debug.WriteLine($"{o.Value} | {o.Value} ({o.Value == o.Value})");
		};
	}
}

public class MyDO : DependencyObject
{
	public static readonly DependencyProperty ValueProperty = DependencyProperty.Register(
		"Value", typeof(object), typeof(MyDO), PropertyMetadata.Create(CreateValue));

	private static object CreateValue()
		=> Guid.NewGuid();

	public object Value
	{
		get => (object)GetValue(ValueProperty);
		set => SetValue(ValueProperty, value);
	}

}
```

will invoke the `CreateValue` **4 times** and output:
`3d1fb4de-d7c5-4bb9-83aa-712bb5bd9c79 | a8190581-142a-4498-8fd9-9ddbacd11fcc (False)`

The `CreateValue` will actually be invoked each time we get the `o.Value` until we "set" is explicitly.

This does not follow the documentation at all, differs to what I've observed in previous versions of windows ... and worst: which is definitively not what anybody would expect! I agree with Jérôme that we should not even try to implement it for now.

@robloo the main purpose of this callback is when you want to provide as default value (i.e. **static**) of a DP an object that thread dependent. When you create a second windows with another UI thread, you must ensure to create another instance of the default value for that thread. An example could be the [`UIElement.UIContext`](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.uielement.uicontext?view=winrt-20348). I know that it is actually not publicly backed by a DP. but it could be internally ... and its only an example as, as Jérôme said, it's really a niche case, I'm not even able to find an explicit example (even if I know they exists).

Youssef1313:
Let me close this for now. Can get to it later when more info is available.

Issue #6763, opened 2021-08-13
Gallery[iOS] Navigationview for Features is not displaying properly while scroll up and down.


sakshi173:
## Current behavior

https://user-images.githubusercontent.com/74563602/129414671-b900ba7a-870e-4c28-b91e-d4aaf246db1f.MOV

## Expected behavior
The Navigation view work properly.

## How to reproduce it (as minimally and precisely as possible)

1. Open the link https://unogallerycanary.z27.web.core.windows.net/
2. Click the Burgrer menu and go to "Feature" section
3. Click on the Up arrow sign on "feature" to open the navigationview.
4. Scroll up and down on the Feature's Navigationview.
5. Observe that the Navigation view list is not displaying properly.

Nuget Package Version(s):

Affected platform(s):

- [X] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] WebAssembly renderers for Xamarin.Forms
- [ ] macOS
- [ ] Skia
  - [ ] WPF
  - [ ] GTK (Linux)
  - [ ] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

IDE:

- [ ] Visual Studio 2017 (version: )
- [ ] Visual Studio 2019 (version: )
- [ ] Visual Studio for Mac (version: )
- [ ] Rider Windows (version: )
- [ ] Rider macOS (version: )
- [ ] Visual Studio Code (version: )

PR #7004, opened 2021-09-09
Fix effective viewport

dr1rrb:
fixes https://github.com/unoplatform/uno/issues/6763

## Bugfix
The `EffectiveViewport` was not computed the same way as UWP

We are considering clipping on all elements while UWP consider clipping only of _scroll host_

## What is the new behavior?
Almost the same behavior as UWP

## PR Checklist

## Other information
**Tests pending**

### DIFF
--- a/src/Uno.UI/UI/Xaml/FrameworkElement.EffectiveViewport.ViewportInfo.cs
+++ b/src/Uno.UI/UI/Xaml/FrameworkElement.EffectiveViewport.ViewportInfo.cs
@@ -1,0 +1,103 @@
+﻿#nullable enable
+
+using System;
+using System.Linq;
+using Windows.Foundation;
+using Uno.Extensions;
+
+namespace Windows.UI.Xaml
+{
+	// Internal interface used to allow communication between the real FrameworkElement
+	// and presenters that are only implementing the IFrameworkElement interface (cf. FrameworkElementMixins.tt).
+	// It must not be used anywhere else out of the file.
+	internal interface IFrameworkElement_EffectiveViewport
+	{
+		void InitializeEffectiveViewport();
+		IDisposable RequestViewportUpdates(IFrameworkElement_EffectiveViewport? child = null);
+		void OnParentViewportChanged(IFrameworkElement_EffectiveViewport parent, ViewportInfo viewport, bool isInitial = false);
+		void OnLayoutUpdated();
+	}
+
+	internal struct ViewportInfo : IEquatable<ViewportInfo>
+	{
+		public static ViewportInfo Empty { get; } = new ViewportInfo { Effective = Rect.Empty, Clip = Rect.Empty };
+
+		/// <summary>
+		/// The element used as reference coordinate space to express the values.
+		/// </summary>
+		public IFrameworkElement_EffectiveViewport? Reference;
+
+		/// <summary>
+		/// The public effective viewport
+		/// </summary>
+		public Rect Effective;
+
+		/// <summary>
+		/// The cumulative clipping applied by all parents scroll ports.
+		/// </summary>
+		public Rect Clip;
+
+		public ViewportInfo(IFrameworkElement_EffectiveViewport reference, Rect viewport)
+		{
+			Reference = reference;
+			Effective = Clip = viewport;
+		}
+
+		public ViewportInfo(IFrameworkElement_EffectiveViewport reference, Rect effective, Rect clip)
+		{
+			Reference = reference;
+			Effective = effective;
+			Clip = clip;
+		}
+
+		public ViewportInfo GetRelativeTo(IFrameworkElement_EffectiveViewport element)
+		{
+			Rect effective = Effective, clip = Clip;
+			if (element is UIElement uiElement && Reference is UIElement usuallyTheParentOfElement)
+			{
+				var parentToElement = UIElement.GetTransform(uiElement, usuallyTheParentOfElement).Inverse();
+
+				if (Effective.IsValid)
+				{
+					effective = parentToElement.Transform(Effective);
+				}
+
+				if (Clip.IsValid)
+				{
+					clip = parentToElement.Transform(Clip);
+				}
+			}
+
+			// If the Reference or the element is not a UIElement (native), we don't have any RenderTransform,
+			// and we assume that the 'element' is stretched in it's parent usually true except for ListView,
+			// but we won't even try to support that case.
+
+			return new ViewportInfo(element, effective, clip);
+		}
+
+		/// <inheritdoc />
+		public override string ToString()
+			=> $"{Effective.ToDebugString()} (clip: {Clip.ToDebugString()})";
+
+		/// <inheritdoc />
+		public override int GetHashCode()
+			=> Effective.GetHashCode();
+
+		/// <inheritdoc />
+		public bool Equals(ViewportInfo other)
+			=> Equals(this, other);
+
+		/// <inheritdoc />
+		public override bool Equals(object obj)
+			=> obj is ViewportInfo vp && Equals(this, vp);
+
+		private static bool Equals(ViewportInfo left, ViewportInfo right)
+			=> left.Effective.Equals(right.Effective);
+
+		public static bool operator ==(ViewportInfo left, ViewportInfo right)
+			=> Equals(left, right);
+
+		public static bool operator !=(ViewportInfo left, ViewportInfo right)
+			=> !Equals(left, right);
+	}
+}
--- a/src/Uno.UI/UI/Xaml/FrameworkElement.EffectiveViewport.cs
+++ b/src/Uno.UI/UI/Xaml/FrameworkElement.EffectiveViewport.cs
@@ -1,15 +1,17 @@
-﻿//#define TRACE_EFFECTIVE_VIEWPORT
+﻿#nullable enable
+
+#define TRACE_EFFECTIVE_VIEWPORT
 
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Numerics;
 using System.Runtime.CompilerServices;
 using Windows.Foundation;
 using Windows.UI.Xaml.Controls.Primitives;
 using Uno;
 using Uno.Disposables;
-using Uno.Extensions;
 using Uno.UI;
 using Uno.UI.Extensions;
 using _This = Windows.UI.Xaml.FrameworkElement;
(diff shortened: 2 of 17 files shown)


src/Uno.UI/UI/Xaml/FrameworkElement.EffectiveViewport.cs
```diff
@@ -1,15 +1,17 @@
-﻿//#define TRACE_EFFECTIVE_VIEWPORT
+﻿#nullable enable
+
+#define TRACE_EFFECTIVE_VIEWPORT
```
dr1rrb, line 3: Going to remove that!

src/Uno.UI/UI/Xaml/FrameworkElement.EffectiveViewport.cs
```diff
@@ -89,23 +76,25 @@ private void ReconfigureViewportPropagation(IFrameworkElement_EffectiveViewport
...
-				else
+				else if (child != null)
 				{
 					TRACE_EFFECTIVE_VIEWPORT("New child requested viewport propagation which has already been enabled, forwarding current viewport to it.");
 
 					// We are already subscribed, the parent won't send any update (and our _parentViewport is expected to be up-to-date).
 					// But if this "reconfigure" was made for a new child (child != null), we have to initialize its own _parentViewport.
 
-					var slot = LayoutInformation.GetLayoutSlot(this); // a.k.a. the implicit viewport to use if none was defined by any parent (usually only few elements at the top of the tree)
+					//var slot = LayoutInformation.GetLayoutSlot(this); // a.k.a. the implicit viewport to use if none was defined by any parent (usually only few elements at the top of the tree)
```
dr1rrb, line 93: About to remove that also

src/Uno.UI/UI/Xaml/FrameworkElement.EffectiveViewport.cs
```diff
@@ -349,11 +318,11 @@ protected void InvalidateViewport()
 		}
 #endif
 
-		[Conditional("TRACE_EFFECTIVE_VIEWPORT")]
+		//[Conditional("TRACE_EFFECTIVE_VIEWPORT")]
```
dr1rrb, line 321: Will uncoment this also

src/Uno.UI/UI/Xaml/UIElement.cs
```diff
@@ -343,7 +343,7 @@ internal static Matrix3x2 GetTransform(UIElement from, UIElement to)
 					offsetY = layoutSlot.Y;
 				}
 
-#if !UNO_HAS_MANAGED_SCROLL_PRESENTER
+#if !UNO_HAS_MANAGED_SCROLL_PRESENTER && !__WASM__
```
carldebilly, line 346: Looks like there's a missing #define for this special case...

carldebilly (approved):

src/Uno.UI/UI/Xaml/FrameworkElement.EffectiveViewport.ViewportInfo.cs
```diff
@@ -0,0 +1,103 @@
...
+using System.Linq;
+using Windows.Foundation;
+using Uno.Extensions;
+
+namespace Windows.UI.Xaml
+{
+	// Internal interface used to allow communication between the real FrameworkElement
+	// and presenters that are only implementing the IFrameworkElement interface (cf. FrameworkElementMixins.tt).
+	// It must not be used anywhere else out of the file.
+	internal interface IFrameworkElement_EffectiveViewport
```
jeromelaban, line 13: Move to another file

src/Uno.UI/UI/Xaml/FrameworkElement.EffectiveViewport.ViewportInfo.cs
```diff
@@ -0,0 +1,103 @@
...
+	{
+		void InitializeEffectiveViewport();
+		IDisposable RequestViewportUpdates(IFrameworkElement_EffectiveViewport? child = null);
+		void OnParentViewportChanged(IFrameworkElement_EffectiveViewport parent, ViewportInfo viewport, bool isInitial = false);
+		void OnLayoutUpdated();
+	}
+
+	internal struct ViewportInfo : IEquatable<ViewportInfo>
```
jeromelaban, line 21: Why a struct ?

jeromelaban (approved):

src/Uno.UI/UI/Xaml/FrameworkElement.EffectiveViewport.ViewportInfo.cs
dr1rrb: Using a `class` for this would most probably be overkill, we could have used a `record`, but this is designed to be a "super `Rect`" that is used as a drop-in replacement in some cases.

src/Uno.UI/UI/Xaml/UIElement.cs
dr1rrb: Right ... but finally I just removed it!

src/Uno.UI/UI/Xaml/Controls/ListViewBase/ListViewBaseScrollContentPresenter.Android.cs
```diff
@@ -10,10 +10,15 @@
...
 namespace Windows.UI.Xaml.Controls
 {
-	public sealed partial class ListViewBaseScrollContentPresenter : ContentPresenter, IScrollContentPresenter
+	public sealed partial class ListViewBaseScrollContentPresenter : IScrollContentPresenter, INativeScrollContentPresenter
 	{
-
-		private Thickness _oldPadding;
+		// NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE NOTE
+		//
+		// This class is about to be replaced by an implementation of INativeScrollContentPresenter which only adapts the 'NativePanel'.
```
davidjohnoliver, line 17: I'm curious what you have in mind here? It would be great to be able to eliminate the need for a custom ScrollViewer template for ListView and GridView, and instead enable the correct behaviour transparently.

src/Uno.UI/UI/Xaml/Window.Android.cs
```diff
@@ -76,6 +78,34 @@ private static Window InternalGetCurrentWindow()
 
 		internal void RaiseNativeSizeChanged()
 		{
+#if __ANDROID_30__
```
davidjohnoliver, line 81: Will these APIs work on devices running Android 29 and below?

src/Uno.UI/UI/Xaml/Window.Android.cs
```diff
@@ -76,6 +78,34 @@ private static Window InternalGetCurrentWindow()
...
+
+			// The 'metric.Bounds' does not include any insets, so we remove the "solid" insets under which we cannot draw anything
+			var windowBounds = new Rect(default, ((Rect)metrics.Bounds).DeflateBy(solidInsets).Size).PhysicalToLogicalPixels();
+			var newBounds = windowBounds; 
+
+			// The visible bounds is the windows bounds on which we remove also
+			// [translucent only inset](all_insets - solid_insets_that_has_already_been_removed_from_all_insets)
+			var visibleBounds = windowBounds.DeflateBy(insets.Minus(solidInsets)).PhysicalToLogicalPixels();
+
+			// The true visible bounds ... is unknown for now!
```
davidjohnoliver, line 106: ?

src/Uno.UI/UI/Xaml/Controls/ListViewBase/ListViewBaseScrollContentPresenter.Android.cs
dr1rrb: Well it's almost already the case as I made the `ListViewSCP` inherit from `SCP` and stripped of  lot of logic which was shared with the `NativeSCP` (moved directly in the SCP).

The ides is that the `SCP` on iOS and Android should only act as a decorator/proxy of my new `INativeSCP`. Then we should have 2 implementations of the `INativeSCP`: one for the default `ScrollViewer` (which _could_ inherit from the `UIScrollView`) and one for the `ListView` (which would adapt the `NativePanel` managed by the `ListView`). There is then no difficulty to instance one of those dynamically depending of the templated parent when we create the visual tree. Actually we are already doing that for the default `ScrollViewer`, we can do the same for the `ListView`.

The main difficulty is to re-dispatch responsibilities. We are dealing with old code here, and I want to make sure that the `INativeSCP` is as minimal as possible.

src/Uno.UI/UI/Xaml/Window.Android.cs
dr1rrb: They are documented to be available starting with API 30, and apparently we didn't had that issue in previous versions so fallbacking on the old code seems legit.

dr1rrb: Yep, that's the reason why the tests was failing on the CI for Android 🤦‍♂️

src/Uno.UI/UI/Xaml/Window.Android.cs
```diff
@@ -76,6 +80,57 @@ private static Window InternalGetCurrentWindow()
...
+			var solidInsets = metrics.WindowInsets.GetInsets(solidInsetsTypes).ToThickness();
+
+			// The 'metric.Bounds' does not include any insets, so we remove the "solid" insets under which we cannot draw anything
+			var windowBounds = new Rect(default, ((Rect)metrics.Bounds).DeflateBy(solidInsets).Size).PhysicalToLogicalPixels();
+
+			// The visible bounds is the windows bounds on which we remove also
+			// [translucent only inset](all_insets - solid_insets_that_has_already_been_removed_from_all_insets)
+			var visibleBounds = windowBounds.DeflateBy(insets.Minus(solidInsets)).PhysicalToLogicalPixels();
+
+			return (windowBounds, visibleBounds, visibleBounds);
```
davidjohnoliver, line 128: I'd like to understand this better, the original distinction between VisibleBounds (public) and TrueVisibleBounds (internal) was introduced to avoid a breaking change in the semantics of VisibleBounds.

What is changing here w.r.t. the old code path - the value of VisibleBounds, or the value of TrueVisibleBounds?

dr1rrb: Discuss, as long as tests are passing, we agreed that it's correct

PR #7074, opened 2021-09-14
fix: Fix InputScope issues on Android and Wasm

### LINKED ISSUE #1389: [WASM] Support TextBox.InputScope to allow input restrictions
An input control that only accepts numbers. For example it can be an input control like TextBox but with the option to only recieve numbers. This is done in Android and iOS.

## Why is this needed:
There are complex forms that need specific input and user could be confused over the the way they should give input, so it can give like a safe way to users to enter the respective input.

## For which Platform:

- [ ] iOS
- [ ] Android
- [x] WebAssembly
- [ ] Windows

Telerik have a control that can do this, but sadly this control is not currently working on WASM.

### LINKED ISSUE #2557: [Android]TextBox.InputScope=Number causes cursor to reset when entering non-numeric character
#  Current behavior
Given this textbox, `<TextBox InputScope="Number" />`, upon entering any non-numeric character the cursor position resets to back 0.

## Expected behavior
Same as uwp, the cursor position should not reset.

```xaml
<TextBox InputScope="Number" />
```

## Environment
Nuget Package: Uno.UI
Package Version(s):  2.1.0-dev.943
Affected platform(s):
- [ ] iOS
- [x] Android
- [ ] WebAssembly
- [ ] macOS
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

Visual Studio:
- [ ] 2017 (version: )
- [x] 2019 (version: )
- [ ] for Mac (version: )

Relevant plugins:
- [ ] Resharper (version: )

Workaround: The same issue does not happen with if `InputScope="TelephoneNumber"` were used instead.
Somewhat related: https://github.com/unoplatform/uno/issues/2038
https://nventive.visualstudio.com/Umbrella/_workitems/edit/166205

### LINKED ISSUE #5207: [Android] TextBox with InputScope as CurrencyAmount move the cursor on the left when pressing comma.
## Current behavior
As you can see, the cursor moves to the left after pressing the comma.

ezgif com-video-to-gif

## Expected behavior

It should not move the cursor what so ever after pressing the comma by default. This behavior does not happend on neither xamarin.classic nor native.

Xamarin.Classic [TextBoxCurrency_Xamarin.zip](https://github.com/unoplatform/uno/files/5960428/TextBoxCurrency_Xamarin.zip)
ezgif com-video-to-gif

Native [TextBoxCurrency_Native.zip](https://github.com/unoplatform/uno/files/5960431/TextBoxCurrency_Native.zip)
ezgif com-video-to-gif

Download this sample [TextBoxCurrency_Uno.zip](https://github.com/unoplatform/uno/files/5960440/TextBoxCurrency_Uno.zip)
Deploy on Android
Try to enter "2,2"

## Workaround

I mean I could use a similar InputScope (for instance `NumberFullWidth`) for now. Ideally, I would need a way to import what works on the xamarin sample into my project as a workaround but I couldn't figure it out if it's even possible.

```xml
<EditText
        android:layout_width="wrap_content"
        android:layout_height="wrap_content"
        android:layout_centerInParent="true"
        android:digits="1234567890,." // <-- Have this one too, so I can actually press the comma key.
        android:inputType="numberDecimal" // <-- If I could somehow use this one.
        android:text="" />
```

## Environment

Nuget Package:

- [ ] Other: 

Nuget Package Version(s): 3.6.0-dev.27

Affected platform(s):

- [ ] iOS
- [x] Android
- [ ] WebAssembly
- [ ] macOS
- [ ] Skia
  - [ ] WPF
  - [ ] GTK (Linux)
  - [ ] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

IDE:

- [x] Visual Studio 2019 (version: 16.8.4)
- [ ] Rider macOS (version: )

Relevant plugins:

- [ ] Resharper (version: )

Youssef1313:
GitHub Issue (If applicable): closes #5207, closes #2557, closes #1389

## PR Type

- Bugfix

On Android:

- `InputScope` maps to `inputTypes` on Android, which does validation.

On Wasm:

- `InputScope` doesn't do anything.

## What is the new behavior?

On Android:

- Use `SetRawInputType`, which only changes the keyboard but allows any input.

On Wasm:

- Use `inputmode` attribute which changes the keyboard but allows any input.

## PR Checklist

## Other information

I didn't find a way to test how the keyboard shows.


src/Uno.UI/UI/Xaml/Controls/TextBox/TextBoxView.wasm.cs
```diff
@@ -134,6 +135,40 @@ internal void SetIsReadOnly(bool isReadOnly)
...
 
+		internal void SetInputScope(InputScope scope)
+		{
+			// https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode
+			// Allowed html values: none, text (default value), decimal, numeric, tel, search, email, url.
+			var scopeNameValue = scope.GetFirstInputScopeNameValue();
+			var inputmodeValue = scopeNameValue switch
+			{
+				InputScopeNameValue.CurrencyAmount => "decimal",
+				InputScopeNameValue.CurrencyAmountAndSymbol => "decimal",
```
Youssef1313, line 146: In UWP, these are certainly not the same. But the values allowed in HTML are very few.

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.Android.cs
```diff
@@ -228,7 +228,7 @@ protected void SetInputScope(InputTypes types)
 		{
 			if (_textBoxView != null)
 			{
-				_textBoxView.InputType = types;
+				_textBoxView.SetRawInputType(types);
```
jeromelaban, line 231: This is likely to be a breaking change for application current expecting uno to do the validation. This new behavior is likely to be the right one, though. We should be placing this behind a feature flag.

Youssef1313: @jeromelaban To be more clear about the current behavior, we indeed was mapping to an Android API that does validation, but *somehow*, Uno managed to skip/override the native validation(probably [this](https://github.com/unoplatform/uno/blob/master/src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.Overrides.Android.cs)? not sure), but anyway the result was the cursor moving to the beginning after every invalid character.

So I don't expect anyone to be relying on the current behavior.

jeromelaban: SIP is Soft Input Panel (the on-screen keyboard). The old behavior was that keys outside the input type would not be entered at all if using a physical keyboard (and the cursor would not more).

Youssef1313: @jeromelaban Thanks, got it and addressed the feedback.

jeromelaban (approved):

jeromelaban:
Looks like the failing tests are related to this issue: TextBox_IsReadOnly_AcceptsReturn_Test and TextBox_AutoGrow_Vertically_Test on android. Can you take a look ?

Youssef1313:
@jeromelaban I looked at `TextBox_IsReadOnly_AcceptsReturn_Test`. It seems like the old behavior doesn't match UWP, but the new behavior does? BUT there might be other non-tested broken cases. I'm also completely unsure how this change can affect AcceptsReturn.


In UWP:

```xaml
<TextBox x:Name="TargetTextBox"
	AcceptsReturn="True"
	IsReadOnly="True"
	Text="lorem&#x0a;ipsum" />
```

and

```xaml
<TextBox x:Name="TargetTextBox"
	Text="lorem&#x0a;ipsum"
	AcceptsReturn="True"
	IsReadOnly="True" />
```

are different. The order of setting properties matters here because `Text` seems to be cutting the extra lines from the property value itself (not only visually) when `AcceptsReturn` is false.

The failing test is setting the text first, so the text gets cut to one line only. So the expectation of height differences is NOT correct for this particular test.

The actual case that might be regressing is when `AcceptsReturn` is set first.

PR #7080, opened 2021-09-14
fix: Handle types correctly in ResourceDictionary

### LINKED ISSUE #7044: `ResourceDictionary` shouldn't have a string key for `Style` without explicit `x:Name`
### Current behavior

```xml
    <Grid x:Name="grid">
        <Grid.Resources>
            <Button x:Name="MyButton" />
            <Style TargetType="StackPanel">
                <Setter Property="Margin" Value="4" />
            </Style>
            <Style TargetType="TextBlock" x:Name="MyStyle">
                <Setter Property="VerticalAlignment" Value="Center" />
                <Setter Property="Margin" Value="0,0,4,0"/>
            </Style>
        </Grid.Resources>
    </Grid>
```

```csharp
var resources = grid.Resources.ToList();
var a = resources[0];
var b = resources[1];
var c = resources[2];
```

All resources have a string Key.

### Expected behavior

Only `MyButton` and `MyStyle` have a string key.

UWP behavior:

image

↑

### Workaround

_No response_

### Works on UWP/WinUI

Yes

### Environment

_No response_

### NuGet package version(s)

_No response_

### Affected platforms

_No response_

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

Youssef1313:
GitHub Issue (If applicable): closes #7044

## PR Type

- Bugfix

Adding a resource with a System.Type key then retrieving it will result in a string being produced.

## What is the new behavior?

Correctly retrieve it as System.Type.

There was two issues here:

1. `ConvertKey(k.Key)`, this was resulting in an implicit conversion from string to `ResourceKey`, resulting in losing the information of the original `ResourceKey`.
2. Usage of `Type.GetType`, which only works in the current assembly or mscorlib (or System.Private.CoreLib).

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/SpecializedResourceDictionary.cs
+++ b/src/Uno.UI/UI/Xaml/SpecializedResourceDictionary.cs
@@ -31,6 +31,7 @@ namespace Windows.UI.Xaml
 		public readonly struct ResourceKey
 		{
 			public readonly string Key;
+			public readonly Type TypeKey;
 			public readonly bool IsType;
 			public readonly uint HashCode;
 
@@ -41,6 +42,7 @@ namespace Windows.UI.Xaml
 			private ResourceKey(bool dummy)
 			{
 				Key = null;
+				TypeKey = null;
 				IsType = false;
 				HashCode = 0;
 			}
@@ -54,12 +56,14 @@ namespace Windows.UI.Xaml
 				if (key is string s)
 				{
 					Key = s;
+					TypeKey = null;
 					IsType = false;
 					HashCode = (uint)(s.GetHashCode() ^ IsType.GetHashCode());
 				}
 				else if (key is Type t)
 				{
 					Key = t.ToString();
+					TypeKey = t;
 					IsType = true;
 					HashCode = (uint)(t.GetHashCode() ^ IsType.GetHashCode());
 				}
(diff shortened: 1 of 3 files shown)


src/Uno.UI/UI/Xaml/SpecializedResourceDictionary.cs
```diff
@@ -31,6 +31,7 @@ internal class SpecializedResourceDictionary
 		public readonly struct ResourceKey
 		{
 			public readonly string Key;
+			public readonly Type TypeKey;
 			public readonly bool IsType;
```
jeromelaban, line 35: We can certainly remove the `IsType` as a result now, as it is only used for type keys explicitly. That's remove a field to copy during by-value parameters.

Youssef1313: @jeromelaban Yeah sure. Will remove it. Do you want the HashCode to use `(TypeKey != null).GetHashCode()` (match current semantics), or `TypeKey.GetHashCode()` (probably better)?

jeromelaban: It's likely that `TypeKey?.GetHashCode() ?? 0` will be good enough.

jeromelaban (approved):
Thanks for your contribution!

PR #7047, opened 2021-09-13
Add back missing `PersonPicture` sample to projitems and fix crashes

Youssef1313:
GitHub Issue (If applicable): 

## PR Type

What kind of change does this PR introduce?

- Bugfix
- Sample test

## What is the current behavior?

- `PersonPicture` sample doesn't exist.
- `PersonPicture` can crash.

## What is the new behavior?

- Add the sample in projitems (it was deleted, likely unintentional, in https://github.com/unoplatform/uno/pull/2252), and update the sample from WinUI.
- Fix crashes in the sample due to incorrect default (null) values of string fields.

## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)
- [ ] [Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [ ] Validated PR `Screenshots Compare Test Run` results.
- [ ] Contains **NO** breaking changes
- [ ] Associated with an issue (GitHub or internal) and uses the [automatic close keywords](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue).
- [ ] Commits must be following the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) specification.

## Other information

Internal Issue (If applicable):

PR #7087, opened 2021-09-15
fix: Use the correct default value for string DPs

### LINKED ISSUE #7085: Some string DPs have incorrect default value (null instead of string.Empty)
### Current behavior

Some string DPs have incorrect default value (null instead of string.Empty). Discovered that in a few places in https://github.com/unoplatform/uno/pull/7047 where the PersonPicture sample crashed with a NRE, and suspecting there are more DPs with the same issue.

### Expected behavior

Default value is `string.Empty` instead of `null`. (possibly for ~~*all*~~ (there are very few exceptions) string DPs?)

_No response_

### Workaround

_No response_

### Works on UWP/WinUI

_No response_

### Environment

_No response_

### NuGet package version(s)

_No response_

### Affected platforms

_No response_

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

Youssef1313:
GitHub Issue (If applicable): closes #7085

## PR Type

## PR Checklist

## Other information


src/Uno.UI/UI/Xaml/Media/Animation/FadeInThemeAnimation.cs
```diff
@@ -26,7 +26,9 @@ public FadeInThemeAnimation()
 		}
 
 		public static DependencyProperty TargetNameProperty { get; } = DependencyProperty.Register(
-			"TargetName", typeof(string), typeof(FadeInThemeAnimation), new FrameworkPropertyMetadata(null));
+#pragma warning disable Uno0002_Internal // String dependency properties (in *most* cases) shouldn't have null default value.
+			"TargetName", typeof(string), typeof(FadeInThemeAnimation), new FrameworkPropertyMetadata(null)); // TODO: What the default value should be? null or string.Empty?
```
jeromelaban, line 30: It looks like on Windows the value is string.Empty. What makes you think otherwise? (same for the cases other files)

Youssef1313: I tried to validate it now and indeed it's string.Empty. The property that I wasn't able to verify on UWP is `Storyboard.TargetNameProperty`

image

But I see now that I should have used `new ObjectAnimationUsingKeyFrames().GetValue(Storyboard.TargetNameProperty)` instead to validate. I'll fix them.

Youssef1313:
@jeromelaban There is likely one more issue related to this area. Code imported from WinUI that contains string fields was usually imported and defaulting to null, which can cause NREs as I saw with PersonPicture in https://github.com/unoplatform/uno/pull/7047/files.

This one is harder to write an analyzer for, it can have lots of false positives. I'll probably do a quick manual review in a new PR, but as we port new controls, this should be reviewed carefully. Almost every WinUI string is never null as they use winrt::hstring (struct)

src/Uno.UI/UI/Xaml/Controls/Button/RadioButton.cs
```diff
@@ -96,7 +96,7 @@ public string GroupName
 				typeof(string), 
 				typeof(RadioButton), 
 				new FrameworkPropertyMetadata(
-					(string)null, 
+					string.Empty,
```
MartinZikmund, line 99: This might a bit trickier. As I have worked on `RadioButton`, I noticed something strange:

```
var radioButton = new RadioButton();
Debug.WriteLine(radioButton.GroupName == ""); // True
Debug.WriteLine(radioButton.GetValue(RadioButton.GroupNameProperty) == null); // True
```

So while the property itself returns empty string, the dependency property has a default value of `null`. I am not sure whether this is just a `RadioButton` oddity, or it may be the case of some other string-based dependency properties. In addition, trying to set `GroupName = null` ends up with an exception.

In my `RadioButton` branch I left the `null` as default value and used `??` in the property getter to return empty string instead of `null`, and throwing in the setter for `null` value.

src/Uno.UI/UI/Xaml/Controls/AppBar/AppBarButton.cs
```diff
@@ -37,7 +37,7 @@ public string Label
 			DependencyProperty.Register(
 				"Label", typeof(string),
 				typeof(AppBarButton),
-				new FrameworkPropertyMetadata(default(string))
+				new FrameworkPropertyMetadata(string.Empty)
```
Youssef1313, line 40: TODO: null is correct. Setter should do `?? string.Empty`

src/Uno.UI/UI/Xaml/Controls/AppBar/AppBarToggleButton.cs
```diff
@@ -40,7 +40,7 @@ public string Label
 			DependencyProperty.Register(
 				"Label", typeof(string),
 				typeof(AppBarToggleButton),
-				new FrameworkPropertyMetadata(default(string))
+				new FrameworkPropertyMetadata(string.Empty)
```
Youssef1313, line 43: TODO: null is correct. Setter should do `?? string.Empty`

src/Uno.UI/UI/Xaml/Controls/AutoSuggestBox/AutoSuggestBox.Properties.cs
```diff
@@ -91,7 +91,7 @@ public string Text
 		Windows.UI.Xaml.DependencyProperty.Register(
 			"TextMemberPath", typeof(string),
 			typeof(global::Windows.UI.Xaml.Controls.AutoSuggestBox),
-			new FrameworkPropertyMetadata(defaultValue: null)
+			new FrameworkPropertyMetadata(defaultValue: string.Empty)
```
Youssef1313, line 94: Hmm.. It looks like most controls in MUXC will have null default value but the setter does `?? ""`.
I'll need to adjust the analyzer for that and fix the PR.

Thanks @MartinZikmund for catching that! 🎉

src/Uno.UI/UI/Xaml/Controls/CalendarView/Primitives/CalendarViewTemplateSettings.cs
```diff
@@ -48,7 +48,7 @@ public string WeekDay1
 			"WeekDay1",
 			typeof(string),
 			typeof(CalendarViewTemplateSettings),
-			new FrameworkPropertyMetadata(default(string)));
+			new FrameworkPropertyMetadata(string.Empty));
```
Youssef1313, line 51: It seems like all these should be regular properties, not DPs. I'm going to leave this as string.Empty instead of adding `?? ""` since the DP doesn't actually exist in UWP.

src/Uno.UI/UI/Xaml/Controls/NavigationView/NavigationView.Properties.cs
```diff
@@ -530,7 +530,7 @@ public NavigationViewTemplateSettings TemplateSettings
 			nameof(PaneTitle),
 			typeof(string),
 			typeof(NavigationView),
-			new FrameworkPropertyMetadata(default(string),
```
Youssef1313, line 533: In UWP this is not a DP. I'll keep it as string.Empty since this is the default of the regular property.

MartinZikmund:
Hey @Youssef1313, would you be able to rebase this PR on latest master please?

Youssef1313:
@jeromelaban @MartinZikmund Back to this again now, I'm not sure if adding an analyzer is a good idea or not. Let me know if you prefer having the analyzer, or just extract out the real fixes from this PR?

jeromelaban:
> @jeromelaban @MartinZikmund Back to this again now, I'm not sure if adding an analyzer is a good idea or not. Let me know if you prefer having the analyzer, or just extract out the real fixes from this PR?

The change is important, though there are some changes that may be breaking behaviors, so we'll likely have to split this PR for those that are not breaking.

MartinZikmund:
@jeromelaban @Youssef1313 I think this can now be rebased and get it ready, for 5.0?

PR #7090, opened 2021-09-15
feat(NumeralSystemTranslator): Implement TranslateNumerals Method

MohammadHadi2031:
Implement TranslateNumerals method for Arab and ArabExt NumeralSystem value

## PR Type

What kind of change does this PR introduce?\

- Feature

Nothing.

## What is the new behavior?

Translates digits of the Latin numerical system into digits of Arab and ArabExt numerical systems.

## PR Checklist

## Other information


MohammadHadi2031:
@MartinZikmund Next step is to add test for other languages. Would you mind checking my code style and naming?

src/Uno.UWP/Globalization/NumberFormatting/NumeralSystemTranslator.cs
```diff
@@ -0,0 +1,167 @@
...
+
+			foreach (var language in languages)
+			{
+				NumeralSystemTranslatorHelper.GetNumeralSystem(language);
+			}
+		}
+
+		public string TranslateNumerals(string value)
+		{
+			if (NumeralSystem == "Arab" ||
```
MartinZikmund, line 63: Please perform the comparison using `.Equals(..., CultureInfo.InvariantCultureIgnoreCase)`

src/Uno.UWP/Globalization/NumberFormatting/NumeralSystemTranslator.cs
```diff
@@ -0,0 +1,167 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Windows.Globalization.NumberFormatting
+{
+	public partial class NumeralSystemTranslator
+	{
+		private static Dictionary<string, char[]> s_numeralSystemToDigitsDictionary =
```
MartinZikmund, line 9: Please use `_underscoreNamingConvention` for private members (and maybe this one can be made `readonly` too?)

src/Uno.UWP/Globalization/NumberFormatting/NumeralSystemTranslator.cs
```diff
@@ -0,0 +1,167 @@
...
+			NumeralSystemTranslatorHelper.CreateNumeralSystemToDigitsDictionary();
+		private string numeralSystem;
```
MartinZikmund, line 11: Also here use `_underscoreField` naming convention :-)

src/Uno.UWP/Globalization/NumberFormatting/NumeralSystemTranslator.cs
MartinZikmund: We use the naming like `s_something` when porting from WinUI, where we tend to retain the C++ conventions used there, but here we just use plain underscore

src/Uno.UWP/Globalization/NumberFormatting/NumeralSystemTranslator.cs
```diff
@@ -0,0 +1,167 @@
...
+					throw new ArgumentException("The parameter is incorrect.");
+				}
+
+				numeralSystem = value;
+			}
+		}
+
+		public IReadOnlyList<string> Languages { get; }
+
+		public string ResolvedLanguage { get; }
```
MartinZikmund, line 29: Please make the ordering in the class as follows:

- Fields
- Constructors
- Events
- Properties
- Methods

MartinZikmund:
@MohammadHadi2031 Added a few comments, looking good so far!

MohammadHadi2031:
@MartinZikmund Thanks you so much. I fixed naming conventions and coding style issues. Please tell me if I miss something or there is new one. 
I Added tests for all numeral systems. Also for initialization with various languages. I wonder if numeral system tests aren't good enough. Please guide me how to add good tests.

src/Uno.UWP/Globalization/NumberFormatting/NumeralSystemTranslatorHelper.cs
```diff
@@ -0,0 +1,1781 @@
...
+
+			Debug.Assert(result.Count == Count);
+			return result;
+		}
+
+		public static string GetNumeralSystem(string languageTag)
+		{
+			languageTag = languageTag.ToLowerInvariant();
+
+			if (_languageTagToNumeralSystemDictionary.TryGetValue(languageTag, out string value))
```
jeromelaban, line 1761: With regards to performance, it's best to use an actual switch/case block. The C# compiler uses a static hash-table that's more efficient at runtime and particularly initialization time.

src/Uno.UWP/Globalization/NumberFormatting/NumeralSystemTranslatorHelper.cs
```diff
@@ -0,0 +1,1781 @@
...
+			}
+
+			throw new ArgumentException("The parameter is incorrect.");
+		}
+
+		public static string GetResolvedLanguage(string languageTag)
+		{
+			languageTag = languageTag.ToLowerInvariant();
+
+			if (_languageTagToResolvedLanguageDictionary.TryGetValue(languageTag, out string value))
```
jeromelaban, line 1773: Same here.

src/Uno.UWP/Globalization/NumberFormatting/NumeralSystemTranslator.cs
```diff
@@ -0,0 +1,185 @@
...
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using Uno.Globalization.NumberFormatting;
+
+namespace Windows.Globalization.NumberFormatting
+{
+	public partial class NumeralSystemTranslator
+	{
+		private readonly static Dictionary<string, char[]> _numeralSystemToDigitsDictionary =
```
MohammadHadi2031, line 11: @jeromelaban Should I use switch/case instead of dictionary here?

MohammadHadi2031:
@jeromelaban I added a review don't miss it :)

src/Uno.UWP/Globalization/NumberFormatting/NumeralSystemTranslator.cs
jeromelaban: Yes, let's do that as well. Thanks!

src/Uno.UI.Tests/Windows_Globalization/Given_NumeralSystemTranslator.cs
```diff
@@ -0,0 +1,1570 @@
...
+		}
+
+		[DataTestMethod]
+		[DynamicData(nameof(When_UsingLanguage_Then_InitializeCorrectly_TestData), DynamicDataSourceType.Method)]
+		public void When_UsingLanguage_Then_InitializeCorrectly(string languageTag, string numeralSystem, string resolveLanguage)
+		{
+			var numeralSystemTranslator = new NumeralSystemTranslator(new string[] { languageTag });
+			Assert.AreEqual(numeralSystem, numeralSystemTranslator.NumeralSystem);
+			Assert.AreEqual(resolveLanguage, numeralSystemTranslator.ResolvedLanguage);
+		}
```
MohammadHadi2031, line 888: @jeromelaban I figure out that Android UI Tests and iOS Automated Runtime Tests failed because of the error below:
Failed: When_UsingLanguage_Then_InitializeCorrectly() [System.Reflection.TargetParameterCountException] 
System.Reflection.TargetParameterCountException: Number of parameters specified does not match the expected number.

but this test passed perfectly when target framework is NET461 or UWP. I don't know what is the reason it failed on Android and IOS. Can you help me?

jeromelaban: It fails on iOS/Android because `DynamicData` is not yet supported by the unit test engine there. `DataRow` is supported, though:

https://github.com/unoplatform/uno/blob/b18a4d0355a07731a3e6d88f058c0db69bfe7567/src/SamplesApp/SamplesApp.UnitTests.Shared/Controls/UnitTest/UnitTestsControl.cs#L566

src/Uno.UWP/Globalization/NumberFormatting/NumeralSystemTranslatorHelper.cs
```diff
@@ -0,0 +1,3566 @@
...
+				case "Tibt":
+					return TibtDigits;
+				case "Vaii":
+					return VaiiDigits;
+				default:
+					throw new ArgumentException("The parameter is incorrect.");
+			}
+		}
+
+		public static string ToPascalCase(string numeralSystem)
```
MohammadHadi2031, line 3485: @jeromelaban I Added this method because NumeralSystem property is case insensitive but It is supposed to display NumeralSystem in a standard way.

src/Uno.UI.Tests/Windows_Globalization/Given_NumeralSystemTranslator.cs
```diff
@@ -0,0 +1,1613 @@
...
+		[DataRow("SAUR", "Saur")]
+		[DataRow("SUND", "Sund")]
+		[DataRow("TALU", "Talu")]
+		[DataRow("TAMLDEC", "TamlDec")]
+		[DataRow("TELU", "Telu")]
+		[DataRow("THAI", "Thai")]
+		[DataRow("TIBT", "Tibt")]
+		[DataRow("VAII", "Vaii")]
+
+		public void When_NumeralSystemCaseIsNotStandard(string numeralSystem, string expected)
```
MohammadHadi2031, line 1605: And the corresponding test

MartinZikmund (approved):
Very nice work!

jeromelaban:
I'm making a change in #7199 so that we can grab the UAP artifacts for easier validation of the runtime tests. We'll merge this PR after a rebuild with the artifact.

src/Uno.UWP/Globalization/NumberFormatting/NumeralSystemTranslator.cs
```diff
@@ -0,0 +1,169 @@
...
+
+		public IReadOnlyList<string> Languages { get; }
+
+		public string ResolvedLanguage { get; }
+
+		private void ValidateLanguages(IEnumerable<string> languages)
+		{
+			if (languages is null)
+			{
+				throw new NullReferenceException("Invalid pointer");
```
Youssef1313, line 49: Should this be `NullArgumentException`?

src/Uno.UWP/Globalization/NumberFormatting/NumeralSystemTranslator.cs
```diff
@@ -0,0 +1,169 @@
...
+using System.Linq;
+using Uno.Globalization.NumberFormatting;
+
+namespace Windows.Globalization.NumberFormatting
+{
+	public partial class NumeralSystemTranslator
+	{
+		private string _numeralSystem;
+
+		public NumeralSystemTranslator() : this(new string[] { "en-US" })
```
Youssef1313, line 13: Wouldn't this allocate unnecessarily? Consider extracting this as an `ImmutableArray` static field.

src/Uno.UWP/Globalization/NumberFormatting/NumeralSystemTranslator.cs
```diff
@@ -0,0 +1,169 @@
...
+			foreach (var language in languages)
+			{
+				NumeralSystemTranslatorHelper.GetNumeralSystem(language);
+			}
+		}
+
+		public string TranslateNumerals(string value)
+		{
+			if (NumeralSystem.Equals("Arab", StringComparison.InvariantCultureIgnoreCase) ||
+				NumeralSystem.Equals("ArabExt", StringComparison.InvariantCultureIgnoreCase))
```
Youssef1313, line 66: I'm not sure if this should be Ordinal or InvariantCulture.

src/Uno.UWP/Globalization/NumberFormatting/NumeralSystemTranslatorHelper.cs
```diff
@@ -0,0 +1,3566 @@
...
+				case "af":
+					return "Latn";
+				case "af-na":
+					return "Latn";
+				case "af-za":
+					return "Latn";
+				case "agq":
+					return "Latn";
+				case "agq-cm":
+					return "Latn";
```
Youssef1313, line 71: Have you got this list from somewhere?

src/Uno.UWP/Globalization/NumberFormatting/NumeralSystemTranslator.cs
MohammadHadi2031: I added a test for this at lin 39 in Given_NameralSystemTranslator.cs

src/Uno.UWP/Globalization/NumberFormatting/NumeralSystemTranslator.cs
MohammadHadi2031: I just googled it and I think it should be Ordinal. NumeralSystem is always pascal case so we can perform case-sensitive comparison. @MartinZikmund Do you agree?

src/Uno.UWP/Globalization/NumberFormatting/NumeralSystemTranslatorHelper.cs
MohammadHadi2031: There is a link in CultureInfo class documentation that shows supported languages by windows. "see the Language tag column in the list of language/region names supported by Windows. Culture names follow the standard defined by BCP 47"
[The link](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-lcid/a9eac961-e77d-41a6-90a5-ce1a8b0cdb9c)

src/Uno.UWP/Globalization/NumberFormatting/NumeralSystemTranslator.cs
MohammadHadi2031: Nice comment. Immutable Array is not supported by Net461 do you agree with private static readonly array?

src/Uno.UWP/Globalization/NumberFormatting/NumeralSystemTranslatorHelper.cs
```diff
@@ -0,0 +1,3566 @@
...
+				case "zu":
+					return "Latn";
+				case "zu-za":
+					return "Latn";
+				default:
+					throw new ArgumentException("The parameter is incorrect.");
+			}
+		}
+
+		public static string GetResolvedLanguage(string languageTag)
```
Youssef1313, line 1727: - `new Windows.Globalization.NumberFormatting.NumeralSystemTranslator(new[] { "zh-Hans-CN" })` doesn't throw in UWP. Looks like this code will throw?. (Note that `zh-Hans-CN` produces `ResolvedLanguage` = `zh-Hans-CN`, but `fr-TEST` will produce `en-US`.

- I think this logic is common between the `Language` class as well, and also many other globalization APIs. So we might in the future want to move it somewhere less-specific to `NumberFormatting`.

MohammadHadi2031: I will check zh-Hans-CN thanks. Do you have complete list of language tags to prevent missing one?

I don't understand this part "but fr-TEST will produce en-US" what is fr-test?

src/Uno.UWP/Globalization/NumberFormatting/NumeralSystemTranslator.cs
MartinZikmund: Yes it seems like `Ordinal` should be enough

src/Uno.UWP/Globalization/NumberFormatting/NumeralSystemTranslator.cs
MartinZikmund: Readonly array should be enough, but I think even the current solution is good enough. I don't see a scenario where there would be thousands of instances of `NumeralSystemTranslator` and the allocation would be a problem 🤔 .

src/Uno.UWP/Globalization/NumberFormatting/NumeralSystemTranslatorHelper.cs
Youssef1313: `fr-TEST` is just an invalid string I tried 😄

Youssef1313: > Do you have complete list of language tags to prevent missing one

I don't, I just knew about it since I tried to work on globalization-related fixes and got this test failing:

https://github.com/unoplatform/uno/blob/master/src/Uno.UI.RuntimeTests/Tests/Windows_Globalization/Given_ApplicationLanguages.cs#L21

Youssef1313: Some information I got from Microsoft:

---

```
The logic of language tags is based on the RFCs and other considerations.  However, because of geopolitical and other considerations, this space is not stable.  Tags can (and do) change between Windows versions. 

 

Since Windows doesn’t agree with other Windows versions, it isn’t practical to expect that another implementation would agree with either of those.

 

fr-Latn-CA has a “suppress-script” flag because fr-CA is sufficient as it’s always Latin.  Some other languages like az-Cyrl or az-Latn keep the script because the language may be written in multiple scripts.

 

Windows also sometimes choses only the language name when there’s typically no ambiguity over language.  Eg: there are lots of “versions” of English; en-CA, en-US, en-AU, etc… but only one version of Japanes:  ja.  The JP in ja-JP is therefore somewhat redundant.

 

Chinese is a little more confusing is different vendors have different expectations of the language tags for Chinese.  In particular, the Cantonese vs Mandarin differences.

 

Additionally, some tags are “aliased” because the standard names &/or the way that the industry handled those languages shifted over time.

 

Furthermore, any tag for which we have limited data may not be normalized in such a fashion.

 

Generally, the recommendation is to be flexible in what language tag(s) are accepted and used as they may change over time.

 

IETF language tag - Wikipedia

 

The “suppress script” tags are defined in IANA.  https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry

 

The tags for which the region is dropped is somewhat arbitrary and may change.  Frankly, for an alternate implementation, I’d probably be happy if the region wasn’t dropped.
```

---

```
Regarding the behavior of the Language constructor in Windows.Globalization:

 

> When we call `new Language("fr-Latn-CA")`, the language tag becomes "fr-CA" (the middle part dropped)
> However, when we call the ctor passing "zh-Hans-CN", nothing drops in LanguageTag


The reason that the script subtag (Latn, Hans) is sometimes dropped (or kept) depends on if the language subtag has a “Suppress-Script” defined by IANA in the Language Subtag Registry.

 

The IANA subtag registry is here: https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry

 

In the BCP47 spec / RFC5646 spec (https://www.rfc-editor.org/rfc/rfc5646.html) in section 4.1 says


The script subtag SHOULD NOT be used to form language tags unless
the script adds some distinguishing information to the tag.
 

So if the language has a suppress-script defined, then it is omitted from the overall tag.

 

For example, for French (fr), there is a suppress script defined of Latin/Latn:

 

%%

Type: language

Subtag: fr

Description: French

Added: 2005-10-16

Suppress-Script: Latn

 

This is why fr-Latn-CA ends up with fr-CA, as the Latn part is redundant.

 

The internal implementation has a partial copy of the IANA subtag registry, which is used to determine the suppress-script (if any).

 

 

> Moreover, calling it  with "az-Latn-AZ", the last part drops:


The dropping of the region part of the tag is unfortunately much less well defined.

AFAIK, there is no requirement for this in the BCP47 spec, though it does say in section 4.1 to:

 

Use as precise a tag as possible, but no more specific than is
justified.  Avoid using subtags that are not important for
distinguishing content in an application.
 

So for some tags, the regional part isn’t considered distinguishing, and it is omitted.

For example, ja-JP drops the region part JP, and is just “ja”, as the thinking was that there isn’t any distingusing data added by the region JP for the language Japanese.

 

There is no external data source for what is considered distingusing for the regoinal subtag, so you’d likely need to create a table to replicate it.
```

---


```
I’d like to reiterate that we pick up updated suppress-script data on a sort of ad-hoc basis.  Best practice would probably just be to use whatever’s current and recognize that it may differ from the behavior of any specific version of Windows.

 

And, for the dropping of the region, you could probably query a current Windows box with all of the registered IANA tags and start with that.  But as Jeff says, there’s no standard and that’s far more likely to change.
```

Youssef1313: Please note that information in the previous comment was an answer to `Windows.Globalization.Language` class. I'm posting assuming they follow the same behavior.

MohammadHadi2031: Thanks for the information. It seems that the resolved language is your default windows language when the input language tag is "tag-InvalidSomething". I think [Language ](https://github.com/unoplatform/uno/blob/287f9785be664ca577775b1bb6b90fe68bbd26fb/src/Uno.UWP/Globalization/Language.cs#L9)class needs work because CultureInfo throws exception when input is "zh-cmn-Hans" but Language doesn't. I'm confused :)

Youssef1313: Yeah `Language` is currently buggy. It has been preventing me from fixing `PrimaryLanguageOverride` (https://github.com/unoplatform/uno/pull/6906)

src/Uno.UWP/Globalization/NumberFormatting/NumeralSystemTranslator.cs
jeromelaban: Extracting to a readonly static field could be enough indeed, even if the number of instances is likely not going to be very large.

src/Uno.UWP/Globalization/NumberFormatting/NumeralSystemTranslatorHelper.cs
jeromelaban: Let's add this link in the code comments, for future updates.

src/Uno.UWP/Globalization/NumberFormatting/NumeralSystemTranslatorHelper.cs
MohammadHadi2031: It seems that for best implementation RFCs should be read this needs time. Can we merge this and then complete the implementation in the future? Or we can mark the NumeralSystemTranslator constructor with parameter as not implemented. Users can set NumeralSystem property manually and call TranslateNumeral.

MohammadHadi2031:
@jeromelaban I wrote some code to generate switch/case block of GetNumeralSystem and GetResolvedLanguage. Also generates test methods. Input is a csv file contains language tags supported by windows. The code extract language tags and generates desired pieces of code. Is it useful to add it to Uno source code?

jeromelaban:
I've run the Unit Tests in the UWP sample app, and there are 11 failures:

image

You can try running the app using the UWP package from the CI artifacts:

image

https://dev.azure.com/uno-platform/Uno%20Platform/_build/results?buildId=34094&view=artifacts&pathAsName=false&type=publishedArtifacts

MohammadHadi2031:
Thanks I will check it

MohammadHadi2031:
@jeromelaban I couldn't install UWP artifact sample app but I've run the unit test on my local machine and I found out not all language tags listed in [here ](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-lcid/a9eac961-e77d-41a6-90a5-ce1a8b0cdb9c)are supported by UWP. When the input language tag is not supported I saw two behaviour:
1) Not supported language is listed in [IANA subtag registry](https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry) the resolved language would be your default system language.
2) Not supported language form is like "LanguageA-InvalidSomething". The resolved language would be "LanguageA"

Now can I remove failed language tags case from switch/case in GetResolveLanguage and GetNumeralSystem methods and maybe fix it in the future?

jeromelaban:
The UWP samples app is not running as part of the CI yet, which is why it's not visible.

You may disable the tests in the UWP head (using `#if NETFX_CORE`) so that we know it's ignore for a good reason.

src/Uno.UI.Tests/Windows_Globalization/Given_NumeralSystemTranslator.cs
```diff
@@ -143,18 +143,22 @@ public void When_LanguagesIsNull_Then_Throw()
...
 		[DataRow("ku", "Arab", "ku")]
 		[DataRow("ku-Arab", "Arab", "ku-Arab")]
 		[DataRow("ku-Arab-IQ", "Arab", "ku-Arab")]
+#if !NETFX_CORE
 		[DataRow("ccp", "ArabExt", "fa")]
 		[DataRow("ccp-Cakm", "ArabExt", "fa")]
 		[DataRow("ccp-Cakm-BD", "ArabExt", "fa")]
 		[DataRow("ccp-Cakm-IN", "ArabExt", "fa")]
+#endif
 		[DataRow("ce-RU", "Latn", "ce-Cyrl-RU")]
```
MohammadHadi2031: @jeromelaban I disabled tests for UWP. Did I do this correctly?

src/Uno.UI.Tests/Windows_Globalization/Given_NumeralSystemTranslator.cs
```diff
@@ -0,0 +1,1622 @@
...
+		[DataRow("ku", "Arab", "ku")]
+		[DataRow("ku-Arab", "Arab", "ku-Arab")]
+		[DataRow("ku-Arab-IQ", "Arab", "ku-Arab")]
+#if !NETFX_CORE
+		[DataRow("ccp", "ArabExt", "fa")]
+		[DataRow("ccp-Cakm", "ArabExt", "fa")]
+		[DataRow("ccp-Cakm-BD", "ArabExt", "fa")]
+		[DataRow("ccp-Cakm-IN", "ArabExt", "fa")]
+#endif
+		[DataRow("ce-RU", "Latn", "ce-Cyrl-RU")]
```
MohammadHadi2031, line 162: @jeromelaban I disabled tests for UWP. Did I do this correctly?

MohammadHadi2031: @jeromelaban  On my computer number of failed tests was 9. I don't know why it was 11 for you. Please tell me if tests failed again.

MohammadHadi2031:
@jeromelaban Is this ready to merge?

jeromelaban:
Thanks for the update. It may definitely be locale/environment related. Here's what I'm seeing now:image

and

image

MohammadHadi2031:
@jeromelaban I disabled the two tests. Can we have a merge today?

MohammadHadi2031:
@jeromelaban Sorry for interruption. Would you mind rerunning failed test. I can't wait for merge

jeromelaban (approved):

jeromelaban:
Thanks for your contribution!

Issue #7123, opened 2021-09-17
Update NumberBox from WinUI sources


Youssef1313:
### Current behavior

https://github.com/unoplatform/uno/pull/7122

### Expected behavior

Revert back

Youssef1313:
I can't really imagine how porting `NumberBox` will break such a test. It's completely irrelevant. The only piece of information I could find is that things fail in `TryStaticRetrieval` (in `ResourceResolver`), but no clue how `NumberBox_v1.xaml` affects that.

Youssef1313:
I narrowed down more to that `GetActiveThemeDictionary` is returning null. It's relying on an invalid cached (null) value.

jeromelaban:
Indeed, that's what I found as well. It feels like we should be observing the theme dictionaries to determine if the value should be updated.

jeromelaban:
btw, It's not the PR itself that broke the test, it just raised an unrelated issue because of initialization order... so it's not your fault ;)

MartinZikmund:
@duke7553 was working on updating the source code of `NumberBox`. If you are updating the styles only, that should be fine, but may be useful to check with him to avoid potential conflicts 🙂

Youssef1313:
@MartinZikmund I recall I updated everything. In case @duke7553 had already done it, it might be good idea to compare both branches to check if any of us has missed anything.

PR #7134, opened 2021-09-18
Invalidate cached theme dictionary when theme dictionaries are added/removed

Youssef1313:
GitHub Issue (If applicable): Part of #7123

## PR Type

- Bugfix

Cached value doesn't get properly updated when needed.

## What is the new behavior?

Update cached value properly.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI.Tests/Windows_UI_Xaml/Given_ResourceDictionary.cs
+++ b/src/Uno.UI.Tests/Windows_UI_Xaml/Given_ResourceDictionary.cs
@@ -795,5 +795,84 @@ namespace Uno.UI.Tests.Windows_UI_Xaml
 			Assert.IsTrue(resources.ContainsKey("TestKey"));
 			Assert.AreEqual(resources["TestKey"], "Test123");
 		}
+
+		[TestMethod]
+		public void When_Theme_Dictionary_Is_Cached_Then_Add_And_Clear_Theme()
+		{
+			var dictionary = new ResourceDictionary();
+			dictionary.TryGetValue("TestKey", out _); // This causes _activeThemeDictionary to be cached.
+
+			var lightTheme = new ResourceDictionary();
+			lightTheme.Add("TestKey", "TestValue");
+			dictionary.ThemeDictionaries.Add("Light", lightTheme); // Cached value is no longer valid due to adding theme.
+
+			// Make sure the cache is updated.
+			Assert.IsTrue(dictionary.TryGetValue("TestKey", out var testValue));
+			Assert.AreEqual("TestValue", testValue);
+
+			dictionary.ThemeDictionaries.Clear(); // Cached value is no longer valid due to clearing themes.
+
+			Assert.IsFalse(dictionary.TryGetValue("TestKey", out _));
+		}
+
+		[TestMethod]
+		public void When_Theme_Dictionary_Is_Cached_Then_Add_And_Remove_Theme()
+		{
+			var dictionary = new ResourceDictionary();
+			dictionary.TryGetValue("TestKey", out _); // This causes _activeThemeDictionary to be cached.
+
+			var lightTheme = new ResourceDictionary();
+			lightTheme.Add("TestKey", "TestValue");
+			dictionary.ThemeDictionaries.Add("Light", lightTheme); // Cached value is no longer valid due to adding theme.
+
+			// Make sure the cache is updated.
+			Assert.IsTrue(dictionary.TryGetValue("TestKey", out var testValue));
+			Assert.AreEqual("TestValue", testValue);
+
+			Assert.IsTrue(dictionary.ThemeDictionaries.Remove("Light")); // Cached value is no longer valid due to removing theme.
+
+			Assert.IsFalse(dictionary.TryGetValue("TestKey", out _));
+		}
+
+
+		[TestMethod]
+		public void When_Default_Theme_Dictionary_Is_Cached()
+		{
+			var defaultDictionary = new ResourceDictionary();
+			defaultDictionary.Add("TestKey", "TestValueFromDefaultDictionary");
+
+			var dictionary = new ResourceDictionary();
+			dictionary.ThemeDictionaries.Add("Default", defaultDictionary);
+			Assert.IsTrue(dictionary.TryGetValue("TestKey", out var testValue));
+			Assert.AreEqual("TestValueFromDefaultDictionary", testValue);
+
+			var lightDictionary = new ResourceDictionary();
+			lightDictionary.Add("TestKey", "TestValueFromLightDictionary");
+			dictionary.ThemeDictionaries.Add("Light", lightDictionary);
+			Assert.IsTrue(dictionary.TryGetValue("TestKey", out testValue));
+			Assert.AreEqual("TestValueFromLightDictionary", testValue);
+		}
+
+		[TestMethod]
+		public void When_Default_Theme_Dictionary_Should_Be_Used_After_Removing_Active_Theme()
+		{
+			var lightDictionary = new ResourceDictionary();
+			lightDictionary.Add("TestKey", "TestValueFromLightDictionary");
+
+			var dictionary = new ResourceDictionary();
+			dictionary.ThemeDictionaries.Add("Light", lightDictionary);
+			Assert.IsTrue(dictionary.TryGetValue("TestKey", out var testValue));
+			Assert.AreEqual("TestValueFromLightDictionary", testValue);
+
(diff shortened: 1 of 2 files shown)


src/Uno.UI.Tests/Windows_UI_Xaml/Given_ResourceDictionary.cs
```diff
@@ -795,5 +795,20 @@ public void When_Custom_Resource_Dictionary_With_Custom_Property_in_Custom_Contr
...
+
+		[TestMethod]
+		public void When_Theme_Dictionary_Is_Cached()
+		{
+			var dictionary = new ResourceDictionary();
+			dictionary.TryGetValue("TestKey", out _); // This causes _activeThemeDictionary to be cached.
+
+			var lightTheme = new ResourceDictionary();
+			lightTheme.Add("TestKey", "TestValue");
+			dictionary.ThemeDictionaries.Add("Light", lightTheme); // Cached value is no longer valid.
```
Youssef1313, line 807: I'm not sure hard coding "Light" here is correct.

src/Uno.UI/UI/Xaml/ResourceDictionary.cs
```diff
@@ -48,7 +72,33 @@ public Uri Source
...
+				_themeDictionaries = new ResourceDictionary();
+				_themeDictionaries.ResourceDictionaryValueChange += (sender, e) =>
+				{
+					// The purpose of this to updated the cached value when it's no longer valid.
+					// See GetActiveThemeDictionary.
+					if (e.ChangedKey is ResourceKey changedKey && changedKey.Equals(_activeTheme))
+					{
+						// Either a key is added (e.NewValue will be non-null), or
+						// a key is removed (e.NewValue will be null).
+						_activeThemeDictionary = e.NewValue;
```
davidjohnoliver, line 90: I wonder what happens if the dictionary was falling back on the "Default" theme (or now should be), though we may be getting into the edge case of an edge case.

I also wonder how the case of nested dictionaries in ThemeDictionaries is handled (both by this and by the existing caching logic), since ResourceDictionary.TryGetValue (which is called by GetActiveThemeDictionary()) will search into nested (merged and theme) dictionaries.

I'm curious what the actual scenario was that you ran into where you noticed this bug?

Youssef1313: > I'm curious what the actual scenario was that you ran into where you noticed this bug?

@davidjohnoliver It was a simple port of NumberBox from microsoft-ui-xaml repo that revealed the bug.

---

> I wonder what happens if the dictionary was falling back on the "Default" theme (or now should be), though we may be getting into the edge case of an edge case.

This is really very interesting. I'll see if I can reproduce in a unit test and look into how would I fix it.

Youssef1313:
@davidjohnoliver Can you see the updated logic please?

src/Uno.UI/UI/Xaml/ResourceDictionary.cs
davidjohnoliver: I get that it was related to the earlier NumberBox PR, but I was wondering about what was actually happening that triggered this case.

Youssef1313: @davidjohnoliver I couldn't really know the answer. @jeromelaban Do you know?

src/Uno.UI/UI/Xaml/ResourceDictionary.cs
```diff
@@ -48,7 +72,41 @@ public Uri Source
...
+					// The purpose of this to updated the cached value when it's no longer valid.
+					// See GetActiveThemeDictionary.
+					if (e.ChangedKey is ResourceKey changedKey && changedKey.Equals(_activeTheme))
+					{
+						if (e.NewValue is null && !_activeTheme.Equals(Themes.Default))
+						{
+							// If the key is removed and we have non-default active theme,
+							// then update cached value with default theme.
+							// We do that here because GetActiveThemeDictionary does that.
+							_activeThemeDictionary = GetThemeDictionary(Themes.Default);
```
davidjohnoliver, line 93: Since you're now setting _activeTheme from GetThemeDictionary(), this will set _activeTheme to Default, which seems strange?

Instead of trying to recalculate _activeThemeDictionary here, can you simply set _activeTheme to ResourceKey.Empty? Then _activeThemeDictionary will be correctly re-resolved the next time `GetActiveThemeDictionary()` is called.

Youssef1313: Doing so is still problematic for the following case:

- Active theme is Light but we only have Default
- We assign `Light` to `_activeTheme`, and the default dictionary to `_activeThemeDictionary` (through the `?? GetThemeDictionary(Themes.Default)`
- Later, "Default" gets removed

For this scenario, we shouldn't have anything available, but we'll be caching the "Default" theme if the `_activeTheme` wasn't set to `Default`. Is there a better approach that handles this scenario?

davidjohnoliver: But it's problematic in the case that we have cached the "Default" dictionary, and later on "Light" gets added, or no?

Can you just invalidate `_activeTheme` when _any_ dictionary is added/removed from ThemeDictionaries? That seems safest, and modifying `ThemeDictionaries` is sufficiently uncommon that it hopefully shouldn't pose a performance problem.

Youssef1313: > But it's problematic in the case that we have cached the "Default" dictionary, and later on "Light" gets added, or no?

In that case, the next time `GetActiveThemeDictionary` is called won't use the cached value because `_activeTheme` is "Default" but we're requesting "Light". So it shouldn't be problematic?

Anyway I agree that invalidating _activeTheme whenever a dictionary is added or removed is safest. But I want to make sure we have a good test coverage in case someone adds the caching back in future :)

davidjohnoliver: > In that case, the next time GetActiveThemeDictionary is called won't use the cached value because _activeTheme is "Default" but we're requesting "Light". So it shouldn't be problematic?

It sounds like it's the caching itself that would be broken in that case, but I can't see the code anymore, so never mind. :D

And yes, whatever the fix, good test coverage is useful. ;)

Youssef1313: @davidjohnoliver Do you mean the current implementation is okay, or I should still be invalidating the cache for any theme dictionary change? or both are okay (in which case I'd like to see if @jeromelaban has any preference)?

src/Uno.UI.Tests/Windows_UI_Xaml/Given_ResourceDictionary.cs
jeromelaban: You could use a nameof from the RequestedTheme, but that's unlikely to change.

src/Uno.UI/UI/Xaml/ResourceDictionary.cs
jeromelaban: This is happening because some resources initialized in the NumberBox updates are changing the initialization order, causing some early evaluations. It could be something related to the lazy evaluation you've been looking into @davidjohnoliver.

src/Uno.UI/UI/Xaml/ResourceDictionary.cs
```diff
@@ -21,6 +21,30 @@ public partial class ResourceDictionary : DependencyObject, IDependencyObjectPar
 		private readonly List<ResourceDictionary> _mergedDictionaries = new List<ResourceDictionary>();
 		private ResourceDictionary _themeDictionaries;
 
+		/// <summary>
+		/// This event is fired when a key that has value of type <see cref="ResourceDictionary"/> is added or changed in the current <see cref="ResourceDictionary" />
+		/// </summary>
+		private event EventHandler<ResourceDictionaryChangedEventArgs> ResourceDictionaryValueChange;
+
+		private class ResourceDictionaryChangedEventArgs
```
Youssef1313, line 29: This class is no longer needed since we invalidate the cache for any changes now. But it may (or may not) be used later. I'm not sure if I should remove it or keep it. @davidjohnoliver @jeromelaban What do you think?

davidjohnoliver: Personally I would remove it if it's not used, it's internal so there's no difficulty in bringing it back if it becomes necessary.

jeromelaban (approved):
Thanks for your contribution!

PR #7137, opened 2021-09-19
`StartStopEventWrapper`

MartinZikmund:

## PR Type

Each sensor class has its own implementation of event start/stop management.

## What is the new behavior?

Using shared start/stop infrastructure, which is covered by tests.

## PR Checklist

### DIFF
--- a/src/Uno.UI.Runtime.Skia.Gtk/Helpers/Dpi/DpiHelper.cs
+++ b/src/Uno.UI.Runtime.Skia.Gtk/Helpers/Dpi/DpiHelper.cs
@@ -10,7 +10,7 @@ namespace Uno.UI.Runtime.Skia.Gtk.Helpers.Dpi;
 
 internal class DpiHelper
 {
-	private readonly StartStopEventWrapper<EventHandler> _dpiChangedWrapper;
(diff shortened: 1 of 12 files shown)


src/Uno.UWP/Devices/Lights/Lamp.unsupported.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Windows.Devices.Lights
+{
+	public partial class Lamp
+	{
+		/// <summary>
+		/// API not supported, always returns null.
+		/// </summary>
+		/// <returns>Null.</returns>
+		public static IAsyncOperation<Lamp> GetDefaultAsync() => Task.FromResult<Lamp>(null).AsAsyncOperation();
```
Youssef1313, line 11: Not implemented attribute?

MartinZikmund: In this case I think it is okay - the sensor itself is not available on the target platforms (WPF, GTK, macOS), so returning `null` is valid - when the device doesn't have a `Lamp`, the API does return `null`. The advantage of this as compared to having the method throw is that the developer can use the code without any `#if` - comfortable that the code will never enter the branch after `if (lamp != null)`

PR #7535, opened 2021-11-26
Support non-default `xmlns` and `ConvertFromStringAttribute` in `XamlReader`

MartinZikmund:
GitHub Issue (If applicable): closes https://github.com/unoplatform/uno/issues/7536

## PR Type

- Bugfix
- Feature

Not supported.

## What is the new behavior?

- When `XamlReader` loads XAML with non-default `xmlns` with "using:" or "clr-namespace:", it can handle it porperly.
- When string value of a property of type annotated with `ConvertFromStringAttribute` is read, it is properly handled in `XamlReader`.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Markup/Reader/XamlObjectBuilder.cs
+++ b/src/Uno.UI/UI/Xaml/Markup/Reader/XamlObjectBuilder.cs
@@ -5,7 +5,6 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using System.Reflection;
-using System.Text;
 using System.Text.RegularExpressions;
 using Uno.Extensions;
 using Uno.UI;
@@ -15,9 +14,8 @@ using Uno.Xaml;
 using Windows.UI.Xaml.Controls;
 using Windows.UI.Xaml.Data;
 using Windows.UI.Xaml.Documents;
-using Windows.UI;
-using Windows.Foundation;
 using Windows.UI.Text;
+using Windows.Foundation.Metadata;
 
 #if XAMARIN_ANDROID
 using _View = Android.Views.View;
@@ -38,7 +36,7 @@ namespace Windows.UI.Xaml.Markup.Reader
 		private Queue<Action> _postActions = new Queue<Action>();
 		private static readonly Regex _attachedPropertMatch = new Regex(@"(\(.*?\))");
 
-		private static Type[] _genericConvertibles = new []
+		private static Type[] _genericConvertibles = new[]
 		{
 			typeof(Media.Brush),
 			typeof(Media.SolidColorBrush),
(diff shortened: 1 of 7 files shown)


src/Uno.UI/UI/Xaml/Markup/Reader/XamlObjectBuilder.cs
```diff
@@ -822,6 +821,39 @@ private object BuildLiteralValue(XamlMemberDefinition member, Type propertyType
...
 		{
+			if (propertyType.GetCustomAttribute<CreateFromStringAttribute>() is { } createFromString)
+			{
+				var sourceType = propertyType;
+				var methodName = createFromString.MethodName;
+				if (createFromString.MethodName.Contains("."))
+				{
+					var splitIndex = createFromString.MethodName.LastIndexOf(".");
+					var typeName = createFromString.MethodName.Substring(0, splitIndex);
+					sourceType = AppDomain.CurrentDomain
```
jeromelaban, line 832: TypeResolver should be able to provide this type of lookup

MartinZikmund: Unfortunately it expects a different format (`ns:Type` or just `Type`), whereas in this case I have full type name

MartinZikmund: This approach did not work due to the change I made in this PR - I added this line - https://github.com/unoplatform/uno/blob/47aed3f70f56edcbced643b9119a7489f3cbfd54/src/Uno.UI/UI/Xaml/Markup/Reader/XamlTypeResolver.cs#L359

This covers the scenario where the xmlns of the file is not standard - e.g. xmlns="using:Uno.UI.Tests.Windows_UI_Xaml_Markup.XamlReaderTests which is something that UWP handles properly. This line prepends the default namespace before the type name so "XYZ" becomes "Uno.UI.......XamlReaderTests.XYZ" - which is needed for Type.GetType(name) to work. However, when the type passed into XamlTypeResolver is already fully qualified, it causes it to become "Uno.UI.......XamlReaderTests.Uno.UI.......XamlReaderTests.XYZ" . Not sure how to resolve this to satisfy both scenarios, as it is not really possible to tell whether the string is a qualified or non-qualified type, other than checking if indexof "." is -1.

However, I added this, which makes a fallback to use the original type and that resolves this:

https://github.com/unoplatform/uno/pull/7535/files#diff-9c43ddb90b5f877f348d20114e11c095881eaea4c50d5df26db6f4d359ee2ae5R378

jeromelaban: That works as well!

jeromelaban:
Looks like there are some nullability errors in the XamlReader

carldebilly (approved):

jeromelaban (approved):

Issue #4309, opened 2020-10-19
Skia Rendering Backend Support


jeromelaban:
This issue is an epic that tracks opened issues and larger items to improve the Skia backend support, for Linux and other platforms.

- [ ] Improved TextBlock support
- [ ] TextBox support
- [ ] Keyboard events support
- [ ] DPI Scaling support
- [ ] Automated testing
- [ ] Compositor Thread
- [ ] Composition Animations support
- [ ] Acrylic brush support
- [ ] ReadyToRun support for x64, arm64
- [ ] Single file packaging
- [ ] Per layer runtime selection (Skia rendering + non-UI per platform)

microhobby:
@jeromelaban, and maybe also @mattleibow, could you help me with some questions?

I think SkiaSharp runs with OpenGL/GPU graphics acceleration, so my question: today the render that we have in Gtk uses OpenGL/GPU? I did not find anything that could affirm this, in some tests I had a certain increase in the use of the GPU (maybe this increase is just from the Gtk shell), but when using in some Linux embedded systems with lower CPU power the performance is much worse,  seems that everything is being done in CPU. Looking at the code I didn't find anything creating the openGL context.

And the second question, if we are not using openGL/GPU in SkiaSharp for Gtk what would be the effort to implement this? (it would be easy, difficult, we already have something to base on ....)

jeromelaban:
@microhobby at this point, the Uno implementation for the Skia backend uses a derivative of the SkiaSharp.Views GTK View which is using software rendering. There's no official OpenGL implementation of the GTK rendering, so we do no know the kind of effort would be needed to enable this. @mattleibow may know more.

In all cases, we have that on the backlog and we'll investigate it as part of the Linux support effort.

mattleibow:
At this point, there was not much demand for GTK views so I just threw in this commented code. https://github.com/mono/SkiaSharp/blob/main/source/SkiaSharp.Views.Forms/SkiaSharp.Views.Forms.GTK/SKGLWidget.cs

It theoretically should be quite easy to use if we use the `GLWidget` from OpenTK. Not exactly sure the issue that I had... Might have been just the fact that there was a few assembly conflicts because Forms used to bundle its own copy in the nuget. But I think that changed.

Until I can get to it, someone could have a stab at this.

[...]

PR #7665, opened 2021-12-17
fix: [GTK] Incorrect window ActualWidth and ActualHeight on WSL and Linux

### LINKED ISSUE #5666: [Skia.Gtk] ScrollViewer ScrollBar not shown
## Current behavior
XAML has:
`ScrollViewer.HorizontalScrollBarVisibility="Visible" `
`ScrollViewer.VerticalScrollBarVisibility="Visible" `

but no scrollbars are not shown on Skia.Gtk aka Linux

## Expected behavior
scrollbars should be shown

[Uno.ZoomPanControl.zip](https://github.com/unoplatform/uno/files/6252953/Uno.ZoomPanControl.zip)

## Workaround
none

## Environment
Nuget Package:

- [ ] Uno.UI / Uno.UI.WebAssembly / Uno.UI.Skia
- [ ] Other: 

Nuget Package Version(s):

Affected platform(s):

- [ ] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] macOS
- [ ] Skia
  - [ ] WPF
  - [x] GTK (Linux)
  - [ ] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

IDE:

- [ ] Visual Studio 2019 (version: )
- [ ] Rider macOS (version: )
- [x] Visual Studio Code (version: )

Relevant plugins:

- [ ] Resharper (version: )

mikebmcl:
GitHub Issue (If applicable): closes #5666 

[ScrollViewer ScrollBar not shown #5666](https://github.com/unoplatform/uno/issues/5666)
which is part of the larger epic
[Skia Rendering Backend Support #4309](https://github.com/unoplatform/uno/issues/4309)

## PR Type

- Bugfix

In Skia.Gtk, the ActualWidth and ActualHeight of the window are set to the width and height values for the Gtk.Window.

When compiled and run on Windows using Gtk for Windows everything appears correct.

On WSL and Linux, likely specific to Wayland-based windowing, the values for the Gtk.Window's width and height includes the window's title bar and other non client area parts that are outside of the drawable area. This causes the Uno MainWindow to report and do layout, etc., using ActualWidth and ActualHeight values that are larger than they should be.

As a result, content that should be rendering inside the window instead renders past the right and bottom of the window resulting in things such as the scrollbars for the ScrollViewer control or the right and bottom part of the border in the SampleApp included in this pull request to appear to be missing.

## What is the new behavior?

The values for ActualWidth and ActualHeight are taken from the window's drawable area in GtkHost rather than using the window's width and height. Scrollbars for full-window ScrollViewers, full-window Borders, and anything else that is the size of or that is laid out to be on the right side or bottom of the window are now visible on WSL.

I am currently unable to test it on Linux but due to the nature of the change it should fix that as well.

The behavior of Skia.Gtk using Gtk for Windows remains identical to its previous behavior and the bug fix only affects Skia.Gtk so it does not change the behavior of any other platforms, tested and verified on UWP, Skia.Wpf, WASM, and Android.

## PR Checklist

- [ ] Docs have been added/updated which fit documentation template (for bug fixes / features)
- [x] Unit Tests and/or UI Tests for the changes have been added (for bug fixes / features) (if applicable)
- [x] Associated with an issue (GitHub or internal) and uses the automatic close keywords.
- [x] Commits must be following the Conventional Commits specification.

I'm not sure if fixing rendering to be correct would be considered a breaking change. It's possible that someone has written code to work around this issue. If so, the result of this change would be that their application would most likely end up having empty space at the right side and bottom of the window rather than filling it. That is what would have happened to mine if I had decided to implement a workaround rather than create this fix.

The change won't cause code that compiles to fail to compile. It's possible that for some users the visual result would be as I described. It's highly unlikely that it would cause a program to crash at runtime unless someone wrote a work around and failed to write adequate error checking as part of it.

As such, I marked that this has no breaking changes. The migration path, if it is breaking, would be for affected users to remove their custom code that currently does rendering adjustment for Gtk.WSL, Gtk.Linux, and any other Gtk + Wayland/etc. combination and simply use the code they use for all other platforms they support since the behavior will now conform to the current behavior of all other platforms.

## Other information

The UI test is in the SampleApp at: Window -> Window_ContentIsFullyVisible.

When testing, note that unlike with this fix where there is some padding between the edge of the window and the Border of the StackPanel, UWP renders the border right against the edge of the window. However, Skia.Gtk WSL, Skia.Gtk Gtk for Windows, Skia.Wpf, WASM, and Android, which are the platforms I tested this on, also have some amount of padding between the edge of their window/screen and the border. So while the behavior is different than UWP, it's different in the same way that all of those other platforms are different, and thus is unrelated to this bug fix.

### DIFF
--- a/src/SamplesApp/UITests.Shared/UITests.Shared.projitems
+++ b/src/SamplesApp/UITests.Shared/UITests.Shared.projitems
@@ -1001,6 +1001,10 @@
       <SubType>Designer</SubType>
       <Generator>MSBuild:Compile</Generator>
     </Page>
+    <Page Include="$(MSBuildThisFileDirectory)Windows_UI_Xaml\WindowTests\Window_ContentIsFullyVisible.xaml">
+      <SubType>Designer</SubType>
+      <Generator>MSBuild:Compile</Generator>
+    </Page>
     <Page Include="$(MSBuildThisFileDirectory)Windows_UI_Xaml\xBindTests\NoPhaseBinding_Large.xaml">
       <SubType>Designer</SubType>
       <Generator>MSBuild:Compile</Generator>
@@ -5085,6 +5089,9 @@
     </Compile>
     <Compile Include="$(MSBuildThisFileDirectory)Windows_UI_Xaml\Clipping\XamlButtonWithClipping_Scrollable.xaml.cs">
       <DependentUpon>XamlButtonWithClipping_Scrollable.xaml</DependentUpon>
+    </Compile>
+    <Compile Include="$(MSBuildThisFileDirectory)Windows_UI_Xaml\WindowTests\Window_ContentIsFullyVisible.xaml.cs">
+      <DependentUpon>Window_ContentIsFullyVisible.xaml</DependentUpon>
     </Compile>
     <Compile Include="$(MSBuildThisFileDirectory)Windows_UI_Xaml\xBindTests\NoPhaseBinding_Large.xaml.cs">
       <DependentUpon>NoPhaseBinding_Large.xaml</DependentUpon>
--- a/src/SamplesApp/UITests.Shared/Windows_UI_Xaml/WindowTests/Window_ContentIsFullyVisible.xaml
+++ b/src/SamplesApp/UITests.Shared/Windows_UI_Xaml/WindowTests/Window_ContentIsFullyVisible.xaml
@@ -1,0 +1,29 @@
+﻿<UserControl
+	x:Class="UITests.Windows_UI_Xaml.WindowTests.Window_ContentIsFullyVisible"
+	xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+	xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+	xmlns:local="using:UITests.Windows_UI_Xaml.WindowTests"
+	xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+	xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+	mc:Ignorable="d"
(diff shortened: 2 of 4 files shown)


jeromelaban:
Thanks for the PR! Conventional commit does not allow for dots between the parentheses, could you use a `_` instead?

The change looks good though, we'll test with the PR's artifacts.

mikebmcl:
I fixed the Conventional Commits parentheses problem. Thanks!

jeromelaban:
Thanks for the update. It's actually the commit's message content that causes this issue (not the PR title). Can you do the same modification there as well ?

mikebmcl:
I edited the commit message to remove all the periods inside parentheses, including removing links that had parentheses in them since I wasn't sure if those were problematic.

I also slightly modified the message to indicate that this may be specific to Wayland because Wayland prefers to leave rendering of the title bar and other window decoration to the client and [the Gtk Wayland backend](https://wiki.gnome.org/Initiatives/Wayland/GTK%2B) does that. So that may be the reason this affects WSL, which uses Wayland, but not Gtk for Windows. It would also affect Linux distros that are configured to use Wayland. I'm unsure if it would be a problem when using X11.

jeromelaban:
I've updated the history for your PR to reflect the change, it should be good to go now.

mikebmcl:
Seems like it didn't like something int the scope (probably the _ ), so I just switched from "fix(Skia_Gtk): ..." to "fix: [GTK] ..." since past PRs have used no scope with the platform in square brackets as part of their message.

jeromelaban (approved):
This is looking great!

Thanks for your contribution!

PR #7735, opened 2022-01-03
fix(dragdrop): Reordering item out of LV items bounds

dr1rrb:
closes https://github.com/unoplatform/nventive-private/issues/329

## Bugfix
Reordering items before or after the bounds of items on the LV has no effect.

* When we leave the bounds of a drop target, we don't reset the `dragUIOverride` object, so the accepted operation (i.e. `None`) is not updated in the `DragView`.
* In a `ListView` when reordering items after the bounds of the last item, the accepted operation remains to `Move` (as expected)  but drop has no effect.

## What is the new behavior?
* On pointer move, if we are over a new drop target, or if we are not able to find a drop target, we clear the `dragUIOvwerride`
* If we are not able to find an item under the pointer when reordering items of a `ListView`, we are selecting the last one.

## PR Checklist

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/ListViewBase/ListViewBase.DragDrop.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/ListViewBase/ListViewBase.DragDrop.cs
@@ -184,10 +184,7 @@ namespace Windows.UI.Xaml.Controls
 			}
 		}
 
-		private static void OnReorderDragUpdated(object sender, _DragEventArgs dragEventArgs) => OnReorderUpdated(sender, dragEventArgs, setVelocity: true);
-		private static void OnReorderDragLeave(object sender, _DragEventArgs dragEventArgs) => OnReorderUpdated(sender, dragEventArgs, setVelocity: false);
-
-		private static void OnReorderUpdated(object sender, _DragEventArgs dragEventArgs, bool setVelocity)
+		private static void OnReorderDragUpdated(object sender, _DragEventArgs dragEventArgs)
 		{
 			var that = sender as ListView;
 			var src = dragEventArgs.DataView.FindRawData(ReorderOwnerFormatId) as ListView;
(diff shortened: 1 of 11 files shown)


src/Uno.UI/UI/Xaml/Controls/ListViewBase/ListViewBase.DragDrop.cs
```diff
@@ -213,17 +210,25 @@ private static void OnReorderUpdated(object sender, _DragEventArgs dragEventArgs
...
+		}
+
+		private static void OnReorderDragLeave(object sender, _DragEventArgs dragEventArgs)
+		{
+			var that = sender as ListView;
+			var src = dragEventArgs.DataView.FindRawData(ReorderOwnerFormatId) as ListView;
+			if (that is null || src != that)
 			{
-				that.SetPendingAutoPanVelocity(PanVelocity.Stationary);
+				dragEventArgs.Log().Warn("Invalid reorder event.");
```
jeromelaban, line 225: use if(this.Log().IsEnabled...)

carldebilly (approved):

jeromelaban (approved):

src/Uno.UI/UI/Xaml/Controls/ListViewBase/VirtualizingPanelLayout.iOS.cs
```diff
@@ -1506,13 +1511,25 @@ private void UpdateLayoutForReordering()
 			if (_reorderingState is { } reorderingState && reorderingState.LayoutAttributes is { } draggedAttributes)
 			{
 				var dropTargetAttributes = GetLayoutAttributesUnderPoint(reorderingState.Location);
+				if (dropTargetAttributes is null)
+				{
+					// If we have a reorderingState, it means that the pointer is still over the LV.
```
davidjohnoliver, line 1516: Could it also be before the first item?

davidjohnoliver (commented):
This is already handled properly on Android and WASM?

src/Uno.UI/UI/Xaml/Controls/ListViewBase/VirtualizingPanelLayout.iOS.cs
dr1rrb: Do we support to vertically center items in a ListView?

dr1rrb:
> This is already handled properly on Android and WASM?

From my tests, yes. It was only on iOS

src/Uno.UI/UI/Xaml/Controls/ListViewBase/VirtualizingPanelLayout.iOS.cs
davidjohnoliver: not to my knowledge, but what if, eg, there's a Padding set on the ListView or its ItemsPresenter?

dr1rrb: Good point ... fixed!

src/Uno.UI/UI/Xaml/Controls/ListViewBase/VirtualizingPanelLayout.iOS.cs
```diff
@@ -1578,20 +1583,33 @@ private void ResetReorderedLayoutAttributes()
...
+
 			foreach (var dict in _itemLayoutInfos.Values)
 			{
 				foreach (var layoutAttributes in dict.Values)
 				{
-					if (DoesLayoutAttributesContainDraggedPoint(point, layoutAttributes))
+					var distance = ((Rect)layoutAttributes.Frame).GetDistance(adjustedPoint);
+					if (distance == 0)
 					{
+						// Fats path: we found the element that is under the element
```
davidjohnoliver, line 1600:
```suggestion
						// Fast path: we found the element that is under the element
```

davidjohnoliver (approved):

PR #7736, opened 2022-01-04
fix: `AutoSuggestBox` should keep focus when popup is opened

### LINKED ISSUE #7675: [Skia.Wpf - Skia.Gtk] AutoSuggestBox doesn't work right
### Current behavior

I have an `AutoSuggestBox`, which contains several objects in its ItemsSource. 
When I click it, both on Wpf and on Gtk, the text box gets focused and I can start typing. After typing the first character, the dropdown/suggest popup shows up. Now the textbox loses focus and I cannot continue typing.

[AutoSuggestBoxRepro.zip](https://github.com/unoplatform/uno/files/7740598/AutoSuggestBoxRepro.zip)

### Expected behavior

I expect to be able to continue typing. 

_No response_

### Workaround

_No response_

### Works on UWP/WinUI

Yes

### Environment

_No response_

### NuGet package version(s)

3.11.6, 
but can also reproduce on 4.1.0-dev.152

(Also on WinUI)

### Affected platforms

### IDE

Visual Studio 2019

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

MartinZikmund:
GitHub Issue (If applicable): closes #7675

## PR Type

- Bugfix

`AutoSuggestBox` loses focus when its popup opens because `IsLightDismissEnabled` defaults to `true` on Uno instead of `false` as in UWP (and `IsLightDismissEnabled` popups may capture focus when opened)

## What is the new behavior?

- `Popup.IsLightDismissEnabled` is now `false` by default to match UWP
- Added `FeatureConfiguration.Popup.EnableLightDismissByDefault` that allows turning the legacy behavior back on where needed
- Added tests to cover the default, and the text change behavior of `AutoSuggestBox`

## PR Checklist

### DIFF
--- a/src/Uno.UI/FeatureConfiguration.cs
+++ b/src/Uno.UI/FeatureConfiguration.cs
@@ -275,9 +275,9 @@ namespace Uno.UI
 
 			/// <summary>
 			/// By default, light dismiss is disabled in UWP/WinUI unless
-			/// IsLightDismissEnabled is explicitly set to true.
-			/// In earlier versions of Uno Platform this property defaulted
-			/// to true, which was incorrect behavior. If your code depends on this
+			/// <see cref="Windows.UI.Xaml.Controls.Primitives.Popup.IsLightDismissEnabled"/> is explicitly set to true.
+			/// In earlier versions of Uno Platform, this property defaulted
+			/// to true, which was an incorrect behavior. If your code depends on this
 			/// legacy behavior, use this property to override it.
 			/// </summary>
 			public static bool EnableLightDismissByDefault { get; set; } = false;
--- a/src/Uno.UI/FeatureConfiguration.cs
+++ b/src/Uno.UI/FeatureConfiguration.cs
@@ -272,6 +272,15 @@ namespace Uno.UI
 			/// </summary>
 			public static bool UseNativePopup { get; set; } = true;
 #endif
+
+			/// <summary>
+			/// By default, light dismiss is disabled in UWP/WinUI unless
+			/// IsLightDismissEnabled is explicitly set to true.
+			/// In earlier versions of Uno Platform this property defaulted
+			/// to true, which was incorrect behavior. If your code depends on this
+			/// legacy behavior, use this property to override it.
+			/// </summary>
+			public static bool EnableLightDismissByDefault { get; set; } = false;
 		}
 
 		public static class ProgressRing
--- a/src/Uno.UI/UI/Xaml/Controls/Popup/Popup.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Popup/Popup.cs
@@ -76,13 +76,12 @@ namespace Windows.UI.Xaml.Controls.Primitives
 
 		private bool GetDefaultValue(DependencyProperty dependencyProperty, out object value)
 		{
-			value = null;			
-
 			if (dependencyProperty == IsLightDismissEnabledProperty)
 			{
-				value = FeatureConfiguration.Popup.EnableLightDismissByDefault ? true : false;
+				value = FeatureConfiguration.Popup.EnableLightDismissByDefault;
 				return true;
 			}
+			value = null;			
 			return false;
 		}
 
--- a/src/Uno.UI/UI/Xaml/Controls/Popup/Popup.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Popup/Popup.cs
@@ -70,7 +70,20 @@ namespace Windows.UI.Xaml.Controls.Primitives
 
 		public Popup()
 		{
+			this.RegisterDefaultValueProvider(GetDefaultValue);
 			Initialize();
+		}
+
+		private bool GetDefaultValue(DependencyProperty dependencyProperty, out object value)
+		{
+			value = null;			
+
+			if (dependencyProperty == IsLightDismissEnabledProperty)
+			{
+				value = FeatureConfiguration.Popup.EnableLightDismissByDefault ? true : false;
+				return true;
+			}
+			return false;
 		}
 
 		private void Initialize()
--- a/src/Uno.UI/UI/Xaml/Controls/Popup/Popup.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Popup/Popup.cs
@@ -70,19 +70,17 @@ namespace Windows.UI.Xaml.Controls.Primitives
 
 		public Popup()
 		{
-			this.RegisterDefaultValueProvider(GetDefaultValue);
 			Initialize();
 		}
 
-		private bool GetDefaultValue(DependencyProperty dependencyProperty, out object value)
+		internal override bool GetDefaultValue2(DependencyProperty property, out object defaultValue)
 		{
-			if (dependencyProperty == IsLightDismissEnabledProperty)
+			if (property == IsLightDismissEnabledProperty)
 			{
-				value = FeatureConfiguration.Popup.EnableLightDismissByDefault;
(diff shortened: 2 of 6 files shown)


src/Uno.UI/FeatureConfiguration.cs
```diff
@@ -265,6 +265,15 @@ public static class Popup
 			/// </summary>
 			public static bool UseNativePopup { get; set; } = true;
 #endif
+
+			/// <summary>
+			/// By default, light dismiss is disabled in UWP/WinUI unless
+			/// IsLightDismissEnabled is explicitly set to true.
+			/// In earlier versions of Uno Platform this property defaulted
```
Youssef1313, line 272: Minor grammar correction

```suggestion
			/// In earlier versions of Uno Platform, this property defaulted
```

src/Uno.UI/FeatureConfiguration.cs
```diff
@@ -265,6 +265,15 @@ public static class Popup
...
+			/// to true, which was incorrect behavior. If your code depends on this
+			/// legacy behavior, use this property to override it.
+			/// </summary>
+			public static bool EnableLightDismissByDefault { get; set; } = false;
```
Youssef1313, line 276: Consider adding a `<remarks>` section indicating that AutoSuggestBox won't work correctly if the property is overridden to true.

src/Uno.UI/Mixins/DependencyPropertyMixins.tt
```diff
@@ -88,7 +88,7 @@
 			.Class("Popup")
 				.Property("IsOpen", "bool", "false")
 				.Property("Child", "UIElement", "null", frameworkPropertyOption: "ValueInheritsDataContext")
-				.Property("IsLightDismissEnabled", "bool", "true")
+				.Property("IsLightDismissEnabled", "bool", "false")
```
Youssef1313, line 91: Why is this false and not depending on `FeatureConfiguration.EnableLightDismissByDefault`?

If a change will be made here, consider having a test for it.

src/Uno.UI/UI/Xaml/Controls/Popup/Popup.cs
```diff
@@ -58,9 +58,22 @@ private bool ShouldShowLightDismissOverlay
...
 			Initialize();
 		}
 
+		private bool GetDefaultValue(DependencyProperty dependencyProperty, out object value)
+		{
+			value = null;			
+
+			if (dependencyProperty == IsLightDismissEnabledProperty)
+			{
+				value = FeatureConfiguration.Popup.EnableLightDismissByDefault ? true : false;
```
Youssef1313, line 71: Would this work the same?

```suggestion
				value = FeatureConfiguration.Popup.EnableLightDismissByDefault;
```

src/Uno.UI/UI/Xaml/Controls/Popup/Popup.cs
```diff
@@ -58,9 +58,22 @@ private bool ShouldShowLightDismissOverlay
 
 		public Popup()
 		{
+			this.RegisterDefaultValueProvider(GetDefaultValue);
 			Initialize();
 		}
 
+		private bool GetDefaultValue(DependencyProperty dependencyProperty, out object value)
+		{
+			value = null;
```
Youssef1313, line 67: Consider moving this line just before the `return false;`

src/Uno.UI/Mixins/DependencyPropertyMixins.tt
Youssef1313: After a second thought, wouldn't this value always be ignored because of the `RegisterDefaultValueProvider` call? Or I'm misunderstanding `RegisterDefaultValueProvider` behavior?

src/Uno.UI/FeatureConfiguration.cs
```diff
@@ -265,6 +265,15 @@ public static class Popup
 			/// </summary>
 			public static bool UseNativePopup { get; set; } = true;
 #endif
+
+			/// <summary>
+			/// By default, light dismiss is disabled in UWP/WinUI unless
+			/// IsLightDismissEnabled is explicitly set to true.
```
Youssef1313, line 271: Consider using `<see cref="..." />`

src/Uno.UI/Mixins/DependencyPropertyMixins.tt
MartinZikmund: Yes, this value is overridden by the default value provider currently. I suppose we will be able to remove the legacy behavior altogether in the future. I used the default value provider as the `FeatureConfiguration` could be changed at any point (even runtime), so this way the default will always reflect the current setting at the point of `Popup` creation

src/Uno.UI/UI/Xaml/Controls/Popup/Popup.cs
MartinZikmund: Lol, of course, I'm stupid 😂😂😂 !

Youssef1313: @MartinZikmund FYI there are cases where the `? true : false` is desired due to a runtime bug (performance issue) (https://github.com/dotnet/runtime/issues/37904) but that seems to happen only when generics is involved.

src/Uno.UI/UI/Xaml/Controls/Popup/Popup.cs
```diff
@@ -58,9 +58,22 @@ private bool ShouldShowLightDismissOverlay
 
 		public Popup()
 		{
+			this.RegisterDefaultValueProvider(GetDefaultValue);
 			Initialize();
 		}
 
+		private bool GetDefaultValue(DependencyProperty dependencyProperty, out object value)
```
Youssef1313, line 65: Unrelated to your changes: Was there any reason to use `object` instead of `bool?` for the out parameter? Or more simply, eliminate the out parameter and return `bool?`

This doesn't seem to be publicly exposed, so it could be changed without any breaking changes.

src/Uno.UI/FeatureConfiguration.cs
```diff
@@ -265,6 +265,15 @@ public static class Popup
 			/// </summary>
 			public static bool UseNativePopup { get; set; } = true;
 #endif
+
+			/// <summary>
+			/// By default, light dismiss is disabled in UWP/WinUI unless
+			/// <see cref="Windows.UI.Xaml.Controls.Primitives.Popup.IsLightDismissEnabled"/> is explicitly set to true.
+			/// In earlier versions of Uno Platform, this property defaulted
+			/// to true, which was incorrect behavior. If your code depends on this
```
jeromelaban, line 273:
```suggestion
			/// to true, which was an incorrect behavior. If your code depends on this
```

src/Uno.UI/FeatureConfiguration.cs
MartinZikmund: Yes, that's the case 👍

src/Uno.UI/UI/Xaml/Controls/Popup/Popup.cs
MartinZikmund: The `GetDefaultValue` matches the `DefaultValueProvider` delegate and is called for all dependency properties - so it has to be `object` to accommodate for any type of DP. I think for example `Control` class has some provider as well, which handles focus-related properties, etc.

jeromelaban (approved):

ajpinedam (approved):

davidjohnoliver (approved):

PR #7752, opened 2022-01-06
fix: `IsTabStop` everywhere and `AllowFocusOnInteraction` on WASM

### LINKED ISSUE #2115: IsTabStop on TextBox not respected on Android
## Current behavior

Setting `IsTabStop="False"` on a text box will prevent the text box from receiving focus on uwp, however on android the textbox still gets focus and will pop up the keyboard when the page loads.
This behaviour occurs whether the page is loaded directly or from within a frame.

According to [the documentation](https://platform.uno/docs/articles/implemented/windows-ui-xaml-controls-control.html), `Control.IsTabStop` is supported on all platforms.

## Expected behavior

The textbox should behave the same on both platforms, ideally following the existing uwp behaviour where it cannot receive focus.

[NavTestTwo.zip](https://github.com/unoplatform/uno/files/3860715/NavTestTwo.zip)
Run the project, notice on uwp the textbox cannot receive focus, while on android the textbox receives focus automatically and brings up the keyboard.
If one removes the text box from the MainPage then the text box on the InnerPage will receive focus by default.

## Environment

Nuget Package:

Uno.Ui - v1.45.0
Uno.Core - v1.28.0

Affected platform(s):

- [ ] iOS
- [x] Android
- [ ] WebAssembly
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

Visual Studio:

- [ ] 2017 (version: )
- [x] 2019 (version: 16.3.8)
- [ ] for Mac (version: )

Also, setting `IsTabStop="True"` on a different control to try and stop the keyboard from showing automatically doesn't work, the text box still receives focus first (on Android).

### LINKED ISSUE #6471: WASM native focus causes problems with `AllowFocusOnInteraction`
## Current behavior

As an example when user taps on a button, we get a native event from the browser originating in the TextBlock inside the button template - and, the actual control that should get focused is the `Button` . However, if the Button has `AllowFocusOnInteraction = false` or `IsTabStop = false`, then nothing should get focused at all. Our current implementation bubbles up to find first focusable control, which does not work properly, as it may focus something way higher above the button (like the root of the Page for example). However, if we stop bubbling at first `IsTabStop = false` or `AllowFocusOnInteraction = false` it breaks controls like `TextBox` which contains a `ScrollViewer` inside, which is also has `IsTabStop = false`. 

## Expected behavior

Should behave like UWP/WinUI.

See above.

## Works on UWP/WinUI

Yes

## Environment

Nuget Package:

- [x] Uno.WinUI / Uno.WinUI.WebAssembly / Uno.WinUI.Skia
- [ ] Other: 

Nuget Package Version(s): latest dev

Affected platform(s):

- [ ] iOS
- [ ] Android
- [x] WebAssembly
- [ ] macOS
- [ ] Skia
  - [ ] WPF
  - [ ] GTK (Linux)
  - [ ] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

MartinZikmund:
GitHub Issue (If applicable): closes #2115, closes #6471

## PR Type

- Bugfix

- `IsTabStop` had weird behavior for input fields on most platforms
- `AllowFocusOnInteraction` did not work reliably on WASM

## What is the new behavior?

- `IsTabStop` is properly respected by input fields
- `AllowFocusOnInteraction = false` properly stops propagation of native focus on WASM

## PR Checklist

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.wasm.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.wasm.cs
@@ -8,7 +8,7 @@ namespace Windows.UI.Xaml.Controls
 	public partial class TextBox : Control
 	{
 		private TextBoxView _textBoxView;
-		
+
 		protected override bool IsDelegatingFocusToTemplateChild() => true; // _textBoxView
 		partial void OnDeleteButtonClickPartial() => FocusTextView();
 		internal bool FocusTextView() => FocusManager.FocusNative(_textBoxView);
@@ -79,7 +79,7 @@ namespace Windows.UI.Xaml.Controls
 
 		partial void UpdateFontPartial()
 		{
-			if(_textBoxView == null)
+			if (_textBoxView == null)
 			{
 				return;
 			}
@@ -112,12 +112,14 @@ namespace Windows.UI.Xaml.Controls
 			ApplyEnabled(e.NewValue);
 		}
 
-		partial void OnIsReadonlyChangedPartial(DependencyPropertyChangedEventArgs e)
+		partial void OnIsReadonlyChangedPartial(DependencyPropertyChangedEventArgs e) => UpdateTextBoxViewIsReadOnly();
+
+		partial void OnIsTabStopChangedPartial() => UpdateTextBoxViewIsReadOnly();
+
+		private void UpdateTextBoxViewIsReadOnly()
 		{
-			if (e.NewValue is bool isReadonly)
-			{
-				ApplyIsReadonly(isReadonly);
-			}
+			var isReadOnly = IsReadOnly || !IsTabStop;
+			ApplyIsReadonly(isReadOnly);
 		}
 
 		partial void OnInputScopeChangedPartial(DependencyPropertyChangedEventArgs e)
--- a/src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.wasm.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.wasm.cs
@@ -118,8 +118,8 @@ namespace Windows.UI.Xaml.Controls
 
 		private void UpdateTextBoxViewIsReadOnly()
 		{
-			var isReadOnly = IsReadOnly || !IsTabStop;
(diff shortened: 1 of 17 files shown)


src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.wasm.cs
```diff
@@ -112,12 +112,14 @@ private protected override void OnIsEnabledChanged(IsEnabledChangedEventArgs e)
...
+
+		partial void OnIsTabStopChangedPartial() => UpdateTextBoxViewIsReadOnly();
+
+		private void UpdateTextBoxViewIsReadOnly()
 		{
-			if (e.NewValue is bool isReadonly)
-			{
-				ApplyIsReadonly(isReadonly);
-			}
+			var isReadOnly = IsReadOnly || !IsTabStop;
```
jeromelaban, line 121: Does not having tabstop really means that the textbox is readonly? This feels like the naming should be different for the function.

MartinZikmund: The difference here should be that when `IsReadOnly = true`, the control should be read-only, but the managed one (e.g. the `TextBox` itself) should still be focusable. In case of `IsTabStop = false` it is not focusable anymore. In terms of the behavior of the `TextBoxView` however (in this case `<input />`) - in both cases we want to make it `readonly` in HTML. So I think the naming in this case should be ok

davidjohnoliver: maybe `isNativeReadOnly`/`ApplyNativeReadOnly`?

and you should also put the explanation from your reply as a source comment ;)

MartinZikmund: Realized the `ApplyIsReadonly` method itself is unnecessary indirection, so moved the code directly to the `UpdateTextBoxViewReadonly` method and adjusted the naming

jeromelaban (approved):
Can you add documentation about this new feature flag?

davidjohnoliver (approved):

PR #7840, opened 2022-01-17
fix(Skia): [WPF] Issue #7829 SynchronizationContext

### LINKED ISSUE #7829: [Skia][WPF] RenderSize is `[0,0]` when generating the control in the event handler
### Current behavior

Considering a code like this:

```csharp
        private void Button_Click(object sender, RoutedEventArgs e)
        {
            Do();
        }

        async void Do()
        {
            var border = new Border()
            {
                Background = new SolidColorBrush(Windows.UI.Colors.Red),
            };
            GeneratedContent.Children.Add(border);
            await Task.Yield();
            Console.WriteLine(border.RenderSize);
        }
```
If you run the Skia.WPF project, the line: `Console.WriteLine (border.RenderSize);` returns 0,0 in UWP works as expected.

### Expected behavior

Give RenderSize

- Download [Dispatcher_Issue.zip](https://github.com/unoplatform/uno/files/7874791/Dispatcher_Issue.zip)
- Add breckpoint at line `Console.WriteLine(border.RenderSize);'
- Run in UWP Project
- Click on Generate
- You can see correctly RenderSize
- Run Skia WPF
- Click on Generate
- RenderSize is 0,0

### Workaround

Add the following code at the beginning of the method how to generate the UI:

```csharp
var ctx = new Uno.UI.Dispatching.CoreDispatcherSynchronizationContext(Uno.UI.Dispatching.CoreDispatcher.Main
				, Uno.UI.Dispatching.CoreDispatcherPriority.Normal);
System.Threading.SynchronizationContext.SetSynchronizationContext(ctx);
```

### Works on UWP/WinUI

Yes

### Environment

### NuGet package version(s)

Tested on `Uno.UI 4.0.11` abd `Uno.UI 4.1.0-dev.324`

### Affected platforms

Skia (WPF)

### IDE

Visual Studio 2022

### IDE version

17.1.0 Preview 1.1

### Relevant plugins

_No response_

WPF replace CoreDispatcherSynchronizationContext with own DispatcherSynchronizationContext

workgroupengineering:
GitHub Issue (If applicable): closes #7829

## PR Type

- Bugfix

## PR Checklist

## Other information

### DIFF
--- a/build/ci/.azure-devops-skia-tests.yml
+++ b/build/ci/.azure-devops-skia-tests.yml
@@ -31,6 +31,7 @@ jobs:
     displayName: Build GTK Head
 
   - task: PublishBuildArtifacts@1
+    displayName: Publish GTK SamplesApp
     condition: always()
     inputs:
       PathtoPublish: $(Build.SourcesDirectory)\src\SamplesApp\SamplesApp.Skia.Gtk\bin\Release\netcoreapp3.1
@@ -41,10 +42,22 @@ jobs:
     displayName: Build WPF Head
 
   - task: PublishBuildArtifacts@1
+    displayName: Publish WPF SamplesApp 
     condition: always()
     inputs:
       PathtoPublish: $(Build.SourcesDirectory)\src\SamplesApp\SamplesApp.Skia.WPF\bin\Release\netcoreapp3.1
       ArtifactName: skia-wpf-samples-app
+      ArtifactType: Container
+
+  - powershell: dotnet msbuild src/Uno.UI.RuntimeTests.Skia.WPF/Uno.UI.RuntimeTests.Skia.WPF.csproj /r /m /p:Configuration=Release /detailedsummary /m # /bl:$(build.artifactstagingdirectory)\build.binlog
+    displayName: Build WPF Runtime Tests
+
+  - task: PublishBuildArtifacts@1
+    displayName: Publish WPF Runtime Tests 
+    condition: always()
+    inputs:
+      PathtoPublish: $(Build.SourcesDirectory)\src\Uno.UI.RuntimeTests.Skia.WPF\bin\Release\
+      ArtifactName: skia-wpf-runtime-tests
       ArtifactType: Container
 
 ##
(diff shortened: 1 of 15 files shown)


workgroupengineering:
Can anyone review this PR?

jeromelaban (commented):
I've added this test, which matches what you've mentioned:
image

But it does not seem to ~~reproduce~~ fix the issue, can you validate on your end?

workgroupengineering:
> I've added this test, which matches what you've mentioned: image
> 
> But it does not seem to ~reproduce~ fix the issue, can you validate on your end?

If run in test it works as expected but if run in [Real Test APP](https://github.com/unoplatform/uno/files/7874791/Dispatcher_Issue.zip) the issue is present.  I have tryed with 4.1.0-dev.461.

jeromelaban:
It would be important to understand what's different in the sample app, it's supposed to behavior the same way. Can you add the test in the runtime tests and see what results you're getting ?

workgroupengineering:
> It would be important to understand what's different in the sample app, it's supposed to behavior the same way. Can you add the test in the runtime tests and see what results you're getting ?

I try.

How can I do to indicate that the test must be performed only in wpf?

I need to create a real WPF window and call event a via the window API.

jeromelaban:
The test should theoretically work on all platforms, but if we need to limit it to Skia, we'll be able to. Can you add it and launch a build? We'll see how far it goes :)

MartinZikmund:
@jeromelaban Is there some way to the test could avoid requiring separate runtime test project for WPF specifically 🤔 ? (it does not seem so based on the fact the test needs to access `WpfHost`, but it would be nicer to be able to avoid the additional project.

jeromelaban:
> @jeromelaban Is there some way to the test could avoid requiring separate runtime test project for WPF specifically 🤔 ? (it does not seem so based on the fact the test needs to access `WpfHost`, but it would be nicer to be able to avoid the additional project.

I don't think it's necessary at this point. The test that requires a window to validate, this feels like a too low-level replacement for a UI test engine that would do this in a generic way. 

The solution to override the synchronization context for all event going out seems a bit too wide as well, particularly if it's about events going out of Uno's APIs, for which the entry point is known. Let's change those entry points for now to set the proper synchronization context. We can create a test that uses the new pointer injection to validate that we get the proper synchronization context out inside of uno pointer or generic input handlers.

Issue #7835, opened 2022-01-16
[WPF]Can't get the clicked element like as UWP


ywmoyue:
### Current behavior

When Tapped or RightTapped is used, TappedRoutedEventArgs.OriginalSource cannot be obtained correctly
![unotest](https://user-images.githubusercontent.com/28388300/149647852-6925cdb0-cf2b-441b-8472-dd21e9147b6b.gif)
.

### Expected behavior

I can get the OriginalSource in UWP.
![uwptest](https://user-images.githubusercontent.com/28388300/149647903-c8439558-903f-486e-a456-1f62f0bc0c30.gif)

### How to reproduce it (as minimally and precisely as possible)

MainPage.xaml :
```xaml
<Page
    x:Class="TestOriginalSource.MainPage"
    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
    xmlns:local="using:TestOriginalSource"
    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
    mc:Ignorable="d">

    <Grid Tapped="Grid_Tapped">
        <Grid.RowDefinitions>
            <RowDefinition/>
            <RowDefinition/>
        </Grid.RowDefinitions>
        <TextBlock x:Name="text" Text="Hello, world!" Margin="20" FontSize="30" />
        <Grid Grid.Row="1" Background="Red"></Grid>
    </Grid>
</Page>
```

MainPage.xaml.cs :
```cs
    public sealed partial class MainPage : Page
    {
        public MainPage()
        {
            this.InitializeComponent();
        }
        private void Grid_Tapped(object sender, TappedRoutedEventArgs e)
        {
            var element = (FrameworkElement)(e.OriginalSource);
            text.Text = element.GetType().Name;
        }
    }
```

### Works on UWP/WinUI

Yes

### NuGet package version(s)

Uno.UI.Skia.Wpf  4.1.0-dev.324
Uno.UI.RemoteControl 4.1.0-dev.324

### Affected platforms

Skia (WPF)

### IDE

Visual Studio 2022, Visual Studio 2019

### IDE version

Microsoft Visual Studio Community 2019 16.11.2 Preview 1.0

Issue #7811, opened 2022-01-13
ScrollBar does not work on Skia


mikebmcl:
### Current behavior

Hovering over a `ScrollBar` (whether as part of a `ScrollViewer` or on its own) on Skia platforms does not cause `ScrollBar.OnPointerEnter` to be called, which means that its visual state never changes so the scroll bar itself never shows up (no rectangle, no thumb button, etc.).

(This is probably a bigger issue than just ScrollBar because it has to do with ControlTemplates and bubbling. I just happened to find it on ScrollBar. It may also affect some controls that are on platforms other than Skia. More about that in the "anything else" section below.)

### Expected behavior

Moving the mouse over the area where a scroll bar should be will cause the visual state transition that makes the scroll bar appear.

### How to reproduce it (as minimally and precisely as possible)

Run the [Uno SamplesApp](https://github.com/unoplatform/uno/tree/master/src/SamplesApp) on Skia (tested on Skia.Gtk and Skia.Wpf), then use a mouse to try to get the scrollbar on the left (where the list of tests is) to show you the scrollbar. Also, if you then use a mouse wheel to scroll through the samples down to the ScrollBar samples, run ScrollBar_Simple, then switch the indicator mode to `MouseIndicator` you can observe the problem with both horizontal and vertical ScrollBar controls.

### Workaround

None that are practical for most users.

If someone is compiling Uno locally and using those local builds, then the work around is to either undo the changes made in PR #7769 - specifically remove lines:
https://github.com/unoplatform/uno/blob/93cff21fb0d05ed8409f008e7ce904aeb0199232/src/Uno.UI/UI/Xaml/UIElement.Pointers.cs#L752 and https://github.com/unoplatform/uno/blob/93cff21fb0d05ed8409f008e7ce904aeb0199232/src/Uno.UI/UI/Xaml/UIElement.Pointers.cs#L766

or to put them inside a `#if !__SKIA__` guard. This may also affect MacOS but I'm unable to test that.

### Environment

Uno.UI / Uno.UI.WebAssembly / Uno.UI.Skia

### NuGet package version(s)

All packages since #7769 was merged (anything since January 7, 2022).

### Affected platforms

Skia (WPF), Skia (GTK on Linux/macOS/Windows)

### IDE

Visual Studio 2022

### IDE version

17.0.4

### Anything else we need to know?

[...]

Issue #7537, opened 2021-11-26
[Skia.Gtk] Pointer events inconsistencies for items contained in ListView


SNival:
### Current behavior

When touching items, the pointer events sequence is
- Pressed, Released then Tapped for an item on the main page
- Pressed then Tapped for items contained in ListView with SelectionMode=Single (missing Released)
- Pressed, Released then Tapped for items contained in ListView with SelectionMode=None

### Expected behavior

For every item, expected sequence is
- PointerPressed, PointerRelease then Tapped (if touched/released on same item)
- PointerPressed (touched item) then PointerReleased (released item) if released outside of the initialy touched item

### How to reproduce it (as minimally and precisely as possible)

[PointerEvents.zip](https://github.com/unoplatform/uno/files/7607857/PointerEvents.zip)
Run the project on a raspberry pi,
- tap in left list (Items A) : good sequence
- tap in right list (Items B): missing pointer released

Run the UWP version : expected behaviour

### Works on UWP/WinUI

Yes

### Environment

Uno.UI / Uno.UI.WebAssembly / Uno.UI.Skia

### Affected platforms

Skia (GTK on Linux/macOS/Windows)

### IDE

Visual Studio 2022, Visual Studio 2019

SNival:
For information, I also tried with version 4.0.9 with same results

PR #7845, opened 2022-01-17
Make sure enter/exit events are raised when needed

dr1rrb:
closes https://github.com/unoplatform/uno/issues/7811
closes https://github.com/unoplatform/uno/issues/7537
closes https://github.com/unoplatform/uno/issues/7835
closes https://github.com/unoplatform/uno/issues/7961

## Bugfix + Feature
1. Enter/exit are not raised on intermediate elements when crossing boundaries of multiple elements at once.
2. We are not able to test pointers on Skia and macOS

1. We are muting all enter/exit events when bubbling them
2. Not support at all

## What is the new behavior?
1. We are muting enter/exit events only if the event effectively updates the **local** state
2. The UWP input injection API is partially implemented and we are not able to multi target some existing UI tests to Skia and macOS

## PR Checklist


src/Uno.UI.RuntimeTests/UITests/_Engine/QueryExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+		=> new QueryEx(app, elementName).Execute();
+
+	public static QueryEx Marked(this SkiaApp app, string elementName)
+		=> new(app, elementName);
+
+	public static void FastTap(this QueryEx query)
+	{
+		var bounds = query.Execute().Single().Rect;
+
+		query.App.TapCoordinates((float)(bounds.X + bounds.Width / 2), (float)(bounds.Y + bounds.Height / 2));
```
carldebilly, line 23: Are you sure it will work all the time, considering some screen pixel density...

carldebilly (approved):

MartinZikmund (approved):

src/Uno.UI.RuntimeTests/UITests/_Engine/QueryExtensions.cs
dr1rrb: It should since we are dealing only with virtual pixels here, Anyway it's only a preliminary work, we will fix later if needed.

jeromelaban (approved):

mikebmcl:
I'm still working on the Skia testing related to #7888 however I noticed something important that can be seen when using the SampleApps Button->Button_Events sample (i.e. you don't need any of the samples I'm using from #7892 ).

Any event that is invoked due to motion, which in the case of Button_Events with its default ClickMode are the ones the UI lists as Pointer Entered and Pointer Leaved, are likely to be raised at times they shouldn't be based on UWP's behavior.

Whenever the pointer (mouse or touch) moves into a button, if it is supposed to raise an event, PointerEnter in this case, it will properly do so. Additionally, whenever the pointer leaves a button, if it is supposed to raise an event, PointerExited in this case, it will properly do so. 

However anytime the pointer is in the control and leaves from one part of the control to enter its container, e.g. from the Button's Content to the ContentPresenter that displays it (since there's Padding) or from the ContentPresenter to the Border that it has, the event will also be raised. Additionally, the first time the pointer enters one of those parts from its container it will not raise the event but after that, any time it enters that part (in addition to every time it leaves) the event will be raised.

The one exception is that if movement of the pointer is fast enough then only the Entered/Exited event that is supposed to be raised will be raised.

For ClickMode.Press and ClickMode.Release (which is the default) the problems only really affect the PointerEntered/Exited events since they're the only movement-based events for those. For ClickMode.Hover, it also affects the Click event because on UWP, Click for ClickMode.Hover is raised whenever the pointer moves into the button (it's identical to PointerEntered's behavior for Hover).

The old behavior was that PointerEntered would raise when the pointer exited the button in addition to when it entered. PointerExited would only raise when exiting. If the pointer moved fast enough it would never raise the events. The events wouldn't be raised when moving over any of the internal parts.

Additionally, if pointer entered into the HitTestVisible (Gold) Border, entered the Button's bounds while staying in that border, then exited from that into the button, it would never raise the events except for PointerEntered where it would raise when exiting the button outside of the Gold Border area. It should raise the events when entering/exiting the Gold Border so the new behavior is closer to correct.

For the ClickMode.Hover Click event, it would behave identically to PointerEntered.

There are some additional things specific to touch input but the new behavior is closer to UWP than the old behavior so it's better in that regard.

I'm just looking at Button right now. I don't know off-hand if there are any other controls that have events that care about movement into or out of them, which is the only significant behavior change I've found. This is all specific to Skia and I've only done Skia/Gtk testing so far. Unless there's another platform that also relies entirely on managed pointers, then this should be a Skia-only behavior change. I don't know if any of this is significant enough to matter and these new changes are pretty awesome. Just wanted to let you know about it in case there's some quick fix for this!

Issue #2116, opened 2019-11-18
WASM pointer events args are serialized/deserialized


dr1rrb:
## What would you like to be added:
Pointer events on WASM should be marshaled as struct instead of being serialized/deserialzied

## Why is this needed:
* For performance
* To be able properly propagate the isOver (needed for move while captured)

## For which Platform:
- [ ] iOS
- [ ] Android
- [x] WebAssembly
- [ ] WebAssembly renderers for Xamarin.Forms
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

## Anything else we need to know?
This is already almost done in this branch: https://github.com/unoplatform/uno/tree/dev/dr/PointerArgsMarshaling

dr1rrb:
As part of this work, we should also try to include this: https://github.com/unoplatform/uno/pull/2802/files#r391884707

PR #7886, opened 2022-01-21
feat(pointers): [PERF] [WASM] Marshal pointer events instead of serializing them

dr1rrb:
closes https://github.com/unoplatform/uno/issues/2116

## Performance
Marshal pointer event args between JS and C# instead of serializing / de-serializing them.

Each native pointer event is serialized into a string representation which is sent to the managed code.

## What is the new behavior?
On startup of the app, a memory space is shared to JS. Then on each native pointer event, we are writing into that space the pointer args info, then we notify the managed code which reads the value and raise the event in managed.

## PR Checklist


src/Uno.UI/UI/Xaml/UIElement.Pointers.wasm.cs
```diff
@@ -16,372 +24,413 @@
...
 		{
-			var parts = eventPayload?.Split(';');
-			if (parts?.Length != 13)
+			_pointerEventResult.Value = new NativePointerEventResult
 			{
-				args = default;
-				return false;
-			}
-
-			args = new NativePointerEventArgs {
```
jeromelaban, line 183: 🚀

jeromelaban (approved):

Issue #7900, opened 2022-01-25
Elements leaving Visual Tree should not be focusable


MartinZikmund:
### Current behavior

In case of Uno Platform, elements which are being unloaded can still get focused (even if briefly).

### Expected behavior

Elements which are being unloaded should not be focusable anymore.

To fix this, we will probably need to adjust the control lifecycle to match UWP/WinUI more closely.

### How to reproduce it (as minimally and precisely as possible)

See `[Ignored]` runtime tests `When_Page_Navigate_XYZ` in `Given_FocusManager.cs`.

### Works on UWP/WinUI

Yes

### Environment

Uno.UI / Uno.UI.WebAssembly / Uno.UI.Skia, Uno.WinUI / Uno.WinUI.WebAssembly / Uno.WinUI.Skia

### Affected platforms

iOS, Android, WebAssembly, WebAssembly renderers for Xamarin.Forms, macOS, Skia (WPF), Skia (GTK on Linux/macOS/Windows), Skia (Tizen)

### IDE

Visual Studio 2022

MartinZikmund:
Still relevant

MartinZikmund:
# Repro generator

This issue was classified as not reproducible with a code sample or test.

**Reason:** Focus management bug — elements leaving visual tree should not be focusable. Requires complex focus state tracking.

PR #7901, opened 2022-01-25
fix: Move focus from page which is navigated away in `Frame`

MartinZikmund:
GitHub Issue (If applicable): closes https://github.com/unoplatform/nventive-private/issues/313, partially #7900

## PR Type

- Bugfix

-  In some cases focus can stay on page which is being navigated from.
- When navigating, focus may move frantically from element to element as they are being unloaded.

## What is the new behavior?

When page is navigated away, focus will move away from it directly to some outer element or `null`. When the next page is navigated to, the focus will in case of `null` move to this new page.

## PR Checklist

### DIFF
--- a/src/Uno.UI.RuntimeTests/Uno.UI.RuntimeTests.csproj
+++ b/src/Uno.UI.RuntimeTests/Uno.UI.RuntimeTests.csproj
@@ -99,25 +99,6 @@
 	<ItemGroup>
 		<Compile Include="Tests\Windows_UI_Xaml_Controls\HtmlElementAttributeTests\Given_HtmlElementAttribute.Wasm.cs" />
 	</ItemGroup>
-	
-	
-	<ItemGroup>
-		<Page Update="Tests\Windows_UI_Xaml\Controls\Animation_Leak.xaml">
-		  <Generator>MSBuild:UpdateDesignTimeXaml</Generator>
-		</Page>
-		<Page Update="Tests\Windows_UI_Xaml\Controls\ContentDialog_Leak.xaml">
-		  <Generator>MSBuild:UpdateDesignTimeXaml</Generator>
-		</Page>
-		<Page Update="Tests\Windows_UI_Xaml\Controls\xLoad_xBind.xaml">
-			<SubType>Designer</SubType>
-		</Page>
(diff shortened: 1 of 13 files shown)


src/Uno.UI/UI/Xaml/UIElement.cs
```diff
@@ -107,6 +107,8 @@ internal virtual bool GetDefaultValue2(DependencyProperty property, out object d
 
 		internal Size AssignedActualSize { get; set; }
 
+		internal bool IsLeavingFrame { get; set; }
```
jeromelaban, line 110: Validate if you can use a flag instead, see with @carldebilly for compatibility.

jeromelaban (approved):

src/Uno.UI.RuntimeTests/Uno.UI.RuntimeTests.csproj
```diff
@@ -111,6 +111,12 @@
 		<Page Update="Tests\Windows_UI_Xaml\Controls\xLoad_xBind.xaml">
 			<SubType>Designer</SubType>
 		</Page>
+		<Page Update="Tests\Windows_UI_Xaml_Input\TestPages\TwoButtonSecondPage.xaml">
```
davidjohnoliver, line 114: I think you can revert these lines

davidjohnoliver (approved):

src/Uno.UI/UI/Xaml/UIElement.cs
MartinZikmund: @carldebilly can I merge this for now or should I wait for your changes first?

MartinZikmund: @carldebilly @jeromelaban seems I cannot easily use the facilities from Carl's PR, for two reasons:

- The `LayoutFlags` are netstd specific, but this property needs to apply to all targets
- It is rather a focus-related thing, not layout

I would say it would be easiest to merge this for now and maybe create some other flags enum for non-layout bools?

jeromelaban: Let's merge as is for now.

Issue #7632, opened 2021-12-11
[Android] Removing items from ComboBox confuses the SelectedIndex


queequac:
### Current behavior

I have a ComboBox control with several entries. Those entries are localized and set via XAML.
When I remove one of the entries, the SelectedIndex properties does not reflect the correct index anymore, but either returns -1 or the original index when the collection was created.
Also setting the selection via index is messed up afterwards.

### Expected behavior

The index reflects the correct position of the item within the Items collection.
Works correctly with Windows UWP or with Uno packages < version 4.0 (more precisely, it worked until 3.11.6)

### How to reproduce it (as minimally and precisely as possible)

Use the following XAML in a page:
```
    <StackPanel Orientation="Vertical">
        <TextBlock Text="{Binding ElementName=combo, Path=SelectedIndex}"/>
        <ComboBox x:Name="combo">
            <ComboBoxItem>One</ComboBoxItem>
            <ComboBoxItem>Two</ComboBoxItem>
            <ComboBoxItem>Three</ComboBoxItem>
        </ComboBox>
    </StackPanel>
```

And simply call in the page's code behind the following line:
`combo.Items.RemoveAt(0);`

### Workaround

Clear the collection (combo.Items.Clear()) and add items again, but that's not really fancy.😉

### Works on UWP/WinUI

Yes

### Environment

Uno.UI / Uno.UI.WebAssembly / Uno.UI.Skia

### NuGet package version(s)

4.0.9

### Affected platforms

Android

### IDE

Visual Studio 2019

### IDE version

16.11.7

carldebilly:
This should been fixed by https://github.com/unoplatform/uno/pull/8227

PR #8068, opened 2022-02-11
fix: Combobox remove selectedindex resolved index collection

CesarRabelo:
#7632 

## PR Type

- Bugfix

When I remove one of the entries, the SelectedIndex properties does not reflect the correct index anymore, but either returns -1 or the original index when the collection was created.

## What is the new behavior?

After remove item now the index is correct.

## PR Checklist

- [X] Validated PR `Screenshots Compare Test Run` results.

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/ItemsControl/ItemsControl.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/ItemsControl/ItemsControl.cs
@@ -61,6 +61,23 @@ namespace Windows.UI.Xaml.Controls
 		internal ScrollViewer ScrollViewer { get; private set; }
 
 		/// <summary>
+		/// Reindex ItemsControl.IndexForContainerProperty after the collection changes.		
+		/// </summary>
+		/// <param name="startingIndex">The minimum index of containers we care about.</param>
+		/// <param name="indexChange">How does the index change.</param>
+		private void SaveContainersAndRepairIndex(int startingIndex, int indexChange)
+		{
+			foreach (var container in MaterializedContainers)
+			{
+				var currentIndex = (int)container.GetValue(ItemsControl.IndexForItemContainerProperty);
+				if (currentIndex >= startingIndex)
+				{
+					container.SetValue(ItemsControl.IndexForItemContainerProperty, currentIndex + indexChange);
+				}
+			}
+		}
+
+		/// <summary>
 		/// This template is stored here in order to allow for 
 		/// FrameworkTemplate pooling to function properly when an ItemTemplateSelector has been
 		/// specified.
@@ -1034,6 +1051,7 @@ namespace Windows.UI.Xaml.Controls
 					LocalCleanupContainer(container);
 					ReassignIndexes(index);
 					RequestLayoutPartial();
+					SaveContainersAndRepairIndex(args.OldStartingIndex, -args.OldItems.Count);
 					return;
 				}
 				else if (args.Action == NotifyCollectionChangedAction.Add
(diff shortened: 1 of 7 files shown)


src/Uno.UI/UI/Xaml/Controls/ItemsControl/ItemsControl.cs
```diff
@@ -62,6 +62,32 @@ public partial class ItemsControl : Control, IItemsControl
 
 		internal ScrollViewer ScrollViewer { get; private set; }
 
+		/// <summary>
+		/// Stores materialized containers starting a given index, so that their
+		/// ItemsControl.IndexForContainerProperty can be updated after the collection changes.
```
davidjohnoliver, line 67: This description doesn't sound accurate, you're 'storing' anything. The method name is also no longer accurate since you're not saving the containers.

src/Uno.UI/UI/Xaml/Controls/ItemsControl/ItemsControl.cs
```diff
@@ -62,6 +62,32 @@ public partial class ItemsControl : Control, IItemsControl
...
 
+		/// <summary>
+		/// Stores materialized containers starting a given index, so that their
+		/// ItemsControl.IndexForContainerProperty can be updated after the collection changes.		
+		/// </summary>
+		/// <param name="startingIndex">The minimum index of containers we care about.</param>
+		/// <param name="indexChange">How does the index change.</param>
+		private void SaveContainersAndRepairIndex(int startingIndex, int indexChange)
+		{
+			Dictionary<DependencyObject, object> _containers = new Dictionary<DependencyObject, object>();
```
davidjohnoliver, line 73: why create this dictionary at all any more? Just set IndexForItemContainerProperty directly from the first foreach loop, you don't need two loops.

davidjohnoliver (commented):
A couple of general remarks that I'm going to split into two comments. 

The first is about commits:

1. Make sure to understand the [Conventional Commits format](https://platform.uno/docs/articles/uno-development/git-conventional-commits.html) and use the appropriate prefix. You should use 'fix' here, not 'feat', since you're fixing a bug, not adding a new feature. 
2. Consider squashing all your changes down to a single commit, to keep the history clean.

davidjohnoliver:
The second comment is that ideally the PR description should explain the reasoning behind the changes. It's good to see that you added unit tests, and the intent of the code is getting a little clearer to me in this PR relative to the previous one, now that you've refactored it, but it would be great to see some commentary to know what you were thinking when you added the code.

src/Uno.UI/UI/Xaml/Controls/ItemsControl/ItemsControl.cs
CesarRabelo: I forgot to review the summary

src/Uno.UI/UI/Xaml/Controls/ItemsControl/ItemsControl.cs
CesarRabelo: I forgot to review the summary

jeromelaban:
> > Thanks for the update. I've taken a look at the unit test, and it does not fail when the fix not applied. You'll probably need to take a look at the containers (using `comboBox.ContainerFromIndex(...)`) and validate that the bound data is the right one.
> > Also, make sure to squash your commits, as individual commits marked as `feature` will appear in the change log, and this PR is only making one specific fix.
> 
> Please @jeromelaban could you explain more what I need to do about (using `comboBox.ContainerFromIndex(...)`)?

You'll need to use `comboBox.ContainerFromIndex(...)` to get the actual containers and check the inner contents there. You can see similar examples here:
https://github.com/unoplatform/uno/blob/cfc5f060100ecc05e0be438487a23638178ab4e5/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_ListViewBase.cs#L1644

jeromelaban:
Thanks Cesar, I've validated your PR again, and the tests are passing even without the fix. The modified dependency property is used by `IndexFromContainer`, which is likely to be the part to be validated in both unit and runtime tests.

CesarRabelo:
Closed PR because issue was solved in another PR https://github.com/unoplatform/uno/issues/7632#issuecomment-1060819925

PR #8106, opened 2022-02-16
fix: TextWrapping="Wrap" does not seem to work for TextBox on Android

### LINKED ISSUE #3648: TextWrapping="Wrap" does not seem to work for TextBox on Android:
## Current behavior

TextWrapping="Wrap" does not seem to work for TextBox on Android:

## Expected behavior

TextWrapping="Wrap" should work for TextBox on Android:

```
<ScrollViewer Grid.Row="2">
            <TextBox  
                HorizontalAlignment="Center"
                TextWrapping="Wrap"
                MaxWidth="720"
                MinWidth="360"
                Padding="8"
                IsReadOnly="True"
                Text="{x:Bind sAbout}"
                />

        </ScrollViewer>
```

## Environment

Nuget Package: Uno.UI 

Package Version(s): 3.0.0-dev.1113

Affected platform(s): Android

- [ ] iOS
- [x] Android
- [ ] WebAssembly
- [ ] macOS
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

Visual Studio:

- [ ] 2017 (version: )
- [ ] 2019 (version: )
- [ ] for Mac (version: )

Relevant plugins:

- [ ] Resharper (version: )

aesteves900:
GitHub Issue (If applicable): closes #3648 

## PR Type

- Bugfix

When the user sets TextWrapping="Wrap" the text is never wrapped on the Android Platform.

## What is the new behavior?

When the user sets TextWrapping="Wrap" the text is wrapped.

## PR Checklist

## Other information

### DIFF
--- a/src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/TextBox/TextBox_Wrap.xaml
+++ b/src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/TextBox/TextBox_Wrap.xaml
@@ -1,0 +1,35 @@
+﻿<UserControl
+    x:Class="UITests.Windows_UI_Xaml_Controls.TextBox.TextBox_Wrap"
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:local="using:UITests.Windows_UI_Xaml_Controls.TextBox"
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    mc:Ignorable="d"
+    d:DesignHeight="300"
+    d:DesignWidth="400">
+
+    <Grid>
+
+		<Grid.RowDefinitions>
+			<RowDefinition Height="50"></RowDefinition>
+			<RowDefinition Height="300"></RowDefinition>
+			<RowDefinition Height="*"></RowDefinition>
+		</Grid.RowDefinitions>
+		<Button x:Name="buttonWrap" Grid.Row="0" Content="Wrap"  Click="OnWrapButtonClick"></Button>
+		<TextBox Grid.Row="1"
+            x:Name="textWrap"
+                HorizontalAlignment="Center"
+                TextWrapping="Wrap"
+                Padding="8"
+               Text="Sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium doloremque laudantium, totam rem aperiam, eaque ipsa quae ab illo inventore veritatis et quasi architecto beatae vitae dicta sunt explicabo. Nemo enim ipsam voluptatem quia voluptas sit aspernatur aut odit aut fugit, sed quia consequuntur magni dolores eos qui ratione voluptatem sequi nesciunt. Neque porro quisquam est, qui dolorem ipsum quia dolor sit amet, consectetur, adipisci velit, sed quia non numquam eius modi tempora incidunt ut labore et dolore magnam aliquam quaerat voluptatem. Ut enim ad minima veniam, quis nostrum exercitationem ullam corporis suscipit laboriosam, nisi ut aliquid ex ea commodi consequatur? Quis autem vel eum iure reprehenderit qui in ea voluptate velit esse quam nihil molestiae consequatur, vel illum qui dolorem eum fugiat quo voluptas nulla pariatur?" />
+
+		<TextBox Grid.Row="2"
+            x:Name="textWrapBind"
+                HorizontalAlignment="Center"
+                TextWrapping="Wrap"
+				Text="{x:Bind LocalTextWrapping, Mode=TwoWay}" 
+                Padding="8" /> 
+		
+	</Grid>
+</UserControl>
--- a/src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/TextBox/TextBox_Wrap.xaml.cs
+++ b/src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/TextBox/TextBox_Wrap.xaml.cs
@@ -1,0 +1,61 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Runtime.InteropServices.WindowsRuntime;
+using Windows.Foundation;
+using Windows.Foundation.Collections;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Controls;
+using Windows.UI.Xaml.Controls.Primitives;
+using Windows.UI.Xaml.Data;
+using Windows.UI.Xaml.Input;
+using Windows.UI.Xaml.Media;
+using Windows.UI.Xaml.Navigation;
+using Uno.UI.Samples.Controls;
+
+// The User Control item template is documented at https://go.microsoft.com/fwlink/?LinkId=234236
+
+namespace UITests.Windows_UI_Xaml_Controls.TextBox
+{
+	[Sample("TextBox")]
+	public sealed partial class TextBox_Wrap : UserControl
+    {
+
+		public static readonly DependencyProperty LocalTextWrappingProperty = DependencyProperty.Register(
+		"LocalTextWrapping", typeof(TextWrapping), typeof(TextBox_Wrap), new PropertyMetadata(TextWrapping.Wrap));
+
+		public TextWrapping LocalTextWrapping
+		{
+			get { return (TextWrapping)GetValue(LocalTextWrappingProperty); }
+			set { SetValue(LocalTextWrappingProperty, value); }
+		}
+
+		public TextBox_Wrap()
+        {
+            this.InitializeComponent();
+
+			textWrapBind.Text = textWrap.Text;
+		}
+
+		private void OnWrapButtonClick(object sender, Windows.UI.Xaml.RoutedEventArgs e)
+		{
+			if (buttonWrap.Content.ToString() == TextWrapping.Wrap.ToString())
+			{
+				textWrap.TextWrapping = TextWrapping.NoWrap;
+				textWrapBind.TextWrapping = TextWrapping.NoWrap;
+				buttonWrap.Content = textWrap.TextWrapping.ToString();
+				LocalTextWrapping = textWrap.TextWrapping;
+				textWrapBind.Text = textWrap.Text;
+			}
+			else
+			{
+				textWrap.TextWrapping = TextWrapping.Wrap;
+				buttonWrap.Content = textWrap.TextWrapping.ToString();
+				textWrapBind.TextWrapping = TextWrapping.Wrap;
+				LocalTextWrapping = textWrap.TextWrapping;
+				textWrapBind.Text = textWrap.Text;
+			}
+		}
+	}
+}
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.cs
@@ -1,4 +1,5 @@
-﻿using System;
+﻿
+using System;
 using System.Threading.Tasks;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Uno.UI.RuntimeTests.Helpers;
@@ -8,6 +9,7 @@ using Windows.UI.Xaml.Media;
 using static Private.Infrastructure.TestServices;
 using Windows.UI.Xaml;
 using Windows.UI;
+using FluentAssertions;
 #if NETFX_CORE
 using Uno.UI.Extensions;
 #elif __IOS__
@@ -293,6 +295,44 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 			Assert.IsTrue(textbox.IsEnabled);
 			Assert.AreEqual(contentPresenter.Foreground, foregroundColor);
 		}
+
+#if __ANDROID__
+		[TestMethod]
+		public async Task When_Text_IsWrapping_Set()
+		{
+			var textbox = new TextBox();
+
+			textbox.Width = 100;
+			StackPanel panel = new() { Orientation = Orientation.Vertical };
+			panel.Children.Add(textbox);
+			WindowHelper.WindowContent = panel;
+			await WindowHelper.WaitForLoaded(textbox);
+			var originalActualHeigth = textbox.ActualHeight;
+			textbox.Text = "Sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium doloremaaaaaaaaaaaaaaaaaa";
+			textbox.TextWrapping = TextWrapping.Wrap;
+			await WindowHelper.WaitForIdle();
+			textbox.ActualHeight.Should().BeGreaterThan(originalActualHeigth);
+		}
+
+
+		[TestMethod]
+		public async Task When_Text_IsNoWrap_Set()
+		{
+			var textbox = new TextBox();
+
+			textbox.Width = 100;
+			StackPanel panel = new() { Orientation = Orientation.Vertical };
+			panel.Children.Add(textbox);
+			WindowHelper.WindowContent = panel;
+			await WindowHelper.WaitForLoaded(textbox);
+			var originalActualHeigth = textbox.ActualHeight;
+			textbox.Text = "Sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium doloremaaaaaaaaaaaaaaaaaa";
+			textbox.TextWrapping = TextWrapping.NoWrap;
+			await WindowHelper.WaitForIdle();
+			textbox.ActualHeight.Should().Be(originalActualHeigth);
+		}
+
+#endif
 
 		[TestMethod]
 		public async Task When_SelectedText_StartZero()
--- a/src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.Android.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.Android.cs
@@ -380,7 +380,10 @@ namespace Windows.UI.Xaml.Controls
 
 		partial void OnTextWrappingChangedPartial(DependencyPropertyChangedEventArgs e)
 		{
-			//TODO : see bug #8178
+			if (_textBoxView != null && e.NewValue is TextWrapping textWrapping)
+			{
+				_textBoxView.SetSingleLine(textWrapping == TextWrapping.NoWrap);
+			}
 		}
 
 		partial void UpdateFontPartial()
--- a/src/Uno.UI/UI/Xaml/Controls/TextBox/TextBoxView.Android.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/TextBox/TextBoxView.Android.cs
@@ -42,7 +42,7 @@ namespace Windows.UI.Xaml.Controls
 			_ownerRef = WeakReferencePool.RentWeakReference(this, owner);
 			InitializeBinder();
 
-			base.SetSingleLine(true);
+			base.SetSingleLine(owner.TextWrapping == TextWrapping.NoWrap);
 
 			//This Background color is set to remove the native android underline on the EditText.
 			this.SetBackgroundColor(Colors.Transparent);
--- a/src/SamplesApp/SamplesApp.UITests/Windows_UI_Xaml_Controls/TextBoxTests/TextBoxTests.cs
+++ b/src/SamplesApp/SamplesApp.UITests/Windows_UI_Xaml_Controls/TextBoxTests/TextBoxTests.cs
@@ -559,6 +559,7 @@ namespace SamplesApp.UITests.Windows_UI_Xaml_Controls.TextBoxTests
 
 			// initial state
 			_app.WaitForElement(target);
+			target.SetDependencyPropertyValue("TextWrapping", "Wrap");
 			var multilineReadonlyTextRect = target.FirstResult().Rect.ToRectangle();
 
 			// remove readonly
@@ -657,6 +658,7 @@ namespace SamplesApp.UITests.Windows_UI_Xaml_Controls.TextBoxTests
 
 			_app.FastTap("btnDouble");
 			_app.WaitForElement("Test");
+			_app.Marked("Test").SetDependencyPropertyValue("TextWrapping", "Wrap");
 			var height2 = _app.GetLogicalRect("Test").Height;
 
 			using var _ = new AssertionScope();


src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/TextBox/TextBox_Wrap.xaml.cs
```diff
@@ -0,0 +1,61 @@
...
+		}
+
+		public TextBox_Wrap()
+        {
+            this.InitializeComponent();
+
+			textWrapBind.Text = textWrap.Text;
+		}
+
+		private void OnWrapButtonClick(object sender, RoutedEventHandler e)
```
jeromelaban, line 41: Make sure to build this sample locally first, this parameter is likely incorrect.

jeromelaban (commented):
Thanks for the PR! Can you add a runtime test that validates the height of the control ? Now that it wraps, its height should change based on the content.

See https://platform.uno/docs/articles/uno-development/creating-runtime-tests.html?q=runtime for details.

src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/TextBox/TextBox_Wrap.xaml
```diff
@@ -0,0 +1,35 @@
...
+			<RowDefinition Height="300"></RowDefinition>
+			<RowDefinition Height="*"></RowDefinition>
+		</Grid.RowDefinitions>
+		<Button x:Name="buttonWrap" Grid.Row="0" Content="Wrap"  Click="OnWrapButtonClick"></Button>
+		<TextBox Grid.Row="1"
+            x:Name="textWrap"
+                HorizontalAlignment="Center"
+                TextWrapping="Wrap"
+                Padding="8"
+               Text="Now what are the possibilities of warp drive? Cmdr Riker's nervous system has been invaded by an unknown microorganism. The organisms fuse to the nerve, intertwining at the molecular level. That's why the transporter's biofilters couldn't extract it. The vertex waves show a K-complex corresponding to an REM state. The engineering section's critical. Destruction is imminent. Their robes contain ultritium, highly explosive, virtually undetectable by your transporter." />
```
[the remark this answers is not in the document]
jeromelaban: It likely is, let's use some https://lipsum.com/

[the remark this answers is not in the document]
aesteves900: @jeromelaban , I found the text in another teste inside the Uno Platform. But thank you for the tip.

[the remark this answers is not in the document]
aesteves900: I will change as well, just to avoid any concern.

src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/TextBox/TextBox_Wrap.xaml.cs
aesteves900: It was fixed on the commit (7c2ec74)

src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/TextBox/TextBox_Wrap.xaml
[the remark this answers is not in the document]
aesteves900: It was fixed on the commit (7c2ec74 and 3bc873b)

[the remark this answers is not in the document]
jeromelaban: Thanks! Make sure to tidy up your commits so they conform to "conventional commits" (the build is failed because of this)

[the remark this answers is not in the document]
davidjohnoliver: this is gibberish, I'm pretty sure it's just sci-fi-themed lorem ipsum :D

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.cs
```diff
@@ -305,5 +305,69 @@ public async Task When_IsEnabled_Set()
...
+			};
+
+			WindowHelper.WindowContent = stackPanel;
+
+			await WindowHelper.WaitForLoaded(textbox);
+
+			textbox.Text = "Sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium doloremaaaaaaaaaaaaaaaaaa";
+			textbox.TextWrapping = TextWrapping.NoWrap;
+
+			Android.Graphics.Rect textBounds = new Android.Graphics.Rect();
```
jeromelaban, line 362: This is likely not going to work on non-android platforms. Did `ActualHeight` not work?

aesteves900: I saw a divergence about the Height measurement in-between TextBox.ActualHeight and the _ textBoxView (TextView wrapper) ActualHeigth, because of this my test fail. Can I equalize both values according to the Wrapper (TextView) values? Take a look at the image below, please

image

jeromelaban: This is the margins integrated in the TextBox. You can only validate that the height is bigger as wrapping is enabled.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.cs
```diff
@@ -312,29 +312,28 @@ public async Task When_Text_IsWrapping_Set()
...
+			WindowHelper.WindowContent = textbox;
 
 			await WindowHelper.WaitForLoaded(textbox);		
 			
 			textbox.Text = "Sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium doloremaaaaaaaaaaaaaaaaaa";
 			textbox.TextWrapping = TextWrapping.Wrap;
+			await WindowHelper.WaitForIdle();
 
-			Android.Graphics.Rect textBounds = new Android.Graphics.Rect();
+			var field = textbox.GetType().GetField("_textBoxView", System.Reflection.BindingFlags.NonPublic
```
jeromelaban, line 327: Avoid reflection whenever possible. In this case, you can find the subview using `FirstFirstChild`, like this:

https://github.com/unoplatform/uno/blob/acc3cab3e2940e9c3ac7845c4d47ffad28389239/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_Native_View.cs#L36

and use `TextBoxView` as the type.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.cs
```diff
@@ -312,29 +312,28 @@ public async Task When_Text_IsWrapping_Set()
...
+				| System.Reflection.BindingFlags.Instance);
 
-			textbox.GetHitRect(textBounds);
+			var textBoxView = field.GetValue(textbox);
 
-			await WindowHelper.WaitForIdle();
-			
-			Assert.IsTrue(textBounds.Height() <= initialHeight);						
+			var linecountProperty = textBoxView.GetType().GetProperty("LineCount");
```
jeromelaban, line 332: LineCount is not the right property, this will not be available on all platforms.

Get `ActualHeight` before and after and check that `ActualHeight` is greater. If it is, the test is successful.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.cs
aesteves900: I did the change one more time and like I reported before ActualHeight never changes after wrapping and unwrapping the text. Still failing to assert condition.

image

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.cs
aesteves900: I did the change one more time and as I reported before ActualHeight never changes after wrapping and unwrapping the text. Still failing to assert condition.

image

jeromelaban: It'd be best to display the values used in IsTrue, or to use IsGreater or similar. Are you reproducing the error on your machine?

aesteves900: Yes, @jeromelaban I could reproduce this error in my machine. It was fixed and passes to the pipeline now!

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.Android.cs
```diff
@@ -371,9 +371,31 @@ private void UpdateInputScope(InputScope inputScope)
...
+			else
+			{
+				if (InternalIsSingleLine.HasValue)
+				{
+					_textBoxView?.SetSingleLine(InternalIsSingleLine.Value);
+				}
+			}
 		}
 
+		private bool? InternalIsSingleLine { get; set; }
```
carldebilly, line 397: This doesn't need to be a property. It can be a field.

```suggestion
		private bool? _isSingleLine;
```

It's unclear to me what the `null` value means, when it's set and it's never reset.

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.Android.cs
```diff
@@ -371,9 +371,31 @@ private void UpdateInputScope(InputScope inputScope)
...
+			{
+
+				if (_textBoxView != null)
+				{
+					_isSingleLine = _textBoxView.IsSingleLine;
+
+					if (_textBoxView.IsSingleLine)
+					{
+						_textBoxView.SetSingleLine(false);
+					}
```
carldebilly, line 385: Maybe this needs comments to explain the pattern.  I don't understand what it is doing.

My questions remains: why do we need such `_isSingleLine` variable? What does the null value means here?

Also... If I trigger wrap/non-wrap on the TextBox, I'm not sure it will reset the `_isSingleLine` to the correct value.

carldebilly: As I said, needs to comment the pattern and explain:
1. why we need to keep the value of `_isSingleLine` in a field
2. what the `null` value mean on this field.
3. ensure the reset is done properly when the control is reused.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.cs
```diff
@@ -305,5 +306,58 @@ public async Task When_IsEnabled_Set()
...
+
+			var actualHeigth = textbox.ActualHeight;
+
+			textbox.Text = "Sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium doloremaaaaaaaaaaaaaaaaaa";
+			textbox.TextWrapping = TextWrapping.Wrap;
+			await WindowHelper.WaitForIdle();
+
+			var textBoxView = textbox.FindFirstChild<TextBoxView>();
+
+			Assert.IsTrue(textBoxView.LineCount > 1);
```
jeromelaban, line 332: Does this work?
```suggestion
			Assert.IsGreater(actualHeigth, textbox.ActualHeight);
```

It'd be better to use this as it will work on all platforms.

aesteves900:
@jeromelaban , In this case, the Assert condition will fail because the ActualHeight never changes after you set text to Wrap. I did the test just to fix the issue about Android. Can I open another issue to check de behavior for all the remaining platforms?

jeromelaban:
It's very curious that the ActualHeight does not change, particularly if the text is supposed to wrap. Doesn't the textbox get taller as a result?

aesteves900:
> It's very curious that the ActualHeight does not change, particularly if the text is supposed to wrap. Doesn't the textbox get taller as a result?

The text will be wrapped in lines according to visible width, it doesn't impact in ActualHeigth. If you have a textbox with 500 of Heigth with text not wrapped and you trigger the Wrap text, TextView natively creates new lines according to the text (visible width), and a scrollbar (if configured) does not change the ActualHeigth. Just one scenario for example.

jeromelaban:
That's fair. In that case, we can read the inner control's ScrollViewer ViewPort size. @dr1rrb will likely have some additional information here.

dr1rrb:
> That's fair. In that case, we can read the inner control's ScrollViewer ViewPort size. @dr1rrb will likely have some additional information here.

What you are looking for is the [`ExtentHeight`](https://github.com/unoplatform/uno/blob/master/src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.cs#L440) of the `ScrollViewer` of the template of the `TextBox`

jeromelaban:
In runtime tests, note that you can get the internal ScrollViewer by using `.FindFirstChild<ScrollViewer>();`.

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.Android.cs
```diff
@@ -371,9 +371,38 @@ private void UpdateInputScope(InputScope inputScope)
...
+				if (_textBoxView != null)
+				{
+					//store  the value configured on the wrapper _textBoxView to restore if user changes from Wrap to UnWrap
+					_isSingleLine = _textBoxView.IsSingleLine;
+
+					//just change if IsSingleLine equals true, to avoid invoke the propertychange in the wrong way
+					if (_textBoxView.IsSingleLine)
+					{
+						//this must be setted to false to Wrap the text
+						_textBoxView?.SetSingleLine(false);
```
Youssef1313, line 386: Null check is unnecessary

```suggestion
						_textBoxView.SetSingleLine(false);
```

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.Android.cs
```diff
@@ -371,9 +371,38 @@ private void UpdateInputScope(InputScope inputScope)
...
+					}
+				}
+			}
+			else
+			{
+				//just check if the "user value" was stored 
+				if (_isSingleLine.HasValue)
+				{
+					//return the value configured by user
+					_textBoxView?.SetSingleLine(_isSingleLine.Value);
```
Youssef1313, line 396: If we have a value for `_isSingleLine`, then we already had a non-null `_textBoxView`, so the null check is unnecessary here too.

```suggestion
					_textBoxView.SetSingleLine(_isSingleLine.Value);
```

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.Android.cs
```diff
@@ -371,9 +371,38 @@ private void UpdateInputScope(InputScope inputScope)
...
+
+					//reset the value
+					_isSingleLine = null;
+				}
+			}
 		}
 
+		private bool? _isSingleLine;
```
Youssef1313, line 404: Consider moving this line to the existing fields at the top:

https://github.com/unoplatform/uno/blob/6aa0f7003820bb5b0eaafcce6902943e9df0247c/src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.Android.cs#L33-L37

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.Android.cs
```diff
@@ -371,9 +371,38 @@ private void UpdateInputScope(InputScope inputScope)
 
 		partial void OnTextWrappingChangedPartial(DependencyPropertyChangedEventArgs e)
 		{
-			//TODO : see bug #8178
+			if (e.NewValue is TextWrapping textWrapping && !string.IsNullOrEmpty(Text) &&
+				(textWrapping is TextWrapping.Wrap || textWrapping is TextWrapping.WrapWholeWords))
```
Youssef1313, line 375: *Not much important:*

If the repository uses C# 9 or higher, I *think* this could be simplified with `or` patterns to:

```suggestion
			if (!string.IsNullOrEmpty(Text) &&
				(e.NewValue is TextWrapping.Wrap or TextWrapping.WrapWholeWords))
```

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.cs
```diff
@@ -306,6 +307,59 @@ public async Task When_IsEnabled_Set()
...
+			await WindowHelper.WaitForLoaded(textbox);
+
+			textbox.Text = "Sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium doloremaaaaaaaaaaaaaaaaaa";
+			textbox.TextWrapping = TextWrapping.NoWrap;
+		
+			await WindowHelper.WaitForIdle();
+
+			var textBoxView = textbox.FindFirstChild<TextBoxView>();
+
+			Assert.IsTrue(textBoxView.LineCount <= 1);
```
Youssef1313, line 358: This means we assert either a zero or one. Shouldn't it be exactly one? or I'm missing something?

Youssef1313 (commented):
Does `SetSingleLine` modify the Text of native control? If so, is that expected?

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.Android.cs
```diff
@@ -371,9 +371,38 @@ private void UpdateInputScope(InputScope inputScope)
...
+			{
+				if (_textBoxView != null)
+				{
+					//store  the value configured on the wrapper _textBoxView to restore if user changes from Wrap to UnWrap
+					_isSingleLine = _textBoxView.IsSingleLine;
```
MartinZikmund, line 380: Why is it necessary to store the "old" value instead of just directly setting true/false?

MartinZikmund (approved):

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.Android.cs
carldebilly: Questions not answered

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.cs
Youssef1313, line 358: I'm confused by `<= 1`
Shouldn't it be just `== 1`? If not, it might be worth explaining that in a comment.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.cs
```diff
@@ -306,6 +307,59 @@ public async Task When_IsEnabled_Set()
...
+			var actualHeigth = textbox.ActualHeight;
+
+			textbox.Text = "Sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium doloremaaaaaaaaaaaaaaaaaa";
+			textbox.TextWrapping = TextWrapping.Wrap;
+			await WindowHelper.WaitForIdle();
+
+			var textBoxView = textbox.FindFirstChild<TextBoxView>();
+
+			Assert.IsTrue(textBoxView.LineCount > 1);
+		}
```
jeromelaban, line 333: Here's how you can change the test, by using the ActualHeight of the control. You need to place the control in a container which lets the TextBox set its own height (`StackPanel` in that case), so that when the wrapping is changed, its height changes.

The test will work on all platforms this way.

```suggestion
		public async Task When_Text_IsWrapping_Set()
		{
			var textbox = new TextBox();
			
			textbox.Width = 100;

			StackPanel panel = new() { Orientation = Orientation.Vertical };
			panel.Children.Add(textbox);

			WindowHelper.WindowContent = panel;

			await WindowHelper.WaitForLoaded(textbox);

			var originalActualHeigth = textbox.ActualHeight;

			textbox.Text = "Sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium doloremaaaaaaaaaaaaaaaaaa";
			textbox.TextWrapping = TextWrapping.Wrap;
			await WindowHelper.WaitForIdle();

			textbox.ActualHeight.Should().BeGreaterThan(originalActualHeigth);
		}

```

aesteves900:
@carldebilly 

1 - why we need to keep the value of _isSingleLine in a field

The SetSingleLine on TextViewBox contributor is set to true, that's blocks the Text Wrapping. To avoid broken other scenarios I just store the default value and backup if the user changes the TextWrapping property. I tried to not store the default value, but I have broken a lot of tests, because of this I stored the value in a short way. My conclusion is when I store the value I'm not changing the warmup of the TextBoxView.

image



2 - what the null value means on this field.

It means that the value was restored., because the field was just used to not create interference on TextBoxView warmup.

3 - ensure the reset is done properly when the control is reused.

The field is reset if the user changes the value of TextWrapping property.

jeromelaban:
> @carldebilly
> 
> 1 - why we need to keep the value of _isSingleLine in a field
> 
> The SetSingleLine on TextViewBox contributor is set to true, that's blocks the Text Wrapping. To avoid broken other scenarios I just store the default value and backup if the user changes the TextWrapping property. I tried to not store the default value, but I have broken a lot of tests, because of this I stored the value in a short way. My conclusion is when I store the value I'm not changing the warmup of the TextBoxView.
> 
> image
> 
> 2 - what the null value means on this field.
> 
> It means that the value was restored., because the field was just used to not create interference on TextBoxView warmup.
> 
> 3 - ensure the reset is done properly when the control is reused.
> 
> The field is reset if the user changes the value of TextWrapping property.

Can you provide which tests were failing?

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.Android.cs
```diff
@@ -371,7 +371,18 @@ private void UpdateInputScope(InputScope inputScope)
 
 		partial void OnTextWrappingChangedPartial(DependencyPropertyChangedEventArgs e)
 		{
-			//TODO : see bug #8178
+			if (e.NewValue is TextWrapping textWrapping && !string.IsNullOrEmpty(Text) &&
+				(textWrapping is TextWrapping.Wrap || textWrapping is TextWrapping.WrapWholeWords))
```
Youssef1313, line 375:
```suggestion
			if (e.NewValue is TextWrapping textWrapping &&
				(textWrapping is TextWrapping.Wrap || textWrapping is TextWrapping.WrapWholeWords))
```

**Reason for the suggested change (please verify it with a test if possible)**

Assume we have a `TextBox` as follows:

```csharp
var textBox = new TextBox { Text = "", TextWrapping = TextWrapping.NoWrap };
```

After that, we change `TextWrapping` to `TextWrapping.Wrap`, we'll not change anything due to `IsNullOrEmpty` check.

Then, if `Text` was changed to a long text that needs wrapping, we'll end up in a state where we have a long text, and wrapping is requested, but it doesn't happen because `IsSingleLine` is `true`

Youssef1313 (changes requested):
The same logic must be applied here:

https://github.com/unoplatform/uno/blob/6d723331c8607e6c3d4a24b517f953a24f1b611c/src/Uno.UI/UI/Xaml/Controls/TextBox/TextBoxView.Android.cs#L45

jeromelaban:
> @jeromelaban , there are two problems, first, it's about the Assert condition that fails in other platforms after your suggestion above, second some RuntimeTests are broken as you can see in the screencap:

Sure, then let's enable the test for the platforms that the validation works on, we'll at least know that the behavior is the right one for those.

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.Android.cs
```diff
@@ -371,9 +371,37 @@ private void UpdateInputScope(InputScope inputScope)
...
+				//just check if the "user value" was stored 
+				if (_isSingleLine.HasValue)
+				{
+					//return the value configured by user
+					_textBoxView.SetSingleLine(_isSingleLine.Value);
+
+					//reset the value
+					_isSingleLine = null;
+				}
+			}
```
jeromelaban, line 400: @aesteves900 you do not need to store the value here, as `TextBoxView` is only manipulated by `TextBox` (unless you've seen otherwise). If the text wrapping does not allow multiline, invoke `SetSingleLine(true)` otherwise `SetSingleLine(false)`.

```suggestion
		   if (_textBoxView != null && e.NewValue is TextWrapping textWrapping)
		   {
		        var needsWrapping = textWrapping is TextWrapping.Wrap || textWrapping is TextWrapping.WrapWholeWords;
               _textBoxView.SetSingleLine(!needsWrapping);
		   }
```

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.Android.cs
jeromelaban: You should be able to invoke `_textBoxView?.SetSingleLine(false);` directly, without storing the `IsSingleLine` as the only interactions for `TextBoxView` come from `TextBox`.

aesteves900:
@jeromelaban , thank you for your suggestion!

I did the change, but it has broken a lot of the tests, just because the initial value was not stored.  (_isSingleLine). What do you think about that? Can I fix these tests or back up to the first solution?


image

jeromelaban:
@aesteves900 have you debugged locally why that happens? Looks like there may be some case missing, maybe because the SetSingleLine should always be called, not just because wrapping is enabled.

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.Android.cs
```diff
@@ -371,7 +371,11 @@ private void UpdateInputScope(InputScope inputScope)
 
 		partial void OnTextWrappingChangedPartial(DependencyPropertyChangedEventArgs e)
 		{
-			//TODO : see bug #8178
+			if (_textBoxView != null && e.NewValue is TextWrapping textWrapping)
+			{
+				var needsWrapping = textWrapping is TextWrapping.Wrap || textWrapping is TextWrapping.WrapWholeWords;
+				_textBoxView.SetSingleLine(!needsWrapping);
```
Youssef1313, line 377: Same thing but a bit smaller and more readable (IMO)

```suggestion
				_textBoxView.SetSingleLine(textWrapping == TextWrapping.NoWrap);
```

I also think we need to do the same in:

https://github.com/unoplatform/uno/blob/5e0c3818f1badff1f148325e7c3ff7b53647d4a1/src/Uno.UI/UI/Xaml/Controls/TextBox/TextBoxView.Android.cs#L45

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBoxView.Android.cs
```diff
@@ -43,7 +43,7 @@ public TextBoxView(TextBox owner)
 			InitializeBinder();
 
 			base.SetSingleLine(true);
```
Youssef1313, line 45: I'm feeling this should be updated too.

```suggestion
			base.SetSingleLine(owner.TextWrapping == TextWrapping.NoWrap);
```

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.Android.cs
```diff
@@ -371,7 +371,10 @@ private void UpdateInputScope(InputScope inputScope)
 
 		partial void OnTextWrappingChangedPartial(DependencyPropertyChangedEventArgs e)
 		{
-			//TODO : see bug #8178
+			if (_textBoxView != null && e.NewValue is TextWrapping textWrapping && e.OldValue is TextWrapping oldTextWrapping && oldTextWrapping != textWrapping)
+			{
+				_textBoxView?.SetSingleLine(textWrapping == TextWrapping.NoWrap);
```
Youssef1313, line 376:
```suggestion
				_textBoxView.SetSingleLine(textWrapping == TextWrapping.NoWrap);
```

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.Android.cs
```diff
@@ -371,7 +371,10 @@ private void UpdateInputScope(InputScope inputScope)
 
 		partial void OnTextWrappingChangedPartial(DependencyPropertyChangedEventArgs e)
 		{
-			//TODO : see bug #8178
+			if (_textBoxView != null && e.NewValue is TextWrapping textWrapping && e.OldValue is TextWrapping oldTextWrapping && oldTextWrapping != textWrapping)
```
jeromelaban, line 374: It's not a big deal, but just to you know, property changed events are only raised when the value has really changed:

https://github.com/unoplatform/uno/blob/d403a71fe22c32f4a3531b202e9a255c6261906c/src/Uno.UI/UI/Xaml/DependencyObjectStore.cs#L1914

 so there's no need to compare it again here:
```suggestion
			if (_textBoxView != null && e.NewValue is TextWrapping textWrapping && e.OldValue is TextWrapping oldTextWrapping)
```

jeromelaban (approved):

jeromelaban:
Let's rerun the tests one more time, but it looks like there's a good reason for the tests to fail:
image


https://dev.azure.com/uno-platform/Uno%20Platform/_TestManagement/Runs?runId=1803190&_a=resultQuery

PR #8177, opened 2022-02-23
fix(wasm_freeze): Fixed sometimes the UI froze when recycling elements.

carldebilly:


francoistanguay (approved):

src/Uno.UI/UI/Xaml/UIElement.wasm.cs
```diff
@@ -426,7 +426,7 @@ public Rect Arranged
 				UpdateDOMProperties();
 			}
 
-			if (FeatureConfiguration.UIElement.UseInvalidateMeasurePath && this.GetParent() is UIElement parent)
+			if (IsLoaded && FeatureConfiguration.UIElement.UseInvalidateMeasurePath && this.GetParent() is UIElement parent)
```
dr1rrb, line 429: Hummm, do we really need to invalidate the layout when **HIT TEST** visibility changes? For me it looks like a wrong copy paste between the `VisibilityChanged` and the `HitTestVisibilityChanged`.

dr1rrb (approved):

PR #8185, opened 2022-02-24
fix: Always set `ItemsControlFromItemContainer`, `VisualTreeHelper.GetParent` must be able to reach `ItemsControl`

### LINKED ISSUE #8184: Item containers should be able to reach `ItemsControl` using `VisualTreeHelper.GetParent`
### Current behavior

The list item containers `Parent` property is `null` and `VisualTreeHelper` is not able to reach the `ItemsControl`

### Expected behavior

While list item container's `Parent` property is `null` when within a `ItemsControl`, it should still be possible to reach the parent via `VisualTreeHelper.GetParent`

_No response_

### Workaround

_No response_

### Works on UWP/WinUI

Yes

### Environment

Uno.UI / Uno.UI.WebAssembly / Uno.UI.Skia, Uno.WinUI / Uno.WinUI.WebAssembly / Uno.WinUI.Skia

### NuGet package version(s)

_No response_

### Affected platforms

iOS, Android, WebAssembly, WebAssembly renderers for Xamarin.Forms, macOS, Skia (WPF), Skia (GTK on Linux/macOS/Windows), Skia (Tizen)

### IDE

Visual Studio 2022

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

MartinZikmund:
GitHub Issue (If applicable): closes #8184, closes [#313](https://github.com/unoplatform/nventive-private/issues/313)

## PR Type

- Bugfix

- In some cases the `ItemsControlFromItemContainer` property for item containers is not initialized
- `VisualTreeHelper.GetParent` on item containers returns `null` in `ItemsControl`

## What is the new behavior?

- `ItemsControlFromItemContainer` is always initialized
- `VisualTreeHelper.GetParent`  returns the owner `ItemsControl` instance

## PR Checklist


src/Uno.UI/UI/Xaml/DependencyObjectExtensions.cs
```diff
@@ -95,12 +91,25 @@ internal static Style StoreGetImplicitStyle(this IDependencyObjectStoreProvider
...
+				}
+
+				if (parent is not null)
+				{
+					yield return parent;
+				}
+
+				current = parent;
+
+			} while (current != null);
```
MartinZikmund, line 112: @jeromelaban @dr1rrb @carldebilly Not sure about this change - should `GetParents()` only include "real" parents and stop when `GetParent()` returns `null` or include the `ItemsControl` as parent for `SelectorItem`?

jeromelaban: This method is an internal version which only tracks DependencyObject hierarchy, not visual tree ones. If VisualTreeHelper needs to behave different, the specificity should be in VisualTreeHelper.

carldebilly (approved):

src/Uno.UI/UI/Xaml/Media/VisualTreeHelper.cs
```diff
@@ -169,12 +170,20 @@ public static IReadOnlyList<Popup> GetOpenPopupsForXamlRoot(XamlRoot xamlRoot)
...
 #if XAMARIN
-			return (reference as _ViewGroup)?
+			realParent = (reference as _ViewGroup)?
 				.FindFirstParent<DependencyObject>();
 #else
-			return reference.GetParent() as DependencyObject;
+			realParent = reference.GetParent() as DependencyObject;
 #endif
+
+			if (realParent is null && reference is SelectorItem selectorItem)
```
dr1rrb, line 181: I would assume this condition is true only when we reach a `NativeListViewBase`. If I'm right, discriminating on `SelectorItem` looks weird / weak. I think here we should just support native elements and pass through them using native API to reach their managed parents.

This is really similar to https://github.com/unoplatform/uno/blob/master/src/Uno.UI/UI/Xaml/UIElement.Android.cs#L227, except this will also compute offsets to compute transform.

src/Uno.UI/UI/Xaml/DependencyObjectExtensions.cs
MartinZikmund: Undone

src/Uno.UI/UI/Xaml/Media/VisualTreeHelper.cs
MartinZikmund: I tried replacing the logic to use `GetVisualParent` on non-Skia and non-WASM targets, however on Skia and WASM it just gets the `DependencyObjectStore` parent, which does not work for list items, so on WASM and Skia I am keeping the existing condition currently. Maybe it then would also make sense to add similar logic into that transform computation, as it now would not work for lists on Skia and WASM either, if I am not mistaken

MartinZikmund:
\azp run

MartinZikmund:
@dr1rrb Adjusted the code to correctly return the parent for `GetVisualTreeParent`. Seems it really was not returning the parent correctly, even in case of iOS and Android (for items the parent was `null`)

src/Uno.UI/Extensions/UIViewExtensions.iOSmacOS.cs
```diff
@@ -295,7 +287,18 @@ public static void AddChild(this _View parent, _View child)
 		/// <summary>
 		/// Get the parent view in the visual tree. This may differ from the logical <see cref="FrameworkElement.Parent"/>.
 		/// </summary>
-		public static _View GetVisualTreeParent(this _View child) => child?.Superview;
+		public static _View GetVisualTreeParent(this _View child)
+		{
+			var visualParent = child?.Superview;
+
+			if (visualParent is null &&
```
dr1rrb, line 294: I would add a comment to explicit the cases where the `visualParent` could be null while `fw.VisualPrent` is not.

src/Uno.UI/Extensions/ViewExtensions.Android.cs
```diff
@@ -390,7 +390,18 @@ public static void AddChild(this ViewGroup parent, View child)
 		/// <summary>
 		/// Get the parent view in the visual tree. This may differ from the logical <see cref="FrameworkElement.Parent"/>.
 		/// </summary>
-		public static ViewGroup? GetVisualTreeParent(this View child) => child?.Parent as ViewGroup;
+		public static ViewGroup? GetVisualTreeParent(this View child)
+		{
+			var visualParent = child?.Parent as ViewGroup;
+
+			if (visualParent is null &&
```
dr1rrb, line 397: Same here :)

dr1rrb (approved):

Issue #6908, opened 2021-08-30
Add support for `Windows.Globalization.NumberFormatting`


MartinZikmund:
## What would you like to be added:

Support for functionality under `Windows.Globalization.NumberFormatting`.

## Why is this needed:

For many scenarios of `NumberBox`, this namespace is crucial.

## For which Platform:

All

marwalsch:
Any update on this?

MohammadHadi2031:
Currently just the "CurrencyFormatter" is not implemented.

MartinZikmund:
Still relevant

MartinZikmund:
# Repro generator

This issue was classified as not reproducible with a code sample or test.

**Reason:** Feature request for Windows.Globalization.NumberFormatting.

PR #8348, opened 2022-03-15
feat(CurrencyFormatter): Implement FormatDouble and ParseDouble

MohammadHadi2031:
GitHub Issue (If applicable): part of #6908

## PR Type

- Feature

Nothing.

## What is the new behavior?
Can format and parse doubles.

## PR Checklist

## Other information


MartinZikmund:
Wow, you are fast!!!!

MartinZikmund (approved):

src/Uno.UWP/Globalization/NumberFormatting/Extensions.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System;
+
+namespace Windows.Globalization.NumberFormatting
```
MartinZikmund, line 3: If possible, can you switch the newly created code files to use file-scoped namespaces 👍 ?

src/Uno.UWP/Globalization/NumberFormatting/CurrencyFormatter.cs
```diff
@@ -0,0 +1,160 @@
...
+namespace Windows.Globalization.NumberFormatting
+{
+	public partial class CurrencyFormatter : INumberParser, INumberFormatter2, INumberFormatter, INumberFormatterOptions, ISignificantDigitsOption, INumberRounderOption, ISignedZeroOption
+	{
+		private const char NoBreakSpaceChar = ' ';
+
+		private readonly FormatterHelper _formatterHelper;
+		private readonly NumeralSystemTranslator _translator;
+
+		private CurrencyData _currencyData;
```
MartinZikmund, line 14: Can be `readonly`

src/Uno.UWP/Globalization/NumberFormatting/Extensions.cs
MohammadHadi2031: Sure

src/Uno.UWP/Globalization/NumberFormatting/CurrencyData.cs
```diff
@@ -0,0 +1,354 @@
+﻿namespace Windows.Globalization.NumberFormatting
```
jeromelaban:
```suggestion
#nullable enable

﻿namespace Windows.Globalization.NumberFormatting
```

jeromelaban (commented):
Thanks a lot for your contribution!

I've pushed a small update to ensure that we can validate the tests when running the UWP samples app as a reference. Let's validate this first to merge the PR.

src/Uno.UWP/Globalization/NumberFormatting/CurrencyFormatter.cs
```diff
@@ -0,0 +1,160 @@
+﻿using System;
```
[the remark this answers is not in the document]
MohammadHadi2031: @jeromelaban Hi, I'm using "#nullable" wherever possible but we can't use it here because when CurrencyFormatter (and other formatters) initializes in UWP app default value for "NumberRounder" property is null. If we want to enable nullable we should change the "INumberRounderOption" Interface from `INumberRounder NumberRounder { get; set; }` to `INumberRounder? NumberRounder { get; set; }`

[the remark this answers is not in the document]
jeromelaban: Let's do this then, it makes sense to add nullability there. Thanks!

src/Uno.UI.Tests/Windows_Globalization/Given_NumeralSystemTranslator.cs
```diff
@@ -32,12 +51,29 @@ public void When_NumeralSystemIsNull_Then_Throw()
...
+			catch (Exception ex)
+			{
+				Assert.AreEqual("The parameter is incorrect.\r\n\r\nlanguages", ex.Message);
+			}
+			
 			Assert.ThrowsException<ArgumentException>(() => new NumeralSystemTranslator(languages));
 		}
 
 		[TestMethod]
 		public void When_LanguagesIsNull_Then_Throw()
```
MohammadHadi2031, line 67: @jeromelaban I'm removing this test for enabling nullable reference type

src/Uno.UI.Tests/Windows_Globalization/Given_NumeralSystemTranslator.cs
```diff
@@ -16,12 +16,31 @@ public class Given_NumeralSystemTranslator
...
+			catch (Exception ex)
+			{
+				Assert.AreEqual("The parameter is incorrect.\r\n\r\nnumeralSystem", ex.Message);
+			}
+			
 			Assert.ThrowsException<ArgumentException>(() => sut.NumeralSystem = numeralSystem);
 		}
 
 		[TestMethod]
 		public void When_NumeralSystemIsNull_Then_Throw()
```
MohammadHadi2031, line 33: @jeromelaban  removing this test too

src/Uno.UWP/Globalization/NumberFormatting/PermilleFormatter.cs
```diff
@@ -1,99 +1,100 @@
...
+	{
+		text = _translator.TranslateBackNumerals(text);
+
+		if (!text.EndsWith(_symbol, StringComparison.Ordinal))
+		{
+			return null;
+		}
 
-			var result = _formatterHelper.ParseDoubleCore(text);
+		text = text.Substring(0, text.Length - _symbol.Length);
```
workgroupengineering, line 89: Like above

workgroupengineering (commented):
Number formatting is a critical area because it is a very common operation. I think it should be optimized as much as possible to avoid wasting memory. CurrencyFormatter should only allocate the final string not the intermediate ones.

src/Uno.UWP/Globalization/NumberFormatting/PermilleFormatter.cs
MohammadHadi2031: @workgroupengineering Is it necessary to use `StringBuilder` here because there is just one line that can be converted to StringBuilder style so if I use StringBuilder the `ToString()` creates new string?

workgroupengineering: you can `text.AsSpan().Slice(0,text.Length - _symbol.Length)`;

MohammadHadi2031: I would rather use `StringBuilder` as you said first.

jeromelaban:
> > @workgroupengineering I changed the code to use StringBuilder. Please inform me if there is any issue or potential improvement
> 
> Instead of creating StringBuilder every time, you can create a StringBuilder pool. If I remember correctly in the UNO BCL there is already an Object Pool manager.

That's correct there is, but it's not meant to work for individual objects like this but rater for arrays. We could have one `StringBuilder` per thread, though, given that the code we're calling is not reentrant.

src/Uno.UWP/Globalization/NumberFormatting/CurrencyFormatter.cs
```diff
@@ -0,0 +1,175 @@
...
+		{
+			stringBuilder.Insert(0, '(');
+			stringBuilder.Append(')');
+		}
+
+		var formatted = stringBuilder.ToString();
+
+		StringBuilderPool.Instance.Return(stringBuilder);
+		return formatted;
+	}
```
MohammadHadi2031, line 124: @workgroupengineering I created simple StringBuilderPool. Also I converted more method to use StringBuilder. I'll be happy to hear from you.

src/Uno.UWP/Globalization/NumberFormatting/StringBuilderPool.cs
```diff
@@ -0,0 +1,45 @@
...
+
+		for (int i = 0; i < _items.Length; i++)
+		{
+			if (_items[i] == null)
+			{
+				_items[i] = obj;
+				return;
+			}
+		}
+	}
```
MohammadHadi2031, line 44: The `StringBuilderPool` is not thread-safe. We can copy & paste from[ Microsoft.Extensions.ObjectPool](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.objectpool) is it good idea?

jeromelaban: There's no need at this point. If we find that this is a performance concern, we'll adjust it at this time.

You can make a `StringBuilder` field marked with `[ThreadStatic]` and make sure to clear the instance it on each formatting call, it'll be good enough for our scenario.

MohammadHadi2031: As the [Documentation](https://docs.microsoft.com/en-us/dotnet/api/system.threadstaticattribute?view=net-6.0#remarks) says:

> Do not specify initial values for fields marked with ThreadStaticAttribute, because such initialization occurs only once, when the class constructor executes, and therefore affects only one thread. If you do not specify an initial value, you can rely on the field being initialized to its default value if it is a value type, or to null if it is a reference type.

if we use `ThreadStaticAttribute` on other threads the value of `StringBuilder` would be null. `[ThreadLocal]`(https://docs.microsoft.com/en-us/dotnet/api/system.threading.threadlocal-1?view=net-6.0) solves this issue. Do you agree to use `ThreadLocal`?

jeromelaban: Either one is good. `ThreadLocal` works as well.

src/Uno.UWP/Globalization/NumberFormatting/CurrencyFormatter.cs
workgroupengineering: Sorry for the delay, I've been busy.

src/Uno.UWP/Globalization/NumberFormatting/StringBuilderPool.cs
workgroupengineering: Maybe you could move StringBuilderPool to Uno's BCL so you can use it elsewhere.

MohammadHadi2031: Good idea but I think it would be dangerous because currently there isn't any way to ensure that StringBuilder1 is used by another code so being in `Windows.Globalization.NumberFormatting` is now better. To avoid the issue we should use ObjectPool as you said first.

src/Uno.UWP/Globalization/NumberFormatting/CurrencyData.cs
```diff
@@ -0,0 +1,356 @@
...
+{
+	public int DefaultFractionDigits { get; set; }
+	public string CurrencyCode { get; set; } = "";
+	public string Symbol { get; set; } = "";
+	public bool IsSymbolAfterNumber { get; set; }
+
+	#region Initialize
+	public static readonly CurrencyData Empty = new CurrencyData();
+
+	private static readonly CurrencyData _allCurrencyData = new CurrencyData { CurrencyCode = "ALL", Symbol = "Lek", DefaultFractionDigits = 2 };
```
jeromelaban, line 15: Did you build these fields out of the results from reading what UWP provides?

jeromelaban (commented):
I ran the UWP test app (in the build artifacts), and I'm getting 38 failures for `Given_CurrencyFormatter`:
image

I'm not sure though by how much the results are specific to my system. Could you validate?

All tests pass for:
- `Given_DecimalFormatter`
- `Given_PercentFormatter`
- `Given_PermilleFormatter`
- `Given_NumeralSystemTranslator`
- `Given_IncrementNumberRounder`

Almost there and thanks for the work!

src/Uno.UWP/Globalization/NumberFormatting/CurrencyData.cs
MohammadHadi2031: @jeromelaban Hi, I used this [link](https://www.eurochange.co.uk/travel-money/world-currency-abbreviations-symbols-and-codes-travel-money) to grab CurrencyCodes and then used [this code](https://github.com/MohammadHadi2031/UwpNumberFormatting/blob/e7b55d9657abfd32f04d67ea4ea62badd58394e5/DecimalFormatter.Tests/CodeGenerator.cs#L16) to generate fields based on what UWP provides.

MohammadHadi2031:
> I ran the UWP test app (in the build artifacts), and I'm getting 38 failures for `Given_CurrencyFormatter`: image
> 
> I'm not sure though by how much the results are specific to my system. Could you validate?
> 
> All tests pass for:
> 
> * `Given_DecimalFormatter`
> * `Given_PercentFormatter`
> * `Given_PermilleFormatter`
> * `Given_NumeralSystemTranslator`
> * `Given_IncrementNumberRounder`
> 
> Almost there and thanks for the work!

I'm trying to figure out what is happening but currently on my computer tests passes. for testing I use a [simple UWP project](https://github.com/MohammadHadi2031/UwpNumberFormatting/tree/main/DecimalFormatter.Tests) and run tests on that project. I downloaded the artifact and actually I'm getting more errors. I'm trying to solve the issue

src/Uno.UWP/Globalization/NumberFormatting/CurrencyData.cs
MohammadHadi2031: @jeromelaban I'm using CurrencyIdentifiers as a reference for currency codes.

MohammadHadi2031:
> I ran the UWP test app (in the build artifacts), and I'm getting 38 failures for `Given_CurrencyFormatter`: image
> 
> I'm not sure though by how much the results are specific to my system. Could you validate?
> 
> All tests pass for:
> 
> * `Given_DecimalFormatter`
> * `Given_PercentFormatter`
> * `Given_PermilleFormatter`
> * `Given_NumeralSystemTranslator`
> * `Given_IncrementNumberRounder`
> 
> Almost there and thanks for the work!

Hi. I found out that CurrencyFormatter uses [CurrencyPositivePattern](https://docs.microsoft.com/en-us/dotnet/api/system.globalization.numberformatinfo.currencypositivepattern?view=net-6.0) and [CurrencyNegativePattern](https://docs.microsoft.com/en-us/dotnet/api/system.globalization.numberformatinfo.currencynegativepattern?view=net-6.0) to format/parse doubles I updated the code but for testing it's little complicated. Now every test has expected value based on my windows globalization settings. for getting test passes independently we should generate expected value based on target windows globalization settings. So for generating correct expected value I should use the code that format double and I want to test that code too! If I generate it this way test always passes on non-UWP platform but may not pass on UWP. It seems that testing on non-UWP platform wouldn't be useful.

src/Uno.UWP/Globalization/NumberFormatting/StringBuilderCache.cs
```diff
@@ -0,0 +1,69 @@
...
+// Source Link: https://github.com/dotnet/runtime/blob/e63d21947e734db2da5093510a6636b5b7fb45b5/src/libraries/Common/src/System/Text/StringBuilderCache.cs
+// Commit: a9c5ead on Feb 10, 2021, https://github.com/dotnet/runtime/commit/a9c5eadd951dcba73167f72cc624eb790573663a
+// 
+#nullable enable
+//
+
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace System.Text
```
jeromelaban, line 11: Great! Let's move this to `Uno.Foundation`, into the `Uno` namespace: https://github.com/unoplatform/uno/tree/master/src/Uno.Foundation/Uno.Core.Extensions/Uno.Core.Extensions

MohammadHadi2031: Done.

MohammadHadi2031:
> > I ran the UWP test app (in the build artifacts), and I'm getting 38 failures for `Given_CurrencyFormatter`: image
> > I'm not sure though by how much the results are specific to my system. Could you validate?
> > All tests pass for:
> > 
> > * `Given_DecimalFormatter`
> > * `Given_PercentFormatter`
> > * `Given_PermilleFormatter`
> > * `Given_NumeralSystemTranslator`
> > * `Given_IncrementNumberRounder`
> > 
> > Almost there and thanks for the work!
> 
> Hi. I found out that CurrencyFormatter uses [CurrencyPositivePattern](https://docs.microsoft.com/en-us/dotnet/api/system.globalization.numberformatinfo.currencypositivepattern?view=net-6.0) and [CurrencyNegativePattern](https://docs.microsoft.com/en-us/dotnet/api/system.globalization.numberformatinfo.currencynegativepattern?view=net-6.0) to format/parse doubles I updated the code but for testing it's little complicated. Now every test has expected value based on my windows globalization settings. for getting test passes independently we should generate expected value based on target windows globalization settings. So for generating correct expected value I should use the code that format double and I want to test that code too! If I generate it this way test always passes on non-UWP platform but may not pass on UWP. It seems that testing on non-UWP platform wouldn't be useful. What should I do?

@jeromelaban Now I generate expected value based on Positive/Negative currency pattern. I manually tested the code by changing these pattern in Windows currency format settings. I think it's ready for review

MartinZikmund (approved):

jeromelaban:
Thanks @MohammadHadi2031! I took a look at the UWP build, and a lot of the tests are failing:

image

Issue #8324, opened 2022-03-13
[Android] Borders are choppy/rendering incorrectly


MartinZikmund:
### Current behavior

On Android rounded corners are now choppy or rendering incorrectly.

``` xaml
<Button CornerRadius="20" Content="Test" ... />
```

![image](https://user-images.githubusercontent.com/1075116/158080567-820c3040-c7ba-47a7-836d-987f20226c24.png)

``` xaml
<Button CornerRadius="2" Content="Test" ... />
```

![image](https://user-images.githubusercontent.com/1075116/158080608-bfa2ad56-4580-4307-a3ea-2fa4229cb10f.png)

Notice the rounded corners show up only on the right side of the buttons, although they should be applied on all corners.

The problem was most likely introduced with https://github.com/unoplatform/uno/pull/8118. It seems the changes need to be enabled conditionally - only for straight borders without rounding. 

When fixing, make sure the changes don't break https://github.com/unoplatform/uno/issues/6490 which was fixed by #8118 (and which was, interestingly, also happening for rounded corners to some extent - see discussion on #6490)

### Expected behavior

Borders should render correctly.

Following screenshots from 4.0.13:

![image](https://user-images.githubusercontent.com/1075116/158080947-192994b6-212e-431d-b80f-95c9591081e5.png)

![image](https://user-images.githubusercontent.com/1075116/158080913-b8d9087b-0448-41ac-afc4-a3ce66eb5ac5.png)

### Works on UWP/WinUI

Yes

### Environment

Uno.UI / Uno.UI.WebAssembly / Uno.UI.Skia, Uno.WinUI / Uno.WinUI.WebAssembly / Uno.WinUI.Skia

### NuGet package version(s)

4.2. dev, works in 4.1.8

### Affected platforms

Android

### IDE

Visual Studio 2022

MartinZikmund:
Not a regression as it works in 4.1.8. Problems are appearing in 4.2 dev

kazo0:
I believe it actually is a regression, this is 100% caused by https://github.com/unoplatform/uno/pull/8118. Looks like I was just doing the wrong thing when trying to reproduce

MartinZikmund:
@kazo0 that's odd, I have downgraded the app where I found the problem from 4.2 dev to 4.1.8 and the problem is not happening there 🤔

kazo0:
[...]

PR #8357, opened 2022-03-16
fix: Avoid disabling Paint AntiAlias when drawing rounded corners

kazo0:
closes https://github.com/unoplatform/uno/issues/8324

## PR Type

- Bugfix

image

## What is the new behavior?

image

### DIFF
--- a/src/SamplesApp/SamplesApp.UITests/Windows_UI_Xaml_Controls/BorderTests/Border_Tests.cs
+++ b/src/SamplesApp/SamplesApp.UITests/Windows_UI_Xaml_Controls/BorderTests/Border_Tests.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
 using NUnit.Framework;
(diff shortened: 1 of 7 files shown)


MartinZikmund (approved):

jeromelaban:
I think we should be adding a test that validates if there are non-plain colors caused by the aliasing. I think the image test helpers can do that.

ajpinedam (approved):
This is a good finding! 🔥

kazo0:
> I think we should be adding a test that validates if there are non-plain colors caused by the aliasing. I think the image test helpers can do that.

@jeromelaban I'm not sure I understand, by non-plain color you mean we just have something like a White background and a Red border and if we detect any pixels that are not White and not full red (255,0,0) then we assume there is anti-aliasing?

jeromelaban:
I think we should be adding a test that validates if there are non-plain colors caused by the aliasing. I think the image test helpers can do that.

> 

Something like this, yes :)

Issue #8289, opened 2022-03-09
[iOS] Buttons may keep unknown visual state after being pressed


jeromelaban:
### Current behavior

Using iOS latest calculator (4.2.0-dev.243)

https://user-images.githubusercontent.com/5839577/157540432-35ab9caa-c8b4-46a6-bee9-438fcc115819.mp4

### Expected behavior

Buttons don't stay half pressed

### Affected platforms

iOS

PR #8358, opened 2022-03-16
fix(pointers): Enter/Exit bubbling when crossing boundaries of OriginalSource

dr1rrb:
closes https://github.com/unoplatform/uno/issues/8289

## Bugfix
`PointerEntered` and `PointerExited` might not be raise nor bubble properly, especially when moving pointer while pressed from an element to another element.

1. When moving from an element to another element while pressed, on platform with "implicit capture" (iOS, Android, wasm-touch-chromium), when we try to "re-dispatch" up and exited, as the element is no longer considered pressed/hovered, we are filtering out the events instead of bubbling them as we should.
2. On iOS, the SCP (`UIScrollView`) might stop bubbling of native events for an unknown reason
3. When an event reach the top of the tree, the `RaiseEvent` method returns `true` which is interpreted has "handled" which alter the re-dispatching of exited event.

## What is the new behavior?
1. Either we make sure to flag "re-dispatched" events in a way that they are not filtering out, either we just disable the fitering when not needed at all.
2. [BREAKING] Pointer events are always bubble in managed by the SCP on iOS
3. The `RaiseEvent` now returns the `Handled` flag when reaching the top of the tree.

## PR Checklist
- [ ] ~~Contains **NO** breaking changes~~ **cf. Below**

## Other information
```

                                   ******************************
                                   ********** BREAKING **********
                                   ******************************
```

On iOS the `ScrollContentPresenter` is now bubbling all pointer events in managed only.
This means that a **native only** element that is hosting a managed `ScrollViewer` won't receive pointer event that occurs above that SV.

_Note: This is expected to not break any app since the native events was already not bubbling properly, so it would have already been broken._

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/ScrollContentPresenter/NativeScrollContentPresenter.iOS.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/ScrollContentPresenter/NativeScrollContentPresenter.iOS.cs
@@ -529,8 +529,11 @@ namespace Windows.UI.Xaml.Controls
 			// Like native dispatch on iOS, we do "implicit captures" the target.
 			if (this.GetParent() is UIElement parent)
 			{
+				// canBubbleNatively: true => We let native bubbling occur properly as it's never swallowed by system
+				//							  but blocking it would be breaking in lot of aspects.
+
 				_touchTarget = parent;
-				_touchTarget.TouchesBegan(touches, evt);
+				_touchTarget.TouchesBegan(touches, evt, canBubbleNatively: true); 
 			}
 		}
 
@@ -539,7 +542,8 @@ namespace Windows.UI.Xaml.Controls
 		{
 			base.TouchesMoved(touches, evt);
 
-			_touchTarget?.TouchesMoved(touches, evt);
+			// canBubbleNatively: false => system might silently swallow pointer after few moves so we prefer to bubble in managed.
+			_touchTarget?.TouchesMoved(touches, evt, canBubbleNatively: false);
 		}
 
 		/// <inheritdoc />
@@ -547,7 +551,8 @@ namespace Windows.UI.Xaml.Controls
 		{
 			base.TouchesEnded(touches, evt);
 
-			_touchTarget?.TouchesEnded(touches, evt);
+			// canBubbleNatively: false => system might silently swallow pointer after few moves so we prefer to bubble in managed.
+			_touchTarget?.TouchesEnded(touches, evt, canBubbleNatively: false);
 			_touchTarget = null;
 		}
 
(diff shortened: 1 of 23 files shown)


src/Uno.UI/UI/Xaml/Controls/ScrollContentPresenter/NativeScrollContentPresenter.iOS.cs
```diff
@@ -529,8 +529,11 @@ public override void TouchesBegan(NSSet touches, UIEvent evt)
 			// Like native dispatch on iOS, we do "implicit captures" the target.
 			if (this.GetParent() is UIElement parent)
 			{
+				// canBubbleNatively: true => We let native bubbling occur properly as it's never swallowed by system
+				//							  but blocking it would be breaking in lot of aspects.
```
jeromelaban, line 533: Which aspects (give some)?

src/Uno.UI/UI/Xaml/Controls/ScrollContentPresenter/NativeScrollContentPresenter.iOS.cs
```diff
@@ -539,15 +542,17 @@ public override void TouchesMoved(NSSet touches, UIEvent evt)
 		{
 			base.TouchesMoved(touches, evt);
 
-			_touchTarget?.TouchesMoved(touches, evt);
+			// canBubbleNatively: false => system might silently swallow pointer after few moves so we prefer to bubble in managed.
```
jeromelaban, line 545:
```suggestion
			// canBubbleNatively: false => The system might silently swallow pointers after a few moves so we prefer to bubble in managed.
```

src/Uno.UI/UI/Xaml/UIElement.Pointers.cs
```diff
@@ -870,7 +870,15 @@ private bool OnPointerDown(PointerRoutedEventArgs args, BubblingContext ctx = de
 			// it due to an invalid state. So here we make sure to not stay in an invalid state that would
 			// prevent any interaction with the application.
 			var isOverOrCaptured = ValidateAndUpdateCapture(args, isOver: true, forceRelease: true);
-			var handledInManaged = SetPressed(args, true, muteEvent: ctx.IsInternal || !isOverOrCaptured);
+			if (!isOverOrCaptured)
+			{
+				// We receive this event due to implicit capture, just ignore it locally and let is bubble
+				// note: If bubbling in managed then we are going to ignore element anyway as ctx is flagged as IsInternal.
```
jeromelaban, line 876:
```suggestion
				// note: If bubbling in managed then we are going to ignore the element anyway as ctx is flagged as IsInternal.
```

jeromelaban (approved):

PR #8368, opened 2022-03-17
fix: Add quotation marks on font family

### LINKED ISSUE #7395: [WASM] Custom font with number in `FontFamily` is not working
### Current behavior

I load & display custom woff2 font successfully to a wasm uno project,  except some FontFamily : 2Dumb  / 1942 report / Capture It 2   etc.    All these fonts have number in FontFamily.  These fonts can be displayed in Transfonter demo page, so I think it's an uno wasm problem.    Doing further research,  if I use 2Dumb woff2 font data, and uses a different FontFamily like "abcfont", it displays correctly.  

### Expected behavior

Load and display custom font with number in FontFamily correctly.

_No response_

### Workaround

_No response_

### Works on UWP/WinUI

_No response_

### Environment

### NuGet package version(s)

4.0.0-dev.5407

### Affected platforms

WebAssembly

### IDE

Visual Studio 2019

### IDE version

16.11.3 Preview 1.0

### Relevant plugins

_No response_

_No response_

CesarRabelo:
GitHub Issue (If applicable): closes #7395 

## PR Type

- Bugfix

Load custom font family starting with number does not work.

## What is the new behavior?

The custom font family starting with number is now loaded.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/ts/WindowManager.ts
+++ b/src/Uno.UI/ts/WindowManager.ts
@@ -565,8 +565,12 @@ namespace Uno.UI {
 			for (let i = 0; i < params.Pairs_Length; i += 2) {
 				const key = pairs[i];
 				const value = pairs[i + 1];
-
-				elementStyle.setProperty(key, value);
+				if (this.isStyleFontFamily(key)) {
+					elementStyle.setProperty(key, this.addQuotationMarks(value));
+				}
+				else {
+					elementStyle.setProperty(key, value);
+				}
 			}
 
 			return true;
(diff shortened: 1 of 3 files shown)


MartinZikmund:
I see you are checking if the font family needs quotation marks, but would it potentially cause any issues if we always added them 🤔 ?

src/Uno.UI/ts/WindowManager.ts
```diff
@@ -565,8 +565,12 @@ namespace Uno.UI {
...
 				const key = pairs[i];
 				const value = pairs[i + 1];
-
-				elementStyle.setProperty(key, value);
+				if (this.isStyleFontFamily(key)) {
+					elementStyle.setProperty(key, this.addQuotationMarks(value));
+				}
+				else {
+					elementStyle.setProperty(key, value);
+				}
```
jeromelaban, line 573: This feels like a performance issue. If that's the case, we'll need to change the style font family setter to be an explicit method for changing fonts through the TS Interop. What do you think @carldebilly ?

CesarRabelo:
> I see you are checking if the font family needs quotation marks, but would it potentially cause any issues if we always added them 🤔 ?

Thanks for your comment @MartinZikmund, but if put in all font-family we will have another bug.

sample-all-font-quotation-marks

MartinZikmund:
> Thanks for your comment @MartinZikmund, but if put in all font-family we will have another bug.

Isn't the reason it doesn't work in the video that there is some kind of encoding happening on the quotation marks? According to [this](https://stylelint.io/user-guide/rules/list/font-family-name-quotes/#options) quotes around font family names are recommended for all font family names except for "vendor prefixed system fonts", which we don't use 🤔 

image

src/Uno.UI/ts/WindowManager.ts
carldebilly: I was thinking the same thing...  the "adjustment" should not be done in the managed side, but in TS/JS.

jeromelaban: @CesarRabelo You'll need to call a new method for changing the font here:
https://github.com/unoplatform/uno/blob/5e0c3818f1badff1f148325e7c3ff7b53647d4a1/src/Uno.UI/UI/Xaml/UIElement.TextHelper.wasm.cs#L93

Then declare it the same class as this method:
https://github.com/unoplatform/uno/blob/5e0c3818f1badff1f148325e7c3ff7b53647d4a1/src/Uno.UI/UI/Xaml/WindowManagerInterop.wasm.cs#L760

You can build a method similar to it.

In typescript, you'll need to create methods similar to these:
https://github.com/unoplatform/uno/blob/76f6ce75fad58e676e56c55dfafeafe73c7078d8/src/Uno.UI/ts/WindowManager.ts#L354-L377

But named `setFontFamily`.

This way, the invoking the method will cost less than when used in general by the `SetStyles` method.

jeromelaban:
See original issue for fix.

PR #8389, opened 2022-03-20
feat: `PowerManager` on WASM, refactoring of existing implementations

### LINKED ISSUE #8388: `PowerManager` API on WASM

Support for `PowerManager` on WASM

## Why is this needed:

Feature parity

## For which Platform:

- [ ] iOS
- [ ] Android
- [x] WebAssembly
- [ ] macOS
- [ ] Skia
  - [ ] WPF
  - [ ] GTK (Linux)
  - [ ] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

MartinZikmund:
GitHub Issue (If applicable): closes #8388

## PR Type

- Feature

Not implemented

## What is the new behavior?

Implemented

## PR Checklist

- [ ] **Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)**


trungnt2910:
What's the status of this PR?

I'm looking at adding support for PowerManager on other platforms but I can see here that the implementation is undergoing quite a few changes.

MartinZikmund:
@trungnt2910  Will try to finish this up asap

doc/articles/features/windows-system-power.md
```diff
@@ -15,5 +15,17 @@ uid: Uno.Features.WSPower
...
 - `RemainingDischargeTime` and `RemainingDischargeTimeChanged` events not supported (platform does not offer appropriate APIs)
 
+**WASM**
+
+Before any of the properties/events of `PowerManager` can be accessed, the class needs to be initialized via a platform-specific `InitializeAsync` method:
+
+```csharp
+#if __WASM__
+var isAvailable = await PowerManager.InitializeAsync();
+#endif
```
Youssef1313, line 25: It's unclear where does the user needs to be put this code. In the class library, `#if __WASM__` wouldn't work.

src/Uno.UWP/System/Power/PowerManager.crossruntime.cs
```diff
@@ -0,0 +1,40 @@
...
+			return false;
+		}
+#else
+		if (typeof(PowerManager).Log().IsEnabled(LogLevel.Error))
+		{
+			typeof(PowerManager).Log().LogError("PowerManager is not implemented on this platform", ex);
+		}
+
+		return false;
+#endif
```
Youssef1313, line 38: I think this wouldn't build in non-Wasm platforms.
I would rename this to `PowerManager.wasm.cs`, and then have a non-async implementation for other platforms that returns true/false based on whether it's implemented.

src/Uno.UWP/System/Power/PowerManager.wasm.cs
```diff
@@ -0,0 +1,132 @@
...
+	static partial void EndBatteryStatus() => EndOnChargingChange();
+
+	private static void StartOnChargingChange()
+	{
+		if (_isChargingChangedSubscribed)
+		{
+			return;
+		}
+
+		WebAssemblyRuntime.InvokeJS($"{JsType}.startChargingChange()");
```
Youssef1313, line 32: This shouldn't be the way to go.

src/Uno.UWP/System/Power/PowerManager.wasm.cs
```diff
@@ -0,0 +1,132 @@
...
+		_isChargingChangedSubscribed = false;
+	}
+
+	static partial void StartRemainingChargePercent() => WebAssemblyRuntime.InvokeJS($"{JsType}.startRemainingChargePercentChange()");
+
+	static partial void EndRemainingChargePercent() => WebAssemblyRuntime.InvokeJS($"{JsType}.endRemainingChargePercentChange()");
+
+	static partial void StartRemainingDischargeTime() => WebAssemblyRuntime.InvokeJS($"{JsType}.startRemainingDischargeTimeChange()");
+
+	static partial void EndRemainingDischargedTime() => WebAssemblyRuntime.InvokeJS($"{JsType}.endRemainingDischargeTimeChange()");
```
Youssef1313, line 54: Same here. We shouldn't use InvokeJS

doc/articles/features/windows-system-power.md
Youssef1313: We probably should make this API available in all platforms for consistency, and let the user access it without the need for any `#if` directives.

MartinZikmund: All platforms except Windows I guess 🤔 . `#if HAS_UNO` will still be needed...

MartinZikmund: Which actually makes me think - maybe we should include `__CROSSRUNTIME__` symbol in Uno apps as well (to be able to `#if` WASM and Skia)

src/Uno.UWP/System/Power/PowerManager.wasm.cs
MartinZikmund: Indeed, it is an old PR and I failed to realize that it is now done with introp 😅

src/Uno.UI/ts/Windows/System/Power/PowerManager.ts
```diff
@@ -0,0 +1,125 @@
...
+			}
+
+			return -1;
+		}
+
+		private static onChargingChange() {
+			if (!PowerManager.dispatchChargingChanged) {
+				PowerManager.dispatchChargingChanged =
+					(<any>Module).mono_bind_static_method(
+						"[Uno] Windows.System.Power.PowerManager:DispatchChargingChanged");
```
jeromelaban, line 102: We must not use mono_bind_static_method anymore (it's not CSP compliant), and use `JSExport` instead. See this:

https://github.com/unoplatform/uno/blob/207f109b8ad86554825b5daeaaf6ee23f3a9c8e6/src/Uno.UI/ts/Windows/Gaming/Input/Gamepad.ts#L67L73.

Also, as this particular PR will not be backported, there's no need to keep the `mono_bind_static_method` version.

doc/articles/features/windows-system-power.md
```diff
@@ -6,14 +6,44 @@ uid: Uno.Features.WSPower
...
+#### WebAssembly
+
+Before any of the properties/events of `PowerManager` can be accessed, the class needs to be initialized via a platform-specific `InitializeAsync` method:
 
-**iOS**
-- `RemainingDischargeTime` and `RemainingDischargeTimeChanged` events not supported (platform does not offer appropriate APIs)
+```csharp
+#if __HAS_UNO__
+var isAvailable = await PowerManager.InitializeAsync();
+#endif
```
jeromelaban, line 46: Why do we need to do this? Is it because the internal APIs are async and WinUI APIs are sync?

MartinZikmund: Yes, that is the case - the class is `static` as a whole and the properties can just be read on Windows automatically. Unfortunately, in case of WASM, the initial retrieval is asynchronous (`navigator.getBattery()` is a promise), so the initialization must happen asynchronously before the properties can be used. For simplicity, I made the `InitializeAsync` method available on all Uno targets and it has a `bool` return that tells you whether the `PowerManager` is available or not.

src/SamplesApp/UITests.Shared/Windows_System/Power/PowerManager.xaml.cs
```diff
@@ -1,67 +1,281 @@
...
-using Windows.UI.Xaml.Media;
-using Windows.UI.Xaml.Navigation;
 using UwpPowerManager = Windows.System.Power.PowerManager;
 
-namespace UITests.Shared.Windows_System.Power
+namespace UITests.Shared.Windows_System.Power;
+
+[Sample("Windows.System", Name = "Power.PowerManager",
+	ViewModelType = typeof(PowerManagerTestsViewModel),
+	Description = "Shows properties of Power manager and handles its events")]
```
jeromelaban, line 15: Let's mark this one as a manual test

jeromelaban (approved):

jeromelaban (approved):

PR #8559, opened 2022-04-15
chore: Remove outdated comment

Youssef1313:
Uno has first fixed a bug here, diverging from Avalonia implementation. But later, Avalonia was fixed in the same way. So this comment no longer applies


kazo0 (approved):

ajpinedam (approved):

jeromelaban:
Thanks!

PR #8561, opened 2022-04-17
fix: Fixed bug with Additional Files metadata override.

### LINKED ISSUE #8176: [Generation] AdditionalFiles should not be removed if not added by uno
### Discussed in https://github.com/unoplatform/uno/discussions/8175

<div type='discussions-op-text'>

<sup>Originally posted by **HavenDV** February 23, 2022</sup>
I noticed that the Uno package started blocking the use of third-party generators. In my case, I was generating View code behind constructor code in my test project. 
Is this a bug or is it a special behavior to limit the possible problems of using other generators in most cases? In second case, just need to create a library that will call these generators and use it already in Uno

I assume it is caused by this code
https://github.com/unoplatform/uno/blob/2ed9559ff94e459938d227a4a309c4ba65e17056/src/SourceGenerators/Uno.UI.SourceGenerators/Content/Uno.UI.SourceGenerators.props#L362-L363

P.S. I did not find any changes in the code that would explicitly prohibit the use of other generators. But I still can't achieve using another generator in a project with Uno.UI.WebAssembly. This may be related to dotnet updates to 6.0.2 or VS to 17.2 preview

Repro:
[ratbuddyssey-master.zip](https://github.com/unoplatform/uno/files/8122613/ratbuddyssey-master.zip)
1. Build Ratbuddyssey.Apps.Uno.WebAssembly. The generator will not show the generated file (it should appear if at least one AdditionalFile exists at the time of generation)
image
2. This is not a bug related to the generation order. If you remove Visible="False" then you can see that AdditionalFile exists before the start of build. But it seems then some MSBuild code removes AdditionalFile at one of the stages of preparation.
3. ViewBase code is generated in another library (Ratbuddyssey.Apps.Uno.Core.Standard) and available for Uno generator

</div>

HavenDV:
GitHub Issue (If applicable): closes #8176

## PR Type

Bugfix

At the moment the Uno package does not take into account that the user may have other generators that have metadata in AdditionalFiles

## What is the new behavior?

_InjectAdditionalFiles Target will correctly take into account the presence of AdditionalFiles by comparing them by full path. This will simply add new metadata to existing files that are also Page/ApplicationDefinition/PRIResource/TSBindingAssemblySource without losing the existing metadata.

## PR Checklist

## Other information


src/SourceGenerators/Uno.UI.SourceGenerators/Content/Uno.UI.SourceGenerators.props
```diff
@@ -351,18 +351,31 @@
...
+
+    <PropertyGroup>
+      <_Pages>@(Page)</_Pages>
+      <_ApplicationDefinitions>@(ApplicationDefinition)</_ApplicationDefinitions>
+      <_PRIResources>@(PRIResource)</_PRIResources>
+      <_TSBindingAssemblySources>@(TSBindingAssemblySource)</_TSBindingAssemblySources>
+    </PropertyGroup>
+    
+    <ItemGroup Label="Add SourceItemGroup metadata to existing additional files">
+      <AdditionalFiles SourceItemGroup="Page" Condition="$([System.String]::new($(_Pages)).Contains(%(AdditionalFiles.FullPath)))" />
```
jeromelaban, line 365: I wonder what's the performance impact of this, since it'll recreate a string for all the pages of the app. Can't it be 
```suggestion
      <AdditionalFiles SourceItemGroup="Page" Condition="$(_Pages.Contains(%(AdditionalFiles.FullPath)))" />
```



That would avoid msbuild to re-create the full string for each AdditionalFile already present?

jeromelaban, line 365: I wonder what's the performance impact of this, since it'll recreate a string for all the pages of the app. Can't it be 
```suggestion
      <AdditionalFiles SourceItemGroup="Page" Condition="$(_Pages.Contains(%(AdditionalFiles.FullPath)))" />
```



That would avoid msbuild to re-create the full string for each AdditionalFile already present?

HavenDV:
It's not entirely clear to me what caused the tests to crash after the last change. Everything works correctly in VS 2022 preview 4.0, but the CI environment may have a different msbuild that doesn't support this (or a nested %()). Should I undo the change?

jeromelaban:
This caused by:
```
The expression """.Contains()" cannot be evaluated. Method 'System.String.Contains' not found.
```

Most likely because `_Pages` or others can be null. Either check for empty in the expression, or make it non-empty before getting to the ItemGroup creation,

HavenDV:
> This caused by:
> 
> ```
> The expression """.Contains()" cannot be evaluated. Method 'System.String.Contains' not found.
> ```
> 
> Most likely because `_Pages` or others can be null. Either check for empty in the expression, or make it non-empty before getting to the ItemGroup creation,

In addition to this message, there is also this, so I do not think that this is the reasonэ
```
src\SourceGenerators\Uno.UI.SourceGenerators\Content\Uno.UI.SourceGenerators.props(365,47): Error MSB4184: The expression ""C:\a\1\s\src\Uno.UI.FluentTheme.v2\themeresources_v2.xaml".Contains()" cannot be evaluated. Method 'System.String.Contains' not found.
```

jeromelaban:
> In addition to this message, there is also such a message here, so I do not think that this is the reason
> 
> ```
> src\SourceGenerators\Uno.UI.SourceGenerators\Content\Uno.UI.SourceGenerators.props(365,47): Error MSB4184: The expression ""C:\a\1\s\src\Uno.UI.FluentTheme.v2\themeresources_v2.xaml".Contains()" cannot be evaluated. Method 'System.String.Contains' not found.
> ```

It's likely because the parameter is null or empty, and there's no signature that matches `Contains` without parameters.

HavenDV:
I have updated the code according to the latest changes in the master branch. I also added the necessary checks.

HavenDV:
There are errors in the build that I don't understand. Perhaps they are caused by the fact that the full path is always used to avoid collisions.

HavenDV:
I checked, the error is caused by the fact that data with an absolute path was added to the dictionary instead of a relative one, as it was before. I will try to find an alternative solution instead of changing the current tests.

HavenDV:
I made some edits to make the final output match the current one. Here is a project where you can experiment and evaluate the changes. 
https://github.com/HavenDV/Uno.Issues/tree/main/uno_removes_additional_files_metadata
Ultimately, the problem was not that Uno was removing metadata, but that Uno was adding duplicate values with other metadata, causing other generators that could work with these files to only get the element with Uno's metadata (the second an element with the same FullPath and other metadata was lost)

src/SourceGenerators/Uno.UI.SourceGenerators/Content/Uno.UI.SourceGenerators.props
```diff
@@ -371,27 +371,51 @@
...
+      <_ApplicationDefinitions>@(ApplicationDefinition)</_ApplicationDefinitions>
+      <_PRIResources>@(PRIResource)</_PRIResources>
+      <_TSBindingAssemblySources>@(TSBindingAssemblySource)</_TSBindingAssemblySources>
+      <_UnoPages>@(UnoPage)</_UnoPages>
+      <_UnoApplicationDefinitions>@(UnoApplicationDefinition)</_UnoApplicationDefinitions>
+      <_AdditionalFiles>@(AdditionalFiles)</_AdditionalFiles>
+    </PropertyGroup>
+
+    <ItemGroup Label="Add SourceItemGroup metadata to existing additional files">
+      <AdditionalFiles SourceItemGroup="Page" Condition="$(_UnoPages) == '' AND $(_Pages) != '' AND $(_AdditionalFiles) != '' AND $(_Pages.Contains(%(AdditionalFiles.FullPath)))" />
```
jeromelaban, line 386: Would you be able to add a regression test in the runtimetests csproj (for instance), which create a custom `AdditionalFile` item, then validates that after the build, it's still set properly (with the right metadata)? Adding two targets, one beforebuild and one after `compile` that fails the build, would make it easier to ensure we're not breaking later on. Thanks!

HavenDV: I want to note that the syntax for updating metadata through Condition looks erroneous if you don’t know about it (because you expect an explicit Update, but I couldn’t make code through Update at the time of the PR, although I think it’s possible, I just need something like this filter the collection). Here are some examples that this is a valid code and it works:
https://github.com/MicrosoftDocs/visualstudio-docs/blob/main/docs/msbuild/msbuild-items.md#updating-metadata-on-items-in-an-itemgroup-of-a-target
https://github.com/dotnet/msbuild/issues/1618#issuecomment-275691043

MartinZikmund:
@HavenDV Is this PR still applicable?

HavenDV:
Hi @MartinZikmund, I've resolved the conflict so it may still be relevant/working, but I don't have the free time to test it right now

jeromelaban:
Thanks for the update. It's still causing issues at this time with some combinations in regression testing. Would you know of a known generator that would break when using Uno?

PR #8625, opened 2022-04-26
feat: ValueSet

pkar70:

## PR Type

- Feature

 No ValueSet implementation (only ctor exist).

## What is the new behavior?
 Simple ValueSet implementation exist.

## PR Checklist

## Other information
 As I'm porting apps from UWP to Uno, I want to not loose some functionality (as RemoteSystems). Of course, Uno has no implementation of such things, so this code would never be called by Android, but I want to decrease number of Uno0001 warnings.

 Recreated, as in previous PR, I made one small change, and commit has >5000 files.


src/Uno.Foundation/Collections/ValueSet.cs
```diff
@@ -10,7 +11,40 @@ namespace Windows.Foundation.Collections
...
 	{
-		
+		private Dictionary<string, object> _dictionary;
+
+		public ValueSet()
+		{
+			_dictionary = new Dictionary<string, object>();
+		}
+
+		public void Add(string key, object value) => _dictionary.Add(key, value);
```
jeromelaban, line 21: I wonder, does the windows API accept a null `value` parameter? If so, we will need to alter the API here for the nullable references to work properly.

pkar70: Yes, both `Dictionary` and `ValueSet` allows `Add("dummy", null)`

pkar70:
Why `Uno.Foundation\Generated\2.0.0.0\Windows.Foundation.Collections\ValueSet.cs(8,157): Error CS0535: 'ValueSet' does not implement interface member 'IDictionary<string, object>.ContainsKey(string)' ?

src/Uno.Foundation/Collections/ValueSet.cs
jeromelaban: Ok, then we'll need to change the definition of this type to accept nullable object:

```csharp
public sealed partial class ValueSet : IPropertySet, IObservableMap<string, object?>, IDictionary<string, object?>, IEnumerable<KeyValuePair<string, object?>>
```

And the rest of the definition that comes with it.

pkar70: ok.

pkar70:
And how it can complain that
`src\Uno.Foundation\Generated\2.0.0.0\Windows.Foundation.Collections\ValueSet.cs(115,65): Error CS0102: The type 'ValueSet' already contains a definition for 'Values'`
 when  this line is within `#if false` ?

jeromelaban:
> And how it can complain that `src\Uno.Foundation\Generated\2.0.0.0\Windows.Foundation.Collections\ValueSet.cs(115,65): Error CS0102: The type 'ValueSet' already contains a definition for 'Values'` when this line is within `#if false` ?

I do not see this error in the build, but rather the fact that the Values is not implemented properly because of the nullability. I'd suggest that you implement the whole type, and remove the generated file completely (including the observable events parts).

Also, as we're changing the signature of the interface, we'll need to exclude the type from the sync generator:
https://github.com/unoplatform/uno/blob/da4338ddc48e0204f0badcb32bdd9ba375a91938/src/Uno.UWPSyncGenerator/Generator.cs#L503-L505
as it would otherwise re-generate an invalid interface (and fail on the WinUI packaging side).

I think we also should drop the `#if NET461`, the class does not need to be excluded there, it's fully generic code.

pkar70:
> > And how it can complain that `src\Uno.Foundation\Generated\2.0.0.0\Windows.Foundation.Collections\ValueSet.cs(115,65): Error CS0102: The type 'ValueSet' already contains a definition for 'Values'` when this line is within `#if false` ?
> 
> I do not see this error in the build, but rather the fact that the Values is not implemented properly because of the nullability. I'd suggest that you implement the whole type, and remove the generated file completely (including the observable events parts).

It is in errors from "Uno.UI - CI (Build .NET 6 Build (Windows) WinUI)".
 
> I think we also should drop the `#if NET461`, the class does not need to be excluded there, it's fully generic code.
Ok. `#if NET461` was before my PR, and these two lines was not touched by me.

jeromelaban:
> > > And how it can complain that `src\Uno.Foundation\Generated\2.0.0.0\Windows.Foundation.Collections\ValueSet.cs(115,65): Error CS0102: The type 'ValueSet' already contains a definition for 'Values'` when this line is within `#if false` ?
> > 
> > 
> > I do not see this error in the build, but rather the fact that the Values is not implemented properly because of the nullability. I'd suggest that you implement the whole type, and remove the generated file completely (including the observable events parts).
> 
> It is in errors from "Uno.UI - CI (Build .NET 6 Build (Windows) WinUI)".
> 
> > I think we also should drop the `#if NET461`, the class does not need to be excluded there, it's fully generic code.
> 
> Ok. `#if NET461` was before my PR, and these two lines was not touched by me.

I'm sure it was not you, but I was asking to change it anyways :) I made some additional changes to include the rest, but it'll be required to add unit tests to validate the observable behavior.

pkar70:
Could you be more specific in regard to "it'll be required to add unit tests to validate the observable behavior" ?

jeromelaban:
> Could you be more specific in regard to "it'll be required to add unit tests to validate the observable behavior" ?

We need to have unit test that validate the behavior of `MapChanged` and everything else in that class.

pkar70:
And I only want to have four less warnings in building my apps... :)

jeromelaban:
> And I only want to have four less warnings in building my apps... :)

Sure. If you want to implement a feature in Uno, this is still what is needed.

pkar70:
can build be restarted?

jeromelaban:
The builds restart automatically up to three times, delayed depending on the CI availability.

src/Uno.UI.RuntimeTests/Tests/Windows_Foundation/Collections/Given_ValueSet.cs
```diff
@@ -0,0 +1,109 @@
...
+		}
+
+		[TestMethod]
+		public async void When_ModificationEvent()
+		{
+
+			// .Add
+			_wasEventFired = false;
+			_testSet.Add("thisisKey", "andValue");
+			await System.Threading.Tasks.Task.Delay(100);
```
jeromelaban, line 42: Why did you need to add an arbitrary delay here?

pkar70: Because we want to check if _wasEventFired is changed by handler, and handler needs time to fire and change _wasEventFired variable.

src/Uno.Foundation/Collections/MapChangedEventArgs.cs
```diff
@@ -0,0 +1,4 @@
+﻿namespace Windows.Foundation.Collections
```
MartinZikmund, line 1: Use file-scoped namespaces here

src/Uno.Foundation/Collections/IMapChangedEventArgs1.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Windows.Foundation.Collections
```
MartinZikmund, line 1: Use file-scoped namespaces here

src/Uno.UI.RuntimeTests/Tests/Windows_Foundation/Collections/Given_ValueSet.cs
MartinZikmund: As the event is executed synchronously, it should be possible to remove all the awaits in this test

src/Uno.UWPSyncGenerator/Generator.cs
```diff
@@ -504,6 +504,10 @@ protected bool SkippedType(INamedTypeSymbol type)
 					// Skipped to include generic variance.
 					return true;
 
+				case "Windows.Foundation.Collections.ValueSet":
```
MartinZikmund, line 507: This is actually a problem with sync generator, it should ignore the nullables, right @jeromelaban ? Maybe worth creating an issue to remove this after it is merged and to fix it in generator.

MartinZikmund (approved):

src/Uno.UWPSyncGenerator/Generator.cs
pkar70: Seems like this comment is not directly related to this PR, so I think I can "resolve conversation".

pkar70:
Can build be restarted?

jeromelaban (approved):

Issue #8643, opened 2022-04-28
[Skia] OpenGL render surface fails with a null reference on WSL


jeromelaban:
### Current behavior

```
Marshaling render signal
GLIB UNHANDLED EXCEPTIONSystem.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at Uno.UI.Runtime.Skia.GLRenderSurface.UnoGLDrawingArea_Render(Object o, RenderArgs args)
   --- End of inner exception stack trace ---
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor, Boolean wrapExceptions)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Delegate.DynamicInvokeImpl(Object[] args)
   at System.Delegate.DynamicInvoke(Object[] args)
   at GLib.Signal.ClosureInvokedCB(Object o, ClosureInvokedArgs args)
```

### Expected behavior

OpenGL rendering works properly

### Environment

Uno.UI / Uno.UI.WebAssembly / Uno.UI.Skia

### NuGet package version(s)

4.2.6

### IDE

Visual Studio 2022

microhobby:
Maybe related to https://github.com/microsoft/wslg/issues/715

jeromelaban:
The null reference happens because there's no OpenGL context when running under Wayland (it's OpenGL ES). 

This line returns null, and causes the rest to fail:
https://github.com/unoplatform/uno/blob/6d97c470f88327496e309cf2823e65dff25b7eb8/src/Uno.UI.Runtime.Skia.Gtk/GLRenderSurface.cs#L82

The suggestion from https://github.com/unoplatform/uno/issues/8660 to set `GDK_BACKEND=x11` as an environment variable does help on WSL and bare Linux but detecting the required conditions may be tricky.

Another option is to use OpenGL ES, but this does not solve the support for WSL, because the hardware abstraction for WSL does not support a high enough version so that Skia can use it. Here's an error that arises when using GLES:

```
Shader compilation error
------------------------
   1	#version 320 es
   2	
   3	precision mediump float;
   4	precision mediump sampler2D;
   5	out mediump vec4 sk_FragColor;
   6	flat in mediump vec4 vcolor_Stage0;
[...]

PR #8591, opened 2022-04-22
fix(skia): SKImageInfo.PlatformColorType is incorrect on macOS/arm64

spouliot:
- Detect macOS + arm64 and reset default `SKColorType` to `Bgra8888`
- Create the `SKBitmap` using the `SKImageInfo` instance

Addresses #8580

GitHub Issue (If applicable): closes #8580

## PR Type

What kind of change does this PR introduce?

- Bugfix

## What is the current behavior?

When using Skia.Gtk on a M1-based Mac computer the red and blue channels are inverted.

## What is the new behavior?

`SKColorType.Bgra8888` is used in the software renderer `RenderSurfaceType.Software` for macOS/arm64.

## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)
- [ ] [Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [ ] Validated PR `Screenshots Compare Test Run` results.
- [ ] Contains **NO** breaking changes
- [ ] Associated with an issue (GitHub or internal) and uses the [automatic close keywords](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue).
- [ ] Commits must be following the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) specification.

## Other information

Internal Issue (If applicable):

PR #8698, opened 2022-05-05
WIP: feat(gtk): Add OpenGL ES support

microhobby:

https://github.com/unoplatform/uno/issues/8643

## PR Type

- Feature

OpenGL hardware acceleration does not work in wayland with GDK_GL gles

## What is the new behavior?

OpenGL hardware acceleration working with wayland GDL_GL gles

## PR Checklist

## Other information

https://github.com/unoplatform/uno/issues/8643


microhobby:
Testing here I also noticed that the colors are kind of weird, the red is blue 🥲 ...

MartinZikmund:
Is it possible to detect which renderer to use by default (unless explicitly overriden by the user)?

jeromelaban:
Thanks for the PR! I had a similar support working locally but I think adding an additional render surface is not needed. We'll be able to have all of it inside the existing OpenGL renderer class and have it use OpenGL or ES based on platform availability. I'll merge what I have with the bits you added, and we'll make some tests around it.

Does your change work with WSL? Or did you try it with Linux only?

src/Uno.UI.Runtime.Skia.Gtk/GLESRenderSurface.cs
```diff
@@ -0,0 +1,197 @@
...
+using Gtk;
+using Silk.NET.OpenGLES;
+using Silk.NET.Core.Loader;
+
+namespace Uno.UI.Runtime.Skia
+{
+
+	internal class GLESRenderSurface : GLArea, IRenderSurface
+	{
+		private const SKColorType colorType = SKColorType.Rgba8888;
```
spouliot, line 28: > Testing here I also noticed that the colors are kind of weird, the red is blue 🥲 ...

This looks like the surface is `Bgra8888` (and not `Rgba8888`). See https://github.com/unoplatform/uno/pull/8591 for a similar case

jeromelaban:
@microhobby I could not get the GL ES to work on WSL as I found previously, so I've updated the PR to enable falling back from GL ES to GL, then to software if nothing else is available. Can you let me know if this still works on your system?

src/Uno.UI.Runtime.Skia.Gtk/GLRenderSurfaceBase.cs
```diff
@@ -133,9 +122,17 @@ private void UnoGLDrawingArea_Render(object o, RenderArgs args)
...
 
-			_gl.Flush();
+			GLFlush();
 		}
 
+		protected abstract void GLClear();
+
+		protected abstract void GLFlush();
+
+		protected abstract (int framebuffer, int stencil, int samples)GetGLBuffers();
```
spouliot, line 132: It seems that all 3 methods implementations are identical* for both GL and GLES subclasses. 

Considering virtual calls are often (if not devirtualized) slower than normal calls I think it might be best to have the base class make the calls directly. 

`*` If the situation ever change (any implementation become different) then the override(s) can be introduced at that time.

jeromelaban: Those are unfortunately not identical, as the namespace is different. I agree that the virtual call will be impacting a bit here. We could do the switching in-class, at the expense of a bit of readability.

microhobby:
@jeromelaban it's working (in WSL 2) if I set the `GDK_GL=gles` env variable and using the `host.RenderSurfaceType = RenderSurfaceType.OpenGLES;`

About the colors I'm checking here https://gitlab.gnome.org/GNOME/gtk/-/issues/303 as the `gtk3-demo` as we have the same behavior when using the OpenGL ES, it seams to be a GTK related issue ...

jeromelaban:
> @jeromelaban it's working (in WSL 2) if I set the `GDK_GL=gles` env variable and using the `host.RenderSurfaceType = RenderSurfaceType.OpenGLES;`

Ah! This is very good to know. This is some additional validation that can be done in the class to enable GL support.

jeromelaban:
> About the colors I'm checking here https://gitlab.gnome.org/GNOME/gtk/-/issues/303 as the `gtk3-demo` has the same behavior when using the OpenGL ES, it seams to be a GTK related issue ...
> 
> >

About this what happens if you change the pixel format, does it help? We could probably handle that specifically when not running under WSL (the colors are not inverted there).

jeromelaban:
I'll merge this PR as-is as it fixes a lot of other issues, but let's keep track of inverted colors in https://github.com/unoplatform/uno/issues/8766.

jeromelaban (approved):

PR #8824, opened 2022-05-19
perf(selectoritem): Use monotonic clock instead of DateTime.Now

ebariche:


MartinZikmund:
Nice 🙂

src/Uno.UI/UI/Xaml/Controls/Primitives/SelectorItem.cs
```diff
@@ -186,7 +187,7 @@ private void UpdateCommonStates(ManipulationUpdateKind manipulationUpdate = Mani
 			{
 				CoreDispatcher.Main.RunAsync(CoreDispatcherPriority.Normal, async () =>
 				{
-					await Task.Delay(delay);
+					await Task.Delay(ToTimeSpan(delay));
```
workgroupengineering, line 190: you can using [Task.Delay(int millisecondsDelay)](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.delay?view=net-6.0#system-threading-tasks-task-delay(system-int32)) to avoid convert to TimeSpan

jeromelaban: In this case, it's not possible because of the time resolution.

src/Uno.UI/UI/Xaml/Controls/Primitives/SelectorItem.cs
```diff
@@ -65,7 +66,7 @@ private enum ManipulationUpdateKind
 
 		private string _currentState;
 		private uint _goToStateRequest;
-		private DateTime _pauseStateUpdateUntil;
+		private long _pauseStateUpdateUntil;
```
jeromelaban, line 69: Let's use a TimeSpan instead, it makes it easier to read, and there's no performance penalty.

jeromelaban: Does it need to be absolute? Do we expose that value anywhere?

dr1rrb: No, could be relative to "something", but we do need to change the logic then and the name of the field :)

jeromelaban (approved):
🚀

PR #8960, opened 2022-06-02
perf(drawables): Generate resolver to avoid reflection at startup

ebariche:


src/Uno.UWP/Helpers/DrawableHelper.Android.cs
```diff
@@ -76,6 +90,8 @@ public static Drawable FromUri(Uri uri)
 			return drawable;
 		}
 
+		public static void SetDrawableResolver(Func<string, int> resolver) => _resolver = resolver;
```
jeromelaban, line 93:
```suggestion
		/// <summary>
        /// Sets a function to resolve drawable Ids from their string equivalent, generally used internally by Uno using a source generator.
        /// </summary>
		public static void SetDrawableResolver(Func<string, int> resolver) => _resolver = resolver;
```

jeromelaban (approved):

PR #9072, opened 2022-06-17
perf(macOS): Avoid calling `NSView.Layer` getter more than once / method

spouliot:

## PR Type

The `NSView.Layer` property getter is often called more than once inside methods.

### Callers
* `UIElement.ApplyNativeClip` (75%)
* `BindableNSView..ctor()` (9.7%)
* `NativeRenderTransformAdapter.CompleteInitialization` (7.2%)
* `FrameworkElement.set_Frame` (6.2%)
* `NativeRenderTransformAdapter.InitializeOrigin` (0.61%)
* `NativeRenderTransformAdapter.Apply` (0.54%)
* `NativeRenderTransformAdapter.Initialized` (0.40%)

All `get_Layer` calls, inside the same method, will return the exact same instance - but the extra native call(s) and even more the managed peer lookup (`GetNativeObject<T>`) are costly and totally avoidable.

Screen Shot 2022-06-17 at 10 50 56 AM

## What is the new behavior?

The `NSView.Layer` property getter is only called once per method - at least the ones identified inside Dope's flamegraph.

This changed the distribution of the callers

* `UIElement.ApplyNativeClip` 75% -> 44%
* `BindableNSView..ctor()` 9.7% -> 15%
* `NativeRenderTransformAdapter.CompleteInitialization` 7.2% -> 6.0%
* `FrameworkElement.set_Frame` 6.2% -> 5.0%
* `NativeRenderTransformAdapter.InitializeOrigin` 0.61% -> 21%
* `NativeRenderTransformAdapter.Apply` 0.54% -> 6.7%
* `NativeRenderTransformAdapter.Initialized` 0.40% -> 3.3%

which means clipping is less a bottleneck that it was before.

Screen Shot 2022-06-17 at 10 51 19 AM

note: Dope being a fixed time benchmark you need to look at relative (percentage) values, not the exact amount of time.

After the change 92% of `get_Layer` time is spent inside unmanaged code, while `GetNativeObject<T>` is a 8.2%.

Before that `GetNativeObject<T>` was taking 81% of the time and unmanaged code was 19%.

## PR Checklist

### Current

```csharp
if (Layer != null)
{
    Layer.MasksToBounds = false;
}
```

The above C# code becomes the following IL code

```il
// if (Layer != null)
IL_002b: ldarg.0
IL_002c: callvirt instance class [Microsoft.macOS]CoreAnimation.CALayer [Microsoft.macOS]AppKit.NSView::get_Layer()
IL_0031: brfalse.s IL_003f

// Layer.MasksToBounds = false;
IL_0033: ldarg.0
IL_0034: callvirt instance class [Microsoft.macOS]CoreAnimation.CALayer [Microsoft.macOS]AppKit.NSView::get_Layer()
IL_0039: ldc.i4.0
IL_003a: callvirt instance void [Microsoft.macOS]CoreAnimation.CALayer::set_MasksToBounds(bool)
```

which shows the two calls to `get_Layer`. This is a performance issue since 
* that native/ObjC call is slow; and
* the lookup (`GetNativeObject<T>`) is also slow

This also fix (albeit very unlikely) a potential `NullReferenceException`, IOW [ToCToU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use), since something else could have executed between the calls and set it the layer to `null` (should not happen since almost everything runs from the main UI thread but CoreAnimation can be surprising).

```csharp
var layer = Layer;
if (layer != null)
{
    layer.MasksToBounds = false;
}
```

only calls the getter once

```il
// CALayer layer = Layer;
IL_002b: ldarg.0
IL_002c: callvirt instance class [Xamarin.Mac]CoreAnimation.CALayer [Xamarin.Mac]AppKit.NSView::get_Layer()
IL_0031: stloc.0
// if (layer != null)
IL_0032: ldloc.0
IL_0033: brfalse.s IL_003c

// layer.MasksToBounds = false;
IL_0035: ldloc.0
IL_0036: ldc.i4.0
IL_0037: callvirt instance void [Xamarin.Mac]CoreAnimation.CALayer::set_MasksToBounds(bool)
```

This change is within the noise for Dope (2 seconds out of 120 seconds is below 5%). However the `dotnet trace` shows the percentages of time spent on `GetNSObject<T>` has been reduced.

A similar PR might be needed for iOS since most of the modified code is not shared.


src/Uno.UI/UI/Xaml/UIElement.macOS.cs
```diff
@@ -262,18 +262,16 @@ private protected override void OnNativeKeyUp(NSEvent evt)
 			{
 				if (!ClippingIsSetByCornerRadius)
 				{
-					if (Layer != null)
-					{
-						this.Layer.Mask = null;
-					}
+					Layer.Mask = null;
```
jeromelaban, line 265: As far as I remember, Layer could be null if not asked for, which is very different from iOS. Has that changed?

spouliot: no, you're 💯% right 😄. The mention

> This property is never nil. 
https://developer.apple.com/documentation/uikit/uiview/1622436-layer?language=objc

only applies to iOS, where it's `readonly`, and not macOS where something else could set if to `nil` at anytime. I'll fix the PR asap

MartinZikmund (approved):
Beautiful!

jeromelaban (approved):

spouliot:
@jeromelaban looks like the two last steps have gone awol... can you `/azp run` again ? thanks!

MartinZikmund:
@spouliot the WIP step and conventional commit validation got stuck. I usually force-push to force it to re-run, can you try?

PR #9073, opened 2022-06-17
fix(iOS): Fire the native SizeChanged event upon init of Window

kazo0:
fixes https://github.com/unoplatform/nventive-private/issues/402

## PR Type

- Bugfix

Attempting to retrieve the Window.Bounds too early results in a Rect of 0x0

## What is the new behavior?

Initialization of the Windows.UI.XamlWindow fires a SizeChanged event to ensure applications will not recieve an empty Rect when trying to access the Window.Bounds


src/Uno.UI/Extensions/ViewHelper.iOSmacOS.cs
```diff
@@ -231,11 +231,12 @@ public static nfloat StackSubViews(_View thisView, float topPadding, float space
 		/// Gets the orientation-dependent screen size
 		/// </summary>
 		/// <returns></returns>
-		public static CGSize GetScreenSize()
+		public static CGSize GetScreenSize(Window window = null)
 		{
 #if __IOS__
-			var width = Window.Current.NativeWindow.Frame.Width;
-			var height = Window.Current.NativeWindow.Frame.Height;
+			var nativeWindow = window?.NativeWindow ?? Window.Current.NativeWindow;
```
jeromelaban, line 237: We should be validating if `Window.Current.NativeWindow` is null as well.

jeromelaban (approved):

MartinZikmund (approved):

MartinZikmund:
One potential issue is that this is a breaking change:

```
Removed method CoreGraphics.CGSize Uno.UI.ViewHelper.GetScreenSize() not found in ignore set.
```

It may be possible that this is used by some nventive libraries? @jeromelaban In such case we could create a separate `internal` overload instead and call it from the parameterless version with `Window.Current` as parameter.

kazo0:
> One potential issue is that this is a breaking change:
> 
> ```
> Removed method CoreGraphics.CGSize Uno.UI.ViewHelper.GetScreenSize() not found in ignore set.
> ```
> 
> It may be possible that this is used by some nventive libraries? @jeromelaban In such case we could create a separate `internal` overload instead and call it from the parameterless version with `Window.Current` as parameter.

Ah damn I didn't realize the ViewHelper was public

MartinZikmund:
@kazo0 Surprised me too when I was changing it 😀

spouliot (approved):

PR #9114, opened 2022-06-23
perf(macOS): Cache DisplayInformation properties

### LINKED ISSUE #9098: [macOS][iOS] Proposal to cache `[NS|UI]Screen.MainScreen` derived data
### Current behavior

`DisplayInformation` is doing too many native (ObjC) calls on values that rarely changes. Here's a sample of the existing code:

```csharp
public uint ScreenHeightInRawPixels
{
	get
	{
		var screenSize = NSScreen.MainScreen.Frame.Size;
		var scale = NSScreen.MainScreen.BackingScaleFactor;
		return (uint)(screenSize.Height * scale);
	}
}
```

That's four native (ObjC) calls to return the `Height`.

The methods could be optimized, individually, to cache `[NSScreen mainScreen]` saving one native (ObjC) call for each invocation, example:

```csharp
public uint ScreenHeightInRawPixels
{
	get
	{
		var screen = NSScreen.MainScreen;
		return (uint)screen.ConvertRectToBacking(screen.Frame).Height;
	}
}
```

That's _only_ three ObjC calls (and avoids the extra step of multiplication). 

Still the following, unavoidable, code pattern remain very inefficient:

```csharp
var width = di.ScreenHeightInRawPixels;
var height = di.ScreenWidthInRawPixels;
```

since it requires to recompute the same data twice, requiring issuing _again_ the same native calls (twice three ObjC calls).

If this code ends up being called inside a loop (issue) then performance can suffer a lot.  The `get_Scale` inside ["issue"](https://github.com/unoplatform/uno/issues/8597) looks like such an occurrence.

### Expected behavior

Querying `DisplayInformation` properties should be _fast_ and not appear as bottleneck inside benchmarks.

IMHO we can assume that

* Screen changes are infrequent (there's often only one screen for the lifetime of the app);
* A small delay (e.g. 500ms) to detect a screen change wont cause visual issues [1];
* The check for the delay is much smaller than the cost of all the native calls involved;

[1] https://github.com/unoplatform/uno/issues/9027 (about scale) was found reading the code, not reported as an issue (as far as I know).

Create a loop over some properties of `DisplayInformation` and print the time it takes to go over it.

### Workaround

Applications can cache the values.

### Works on UWP/WinUI

_No response_

### Environment

### NuGet package version(s)

_No response_

### Affected platforms

iOS, macOS

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

This affect fixing https://github.com/unoplatform/uno/issues/9027 since it means replacing  the use of **very fast** `readonly` fields with **very slow** calls to `DisplayInformation` properties.

Once the replacement is done the `MainScreen` becomes a large bottleneck inside the Dope benchmark (16 seconds out of the 120 seconds run).

Screen Shot 2022-06-22 at 8 36 30 PM

## Proposal

* Calculate all `[NS|UI]Screen`-based values in a single method
* Call that method if more than 500ms (for example) has gone by since last time it was called
* Change properties to return the cached values

spouliot:
GitHub Issue (If applicable): closes #9098

## PR Type

Calling `DisplayInformation` properties requires several, slow native (ObjC) calls.

## What is the new behavior?

Subscribe to notifications to let us know the `NSWindow` (macOS) or `UIScreen` (macOS) have changed screen (or resolution) so we can cache the values between calls.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UWP/Graphics/Display/DisplayInformation.iOS.cs
+++ b/src/Uno.UWP/Graphics/Display/DisplayInformation.iOS.cs
@@ -1,5 +1,4 @@
-﻿#if __IOS__
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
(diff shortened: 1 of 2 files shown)


spouliot:
Draft until the iOS version is added.

jeromelaban (approved):

PR #9115, opened 2022-06-23
feat: Initial support for XAML resources trimming on Skia NET 6

ebariche:


src/Uno.UI.FluentTheme.v1/AssemblyInfo.cs
```diff
@@ -8,13 +8,14 @@
 [assembly: InternalsVisibleTo("Uno.Wasm")]
 [assembly: InternalsVisibleTo("Uno.UI.Tests")]
 
-#if NET6_0_OR_GREATER
-[assembly: System.Reflection.AssemblyMetadata("IsTrimmable", "True")]
-#elif __IOS__
+[assembly: AssemblyMetadata("IsTrimmable", "True")]
```
spouliot, line 11: This attribute is ignored by the linker in the legacy SDK (before net6)
IOW it does not matter (much) if it stays under `NET6_0_OR_GREATER` or not

src/Uno.UI.Toolkit/AssemblyInfo.cs
```diff
@@ -1,4 +1,7 @@
-﻿using System.Runtime.CompilerServices;
+﻿using System.Reflection;
+using System.Runtime.CompilerServices;
+
+[assembly: AssemblyMetadata("IsTrimmable", "True")]
```
spouliot, line 4: Only net6+ will trim this assembly.
IOW The pre-net6 linker won't process this assembly.
Is that ok ?

src/Uno.UI.FluentTheme.v1/AssemblyInfo.cs
jeromelaban: We need this to be enabled for the netstandard2.0 target (for skia+gtk), this is why it's forcibly added for all targets.

jeromelaban (approved):

Issue #8339, opened 2022-03-15
Breaking changes rollup


MartinZikmund:
- [ ] `ElevatedView`'s `CornerRadius` property is already declared in base `Control` - it should be enough to observe its changes

## For which Platform:

All

Youssef1313:
Consider removing this in the next major release:

https://github.com/unoplatform/uno/blob/4666c66d75ea47c9156ea79c5d1e83465fdf570d/src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlCodeGeneration.cs#L871-L873

Youssef1313:
Also https://github.com/unoplatform/uno/blob/master/src/Uno.Foundation/HResult.cs

I think it should be removed. It's unused, and https://docs.microsoft.com/en-us/uwp/api/windows.foundation.hresult?view=winrt-22621 says:

> When programming with .NET, this class is hidden

Youssef1313:
It might also be a good idea to expose `HtmlImage` publicly

https://github.com/unoplatform/uno/blob/88d7a4024ccdcaa07a6b8cd056ba55ef4e92865e/src/Uno.UI/UI/Xaml/Controls/Image/Image.wasm.cs#L17-L22

It should be used only by Uno, I think

Youssef1313:
Also https://github.com/unoplatform/uno/blob/88d7a4024ccdcaa07a6b8cd056ba55ef4e92865e/src/Uno.UI/UI/Xaml/Controls/Control/Control.cs#L997-L998

Youssef1313:
`CornerRadius` and `CornerRadiusProperty` should be removed from `ElevatedView`. See https://github.com/unoplatform/uno/pull/10138#discussion_r996273357

Youssef1313:
@MartinZikmund I assume these will go into Uno 5 (next major release)? If so, should we have a branch to work on these earlier?

MartinZikmund:
> @MartinZikmund I assume these will go into Uno 5 (next major release)? If so, should we have a branch to work on these earlier?

I think that is a good idea, we could prep them in this branch https://github.com/unoplatform/uno/tree/dev/5.0-breaking-changes

jeromelaban:
For breaking changes let's use a branch named `feature/breaking-changes` instead, the CI is already setup properly.

MartinZikmund:
# Repro generator

This issue was classified as not reproducible with a code sample or test.

**Reason:** Breaking changes tracking epic.

PR #9126, opened 2022-06-24
fix(iOSmacOS): Obsolete the use of readonly fields

### LINKED ISSUE #9027: [macos][ios] Issues with ViewHelper
### Current behavior

This code does not always work as expected:

```csharp
#if __IOS__
		public static readonly nfloat MainScreenScale = UIScreen.MainScreen.Scale;
		public static readonly bool IsRetinaDisplay = UIScreen.MainScreen.Scale > 1.0f;
#elif __MACOS__
		public static readonly nfloat MainScreenScale = NSScreen.MainScreen.UserSpaceScaleFactor;
		public static readonly bool IsRetinaDisplay = NSScreen.MainScreen.UserSpaceScaleFactor > 1.0f;
#endif
```

Both fields are evaluated and assigned only when the (static) constructor is executed.
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/readonly

Which means they will always match the **then** current (at execution time) screen/monitor. 

IOW if the app starts on a 4K monitor and is moved to an HD monitor the (cached) values will still reflect the 4K monitor.

Note: this is also a problem for iOS / iPadOS where multiple screens are becoming more commons.

### Expected behavior

Calling the fields should return the correct values for the currently used screen/monitor (at time of the call).

_No response_

### Workaround

Avoid using the fields and call the API.

### Works on UWP/WinUI

_No response_

### Environment

### NuGet package version(s)

_No response_

### Affected platforms

iOS, macOS

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

Since the fields are public removing them would be a **breaking change**.

However they will never work as someone would expect them since they can only be assigned inside the constructor.

## Proposal

1. Decorate the readonly fields with `[Obsolete]`. This avoids the breaking changes and will notify developers to use something else;

2. Provide new properties (not fields) that can be used to replace both fields;

3. Ensure the old fields are not used inside any calculation inside Uno;

spouliot:
GitHub Issue (If applicable): closes #9027

## PR Type

- Bugfix

`ViewHelper` has C# `readonly` fields for `MainScreenScale` and `IsRetinaDisplay` that are evaluated only once. IOW they are set to the current screen when the evaluation is done and moving the window to a different screen will not update the values.

## What is the new behavior?

Both `readonly` fields are now decorated with `[Obsolete]` attributes giving update instructions to directly use (now cached) `DisplayInformation` properties. This ensure the current (screen) values will be given for every call during the app lifetime.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/Extensions/ViewHelper.iOSmacOS.cs
+++ b/src/Uno.UI/Extensions/ViewHelper.iOSmacOS.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.Drawing;
 using System.Runtime.CompilerServices;
 using Uno.UI.Extensions;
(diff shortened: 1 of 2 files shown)


src/Uno.UI/Extensions/ViewHelper.iOSmacOS.cs
```diff
@@ -27,15 +28,19 @@ namespace Uno.UI
 	public static class ViewHelper
 	{
 #if __IOS__
+		[Obsolete("This return the value from the original screen. Use 'DisplayInformation.RawPixelsPerViewPixel' to get the value for the current screen.")]
```
jeromelaban, line 31: We generally don't mark fields as obsolete because it's breaking for warning-as-error-enabled builds. Let's mark those fields are non-browsable for now, and we can introduce a breaking change in Uno 5 (https://github.com/unoplatform/uno/issues/8339).

jeromelaban (approved):
Nice ! Let's validate the screenshot comparison results as well.

jeromelaban:
Screenshots look good!

PR #9179, opened 2022-06-30
fix(iOSmacOS): Correct computation for LogicalToPhysicalPixels

### LINKED ISSUE #9131: [iOS][macOS] Incorrect computation for `LogicalToPhysicalPixels`
### Current behavior

Given a `CGRect` of `{{X=451.5,Y=1865,Width=182,Height=30}}`

The current code returns `{{X=451,Y=1865,Width=182,Height=30}}`

Which I, like `CGRectIntegral`, believe to be incorrect 😄 

While reducing `X` to `451` is correct, it requires increasing the `Width` to cover for the half-pixel decrease of `X`.

That _temporarily_ makes `Width == 182.5` and, _integrally_, it should become `183` to fully cover the rectangle.

### Expected behavior

Given a `CGRect` of `{{X=451.5,Y=1865,Width=182,Height=30}}`

`CGRectIntegral` [1] returns `{{X=451,Y=1865,Width=183,Height=30}}`

_No response_

### Workaround

_No response_

### Works on UWP/WinUI

_No response_

### Environment

_No response_

### NuGet package version(s)

_No response_

### Affected platforms

_No response_

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

Was there a reason not to use the `CGRectIntegral` [1] API ? Scaling ?

I have not (yet) done any benchmark to see if the managed version is faster (which might be another reason). The current implementation does not show up inside Dope (even if used) so it's not AFAIK a performance (but a correctness) issue.

[1] Inside Xamarin SDK it exists as an extension method of `CGRect` called `Integral`.

spouliot:
GitHub Issue (If applicable): closes #9131

## PR Type

- Bugfix

The size of the `CGRect`, `Width` and `Height`, were used like a point in the calculations.

This means a correct to `X` or `Y` was not reflected in the `Width` and `Height` properties.

Example:

Input: `{{X=451.5,Y=1865,Width=182,Height=30}}`
Output: `{{X=451,Y=1865,Width=182,Height=30}}`

Anything in the last half pixel (width) is _lost_ (as not part of the returned `CGRect`).

## What is the new behavior?

Calculations are done on two points (X1,Y1) and (X2,Y2). The a `CGRect` is created from those two points - IOW Width and Height are recalculated.

Example:

Input: `{{X=451.5,Y=1865,Width=182,Height=30}}`
Output: `{{X=451,Y=1865,Width=183,Height=30}}`

A width of 183 encompass all the input rect and returns an integral rectangle.

Also
* when the scale is 1.0 (default) then avoid multiplications and divisions
* since pixels are always positive we can re-use the `Truncate` results instead of extraneous `Floor` calls

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Media/Given_RevealBrush.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Media/Given_RevealBrush.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Windows.Graphics.Display;
 using Windows.UI;
 using Windows.UI.Xaml.Controls;
 using Windows.UI.Xaml.Media;
@@ -42,7 +43,13 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Media
 			await WindowHelper.WaitForEqual(57, () => gridWithRevealBackground.ActualWidth);
 			Assert.AreEqual(34, gridWithRevealBackground.ActualHeight);
 
-			Assert.AreEqual(43, siblingBorder.ActualWidth);
+			// If `siblingBorder.X` is fractional (not always visible inside `Frame`) then the `Width` will be
+			// Different screen size of high-res (scale > 1) devices can trigger this (fractional `X`) condition
+			// along where the canvas start position is (relative to the unit test runner)
+			// See https://github.com/unoplatform/uno/pull/9179 for more details
+			var delta = DisplayInformation.GetForCurrentView().RawPixelsPerViewPixel == 2 ? 0.5d : 0.0d;
+
+			Assert.AreEqual(43, siblingBorder.ActualWidth, delta);
 			Assert.AreEqual(22, siblingBorder.ActualHeight);
 		}
 	}
--- a/src/Uno.UI/Extensions/ViewHelper.iOSmacOS.cs
+++ b/src/Uno.UI/Extensions/ViewHelper.iOSmacOS.cs
@@ -44,21 +44,10 @@ namespace Uno.UI
 		public static readonly bool IsRetinaDisplay = MainScreenScale > 1.0f;
 #endif
 
-		private static double _rectangleRoundingEpsilon = 0.05;
-		private static double _scaledRectangleRoundingEpsilon = _rectangleRoundingEpsilon * DisplayInformation.GetForCurrentView().RawPixelsPerViewPixel;
-
 		/// <summary>
 		/// This is used to correct some errors when using Floor and Ceiling in LogicalToPhysicalPixels for CGRect.
 		/// </summary>
-		public static double RectangleRoundingEpsilon
-		{
-			get { return _rectangleRoundingEpsilon; }
-			set
-			{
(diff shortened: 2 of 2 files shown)


spouliot:
There's a test failure with the PR.

```
1) Failed : SamplesApp.UITests.Runtime.RuntimeTests.RunRuntimeTests
1 unit test(s) failed (count=1).
	Failing Tests:
	1. When_RevealBrush_Assigned()


---
	Details:
Failed: When_RevealBrush_Assigned() [Microsoft.VisualStudio.TestTools.UnitTesting.AssertFailedException] 
 Microsoft.VisualStudio.TestTools.UnitTesting.AssertFailedException: Assert.AreEqual failed. Expected:<43>. Actual:<43.5>. 
```

I had problem to duplicate it locally. Turns out the build bots uses `iPad Pro (12.9-inch) (4th generation)` to run the tests (and that fails) unlike the default (VSfM) being an `iPhone SE (3rd generation)` (that does **not** fail).

The problematic value is `{{X=228.25,Y=34,Width=43,Height=22}}`. That `.25` on `X` means that a `Width` of `43` is not enough to cover the original rectangle area once `X` is reduced to `228`.

This PR returns `43.5` (fine because scale == 2) so that the entire original area is covered by the new values. However the unit tests is not happy...

Right now (need more investigation) I think my change is right - and that other platforms are _simply_ not failing since `X` is different (and not checked by the unit test).

spouliot:
Resizing the area between the unit test runner and the canvas is enough to trigger a success (or a failure) since it moves the origin of the shapes (right side).

iPhone working value `{{X=68,Y=34,Width=43,Height=22}}`
iPhone not-working value `{{X=55.5,Y=34,Width=43.5,Height=22}}`

The default placement (value) of that bar for the iPhone (working by default) and iPad (not working by default) are why the test works (or not) on different simulators.

Best thing that comes to mind is to adjust the test, along with comments. Something like:

```diff
diff --git a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Media/Given_RevealBrush.cs b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Media/Given_RevealBrush.cs
index b5ec75b73..d0b5e06dd 100644
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Media/Given_RevealBrush.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Media/Given_RevealBrush.cs
@@ -42,7 +42,12 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Media
                        await WindowHelper.WaitForEqual(57, () => gridWithRevealBackground.ActualWidth);
                        Assert.AreEqual(34, gridWithRevealBackground.ActualHeight);

-                       Assert.AreEqual(43, siblingBorder.ActualWidth);
+                       var x = siblingBorder.Frame.X;
+                       // if `X` is fractional then the `Width` will be, see https://github.com/unoplatform/uno/pull/9179
+                       // different screen size (e.g. iPhone vs iPad) of highres devices (scale > 1) can trigger this
+                       var w = (Math.Truncate(x) == x) ? 43.0 : 43.5;
+
+                       Assert.AreEqual(w, siblingBorder.ActualWidth);
                        Assert.AreEqual(22, siblingBorder.ActualHeight);
                }
        }
```

Ideas ? Comments ?

carldebilly (approved):

PR #9223, opened 2022-07-06
fix(PasswordBox): Avoid the content of PasswordBox being cleared on iOS

kazo0:
closes https://github.com/unoplatform/nventive-private/issues/403

## PR Type

- Bugfix

Typing "123" into the PasswordBox, removing focus from the PasswordBox, and then returning focus to the PasswordBox and attempting to enter more text will cause the previous string of "123" to be cleared.

## What is the new behavior?

Don't clear PasswordBox automatically


MartinZikmund (approved):

jeromelaban:
@mergifyio backport release/stable/4.4

PR #8612, opened 2022-04-25
feat: Enable multiple `XamlRoots`

MartinZikmund:
GitHub Issue (If applicable): closes #

## PR Type

What kind of change does this PR introduce?

- Feature

## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)
- [ ] [Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [ ] Validated PR `Screenshots Compare Test Run` results.
- [ ] Contains **NO** breaking changes
- [ ] Associated with an issue (GitHub or internal) and uses the [automatic close keywords](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue).
- [ ] Commits must be following the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) specification.

## Other information

Internal Issue (If applicable):

PR #9295, opened 2022-07-16
Fixes runtimes skia wpf warnings

workgroupengineering:

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI.Runtime.Skia.Wpf/WpfHost.cs
+++ b/src/Uno.UI.Runtime.Skia.Wpf/WpfHost.cs
@@ -54,7 +54,6 @@ namespace Uno.UI.Skia.Platform
 	{
 		private const string NativeOverlayLayerPart = "NativeOverlayLayer";
 
-		private readonly bool designMode;
 		private readonly Func<UnoApplication> _appBuilder;
 		private CompositeDisposable _registrations = new();
 
@@ -63,13 +62,9 @@ namespace Uno.UI.Skia.Platform
 		[ThreadStatic] private static WpfHost _current;
 
 		private WpfCanvas? _nativeOverlayLayer = null;
-		private WriteableBitmap bitmap;
 		private bool ignorePixelScaling;
 		private FocusManager? _focusManager;
 		private bool _isVisible = true;
-
-		private DisplayInformation _displayInformation;
-		private SKColor _backgroundColor;
 
 		static WpfHost()
 		{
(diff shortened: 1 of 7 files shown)


src/Uno.UI.Runtime.Skia.Wpf/WpfHost.cs
```diff
@@ -307,8 +298,6 @@ private void WpfHost_SizeChanged(object sender, SizeChangedEventArgs e)
 			);
 		}
 
-		public SKSize CanvasSize => bitmap == null ? SKSize.Empty : new SKSize(bitmap.PixelWidth, bitmap.PixelHeight);
```
Youssef1313, line 310: Is a public API change intended here?

workgroupengineering: I removed it because it's not used anywhere in uno repository. If it were used by someone there would already be an issue because at the moment it always returns an empty rectangle.

Youssef1313: @workgroupengineering This didn't use to be the case before https://github.com/unoplatform/uno/pull/8612 😕 

`bitmap` was previously getting assigned in `OnRender`
`designMode` used to be used too.

cc @MartinZikmund

workgroupengineering: after #8612
bitmap is manage from UnoWpfRenderer
is assigned only and never uses other parts.

Youssef1313: @workgroupengineering I mean if someone was using the `CanvasSize` API, then #8612 should have broken them. So instead of removing the field and `CanvasStyle`, I think the correct action might be returning the `bitmap` from `UnoWpfRenderer.Render` so that we keep `CanvasSize` functioning as before.

Youssef1313: Or probably move `CanvasSize` to `UnoWpfRenderer` and call it from `WpfHost`

MartinZikmund: If we want to avoid breaking change here, we can do it like suggested and add the actual removal as planned breaking change for next major release

MartinZikmund (approved):

PR #9395, opened 2022-07-30
feat: Allow overriding default scaling on Framebuffer

### LINKED ISSUE #9394: Allow me to set DPI for framebuffer renderer
When creating the framebuffer host, the renderer is hardcoded to a scale factor of 1:
https://github.com/unoplatform/uno/blob/e4df979d3bb3a931f57bd266182307bfb8de6db0/src/Uno.UI.Runtime.Skia.Linux.FrameBuffer/Renderer.cs#L64

Can we have an overload to `FrameBufferHost` that allows us to set the DPI we'd like to use?
So this:
`var host = new FrameBufferHost(() => new App());`
would change to this:
`var host = new FrameBufferHost(() => new App(), 192f);`

Ie change `public FrameBufferHost(Func<WUX.Application> appBuilder)` to 
`public FrameBufferHost(Func<WUX.Application> appBuilder, float dpi=96f;`

In turn this value will be forwarded to the Renderer initialized in the initialize method:
https://github.com/unoplatform/uno/blob/e4df979d3bb3a931f57bd266182307bfb8de6db0/src/Uno.UI.Runtime.Skia.Linux.FrameBuffer/FramebufferHost.cs#L69

## Why is this needed:
On some IoT device screens, the default size of text and buttons is just way too small. Being able to create the device with a better DPI setting would be preferable. Or if possible, detect the physical size of the screen, and calculate a proper dpi scaling (albeit this might be a breaking behavior). My 2" 480x480pixel screen creates almost unusably small buttons and unreadable text, so a lot of styling is required.

## For which Platform:

- [ ] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] macOS
- [ ] Skia
  - [ ] WPF
  - [x] GTK (Linux)
  - [ ] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

MartinZikmund:
GitHub Issue (If applicable): closes #9394

## PR Type

- Feature

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI.Runtime.Skia.Linux.FrameBuffer/Renderer.cs
+++ b/src/Uno.UI.Runtime.Skia.Linux.FrameBuffer/Renderer.cs
@@ -12,11 +12,12 @@ namespace Uno.UI.Runtime.Skia
 {
 	class Renderer
 	{
+		private readonly FrameBufferHost _host;
 		private FrameBufferDevice _fbDev;
 		private SKBitmap? bitmap;
 		private int renderCount = 0;
 
-		public Renderer()
+		public Renderer(FrameBufferHost host)
 		{
 			_fbDev = new FrameBufferDevice();
 			_fbDev.Init();
@@ -24,6 +25,7 @@ namespace Uno.UI.Runtime.Skia
 			var resolution = _fbDev.ScreenSize;
 			
 			WUX.Window.Current.OnNativeSizeChanged(new Windows.Foundation.Size(resolution.Width, resolution.Height));
+			_host = host;
 		}
 
 		public Size PixelSize => _fbDev.ScreenSize;
@@ -39,15 +41,15 @@ namespace Uno.UI.Runtime.Skia
 				this.Log().Trace($"Render {renderCount++}");
 			}
 
-			var dpi = 1;
+			var scale = _host.ScaleOverride ?? 1f;
 
 			var resolution = _fbDev.ScreenSize;
 
 			width = (int)resolution.Width;
 			height = (int)resolution.Height;
 
-			var scaledWidth = (int)(width * dpi);
-			var scaledHeight = (int)(height * dpi);
+			var scaledWidth = (int)(width * scale);
+			var scaledHeight = (int)(height * scale);
 
 			var info = new SKImageInfo(scaledWidth, scaledHeight, SKImageInfo.PlatformColorType, SKAlphaType.Premul);
 
@@ -61,7 +63,7 @@ namespace Uno.UI.Runtime.Skia
 			{
 				surface.Canvas.Clear(SKColors.White);
 
-				surface.Canvas.Scale((float)dpi);
+				surface.Canvas.Scale(scale);
 
 				WUX.Window.Current.Compositor.Render(surface);
 
--- a/src/Uno.UI.Runtime.Skia.Linux.FrameBuffer/FramebufferHost.cs
+++ b/src/Uno.UI.Runtime.Skia.Linux.FrameBuffer/FramebufferHost.cs
@@ -41,6 +41,8 @@ namespace Uno.UI.Runtime.Skia
 			_eventLoop = new EventLoop();
 		}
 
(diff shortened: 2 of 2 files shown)


MartinZikmund:
@dotMorten Once this build passes and test NuGet packages are generated, will you please be able to test if this works 😊 ?

src/Uno.UI.Runtime.Skia.Linux.FrameBuffer/Renderer.cs
```diff
@@ -39,15 +41,15 @@ void Invalidate()
...
 
 			width = (int)resolution.Width;
 			height = (int)resolution.Height;
 
-			var scaledWidth = (int)(width * dpi);
-			var scaledHeight = (int)(height * dpi);
+			var scaledWidth = (int)(width * scale);
+			var scaledHeight = (int)(height * scale);
 
 			var info = new SKImageInfo(scaledWidth, scaledHeight, SKImageInfo.PlatformColorType, SKAlphaType.Premul);
```
dotMorten, line 54: Isn’t this supposed to be the physical size ? If I set the scale factor to 2, the physical size of the display didn’t change. If it’s device independent units then it should have gotten smaller instead.

spouliot (approved):

src/Uno.UI.Runtime.Skia.Linux.FrameBuffer/Renderer.cs
dotMorten: Looking over skimageinfo this is definitely physical pixels and there the number passed into it shouldn’t change. I think the original code was just written with misleading variable names and worked because the factor was 1. The only scaling factor that needs changing I think is for the canvas itself and nothing else.

dotMorten:
Perhaps it’s better to read the system dpi setting in framebufferhost? https://stackoverflow.com/questions/71120399/how-get-correct-dpi-information-on-linux

src/Uno.UI.Runtime.Skia.Linux.FrameBuffer/Renderer.cs
MartinZikmund: You are right... and I think the the `DisplayInformation` would need to be updated to apply the override too 🤔

MartinZikmund:
@dotMorten Ah, that would definitely be the best, unfortunately I don't have any device to test on, so I was trying in the blind - and it didn't work out 😅 . I think @jeromelaban has a test device so he could probably try to look into it next week. Alternatively - if you were interested to contribute this feature, I am happy to help you out to get the solution running locally.

jeromelaban:
> Perhaps it’s better to read the system dpi setting in framebufferhost?
> 
> * https://stackoverflow.com/questions/71120399/how-get-correct-dpi-information-on-linux
> * https://winaero.com/find-change-screen-dpi-linux/amp/

We may not be able to read the DPI information this way (X11 is not supposed to be available when the framebuffer is used, or may not be installed at all). Still, the framebuffer API provides a physical size value, we may be able to determine a default scale based on that, if the information is accurate.

src/Uno.UI.Runtime.Skia.Linux.FrameBuffer/Renderer.cs
jeromelaban: `DisplayInformation ` definitely needs to be updated properly, indeed.

jeromelaban:
So, it seems that `fbinfo.width/height` can provide proper values on some devices. I got exact size values on a physical device (e.g. `Screen dimensions: 154x86 mm`, and got `-1x-1` on an Hyper-V virtual machine's screen). We should be able to guess a scale based on that, and provide it to `DisplayInformation` as well.

dotMorten:
@jeromelaban Cool! I do wonder if there should be an environment variable you should be able to set to override or provide this, so you can still configure per-device without having to recompile.

jeromelaban:
I've implemented this and some more in https://github.com/unoplatform/uno/pull/9418.

Issue #8910, opened 2022-05-28
[Skia] `ApplicationData.Current.LocalFolder` redirects to %appdata%


jeromelaban:
### Current behavior

`ApplicationData.Current.LocalFolder` redirects to %appdata%, but should redirect to a more app-specific folder.

### Expected behavior

Writing to `ApplicationData.Current.LocalFolder` should be specific to the running app.

## Note for contributors

The behavior for this is currently undefined, as it is mapped to the UWP behavior. We should be mapping to the behavior that `ApplicationData.Current.LocalFolder` shows when running a unpackaged WinAppSDK app for the same API.

MartinZikmund:
This will be a breaking change, but should probably be done soon.

MartinZikmund:
@jeromelaban I think this is one of the breaking changes that are really needed for the next major release

MartinZikmund:
Currently unpackaged WinAppSDK apps just crash when `ApplicationData.Current` is used. See https://github.com/microsoft/WindowsAppSDK/issues/2639, https://github.com/microsoft/WindowsAppSDK/issues/2621 .

Because we have the `Package.appxmanifest`, I think for the default behavior we can read the publisher and app id from there, and also provide a `FeatureConfiguration` to let the developer to override the path and store the app data/settings somewhere else as needed, does that sound like a good solution @jeromelaban ?

jeromelaban:
altering the path using publisher and appid makes sense yes, as for feature configuration, I wonder if an helper class would be more appropriate?

PR #9532, opened 2022-08-18
Add support for more `Package` properties

MartinZikmund:

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UWP/ApplicationModel/Package.Other.cs
+++ b/src/Uno.UWP/ApplicationModel/Package.Other.cs
@@ -1,132 +1,95 @@
-#if !(__APPLE_UIKIT__ || __ANDROID__ || __TVOS__)
 #nullable enable
+#if !(__APPLE_UIKIT__ || __ANDROID__ || __MACOS__)
 using System;
+using System.Collections.Generic;
 using System.Reflection;
+using System.Security.Cryptography;
 using System.Xml;
+using Uno.Extensions;
 using Uno.Foundation.Logging;
+using Uno.UI;
+using Windows.ApplicationModel.Email.DataProvider;
+using Windows.Storage;
 
-namespace Windows.ApplicationModel;
-
-public partial class Package
+namespace Windows.ApplicationModel
 {
-	private const string PackageManifestName = "Package.appxmanifest";
-
-	private static Assembly? _entryAssembly;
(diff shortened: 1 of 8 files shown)


src/Uno.UWP/ApplicationModel/Package.Other.cs
```diff
@@ -21,9 +21,9 @@ public partial class Package
 		private string _logo = "ms-appx://logo";
 		private bool _manifestParsed;
 
-		private bool GetInnerIsDevelopmentMode() => false;
+		private bool GetIsDevelopmentMode() => false;
 
-		private DateTimeOffset GetInstallDate() => DateTimeOffset.Now;
+		private DateTimeOffset GetInstallDate() => new DateTimeOffset(2000, 1, 1, 0, 0, 0, TimeSpan.Zero);
```
jeromelaban, line 26: I wonder if we could create a file (or read some other file) on first launch then read the date so we can provide that date more reliably?

MartinZikmund: That is a good idea! It would be the first launch date, but that is still more accurate than nothing 😀 . However, first we need to resolve https://github.com/unoplatform/uno/issues/8910 so that it is really app specific.

pkar70:
I think this is bad.
You return some properties that leads to errors in app logic.
GetInstallDate() => new DateTimeOffset(2000, 1, 1, 0, 0, 0, TimeSpan.Zero); - but this date is not install date. So, any logic (as e.g. "only allow work for 3 months since install, as this is demo version of app"), would fail and app would not work.
You return InstalledPath on Android - so, app that expects that it can e.g. open file from this location would crash (e.g. to open Manifest file from InstalledPath to check build date), as Android doesn't extract files from apk.

src/Uno.UWP/ApplicationModel/Package.Other.cs
pkar70: Why use such placeholders? It is simply not supported atrribute.
App programmer should be warned on compile time that he/she should check app logic that uses GetInstalledDate.
E.g.: tryware, and logic as 
` if(date.Now - GetInstalledDate > 14 days) showerror("free use period ends, please buy this app")`

Issue #693, opened 2019-03-15
[WASM] Font preloading support 


jeromelaban:
## I'm submitting a...
- Feature request

## Current behavior
Fonts specified as WebFonts do not get loaded synchronously, making TextBlock based control measure their content incorrectly. Loading fonts using a data-uri is not a workaround.

## Expected behavior
Fonts are loaded and measured properly.

## Possible workaround

This works as a workaround, forcing fonts to load very early in the app's life cycle:
```css
body::after {
    font-family: 'Symbols';
    background: transparent;
    content: "";
    opacity: 0;
    pointer-events: none;
    position: absolute;
}
```

## Environment

```
Nuget Package: 

Package Version(s): 

Affected platform(s):
- [ ] iOS
- [ ] Android
- [x] WebAssembly
- [ ] Windows
- [ ] Build tasks

Visual Studio
- [ ] 2017 (version: )
- [ ] 2017 Preview (version: )
- [ ] for Mac (version: )

Relevant plugins
- [ ] Resharper (version: )
```

zipswich:
![image](https://user-images.githubusercontent.com/6095671/67943651-d9e52600-fbb0-11e9-804a-8a67fc81abe9.png)

carldebilly:
`FontFace.load()` could probably be used to fix this problem.
Doc: https://developer.mozilla.org/en-US/docs/Web/API/FontFace/load

carldebilly:
The [`FontFaceSet`](https://developer.mozilla.org/en-US/docs/Web/API/FontFaceSet) may also help on this issue.

agneszitte-nventive:
@jeromelaban / @carldebilly Interesting article here in my opinion:
https://nooshu.com/blog/2021/01/23/the-importance-of-font-face-source-order-when-used-with-preload/

agneszitte-nventive:
(@jeromelaban FYI, there is this item also https://github.com/unoplatform/Uno.Themes/issues/733)

jeromelaban:
Available in 4.6

PR #9543, opened 2022-08-18
[WASM] Font preloading support

agneszitte-nventive:
GitHub Issue (If applicable): [#693](https://github.com/unoplatform/uno/issues/693#)

## PR Type

- Feature

Missing WASM Font preloading support

## What is the new behavior?

Added WASM Font preloading support

## PR Checklist


carldebilly (approved):

src/Uno.UI/UI/Xaml/FontFamily.wasm.cs
```diff
@@ -1,58 +1,48 @@
...
 		/// </summary>
-		internal string ParsedSource { get; private set; }
+		public static void Preload(FontFamily family) => family._loader.LaunchLoading();
 
-		private string ParseFontFamilySource(string familyName)
+		/// <summary>
+		/// Use this to launch the loading of a font before it is actually required to
+		/// minimize loading time and prevent potential flicking.
+		/// </summary>
+		public static void Preload(string familyName) => Preload(new FontFamily(familyName));
```
MartinZikmund, line 32: I think that these methods need to be `internal` and then need to be made `public` using an API inside `Uno.UI.Runtime.WebAssembly`, right @jeromelaban? Otherwise they will not actually be accessible (because the magic with reference API, only things that are netstdref are actually visible). I hit the same problem when I implemented `CookieManager` which is WASM specific - it was usable from Samples app (as it references Uno.UI.Wasm directly), but then not usable in actual apps, and had to move it to `Uno.UI.Runtime.WebAssembly` for that reason.

src/Uno.UI/UI/Xaml/FontFamily.wasm.cs
```diff
@@ -1,58 +1,48 @@
...
+		internal string CssFontName => _loader.CssFontName;
 
 		/// <summary>
-		/// Contains the parsed font family for use in WASM
-		/// (matches CSS @font-face's font-family)
+		/// Use this to launch the loading of a font before it is actually required to
+		/// minimize loading time and prevent potential flicking.
 		/// </summary>
-		internal string ParsedSource { get; private set; }
+		public static void Preload(FontFamily family) => family._loader.LaunchLoading();
```
MartinZikmund, line 26: Also, could there be a "async" variant of the `Preload` method as well? So that the user could actually `await` its completion to be sure that all required fonts are ready - this could then be used in `OnLaunched` before the first page navigation.

src/Uno.UI/UI/Xaml/FontFamily.Loader.wasm.cs
```diff
@@ -0,0 +1,158 @@
...
+			internal static FontLoader GetLoaderForFontFamily(FontFamily forFamily)
+			{
+				if (_loaders.TryGetValue(forFamily, out var loader))
+				{
+					// There is already a loader for this font family
+					return loader;
+				}
+
+				loader = new FontLoader(forFamily);
+				_loaders.Add(forFamily, loader);
```
Youssef1313, line 26: Any concurrency concerns here? 
I think this is problematic if multiple threads execute this line, but I'm not sure if this can happen.

jeromelaban: This cannot happen in this context, the code is always called from the UI thread.

src/Uno.UI/UI/Xaml/FontFamily.wasm.cs
jeromelaban: Those methods (even with the newer changed signatures) are accessible from the main app, but indeed won't be from a library. The point is for these methods to be called during startup in the main project, so if we need to move those to the runtime assembly, we'll be able to do so in the future.

src/Uno.UI/UI/Xaml/FontFamily.wasm.cs
jeromelaban: Done!

src/Uno.UI/UI/Xaml/FontFamily.cs
```diff
@@ -42,7 +45,7 @@ public override bool Equals(object obj)
 				return true;
 			}
 
-			return !ReferenceEquals(a, null) && a.Equals(b);
+			return a is not null && a.Equals(b);
```
dr1rrb, line 48: I think that this equivalent to `a.Equals(null)` which is not the same as previous (as we have override the `Equals`). Regarding the current implementation the result will be the same (unless `Equals` is overridden in a sub-class), but even if less cute, I would keep `ReferenceEquals` as it's more performant ... unless that behavior change is expected of course!

(BTW personally when I override `Equals` implement equality operators, I prefer to create a static method `Equals(MyType left, MyType right)` which use only `Equals` and/or `ReferenceEquals`, and redirect all other to this one. This avoid unexpected recursive calls.)

src/Uno.UI/UI/Xaml/FontFamily.cs
```diff
@@ -52,7 +55,7 @@ public override bool Equals(object obj)
 				return false;
 			}
 
-			return ReferenceEquals(a, null) || !a.Equals(b);
+			return a is null || !a.Equals(b);
```
dr1rrb, line 58: Same thing here for the `ReferenceEquals`

src/Uno.UI/UI/Xaml/FontFamilyHelper.cs
```diff
@@ -1,58 +1,58 @@
...
+	/// This methods removes the font files extensions, typically .otf or .ttf because in iOS
+    /// you need to refer to a FontFamily via its name without extension and in Android you need the extension.
+	/// </summary>
+	/// <param name="familyName"></param>
+	/// <returns></returns>
+	public static string RemoveExtension(string familyName)
+	{
+		return familyName
+			.Replace(".otf", string.Empty)
+			.Replace(".ttf", string.Empty);
```
dr1rrb, line 33: `System.IO.Path.GetFileNameWithoutExtension()` (not sure it exists on all platform, if not you should add the `StringComparison.OrdinalIgnoreCase`)

src/Uno.UI/UI/Xaml/FontFamilyHelper.cs
```diff
@@ -1,58 +1,58 @@
...
-			if (hashIndex != -1)
-			{
-				familyName = familyName.Substring(0, hashIndex);
-			}
-			return familyName;
+		if (hashIndex != -1)
+		{
+			familyName = familyName.Substring(0, hashIndex);
 		}
+		return familyName;
```
dr1rrb, line 56: As the class seems to need to be `public`, should we at least make those method `internal` ?

src/Uno.UI/UI/Xaml/FontFamilyLoader.wasm.cs
```diff
@@ -0,0 +1,244 @@
...
+using Uno.Foundation.Logging;
+using Uno.UI.DataBinding;
+using Windows.Storage.Helpers;
+
+namespace Windows.UI.Xaml.Media;
+
+/// <summary>
+/// WebAssembly-specific asynchronous font loader
+/// </summary>
+class FontFamilyLoader
```
dr1rrb, line 16:
```suggestion
internal class FontFamilyLoader
```

doc/articles/features/custom-fonts.md
```diff
@@ -92,6 +135,33 @@ Will match:
...
+
+public static void main(string[] orgs)
+{
+    // Add this in your application to preload a font.
+    // You can add more than one, but preload too many fonts could hurt user experience.
+    // IMPORTANT: The string parameter should be exactly the same string (including casing)
+    //            used as FontFamily in the application.
+    Windows.UI.Xaml.Media.FontFamilyHelper.Preload("ms-appx:///Assets/Fonts/yourfont01.ttf#ApplicationFont01");
+    Windows.UI.Xaml.Media.FontFamilyHelper.Preload("https://fonts.cdnfonts.com/s/71084/antikythera.woff#Antikythera");
+    Windows.UI.Xaml.Media.FontFamilyHelper.Preload("Roboto");
```
dr1rrb, line 160: Should we automatically preload all fonts located in the `Assets` folder?

dr1rrb (approved):

src/Uno.UI/UI/Xaml/FontFamily.cs
jeromelaban: `c is null` is equvalient to `object.ReferenceEquals(c, null)`

src/Uno.UI/UI/Xaml/FontFamilyHelper.cs
jeromelaban: https://github.com/unoplatform/uno/issues/8339

doc/articles/features/custom-fonts.md
jeromelaban: Let's review that after the canaries validations

src/Uno.UI/UI/Xaml/FontFamilyHelper.wasm.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Windows.UI.Xaml;
+
+public partial class FontFamilyHelper
+{
+	/// <summary>
+	/// Pre-loads a font to minimize loading time and prevent potential text re-layouts.
+	/// </summary>
+	/// <returns>True is the font loaded successfuly, otherwise false.</returns>
+	public static Task<bool> Preload(FontFamily family)
```
MartinZikmund, line 15: I suggest renaming this to `PreloadAsync` so that it is clear that it can be awaited (also to stay in line with the WinUI/UWP naming convention). Also, the docs don't include the `await` yet, maybe it should be added there too.

Youssef1313: @MartinZikmund There is analyzer to enforce that. Is it interesting to enable it for the repo?

https://github.com/microsoft/vs-threading/blob/main/doc/analyzers/VSTHRD200.md

jeromelaban: I don't think we'll be able to enable it for the whole repro (many APIs need to stay as-is to avoid breaking changes), but it may be interesting in other cases, like tools.

PR #9559, opened 2022-08-20
fix: Updated workaround for iOS 16 `UITextInput.SelectedTextRange` crashes

spouliot:

## PR Type

- Bugfix

`UITextInput.SelectedTextRange` crashes on iOS 16 betas

## What is the new behavior?

Updated workaround to avoid the runtime crash while keeping the `UITextInput.SelectedTextRange` functionality available.

## PR Checklist

## Other information

Pro: It does not break app compatibility, selection works

Con: This does not work _currently_ with the dynamic registrar [1] which is the default when building for iOS simulators. OTOH there's a workaround for that [2].

[1] Unless building with the iOS SDK for Xcode 14 (currently only previews) once https://github.com/xamarin/xamarin-macios/pull/15712 is merged

[2] Adding `--registrar=static` to the simulator builds will fix this, sadly build times will be longer (so better use [1] asap)

### DIFF
--- a/build/PackageDiffIgnore.xml
+++ b/build/PackageDiffIgnore.xml
@@ -8903,6 +8903,12 @@
 			<Member
 				fullName="System.Void Windows.UI.Xaml.Controls.SinglelineTextBoxView::set_SelectedTextRange(UIKit.UITextRange value)"
 				reason="Workaround for iOS 16" />
+			<Member
+				fullName="UIKit.UITextRange Windows.UI.Xaml.Controls.MultilineTextBoxView::get_SelectedTextRange()"
+				reason="Workaround for iOS 16" />
+			<Member
+				fullName="System.Void Windows.UI.Xaml.Controls.MultilineTextBoxView::set_SelectedTextRange(UIKit.UITextRange value)"
+				reason="Workaround for iOS 16" />
 			<!-- END iOS 16 workaround -->
 		</Methods>
 		<Properties>
@@ -8918,6 +8924,9 @@
 			<!-- BEGIN iOS 16 workaround -->
 			<Member
 				fullName="UIKit.UITextRange Windows.UI.Xaml.Controls.SinglelineTextBoxView::SelectedTextRange()"
+				reason="Workaround for iOS 16" />
+			<Member
+				fullName="UIKit.UITextRange Windows.UI.Xaml.Controls.MultilineTextBoxView::SelectedTextRange()"
 				reason="Workaround for iOS 16" />
 			<!-- END iOS 16 workaround -->
 		</Properties>
--- a/src/SamplesApp/SamplesApp.iOS/SamplesApp.iOS.csproj
+++ b/src/SamplesApp/SamplesApp.iOS/SamplesApp.iOS.csproj
@@ -28,7 +28,7 @@
     <MtouchArch>x86_64</MtouchArch>
     <MtouchLink>None</MtouchLink>
     <MtouchDebug>true</MtouchDebug>
-    <MtouchExtraArgs>--dlsym:Microsoft.Win32.Registry.dll --xml=./LinkerExclusions.xml --linkskip=$(AssemblyName) --setenv=MONO_GC_PARAMS=soft-heap-limit=1024m,nursery-size=64m,evacuation-threshold=66,major=marksweep,concurrent-sweep</MtouchExtraArgs>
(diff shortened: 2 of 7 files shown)


MartinZikmund (approved):

jeromelaban:
It looks like there's something the runtime tests don't like, but I'm not seeing any exceptions in the apps log. Maybe it's changing the runtime behavior?

spouliot:
@jeromelaban it should not (change the behaviour). I'll investigate this asap, thanks!

jeromelaban:
So the crash here is like this:

```
Xamarin.iOS: Don't know how to marshal a return value of type 'System.IntPtr'. Please file a bug with a test case at https://github.com/xamarin/xamarin-macios/issues/new
```

spouliot:
@jeromelaban doh! That's why I could not duplicate it. That's the issue [I fixed in xamarin-macios](https://github.com/xamarin/xamarin-macios/pull/15712) and requires using `--registrar=static` for simulator builds (until the fix is available)

jeromelaban:
Excellent! Let's see if the CI is happier now ;) Thanks!

spouliot:
@jeromelaban the CI still need another commit to add this flag. Resetting my mono tree to test it right now :)

jeromelaban:
@spouliot I can confirm that it works with the static registrar !

jeromelaban:
SelectedRange's value is used here, so it seems we still need to read the proper content: https://github.com/unoplatform/uno/blob/300d0e10325d447a89d199b4fae93dab3a97908d/src/Uno.UI/Controls/Window.iOS.cs#L389. Do you need to expose that property as the original name? Since there's the attribute specifying the native name, it may not be required and we can put that property private with another name?

spouliot:
@jeromelaban I forgot to include that change with my earlier commit.

As far as I know (and tested) there's no issue marshalling the native `UITextRange` subclass into managed code. The problem seems _only related_ when a managed `UITextRange` instance is sent to native code. That's why only the setter was causing crash.

build/PackageDiffIgnore.xml
```diff
@@ -8903,6 +8903,12 @@
 			<Member
 				fullName="System.Void Windows.UI.Xaml.Controls.SinglelineTextBoxView::set_SelectedTextRange(UIKit.UITextRange value)"
 				reason="Workaround for iOS 16" />
+			<Member
```
jeromelaban, line 8906: I wonder, wouldn't we be able to keep this member as-is (albeit with a `new`) and create a new private member with a new name that has the IntPtr property?

spouliot: I guess so... but that would still be a breaking change as it would remove the `override`. Maybe the tool does not care ? OTOH any binary code that refers to the members would not be happy.

jeromelaban: That's fine anyways, that property is on a type that is not part of the WinUI API. I guess that it's mostly to keep the IntPtr trick inside the `SinglelineTextBoxView` class.

jeromelaban: So I tried implementing this:

```csharp
		public new UITextRange SelectedTextRange
		{
			get => Runtime.GetNSObject<UITextRange>(SelectedTextRangeWorkaround);
			set => base.SelectedTextRange = value;
		}

		/// <summary>
		/// Workaround for https://github.com/unoplatform/uno/issues/9430
		/// </summary>
		[Export("selectedTextRange")]
		private IntPtr SelectedTextRangeWorkaround
		{
			get
			{
				return SinglelineTextBoxView.IntPtr_objc_msgSendSuper(SuperHandle, Selector.GetHandle("selectedTextRange"));
			}
			set
			{
				if (SelectedTextRangeWorkaround != value)
				{
					SinglelineTextBoxView.void_objc_msgSendSuper(SuperHandle, Selector.GetHandle("setSelectedTextRange:"), value);
					_textBox.GetTarget()?.OnSelectionChanged();
				}
			}
		}
```

But this causes:
```
error MT4119: Could not register the selector 'selectedTextRange' of the member 'Windows.UI.Xaml.Controls.MultilineTextBoxView.get_SelectedTextRange' because the selector is already registered on the member 'get_SelectedTextRangeWorkaround'.
error MT4119: Could not register the selector 'setSelectedTextRange:' of the member 'Windows.UI.Xaml.Controls.MultilineTextBoxView.set_SelectedTextRange' because the selector is already registered on the member 'set_SelectedTextRangeWorkaround'.
```

Since` SinglelineTextBoxView` is an implementation detail, we can keep the current changes, and roll back whenever possible.

spouliot: In general (there's always exceptions) you cannot export from C#  the same selector twice (inside the same type).

Calling (managed to native) a selector is not a problem. However being called from native code requires a single entry point (method) for the selector. If there's more than one then you would not know which method would be called, hence the registrar complaining with a build error.

PR #9696, opened 2022-09-01
fix(Wasm): Fix DoubleTapped not firing with touch

### LINKED ISSUE #8042: [WASM] Cannot double tapped with Touch
### Current behavior

When double tapping an element, the `Tapped` event is fired twice, but the `DoubleTapped` is not.

### Expected behavior

🙃

_No response_

### Workaround

_No response_

### Works on UWP/WinUI

_No response_

### Environment

_No response_

### NuGet package version(s)

4.1

### Affected platforms

_No response_

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

Pointer captures also behaves strangely when using touch pointer.

Youssef1313:
GitHub Issue (If applicable): closes #8042

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/UIElement.Pointers.wasm.cs
+++ b/src/Uno.UI/UI/Xaml/UIElement.Pointers.wasm.cs
@@ -251,10 +251,10 @@ public partial class UIElement : DependencyObject
 	{
 		if (_nativeToManagedPointerId.TryGetValue(nativeId, out var managedId))
 		{
-			return managedId;
+			return managedId.Id;
 		}
 
-		var managedId = new PointerIdentifier(nativeId.Type, ++_lastUsedId);
+		managedId = new PointerIdentifier(nativeId.Type, ++_lastUsedId);
 		_managedToNativePointerId[managedId] = nativeId;
 		_nativeToManagedPointerId[nativeId] = managedId;
 		
--- a/src/Uno.UI/UI/Xaml/UIElement.Pointers.wasm.cs
+++ b/src/Uno.UI/UI/Xaml/UIElement.Pointers.wasm.cs
@@ -243,7 +243,7 @@ public partial class UIElement : DependencyObject
 
 	private static readonly HashSet<uint> _activePointers = new();
 	private static readonly Dictionary<uint, uint> _nativeToManagedPointerId = new();
-	private static int _lastUsedId = 0;
+	private static int _lastUsedId;
 
 	private static uint TransformPointerId(uint nativePointerId)
 	{
--- a/src/Uno.UI/UI/Xaml/UIElement.Pointers.wasm.cs
+++ b/src/Uno.UI/UI/Xaml/UIElement.Pointers.wasm.cs
@@ -18,6 +18,8 @@ using Uno.UI.Extensions;
 using Uno.UI.Xaml;
 using Uno.UI.Xaml.Core;
 
+using PointerIdentifier = Windows.Devices.Input.PointerIdentifier;
+
 #if HAS_UNO_WINUI
 using Microsoft.UI.Input;
 #else
@@ -241,19 +243,18 @@ public partial class UIElement : DependencyObject
 		}
 	}
 
-	private static readonly HashSet<uint> _activePointers = new();
-	private static readonly Dictionary<uint, uint> _nativeToManagedPointerId = new();
+	private static readonly Dictionary<PointerIdentifier, uint> _nativeToManagedPointerId = new();
 	private static uint _lastUsedId;
 
-	private static uint TransformPointerId(uint nativePointerId)
+	private static uint TransformPointerId(PointerIdentifier nativePointerIdentifier)
 	{
-		if (_nativeToManagedPointerId.TryGetValue(nativePointerId, out var pointerId))
+		if (_nativeToManagedPointerId.TryGetValue(nativePointerIdentifier, out var pointerId))
 		{
 			return pointerId;
 		}
 
 		pointerId = ++_lastUsedId;
-		_nativeToManagedPointerId[nativePointerId] = pointerId;
+		_nativeToManagedPointerId[nativePointerIdentifier] = pointerId;
 		return pointerId;
 	}
 
@@ -263,8 +264,6 @@ public partial class UIElement : DependencyObject
 		{
 			_lastUsedId--;
 		}
-
-		_activePointers.Remove(pointerId);
 	}
 
 	private static PointerRoutedEventArgs ToPointerArgs(
@@ -275,13 +274,9 @@ public partial class UIElement : DependencyObject
 		const int cancel = (int)NativePointerEvent.pointercancel;
 		const int exitOrUp = (int)(NativePointerEvent.pointerout | NativePointerEvent.pointerup);
 
-		var pointerId = TransformPointerId((uint)args.pointerId);
-		if ((NativePointerEvent)args.Event == NativePointerEvent.pointerdown)
-		{
-			_activePointers.Add(pointerId);
-		}
-
 		var pointerType = (PointerDeviceType)args.deviceType;
+		var pointerId = TransformPointerId(new PointerIdentifier((Windows.Devices.Input.PointerDeviceType)pointerType, (uint)args.pointerId));
+		
 		var src = GetElementFromHandle(args.srcHandle) ?? (UIElement)snd;
 		var position = new Point(args.x, args.y);
 		var isInContact = args.buttons != 0;
--- a/src/Uno.UI/UI/Xaml/UIElement.Pointers.wasm.cs
+++ b/src/Uno.UI/UI/Xaml/UIElement.Pointers.wasm.cs
@@ -241,19 +241,24 @@ public partial class UIElement : DependencyObject
 		}
 	}
 
-	private static readonly HashSet<uint> s_activePointers = new();
-	private static Dictionary<uint, uint> s_nativeToManagedPointerId = new();
+	private static readonly HashSet<uint> _activePointers = new();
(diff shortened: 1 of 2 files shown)


src/Uno.UWP/UI/Input/GestureRecognizer.cs
```diff
@@ -241,7 +241,9 @@ private static ulong GetPointerIdentifier(PointerPoint point)
 
 			var props = point.Properties;
 
-			ulong identifier = point.PointerId;
+			// Skipping pointer id for touch for now.
+			// It changes on every touch, preventing double tap from working correctly.
+			ulong identifier = point.PointerDeviceType == PointerDeviceType.Touch ? 0 : point.PointerId;
```
Youssef1313, line 246: Not sure if I'm causing any other problems with that.

cc @dr1rrb

dr1rrb: Yes you will most-probably break multi touch support here. We should fix the pointer ID itself instead ... on which browser do you have this issue?

dr1rrb (changes requested):

Youssef1313:
@dr1rrb I was testing on Edge, but I believe this will happen on any browser. It seems that two touches can never generate the same pointer id.

Youssef1313:
> Yes you will most-probably break multi touch support here.

Where is that implemented? I can change the fix to ignore touch pointer id for double tapping specifically (ie, make the fix [here](https://github.com/unoplatform/uno/blob/9ae73a42303894d084634ea018457a3974de3e8c/src/Uno.UWP/UI/Input/GestureRecognizer.Gesture.cs#L337) instead).

dr1rrb:
> @dr1rrb I was testing on Edge, but I believe this will happen on any browser. It seems that two touches can never generate the same pointer id.

Well if there is no way to get a valid pointer ID from the browser, we can just count the number of active touch pointers on the screen and use it as ID. The issue is to make sure to track each finger properly, for instance:
1. press finger 1 (Id = activeTouches++ == 1)
2. press finger 2 (Id = activeTouches++ == 2)
3. release finger 1
4. release finger 2
We must ensure that the pointer ID for the 4th event (release finger 2)  will effectively be with pointerID = 2 even if we have only one finger remaining on the screen. Also if we add a new finger before releasing finger 2, ptID as to be 3!

Currently we are directly mapping the `pointerId` of the browser (https://github.com/unoplatform/uno/blob/master/src/Uno.UI/ts/Windows/UI/Xaml/UIElement.Pointers.ts#L214 and in managed code https://github.com/unoplatform/uno/blob/master/src/Uno.UI/UI/Xaml/UIElement.Pointers.wasm.cs#L142). We could / should have a global mapping of active "native pointers" to "managed pointer id" (that's basically what we are doing on skia).

Youssef1313:
@dr1rrb Can you take another look?

src/Uno.UI/UI/Xaml/UIElement.Pointers.wasm.cs
```diff
@@ -249,7 +264,16 @@ public static void OnNativePointerEvent()
...
 
-		var pointerId = (uint)args.pointerId;
+		var pointerId = TransformPointerId((uint)args.pointerId);
+		if ((NativePointerEvent)args.Event == NativePointerEvent.pointerdown)
+		{
+			s_activePointers.Add(pointerId);
+		}
+		else if ((NativePointerEvent)args.Event == NativePointerEvent.pointerup)
+		{
+			s_activePointers.Remove(pointerId);
```
dr1rrb, line 274: Yeah that's the good way to go 👍

The issue here is that you will remove the entry on the first "pointerup" before the event bubbles. But events are bubbling in native on wasm (unless they are flagged as handled) so this method is going to be invoked also for all parent elements (as soon as they have pointer event handler registered) which might drive to have inconsistent ID.

Fortunately we now a unique root element where we can handle such kind of requirement, the `RootVisual`. So you should move the cleanup logic here: https://github.com/unoplatform/uno/blob/master/src/Uno.UI/UI/Xaml/Internal/RootVisual.cs#L143. You will also have to cleanup on `PointerCancelled`.

Also, I would do that only for touch pointers if possible, it would avoid useless dictionary lookup.

src/Uno.UI/UI/Xaml/UIElement.Pointers.wasm.cs
```diff
@@ -241,6 +241,21 @@ public static void OnNativePointerEvent()
 		}
 	}
 
+	private static readonly HashSet<uint> s_activePointers = new();
```
dr1rrb, line 244: Note: We usually don't prefix static member by `s_` in uno

src/Uno.UI/UI/Xaml/UIElement.Pointers.wasm.cs
Youssef1313: > Also, I would do that only for touch pointers if possible

What about "Pen" pointers? I don't have a stylus device to be able to test :(

src/Uno.UI/UI/Xaml/UIElement.Pointers.wasm.cs
```diff
@@ -241,6 +241,21 @@ public static void OnNativePointerEvent()
...
+	private static Dictionary<uint, uint> s_nativeToManagedPointerId = new();
+
+	private static uint TransformPointerId(uint nativePointerId)
+	{
+		if (s_nativeToManagedPointerId.TryGetValue(nativePointerId, out var pointerId))
+		{
+			return pointerId;
+		}
+
+		pointerId = s_activePointers.Count == 0 ? 1 : s_activePointers.Max() + 1;
```
dr1rrb, line 254: Also: I think that the `Max()` will enumerate the collection. It would be great to just reset to 0 when we remove the last ID from the `s_nativeToManagedPointerId`.

src/Uno.UI/UI/Xaml/UIElement.Pointers.wasm.cs
dr1rrb: Well, I could test but, yeah it's probably safer to it for all pointer types then. Then should we consider to include the type of the pointer? I.e. we should not confuse finger_1 and mouse_1 (on Android, I used a `long` and kept the pointer type in bit 33 :) )

Youssef1313: > Then should we consider to include the type of the pointer? I.e. we should not confuse finger_1 and mouse_1 

Shouldn't the native pointer id already be different?

src/Uno.UI/UI/Xaml/UIElement.Pointers.wasm.cs
dr1rrb: You can keep a `_activePointerCount` that you reset to 0 only when you clear the `_activePointers` list :)

src/Uno.UI/UI/Xaml/UIElement.Pointers.wasm.cs
dr1rrb: Depending of the platform not necessarily. As I'm not sure of the behavior for each browser on each OS (yes same browser on different OS sometimes behaves differently), I would try to be the safest possible.

BTW, just realized that I do already have a an helper for that : https://github.com/unoplatform/uno/blob/master/src/Uno.UWP/Devices/Input/PointerIdentifier.cs

src/Uno.UI/UI/Xaml/UIElement.Pointers.wasm.cs
Youssef1313: I think that's not equivalent, unless I'm missing something.

Consider two cases:

#### Case 1

- press finger 1
- press finger 2
- release finger 1
- press finger 3 (here I expect the id to be 3)

#### Case 2

- press finger 1
- press finger 2
- release finger 2
- press finger 3 (here I expect the id to be 2)

If I used `_activePointerCount`, the id will be 3 in both cases I think?

dr1rrb: Yes the ID should be 3, but here it depends if you decrement the `_activePointerCount` when finger 2 is released or not. You should not and reset it to 0 only when all fingers are removed from the screen (i.e. `s_activePointers is {Count: 0}`).

I agree that `_activePointerCount` might not be the right name though.

Youssef1313: @dr1rrb Fixed in last commit. Not sure if the logic I wrote is correct or if it matches your intent. Can you take a look? Thanks!

Youssef1313:
> src\Uno.UI\UI\Xaml\UIElement.Pointers.wasm.cs(248,41): Error CS0246: The type or namespace name 'PointerIdentifier' could not be found (are you missing a using directive or an assembly reference?)

Happening for WinUI only.

@jeromelaban Are we missing a `Windows` replacement with `Microsoft`?

Youssef1313:
@jeromelaban It looks like `PointerDeviceType` is always in `Windows.Devices.Input` namespace, but for something the `using Windows.Devices.Input` gets replaced to `Microsoft.UI.Input`.

That's what I could tell from the error:

> src\Uno.UI\UI\Xaml\UIElement.Pointers.wasm.cs(278,60): Error CS1503: Argument 1: cannot convert from 'Microsoft.UI.Input.PointerDeviceType' to 'Windows.Devices.Input.PointerDeviceType'

I'm not very sure about the WinUI conversion tool and how it's supposed to work. Can you help? Thanks!

Youssef1313:
@dr1rrb Can you take a look please? Thanks!

PR #9698, opened 2022-09-01
feat: Use dotnet SDK linker when available

ebariche:
Use the .NET linker to improve the performance, as the embedded linker was an earlier build which has performance issues.


jeromelaban (approved):
Nice!

spouliot (approved):

MartinZikmund (approved):

PR #9707, opened 2022-09-02
fix(AutoSuggestBox): Close suggestion list on losing focus

### LINKED ISSUE #8836: AutoSuggestBox Popup closing after loosing focus not work
### Current behavior

On UWP, when the user change focus to anywhere else than AutosuggestBox components, the popup is closed. 
On other platforms it doesn't work.

### Expected behavior

Must work as similar as possible to UWP.

Just create a new AutoSuggestBox and give it a source to search.
When you type anything, the popup will appear. 
Then, try to click anywhere else than the textbox itself and the popup.

### Workaround

Pressing ESC on SKIA and WASM works pretty well. But not loosing focus by mouse events.
For Android it is worst. AutoSuggest not fires the KeyPress event when the SoftKeyboard is used, an the ESC key does nothing.

### Works on UWP/WinUI

_No response_

### Environment

_No response_

### NuGet package version(s)

4.3.0-dev.288

### Affected platforms

_No response_

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

Youssef1313:
GitHub Issue (If applicable): closes #8836

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/AutoSuggestBox/AutoSuggestBox.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/AutoSuggestBox/AutoSuggestBox.cs
@@ -229,6 +229,7 @@ namespace Windows.UI.Xaml.Controls
 			if (_textBox != null)
 			{
 				_textBox.KeyDown += OnTextBoxKeyDown;
+				_textBox.LostFocus += OnTextBoxLostFocus;
 			}
 
 			if (_queryButton != null)
@@ -252,6 +253,7 @@ namespace Windows.UI.Xaml.Controls
 			if (_textBox != null)
 			{
 				_textBox.KeyDown -= OnTextBoxKeyDown;
+				_textBox.LostFocus -= OnTextBoxLostFocus;
 			}
 
 			if (_queryButton != null)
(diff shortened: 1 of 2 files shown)


iury-kc:
Great!! Thanks @Youssef1313 !

src/Uno.UI/UI/Xaml/Controls/AutoSuggestBox/AutoSuggestBox.cs
```diff
@@ -229,6 +229,7 @@ void RegisterEvents()
 			if (_textBox != null)
 			{
 				_textBox.KeyDown += OnTextBoxKeyDown;
+				_textBox.LostFocus += OnTextBoxLostFocus;
```
MartinZikmund, line 232: I think this solution may be problematic in case the user wants to use keyboard to navigate the suggestions - I would think that in such case the `TextBox` will lose focus and that will then cause the suggestion list to close, is that not the case? Maybe focus navigation is not yet correctly implemented in Uno for `AutoSuggestBox`

MartinZikmund: Ah right, probably not - seems the WinUI control keeps focus on the input when keys are pressed and Tab actually causes the suggestion list to close 👍

MartinZikmund (approved):

jeromelaban (approved):

PR #1796, opened 2019-10-10
Android implementation for Windows.Data.Pdf

artemious7:
GitHub Issue (If applicable): #1629 

## PR Type

What kind of change does this PR introduce?
- Feature

## What is the current behavior?
No implementation for Windows.Data.Pdf 

## What is the new behavior?
- Added implementation for Windows.Data.Pdf for Android (where a correspondent APIs available). 
- Added extensions to convert `Stream` to `IRandomAccessStream` and vice versa (which is not a real implementation, but just a stub to pass `Stream` to methods accepting only `IRandomAccessStream`).

## PR Checklist

Please check if your PR fulfills the following requirements:

- [x] Tested code with current [supported SDKs](https://github.com/unoplatform/uno/blob/master/README.md#uno-features)
(not sure what current supported SDKs means, but I have tested it on my Android project)
- [ ] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)
- [ ] [Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [ ] [Wasm UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md#running-the-webassembly-ui-tests-snapshots) are not showing unexpected any differences. Validate PR `Screenshots Compare Test Run` results.
- [ ] Contains **NO** breaking changes
- [ ] Updated the [Release Notes](https://github.com/unoplatform/uno/tree/master/doc/ReleaseNotes)
- [x] Associated with an issue (GitHub or internal)

## Other information

Internal Issue (If applicable):

Issue #1629, opened 2019-09-30
Implement Windows.Data.Pdf


artemious7:
## What would you like to be added:
Implement Windows.Data.Pdf APIs (PdfDocument, ..)

## Why is this needed:
To build a PDF-viewer app in Uno. PdfDocument allows to render PDF files to image files and display them in the app directly without using an external app or WebView.

## For which Platform:

- [x] iOS
- [x] Android
- [x] WebAssembly ? (not yet sure if it's possible)
- [ ] Windows

## Anything else we need to know?
There's already [a simple plugin for rendering PDFs cross-platform](https://github.com/dotnet-ad/Xamarin.Plugins), which wraps native APIs for Android, iOS and UWP. However, it has some issues and the project is abandoned for 4 years already. It also uses its own interface and abstractions instead of UWP ones. And it doesn't allow to choose different rendering options. 

I am going to implement this feature, because I need for my app and I need the rendering options. Can you please, provide me some guidance on how to do it right?

jeromelaban:
@artemious7 thanks for taking a stab at it!

Seems like a straight forward implementation that only rasterizes the PDF. The basic task is to take a sample that works on the platforms you need:
- Create a [set of samples](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md#working-with-the-samples-applications)
- Implement the classes that are required (from the generated files). You can copy and paste the files and remove the `[NotImplemented]` attributes where needed.

artemious7:
How do we workaround unavailability of `IRandomAccessStream` ? `Windows.Data.Pdf.PdfPage` has only this way of rendering a pdf page, requiring this type:

        public IAsyncAction RenderToStreamAsync(IRandomAccessStream outputStream, PdfPageRenderOptions options)

The caller will use Stream instead of IRandomAccessStream, and the actual implementation will also use Stream. So I made a simple workaround - extensions methods that would wrap the Stream and unwrap it.

    public static class RandomAccessStreamExtensions
    {
        public static IRandomAccessStream WrapAsRandomAccessStream(this Stream stream)
        {
            if (stream is null)
            {
                throw new ArgumentNullException(nameof(stream));
            }
            return new RandomAccessStreamWrapper(stream);
        }

        public static Stream UnwrapAsStream(this IRandomAccessStream stream)
        {
            if (stream is null)
            {
                throw new ArgumentNullException(nameof(stream));
            }

[...]

PR #9718, opened 2022-09-02
feat: Android implementation for Windows.Data.Pdf

workgroupengineering:
GitHub Issue (If applicable): Part of #1629 (This PR implements Android only), closes #1796

## PR Type

- Feature

not impemented

## What is the new behavior?

implementation for Windows.Data.Pdf on Android continuation of #1796

## PR Checklist

## Other information


workgroupengineering:
Commit b63abd55764461041e62aaf941537b3bb934082e will be dropped when PR # 9623 is merged.

src/Uno.UWP/Data/Pdf/PdfPage.Android.cs
```diff
@@ -0,0 +1,164 @@
...
+	public uint Index => (uint)pdfPage.Index;
+
+	[NotImplemented]
+	public float PreferredZoom
+	{
+		get
+		{
+			throw new NotImplementedException("The member float PdfPage.PreferredZoom is not implemented in Uno.");
+		}
+	}
```
jeromelaban, line 37:
```suggestion
	public float PreferredZoom 
	   => throw new NotImplementedException("The member float PdfPage.PreferredZoom is not implemented in Uno.");
```

Same for the similar others.

src/Uno.UWP/Data/Pdf/PdfDocument.Android.cs
```diff
@@ -0,0 +1,133 @@
...
+	public static Foundation.IAsyncOperation<PdfDocument> LoadFromStreamAsync(Storage.Streams.IRandomAccessStream inputStream, string password)
+	{
+		if (!string.IsNullOrWhiteSpace(password))
+		{
+			// password protected PDF files are not supported by PdfRenderer in Android
+			throw new NotImplementedException("The member IAsyncOperation<PdfDocument> PdfDocument.LoadFromStreamAsync(IRandomAccessStream inputStream, string password) is not implemented in Uno.");
+		}
+
+		var tcs = new TaskCompletionSource<PdfDocument>();
+		Task.Factory.StartNew(static arg =>
```
jeromelaban, line 75: Considering the low number of invocations of this particular method, using a non-static lambda will make the code a lot easier to read. Also, were you unable to use `AsyncOperation.FromTask`?

jeromelaban (commented):
Very nice !

src/Uno.UI.RuntimeTests/Tests/Windows_Data/Pdf/Given_PdfDocument.cs
```diff
@@ -0,0 +1,275 @@
...
+	private const string PdfDocument_Name_Protected = "UnoA4_Protected.pdf";
+	private const string ReferencePageImageUri = "UnoA4.png";
+	private const string ReferencePageImage_ProtectedUri = "UnoA4_Protected.png";
+
+	/// <summary>
+	/// Calculates root mean square error for given two images.
+	/// Based roughly on ImageMagick implementation to ensure consistency.
+	/// If the error is greater than or equal to 0.022, the differences are visible to human eyes.
+	/// </summary>
+	private static double CompareImages(RawBitmap actual, RawBitmap expected)
```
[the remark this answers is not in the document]
workgroupengineering: There is HasPixels but you can set the tolerance at the pixel level but not the whole image. The images, although different, are identical to the human eye. If the error returned by the function is greater than 0.022, the differences are visible to the human eye.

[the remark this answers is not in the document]
workgroupengineering: I had thought of moving it to `ImageAssert` but I am not convinced by the name `CompareImages`

src/Uno.UWP/Data/Pdf/PdfDocument.Android.cs
```diff
@@ -0,0 +1,133 @@
...
+						try
+						{
+							var pdfRenderer = new PdfRenderer(t.Result);
+							var pdfDocument = new PdfDocument(pdfRenderer);
+							context.TrySetResult(pdfDocument);
+						}
+						catch (Exception ex)
+						{
+							context.SetException(ex);
+						}
```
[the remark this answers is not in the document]
workgroupengineering: to catch any exception from the previous block and propagate it through `TaskCompletionSource.SetException`

src/Uno.UWP/Data/Pdf/PdfDocument.Android.cs
workgroupengineering: I did not know `AsyncOperation.FromTask` :sweat_smile:.  Do I use it instead of `TaskCompletionSource`?

src/Uno.UWP/Data/Pdf/PdfPage.Android.cs
```diff
@@ -0,0 +1,164 @@
...
+using Windows.UI;
+using Color = Windows.UI.Color;
+using Rect = Windows.Foundation.Rect;
+
+namespace Windows.Data.Pdf;
+
+public partial class PdfPage : IDisposable
+{
+	private readonly PdfRenderer.Page pdfPage;
+	private const float pointToPixelFactor = 96.00000000f / 72.0000000f;
```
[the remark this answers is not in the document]
workgroupengineering: To transform at the same scale as UWP

src/Uno.UWP/Data/Pdf/PdfDocument.Android.cs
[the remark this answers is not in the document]
jeromelaban: Sure but you could do this using a try/catch and use TrySetException in the catch

src/Uno.UWP/Data/Pdf/PdfDocument.Android.cs
jeromelaban: It's best to use this yes, as it also provides cancellationtoken.

src/Uno.UI.RuntimeTests/Tests/Windows_Data/Pdf/Given_PdfDocument.cs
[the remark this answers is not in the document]
dr1rrb: Yes we do already have such comparison methods here https://github.com/unoplatform/uno/blob/master/src/SamplesApp/SamplesApp.UITests/TestFramework/ImageAssert.cs#L49 They are not yet ported to the RuntimeTests but it would be great to port them :)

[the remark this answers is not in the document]
workgroupengineering: Hi @dr1rrb ,
I've seen the method you suggested, but it doesn't do exactly what I need.

AreAlmostEqual: verifies for each single pixel of the images to be compared that the sum of the absolute values of the differences between the individual components of the pixels is less than the specified tolerance.

CompareImages: Computes the [root mean square error](https://en.wikipedia.org/wiki/Root-mean-square_deviation) for two given images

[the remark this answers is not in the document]
dr1rrb: Yeah I always considered that our comparison of pixels was pretty weak. What would be nice then it to add that comparison type to the [`ColorToleranceKind`](https://github.com/unoplatform/uno/blob/f1b6f49acc01d584939b4dcc3d7dec885fd09ed1/src/SamplesApp/SamplesApp.UITests/TestFramework/ImageAssert.ExpectedPixels.cs#L256) and adjust the comparison to support such comparison (https://github.com/unoplatform/uno/blob/master/src/SamplesApp/SamplesApp.UITests/TestFramework/ImageAssert.Validations.cs#L44) So we would take advantage of it in all UI tests using the same "syntax" to specify comparison regions/rect etc.

But I agree that it might be not be an easy fix and we prefer to do that in another PR, your choice.

(Note: I would assume that the "root mean square error" would not make sense with `LocationToleranceKind.PerPixel`, I would throw an exception if user specified both).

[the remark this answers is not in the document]
workgroupengineering: Ok, I'll try to do another PR for ImageAssert in the second time. Thanks for the indications.

src/Uno.UWP/Data/Pdf/PdfPage.Android.cs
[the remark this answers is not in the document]
jeromelaban: Can you add a comment to explain it as such?

src/Uno.UI.RuntimeTests/Tests/Windows_Data/Pdf/Given_PdfDocument.cs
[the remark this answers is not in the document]
jeromelaban: It would still be best to move it to `ImageAssert`, and provide the `ImperceptibilityDeviation` value as a parameter so it can assert directly there. You can also have it call `RawImage.Populate` so it's not called in the test.

src/Uno.UWP/Generated/3.0.0.0/Windows.Data.Pdf/PdfDocument.cs
```diff
@@ -2,67 +2,67 @@
...
 	[global::Uno.NotImplemented]
-	#endif
-	public  partial class PdfDocument 
+#endif
+	public partial class PdfDocument
 	{
-		#if __ANDROID__ || __IOS__ || NET461 || __WASM__ || __SKIA__ || __NETSTD_REFERENCE__ || __MACOS__
-		[global::Uno.NotImplemented("__ANDROID__", "__IOS__", "NET461", "__WASM__", "__SKIA__", "__NETSTD_REFERENCE__", "__MACOS__")]
+#if false || __IOS__ || NET461 || __WASM__ || __SKIA__ || __NETSTD_REFERENCE__ || __MACOS__
+		[global::Uno.NotImplemented]
```
Youssef1313, line 11: Why are the attribute arguments removed? I recall it was problematic if they didn't exist (even with the `#if` above).

src/Uno.UWP/Data/Pdf/PdfPageRenderOptions.Android.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+using Windows.Foundation;
+using Windows.UI;
+
+namespace Windows.Data.Pdf;
+
+public partial class PdfPageRenderOptions
```
Youssef1313, line 7:
```suggestion
public sealed partial class PdfPageRenderOptions
```

src/Uno.UWP/Data/Pdf/PdfPageRenderOptions.Android.cs
```diff
@@ -0,0 +1,20 @@
...
+
+	public bool IsIgnoringHighContrast { get; set; }
+
+	public uint DestinationWidth { get; set; }
+
+	public uint DestinationHeight { get; set; }
+
+	public Guid BitmapEncoderId { get; set; }
+
+	public Color BackgroundColor { get; set; }
```
Youssef1313, line 19: Documentation says:

> The default background color is [White](https://docs.microsoft.com/en-us/uwp/api/windows.ui.colors.white?view=winrt-22621#windows-ui-colors-white)

src/Uno.UWP/Data/Pdf/PdfPageRenderOptions.Android.cs
```diff
@@ -0,0 +1,20 @@
...
+using Windows.Foundation;
+using Windows.UI;
+
+namespace Windows.Data.Pdf;
+
+public partial class PdfPageRenderOptions
+{
+	public Rect SourceRect { get; set; }
+
+	public bool IsIgnoringHighContrast { get; set; }
```
Youssef1313, line 11: Consider verifying whether the default is `true` or `false` on UWP. If it's `false, we're good. If it's true, we need to set it.

Youssef1313: Is this property actually used? If not, I think it should remain as NotImplemented

src/Uno.UWP/Data/Pdf/PdfPageRenderOptions.Android.cs
```diff
@@ -0,0 +1,20 @@
...
+
+	public uint DestinationWidth { get; set; }
+
+	public uint DestinationHeight { get; set; }
+
+	public Guid BitmapEncoderId { get; set; }
```
Youssef1313, line 17: Same comment here. If this property is unused, it should remain as NotImplemented

src/Uno.UWP/Data/Pdf/PdfPageDimensions.Android.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Windows.Foundation;
+
+namespace Windows.Data.Pdf;
+
+public partial class PdfPageDimensions
```
Youssef1313, line 5:
```suggestion
public sealed partial class PdfPageDimensions
```

src/Uno.UWP/Data/Pdf/PdfPageDimensions.Android.cs
```diff
@@ -0,0 +1,12 @@
...
+{
+	public Rect ArtBox { get; }
+	public Rect BleedBox { get; }
+	public Rect CropBox { get; }
+	public Rect MediaBox { get; }
+	public Rect TrimBox { get; }
```
Youssef1313, line 11: It looks like only `TrimBox` is implemented?

workgroupengineering:
What did I forget to answer? I seem to have applied all the review comments.

jeromelaban:
The build was not passing at that time, we'll need to review the Sample from the APK now :)

jeromelaban:
I've checked quickly the samples app on a device, and it works! Very nice! It should be interesting to make sure that the ScrollViewer that show the document is able to scroll horizontally, otherwise it's cropped. I'm not too sure about the API, but is there a way to show the pdf originally scaled to the available view size?

workgroupengineering:
> I've checked quickly the samples app on a device, and it works! Very nice! It should be interesting to make sure that the ScrollViewer that show the document is able to scroll horizontally, otherwise it's cropped. I'm not too sure about the API, but is there a way to show the pdf originally scaled to the available view size?

I have applied your suggestions. What are the doubts about implementing the API?

src/Uno.UWP/Data/Pdf/PdfDocument.Android.cs
```diff
@@ -0,0 +1,112 @@
...
+				byte[] buf = new byte[4096];
+				var len = 0;
+				while (!token.IsCancellationRequested && (len = input.Read(buf, 0, buf.Length)) > 0)
+				{
+					outputStream.Write(buf, 0, len);
+				}
+				outputStream.Flush();
+				outputStream.Close();
+				return ParcelFileDescriptor.Open(file, ParcelFileMode.ReadOnly);
+			}, token);
```
Youssef1313, line 111: Why this returns a Task? I think this can just return `ParcelFileDescriptor`

src/Uno.UWP/Data/Pdf/PdfDocument.Android.cs
```diff
@@ -0,0 +1,112 @@
...
+				 * PdfRenderer needs a seek-able ParcelFileDescriptor,
+				 * to get it from a stream (without using hacks),
+				 * you need to copy the stream to a temporary file.
+				*/
+				var fileName = Path.GetTempFileName();
+				var file = new Java.IO.File(fileName);
+				file.DeleteOnExit();
+				var outputStream = new Java.IO.FileOutputStream(file);
+
+				byte[] buf = new byte[4096];
```
Youssef1313, line 102: Could this be `stackalloc`ed?

src/Uno.UWP/Data/Pdf/PdfDocument.Android.cs
```diff
@@ -0,0 +1,112 @@
...
+			// password protected PDF files are not supported by PdfRenderer in Android
+			throw new NotImplementedException("The member IAsyncOperation<PdfDocument> PdfDocument.LoadFromStreamAsync(IRandomAccessStream inputStream, string password) is not implemented in Uno.");
+		}
+		return AsyncOperation.FromTask(async ct =>
+		{
+			var parcel = await GetParcelFileDescriptorFromStreamAsync(inputStream.AsStream(), ct);
+			var pdfRenderer = new PdfRenderer(parcel);
+			var pdfDocument = new PdfDocument(pdfRenderer);
+			return pdfDocument;
+		});
```
Youssef1313, line 81: If you'll make `GetParcelFileDescriptorFromStreamAsync` non-async, I think you can use it as follows, and also remove the `CancellationToken` parameter

```suggestion
		return AsyncOperation.FromTask(_ =>
		{
			var parcel = GetParcelFileDescriptorFromStream(inputStream.AsStream());
			var pdfRenderer = new PdfRenderer(parcel);
			var pdfDocument = new PdfDocument(pdfRenderer);
			return Task.FromResult(pdfDocument);
		});
```

src/Uno.UI.RuntimeTests/Tests/Windows_Data/Pdf/Given_PdfDocument.cs
```diff
@@ -0,0 +1,224 @@
...
+
+		Assert.IsNotNull(resourceFullName, $"Unable to find resource named {resourceName}.");
+
+		return assembly.GetManifestResourceStream(resourceFullName);
+	}
+
+	private static async Task CheckDocumentAsync(PdfDocument? document
+		, string referencePageImage
+		, PdfPageRenderOptions? options = default
+		, bool hasPassword = false)
```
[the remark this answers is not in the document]
workgroupengineering: I prefer this style because it is clearer when it shows the differences. eg:

```csharp
private static async Task CheckDocumentAsync(PdfDocument? document,
		string referencePageImage,
		PdfPageRenderOptions? options = default,
		bool hasPassword = false)
```
```csharp
private static async Task CheckDocumentAsync(PdfDocument? document,
		string referencePageImage,
		PdfPageRenderOptions? options = default,
		bool hasPassword = false,
                bool extra = true)
```

```diff
private static async Task CheckDocumentAsync(PdfDocument? document,
		string referencePageImage,
		PdfPageRenderOptions? options = default,
-               bool hasPassword = false)
+	        bool hasPassword = false,
+               bool extra = true)
```

```csharp
private static async Task CheckDocumentAsync(PdfDocument? document
		, string referencePageImage
		, PdfPageRenderOptions? options = default
		, bool hasPassword = false
)
```
```csharp
private static async Task CheckDocumentAsync(PdfDocument? document
		, string referencePageImage
		, PdfPageRenderOptions? options = default
		, bool hasPassword = false
                , bool extra = true
)
```

```diff
private static async Task CheckDocumentAsync(PdfDocument? document
		, string referencePageImage
		, PdfPageRenderOptions? options = default
		, bool hasPassword = false
+               , bool extra = true
)
```

in the second case it is immediately identified that only the extra parameter has been added.

I don't know what one's policy is for these situations.

Tell me if I can stay like this or I change it.

src/Uno.UWP/Data/Pdf/PdfDocument.Android.cs
workgroupengineering: Because with very large pdfs, copying the stream to the temporary file could take a long time.

src/Uno.UI.RuntimeTests/Tests/Windows_Data/Pdf/Given_PdfDocument.cs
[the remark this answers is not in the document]
Youssef1313: I can see the benefit of your style. But that's not common and is less consistent with the rest of the codebase.

src/Uno.UWP/Data/Pdf/PdfDocument.Android.cs
workgroupengineering: I verified. I can't use it because [Write(ReadOnlySpan<Byte>)](https://learn.microsoft.com/en-us/dotnet/api/system.io.filestream.write?view=net-6.0#system-io-filestream-write(system-readonlyspan((system-byte)))) is not supported in Xamarin.Android.

Do you have any suggestions?

src/Uno.UI.RuntimeTests/Tests/Windows_Data/Pdf/Given_PdfDocument.cs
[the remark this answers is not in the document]
jeromelaban: Placing commas is common in other languages, it allows for commenting lines more easily without having to modify the previous one. I don't mind that syntax, using it myself :)

src/Uno.UWP/Data/Pdf/PdfDocument.Android.cs
```diff
@@ -0,0 +1,112 @@
...
+public partial class PdfDocument : IDisposable
+{
+	private readonly PdfRenderer _pdfRenderer;
+
+	private PdfDocument(PdfRenderer pdfRenderer)
+	{
+		this._pdfRenderer = pdfRenderer ?? throw new ArgumentNullException(nameof(pdfRenderer));
+	}
+
+	public bool IsPasswordProtected => false;
```
Youssef1313, line 20: Is this property always false?

src/Uno.UWP/Data/Pdf/PdfDocument.Android.cs
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	public static IAsyncOperation<PdfDocument> LoadFromFileAsync(Storage.IStorageFile file)
+	{
+		return LoadFromFileAsync(file, null);
+	}
+
+	public static IAsyncOperation<PdfDocument> LoadFromFileAsync(Storage.IStorageFile file, string password)
+	{
+		if (!string.IsNullOrWhiteSpace(password))
```
Youssef1313, line 42:
```suggestion
		if (password is not null)
```

src/Uno.UWP/Data/Pdf/PdfDocument.Android.cs
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	public static IAsyncOperation<PdfDocument> LoadFromStreamAsync(Storage.Streams.IRandomAccessStream inputStream)
+	{
+		return LoadFromStreamAsync(inputStream, default);
+	}
+
+	public static IAsyncOperation<PdfDocument> LoadFromStreamAsync(Storage.Streams.IRandomAccessStream inputStream, string password)
+	{
+		if (!string.IsNullOrWhiteSpace(password))
```
Youssef1313, line 70:
```suggestion
		if (password is not null)
```

src/Uno.UWP/Data/Pdf/PdfPage.Android.cs
```diff
@@ -0,0 +1,146 @@
...
+		if (outputStream is null)
+		{
+			throw new ArgumentNullException(nameof(outputStream));
+		}
+
+		var options = new PdfPageRenderOptions
+		{
+			DestinationWidth = (uint)Size.Width,
+			DestinationHeight = (uint)Size.Height
+		};
```
Youssef1313, line 45: Avoids re-calculating the size.

```suggestion
		var size = Size;
		var options = new PdfPageRenderOptions
		{
			DestinationWidth = (uint)size.Width,
			DestinationHeight = (uint)size.Height
		};
```

src/Uno.UWP/Data/Pdf/PdfPageDimensions.Android.cs
```diff
@@ -0,0 +1,18 @@
...
+	public Rect ArtBox { get; }
+	[NotImplemented("__ANDROID__")]
+	public Rect BleedBox { get; }
+	[NotImplemented("__ANDROID__")]
+	public Rect CropBox { get; }
+	[NotImplemented("__ANDROID__")]
+	public Rect MediaBox { get; }
+	[NotImplemented("__ANDROID__")]
+	public Rect TrimBox { get; }
+}
```
Youssef1313, line 18: If this is not implemented, why it's moved out of the generated file?

workgroupengineering: Because I don't want to throw an exception and report that it's not supported.

src/Uno.UWP/Data/Pdf/PdfDocument.Android.cs
workgroupengineering: Because Android PdfRenderer does not support password protected pdf. As a result, IsPasswordProtected is always false.

src/Uno.UWP/Data/Pdf/PdfDocument.Android.cs
workgroupengineering: Thank you for report. The "password is not null" is equivalent to "ReferenceEquals(null, password)". UWP considering empty string as no password set. I fix the issue replacing IsNullOrWhiteSpace with IsNullOrEmpty.

src/Uno.UWP/Data/Pdf/PdfDocument.Android.cs
workgroupengineering: Thank you for report. The "password is not null" is equivalent to "ReferenceEquals(null, password)". UWP considering empty string as no password set. I fix the issue replacing IsNullOrWhiteSpace with IsNullOrEmpty.

Youssef1313:
@workgroupengineering I updated the PR description to not close https://github.com/unoplatform/uno/issues/1629 since the issue involves iOS and possibly Wasm, not Android only.

workgroupengineering:
can someone restart the CI?

jeromelaban:
Very nice! It's getting there!

Two observations:
- Resizing of the image does not work, though I don't know if it's a ScrollViewer issue on Uno's end, or if it's some missing configuration
- The "Arrange to Width" does not seem to be working properly, where the width of the PDF is larger than the available size on screen. Is this expected?

workgroupengineering:
> Very nice! It's getting there!
> 
> Two observations:
> 
>     * Resizing of the image does not work, though I don't know if it's a ScrollViewer issue on Uno's end, or if it's some missing configuration
> 
>     * The "Arrange to Width" does not seem to be working properly, where the width of the PDF is larger than the available size on screen. Is this expected?

Thanks for the review,
I double check with UWP

workgroupengineering:
I verified.
- It seems to be a scrollviewer problem.
- I added a test to verify that scaling works correctly.

jeromelaban:
@workgroupengineering thanks for taking the time to rebase on master! We'll take a look again and see if there's anything else to adjust.

src/Uno.UWP/Data/Pdf/PdfDocument.Android.cs
```diff
@@ -0,0 +1,112 @@
...
+			{
+				/* PdfRenderer https://developer.android.com/reference/android/graphics/pdf/PdfRenderer#PdfRenderer(android.os.ParcelFileDescriptor)
+				 * PdfRenderer needs a seek-able ParcelFileDescriptor,
+				 * to get it from a stream (without using hacks),
+				 * you need to copy the stream to a temporary file.
+				*/
+				var fileName = Path.GetTempFileName();
+				var file = new Java.IO.File(fileName);
+				file.DeleteOnExit();
+				var outputStream = new Java.IO.FileOutputStream(file);
```
jeromelaban, line 100: If it is disposable (not sure), let's dispose it:

```suggestion
				using var outputStream = new Java.IO.FileOutputStream(file);
```

jeromelaban (approved):
This is looking very good! Just one small adjustment and we can merge this!

PR #9726, opened 2022-09-03
perf: Avoid Regex for indentation when possible

### LINKED ISSUE #9715: IndentedStringBuilder uses Regex to perform indentation
### Current behavior

This method is used by `IndentedStringBuilder` and causes a lot of string allocations and performance impact by the use of Regex:
https://github.com/unoplatform/uno/blob/b671f1de73245cf57265af112bd5b3c9da4e403f/src/Uno.Foundation/Uno.Core.Extensions/Uno.Core.Extensions/StringExtensions.cs#L242

The idea behind this feature is that if there are multiple lines, all these lines should be indented the same, but in most of the generator's use of `IndentedStringBuilder`  single line strings are created. We should be changing this code to either detect if there are actually new lines in the string to avoid the use of Regex, or simply make a second version of index called `IndentMultiLine` or similar which allows for single-line indentation to be used by `IndentedStringBuilder`.

Related to https://github.com/unoplatform/uno/issues/7506

Youssef1313:
GitHub Issue (If applicable): closes #9715

## PR Type

## PR Checklist

## Other information


src/SourceGenerators/Uno.UI.SourceGenerators/DependencyObject/DependencyObjectGenerator.cs
```diff
@@ -247,30 +247,30 @@ private void WriteiOSMoveToSuperView(INamedTypeSymbol typeSymbol, IndentedString
 
 				if (overridesWillMoveToSuperview)
 				{
-					builder.AppendFormatInvariant(@"
+					builder.AppendIndented(@"
```
jeromelaban, line 250: This is a multiline block. I don't think we should be having a "multiline" version of `AppendIndented`, but rather search for the first newline inside the block then switch to the multiline version if there is one. It'll make authoring a lot easier, even if at the expense of a lookup when adding the line.

Youssef1313: @jeromelaban I wanted to keep `AppendIndented` as fast as possible. That's why I introduced a specific API for multiline.

Regarding authoring, actually most multiline appends are *already* broken (with and without this PR). The passed multi-line verbatim string is already indented in most cases, and we add extra indentation to that. So the authoring was and still an issue. Let me know what do you think.

src/Uno.Foundation/Uno.Core.Extensions/Uno.Core.Extensions.Compatibility/Extensions/IndentedStringBuilder.cs
```diff
@@ -39,54 +39,62 @@ public IndentedStringBuilder(StringBuilder stringBuilder)
...
-		public virtual void AppendFormat(IFormatProvider formatProvider, string pattern, params object[] replacements)
+		public void AppendIndented(string text)
 		{
-			_stringBuilder.AppendFormat(formatProvider, pattern.Indent(CurrentLevel), replacements);
+			AppendIndented(text, CurrentLevel);
+		}
+
+		private void AppendIndented(string text, int indentCount)
+		{
+			_stringBuilder.Append(new string('\t', indentCount)); // TODO: Which is faster? this? or looping and appending a single tab each iteration?
```
ebariche, line 84: We should use this : https://docs.microsoft.com/en-us/dotnet/api/system.text.stringbuilder.append?view=net-6.0#system-text-stringbuilder-append(system-char-system-int32)

src/Uno.Foundation/Uno.Core.Extensions/Uno.Core.Extensions.Compatibility/Extensions/IndentedStringBuilder.cs
```diff
@@ -39,54 +39,62 @@ public IndentedStringBuilder(StringBuilder stringBuilder)
...
+			_stringBuilder.Append(text);
+		}
+
+		public void AppendFormatIndented(IFormatProvider formatProvider, string text, params object[] replacements)
+		{
+			_stringBuilder.Append(new string('\t', CurrentLevel)); // TODO: Which is faster? this? or looping and appending a single tab each iteration?
```
ebariche, line 90: Same here

Youssef1313: Thanks @ebariche. Fixed

src/SourceGenerators/Uno.UI.SourceGenerators/DependencyObject/DependencyObjectGenerator.cs
jeromelaban: Makes sense, let's keep as is for now then. Thanks!

jeromelaban (approved):

PR #9727, opened 2022-09-04
perf: Cache XAML parsing

### LINKED ISSUE #9716: Cache XAML parsing
### Current behavior

When using the XAML generator, in the context of Design Time, the instance of the generator is kept alive for an extended period of time, and generation performance is important. 

The XAML Generator parses XAML files on every run, but it may be that in many cases, the parsing could be cached and reused for any subsequent run.

We should be able to modify:
https://github.com/unoplatform/uno/blob/5fb1bca0614e83a5ae3a878dd48abe62a2400892/src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileParser.cs#L36

to include a cache based on path and hash of the file to provide an already parsed result, likely as a static member.

It would also be important to provide a cache pruning mechanism based on last-used time (e.g. remove the file if not used after 1 hour, TBD more precisely)

Related to https://github.com/unoplatform/uno/issues/7506

Youssef1313:
GitHub Issue (If applicable): closes #9716

## PR Type

- Performance

## PR Checklist

## Other information

### DIFF
--- a/src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileDefinition.cs
+++ b/src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileDefinition.cs
@@ -47,11 +47,6 @@ namespace Uno.UI.SourceGenerators.XamlGenerator
 			get => _shortId ?? -1;
 			set
 			{
-				if (_shortId != null)
-				{
-					throw new InvalidOperationException($"{nameof(ShortId)} should not be set more than once.");
-				}
-
 				_shortId = value;
 			}
 		}
--- a/src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileDefinition.cs
+++ b/src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileDefinition.cs
@@ -37,20 +37,6 @@ namespace Uno.UI.SourceGenerators.XamlGenerator
 		/// </summary>
 		public string UniqueID { get; }
 
-		private int? _shortId;
-
-		/// <summary>
-		/// Compact unique ID, used to name associated global members.
-		/// </summary>
-		public int ShortId
-		{
-			get => _shortId ?? -1;
-			set
-			{
-				_shortId = value;
-			}
-		}
-
 		public bool Equals(XamlFileDefinition? other)
 		{
 			if (other is null)
--- a/src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileParser.Cache.cs
+++ b/src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileParser.Cache.cs
@@ -1,0 +1,154 @@
+﻿extern alias __uno;
+#nullable enable
+
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Diagnostics;
+
+namespace Uno.UI.SourceGenerators.XamlGenerator;
+
+internal partial class XamlFileParser
+{
+	private static class Hash
+	{
+		/// <summary>
+		/// The offset bias value used in the FNV-1a algorithm
+		/// See http://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function
+		/// </summary>
+		internal const int FnvOffsetBias = unchecked((int)2166136261);
+
+		/// <summary>
+		/// The generative factor used in the FNV-1a algorithm
+		/// See http://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function
+		/// </summary>
+		internal const int FnvPrime = 16777619;
+
+		/// <summary>
+		/// Compute the FNV-1a hash of a sequence of bytes
+		/// See http://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function
+		/// </summary>
+		/// <param name="data">The sequence of bytes</param>
+		/// <returns>The FNV-1a hash of <paramref name="data"/></returns>
+		internal static int GetFNVHashCode(ImmutableArray<byte> data)
+		{
+			int hashCode = Hash.FnvOffsetBias;
+
+			for (int i = 0; i < data.Length; i++)
+			{
+				hashCode = unchecked((hashCode ^ data[i]) * Hash.FnvPrime);
+			}
+
+			return hashCode;
+		}
+	}
+
+	private static class ByteSequenceComparer
+	{
+		// https://github.com/dotnet/roslyn/blob/6f47a0b611ab83ad1b7f8b404cc74098d37e283d/src/Compilers/Core/Portable/Collections/ByteSequenceComparer.cs#L24-L45
+		internal static bool Equals(ImmutableArray<byte> x, ImmutableArray<byte> y)
+		{
+			if (x == y)
+			{
+				return true;
+			}
+
+			if (x.IsDefault || y.IsDefault || x.Length != y.Length)
+			{
+				return false;
+			}
+
+			for (var i = 0; i < x.Length; i++)
+			{
+				if (x[i] != y[i])
+				{
+					return false;
+				}
+			}
+
+			return true;
+		}
+
+		// https://github.com/dotnet/roslyn/blob/6f47a0b611ab83ad1b7f8b404cc74098d37e283d/src/Compilers/Core/Portable/Collections/ByteSequenceComparer.cs#L101-L105
+		internal static int GetHashCode(ImmutableArray<byte> x)
+		{
+			Debug.Assert(!x.IsDefault);
+			return Hash.GetFNVHashCode(x);
+		}
+	}
+
+	private struct CachedFileKey : IEquatable<CachedFileKey>
+	{
+		public CachedFileKey(string includeXamlNamespaces, string excludeXamlNamespaces, string file, ImmutableArray<byte> checksum)
+		{
+			IncludeXamlNamespaces = includeXamlNamespaces;
+			ExcludeXamlNamespaces = excludeXamlNamespaces;
+			File = file;
+			Checksum = checksum;
+		}
+
+		public string IncludeXamlNamespaces { get; }
+		public string ExcludeXamlNamespaces { get; }
+		public string File { get; }
+		public ImmutableArray<byte> Checksum { get; }
+
+		public override bool Equals(object? obj) => obj is CachedFileKey key && Equals(key);
+
+		public bool Equals(CachedFileKey other) =>
+			IncludeXamlNamespaces == other.IncludeXamlNamespaces &&
+			ExcludeXamlNamespaces == other.ExcludeXamlNamespaces &&
+			File == other.File &&
+			ByteSequenceComparer.Equals(Checksum, other.Checksum);
+
+		public override int GetHashCode()
+		{
+			var hashCode = -145098327;
+			hashCode = hashCode * -1521134295 + EqualityComparer<string>.Default.GetHashCode(IncludeXamlNamespaces);
+			hashCode = hashCode * -1521134295 + EqualityComparer<string>.Default.GetHashCode(ExcludeXamlNamespaces);
+			hashCode = hashCode * -1521134295 + EqualityComparer<string>.Default.GetHashCode(File);
+			hashCode = hashCode * -1521134295 + ByteSequenceComparer.GetHashCode(Checksum);
+			return hashCode;
+		}
+
+		public static bool operator ==(CachedFileKey left, CachedFileKey right) => left.Equals(right);
+
+		public static bool operator !=(CachedFileKey left, CachedFileKey right) => !(left == right);
+	}
+
+	private struct CachedFile : IEquatable<CachedFile>
+	{
+		public DateTimeOffset LastTimeUsed { get; }
+		public XamlFileDefinition XamlFileDefinition { get; }
+
+		public CachedFile(DateTimeOffset lastTimeUsed, XamlFileDefinition xamlFileDefinition)
+		{
+			LastTimeUsed = lastTimeUsed;
+			XamlFileDefinition = xamlFileDefinition;
+		}
+
+		public CachedFile WithUpdatedLastTimeUsed()
+		{
+			return new CachedFile(DateTimeOffset.Now, XamlFileDefinition);
+		}
+
+		public override bool Equals(object? obj)
+			=> obj is CachedFile file && Equals(file);
+
+		public bool Equals(CachedFile other)
+		{
+			return LastTimeUsed.Equals(other.LastTimeUsed) &&
+				EqualityComparer<XamlFileDefinition>.Default.Equals(XamlFileDefinition, other.XamlFileDefinition);
+		}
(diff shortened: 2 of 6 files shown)


Youssef1313 (commented):
We need to do some manual testing to make sure this works in a good way.

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileParser.cs
```diff
@@ -2,25 +2,46 @@
...
 
 namespace Uno.UI.SourceGenerators.XamlGenerator
 {
 	internal class XamlFileParser
 	{
+		private struct CachedFile
+		{
+			public string[] IncludeXamlNamespaces { get; }
+			public string[] ExcludeXamlNamespaces { get; }
+			public DateTime LastTimeUsed { get; }
```
jeromelaban, line 27: Let's use `DateTimeOffset` instead

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileParser.cs
```diff
@@ -2,25 +2,46 @@
...
 using Uno.Roslyn;
+using Uno.UI.SourceGenerators.XamlGenerator.Utils;
+using Uno.UI.SourceGenerators.XamlGenerator.XamlRedirection;
+
 
 namespace Uno.UI.SourceGenerators.XamlGenerator
 {
 	internal class XamlFileParser
 	{
+		private struct CachedFile
```
jeromelaban, line 23: We also need to validate the actual hash of the file's contents, otherwise we will not get file updates. We can build the `GetHashCode`/`Equals` overrides based on `IncludeXamlNamespaces`, `ExcludeXamlNamespaces`, the file content hash and the file path so that we can use it to get the value from the dictionary.

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileParser.cs
```diff
@@ -2,25 +2,102 @@
...
+			public override bool Equals(object? obj)
+				=> obj is CachedFileKey key && Equals(key);
+
+			public bool Equals(CachedFileKey other)
+			{
+				return EqualityComparer<string[]>.Default.Equals(IncludeXamlNamespaces, other.IncludeXamlNamespaces) &&
+					EqualityComparer<string[]>.Default.Equals(ExcludeXamlNamespaces, other.ExcludeXamlNamespaces) &&
+					File == other.File &&
+					ContentsHash == other.ContentsHash;
+			}
```
Youssef1313, line 47: Do I need to compare the actual contents here (in case of hash collisions?)

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileParser.cs
```diff
@@ -20,28 +20,84 @@ namespace Uno.UI.SourceGenerators.XamlGenerator
...
+				return LastTimeUsed.Equals(other.LastTimeUsed) &&
+					EqualityComparer<XamlFileDefinition>.Default.Equals(XamlFileDefinition, other.XamlFileDefinition);
+			}
+
+			public override int GetHashCode()
+			{
+				var hashCode = 368878473;
+				hashCode = hashCode * -1521134295 + LastTimeUsed.GetHashCode();
+				hashCode = hashCode * -1521134295 + EqualityComparer<XamlFileDefinition>.Default.GetHashCode(XamlFileDefinition);
+				return hashCode;
```
jeromelaban, line 93: Which implementation is this based on? (the numbers specifically)

Youssef1313: Roslyn feature in Visual Studio to generate Equals and GetHashCode. Just <kbd>Ctrl</kbd>+<kbd>.</kbd> on struct name that doesn't have any equality contracts (interface, operators, object Equals override, and object GetHashCode override), and VS will generate this.

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileParser.cs
```diff
@@ -2,25 +2,102 @@
...
 using Uno.Roslyn;
+using Uno.UI.SourceGenerators.XamlGenerator.Utils;
+using Uno.UI.SourceGenerators.XamlGenerator.XamlRedirection;
+
 
 namespace Uno.UI.SourceGenerators.XamlGenerator
 {
 	internal class XamlFileParser
 	{
+		private struct CachedFileKey : IEquatable<CachedFileKey>
```
jeromelaban, line 23: Let's move those two internals in a separate file partial to `XamlFileParser`, like `XamlFileParser.Cache.cs`.

Youssef1313: Sure

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileParser.cs
Youssef1313: fwiw, it seems that `-1521134295` is hard-coded in Roslyn, while the initial `hashCode` value (`368878473`) is not hard-coded.

Feature tests are in https://github.com/dotnet/roslyn/blob/main/src/EditorFeatures/CSharpTest/GenerateFromMembers/GenerateEqualsAndGetHashCodeFromMembers/GenerateEqualsAndGetHashCodeFromMembersTests.cs

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileParser.Cache.cs
```diff
@@ -0,0 +1,88 @@
+﻿extern alias __uno;
+#nullable enable
+
+using System;
+using System.Collections.Generic;
+
+namespace Uno.UI.SourceGenerators.XamlGenerator
```
jeromelaban, line 7:
```suggestion
namespace Uno.UI.SourceGenerators.XamlGenerator;
```

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileParser.cs
```diff
@@ -45,6 +43,11 @@ public XamlFileDefinition[] ParseFiles(string[] xamlSourceFiles, System.Threadin
 				.ToArray()!;
 		}
 
+		private static void ClearCache()
+		{
+			_cachedFiles.Remove(kvp => DateTimeOffset.Now - kvp.Value.LastTimeUsed > new TimeSpan(hours: 1, minutes: 0, seconds: 0));
```
jeromelaban, line 48: Let's extract the TimeSpan a readonly static value.

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileParser.Cache.cs
```diff
@@ -0,0 +1,88 @@
...
+				IncludeXamlNamespaces = includeXamlNamespaces;
+				ExcludeXamlNamespaces = excludeXamlNamespaces;
+				File = file;
+				ContentsHash = contentsHash;
+			}
+
+			public string[] IncludeXamlNamespaces { get; }
+			public string[] ExcludeXamlNamespaces { get; }
+			public string File { get; }
+			public int ContentsHash { get; }
```
jeromelaban, line 24: Let's add the original file last modified time as well (if we can). I'm not sure if `AdditionalFile` supports that.

Youssef1313: There is no way to get that from `AdditionalFiles`.

I also think this can break the cache if a modification has been made then reverted. The last modified time will change but the contents are not and the cache should ideally be used.

jeromelaban: Interesting, then maybe that's when `AdditionalFiles` makes sense, since `AdditionalText` has a `GetChecksum` method we probably should be using it for the cache instead of building it manually.

Youssef1313: Interesting. Though I can't do that until `Uno.SourceGeneration` pushes a new version with `AdditionalFiles` support.

Youssef1313: Just figured out that Uno.SourceGeneration already pushed a dev version to azure feed (I was looking on NuGet.org). I'll prepare a separate PR to use AdditionalFiles and after that I'll get back to this

jeromelaban: I've just pushed the new build to nuget.org

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileParser.Cache.cs
```diff
@@ -0,0 +1,86 @@
...
+		public string File { get; }
+		public ImmutableArray<byte> Checksum { get; }
+
+		public override bool Equals(object? obj) => obj is CachedFileKey key && Equals(key);
+
+		public bool Equals(CachedFileKey other) =>
+			IncludeXamlNamespaces == other.IncludeXamlNamespaces &&
+			ExcludeXamlNamespaces == other.ExcludeXamlNamespaces &&
+			File == other.File &&
+			Checksum.Equals(other.Checksum);
```
Youssef1313, line 33: TODO: This is a reference equality for the underlying array of `ImmutableArray`.
This might not be correct.

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileParser.Cache.cs
```diff
@@ -0,0 +1,86 @@
...
+namespace Uno.UI.SourceGenerators.XamlGenerator;
+
+internal partial class XamlFileParser
+{
+	private struct CachedFileKey : IEquatable<CachedFileKey>
+	{
+		public CachedFileKey(string includeXamlNamespaces, string excludeXamlNamespaces, string file, ImmutableArray<byte> checksum)
+		{
+			IncludeXamlNamespaces = includeXamlNamespaces;
+			ExcludeXamlNamespaces = excludeXamlNamespaces;
```
Youssef1313, line 17: Storing these as `string`s here instead of `string[]` for proper comparison and hash code.

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileParser.cs
```diff
@@ -14,26 +16,34 @@
...
-			_excludeXamlNamespaces = excludeXamlNamespaces;
-			_includeXamlNamespaces = includeXamlNamespaces;
-			this._metadataHelper = roslynMetadataHelper;
+			_excludeXamlNamespacesProperty = excludeXamlNamespaces;
+			_excludeXamlNamespaces = excludeXamlNamespaces.Split(new[] { ',' }, StringSplitOptions.RemoveEmptyEntries);
+
+			_includeXamlNamespacesProperty = includeXamlNamespaces;
+			_includeXamlNamespaces = includeXamlNamespaces.Split(new[] { ',' }, StringSplitOptions.RemoveEmptyEntries);
+			
+			_metadataHelper = roslynMetadataHelper;
```
Youssef1313, line 46: @jeromelaban `_metadataHelper` is used in parsing for conditional namespaces, for example:

https://github.com/unoplatform/uno/blob/88d7a4024ccdcaa07a6b8cd056ba55ef4e92865e/src/Uno.UI.Tests/Windows_UI_Xaml/Controls/When_Conditional_IsTypePresent_Page.xaml#L5

It's not clear to me if it's possible to have a wrong cache entry due to that?

jeromelaban: Good catch! It theoretically could, particularly if the cache is reused between versions of Uno. This feature only works on windows when used with types coming from UWP/WinUI itself, but not for other types. One way to fix this would be to include the information version number of the `Uno.UI` assembly (when building outside of Uno.UI itself) and use it as a cache key as well.

Youssef1313: @jeromelaban Is `Uno.UI` version the only factor here? I thought things like conditional directives, etc, could affect the outcome of `IsTypePresent` as well?

One idea is to not cache files using `IsTypePresent`, I assume most files don't use it so we'll still do a good amount of caching. Can you confirm if this approach sounds good?

Youssef1313:
@jeromelaban Is this ready to merge?

jeromelaban (approved):

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileDefinition.cs
```diff
@@ -47,11 +47,6 @@ public int ShortId
 			get => _shortId ?? -1;
 			set
 			{
-				if (_shortId != null)
-				{
-					throw new InvalidOperationException($"{nameof(ShortId)} should not be set more than once.");
-				}
```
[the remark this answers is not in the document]
jeromelaban: @Youssef1313 is this still valid ?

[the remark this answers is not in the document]
jeromelaban: I think this may cause race conditions indeed. We probably should be removing this completely and use a Guid instead which is compatible with :
https://github.com/unoplatform/uno/blob/07f529ff81d03f90552d4321bd43c2b2dbab123a/src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.cs#L1409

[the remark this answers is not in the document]
Youssef1313: I used the existing UniqueID instead of adding a Guid property, but let me know if you think a guid is better.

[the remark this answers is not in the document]
jeromelaban: It's pretty much the same at this point, it's a stable ID, only that it's a bit large. Let's take a look at what the assemblies look like with that new ID.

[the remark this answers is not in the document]
jeromelaban: It looks like this:
image

But overall, I think this line should not use any ID because it's already in a class that's defined with the same ID and that the Index is good enough to discriminate. Let's try doing this instead:

```csharp
 return $"Get_{index}"; 
```

jeromelaban (approved):

PR #9735, opened 2022-09-05
Centralized versions properties

Youssef1313:

## PR Type

## PR Checklist

## Other information


MartinZikmund:
Why cannot the versions be put in `Directory.Build.targets` instead 🤔 ?

Youssef1313:
That's an option, it's already used in the repo, though we are not using it for all packages.

One downside of that is if we use the same version of a given package for all projects, except a single project. ie, suppose we have `<PackageReference Include="MyPackage" />` in 50 project, but one project have `<PackageReference Include="MyPackage" Version="1.0.0" />`

I assume `1.0.0` won't be used, and the version defined as `<PackageReference Update="MyPackage" Version=".." />` in `Directory.Build.targets` will be used?

Note: the approach of `Versions.props` is used a lot in dotnet organization repositories (almost in every repo?), so there might be some other benefits. I'll try to ask about that.

MartinZikmund:
@Youssef1313 I think that one disadvantage of this could be that Dependabot will probably not be able to update packages in this format (while it can handle `Directory.Build.targets`). But in either case - it would be best to use just one approach - so if we do switch to `Versions.props`, we should do so for all packages to make things simpler.

Youssef1313:
Good point about dependababot. I'm not really sure if it can handle this or not, but it's possible that they already support that. One other point I'm not sure about is the order where `<PackageReference Include="..."`s and `<PackageReference Update="..."`s are imported by MSBuild. I think the order might matter. We want `Include`s to be first. Is that always the case?

I'll wait on this PR until hearing from @jeromelaban.
The changes here were automated by a tool I wrote, so it shouldn't me much tedious to tweak it to whatever approach we choose here.

Here is the code for reference (I didn't consider code-quality much while writing as I'm expecting this tool be a one-off usage :smile:)

```csharp
using System.Diagnostics;
using System.Text;
using System.Xml.Linq;

var targets = Directory.EnumerateFiles(@"C:\Users\PC\Desktop\uno\", "*.targets", SearchOption.AllDirectories);
var props = Directory.EnumerateFiles(@"C:\Users\PC\Desktop\uno\", "*.props", SearchOption.AllDirectories);
var csproj = Directory.EnumerateFiles(@"C:\Users\PC\Desktop\uno\", "*.csproj", SearchOption.AllDirectories);

var allXmls = targets.Concat(props).Concat(csproj).ToArray();

var updates = new Dictionary<string, ReferenceTypePair<HashSet<string>, HashSet<XElement>>>();
var includes = new Dictionary<string, ReferenceTypePair<HashSet<string?>, HashSet<XElement>>>();

var documents = new Dictionary<XDocument, string>();
var updatedDocuments = new HashSet<XDocument>();

foreach (var xmlFile in allXmls)
{
    if (xmlFile.Contains(@"\obj\"))
    {
        continue;
    }

    if (xmlFile.StartsWith(@"C:\Users\PC\Desktop\uno\src\SolutionTemplate\"))
    {
        continue;
    }

    var document = XDocument.Load(xmlFile, LoadOptions.PreserveWhitespace);
    documents.Add(document, xmlFile);
    var packageReferences = document.Descendants("PackageReference").ToArray();

    foreach (var reference in packageReferences)
    {
        var include = reference.Attribute("Include");
        var update = reference.Attribute("Update");
        Debug.Assert(include is null ^ update is null);
        var version = reference.Attribute("Version")?.Value ?? reference.Element("Version")?.Value ?? reference.Attribute("version")?.Value ?? reference.Element("version")?.Value;
        if (include is null)
        {
            Debug.Assert(update is not null);
            Debug.Assert(version is not null);
            tryAdd(updates, update.Value.Trim(), version, reference);
        }
        else if (update is null)
        {
            Debug.Assert(include is not null);
            tryAdd(includes, include.Value.Trim(), version, reference);
        }
    }
}

var versionsFile = new StringBuilder();
versionsFile.AppendLine("<Project>");
versionsFile.AppendLine("  <PropertyGroup>");
foreach (var include in includes.Where(kvp => kvp.Value.Item1.Count == 1))
{
    var versionsPropertyName = include.Key.Replace(".", "") + "Version";
    var versionsPropertyValue = include.Value.Item1.Single();
    if (versionsPropertyValue is null)
    {
        if (updates[include.Key].Item1.Count == 0)
        {
            Debug.Fail("No version was found...");
        }
        else if (updates[include.Key].Item1.Count == 1)
        {
            versionsPropertyValue = updates[include.Key].Item1.Single();
        }
        else
        {
            continue;
        }
    }
    
    foreach (var xelement in include.Value.Item2)
    {
        updatedDocuments.Add(xelement.Document!);
        if (xelement.Attribute("Version") is { } versionAttribute)
        {
            versionAttribute.Value = $"$({versionsPropertyName})";
        }
        else if (xelement.Element("Version") is { } elementVersion)
        {
            elementVersion.Value = $"$({versionsPropertyName})";
        }
        else
        {
            xelement.SetAttributeValue("Version", $"$({versionsPropertyName})");
        }
    }

    versionsFile.AppendLine($"    <{versionsPropertyName}>{versionsPropertyValue}</{versionsPropertyName}>");
}

versionsFile.AppendLine("  </PropertyGroup>");
versionsFile.AppendLine("</Project>");

foreach (var updatedDocument in updatedDocuments)
{
    File.WriteAllText(documents[updatedDocument], updatedDocument.ToString(SaveOptions.DisableFormatting), Encoding.UTF8);
}

File.WriteAllText("C:\\Users\\PC\\Desktop\\uno\\Versions.props", versionsFile.ToString(), Encoding.UTF8);

void tryAdd<T>(Dictionary<string, ReferenceTypePair<HashSet<T>, HashSet<XElement>>> dictionary, string key, T version, XElement xelement)
{
    if (dictionary.TryGetValue(key, out var tuple))
    {
        tuple.Item1.Add(version);
        tuple.Item2.Add(xelement);
    }
    else
    {
        var versions = new HashSet<T>();
        versions.Add(version);

        var xelements = new HashSet<XElement>();
        xelements.Add(xelement);

        dictionary.Add(key, new ReferenceTypePair<HashSet<T>, HashSet<XElement>>(versions, xelements));
    }
}

class ReferenceTypePair<T1, T2>
{
    public ReferenceTypePair(T1 item1, T2 item2)
    {
        Item1 = item1;
        Item2 = item2;
    }

    public T1 Item1 { get; set; }
    public T2 Item2 { get; set; }
}
```

Youssef1313:
@MartinZikmund @jeromelaban Also there is a NuGet feature for that:

https://docs.microsoft.com/en-us/nuget/consume-packages/central-package-management

But I'm not sure if we should use it. It's still preview as far as I know.

Youssef1313:
> Note: the approach of `Versions.props` is used a lot in dotnet organization repositories (almost in every repo?), so there might be some other benefits. I'll try to ask about that.

Got an answer from Roslyn team member about that. `Update` didn't exist when they started using the `Versions.props` pattern.


image

image

So it's most likely we will discard the approach I started in this PR. But whether to use `Update` or use a preview feature is something we need to discuss with @jeromelaban, I think.

jeromelaban:
> Also there is a NuGet feature for that

That's correct, we'll end up using the nuget central version management but since it has to have correct support in the VS IDE, I'd rather see it out of preview to have us use it. Also as @MartinZikmund said, we can't really be super fancy about this because of Dependabot :)

For now, the update approach is good enough even though it does not work automatically properly for legacy projects.

Youssef1313:
Replaced by https://github.com/unoplatform/uno/pull/9883

PR #9834, opened 2022-09-14
perf: Misc performance improvements for generators

Youssef1313:

## PR Type

## PR Checklist

## Other information


src/SourceGenerators/Uno.UI.SourceGenerators/Helpers/SymbolExtensions.cs
```diff
@@ -371,33 +357,17 @@ public static IEnumerable<INamedTypeSymbol> GetAllInterfaces(this ITypeSymbol? s
...
+					yield return @interface;
+				}
 			}
 		}
 
 		public static bool IsNullable(this ITypeSymbol type)
 		{
 			return ((type as INamedTypeSymbol)?.IsGenericType ?? false)
-				&& type.OriginalDefinition.ToDisplayString().Equals("System.Nullable<T>", StringComparison.OrdinalIgnoreCase);
+				&& type.OriginalDefinition.SpecialType == SpecialType.System_Nullable_T;
```
Youssef1313, line 370: `ToDisplayString` is a bit slow. Comparing SpecialType is both more reliable and more efficient.

src/SourceGenerators/Uno.UI.SourceGenerators/Helpers/HashBuilder.cs
```diff
@@ -15,8 +15,8 @@ internal class HashBuilder
 		/// <summary>
 		/// Build a stable ID from the provided symbol
 		/// </summary>
-		public static string BuildIDFromSymbol(ITypeSymbol typeSymbol)
-			=> typeSymbol.GetFullName() + "_" + Build(typeSymbol.MetadataName + typeSymbol.ContainingAssembly.MetadataName);
+		public static string BuildIDFromSymbol(INamedTypeSymbol typeSymbol)
+			=> typeSymbol.GetFullMetadataNameForFileName() + "_" + Build(typeSymbol.MetadataName + typeSymbol.ContainingAssembly.MetadataName);
```
Youssef1313, line 19: `GetFullName` is too slow. Other callsites should possibly be reviewed for performance improvements.

src/SourceGenerators/Uno.UI.SourceGenerators/Helpers/SymbolExtensions.cs
```diff
@@ -371,33 +357,17 @@ public static IEnumerable<INamedTypeSymbol> GetAllInterfaces(this ITypeSymbol? s
 					yield return (INamedTypeSymbol)symbol;
 				}
 
-				do
+				foreach (var @interface in symbol.AllInterfaces)
```
Youssef1313, line 360: AllInterfaces should be faster

src/Uno.Foundation/Uno.Core.Extensions/Uno.Core.Extensions/StringExtensions.cs
```diff
@@ -28,11 +28,6 @@ namespace Uno.Extensions
 {
 	internal static partial class StringExtensions
 	{
-#if (!SILVERLIGHT && !WINDOWS_UWP && HAS_COMPILEDREGEX) || WINDOWS_PHONE || HAS_COMPILEDREGEX
-        private static readonly Regex _newLineRegex = new Regex(@"^", RegexOptions.Compiled | RegexOptions.Multiline);
-#else
-		private static readonly Lazy<Regex> _newLineRegex = new Lazy<Regex>(() => new Regex(@"^", RegexOptions.Multiline));
-#endif
```
Youssef1313, line 35: 🎉 🎉

src/SourceGenerators/Uno.UI.SourceGenerators/Helpers/DependenciesInitializer.cs
```diff
@@ -33,7 +33,7 @@ internal static void Init()
 
 					var files = from file in Directory.EnumerateFiles(basePath, "*.dll")
 								let fileName = Path.GetFileName(file)
-								where fileName.StartsWith("Uno.")
+								where fileName.StartsWith("Uno.") || fileName == "Microsoft.CodeAnalysis.AnalyzerUtilities.dll"
```
jeromelaban, line 36: Why is this needed? Let's add a comment.

src/Uno.Foundation/Uno.Core.Extensions/Uno.Core.Extensions.Compatibility/Extensions/IndentedStringBuilder.cs
```diff
@@ -25,6 +25,55 @@ namespace Uno.Extensions
 	/// </summary>
 	internal sealed class IndentedStringBuilder : IIndentedStringBuilder
 	{
+		// https://www.meziantou.net/split-a-string-into-lines-without-allocation.htm
+		// Must be a ref struct as it contains a ReadOnlySpan<char>
+		private ref struct LineSplitEnumerator
```
jeromelaban, line 30: Looks like it's very useful! It would make sense to put it in the Uno.Extensions folder: https://github.com/unoplatform/uno/tree/master/src/Uno.Foundation/Uno.Core.Extensions/Uno.Core.Extensions

src/Uno.Foundation/Uno.Core.Extensions/Uno.Core.Extensions/StringExtensions.cs
jeromelaban: Nice!

jeromelaban (approved):

src/SourceGenerators/Uno.UI.SourceGenerators/Helpers/DependenciesInitializer.cs
Youssef1313: I faced an error that this dll couldn't be found, so I loaded it here.

src/Uno.Foundation/Uno.Core.Extensions/Uno.Core.Extensions.Compatibility/Extensions/IndentedStringBuilder.cs
```diff
@@ -100,16 +150,18 @@ public void AppendLine()
...
-		/// newline, call <see cref="AppendLine()"/> after this method.
-		/// </remarks>
 		public void AppendMultiLineIndented(string text)
 		{
-			_stringBuilder.Append(text.Indent(CurrentLevel));
+			// LineSplitEnumerator is a struct so there is no allocation here
+			var enumerator = new LineSplitEnumerator(text.AsSpan());
+			foreach (var line in enumerator)
+			{
+				AppendIndented(line.ToString());
```
Youssef1313, line 162: TODO: Make sure `.ToString()` isn't causing too much overhead. It looks like converting a span back to string isn't cheap, we need to compare this with the regex approach.

workgroupengineering: Have you see [this](https://docs.microsoft.com/en-us/dotnet/api/system.text.stringbuilder.append?view=net-6.0#system-text-stringbuilder-append(system-readonlyspan((system-char)))).

Youssef1313: @workgroupengineering This is not available on .NET Framework or .NET Standard 2.0

I'm looking for two alternatives, but need to do proper benchmarking.

```csharp
            unsafe
            {
                fixed (char* ptr = &MemoryMarshal.GetReference(span))
                {
                    builder.Append(ptr, span.Length);
                }
            }
```

or

```csharp
            var array = ArrayPool<char>.Shared.Rent(text.Length);
            text.CopyTo(array);
            builder.Append(array);
            ArrayPool<char>.Shared.Return(array);
```

src/SourceGenerators/Uno.UI.SourceGenerators/Helpers/HashBuilder.cs
```diff
@@ -28,25 +37,32 @@ public static string BuildIDFromSymbol(ITypeSymbol typeSymbol)
 		/// <param name="input">The string to hash</param>
 		/// <param name="algoritm">The algorithm to use</param>
 		/// <returns>A hex-formatted string of the hash</returns>
-		public static string Build(string input, HashAlgorithm algoritm = null)
+		public static string Build(string input)
```
Youssef1313, line 40: #### Small improvements

1. We were doing unnecessary null check. We always use SHA256 so I removed the optional `algorithm` parameter
2. We were calling `Math.Min` on every iteration. Now the result is stored in a local

#### Larger improvements

We were allocating `StringBuilder` on every call. Then on every iteration we allocate a string of Length 2 containing the hex representation of the current byte.
Finally, at the end we allocate a string from the `StringBuilder`.
Now, we only allocate the final string, and use static fields for lookup.

src/SourceGenerators/Uno.UI.SourceGenerators/DependencyObject/DependencyObjectGenerator.cs
```diff
@@ -140,29 +140,30 @@ private void ProcessType(INamedTypeSymbol typeSymbol)
...
-					builder.AppendLineIndented($"using Uno.UI.Controls;");
-					builder.AppendLineIndented($"using Uno.UI.DataBinding;");
-					builder.AppendLineIndented($"using Windows.UI.Xaml;");
-					builder.AppendLineIndented($"using Windows.UI.Xaml.Data;");
-					builder.AppendLineIndented($"using Uno.Diagnostics.Eventing;");
-					builder.AppendLineIndented("#if __MACOS__");
-					builder.AppendLineIndented("using AppKit;");
-					builder.AppendLineIndented("#endif");
+					builder.Append(@"// <auto-generated>
+// ******************************************************************
```
Sergio0694, line 144: Not familiar with this project setup, but I see it's using the 4.4 toolset, doesn't that mean we can also use raw multiline string literals here then? Would make this bit a bit nicer, with proper indentation in the code 🙂

src/SourceGenerators/Uno.UI.SourceGenerators/DependencyObject/DependencyObjectGenerator.cs
```diff
@@ -455,31 +456,31 @@ private void WriteAttachToWindow(INamedTypeSymbol typeSymbol, IndentedStringBuil
 					WriteRegisterLoadActions(typeSymbol, builder);
 
 					builder.AppendMultiLineIndented($@"
```
Sergio0694, line 458: Same as above and in the rest of the PR, could we use a raw multiline string literal with `$$` to avoid the `{{`-s?

src/SourceGenerators/Uno.UI.SourceGenerators/Helpers/HashBuilder.cs
```diff
@@ -1,22 +1,27 @@
...
 using System.Security.Cryptography;
 using System.Text;
-using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
 
 namespace Uno.UI.SourceGenerators.Helpers
 {
 	internal class HashBuilder
 	{
+		private static uint[] _Lookup32 = Enumerable.Range(0, 256).Select(i =>
```
Sergio0694, line 12: Make this readonly so it can get optimized at Tier-1 time.

src/SourceGenerators/Uno.UI.SourceGenerators/Helpers/HashBuilder.cs
```diff
@@ -1,22 +1,27 @@
...
+		{
+			string s = i.ToString("x2");
+			return ((uint)s[0]) + ((uint)s[1] << 16);
```
Sergio0694, line 15: This could be doable without the string allocation, but probably isn't worth the effort given it's a one-time initialization thing anyway.

src/SourceGenerators/Uno.UI.SourceGenerators/Helpers/HashBuilder.cs
```diff
@@ -1,22 +1,27 @@
...
+		}).ToArray();
+
+		private static readonly unsafe uint* _lookup32UnsafeP = (uint*)GCHandle.Alloc(_Lookup32, GCHandleType.Pinned).AddrOfPinnedObject();
```
Sergio0694, line 18: I'm not sure I understand the point of this. I mean the memory leak isn't an issue given this will live for the whole process duration, but it's still not ideal, and I'm not sure what problem is it trying to solve. If that's just a lookup table, what's the point of pinning it here in the first place? You can just index it directly as an array, no?

src/SourceGenerators/Uno.UI.SourceGenerators/Helpers/HashBuilder.cs
```diff
@@ -28,25 +33,32 @@ public static string BuildIDFromSymbol(ITypeSymbol typeSymbol)
...
-				// Use the first 16 bytes of the hash
-				for (int i = 0; i < Math.Min(data.Length, 16); i++)
+				var min = Math.Min(data.Length, 16);
+				var result = new string((char)0, min * 2);
+				// efficient implementation to convert a byte array to hex string.
+				// See https://github.com/patridge/PerformanceStubs/blob/574c79441c18220f7841bdd1c25db0f6d7752876/README.markdown#converting-a-byte-array-to-a-hexadecimal-string
+				unsafe
 				{
-					builder.Append(data[i].ToString("x2", CultureInfo.InvariantCulture));
+					var lookupP = _lookup32UnsafeP;
```
Sergio0694, line 49: As an alternative and safer option, you could also check out what I'm doing in ComputeSharp ([here](https://github.com/Sergio0694/ComputeSharp/blob/ccd0a5da74fd45fd201938cf48caa76e26806065/src/ComputeSharp.D2D1.SourceGenerators/ID2D1ShaderGenerator.CreateLoadBytecodeMethod.Syntax.cs#L192)). You could just reuse that lookup table (removing the leading `0x` if you don't need them) and just index from there, without the need for pointers nor pinning at all. It'd still avoid allocating strings for each value in this loop just the same.

Sergio0694 (commented):
Left some general comments 🙂

src/SourceGenerators/Uno.UI.SourceGenerators/DependencyObject/DependencyObjectGenerator.cs
Youssef1313: We're using C# 10:

https://github.com/unoplatform/uno/blob/18f9d47cf7b944820a49dec723ebc574e1cd936f/src/Directory.Build.props#L51

I think it should be no problem to update to C# 11. I'll do that in a separate PR.

src/SourceGenerators/Uno.UI.SourceGenerators/Helpers/HashBuilder.cs
Youssef1313: Nice! I benchmarked and the lookup table approach has the same performance. I used it in the last commit.

Youssef1313:
Thanks for the review @Sergio0694, very much appreciated!

src/SourceGenerators/Uno.UI.SourceGenerators/Helpers/HashBuilder.cs
```diff
@@ -28,25 +45,29 @@ public static string BuildIDFromSymbol(ITypeSymbol typeSymbol)
...
 				var data = algoritm.ComputeHash(Encoding.UTF8.GetBytes(input));
 
-				// Create a new Stringbuilder to collect the bytes
-				// and create a string.
-				var builder = new StringBuilder();
-
-				// Use the first 16 bytes of the hash
-				for (int i = 0; i < Math.Min(data.Length, 16); i++)
+				var min = Math.Min(data.Length, 16);
+				var result = new string((char)0, min * 2);
```
spouliot, line 56: if you're using `unsafe` then maybe you should also `stackalloc` with a `Span<Char>`?

src/SourceGenerators/Uno.UI.SourceGenerators/Helpers/HashBuilder.cs
```diff
@@ -28,25 +45,29 @@ public static string BuildIDFromSymbol(ITypeSymbol typeSymbol)
...
+				unsafe
 				{
-					builder.Append(data[i].ToString("x2", CultureInfo.InvariantCulture));
+					fixed (char* resultP = result)
+					{
+						for (int i = 0; i < min; i++)
+						{
+							var s = _formattedBytes[data[i]];
```
spouliot, line 63: you don't really need a lookup table here, e.g.
Each char would become `(char)('0' + data [i])`

Youssef1313: The result of `_formattedBytes[..]` is a 2-char string. I'm not sure how `(char)('0' + data [i])` is equivalent?

src/SourceGenerators/Uno.UI.SourceGenerators/Helpers/HashBuilder.cs
Youssef1313: @spouliot We'll endup calling `.ToString()`, I think. So I guess creating it as string in the first place could be faster, but I'm not really sure. Haven't benchmarked.

Are there any concerns with the current approach?

Sergio0694, line 56: Missed this before, this is undefined behavior and you should definitely never do this. The best you can do on .NET Standard 2.0 is renting a `char[]`  from the pool, writing data to it, then slicing as a span and calling `ToString()`, then returning the array to the pool. But you should never, for any reason, mutate data from an existing `string` instance.

Sergio0694 (changes requested):

src/SourceGenerators/Uno.UI.SourceGenerators/Helpers/HashBuilder.cs
spouliot: that would works if you change to a `Span<char>` since it would require a single assignation per `char`

src/SourceGenerators/Uno.UI.SourceGenerators/Helpers/HashBuilder.cs
spouliot: yes, the string has to be created, eventually :)
but the stackalloc is cheap, use `char` (easier logic) and like @Sergio0694 commented it avoid modifying the create string (bad)

Youssef1313: @Sergio0694 Why not use Span directly instead of renting an array?

```
        [Benchmark]
        public string RentArr()
        {
            using (var algorithm = SHA256.Create())
            {
                var data = algorithm.ComputeHash(Encoding.UTF8.GetBytes(Input));

                var min = Math.Min(data.Length, 16);
                var array = ArrayPool<char>.Shared.Rent(min * 2);
                var span = array.AsSpan();
                unsafe
                {
                    for (int i = 0; i < min; i++)
                    {
                        var s = formattedBytes[data[i]];
                        span[i * 2] = s[0];
                        span[i * 2 + 1] = s[1];
                    }
                }

                ArrayPool<char>.Shared.Return(array);
                return span.ToString();
            }
        }

        [Benchmark]
        public string UseSpan()
        {
            using (var algorithm = SHA256.Create())
            {
                var data = algorithm.ComputeHash(Encoding.UTF8.GetBytes(Input));

                var min = Math.Min(data.Length, 16);
                
                Span<char> span = stackalloc char[min * 2];
                unsafe
                {
                    for (int i = 0; i < min; i++)
                    {
                        var s = formattedBytes[data[i]];
                        span[i * 2] = s[0];
                        span[i * 2 + 1] = s[1];
                    }
                }

                
                return span.ToString();
            }
        }
```

Performance-wise, the difference is negligible (using span directly is ~1.473 us, and renting array is ~1.563 us when running the benchmark on my machine). Is renting preferred for some reason?

Sergio0694: If the length has an upper bound, sure, using `stackalloc` is fine. Just hardcode it to 32 though.
Also you can drop the `unsafe` block entirely then 🙂

src/SourceGenerators/Uno.UI.SourceGenerators/Helpers/HashBuilder.cs
```diff
@@ -28,25 +45,24 @@ public static string BuildIDFromSymbol(ITypeSymbol typeSymbol)
...
-					builder.Append(data[i].ToString("x2", CultureInfo.InvariantCulture));
+					var s = _formattedBytes[data[i]];
+					span[i * 2] = s[0];
+					span[i * 2 + 1] = s[1];
 				}
 
-				// Return the hexadecimal string.
-				return builder.ToString();
+
+				return span.Slice(0, min).ToString();
```
Youssef1313, line 65: This bug was caught by the test `When_Grid_Uses_Both_Syntaxes` 🎉 

Looks like https://github.com/unoplatform/uno/pull/9801 was useful :) :)

```suggestion
				return span.Slice(0, min * 2).ToString();
```

src/SourceGenerators/Uno.UI.SourceGenerators/Helpers/HashBuilder.cs
```diff
@@ -28,25 +45,24 @@ public static string BuildIDFromSymbol(ITypeSymbol typeSymbol)
...
+				var data = algorithm.ComputeHash(Encoding.UTF8.GetBytes(input));
 
-				// Create a new Stringbuilder to collect the bytes
-				// and create a string.
-				var builder = new StringBuilder();
+				var min = Math.Min(data.Length, 16);
 
-				// Use the first 16 bytes of the hash
-				for (int i = 0; i < Math.Min(data.Length, 16); i++)
+				Span<char> span = stackalloc char[32];
```
jeromelaban, line 56: 32 seems arbitrary (even if linked implicitly the chosen algorithm). Should it be set to a computed value?

Sergio0694: The value has to be bounded anyway because you can't just use `stackalloc` with an unchecked value (as it could stack overflow, which you cannot recover from), and the value here was clamped to 32 which is very small anyway. Using `stackalloc` with a fixed amount results in much better codegen, as the JIT can unroll the loop and avoid some additional checks it has to do otherwise, and with the constant being so small anyway, even if you have some leftover bytes at the end it's still worth it. You can see a codegen comparison [here](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAMIR8AB14AbGFADKMgG68wMXAG4aNANqyA1rzEAZCGGyTcASS68MAXRrMkTFAwAiATy7Z8ygBS8eBgQAShoAbxoGKIYAjAZpLgBzDAALBgBeBgBZbFSWLIDfBDQGAGZSYI1qaIZI6NkxbC4AHmB3DBgAPgZcRq4MnoxsMF0zSRMGNo6tBOSU2yqauqiAMQgIX16mypoAX01qHX0jEzNLazsHcidiFxXeBBgAE19Q6gjq+r7W9q6evoGuCGIzGEymMC05QWy1qn1W602fR21H21G0WRgqQgTws4kkvgxWJxeIA8mI+BAuLgWAA5CBWSQBAKJYL2WjXZwMNYbBpNH4dbpbLhvD67IA===) 🙂

jeromelaban: Thanks!

Youssef1313:
@jeromelaban This is green now. Is it ready to merge?

src/SourceGenerators/Uno.UI.SourceGenerators/Helpers/HashBuilder.cs
Youssef1313: @spouliot I don't get it. If `data[i]` for example is 15, we'll want to get `0F`
How would `'0' + 15` get me `0F`? Anything I'm missing?

spouliot: I mean something like

```csharp
	[Benchmark]
	public string UseSpanWithoutTable()
	{
		using (var algorithm = SHA256.Create())
		{
			var data = algorithm.ComputeHash(Encoding.UTF8.GetBytes(Input));

			var min = Math.Min(data.Length, 16);
			
			Span<char> span = stackalloc char[min * 2];
			int i = 0, j = 0;
			while (i < min) {
				var b = data [i++];
				var nib = (b >> 4);
				span[j++] = nib > 9 ? (char)(nib + 'W') : (char)(nib + '0');
				nib = (b & 0xf);
				span[j++] = nib > 9 ? (char)(nib + 'W') : (char)(nib + '0');
			}
			
			return span.ToString();
		}
	}
```

|              Method |     Mean |   Error |  StdDev |
|-------------------- |---------:|--------:|--------:|
|             RentArr | 587.8 ns | 1.49 ns | 1.24 ns |
|             UseSpan | 591.9 ns | 3.44 ns | 3.21 ns |
| UseSpanWithoutTable | 574.9 ns | 1.60 ns | 1.34 ns |

It's not terribly faster [1] than the use of span but it requires less code (no table) and memory.

[1] to be honest the majority of the time is likely spent on `ComputeHash` and not on the string conversion 😄

Youssef1313: Thanks for clarifying @spouliot 
I switched to this approach in the last commit

jeromelaban (approved):

PR #9856, opened 2022-09-15
draft

Youssef1313:

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/SourceGenerators/Uno.UI.SourceGenerators/Helpers/HashBuilder.cs
+++ b/src/SourceGenerators/Uno.UI.SourceGenerators/Helpers/HashBuilder.cs
@@ -16,7 +16,7 @@ namespace Uno.UI.SourceGenerators.Helpers
 		/// Build a stable ID from the provided symbol
 		/// </summary>
 		public static string BuildIDFromSymbol(ITypeSymbol typeSymbol)
-			=> typeSymbol.GetFullName() + "_" + Build(typeSymbol.MetadataName + typeSymbol.ContainingAssembly.MetadataName);
+			=> typeSymbol.GetFullName();
 
 		/// <summary>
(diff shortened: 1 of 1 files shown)


Youssef1313:
@jeromelaban CAn you restart the CI here please?

jeromelaban:
The problem here is that if there are two types with the same name in different assemblies, there may be a clash. It's not something we validate inside uno.

Youssef1313:
@jeromelaban If they are in different assemblies, that should be different compilations too? What's the problem if the same name is used across different compilations?

jeromelaban:
The problem here is disambiguation of the same FullName in different assemblies, but in the same compilation.

Youssef1313:
@jeromelaban I'm not following with that. We generate code for symbols that are declared in source, shouldn't they all have the same assembly, specifically, it's [`compilation.Assembly`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.compilation.assembly?view=roslyn-dotnet-4.3.0#microsoft-codeanalysis-compilation-assembly). If a symbol doesn't belong to `compilation.Assembly`, I think we will never generate code for it.

jeromelaban:
> @jeromelaban I'm not following with that. We generate code for symbols that are declared in source, shouldn't they all have the same assembly, specifically, it's [`compilation.Assembly`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.compilation.assembly?view=roslyn-dotnet-4.3.0#microsoft-codeanalysis-compilation-assembly). If a symbol doesn't belong to `compilation.Assembly`, I think we will never generate code for it.

That's fair, I had in mind that the original use was for symbols coming from other assemblies, but this is not the case anymore. Maybe it should be renamed to `GetGeneratedFileNameHint()` or something to make it clear, or removed altogether and used inline.

jeromelaban:
@Youssef1313 still needed?

PR #9860, opened 2022-09-15
build: Include only generator dll

Youssef1313:
## PR Type

## PR Checklist

## Other information


jeromelaban:
@Youssef1313 Roslyn is likely loading those files in the current appdomain, making the generator work as a result. We could make this work differently by adjusting this: 
https://github.com/unoplatform/uno/blob/cf70e8e8ec4389d80269122280b5083066724958/src/SourceGenerators/Uno.UI.SourceGenerators/Helpers/DependenciesInitializer.cs#L16-L51

Youssef1313:
I think we should never need `DependenciesInitializer` actually. I'll see how this can be improved. One possible challenge is that Uno.UI.SourceGenerators is consumable via Uno.UI package and also via `ProjectReference`s in our solution. That makes handling its dependencies a bit harder.

I also don't think generators should need to manually include themselves via `<Analyzer Include=...`

jeromelaban:
> @jeromelaban The `<Analyzer Include=...` part is only for Roslyn generators right? It has nothing to do with `Uno.SourceGeneration`?

Correct. Uno.Sourcegeneration does not use the `Analyzer` msbuild item.

PR #9873, opened 2022-09-17
build: Try to simplify generators props

Youssef1313:
I don't see how language version is related here.

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/SourceGenerators/Uno.UI.SourceGenerators/Content/Uno.UI.SourceGenerators.props
+++ b/src/SourceGenerators/Uno.UI.SourceGenerators/Content/Uno.UI.SourceGenerators.props
@@ -8,17 +8,7 @@
 	<PropertyGroup>
 		<_isRoslynAnalyzerAvailable Condition="'$(MSBuildVersion)' &gt;= '16.8'">true</_isRoslynAnalyzerAvailable>
 
-		<_canUseRoslynAnalyzer Condition="
-							'$(LangVersion)' == 'preview'
-							Or '$(LangVersion)' == 'latest'
-							Or '$(LangVersion)' == 'latestMajor'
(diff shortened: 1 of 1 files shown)


src/SourceGenerators/Uno.UI.SourceGenerators/Content/Uno.UI.SourceGenerators.props
```diff
@@ -8,17 +8,7 @@
 	<PropertyGroup>
 		<_isRoslynAnalyzerAvailable Condition="'$(MSBuildVersion)' &gt;= '16.8'">true</_isRoslynAnalyzerAvailable>
 
-		<_canUseRoslynAnalyzer Condition="
```
jeromelaban, line 11: This is needed for all targets that explicitly specify a LangVersion, where there are specific behaviors that cause issues with generators.

Why do you think this change is needed?

Youssef1313: > where there are specific behaviors that cause issues with generators.

@jeromelaban Can you elaborate more on that?

> Why do you think this change is needed?

To my knowledge, source generators is a compiler feature, not a language feature. ie, it should work whatever the language version is, as long as the compiler version used supports it. So, my motivation was why disable using roslyn generators if it can be used?

jeromelaban: Source generators can behave differently based on the version, though I don't recall exactly which ones. Still, this is a breaking change because it changes the generation behavior for `LangVersion` below 9 (and there are many consumers for that). Let's keep it that way for now.

PR #9895, opened 2022-09-19
Clipboard methods for `TextBox` and `PasswordBox`

### LINKED ISSUE #9894: Support for `TextBox` and `PasswordBox` clipboard methods

Support for Clipboard related methods in `TextBox` and `PasswordBox`

## Why is this needed:

WinAppSDK parity

## For which Platform:

All

MartinZikmund:
GitHub Issue (If applicable): closes #9894

## PR Type

- Feature

Not supported.

## What is the new behavior?

Supported.

## PR Checklist


src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.cs
```diff
@@ -1078,11 +1079,61 @@ public void Select(int start, int length)
...
 
-
 		partial void SelectPartial(int start, int length);
 
 		partial void SelectAllPartial();
 
+		/// <summary>
+		/// Copies content from the OS clipboard into the text control.
+		/// </summary>
+		public async void PasteFromClipboard()
```
Youssef1313, line 1089: Being `async`, I think this doesn't match UWP/WinUI? Is that a problem?

MartinZikmund: I need to document this, but it will need to stay async due to the fact that WASM uses async for clipboard methods. On the other hand it should not be a huge problem as long as developers are aware that it might take a few ticks before the value is actually pasted.

Youssef1313: `async void` should really be avoided because it cannot be awaited.

If it's possible to diverge from UWP/WinUI here, I'd make it `async Task` instead. Otherwise, try with approaches like `Task.Run(async () => await content.GetTextAsync()).Wait()`.

jeromelaban: We should rather dispatch it again on the dispatcher than use `async void`.

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.cs
```diff
@@ -1078,11 +1079,64 @@ public void Select(int start, int length)
...
 		partial void SelectPartial(int start, int length);
 
 		partial void SelectAllPartial();
 
+		/// <summary>
+		/// Copies content from the OS clipboard into the text control.
+		/// </summary>
+		public void PasteFromClipboard()
+		{
+			Dispatcher.RunAsync(CoreDispatcherPriority.High, async () =>
```
Youssef1313, line 1091: I think this should fix the build error

```suggestion
			_ = Dispatcher.RunAsync(CoreDispatcherPriority.High, async () =>
```

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.cs
```diff
@@ -1078,11 +1079,64 @@ public void Select(int start, int length)
...
+			}
+		}
+
+		/// <summary>
+		/// Moves the selected content to the OS clipboard and removes it from the text control.
+		/// </summary>
+		public void CutSelectionToClipboard()
+		{
+			CopySelectionToClipboard();
+			Text = Text.Remove(SelectionStart, SelectionLength);
```
Youssef1313, line 1137: Can `SelectionStart` be -1? What happens in this case for Windows and do we have the same behavior here?

jeromelaban: I don't think it can be -1, mostly because it indicates the caret position, yet it may be interesting to validate this behavior with runtime tests (even if it's only for skia targets?).

Youssef1313: @jeromelaban I asked because other methods have `selectionStart >= 0` checks

MartinZikmund: Yes, indeed cannot be -1, simplified the code to account for that

PR #9902, opened 2022-09-20
fix(VisibleBoundsPadding): Ensure the proper root is used to compute relative bounds

kazo0:
closes https://github.com/unoplatform/nventive-private/issues/412

## PR Type

- Bugfix

VisibleBoundsPadding calculates improper offsets when the control is within a native iOS Modal window. We were using `TransformToVisual` relative to `Windows.Current.Content` which results in an offset of X=16, Y=47 since the Window.Content (white background) is no longer at position 0,0 while a modal (red background) is being displayed

MicrosoftTeams-image (2)

## What is the new behavior?

Use `TransformToVisual(null)` to ensure that the root of the XAML tree is used for the transformation


jeromelaban (approved):

kazo0:
@Mergifyio backport release/stable/4.4

PR #9925, opened 2022-09-22
build: Move to dotnet-g4 instead of our T4Generator

### LINKED ISSUE #9845: Remove dependency on net461 T4 engine

The current implementation of the T4 generator uses .NET Framework desktop-only compatible binaries, causing issues when building under Linux.

Change the process to use [dotnet-t4](https://github.com/mono/t4) instead would make it faster and more compatible.

At this time, we can't remove the T4 templates generation because that generated code is used by the Roslyn generators afterwards and we can't yet have dependencies between generators.

## Why is this needed:
Linux support, faster compilation.

Youssef1313:
GitHub Issue (If applicable): closes #9845

## PR Type

## PR Checklist

## Other information


src/Uno.UI/MixinGeneration.targets
```diff
@@ -2,9 +2,6 @@
 <Project>
 
   <PropertyGroup>
-	<T4Path>$(MSBuildThisFileDirectory)..\T4Generator\bin\$(Configuration)</T4Path>
-	<T4Bin>$(T4Path)\T4Generator.exe</T4Bin>
```
jeromelaban, line 6: I don't think having an external dependency will make it easy on developers to work with Uno. We should be validating the presence of the tool being installed as follows:
https://github.com/unoplatform/uno/blob/e2248d2a89f089915e1c5c93ddf4aab98ef671bd/build/test-scripts/wasm-run-automated-uitests.sh#L13

Let's install it in obj (Interdemediate path) to avoid git detection.

src/Uno.UI/MixinGeneration.targets
```diff
@@ -122,7 +122,7 @@
 	<Message Text="Generating mixins" Importance="high" />
 
 	<!-- This allows for the mixins to generate in parallel -->
-	<MSBuild BuildInParallel="true" Projects="@(MixinDefinition)" StopOnFirstFailure="True" Targets="_InnerGenerateMixins" />
+	<MSBuild BuildInParallel="false" Projects="@(MixinDefinition)" StopOnFirstFailure="True" Targets="_InnerGenerateMixins" />
```
jeromelaban, line 125: Why do you need to remove it ?

Youssef1313: Removing temporarily.

I think we are hitting a race fixed in https://github.com/mono/t4/pull/101.

The PR on mono/t4 was merged in March 2021, and the latest version on NuGet.org is released in February 2021.

I want to get things to pass first, then I'll see how I can consume a more recent release.

Youssef1313:
@jeromelaban There are different failures in different builds.

- Build net6 Binaries UWP

  ```
  Generating mixins
  The command could not be loaded, possibly because:
    * You intended to execute a .NET application:
        The application 'tool' does not exist.
    * You intended to execute a .NET SDK command:
        A compatible .NET SDK was not found.
  
  Requested SDK version: 6.0.100-preview.6.21355.2
  global.json file: C:\a\1\s\src\global.json
  
  Installed SDKs:
  
  Install the [6.0.100-preview.6.21355.2] .NET SDK or update [C:\a\1\s\src\global.json] to match an installed SDK.
  
  Learn about SDK resolution:
  6.0.401 [C:\hostedtoolcache\windows\dotnet\sdk]
  https://aka.ms/dotnet/sdk-not-found
  ```

  Is this because we reference 6.0.100-preview [here](https://github.com/unoplatform/uno/blob/4db60f462ebc085320eb5be3f855a3b33d26927b/src/global-net6.json#L7), and 6.0.401 is found? Should we update global-net6.json in that case?

---------

- Documentation WebSite

  ```
  19:60>Project "D:\a\1\s\src\Uno.UI\Uno.UI.csproj" (19:60) is building "D:\a\1\s\src\Uno.UI\MixinGeneration.targets" (96) on node 1 (_InnerGenerateMixins target(s)).
    96>_InnerGenerateMixins:
         D:\a\1\s\src\Uno.UI\dotnet-t4/t4 D:\a\1\s\src\Uno.UI\Mixins\DependencyPropertyMixins.tt -o D:\a\1\s\src\Uno.UI\Mixins\DependencyPropertyMixins
         System.AggregateException: One or more errors occurred. (Not a valid .NET Core host)
          ---> Mono.TextTemplating.TemplatingEngineException: Not a valid .NET Core host
            at Mono.TextTemplating.TemplatingEngine.GetOrCreateCompiler() in /_/Mono.TextTemplating/Mono.TextTemplating/TemplatingEngine.cs:line 67
            at Mono.TextTemplating.TemplatingEngine.CompileCode(IEnumerable`1 references, TemplateSettings settings, CodeCompileUnit ccu, CancellationToken token) in /_/Mono.TextTemplating/Mono.TextTemplating/TemplatingEngine.cs:line 306
            at Mono.TextTemplating.TemplatingEngine.CompileTemplateInternal(ParsedTemplate pt, String content, ITextTemplatingEngineHost host, TemplateSettings settings, CancellationToken token) in /_/Mono.TextTemplating/Mono.TextTemplating/TemplatingEngine.cs:line 271
            at Mono.TextTemplating.TemplatingEngine.ProcessTemplateAsync(ParsedTemplate pt, String content, TemplateSettings settings, ITextTemplatingEngineHost host, CancellationToken token) in /_/Mono.TextTemplating/Mono.TextTemplating/TemplatingEngine.cs:line 88
            at Mono.TextTemplating.TemplateGenerator.ProcessTemplateAsync(ParsedTemplate pt, String inputFileName, String inputContent, String outputFileName, TemplateSettings settings, CancellationToken token) in /_/Mono.TextTemplating/Mono.TextTemplating/TemplateGenerator.cs:line 251
            --- End of inner exception stack trace ---
            at Mono.TextTemplating.TextTransform.MainInternal(String[] args)
            at Mono.TextTemplating.TextTransform.Main(String[] args) in /_/dotnet-t4/TextTransform.cs:line 42
  ##[error]src\Uno.UI\MixinGeneration.targets(145,4): Error MSB3073: The command "D:\a\1\s\src\Uno.UI\dotnet-t4/t4 D:\a\1\s\src\Uno.UI\Mixins\DependencyPropertyMixins.tt -o D:\a\1\s\src\Uno.UI\Mixins\DependencyPropertyMixins" exited with code -1.
    96>D:\a\1\s\src\Uno.UI\MixinGeneration.targets(145,4): error MSB3073: The command "D:\a\1\s\src\Uno.UI\dotnet-t4/t4 D:\a\1\s\src\Uno.UI\Mixins\DependencyPropertyMixins.tt -o D:\a\1\s\src\Uno.UI\Mixins\DependencyPropertyMixins" exited with code -1.
    96>Done Building Project "D:\a\1\s\src\Uno.UI\MixinGeneration.targets" (_InnerGenerateMixins target(s)) -- FAILED.
    19>Done Building Project "D:\a\1\s\src\Uno.UI\Uno.UI.csproj" (GenerateMixins target(s)) -- FAILED.
     1>Done Building Project "D:\a\1\s\Build\Uno.UI.Build.csproj" (GenerateDoc target(s)) -- FAILED.

  ```

  I have no idea what that is.

----------

- Build WebAssembly UI Tests

  ```
  Generating mixins
  Specify which project file to use because this '/__w/1/s/src/Uno.UI' contains more than one project file.
/__w/1/s/src/Uno.UI/MixinGeneration.targets(124,2): error MSB3073: The command "dotnet tool update dotnet-t4 --version 2.3.0-preview-0041-g0bb7f1d477 --tool-path /__w/1/s/src/Uno.UI/dotnet-t4" exited with code 1. [/__w/1/s/src/Uno.UI/Uno.UI.Wasm.csproj]
  ```

  I don't see a way for `dotnet tool update` to specify a project. Why even it needs a project? 

-----------

@jeromelaban Do you know?

jeromelaban:
You'll need to install the tool with `dotnet tool install` and check if the install `dotnet-t4` tool file exists before running the installer. 

For the global.json, let's update to match the installed version (6.0.400 or 6.0.401).

Youssef1313:
@jeromelaban This is now failing only on WinUI.

I tried cherry-picked the changes from this PR into a clone of `generated/master/winui-autoconvert` branch and I don't see those errors locally.

Not sure how to investigate this further. I looked into the binlog but wasn't able to find something interesting.

src/Uno.UI/MixinGeneration.targets
```diff
@@ -110,28 +99,31 @@
...
 	  <MixinDefinition Include="$(MSBuildThisFileFullPath)" Condition="$(IsXamarinMac)">
-		<Properties>InputFile=$(MSBuildThisFileDirectory)Mixins\macOS\FrameworkElementMixins.tt;OutputPath=$(MSBuildThisFileDirectory)Mixins\macOS;Configuration=$(Configuration);Platform=$(Platform)</Properties>
+		<Properties>InputFile=$(MSBuildThisFileDirectory)Mixins\macOS\FrameworkElementMixins.tt;OutputPath=$(MSBuildThisFileDirectory)Mixins\macOS\FrameworkElementMixins.cs;Configuration=$(Configuration);Platform=$(Platform)</Properties>
 	  </MixinDefinition>
 	</ItemGroup>
 
+
 	<Message Text="Generating mixins" Importance="high" />
 
+	<Exec Command="dotnet tool install dotnet-t4 --version 2.3.0-preview-0041-g0bb7f1d477 --tool-path dotnet-t4" IgnoreExitCode="true" WorkingDirectory="$(MSBuildThisFileDirectory)..\" />
```
jeromelaban, line 125: IntermediatePath did not work? Also, an actual error should be reported. You should be able to use `Condition="exists(..\dotnet-t4\t4)"` to avoid "already install" errors.

Youssef1313: I haven't tried yet. This suggestion is not related to the failure in WinUI builds right?

Youssef1313: For some reason, both `BaseIntermediateOutputPath` and `IntermediateOutputPath` are evaluated as empty when the target is run.

image

Youssef1313: > You should be able to use Condition="exists(..\dotnet-t4\t4)" to avoid "already install" errors.

This won't work well if the developer has installed the tool as global tool I think. The t4 executable will not be found in the expected directory, though `dotnet tool install` can still fail because the tool is installed already.

jeromelaban: > This won't work well if the developer has installed the tool as global tool I think. The t4 executable will not be found in the expected directory, though `dotnet tool install` can still fail because the tool is installed already.

Local install is not impacted by a global install, the installer validates for the specified path only.

> For some reason, both `BaseIntermediateOutputPath` and `IntermediateOutputPath` are evaluated as empty when the target is run.

Then let's use `obj` hard-coded (we won't change the path any time soon anyways). I'd rather have this as an artifact of the local build than a top level dependency (for now at least).

Youssef1313:
There is a build failure on Skia. I think it's a race condition that happens when two builds are generating mixins.

- Project 1 invokes dotnet tool install
- Project 2 invokes dotnet tool install
- Project 1 finds the tool isn't install, and attempts to install
- Project 2 finds the tool isn't install, and attempts to install
- Project 1 starts installing
- Project 2 starts installing, but it fails for some reason because Project 1 already started installing and some files/directories were already created
- Project 2 goes to the next step (since I ignore install failures)
- Project 2 can't execute the tool, because the installation isn't yet complete.
- Project 1 finishes installing
- Project 1 executes the tool successfully.

The fix here is that we really need different locations under obj for each build (ie, `IntermediateOutputPath`). I'll take whatever logic dotnet/sdk has if we find it empty.

src/Uno.UI/MixinGeneration.targets
Youssef1313: I figured out why `IntermediateOutputPath` didn't work. The `_InnerGenerateMixins` target is run by `MSBuild` task, and for each MixinDefinition, we need to explicitly pass `IntermediateOutputPath`.

So remaining thing here is WinUI failure if it's still present with the latest changes.

Youssef1313:
@jeromelaban More builds broke after switching to `IntermediateOutputPath`.

I'm not entirely sure what's going on here.

Youssef1313:
@jeromelaban This is still failing 😕

jeromelaban:
Sure, but this time it's caused by the generated T4 content, which is likely incorrect as dotnet-t4 is different from what the previous tool was doing.

Youssef1313:
> Sure, but this time it's caused by the generated T4 content, which is likely incorrect as dotnet-t4 is different from what the previous tool was doing.

@jeromelaban This is about the failure in WinUI builds right? But the failure in other builds look to me like the race condition mentioned in https://github.com/unoplatform/uno/pull/9925#issuecomment-1258095148 as far as I understand.

jeromelaban:
Yes, correct, the race condition needs to be fixed by installing the tool from uno.ui.build.csproj by running the new target explicitly.

Youssef1313:
> Yes, correct, the race condition needs to be fixed by installing the tool from uno.ui.build.csproj by running the new target explicitly.

Usually when working locally, Uno.UI.Build isn't built. So developers will now have to `dotnet tool install` the tool manually right?

jeromelaban:
Yes, correct, the race condition needs to be fixed by installing the tool from uno.ui.build.csproj by running the new target explicitly.

> > Yes, correct, the race condition needs to be fixed by installing the tool from uno.ui.build.csproj by running the new target explicitly.
> 
> Usually when working locally, Uno.UI.Build isn't built. So developers will now have to `dotnet tool install` the tool manually right?

When running locally, there won't be a race condition as only one project is built at a time (on CI, wasm and skia are built concurrently, for instance).

build/ci/.azure-devops-package-generic.yml
```diff
@@ -83,6 +83,10 @@ jobs:
       logProjectEvents: false
       createLogFile: false
 
+  - bash: |
+      dotnet tool install --global Basic.CompilerLog
+      complog create --projectName=Uno.UI $(build.artifactstagingdirectory)\build-$(GitVersion.FullSemVer)-$(XAML_FLAVOR_BUILD).binlog
```
Youssef1313, line 88: Let's see if this works and if I'll be able to get something useful from it. It's possible to turn this into a Microsoft.CodeAnalysis.Compilation, which could make investigation much easier!

build/ci/.azure-devops-package-generic.yml
```diff
@@ -83,6 +83,14 @@ jobs:
...
       createLogFile: false
 
+  - task: CmdLine@2
+    inputs:
+      workingDirectory: $(build.artifactstagingdirectory)
+      script: |
+        dotnet tool install --global Basic.CompilerLog
+        complog create --projectName=Uno.UI build-$(GitVersion.FullSemVer)-$(XAML_FLAVOR_BUILD).binlog
+    condition: always()
+
```
Youssef1313, line 93: @jeromelaban Let me know if you want me to keep this, delete it, or comment it.

jeromelaban: Let's comment it out for now, but it will be very useful for some advanced debugging,

build/Uno.UI.Build.csproj
```diff
@@ -223,7 +223,13 @@
 
 	</Target>
 
+	<PropertyGroup>
+		<_UnoDotnetT4ToolBasePath>$(MSBuildThisFileDirectory)..\obj\dotnet-tools</_UnoDotnetT4ToolBasePath>
+		<_DotnetT4InstallCommand>dotnet tool install dotnet-t4 --version 2.3.0-preview-0041-g0bb7f1d477 --add-source $(MSBuildThisFileDirectory)..\src\PackageCache --tool-path $(_UnoDotnetT4ToolBasePath)</_DotnetT4InstallCommand>
+	</PropertyGroup>
```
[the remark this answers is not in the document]
Youssef1313: @jeromelaban Wasn't the idea to add the installation here to avoid the race condition when building multiple Uno.UI targets in parallel?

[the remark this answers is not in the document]
jeromelaban: You can invoke the installation target `UnoInstallDotnetT4` in Uno.UI.csproj explicitly (it'll will skip all the other build targets)

Youssef1313:
@jeromelaban This is green now :)

Issue #9811, opened 2022-09-12
[UI Tests] Move image based tests to runtime tests


jeromelaban:
## What would you like to be added:
Many UI tests on Android and iOS are using "remote" screenshots which take a significant amount of time to run (1 second per screenshot in general) that do not validate actual UI interactions.

Moving these UI tests (most of which are using [`ImageAssert`](https://github.com/unoplatform/uno/blob/3595e8a61013d5885147c071f2242beb2d4c4b9a/src/SamplesApp/SamplesApp.UITests/TestFramework/ImageAssert.cs#L20)) to be [`Runtime Tests`](https://platform.uno/docs/articles/uno-development/creating-runtime-tests.html) would help decrease significantly the CI build time.

Here's a (non-exhaustive) list of test that could be migrated:
- https://github.com/unoplatform/uno/blob/c1c6fdb534c0ff7762bd46dbecc1017b4c5392ac/src/SamplesApp/SamplesApp.UITests/Windows_UI_Xaml_Media/ImageBrushTests/ImageBrush_Stretch.cs#L20
- https://github.com/unoplatform/uno/blob/6e093f7bae3116d004e0b54a66467193dcd29a96/src/SamplesApp/SamplesApp.UITests/Windows_UI_Xaml_Controls/CanvasTests/Canvas_Measurement_Tests.cs#L17

## Why is this needed:

More performant CI, with more platforms to be validated as part of the runtime tests.

amelield:
When_Stretch() is using screenshots. So if the screenshot doesn't work in web assembly, should exclude them from that test? I was trying to find a workaround that reproduce the same results(without much success). Or do anyone know a good way to make it works.

jeromelaban:
That's correct. WebAssembly cannot make screenshots at this time, so we need to skip this test using `#if !__WASM__`.

amelield:
In Skia, I can't find what kind of import or interface could make TakeScreenshot(String) works; I see there are possibilities (in Runtimes or others places) but non of those I tried fixed the TakeScreenshot not existing in context.

jeromelaban:
Here are a few of those, using the `RenderTargetBitmap` API (an equivalent of `TakeScreenshot`, but in-app):
https://github.com/unoplatform/uno/blob/aa9311da1926bd52df96ec152cb7585d3c8d74a3/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_FrameworkElement_Opacity.cs#L37-L72

[...]

PR #9939, opened 2022-09-22
chore(test): Move ImageBrush_Stretch to runtime test

amelield:
Issue  #9811

## PR Type

The test Was in Sample app test. The test is broken in different subtests according to the stretch tested.
The test did not work on Android du to failing Stretch.None  section of the test.

## What is the new behavior?

The test is now in Runtime app test. The test doe not work on WASM anymore du to the lack of an equivalent of screenshot.

Stretch.Uniform does not work on skia anymore since it is making a mosaic pattern. 
I included MACOS in functioning build but I could not test it. 
I did not includes Android since it was already failing on of the test previously and I was not able to test it, my build was running in a not enough memory exception.  

I did not move the images from Sample App folder in case it would be important for other tests. 

In order to make this test actually test something in Runtimes, I had, with a lot of help of Xiaotian modify the Runtime AppTest class. Because of the usage of async functions some of these test where on a different thread and did pass the test in Runtimes no matter what parameter they had. Since I did not found a way to make that test fail without the fix, The fix of the test is also included in that task.  That fix required some modification on the  Given_FrameworkElement_Opacity that previously never failed either. I can also make a separate pull request if you prefer.

## PR Checklist

## Other information

Internal Issue (If applicable): #9811

### DIFF
--- a/src/SamplesApp/SamplesApp.UITests/Windows_UI_Xaml_Media/ImageBrushTests/ImageBrush_Stretch.cs
+++ b/src/SamplesApp/SamplesApp.UITests/Windows_UI_Xaml_Media/ImageBrushTests/ImageBrush_Stretch.cs
@@ -16,7 +16,11 @@ namespace SamplesApp.UITests.Windows_UI_Xaml_Media.ImageBrushTests
 		[Test]
 		[AutoRetry]
 
-		[ActivePlatforms(Platform.Browser, Platform.iOS)] // Android behaves differently. See https://github.com/unoplatform/uno/pull/7238#issuecomment-937667565
+		// Other platforms are tested in RuntimeTests:
+		// src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Media/Given_ImageBrushStretch.cs
+		// This UI Test can be deleted once we can test Wasm in RuntimeTests.
+		// This is currently blocked due to lack of support for RenderTargetBitmap on Wasm.
+		[ActivePlatforms(Platform.Browser)] 
 		public void When_Stretch()
 		{
 			Run("UITests.Windows_UI_Xaml_Media.ImageBrushTests.ImageBrush_Stretch");
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_FrameworkElement_Opacity.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_FrameworkElement_Opacity.cs
@@ -48,27 +48,27 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 
 			await renderer.RenderAsync(SUT);
 
-			var si = new RawBitmap(renderer, SUT);
+			var si = await RawBitmap.From(renderer, SUT);
 
 			var width = SUT.tbOpacity1_0.ActualWidth;
 			var height = SUT.tbOpacity1_0.ActualHeight;
-			await ImageAssert.HasColorAtChild(si, SUT.tbOpacity1_0, (width / 4) * 3, height / 2, Colors.Black);
+			ImageAssert.HasColorAtChild(si, SUT.tbOpacity1_0, (width / 4) * 3, height / 2, Colors.Black);
 
 			width = SUT.tbOpacity0_5.ActualWidth;
 			height = SUT.tbOpacity0_5.ActualHeight;
-			await ImageAssert.HasColorAtChild(si, SUT.tbOpacity0_5, (width / 4) * 3, height / 2, "#FF808080");
+			ImageAssert.HasColorAtChild(si, SUT.tbOpacity0_5, (width / 4) * 3, height / 2, "#FF808080");
 
 			width = SUT.tbOpacity0_1.ActualWidth;
 			height = SUT.tbOpacity0_1.ActualHeight;
-			await ImageAssert.HasColorAtChild(si, SUT.tbOpacity0_1, (width / 4) * 3, height / 2, "#0xFFE6E6E6");
+			ImageAssert.HasColorAtChild(si, SUT.tbOpacity0_1, (width / 4) * 3, height / 2, "#FFE6E6E6");
 
 			width = SUT.border0_5.ActualWidth;
 			height = SUT.border0_5.ActualHeight;
-			await ImageAssert.HasColorAtChild(si, SUT.border0_5, 2, 2, "#FFFF8080");
+			ImageAssert.HasColorAtChild(si, SUT.border0_5, 2, 2, "#FFFF8080");
 
 			width = SUT.ImageOpacity0_5.ActualWidth;
 			height = SUT.ImageOpacity0_5.ActualHeight;
-			await ImageAssert.HasColorAtChild(si, SUT.ImageOpacity0_5, width / 2, height / 2, "#FFFEF3C2");
+			ImageAssert.HasColorAtChild(si, SUT.ImageOpacity0_5, width / 2, height / 2, "#FFFEF3C2");
 		}
 
 		[TestMethod]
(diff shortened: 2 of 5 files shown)


Youssef1313 (approved):
Looking good to me! :tada:

src/SamplesApp/SamplesApp.UITests/Windows_UI_Xaml_Media/ImageBrushTests/ImageBrush_Stretch.cs
```diff
@@ -1,55 +0,0 @@
...
-		private const string Yellowish = "#FFFEF200";
-		private const string Greenish = "#FF0ED145";
-
-		[Test]
-		[AutoRetry]
-
-		[ActivePlatforms(Platform.Browser, Platform.iOS)] // Android behaves differently. See https://github.com/unoplatform/uno/pull/7238#issuecomment-937667565
-		public void When_Stretch()
-		{
-			Run("UITests.Windows_UI_Xaml_Media.ImageBrushTests.ImageBrush_Stretch");
```
Youssef1313, line 22: I'm not sure if we should remove the sample from SamplesApp.

@jeromelaban What do you think?

amelield: According to the issue the goal was to "decrease significantly the CI build time". I may be counterproductive to keep both, no?

Youssef1313: @amelield I mean https://github.com/amelield/uno/blob/dev/ALD/test-move/src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Media/ImageBrushTests/ImageBrush_Stretch.xaml and https://github.com/amelield/uno/blob/dev/ALD/test-move/src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Media/ImageBrushTests/ImageBrush_Stretch.xaml.cs which still exist.

But please wait for @jeromelaban opinion first before deleting them.

amelield: After seeing the CI result I will remove MacOs from the test. The When_Opacity test now fail I have not changed the value, I think I just tested with When_Inner_Opacity() , my bad. Do I change the expected value for the one the CI wants #FFFF0000, when #FFE6E6E6 it is written? The runtime test is giving me the same result. I also would wait to see if I delete the files or not and just make a commit with all the fixes.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Media/Given_ImageBrushStretch.cs
```diff
@@ -0,0 +1,90 @@
...
+	{
+#if __SKIA__  || __MACOS__  // WASM does not have an equivalent of renderAsync  and Android behaves differently. See https://github.com/unoplatform/uno/pull/7238#issuecomment-937667565
+		[DataRow(Stretch.None)]
+		[DataRow(Stretch.Fill)]
+		[DataRow(Stretch.UniformToFill)]
+#if !__SKIA__ //Stretch.Uniform create a Mosaic in Skia which is not the desirable comportement 
+		[DataRow(Stretch.Uniform)]
+#endif
+		[DataRow(Stretch.None)]
+
```
Youssef1313, line 29: It looks like Stretch.None is duplicated and the test will run twice unnecessarily.

src/SamplesApp/SamplesApp.UITests/Windows_UI_Xaml_Media/ImageBrushTests/ImageBrush_Stretch.cs
Youssef1313: I don't think you need to change `When_Opacity` test. Your changes shouldn't be affecting it and it's likely an intermittent failure.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Media/Given_ImageBrushStretch.cs
```diff
@@ -0,0 +1,90 @@
...
+using ImageBrush = Windows.UI.Xaml.Media.ImageBrush;
+
+
+namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Media
+{
+	[TestClass]
+	[RunsOnUIThread]
+	public class Given_ImageBrushStretch
+	{
+#if __SKIA__  || __MACOS__  // WASM does not have an equivalent of renderAsync  and Android behaves differently. See https://github.com/unoplatform/uno/pull/7238#issuecomment-937667565
```
Youssef1313, line 21: What about iOS?

amelield: You are right I will see if the CI like this one

src/SamplesApp/SamplesApp.UITests/Windows_UI_Xaml_Media/ImageBrushTests/ImageBrush_Stretch.cs
jeromelaban: Indeed, this file cannot be removed at this time. In this PR, we can now mark the test with `[ActivePlatforms(Platform.Browser)]` so it runs only on wasm. We'll delete this file once we're able to use `WriteableBitmap` on Wasm.

amelield: [ActivePlatforms(Platform.Browser)] does not work in runtimes and it was already in a #if __SKIA__ section (that cover all the test of that class) could I just comment that single test section for the moment being(with an actual comment)? All other subtest are passing.

Youssef1313: @amelield I think what @jeromelaban meant is to have the test both as UI test and runtime test.

- UI test: Runs only on Wasm.
- Runtime test: Runs on other platforms.

@jeromelaban Could you confirm?

amelield: That make sense for the sample app right. I mixed up the two problems.

Youssef1313 (changes requested):
I think the remaining thing here is updating this line:

https://github.com/unoplatform/uno/blob/59e3c94a483508686e84d175fc40f3c82bc513ee/src/SamplesApp/SamplesApp.UITests/Windows_UI_Xaml_Media/ImageBrushTests/ImageBrush_Stretch.cs#L19

to something like this:

```csharp
// Other platforms are tested in RuntimeTests:
// src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Media/Given_ImageBrushStretch.cs
// This UI Test can be deleted once we can test Wasm in RuntimeTests.
// This is currently blocked due to lack of support for RenderTargetBitmap on Wasm.
[ActivePlatforms(Platform.Browser)]
```

src/SamplesApp/SamplesApp.UITests/Windows_UI_Xaml_Media/ImageBrushTests/ImageBrush_Stretch.cs
Youssef1313: > I don't think you need to change `When_Opacity` test. Your changes shouldn't be affecting it and it's likely an intermittent failure.

It's not an intermittent failure, I was wrong.

Though I don't clearly see why the test broke.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_FrameworkElement_Opacity.cs
```diff
@@ -48,27 +48,27 @@ public async Task When_Opacity()
...
 
 			width = SUT.tbOpacity0_5.ActualWidth;
 			height = SUT.tbOpacity0_5.ActualHeight;
-			await ImageAssert.HasColorAtChild(si, SUT.tbOpacity0_5, (width / 4) * 3, height / 2, "#FF808080");
+			ImageAssert.HasColorAtChild(si, SUT.tbOpacity0_5, (width / 4) * 3, height / 2, "#FF808080");
 
 			width = SUT.tbOpacity0_1.ActualWidth;
 			height = SUT.tbOpacity0_1.ActualHeight;
-			await ImageAssert.HasColorAtChild(si, SUT.tbOpacity0_1, (width / 4) * 3, height / 2, "#0xFFE6E6E6");
+			ImageAssert.HasColorAtChild(si, SUT.tbOpacity0_1, (width / 4) * 3, height / 2, "#0xFFE6E6E6");
```
Youssef1313, line 63: @amelield I took a look at this failure. The test is failing because of the extra `0x`. It should be `#0xFFE6E6E6`.

The color parsing ends up here:

https://github.com/unoplatform/uno/blob/1e9a8111358df0eb47cfb92f21c7a85c42efd9a1/src/Uno.UWP/UI/Colors.cs#L117-L123

It's very very very weird this code path is returning Red. What's special with Red to be a "fallback" in this error scenario?

The reason why this has failed in this PR but not in master is we never did the assertion before. The `HasColorAtChild` method was returning a Task, but we never `await`ed it.

amelield: Thank you very much I will remove the 0x from the test. Finger crossed

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Media/Given_ImageBrushStretch.cs
```diff
@@ -0,0 +1,89 @@
...
+#endif
+		[DataRow(Stretch.None)]
+
+		[TestMethod]
+		public async Task When_Stretch(Stretch stretch)
+		{
+			const string Redish = "#FFEB1C24";
+			const string Yellowish = "#FFFEF200";
+			const string Greenish = "#FF0ED145";
+			const string White = "#FFFFFFFF";
```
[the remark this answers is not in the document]
amelield: In the case of Skia, only one subtest is not working. Do we want to cancel all tests because of that?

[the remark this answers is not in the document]
jeromelaban: That's fair, let's add an if around with the stretch that's not working :)

PR #9948, opened 2022-09-23
fix(gen): Report error diagnostic on SnapShotTestGenerator exceptions and avoid Workspaces dependency

Youssef1313:
### First commit 4a0580d

The generator was crashing, and the exception was turning into a compiler warning. Since we don't yet treat all warnings as errors, this was missed.

It was likely broken due to version mismatch described here https://github.com/unoplatform/uno/blob/2738b04461e4e798a77afce8afd4d2fb8065b795/src/SamplesApp/SamplesApp.UITests.Generator/Uno.Samples.UITest.Generator.csproj#L17-L21

This commit (4a0580d) turns the exception into an error diagnostic, so that CI fails if any exception is thrown.

In the next commits, I'll get rid of the Workspaces dependency completely.

------

### Second commit b40a776

This commit moves us away from using Workspaces APIs.

------

### Third commit c1b654e

This commit is minor code cleanup.

### DIFF
--- a/src/SamplesApp/SamplesApp.UITests.Generator/SnapShotTestGenerator.cs
+++ b/src/SamplesApp/SamplesApp.UITests.Generator/SnapShotTestGenerator.cs
@@ -1,4 +1,6 @@
-﻿using System;
+﻿#nullable enable
+
+using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.IO;
@@ -23,6 +25,8 @@ namespace Uno.Samples.UITest.Generator
 			DiagnosticSeverity.Error,
 			isEnabledByDefault: true);
 
+		private static readonly string[] _defaultCategories = new[] { "Default" };
+
 		private const int GroupCount = 5;
 
 		public void Initialize(GeneratorInitializationContext context)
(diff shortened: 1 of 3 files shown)


jeromelaban (approved):
Nice work!

src/SamplesApp/SamplesApp.UITests.Generator/SnapShotTestGenerator.cs
```diff
@@ -215,45 +220,16 @@ private static object Sanitize(string category)
...
-				project = project.AddMetadataReference(MetadataReference.CreateFromFile(new Uri(file).LocalPath));
-			}
-
-			project = AddFiles(context, project, "UITests.Shared");
-			project = AddFiles(context, project, "SamplesApp.UnitTests.Shared");
-
-			var compilation = project.GetCompilationAsync().Result;
-
-			return compilation;
+			return context.Compilation.AddSyntaxTrees(GetSyntaxTrees(context, "UITests.Shared").Concat(GetSyntaxTrees(context, "SamplesApp.UnitTests.Shared")));
```
jeromelaban, line 223: Actually, this does not seem to work properly. Take a look at the snapshot tests on iOS, they don't run the tests, likely showing that the files are not picked up properly.

Youssef1313: Umm... That's strange

image

It seems to be working on my side.

image

Youssef1313:
> ##[warning]CSC(0,0): Warning CS8032: An instance of analyzer Uno.Samples.UITest.Generator.SnapShotTestGenerator cannot be created from D:\a\1\s\src\SamplesApp\SamplesApp.UITests.Generator\bin\Release\Uno.Samples.UITest.Generator.dll : Could not load file or assembly 'Microsoft.CodeAnalysis, Version=4.4.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35' or one of its dependencies. The system cannot find the file specified..

Found it. It would have been really much easier if this occurred as a warning. Hence, I'll add this specific warning to the list of WarningsAsErrors

Youssef1313:
@jeromelaban 

>   Test Count: 250, Passed: 228, Failed: 0, Warnings: 0, Inconclusive: 0, Skipped: 22

🎉 🎉 

I actually didn't fix/change anything
We were previously using a lower compiler version which got fixed by https://github.com/unoplatform/uno/pull/9954

jeromelaban:
Very good! Thanks :)

Issue #5440, opened 2021-03-10
CornerRadius doesn't behave the same as UWP when very large


jeanplevesque:
## Current behavior

iOS & Android:
![image](https://user-images.githubusercontent.com/39710855/110697054-d193f680-81b9-11eb-8263-cbb9ba3204cd.png)

## Expected behavior

UWP:
![image](https://user-images.githubusercontent.com/39710855/110696652-5599ae80-81b9-11eb-9b66-ea093aa9f96e.png)

## How to reproduce it (as minimally and precisely as possible)

```xml
<Border CornerRadius="128"
	Width="128"
	Height="64"
	Background="Black" />
```

## Environment

Nuget Package:

- [x] Uno.UI / Uno.UI.WebAssembly / Uno.UI.Skia
- [ ] Uno.WinUI / Uno.WinUI.WebAssembly / Uno.WinUI.Skia
- [ ] Uno.SourceGenerationTasks
- [ ] Uno.UI.RemoteControl / Uno.WinUI.RemoteControl
- [ ] Other: 

Nuget Package Version(s): **3.5.1**

Affected platform(s):

- [x] iOS
- [x] Android

The following were not tested.

- [x] WebAssembly
- [ ] WebAssembly renderers for Xamarin.Forms
- [ ] macOS
- [ ] Skia
  - [ ] WPF
  - [ ] GTK (Linux)
  - [ ] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

IDE:

- [ ] Visual Studio 2017 (version: )
- [x] Visual Studio 2019 (version: 16.8.6)
- [ ] Visual Studio for Mac (version: )
- [ ] Rider Windows (version: )
- [ ] Rider macOS (version: )
- [ ] Visual Studio Code (version: )

Relevant plugins:

- [ ] Resharper (version: )

## Anything else we need to know?

This strategy of making elliptic borders is useful for scale up animations because the Border draws better.
A scaled up Ellipse becomes pixelated and a scaled up Border doesn't become pixelated.

mikernet:
`3.6.0-dev` also developed a worse regression at some point after 3.5.1, at least on iOS where I am currently testing - if you do something like:

```xml
<Border Margin="20" 
        CornerRadius="9999"
	Width="128"
	Height="64"
	Background="Black" />
```

The shape starts to do very odd things like adding a wide shape on top of the circle that spans across the page.

agneszitte-nventive:
[...]

PR #9958, opened 2022-09-25
fix(Wasm): Fix BorderRadius not rendering properly when it's greater than half of width or height

Youssef1313:
GitHub Issue (If applicable): The Wasm part of #5440. Other platforms still need a fix. I'll take a look to see if I'm able to.

## PR Type

- Bugfix

If the CornerRadius is larger than half of element width or larger than half of element height, the clipping is done as a circular shape, which doesn't match UWP/WinUI behavior.

## What is the new behavior?

The CornerRadius is now applied correctly as an elliptic shape.

image

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/Border/BorderLayerRenderer.wasm.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Border/BorderLayerRenderer.wasm.cs
@@ -55,7 +55,7 @@ namespace Windows.UI.Xaml.Shapes
 			}
 			else
 			{
-				var borderRadiusCssString = $"{cornerRadius.TopLeft.ToStringInvariant()}px {cornerRadius.TopRight.ToStringInvariant()}px {cornerRadius.BottomRight.ToStringInvariant()}px {cornerRadius.BottomLeft.ToStringInvariant()}px";
+				var borderRadiusCssString = $"min(50%,{cornerRadius.TopLeft.ToStringInvariant()}px) min(50%,{cornerRadius.TopRight.ToStringInvariant()}px) min(50%,{cornerRadius.BottomRight.ToStringInvariant()}px) min(50%,{cornerRadius.BottomLeft.ToStringInvariant()}px)";
 				element.SetStyle(
 					("border-radius", borderRadiusCssString),
 					("overflow", "hidden")); // overflow: hidden is required here because the clipping can't do its job when it's non-rectangular.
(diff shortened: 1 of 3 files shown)


Youssef1313 (commented):
@amelield @jeromelaban I think the test I'm modifying here is a candidate to be a runtime test (except for Wasm).

jeromelaban:
@Youssef1313 make sure to validate the screenshot tests for regressions.

src/Uno.UI/UI/Xaml/Controls/Border/BorderLayerRenderer.wasm.cs
```diff
@@ -55,7 +55,7 @@ public static void SetCornerRadius(UIElement element, CornerRadius cornerRadius)
 			}
 			else
 			{
-				var borderRadiusCssString = $"{cornerRadius.TopLeft.ToStringInvariant()}px {cornerRadius.TopRight.ToStringInvariant()}px {cornerRadius.BottomRight.ToStringInvariant()}px {cornerRadius.BottomLeft.ToStringInvariant()}px";
+				var borderRadiusCssString = $"min(50%,{cornerRadius.TopLeft.ToStringInvariant()}px) min(50%,{cornerRadius.TopRight.ToStringInvariant()}px) min(50%,{cornerRadius.BottomRight.ToStringInvariant()}px) min(50%,{cornerRadius.BottomLeft.ToStringInvariant()}px)";
```
carldebilly, line 58: Is there a way to make this string handing in TS instead? It should be less expensive.

carldebilly (approved):

src/Uno.UI/UI/Xaml/Controls/Border/BorderLayerRenderer.wasm.cs
Youssef1313: @carldebilly We are already calling `SetStyles` in this code path to set `overflow: hidden`. I guess if I'll be adding `min(50%, ...)` in JS this will be a separate interop call which is more expensive? Is there something I'm missing?

Youssef1313:
> @Youssef1313 make sure to validate the screenshot tests for regressions.

Is this about the snapshot-compare from CI artifacts?

image

Do I only care for those who have the third column red?

MartinZikmund (approved):

jeromelaban:
That's correct, try seeing if there are changes that do not make sense, that are specifically related to border rendering.

Youssef1313:
There is:

image


vs

image

and also

image

vs

image

Nothing looks interesting?

Youssef1313:
> Best is to look at the xor diff image, what do you see ?

Is that somewhere in snapshot-compare-results from CI?

jeromelaban:
It's available when clicking on the `D` link in the snapshot compare report on each line.

Youssef1313:
image

I don't know if this one is interesting. To the eyes, the two screenshots are identical.

This is for UITests.Windows_UI_Xaml_Controls.TextBox.TextBox_CornerRadius.png

jeromelaban:
It shows that there's a subtle difference in the rendering, likely a result of your change. It's good to see that it's being caught visually, but indeed it's not a deal breaker.

Youssef1313:
@jeromelaban So is there anything remaining for me here?

jeromelaban (approved):
This is definitely an improvement:
image

Nice work!

Issue #7893, opened 2022-01-23
x:Bind Pathless Casting (this) references cause source generator error


hawkerm:
### Current behavior

Was using [pathless casting](https://docs.microsoft.com/en-us/windows/uwp/xaml-platform/x-bind-markup-extension#pathless-casting) in UWP data template to bind to the 'this' object of the data template:
```xml
<DataTemplate x:DataType="local:SomeTypeName">
  <local:MyControl MyProperty="{x:Bind (local:SomeTypeName)}"/>
</DataTemplate>
```

This gave me the following error when trying to run in WASM:

```
Uno.UI.SourceGenerators.XamlGenerator.XamlCodeGenerator\MainPage_bf38630c1c1d5b8ecc552685d35228f7.cs(914,322,914,324): error CS7000: Unexpected use of an aliased name
```

Switching out to old-style `{Binding}` syntax allowed me to compile.

### Expected behavior

Should accept the pathless casting syntax for x:Bind.

### Workaround

Switch x:Bind pathless cast to Binding if only doing direct. Not sure if there's the same issue for function binding which I'm not sure how you'd work around.

### Works on UWP/WinUI

Yes

### Environment

Uno.UI / Uno.UI.WebAssembly / Uno.UI.Skia

### NuGet package version(s)

4.0.11

### Affected platforms

WebAssembly

### IDE

Visual Studio 2022

### IDE version

17.0.5

jeromelaban:
Thanks for the report! Would you have the exact type that was bound? We're looking for an exact repro, as previous ones we got were not successful.

hawkerm:
It's my own type implemented in my project:

```cs
public partial class MyClass : ObservableObject
{
```

With a couple of simple properties (string and custom enum), nothing more.

It is being built using the new MVVM Toolkit source generators (initially V1, now V2), so maybe there's a conflict between the two source generators somehow? (If you're not seeing it with regular types?)

Though I tried switching it to a regular `ObservableObject` without the partial class and think I'm still getting the same message. So it may _not_ be related.

If I remove the pathless cast to the whole type, and instead use a specific property, then I don't get the error message.

jeromelaban:
Thanks for the update! Yes it may seem that the parser is not handling this very well.

## Note for contributors
This is the location of the parser tests: 

[...]

PR #10013, opened 2022-09-28
feat(xbind): Add pathless casting support

jeromelaban:
GitHub Issue (If applicable): closes https://github.com/unoplatform/uno/issues/7893

## PR Type

- Feature

## What is the new behavior?

Adds support for https://docs.microsoft.com/en-us/windows/uwp/xaml-platform/x-bind-markup-extension#pathless-casting

## PR Checklist

## Other information

### DIFF
--- a/src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/Utils/XBindExpressionParser.cs
+++ b/src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/Utils/XBindExpressionParser.cs
@@ -104,15 +104,27 @@ namespace Uno.UI.SourceGenerators.XamlGenerator.Utils
 				var e = base.VisitMemberAccessExpression(node);
 				var isValidParent = !Helpers.IsInsideMethod(node).result && !Helpers.IsInsideMemberAccessExpression(node).result;
 				var isParentMemberStatic = node.Expression is MemberAccessExpressionSyntax m && _isStaticMember(m.ToFullString());
+				var isPathLessCast = Helpers.IsPathLessCast(node);
 
 				if (e!= null && isValidParent && !_isStaticMember(node.Expression.ToFullString()) && !isParentMemberStatic)
 				{
-					var expression = e.ToFullString();
-					var contextBuilder = _isStaticMember(expression) ? "" : ContextBuilder;
-					var output = ParseCompilationUnit($"class __Temp {{ private Func<object> __prop => {contextBuilder}{expression}; }}");
+					if (isPathLessCast.result)
+					{
+						var expression = e.ToFullString();
+						var output = ParseCompilationUnit($"class __Temp {{ private Func<object> __prop => ({isPathLessCast.expression?.Expression}){_contextName}; }}");
 
-					var o2 = output.DescendantNodes().OfType<ArrowExpressionClauseSyntax>().First().Expression;
-					return o2;
+						var newSyntax = output.DescendantNodes().OfType<ArrowExpressionClauseSyntax>().First().Expression;
+						return newSyntax;
+					}
+					else
+					{
+						var expression = e.ToFullString();
+						var contextBuilder = _isStaticMember(expression) ? "" : ContextBuilder;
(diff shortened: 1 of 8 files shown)


carldebilly (approved):

dr1rrb (approved):

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/Utils/XBindExpressionParser.cs
```diff
@@ -104,15 +104,27 @@ private string ContextBuilder
...
+						return newSyntax;
+					}
+					else
+					{
+						var expression = e.ToFullString();
+						var contextBuilder = _isStaticMember(expression) ? "" : ContextBuilder;
+						var output = ParseCompilationUnit($"class __Temp {{ private Func<object> __prop => {contextBuilder}{expression}; }}");
+
+						var newSyntax = output.DescendantNodes().OfType<ArrowExpressionClauseSyntax>().First().Expression;
+						return newSyntax;
```
Youssef1313, line 126: Looks like this is more work than needed? Isn't `return SyntaxFactory.ParseExpression($"{contextBuilder}{expression};");` what we need?
We are creating a whole compilation unit just to get the expression syntax from it. Any thing I'm missing?

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/Utils/XBindExpressionParser.cs
```diff
@@ -259,6 +271,31 @@ internal static (bool result, CastExpressionSyntax? expression) IsInsideCast(Syn
...
+			{
+				var currentNode = node.Parent;
+
+				do
+				{
+					if (currentNode is ArgumentSyntax arg
+						&& arg.Expression is ParenthesizedExpressionSyntax expressionSyntax)
+					{
+						return (true, expressionSyntax);
+					}
```
Youssef1313, line 285: Consider validating the multiple redundant parentheses case. ie, if `expressionSyntax.Expression` is another `ParenthesizedExpressionSyntax` (etc..), should we unwrap them all and return a non-ParenthesizedExpressionSyntax ?

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/Utils/XBindExpressionParser.cs
jeromelaban: It's only a `CompilationUnit` syntax, and the idea behind it is to have a proper resolution context so that the nodes returned are from a method body, not just a top-level statement that may not make sense alone.

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/Utils/XBindExpressionParser.cs
jeromelaban: It cannot happen in that context, the user will not provide advanced constructs in that context (they'll fail later on a compile time).

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/Utils/XBindExpressionParser.cs
jeromelaban: Still, this part is duplicated many times, it's time for a refactor ;)

Youssef1313: Is there any known case where ParseExpression will not produce the expected result?

I did a quick benchmark and the ParseExpression approach appears to be faster:

image

Not sure if the difference is going to be noticeable in our case, but something to consider if it maintains the code correctness.

jeromelaban: It's an interesting benchmark. It's only used in smaller use cases when x:Bind expressions are used, but it can make it faster anyways, I'll check in the other PR. Thanks!

PR #10033, opened 2022-10-02
feat(Clipboard): WASM rich clipboard support

### LINKED ISSUE #10032: [WASM] Multiple data type support for Clipboard
## For which Platform:

- [ ] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] macOS
- [ ] Skia
  - [ ] WPF
  - [ ] GTK (Linux)
  - [ ] Linux Framebuffer
  - [ ] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

trungnt2910:
Added support for content other than plain text on WASM, using the JavaScript raw `Clipboard.read()` and `Clipboard.write()` APIs.

The clipboard can now read/write HTML text and PNG images, but is subject to a few restrictions of the underlying APIs.

GitHub Issue (If applicable): closes #10032

## PR Type

- Feature

WASM `Clipboard` API can only read plain text content.

## What is the new behavior?

WASM `Clipboard` API can now read HTML and PNG content.
Other content (RTF rich text, other image formats,...) cannot be read because [major browsers don't seem to support content other than plain, html and png](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/clipboard/clipboard_writer.cc;drc=e882b8e4a8272f65cb14c608d3d2bc4f0512aa20;l=304).

A screenshot of the Uno Gallery sample app at https://github.com/unoplatform/Uno.Gallery/pull/467 combined with code from this pull request:

image

Unlike on UWP and Skia.WPF/GTK, the content types of the clipboard is not known after `Clipboard.GetContent` is called, but rather resolved asynchronously. Therefore, there can be false positives, such as `dataPackageView.Contains(StandardDataFormats.Text)` returning `true` despite the Clipboard containing only a single image. In such cases, an empty string, or stream in case of a bitmap, is returned.

To resolve this problem, call any of the `Get...` methods of the `DataPackageView` _before_ checking for `Contains`:

```csharp
	private async void PasteText_Click(object sender, RoutedEventArgs args)
	{
		var package = Clipboard.GetContent();
                _ = await package.GetTextAsync();
                // Now things should be valid.
		if (package.Contains(StandardDataFormats.Bitmap))
		{
			var streamRef = await package.GetBitmapAsync();
		}
	}
```

This should _not_ be a breaking change, as `package.Contains(StandardDataFormats.Text)` in the official Uno packages currently always returns `true` regardless of whether text is available on the clipboard or not.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UWP/ApplicationModel/DataTransfer/Clipboard.wasm.cs
+++ b/src/Uno.UWP/ApplicationModel/DataTransfer/Clipboard.wasm.cs
@@ -3,9 +3,6 @@
 using System;
 using System.IO;
 using System.Linq;
-using System.Runtime.InteropServices;
-using System.Runtime.Serialization;
-using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Uno.Disposables;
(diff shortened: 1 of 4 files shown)


src/Uno.UWP/ApplicationModel/DataTransfer/Clipboard.wasm.cs
```diff
@@ -57,6 +223,50 @@ private static void SetClipboardText(string text)
...
+		private static async Task<MarshalledData> GetClipboardData(string mime, CancellationToken ct)
+		{
+			var escapedMime = WebAssemblyRuntime.EscapeJs(mime);
+			var command = $@"
+				(async () => {{
+					var result = await {JsType}.getData(""{escapedMime}"");
+					if (!result) return null;
+					return JSON.stringify(result);
+				}})()
+				";
```
jeromelaban, line 258: This block should not be written in C# and use the "JsType" type in typecript instead.

src/Uno.UWP/ApplicationModel/DataTransfer/Clipboard.wasm.cs
```diff
@@ -1,21 +1,159 @@
...
+using Windows.Storage.Streams;
 
 namespace Windows.ApplicationModel.DataTransfer
 {
 	public static partial class Clipboard
 	{
 		private const string JsType = "Uno.Utils.Clipboard";
 
+		[DataContract]
+		private class MarshalledData: IDisposable
```
jeromelaban, line 23: Move this type to a Clipboard.MarshalledData.wasm.cs for readability. You can keep it as a nested type.

src/Uno.UWP/ApplicationModel/DataTransfer/Clipboard.wasm.cs
```diff
@@ -1,21 +1,159 @@
...
+
+			public void Dispose()
+			{
+				Dispose(disposing: true);
+				GC.SuppressFinalize(this);
+			}
+		}
+
+		[DataContract]
+		private class MarshalledDataWithMime: MarshalledData
```
jeromelaban, line 89: Same here, move to a different file.

src/Uno.UWP/ApplicationModel/DataTransfer/Clipboard.wasm.cs
trungnt2910: I wonder if such a block deserves its own TypeScript function, as the only purpose it serves is to marshal the JavaScript object to something C# can read.

src/Uno.UWP/ApplicationModel/DataTransfer/Clipboard.wasm.cs
trungnt2910: Done.

src/Uno.UWP/ApplicationModel/DataTransfer/Clipboard.wasm.cs
trungnt2910: Done.

MartinZikmund (approved):

jeromelaban:
We're making large changes to the .NET 7 support which will remove the use of JSEval for any interop (so that we support #11769). It's fine if this PR gets adjusted to only work on net7+. This will invalidate the comment about TSInteropMarshaller, as TSImport/TSExport are a lot easier to use.

jeromelaban:
Thanks for your contribution! 

Since we introduced CSP support, it is now required to use TSImport/TSExport to implement this feature, which is very different from the code from this PR. If you wish to continue the work on this PR, please reopen an new one and let us know!

PR #10042, opened 2022-10-03
feat(WASM): Support setting the screen orientation

### LINKED ISSUE #3177: Add screen orientation support on WASM

Support for screen orientation via [Screen.orientation API](https://developer.mozilla.org/en-US/docs/Web/API/Screen/orientation).

Can also include screen orientation locking/unlocking (`AutoRotationPreferences` in UWP)

## Why is this needed:

Feature parity

## For which Platform:

- [ ] iOS
- [ ] Android
- [x] WebAssembly
- [ ] macOS
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

trungnt2910:
Supports setting the screen orientation through the `DisplayInformation.AutoRotationPreferences` on supported browsers.

Internally the implementation calls the JavaScript `screen.orientation.lock` API.

GitHub Issue (If applicable): closes #3177

## PR Type

- Feature

Uno WASM applications ignore the `DisplayInformation.AutoRotationPreferences` property.

## What is the new behavior?

Uno WASM applications attempts to use the `screen.orientation.lock` JavaScript API to set the screen orientation on supported devices.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UWP/Graphics/Display/DisplayInformation.wasm.cs
+++ b/src/Uno.UWP/Graphics/Display/DisplayInformation.wasm.cs
@@ -1,8 +1,10 @@
 ﻿#if __WASM__
-using Uno;
-using Uno.Foundation;
 using System;
 using System.Globalization;
+using System.Threading;
+using System.Threading.Tasks;
+using Uno;
+using Uno.Foundation;
 
 namespace Windows.Graphics.Display
 {
@@ -134,6 +136,13 @@ namespace Windows.Graphics.Display
 			WebAssemblyRuntime.InvokeJS(command);
 		}
 
+		static partial void SetOrientationPartial(DisplayOrientations orientations)
+		{
+			Uno.UI.Dispatching.CoreDispatcher.Main.RunAsync(
+				Uno.UI.Dispatching.CoreDispatcherPriority.High,
+				(ct) => SetOrientationAsync(orientations, ct));
+		}
+
 		private static bool TryReadJsFloat(string property, out float value) =>
 			float.TryParse(WebAssemblyRuntime.InvokeJS(property), NumberStyles.Any, CultureInfo.InvariantCulture, out value);
 
@@ -151,6 +160,11 @@ namespace Windows.Graphics.Display
 				"landscape-secondary" => DisplayOrientations.LandscapeFlipped,
 				_ => DisplayOrientations.None,
 			};
+		}
+
+		private static Task SetOrientationAsync(DisplayOrientations orientations, CancellationToken ct)
+		{
+			return WebAssemblyRuntime.InvokeAsync($"{JsType}.setOrientationAsync({(int)orientations})", ct);
 		}
 	}
 }
--- a/src/Uno.UI/WasmScripts/Uno.UI.d.ts
+++ b/src/Uno.UI/WasmScripts/Uno.UI.d.ts
@@ -864,19 +864,30 @@ declare namespace Windows.Gaming.Input {
     }
 }
 declare namespace Windows.Graphics.Display {
-    class DisplayInformation {
+    enum DisplayOrientations {
+        None = 0,
+        Landscape = 1,
+        Portrait = 2,
+        LandscapeFlipped = 4,
+        PortraitFlipped = 8
+    }
+    export class DisplayInformation {
         private static readonly DpiCheckInterval;
         private static lastDpi;
         private static dpiWatcher;
         private static dispatchOrientationChanged;
         private static dispatchDpiChanged;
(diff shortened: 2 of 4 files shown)


src/Uno.UWP/Graphics/Display/DisplayInformation.wasm.cs
```diff
@@ -152,6 +161,24 @@ private static DisplayOrientations ParseJsOrientation(string jsOrientation)
...
+				DisplayOrientations.Portrait => "portrait-primary",
+				DisplayOrientations.PortraitFlipped => "portrait-secondary",
+				_ => "any"
+			};
+		}
+
+		private static Task SetOrientationAsync(DisplayOrientations orientations, CancellationToken ct)
+		{
+			var jsOrientation = WebAssemblyRuntime.EscapeJs(ToJsOrientation(orientations));
+			return WebAssemblyRuntime.InvokeAsync($"{JsType}.setOrientationAsync(\"{jsOrientation}\")", ct);
```
Youssef1313, line 180: Not sure if it's worth doing a check to call this only if `jsOrientation` is different than the existing old value

trungnt2910: To get the existing value we would need a Javascript interop call anyway.

src/Uno.UWP/Graphics/Display/DisplayInformation.wasm.cs
```diff
@@ -152,6 +161,24 @@ private static DisplayOrientations ParseJsOrientation(string jsOrientation)
...
+		private static string ToJsOrientation(DisplayOrientations orientations)
+		{
+			return orientations switch
+			{
+				DisplayOrientations.Landscape => "landscape-primary",
+				DisplayOrientations.LandscapeFlipped => "landscape-secondary",
+				DisplayOrientations.Portrait => "portrait-primary",
+				DisplayOrientations.PortraitFlipped => "portrait-secondary",
+				_ => "any"
+			};
```
Youssef1313, line 174: `DisplayOrientations` is a flags enum. So, multiple values can be specified.

I think we should well understand the meaning of them, and use the corresponding JS orientation.

src/Uno.UWP/Graphics/Display/DisplayInformation.wasm.cs
```diff
@@ -152,6 +161,24 @@ private static DisplayOrientations ParseJsOrientation(string jsOrientation)
...
+
+		private static string ToJsOrientation(DisplayOrientations orientations)
+		{
+			return orientations switch
+			{
+				DisplayOrientations.Landscape => "landscape-primary",
+				DisplayOrientations.LandscapeFlipped => "landscape-secondary",
+				DisplayOrientations.Portrait => "portrait-primary",
+				DisplayOrientations.PortraitFlipped => "portrait-secondary",
+				_ => "any"
```
Youssef1313, line 173: Did you find "any" in some documentation or just arbitrary string?

trungnt2910: It's documented [here](https://developer.mozilla.org/en-US/docs/Web/API/ScreenOrientation/lock).

src/Uno.UWP/Graphics/Display/DisplayInformation.wasm.cs
trungnt2910: It seems like if multiple values are specified, the system gets to choose which orientation.

So `Portrait | PortraitFlipped` can be mapped to the JS `portrait`, while `Landscape | LanscapeFlipped` can be mapped to the JS `landscape`.

However, for other values, I don't think there's a suitable mapping. A solution might be passing a list to the TypeScript function, and then return when setting to one of the values succeed?

Youssef1313: @MartinZikmund Could you advise here?

trungnt2910: I've removed the mentioned code. Instead, I'm passing the value as-is to the JavaScript code (passing an int is much cheaper than passing an array of strings, which would require JSON serialization).

The JavaScript code would then decode the enum to a list of `OrientationLockType`, and attempt to lock the orientation at one of the types specified.

trungnt2910: This code is outdated, so I'll mark this as resolved.

MartinZikmund (approved):
Very nice! It is interesting that browsers require full-screen 😀 . Approving!

PR #10060, opened 2022-10-04
fix: Ensure ShowAsync returns if popup is forcibly closed

### LINKED ISSUE #10038: [Navigation] NavigateViewModelForResultAsync .AsResult never finishes if you close content dialog by putting app in background
## Current behavior
backgroundbugq

## Expected behavior

It should have the same behavior as when you just navigateback. Or the Content dialog should not close itself

Steps to reproduce:
Open the app,
press on the button
put the app in background
notice the button is now disabled, because task is still running.

To Make another sample app, you just need to open a contentDialog (it's important that it's a content dialog, this problem does not happen with a page) with navigateViewModelForResultAsync (...........) .AsResult and await the result.

The task never ends.

[MyAppTest2.zip](https://github.com/unoplatform/uno.extensions/files/9685275/MyAppTest2.zip)

## Environment

Nuget Package:
Uno.Extensions / winui

Package Version(s):

	  PackageReference Update="Uno.Extensions.Configuration" Version="2.3.0-dev.69" />
	  PackageReference Update="Uno.Extensions.Core" Version="2.3.0-dev.69" />
	  PackageReference Update="Uno.Extensions.Hosting" Version="2.3.0-dev.69" />
	  PackageReference Update="Uno.Extensions.Hosting.WinUI" Version="2.3.0-dev.69" />
	  PackageReference Update="Uno.Extensions.Http" Version="2.3.0-dev.69" />
	  PackageReference Update="Uno.Extensions.Http.Refit" Version="2.3.0-dev.69" />
	  PackageReference Update="Uno.Extensions.Localization.WinUI" Version="2.3.0-dev.69" />
	  PackageReference Update="Uno.Extensions.Logging.Serilog" Version="2.3.0-dev.69" />
	  PackageReference Update="Uno.Extensions.Logging.WinUI" Version="2.3.0-dev.69" />
	  PackageReference Update="Uno.Extensions.Navigation" Version="2.3.0-dev.69" />
	  PackageReference Update="Uno.Extensions.Navigation.Toolkit.WinUI" Version="2.3.0-dev.69" />
	  PackageReference Update="Uno.Extensions.Navigation.WinUI" Version="2.3.0-dev.69" />
	  PackageReference Update="Uno.Extensions.Reactive" Version="2.3.0-dev.69" />
	  PackageReference Update="Uno.Extensions.Reactive.WinUI" Version="2.3.0-dev.69" />
	  PackageReference Update="Uno.Extensions.Serialization" Version="2.3.0-dev.69" />
	  PackageReference Update="Uno.Extensions.Serialization.Http" Version="2.3.0-dev.69" />
	  PackageReference Update="Uno.Extensions.Serialization.Refit" Version="2.3.0-dev.69" />

Affected platform(s):

- [X] iOS (could not open it on my repro app, but it happens on another app)
- [X] Android
- [ ] WebAssembly
- [ ] WebAssembly renders for Xamarin.Forms
- [ ] Windows
- [ ] Build tasks

Visual Studio:

- [ ] 2017 (version: )
- [X] 2022 (version: 17.3.4 )
- [ ] for Mac (version: )

Relevant plugins:

- [ ] Resharper (version: )

kazo0:
closes #10038

## PR Type

- Bugfix

The ContentDialog's popup may be forcibly closed by things such as the app moving to the background, when this happens, the ShowAsync method never returns

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/ContentDialog/ContentDialog.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/ContentDialog/ContentDialog.cs
@@ -70,6 +70,15 @@ namespace Windows.UI.Xaml.Controls
 					that.UpdateVisualState();
 				}
 			};
+
+			_popup.Closed += (s, e) =>
+			{
+				if (thisRef.Target is ContentDialog that)
+				{
+					that.Hide();
+				}
+			};
+
 			this.KeyDown += OnPopupKeyDown;
 			var inputPane = InputPane.GetForCurrentView();
 			inputPane.Showing += (o, e) =>
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_ContentDialog.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_ContentDialog.cs
@@ -327,6 +327,56 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 				SUT.Hide();
 			}
 		}
+
+		[TestMethod]
+#if __MACOS__
+		[Ignore("Currently fails on macOS, part of #9282 epic")]
+#endif
+		public async Task When_Popup_Closed()
+		{
+			var closedEvent = new Event();
+			var openedEvent = new Event();
+			var closedRegistration = new SafeEventRegistration<ContentDialog, TypedEventHandler<ContentDialog, ContentDialogClosedEventArgs>>("Closed");
+			var openedRegistration = new SafeEventRegistration<ContentDialog, TypedEventHandler<ContentDialog, ContentDialogOpenedEventArgs>>("Opened");
+
+			var SUT = new MyContentDialog
+			{
+				Title = "Dialog title",
+				Content = "Dialog content",
+				CloseButtonText = "Close",
+			};
+
+			closedRegistration.Attach(SUT, (s, e) => closedEvent.Set());
+			openedRegistration.Attach(SUT, (s, e) => openedEvent.Set());
+
+			try
+			{
+				var showAsyncResult = SUT.ShowAsync().AsTask();
+
+				await openedEvent.WaitForDefault();
+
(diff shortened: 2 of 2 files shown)


src/Uno.UI/UI/Xaml/Controls/ContentDialog/ContentDialog.cs
```diff
@@ -69,6 +69,15 @@ public ContentDialog() : base()
 					that.UpdateVisualState();
 				}
 			};
+
+			_popup.Closed += (s, e) =>
+			{
+				if (thisRef.Target is ContentDialog that)
+				{
+					that.Hide();
+				}
```
Youssef1313, line 78: Do we need something similar for other controls that internally create a Popup?

kazo0:
tests coming shortly

src/Uno.UI/UI/Xaml/Controls/ContentDialog/ContentDialog.cs
jeromelaban: I'd guess that flyout would need the same treatment. Also, a runtime test would be needed :)

kazo0: I'm not sure we'd need this for other things such as Flyout. The problem here is because we are expecting a result to be returned from the ContentDialog once it closes. Without this logic, the `ShowAsync()` call never returns the expected `ContentDialogResult`. Whereas Flyout just has a void ShowAt, so nothing is waiting for the result of the method. And Flyout already takes action to hide itself when its Popup is closed, [here](https://github.com/unoplatform/uno/blob/e8d56677ba8af1cca929e2ded5693d69c8ab764b/src/Uno.UI/UI/Xaml/Controls/Flyout/FlyoutBase.cs#L409-L413)

MartinZikmund (approved):
As Jerome mentioned, runtime test would be useful for this, to preserve the behavior when we port `Popup` in the future.

spouliot (approved):

PR #10062, opened 2022-10-04
fix(ContentDialog): Handle back button being pressed

kazo0:
closes https://github.com/unoplatform/uno.extensions/issues/768

## PR Type

- Bugfix

A back navigation or app closure will occur when tapping the back button while a contentdialog is being displayed

## What is the new behavior?

while dialog is open, back should just close the dialog and do nothing else

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/ContentDialog/ContentDialog.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/ContentDialog/ContentDialog.cs
@@ -14,6 +14,7 @@ using Windows.System;
 using Uno.UI.DataBinding;
 using Windows.UI.Xaml.Controls.Primitives;
 using Windows.UI.ViewManagement;
+using Windows.UI.Core;
 
 #if HAS_UNO_WINUI
 using WindowSizeChangedEventArgs = Microsoft.UI.Xaml.WindowSizeChangedEventArgs;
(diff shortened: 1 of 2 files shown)


kazo0:
Tests coming shortly

src/Uno.UI/UI/Xaml/Controls/ContentDialog/ContentDialog.cs
```diff
@@ -345,8 +346,34 @@ private void RegisterEvents()
 				window.SizeChanged -= WindowSizeChanged;
 			});
 
+			if (SystemNavigationManager.GetForCurrentView() is { } navManager)
```
jeromelaban, line 349: I wonder if UWP did it that way... We would have to test on Win10 which still has that mode :)

kazo0: They don't do it this way, ideally the BackRequested event is never fired in this scenario. We would need to have some sort of new internal event that gets fired on back pressed, and then BackRequested would only fire if that new event's Handled was still false

MartinZikmund (approved):

src/Uno.UI/UI/Xaml/Controls/ContentDialog/ContentDialog.cs
jeromelaban: Makes sense. Let's explain this in the code as a comment.

jeromelaban (approved):

PR #10079, opened 2022-10-06
ci: Enforce whitespace formatting in CI

Youssef1313:

## PR Type

## PR Checklist

## Other information


src/AddIns/Uno.UI.Lottie/LottieVisualSourceProvider.cs
```diff
@@ -7,7 +7,7 @@
 #pragma warning disable 105
 // Keep the space as this will not be replaced by the
 // WinUI upgrade tool.
-using Microsoft. UI.Xaml.Controls;
+using Microsoft.UI.Xaml.Controls;
```
Youssef1313, line 10: @jeromelaban Can I have more info about the problem that's being worked-around with the extra space here?

jeromelaban:
Changes overall make sense (though we should not change the System.Xaml source formatting as it comes from Mono), but the remaining problem I find here is that Visual Studio won't notify of formatting issues right away. Can you find a way to enforce it properly (Stylecop maybe)?

src/AddIns/Uno.UI.Lottie/LottieVisualSourceProvider.cs
Youssef1313: > What was the error caused by the removal?

I don't recall exactly. It's some type not found or something similar.
Anyway, I adopted adding a comment instead of the space, so this satisfies the formatting and not cause other issues.

Youssef1313:
> Changes overall make sense (though we should not change the System.Xaml source formatting as it comes from Mono), but the remaining problem I find here is that Visual Studio won't notify of formatting issues right away. Can you find a way to enforce it properly (Stylecop maybe)?

Enabling IDE0055 on build should be the way to go for this.

Youssef1313:
@jeromelaban We are hitting https://github.com/dotnet/roslyn/issues/62612

Youssef1313:
Will replace with another PR.

PR #10087, opened 2022-10-07
fix(NativeFramePresenter): Create Page instances at time of insertion into the Frame.BackStack

kazo0:
Relates to https://github.com/unoplatform/uno.toolkit.ui/issues/346

## PR Type

- Bugfix

Manually editing the Frame.BackStack while use the NativeFramePresenter on iOS leads to navigation problems. ViewControllers in the NavigationController stack will not contain references to the proper Page instances. This is causing the Frame BackStack and the NavigationController's ViewControllers to become out of sync with one another

## What is the new behavior?

Upon any change to the Frame's BackStack, the iOS NativeFramePresenter will attempt to instantiate Pages that have been added to the BackStack through an insertion of a new `PageStackEntry`.

We are also introducing a new Helper method for the `FrameNavigationHelper` that will allow the NativeFramePresenter within Toolkit, or any other Frame-reliant navigation mechanism, to achieve the same logic.

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/Frame/Frame.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Frame/Frame.cs
@@ -347,13 +347,11 @@ namespace Windows.UI.Xaml.Controls
 
 			if (CurrentEntry.Instance == null)
 			{
-				var page = CreatePageInstanceCached(entry.SourcePageType);
-				if (page == null)
+				if (EnsurePageInitialized(entry) is not { } page)
 				{
 					return false;
 				}
 
-				page.Frame = this;
 				CurrentEntry.Instance = page;
 			}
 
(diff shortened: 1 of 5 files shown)


src/Uno.UI/UI/Xaml/Controls/Frame/Frame.cs
```diff
@@ -432,6 +430,17 @@ private void ReleasePages(IList<PageStackEntry> pageStackEntries)
 
 		public void SetNavigationState(string navigationState) => _navigationState = navigationState;
 
+		internal Page EnsurePageInitialized(PageStackEntry entry)
+		{
+			if (entry is { Instance: null } && CreatePageInstanceCached(entry.SourcePageType) is { } page)
```
ajpinedam, line 435: Should this validation be separate? Like

```csharp
if (entry is { Instance: null })
{
    entry.Instance = page;
}

if(CreatePageInstanceCached(entry.SourcePageType) is { } page) 
{
    page.Frame = this;
}
```

ajpinedam (approved):
Code looks good! Thanks.-

src/Uno.UI/UI/Xaml/Controls/Frame/Frame.cs
MartinZikmund: I think the current logic is correct - as it needs to initialize the page first and then set the instance

MartinZikmund (approved):

PR #10095, opened 2022-10-08
ci: Remove unused feeds

jeromelaban:


Youssef1313 (approved):
Nice! Are we not using packages from these feeds?

jeromelaban:
@Youssef1313 one of those feed went down, breaking the build. We're not using those anymore, so it'll speed up the build a bit 🙂

Youssef1313:
> @Youssef1313 one of those feed went down, breaking the build. We're not using those anymore, so it'll speed up the build a bit 🙂

It's temporary, but definitely good to remove if they're unused. This should reduce the number of HTTP requests in restore significantly

PR #10144, opened 2022-10-15
chore: Cleanup #if conditionals

Youssef1313:
Not much valuable, just minor cleanup.

## PR Type

## PR Checklist

## Other information


Youssef1313 (commented):
@jeromelaban `PlatformItemGroups.props` and `Uno.CrossTargetting.props` are not included in some projects (e.g, SamplesApp), so `*.platform.cs` doesn't work for these cases.

Was there a reason this is not included in `Directory.Build.props`?

MartinZikmund (approved):

jeromelaban (approved):
Nice!

MartinZikmund:
@Youssef1313 seems there are some issues with `PlatformItemGroups.props`

src/Directory.Build.props
```diff
@@ -1,5 +1,29 @@
...
+
+    <IsXamarinMac>false</IsXamarinMac>
+    <IsXamarinMac Condition="
+		$(TargetFramework.ToLower().StartsWith('xamarinmac'))
+		or '$(TargetFramework)'=='net6.0-macos'
+		or '$(TargetFramework)'=='net7.0-macos'
+		">true</IsXamarinMac>
+  </PropertyGroup>
+
+  <Import Project="PlatformItemGroups.props" />
```
Youssef1313, line 26: @jeromelaban I added this import to ensure that platform file globbing works for all projects, and I moved the properties up so that they are available before they are used.

Youssef1313: Next step to be done in a future PR is cleaning this up a little bit. The property names are not accurate. The TFMs `net6.0-android` and `net7.0-android` are not mono, so `IsMonoAndroid` should be renamed to `IsAndroid`. I'll do that separately as this PR has already enough of changes.

MartinZikmund: Still seems to be acting up a bit 🤔

Youssef1313: MSBuild ordering of things has always confused me 😕 

I'm trying now with having this Import in targets instead of props (which should be imported a bit later).

If it's not late enough, I'll revert and add the import to the individual projects that need them.

jeromelaban: Add this import at the top level is likely to cause a bunch of issues, but we can try :) As for IsMonoAndroid and others, this can cause issues with the import ordering, where this file may be included before any of the sub-properties (derived from TargetFramework) are being set.

Xiaoy312 (approved):

PR #10149, opened 2022-10-16
fix: Invalid `Image` should not display

MartinZikmund:

## PR Type

- Bugfix

When `Image` source is invalid, it does display "broken image" icon on WASM. If empty, it shows border.

image

## What is the new behavior?

When `Image` source is invalid, it does not display at all.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/Image/HtmlImage.wasm.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Image/HtmlImage.wasm.cs
@@ -1,0 +1,12 @@
+﻿#nullable enable
+
+namespace Windows.UI.Xaml.Controls;
+
+public class HtmlImage : UIElement
+{
+	public HtmlImage() : base("img")
+	{
+		// Avoid the "drag effect" which is set by default in browsers
+		SetAttribute("draggable", "false");
+	}
+}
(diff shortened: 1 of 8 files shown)


jeromelaban (approved):

jeromelaban:
Nice!

Xiaoy312 (approved):

src/Uno.UI/UI/Xaml/Controls/Image/HtmlImage.wasm.cs
```diff
@@ -0,0 +1,12 @@
+﻿#nullable enable
+
+namespace Windows.UI.Xaml.Controls;
+
+public class HtmlImage : UIElement
+{
+	public HtmlImage() : base("img")
+	{
+		// Avoid the "drag effect" which is set by default in browsers
+		SetAttribute("draggable", "false");
```
carldebilly, line 10: Some browsers may also require css for this like  `user-drag: none` and the prefixed versions.

https://caniuse.com/?search=user-drag

carldebilly (approved):

src/Uno.UI/UI/Xaml/Controls/Image/HtmlImage.wasm.cs
jeromelaban: Looks like we should set it by default in css?

MartinZikmund:
@carldebilly @jeromelaban Ok, so the actual reason was a bit different - in fact the `GetPhysicalRect` returns a rect which is evaluated when accessed. So even though I queried it before the panel was hidden, it was evaluated only after - and when visibility is `Collapsed`, the element is hidden and its `left` and `top` in CSS are shifted to -10000 - which is what the rect then returns 😀

jeromelaban:
That is a tricky one indeed, I don't think can change much for that, considering that even in other frameworks, invisible elements don't even provide the element to start with (for android I'm sure of, at least).

PR #10260, opened 2022-10-26
fix: Adjust corner-radius border rendering on Skia

MartinZikmund:

## PR Type

- Bugfix

Borders on Skia did not render correctly

## What is the new behavior?

Borders render correctly

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/Border/BorderLayerRenderer.skia.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Border/BorderLayerRenderer.skia.cs
@@ -118,6 +118,7 @@ namespace Windows.UI.Xaml.Shapes
 				var borderShape = compositor.CreateSpriteShape();
 				var backgroundShape = compositor.CreateSpriteShape();
 				var outerShape = compositor.CreateSpriteShape();
+				var clipShape = compositor.CreateSpriteShape();
 
 				// Border brush
 				Brush.AssignAndObserveBrush(borderBrush, compositor, brush => borderShape.FillBrush = brush)
@@ -134,9 +135,16 @@ namespace Windows.UI.Xaml.Shapes
 						.DisposeWith(disposables);
 				}
 
-				var borderPath = GetRoundedRect(cornerRadius, innerCornerRadius, area, adjustedArea);
-				var backgroundPath = GetRoundedPath(cornerRadius, adjustedArea);
-				var outerPath = GetRoundedPath(cornerRadius, area);
+				var outerRadii = GetOuterRadii(cornerRadius);
+				var innerRadii = GetInnerRadii(cornerRadius, borderThickness);
+
+				var borderPath = GetRoundedRect(outerRadii, innerRadii, borderThickness, area, adjustedArea);
+				
+				var backgroundPath = state.BackgroundSizing == BackgroundSizing.InnerBorderEdge ?
+					GetRoundedPath(adjustedArea.ToSKRect(), innerRadii) :
+					GetRoundedPath(adjustedArea.ToSKRect(), outerRadii);
+
+				var outerPath = GetRoundedPath(area.ToSKRect(), outerRadii);
 
 				backgroundShape.Geometry = compositor.CreatePathGeometry(backgroundPath);
 				borderShape.Geometry = compositor.CreatePathGeometry(borderPath);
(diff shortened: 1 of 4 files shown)


src/Uno.UI/UI/Xaml/Controls/Border/BorderLayerRenderer.skia.cs
```diff
@@ -321,31 +314,98 @@ private static Rect CreateImageLayer(Compositor compositor, CompositeDisposable
...
+			var roundRect = new SKRoundRect();
+			roundRect.SetRectRadii(area, radii);
+			geometry.AddRoundRect(roundRect);
 			geometry.Close();
 
 			return new CompositionPath(geometrySource);
 		}
 
-		private static CompositionPath GetRoundedRect(CornerRadius cornerRadius, CornerRadius innerCornerRadius, Rect area, Rect insetArea)
+		internal static SKPoint[] GetOuterRadii(CornerRadius cornerRadius, Thickness borderThickness, float maxHorizontalRadius, float maxVerticalRadius)
```
jeromelaban, line 333: It'd probable be causing less allocations if the return value was a record struct (the size is always 4).

MartinZikmund: @jeromelaban Agreed, but the SkiaSharp `SetRectRadii` method accepts only an array with exactly 4 elements. It would be nice to reuse the array but array pool can return larger arrays, which would cause exception.

MartinZikmund: @jeromelaban  As I think about it, it should be safe to have two static fields for the arrays, because the method can only run on UI thread, so there should be no race condition, no?

jeromelaban: Indeed, because it's the UI Thread, it should be fine.

src/Uno.UI/UI/Xaml/Controls/Border/BorderLayerRenderer.skia.cs
```diff
@@ -321,31 +314,27 @@ private static Rect CreateImageLayer(Compositor compositor, CompositeDisposable
...
 		/// Creates a rounded-rectangle path from the nominated bounds and corner radius.
 		/// </summary>
-		private static CompositionPath GetRoundedPath(CornerRadius cornerRadius, Rect area, SkiaGeometrySource2D geometrySource = null)
+		private static CompositionPath GetRoundedPath(SKRect area, SKPoint[] radii, SkiaGeometrySource2D geometrySource = null)
 		{
 			geometrySource ??= new SkiaGeometrySource2D();
 			var geometry = geometrySource.Geometry;
 
 			// How ArcTo works:
 			// http://www.twistedape.me.uk/blog/2013/09/23/what-arctopointdoes/
```
spouliot, line 323: minor: comment is now useless since the code was removed

spouliot (approved):

src/Uno.UI/UI/Xaml/Controls/Border/BorderLayerRenderer.skia.cs
```diff
@@ -18,6 +18,9 @@ namespace Windows.UI.Xaml.Shapes
 {
 	partial class BorderLayerRenderer
 	{
+		private static SKPoint[] _outerRadiiStore = new SKPoint[4];
+		private static SKPoint[] _innerRadiiStore = new SKPoint[4];
```
spouliot, line 22: candidate for `stackalloc` inside the methods ?
It would avoid potential multithreading issue (even if UI code should not use multiple threads)

MartinZikmund: I think `stackalloc` does not allow assigning to an array, or does it? The reason is that I need an array to be able to pass to the Skia `SetRectRadii` method

spouliot: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-7.3/stackalloc-array-initializers

MartinZikmund: Will keep as is for now and will adjust in future PR which will also include WASM

MartinZikmund: @spouliot I wasn't able to figure out the syntax - based on the linked article, the target can be `Span<T>` or `T*`, but not sure how to get an actual array

spouliot: >  Skia SetRectRadii method

sorry missed that part :( and thought you controlled the usage of the variable

PR #10269, opened 2022-10-26
Dev/jela/skia font default

jeromelaban:
## PR Type

- Bugfix

## What is the new behavior?

Use fallback font when Skia cannot load the requested font.

## PR Checklist

## Other information


src/Uno.UI/UI/Xaml/Documents/Inline.skia.cs
```diff
@@ -115,17 +118,18 @@ protected override void OnFontWeightChanged()
 			else
 			{
 				skTypeFace = SKTypeface.FromFamilyName(name, skWeight, skWidth, skSlant);
+			}
 
+			if (skTypeFace == null)
+			{
 				// FromFontFamilyName may return null: https://github.com/mono/SkiaSharp/issues/1058
```
Youssef1313, line 125: Is this comment outdated?

src/Uno.UI/UI/Xaml/Documents/Inline.skia.cs
```diff
@@ -115,17 +118,18 @@ protected override void OnFontWeightChanged()
 			else
 			{
 				skTypeFace = SKTypeface.FromFamilyName(name, skWeight, skWidth, skSlant);
+			}
 
+			if (skTypeFace == null)
```
Youssef1313, line 123: I don't understand why this is moved outside the `else`?

The code comment says this can be null when `FromFamilyName` is called, so this only makes sense in the `else` block.

I also don't understand where the behavior change comes from?

Youssef1313: If this is reachable because `SKTypeface.FromFile(filePath)` can be null, it would be good to update the below comment so that it's clear.

MartinZikmund (approved):

spouliot (approved):

Issue #10182, opened 2022-10-18
Referencing Uno.*UI transitively should fail clearly in a WinAppSDK app


jeromelaban:
Uno nuget *.targets files should detect that a WinAppSDK app is being built, and should fail with an explicit error message.

---
_Originally posted by @jeromelaban in https://github.com/unoplatform/uno/discussions/10180#discussioncomment-3908580_

Thanks for the update. By reading the logs, you'll be able to see that Uno is transitively referenced in the Windows app, where it should not:

```
C:\Users\Jaden\source\repos\OnionMedia\OnionMedia.Uno\OnionMedia.Uno.Shared\UserControls\TimeRangeSelector.xaml error WMC9999: Type universe cannot resolve assembly: Uno.UI, Version=255.255.255.255, Culture=neutral, PublicKeyToken=null. [C:\Users\Jaden\source\repos\OnionMedia\OnionMedia.Uno\OnionMedia.Uno.Windows\OnionMedia.Uno.Windows.csproj]
```

This is caused by the Community toolkit reference. Make sure to [follow this guide](https://platform.uno/docs/articles/uno-community-toolkit.html?tabs=tabid-winui) on how to integrate the community toolkit.

This is an interesting error still, we probably should be try to error out more clearly when uno is being referenced in a WinUI app, something that definitely cannot work.

PR #10310, opened 2022-11-01
fix(reg): Remove invalid removal of uno.ui files when building with WinAppSDK

jeromelaban:
The current change removes all binaries from the net5.0-windows target, which is incorrect as it also removes uno.toolkit.ui (as it includes ElevatedView and VisibleBoundsPadding).

The proper fix is to specifically modify Uno.WindowsCommunityToolkit, which does not include a Windows target. This causes netstandard2.0 binaries to incorrectly show up in the build.

Related to https://github.com/unoplatform/uno/issues/10182 and https://github.com/unoplatform/uno/pull/10193

### DIFF
--- a/build/uno.winui.winappsdk.targets
+++ b/build/uno.winui.winappsdk.targets
@@ -1,23 +1,4 @@
 ﻿<?xml version="1.0" encoding="utf-8" ?>
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
-	<PropertyGroup>
-		<_UnoRemoveReferences_BeforeTargets>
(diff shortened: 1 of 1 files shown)


spouliot:
> The proper fix is to specifically modify Uno.WindowsCommunityToolkit, which does not include a Windows target.

Should there an issue open for this ? I did not see one in https://github.com/unoplatform/Uno.WindowsCommunityToolkit/issues

jeromelaban:
I forgot to mention it, yes it's fixed there: https://github.com/unoplatform/Uno.WindowsCommunityToolkit/pull/184

spouliot (approved):

PR #10318, opened 2022-11-03
build: Enable CA1304 (Specify CultureInfo)

Youssef1313:

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI.RemoteControl/RemoteControlClient.cs
+++ b/src/Uno.UI.RemoteControl/RemoteControlClient.cs
@@ -2,6 +2,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Net;
@@ -76,7 +77,7 @@ namespace Uno.UI.RemoteControl
 			try
 			{
 #if __WASM__
-				var isHttps = WebAssemblyRuntime.InvokeJS("window.location.protocol == 'https:'").ToLower() == "true";
+				var isHttps = WebAssemblyRuntime.InvokeJS("window.location.protocol == 'https:'").ToLower(CultureInfo.InvariantCulture) == "true";
 #else
 				const bool isHttps = false;
 #endif
--- a/src/Uno.UI.RemoteControl/RemoteControlClient.cs
+++ b/src/Uno.UI.RemoteControl/RemoteControlClient.cs
@@ -77,7 +77,7 @@ namespace Uno.UI.RemoteControl
 			try
 			{
 #if __WASM__
-				var isHttps = WebAssemblyRuntime.InvokeJS("window.location.protocol == 'https:'").ToLower(CultureInfo.InvariantCulture) == "true";
+				var isHttps = WebAssemblyRuntime.InvokeJS("window.location.protocol == 'https:'").Equals("true", StringComparison.OrdinalIgnoreCase);
 #else
 				const bool isHttps = false;
 #endif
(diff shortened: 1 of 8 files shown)


src/Uno.UI.RemoteControl/RemoteControlClient.cs
```diff
@@ -76,7 +77,7 @@ private async Task StartConnection()
 			try
 			{
 #if __WASM__
-				var isHttps = WebAssemblyRuntime.InvokeJS("window.location.protocol == 'https:'").ToLower() == "true";
+				var isHttps = WebAssemblyRuntime.InvokeJS("window.location.protocol == 'https:'").Equals("true", StringComparison.OrdinalIgnoreCase);
```
Youssef1313, line 80: I could do `.ToLower(CultureInfo.InvariantCulture)`, but this allocates a new string for no reason. String comparison should be better here.

jeromelaban (approved):

src/Uno.UI/UI/Xaml/Controls/PersonPicture/InitialsGenerator.cs
```diff
@@ -88,7 +89,7 @@ internal static string InitialsFromContactObject(Contact contact)
 					string result = GetFirstFullCharacter(firstName);
 					result += GetFirstFullCharacter(lastName);
 
-					return result.ToUpper();
+					return result.ToUpper(CultureInfo.InvariantCulture);
```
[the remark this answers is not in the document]
Youssef1313:
```suggestion
					return result.ToUpperInvariant();
```

src/Uno.UI.RemoteControl/RemoteControlClient.cs
Youssef1313: `bool.TryParse` seems significantly slower.

image

src/Uno.UI/UI/Xaml/Controls/PersonPicture/InitialsGenerator.cs
```diff
@@ -128,7 +129,7 @@ public static string InitialsFromDisplayName(string contactDisplayName)
 
 					string result = GetFirstFullCharacter(firstWord);
 
-					return result.ToUpper();
+					return result.ToUpper(CultureInfo.InvariantCulture);
```
Youssef1313, line 132:
```suggestion
					return result.ToUpperInvariant();
```

src/Uno.UI/UI/Xaml/Controls/PersonPicture/InitialsGenerator.cs
```diff
@@ -142,7 +143,7 @@ public static string InitialsFromDisplayName(string contactDisplayName)
 					string result = GetFirstFullCharacter(firstWord);
 					result += GetFirstFullCharacter(lastWord);
 
-					return result.ToUpper();
+					return result.ToUpper(CultureInfo.InvariantCulture);
```
Youssef1313, line 146:
```suggestion
					return result.ToUpperInvariant();
```

src/Uno.UI.RemoteControl/RemoteControlClient.cs
spouliot: Interesting :-) 

I don't recall where (or when) I read that... it might be on 7.0 or it just not as fast anyway since it has to return 2 values (but it might faster than previous net releases). Anyway I doubt it would make a difference when the code is about to invoke some javascript :)

jeromelaban: That particular code is only called once, in debug, for javascript invocation for which the interop will take signficantly more than either solution ;) In that case, (and martin may remember why he wrote it that way), using tryparse is more readable.

Issue #8597, opened 2022-04-22
Performance improvements


jeromelaban:
- [ ] [Android] FrameworkElement.TryCreateKeyRegistration is registering add_KeyPress (causing JNI interop)
- [ ] [Android] TextBoxView.SetCursorColor is creating a BlendModeColorFilter (Could it be pooled ?)
- [ ] [Android] ComboBox should initialize popup lazily
- [ ] [Android] TextBlock uses `View.ImportantForAccessibility` and should be using a default value 
- [ ] `AssignAndObserveBursh` may be able to avoid `RegisterPropertyChangedCallback` and use override instead?
- [ ] `FrameworkElement.PerformOnLoaded` uses `get_Parent` and be able to use a cached value?
- [ ] NumberBox should be able to use `OnKeyDown` override instead of registering to the `KeyDown` event?
- [ ] SelectoItem.UpdateCommonStates uses `DateTime.Now` causing expensive ICU work
- [ ] Remove all `RegisterLoadAction` uses
- [ ] `ConerRadiusFilter` should not be created eagerly during fluent resources initialization
- [ ] Generate the code code `DrawableHelper.Drawables` to avoid reflection on the `Resources` class
- [ ] `GetWinuiThemeResourceUri` should not be formatting a string
- [ ] `NativeInstanceHelper` should delegate its initialization (particularly the GetFolderPath invocation)
- [ ] [Android] `InnerBuildPaint` should be moved to java to reduce the interop calls

spouliot:
I can't edit the issue but `iOS get_Scale` should not be happening anymore since https://github.com/unoplatform/uno/pull/9114 was merged (and calls to `UIScreen.MainScreen` are cached).

spouliot:
I can't edit the issue but the bottleneck of `iOS get_Scale` should not be happening anymore since https://github.com/unoplatform/uno/pull/9114 was merged (and the use of `UIScreen.MainScreen` is cached).

PR #10428, opened 2022-11-14
perf: Reuse android java String instances in TextBlock

jeromelaban:
GitHub Issue (If applicable): #8597 

## PR Type

## What is the new behavior?

- Reuse Java.String instances for fast-path TextBlock rendering
- Fix `ApiInformation` initialization sequence for android
- Fix `MemoryManager` values update for android and `ArrayPool` automatic memory management for android

## PR Checklist

## Other information


src/Uno.Foundation/Metadata/ApiInformation.cs
```diff
@@ -29,7 +29,10 @@ internal static void RegisterAssembly(Assembly assembly)
 		{
 			lock (_assemblies)
 			{
-				_assemblies.Add(assembly);
+				if (!_assemblies.Contains(assembly))
+				{
+					_assemblies.Add(assembly);
+				}
```
Youssef1313, line 35: Will a `HashSet` perform better?

src/Uno.UI/UI/Xaml/Controls/TextBlock/JavaStringCache.android.cs
```diff
@@ -0,0 +1,189 @@
...
+		private static int _trimCallbackCreated;
+
+		private record KeyEntry(string Value, Java.Lang.String NativeValue)
+		{
+			public TimeSpan LastUse { get; set; } = _watch.Elapsed;
+		}
+
+		static JavaStringCache()
+		{
+			_automaticManagement = WinRTFeatureConfiguration.ArrayPool.EnableAutomaticMemoryManagement && _platformProvider.CanUseMemoryManager;
```
Youssef1313, line 55: It it possible to assign this inline and avoid the static constructor?

src/Uno.UI/UI/Xaml/Controls/TextBlock/JavaStringCache.android.cs
```diff
@@ -0,0 +1,189 @@
+﻿#nullable enable
```
Youssef1313, line 1: Could you rename to upper case `.Android.cs` so that the usage is consistent? and also the labeler workflow might be case-sensitive since we run it on Ubuntu.

src/Uno.UWP/System/MemoryManager.Android.cs
```diff
@@ -15,9 +15,14 @@ public partial class MemoryManager
...
 		private static global::System.Diagnostics.Stopwatch _updateWatch = global::System.Diagnostics.Stopwatch.StartNew();
-		private static long _lastUpdate = long.MinValue;
+		private static TimeSpan _lastUpdate = TimeSpan.FromSeconds(-_updateInterval.TotalSeconds);
 
-		private readonly static long _updateResolution = global::System.Diagnostics.Stopwatch.Frequency;
+		private readonly static TimeSpan _updateInterval = TimeSpan.FromSeconds(10);
+
+		static MemoryManager()
+		{
+			IsAvailable = true;
```
Youssef1313, line 24: Can this be assigned inline?

src/Uno.Foundation/Metadata/ApiInformation.cs
jeromelaban: We're adding 3 or 4 values, it's likely not going to change much.

src/Uno.UWP/System/MemoryManager.Android.cs
jeromelaban: It cannot, because the variable is defined on the generic side.

src/Uno.UI/UI/Xaml/Controls/TextBlock/JavaStringCache.android.cs
Youssef1313: > but the static constructor will still be here

Why does it need to stay? Wouldn't it be empty if we inlined this assignment?

Youssef1313, line 55: Actually, do we guarantee that this is executed *after* the user changed `WinRTFeatureConfiguration.ArrayPool.EnableAutomaticMemoryManagement`? I think `_automaticManagement` might be better as a property and execute the getter every time the value is read so that we can detect changes of `EnableAutomaticMemoryManagement`

jeromelaban: Visually yes, but in IL it won't change much, the .cctor will still be generated.

jeromelaban: There's no need to have this dynamically enabled at this point (managing the state of the change would not be useful). Most feature configuration behave this way, and this one is not different.

jeromelaban: Another thing here is the fact that ordering may become a problem, since this particular line depends on an inline initialized member, so I think it's best to keep it this way.

Youssef1313: @jeromelaban Yes it is still generated, but we lost `beforefieldinit` (https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1810 has some details).

The performance difference might be too small, so it doesn't matter that much.

PR #10533, opened 2022-11-22
`Popup` placement and positioning

### LINKED ISSUE #10459: Popup.PlacementTarget support
Popup.PlacementTarget support

## Why is this needed:
Autoplacement of the popup related to the UIElement

MartinZikmund:
GitHub Issue (If applicable): closes #10459

## PR Type

## PR Checklist

## Other information


MartinZikmund:
@FrozDark you are right, thanks, linked it

src/Uno.UI/UI/Xaml/Controls/Popup/Popup.cs
```diff
@@ -147,6 +139,66 @@ public LightDismissOverlayMode LightDismissOverlayMode
...
+
+		/// <summary>
+		/// Identifies the DesiredPlacement dependency property.
+		/// </summary>
+		public static DependencyProperty DesiredPlacementProperty { get; } =
+			DependencyProperty.Register(
+				nameof(DesiredPlacement),
+				typeof(PopupPlacementMode),
+				typeof(Popup),
+				new FrameworkPropertyMetadata(default(PopupPlacementMode)));
```
Youssef1313, line 178: I think it's more readable to be explicit enum value instead of `default(EnumType)`.

src/Uno.UI/UI/Xaml/Controls/Popup/PopupPanel.Placement.cs
```diff
@@ -0,0 +1,384 @@
...
+
+partial class PopupPanel
+{
+	// This is a base popup panel to calculate the placement near an anchor control.
+	// This class exists mostly to reuse the same logic between a Flyout and a ToolTip.
+	// This class should eventually be removed, and Uno should match WinUI's approach, where Flyout sets Popup.HorizontalOffset and VerticalOffset
+	// as well as Width and Height on FlyoutPresenter when it opens, and then allows the popup layouting to do its job.
+
+	private static readonly Dictionary<FlyoutBase.MajorPlacementMode, Memory<FlyoutBase.MajorPlacementMode>> PlacementsToTry =
+		new Dictionary<FlyoutBase.MajorPlacementMode, Memory<FlyoutBase.MajorPlacementMode>>()
```
Youssef1313, line 27:
```suggestion
		new()
```

src/Uno.UI/UI/Xaml/Controls/Popup/PopupPanel.cs
```diff
@@ -125,10 +125,15 @@ protected override Size ArrangeOverride(Size finalSize)
...
+				if (Popup.PlacementTarget is not null)
+				{
+					return PlacementArrangeOverride(Popup, finalSize);
+				}
+				
 				// Gets the location of the popup (or its Anchor) in the VisualTree, so we will align Top/Left with it
 				// Note: we do not prevent overflow of the popup on any side as UWP does not!
 				//		 (And actually it also lets the view appear out of the window ...)
-				var anchor = Popup.Anchor ?? Popup;
+				var anchor = Popup.PlacementTarget ?? Popup;
```
Youssef1313, line 136: It looks like `Popup.PlacementTarget` will always be null here?

```suggestion
				var anchor = Popup;
```

jeromelaban (commented):
Looking good, can we reuse flyout positioning tests for popups?

MartinZikmund:
> Looking good, can we reuse flyout positioning tests for popups?

Possibly yes! Will test against UWP

src/Uno.UI/UI/Xaml/Controls/Popup/PopupPanel.Placement.cs
```diff
@@ -0,0 +1,384 @@
...
+		var anchor = popup.PlacementTarget;
+		if (anchor == null)
+		{
+			return default;
+		}
+
+		return anchor.GetBoundsRectRelativeTo(this);
+	}
+
+	protected virtual Rect CalculatePopupPlacement(Popup popup, Size desiredSize, Size maxSize)
```
Youssef1313, line 115: It seems this is only called in this class, consider:

```suggestion
	private Rect CalculatePopupPlacement(Popup popup, Size desiredSize, Size maxSize)
```

src/Uno.UI/UI/Xaml/Controls/Popup/PopupPanel.Placement.cs
```diff
@@ -0,0 +1,384 @@
...
+	protected virtual bool FullPlacementRequested { get; }
+
+	internal virtual FlyoutBase Flyout => null;
+
+	private Size PlacementArrangeOverride(Popup popup, Size finalSize)
+	{
+		foreach (var child in Children)
+		{
+			var desiredSize = child.DesiredSize;
+			var maxSize = (child as FrameworkElement).GetMaxSize(); // UWP takes FlyoutPresenter's MaxHeight and MaxWidth into consideration, but ignores Height and Width
```
Youssef1313, line 84: In this line, if `child` is not `FrameworkElement`, the extension `GetMaxSize` will throw `NullReferenceException`

- If you can think of a scenario where `child` is not `FrameworkElement`, it should be tested and the code should be updated accordingly.
- Otherwise, a direct cast should be better than `as` in that if this fails, the exception will be clear about the type of `child`.

    ```suggestion
    			var maxSize = ((FrameworkElement)child).GetMaxSize(); // UWP takes FlyoutPresenter's MaxHeight and MaxWidth into consideration, but ignores Height and Width
    ```

src/Uno.UI/UI/Xaml/Controls/Popup/PopupPanel.Placement.cs
```diff
@@ -0,0 +1,384 @@
...
+#if __ANDROID__ || __IOS__
+		if (NativeAnchor != null)
+		{
+			return NativeAnchor.GetBoundsRectRelativeTo(this);
+		}
+#endif
+		var anchor = popup.PlacementTarget;
+		if (anchor == null)
+		{
+			return default;
```
Youssef1313, line 109:
```suggestion
			return null;
```

src/Uno.UI/UI/Xaml/Controls/Popup/PopupPanel.Placement.cs
```diff
@@ -0,0 +1,384 @@
...
+		}
+
+		return anchor.GetBoundsRectRelativeTo(this);
```
Youssef1313, line 112:
```suggestion
		return popup.PlacementTarget?.GetBoundsRectRelativeTo(this);
```

src/Uno.UI/UI/Xaml/Controls/Popup/PopupPanel.Placement.cs
```diff
@@ -0,0 +1,384 @@
...
+	}
+
+	protected virtual Rect CalculatePopupPlacement(Popup popup, Size desiredSize, Size maxSize)
+	{
+		if (!(GetAnchorRect(popup) is { } anchorRect))
```
Youssef1313, line 117:
```suggestion
		if (GetAnchorRect(popup) is not { } anchorRect)
```

src/Uno.UI/UI/Xaml/Controls/Popup/PopupPanel.Placement.cs
```diff
@@ -0,0 +1,384 @@
...
+	private Rect? GetAnchorRect(Popup popup)
+	{
+#if __ANDROID__ || __IOS__
+		if (NativeAnchor != null)
+		{
+			return NativeAnchor.GetBoundsRectRelativeTo(this);
+		}
+#endif
+		var anchor = popup.PlacementTarget;
+		if (anchor == null)
```
Youssef1313, line 107: `GetAnchorRect` is called by `CalculatePopupPlacement`, which is called by `PlacementArrangeOverride`, which is called under a `Popup.PlacementTarget is not null` condition. Consider passing the known non-null placement target from `PlacementArrangeOverride` down to `GetAnchorRect` and use it here. That way we can change the return type of `GetAnchorRect` from `Rect?` to `Rect`, and then avoid the null check done at the callsite of `GetAnchorRect`

src/Uno.UI/UI/Xaml/Controls/Popup/PopupPanel.Placement.cs
MartinZikmund: The method wasn't renamed before, but there is an override for native picker popups in iOS, so need to keep this overridable 👍

src/Uno.UI/UI/Xaml/Controls/Popup/PopupPanel.Placement.cs
```diff
@@ -0,0 +1,377 @@
...
+					var controlXPos = finalPosition.X;
+					fits = TestAndCenterAlignWithinLimits(
+						anchorRect.X,
+						anchorRect.Width,
+						desiredSize.Width,
+						visibleBounds.Left,
+						visibleBounds.Right,
+						justification,
+						ref controlXPos
+					);
```
Youssef1313, line 225: It looks like `TestAndCenterAlignWithinLimits` will always assign a new value to the ref parameter. I think it can be made `out` and the initial value should be unnecessary?

MartinZikmund: Yes indeed, it should have been `out`

src/Uno.UI/UI/Xaml/Controls/Popup/PopupPanel.cs
MartinZikmund: Now the logic is changed a bit to work around `DatePickerFlyout`, so had to keep this in the end.

spouliot (approved):

ajpinedam (approved):

PR #10569, opened 2022-11-25
feat: Annotate sensors

ebariche:


src/Uno.UWP/Devices/Sensors/LightSensor.cs
```diff
@@ -17,6 +17,9 @@ public partial class LightSensor
 
 		private readonly StartStopEventWrapper<TypedEventHandler<LightSensor, LightSensorReadingChangedEventArgs>> _readingChangedWrapper;
 
+#if __WASM__
+		[DynamicDependency(nameof(DispatchReading))]
```
jeromelaban, line 21: Can we put it on `TryCreateInstance` instead? It'll stay on the wasm file.

src/Uno.UWP/Devices/Sensors/Accelerometer.wasm.cs
```diff
@@ -82,7 +84,6 @@ private void DetachDeviceMotion()
 		/// <param name="y">Accelerometer Y</param>
 		/// <param name="z">Accelerometer Z</param>
 		/// <returns>0 - needed to bind method from WASM</returns>
-		[Preserve]
```
spouliot, line 85: ^ do you still care for the legacy Xamarin SDK ?
because the newer attribute(s) is/are not supported

ebariche: Actually https://github.com/unoplatform/uno/blob/master/src/Uno.Foundation/PreserveAttribute.cs is being used 🤔

spouliot: That's fine. The old linker accepts any `[Preserve]` attribute (by name, not by type) so it can work with .NET Standard (or older FX Xamarin had to support)

spouliot: Also I think the correct syntax should have been `[Preserve(Conditional = true)]`

This means the decorated member will be preserved if the type is used.
If the type is not marked then the member will not be preserved.

jeromelaban: The preserve attribute is interpreted by a custom step in the xamarin tooling only, so we can remove that one altogether for wasm targets.

ebariche: What this does, nobody knows, I'll delete it :)

spouliot: oh, missed that it was wasm only :)
yeah, delete it!

jeromelaban (approved):

PR #10671, opened 2022-12-06
perf: Geolocator JSImport bindings

ebariche:


src/Uno.UWP/Devices/Geolocation/Geolocator.wasm.cs
```diff
@@ -60,9 +74,13 @@ public static IAsyncOperation<GeolocationAccessStatus> RequestAccessAsync()
 
 					if (_pendingAccessRequests.Count == 1)
 					{
+#if NET7_0_OR_GREATER
+						NativeMethods.RequestAccess();
+#else
 						//there are no access requests currently waiting for resolution, we need to invoke the check in JS
 						var command = $"{JsType}.requestAccess()";
 						InvokeJS(command);
+#endif
```
jeromelaban, line 83: I wonder if we should be moving the whole block in to the NativeMethods.RequestAccess(), so everything's colocated and easier to reason with?

ebariche: This is only temporary, while the JSImportGenerator gets fixed. I want `NativeMethods` to be a nested class. The only change required then will be to remove this line :

```csharp
using NativeMethods = __Windows.Devices.Geolocation.Geolocator.NativeMethods;
```

I want refactoring to be as painless/bugfree as possible.

jeromelaban: Looks like it built properly :) 

The comment I made was about moving the whole interop code (including the one doing InvokeJS to the separate class, so all interop stays in one class, and the `NET7_0_OR_GREATER` is kept there.

jeromelaban (approved):

PR #10802, opened 2022-12-22
perf: WindowManager bindings II

ebariche:


src/Uno.UI/UI/Xaml/WindowManagerInterop.wasm.cs
```diff
@@ -1522,9 +1540,25 @@ internal enum HtmlPointerButtonUpdate
 #if NET7_0_OR_GREATER
 		internal static partial class NativeMethods
 		{
+			[JSImport("globalThis.Uno.UI.WindowManager.current.arrangeElementNativeFast")]
+			internal static partial void ArrangeElement(
+				double htmlId,
```
jeromelaban, line 1545: Why double?

ebariche: Typo :)

src/Uno.UI/ts/WindowManager.ts
```diff
@@ -733,6 +744,37 @@ namespace Uno.UI {
...
+			style.width = width === NaN ? "auto" : width + "px";
+			style.height = height === NaN ? "auto" : height + "px";
+
+			if (clip) {
+				style.clip = `rect(${clipTop}px, ${clipRight}px, ${clipBottom}px, ${clipLeft}px)`;
+			} else {
+				style.clip = "";
+			}
+
+			this.setAsArranged(element);
```
jeromelaban, line 775: It's probably best to avoid duplicating the code above, it's tricky enough :)

ebariche: Fair :)

src/Uno.UI/UI/Xaml/WindowManagerInterop.wasm.cs
```diff
@@ -1522,9 +1542,25 @@ internal enum HtmlPointerButtonUpdate
 #if NET7_0_OR_GREATER
```
Youssef1313, line 1542: I think it should be possible to get the benefits when not targeting .NET 7.

Two options I can think of:

#### Get the benefits when using .NET 7 SDK, regardless of the target framework.

The `JSImportGenerator` ships with the SDK, and I think it's quite unrelated to the TFM.
So we can define `JSImportAttribute` when `!NET7_0_OR_GREATER`

In this case, we'll need to only use the optimized code path when the .NET SDK is 7.0+.

#### Always get the benefits

This can be achieved by porting the `JSImportGenerator` implementation here.

jeromelaban: We discussed about this internally, and while it's possible, it's an additional level of complexity (and debugging) that will not be serving many developers (a Wasm head targeting `net6.0` with boostrapper 7.0.x where upgrading to `net7.0` is very easy).

jeromelaban (approved):

PR #10840, opened 2022-12-28
fix: Add missing ContentProperty attributes

### LINKED ISSUE #10839: Some controls are missing `ContentProperty` attributes
### Current behavior

_No response_

### Expected behavior

_No response_

_No response_

### Workaround

_No response_

### Works on UWP/WinUI

None

### Environment

_No response_

### NuGet package version(s)

_No response_

### Affected platforms

_No response_

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

MartinZikmund:
GitHub Issue (If applicable): closes #10839

## PR Type

## PR Checklist

## Other information


Youssef1313 (approved):
LGTM

Are there known bugs caused by the missing attribute?

src/Uno.UI/UI/Xaml/FrameworkTemplate.cs
```diff
@@ -24,7 +25,8 @@ namespace Windows.UI.Xaml
 	/// Defines a builder to be used in <see cref="FrameworkTemplate"/>
 	/// </summary>
 	public delegate View? FrameworkTemplateBuilder(object? owner);
-	
+
+	[ContentProperty(Name = "Template")]
```
jeromelaban, line 29: There's no `Template` property, where did you find this one?

jeromelaban: It may be an artifact of the way the `FrameworkTemplate` class is defined in C++. Thing is that this property is not visible in the C# projection, so I don't think we should be showing it either.

Youssef1313: @jeromelaban Why wouldn't this have effect on user-defined classes that inherit `FrameworkTemplate`?

---

> Thing is that this property is not visible in the C# projection

I'm not sure what you mean, I see it in VS, do you mean something else?

image

jeromelaban: I mean that the `Template` property does not exist, even if the `ContentProperty` refers to it.

Youssef1313: Yes, I see the property doesn't exist in `FrameworkTemplate`

What I mean is:

```csharp
// user-code
public class MyClass : FrameworkTemplate
{
    // define a property named "Template"
    // Even though MyClass doesn't have ContentProperty attribute, it is inherited from FrameworkTemplate.
}
```

Can't this happen?

jeromelaban: I see. This is a scenario that is not supported by user code, mainly because there's no public constructor available on WinUI.

Youssef1313: @jeromelaban Why does it have to be public? Isn't a protected constructor sufficient? https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.frameworktemplate.-ctor?view=windows-app-sdk-1.2

jeromelaban: Ah! This may have changed from UWP then :) But anyways, it does not change the fact that the property does not exist, but it does not seem to cause issues at this point, so we can keep it as-is.

Youssef1313 (commented):
Let's wait for https://github.com/unoplatform/uno/pull/11254

jeromelaban:
Let's also merge it in 5.0

PR #10893, opened 2023-01-05
fix(iOS): accelerometer crash when reporting changes

### LINKED ISSUE #10217: [Gallery][iOS]Accelerometer- App crashes when click on ' Start observing Accelerometer reading Changes' Button.
### Current behavior

https://user-images.githubusercontent.com/74563602/197278195-6826427d-580c-46a3-bee0-0a81d1f922af.MOV

### Expected behavior

Gallery app should not crash.

1. Open Uno Gallery APP
2. Tap Burger Menu on the left top right.
3. Tap "Accelerometer" from the "Non-UI Features" list.
4. Add some time duration in the 'Report Interval' section.
5. Click the 'Start observing Accelerometer reading Changes' button.
6. Update the time duration by clicking on the top or down arrow buttons on the Report Interval drop-down.
7. Notice that after a few seconds the Gallery app gets crashes.

### Workaround

_No response_

### Works on UWP/WinUI

_No response_

### Environment

_No response_

### NuGet package version(s)

_No response_

### Affected platforms

iOS

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

ajpinedam:
GitHub Issue (If applicable): closes #10217 

## PR Type

- Bugfix

## What is the current behaviour?
Application crash when reading Accelerometer values on iOS

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UWP/Devices/Sensors/Accelerometer.iOS.cs
+++ b/src/Uno.UWP/Devices/Sensors/Accelerometer.iOS.cs
@@ -37,7 +37,11 @@ namespace Windows.Devices.Sensors
 
 		private void StartReadingChanged()
 		{
-			_motionManager.StartAccelerometerUpdates(new NSOperationQueue(), AccelerometerDataReceived);
+			var nsoperationqueue = NSOperationQueue.CurrentQueue == null || NSOperationQueue.CurrentQueue == NSOperationQueue.MainQueue
+				? new NSOperationQueue()
+				: NSOperationQueue.CurrentQueue;
+
+			_motionManager.StartAccelerometerUpdates(nsoperationqueue, AccelerometerDataReceived);
 		}
 
 		private void StopReadingChanged()
--- a/src/Uno.UWP/Devices/Sensors/Accelerometer.cs
+++ b/src/Uno.UWP/Devices/Sensors/Accelerometer.cs
@@ -2,6 +2,7 @@
 using System;
 using System.Collections.Generic;
 using System.Text;
+using Windows.UI.Core;
 
 namespace Windows.Devices.Sensors
 {
@@ -97,7 +98,16 @@ namespace Windows.Devices.Sensors
 
 		private void OnReadingChanged(AccelerometerReading reading)
 		{
-			_readingChanged?.Invoke(this, new AccelerometerReadingChangedEventArgs(reading));
+			if (CoreDispatcher.Main.HasThreadAccess)
+			{
+				_readingChanged?.Invoke(this, new AccelerometerReadingChangedEventArgs(reading));
+			}
+			else
+			{
+				_ = CoreDispatcher.Main.RunAsync(CoreDispatcherPriority.Normal, () => {
+					_readingChanged?.Invoke(this, new AccelerometerReadingChangedEventArgs(reading));
+				});
+			}
 		}
 
 		internal void OnShaken(DateTimeOffset timestamp)


src/Uno.UWP/Devices/Sensors/Accelerometer.iOS.cs
```diff
@@ -37,7 +37,11 @@ private static Accelerometer TryCreateInstance()
 
 		private void StartReadingChanged()
 		{
-			_motionManager.StartAccelerometerUpdates(new NSOperationQueue(), AccelerometerDataReceived);
+			var nsoperationqueue = NSOperationQueue.CurrentQueue == null || NSOperationQueue.CurrentQueue == NSOperationQueue.MainQueue
+				? new NSOperationQueue()
+				: NSOperationQueue.CurrentQueue;
+
+			_motionManager.StartAccelerometerUpdates(nsoperationqueue, AccelerometerDataReceived);
```
Youssef1313, line 44: I'm not sure what is the effect of this change, but:

- this logic is duplicated in the following as well? https://github.com/unoplatform/uno/blob/6b686ea22265a11ff0dd359f2725fda81725aea9/src/Uno.UWP/Devices/Sensors/SimpleOrientationSensor.iOS.cs#L21
- should this be in https://github.com/unoplatform/uno/blob/6b686ea22265a11ff0dd359f2725fda81725aea9/src/Uno.UWP/Devices/Sensors/Barometer.iOS.cs#L31 https://github.com/unoplatform/uno/blob/6b686ea22265a11ff0dd359f2725fda81725aea9/src/Uno.UWP/Devices/Sensors/Gyrometer.iOS.cs#L32 https://github.com/unoplatform/uno/blob/6b686ea22265a11ff0dd359f2725fda81725aea9/src/Uno.UWP/Devices/Sensors/Magnetometer.iOS.cs#L36
- We should consider extracting this to a common helper method.

ajpinedam: @Youssef1313  thanks for the review.

The change looks to prevent creating a new Operation Queue instance (Reusing the current if any is available), and it ensures we don't schedule the Accelerometer operation in the Main Queue.

About the duplicate logic and helper function: I can include these suggestions in another PR I have to refactor some parts of the `Accelerometer`. 

I tend to avoid any refactoring work while fixing a bug, especially when the refactoring touch areas that are not related to where the fix was made.

Youssef1313 (approved):

MartinZikmund (changes requested):
I'm sorry I didn't notice this issue earlier, but it cannot be fixed within Uno, but rather within the Gallery sample. As per the UWP docs and the UWP samples, the event handler should reschedule to Dispatcher thread if needed - https://github.com/microsoft/Windows-universal-samples/blob/ad9a0c4def222aaf044e51f8ee0939911cb58471/Samples/Accelerometer/cs/Scenario1_DataEvents.xaml.cs#L86 . The reason is that for high-throughput scenarios where you want to just process accelerometer events as they come, you definitely prefer to be on the background thread. This way the Dispatcher would get swamped with potentially hundreds of events and the app would start freezing. So I suggest closing this PR and instead adding the `Dispatcher.RunAsync` to the Gallery app

Youssef1313 (commented):
@MartinZikmund Would that apply to https://github.com/unoplatform/uno/blob/7e3668befe70afaf9228e712e43531b0dd7720dc/src/Uno.UWP/Devices/Sensors/SimpleOrientationSensor.cs#L69-L79 ? [UWP Sample](https://github.com/microsoft/Windows-universal-samples/blob/ad9a0c4def222aaf044e51f8ee0939911cb58471/Samples/SimpleOrientationSensor/cs/Scenario1_DataEvents.xaml.cs#L140-L146)

This code is already in main, should it be changed?

MartinZikmund:
> 

Indeed, that is also incorrect - 

> @MartinZikmund Would that apply to
> 
> https://github.com/unoplatform/uno/blob/7e3668befe70afaf9228e712e43531b0dd7720dc/src/Uno.UWP/Devices/Sensors/SimpleOrientationSensor.cs#L69-L79
> 
> ? [UWP Sample](https://github.com/microsoft/Windows-universal-samples/blob/ad9a0c4def222aaf044e51f8ee0939911cb58471/Samples/SimpleOrientationSensor/cs/Scenario1_DataEvents.xaml.cs#L140-L146)
> This code is already in main, should it be changed?

Yes indeed, that should be changed to match the other sensors - e.g. trigger on background thread and then the user has to apply `Dispatcher` themselves.

agneszitte-nventive (approved):

MartinZikmund:
As discussed, we need to make sure that the behavior matches WinUI here, so I close this PR for now

PR #11012, opened 2023-01-19
ci: Add hints to `OptimizeImplicitlyTriggeredBuild` in cross targeting override

jeromelaban:
@mrlacey some hints for performance

### DIFF
--- a/doc/articles/uno-development/building-uno-ui.md
+++ b/doc/articles/uno-development/building-uno-ui.md
@@ -50,6 +50,7 @@ The step by step process is:
 1. To confirm that everything works:
    - For iOS/Android/macOS you can right-click on the `Uno.UI` project in the Solution Explorer and 'Build'.
    - For WebAssembly and Skia you can right-click on the `Uno.UI.Runtime.WebAssembly` or `Uno.UI.Runtime.Skia.[Gtk|Wpf]` project in the Solution Explorer and 'Build'.
+1. Optionally adjust additional parameters in `crosstargeting_override.props`, such as `UnoDisableNetAnalyzers` which can improve the build time during debugging sessions.
 
 Once you've built successfully, for the next steps, [consult the guide here](debugging-uno-ui.md) for debugging Uno.UI.
 
--- a/src/crosstargeting_override.props.sample
+++ b/src/crosstargeting_override.props.sample
@@ -54,9 +54,9 @@
 
     <!--
     Uncomment this line if you're iterating quickly in your builds. This 
-    will disable most of the analyzers, and you'll need to comment that property
+    will disable most of the C# analyzers, and you'll need to comment that property
     before making your pull request.
     -->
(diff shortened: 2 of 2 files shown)


doc/articles/uno-development/building-uno-ui.md
```diff
@@ -50,6 +50,7 @@ The step by step process is:
 1. To confirm that everything works:
    - For iOS/Android/macOS you can right-click on the `Uno.UI` project in the Solution Explorer and 'Build'.
    - For WebAssembly and Skia you can right-click on the `Uno.UI.Runtime.WebAssembly` or `Uno.UI.Runtime.Skia.[Gtk|Wpf]` project in the Solution Explorer and 'Build'.
+1. Optionally adjust additional parameters in `crosstargeting_override.props`, such as `UnoDisableNetAnalyzers` which can improve the build time during debugging sessions.
```
Youssef1313, line 53: VS disables all analyzers when for builds initiated by "Debug tests" or "Run with debugger" already. So analyzers don't run unless an explicit build (Ctrl+Shift+B or right click project then Build).

So the comment is quite misleading IMO.

Youssef1313: @jeromelaban @mrlacey Can I know more about the performance problem being referred to here? A binlog will also help to properly report the specific slow analyzer to Microsoft.

Youssef1313: I revised the logic responsible for that in Roslyn, and build optimization when building for debugging is only done when `TreatWarningsAsErrors` isn't true.

To force build optimization, `OptimizeImplicitlyTriggeredBuild` should be set to true.

I think the better fix is to set `OptimizeImplicitlyTriggeredBuild` to true in `Directory.Build.props` (just below `TreatWarningsAsErrors`). This way, all developers will have the same experience, and they get it "by default" (without having to modify `crosstargeting_override.props`).

Youssef1313: After adding `<OptimizeImplicitlyTriggeredBuild>true</OptimizeImplicitlyTriggeredBuild>`, you should see the following message if the build is initiated by "Debug test" or similar:

> Skipping analyzers to speed up the build. You can execute 'Build' or 'Rebuild' command to run analyzers.

jeromelaban: Interesting, then maybe we should set the property instead :) Thanks for the insights!

Youssef1313: @jeromelaban I'd set it unconditionally in Directory.Build.props

jeromelaban: Let's have it in the override file for now, and if it does not cause local build issues (or CI issues) we can revisit.

spouliot (approved):
❤️

PR #11033, opened 2023-01-23
perf: WindowManager bindings V

ebariche:


jeromelaban (approved):
🚀

src/Uno.UI/ts/WindowManager.ts
```diff
@@ -30,7 +30,10 @@ namespace Uno.UI {
 		private static readonly unoUnarrangedClassName = "uno-unarranged";
 		private static readonly unoCollapsedClassName = "uno-visibility-collapsed";
 
-		private static _cctor = (() => {
+		private static _cctor = (async () => {
+
+			await ExportManager.initialize();
```
jeromelaban, line 35: Won't this create a race condition with the next methods to execute? The internals are async in the runtime, but it might fail in the future, I'd think.

PR #11074, opened 2023-01-27
perf: Improve xaml generator performance

Youssef1313:
Previously, every run of the generator was collecting a dictionary of all symbols in compilation references. This is very expensive, and the generator is run on every change in the IDE. This is both slow and very memory consuming which can in turn cause high GC pause times.

## PR Type

## PR Checklist

## Other information


src/SolutionTemplate/Uno.ProjectTemplates.Dotnet/content/unoapp-uwp-net6/.vsconfig
```diff
@@ -3,92 +3,36 @@
   "components": [
     "Microsoft.VisualStudio.Component.CoreEditor",
     "Microsoft.VisualStudio.Workload.CoreEditor",
-    "Microsoft.NetCore.Component.Runtime.3.1",
```
Youssef1313, line 6: I didn't change anything in `.vsconfig`. These were done automatically when I built from VS.

Youssef1313:
@jeromelaban This is ready for review.

jeromelaban:
This is unfortunately a large breaking change. There are some legacy codebases which rely on the use of fuzzy matching.

This may be a change that we can include in the next major breaking change pass.

That being said, the change about the `Binding` attribute may be included sooner, as it is not breaking.

Youssef1313:
I tried to reduce the impact of the breaking change by adding additional namespaces that are searched by default. We can also introduce an MSBuild property `UnoXamlDefaultNamespaces` to make it easier for developers to overcome the breaking change. So, if someone largely broke, they can add `<UnoXamlDefaultNamespace>CommaSeparatedNamespaces</UnoXamlDefaultNamespace>.

Alternatively, we can introduce a `UnoEnableXamlFuzzyMatching` property and document that it affects performance badly and that it will be removed in the next major release. This property will restore the fuzzy matching.

Youssef1313:
Also worth noting that code relying on this wouldn't compile on Windows.

Youssef1313:
Worth noting that the GC pause time I found was too large, and these allocations are the major contributor.

image

jeromelaban:
I don't mind if we introduce this breaking change in the next major, it's a good one. Introducing the msbuild property is a good idea, we'll make it enabled by default for 4.x, and mention that in the docs that devs will need to test it when moving to the later version.

Youssef1313:
> I don't mind if we introduce this breaking change in the next major, it's a good one. Introducing the msbuild property is a good idea, we'll make it enabled by default for 4.x, and mention that in the docs that devs will need to test it when moving to the later version.

Great! I'll add this property and default it to true.

Youssef1313:
@jeromelaban Can you take another look?

jeromelaban:
This flag also raises an additional blind spot where we suggest in the documentation to use namespaces to exclude platform specific code, but without providing a way to specify a namespace for lookups. We'll need to find a proper way to handle this, following what was discussed a while back in https://github.com/unoplatform/uno/issues/789

Youssef1313:
> This flag also raises an additional blind spot where we suggest in the documentation to use namespaces to exclude platform specific code, but without providing a way to specify a namespace for lookups. We'll need to find a proper way to handle this, following what was discussed a while back in #789

I guess if one will use something from the default namespaces, then `xmlns:android="http://uno.ui/android"` should work. Otherwise, `xmlns:android="using:NamespaceHere"` will work (already supported I think? - but not documented)

And finally, if platform-specific code is needed with different namespaces, probably do one of the two things:

- Use `IsTypePresent` and pass android-only symbol.
- Use multiple namespaces, `<IncludeXamlNamespaces Include="android_MyNamespace" />` where in XAML this will be `xmlns:android_MyNamespace="using:MyNamespace`, etc for other namespaces.

@jeromelaban Does that sound good? If so, I can document that when removing the new flag in breaking changes branch.

agneszitte-nventive (approved):

jeromelaban:
This is not enough, no, because `android` is meant to be just for one namespace, which means that developers will need to include more `android` like namespaces as defined here: 

https://github.com/unoplatform/uno/blob/8ddbd9e87c815d69403bc0136e8adbb4f3a2d84c/src/SourceGenerators/Uno.UI.SourceGenerators/Content/Uno.UI.SourceGenerators.props#L161-L190

This is not particular user-friendly, and we'll need to come up with another way to exclude namespaces based on a specific format, maybe in the Uri or something else, which was partially the discussion in https://github.com/unoplatform/uno/issues/789.

Youssef1313:
Two different errors in CI. One that I cannot reproduce locally, and the other is because of incomplete `IncludeXamlNamespaces`/`ExcludeXamlNamespaces` in the outer build (not sure if that is the correct terminology, but I'm referring to the build that has all TFMs set but no specific TFM to be build, which at later point is dispatched to inner builds). The task runs on the build before it is dispatched so we don't (can't) have all xaml namespaces set. I'll update to rely on `Ignorable` as you suggested before @jeromelaban 

Then will see if the error that I cannot reproduce locally will happen.

Youssef1313:
@jeromelaban Finally, this is green now 🎉

src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/ListView/HorizontalListView_Templates.xaml
```diff
@@ -9,7 +9,7 @@
 	xmlns:ios="http://uno.ui/ios"
 	xmlns:uConverters="using:Uno.UI.Samples.Converters"
 	xmlns:win="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
-	xmlns:android="http://uno.ui/android"
+	xmlns:android="http://uno.ui/android#using:Uno.UI.Controls.Legacy"
```
jeromelaban, line 12: We need to document this in https://github.com/unoplatform/uno/blob/master/doc/articles/platform-specific-xaml.md

src/Directory.Build.props
```diff
@@ -58,6 +58,8 @@
 
     <DisableImplicitNamespaceImports>true</DisableImplicitNamespaceImports>
 
+    <UnoEnableXamlFuzzyMatching>true</UnoEnableXamlFuzzyMatching>
```
Youssef1313, line 61:
```suggestion
    <UnoEnableXamlFuzzyMatching>false</UnoEnableXamlFuzzyMatching>
```

src/SourceGenerators/Uno.UI.SourceGenerators/BindableTypeProviders/BindableTypeProvidersGenerationTask.cs
```diff
@@ -85,9 +84,8 @@ internal void Generate(GeneratorExecutionContext context)
 						_assemblyName = context.GetMSBuildPropertyValue("AssemblyName");
 
 						_defaultNamespace = context.GetMSBuildPropertyValue("RootNamespace");
-						_namedSymbolsLookup = context.Compilation.GetSymbolNameLookup();
 
-						_bindableAttributeSymbol = FindBindableAttributes();
+						_bindableAttributeSymbol = context.Compilation.GetTypeByMetadataName(XamlConstants.Types.BindableAttribute);
```
jeromelaban, line 88: We should also be searching for `System.ComponentModel.BindableAttribute`, as a compatibility measure as there are some non-winui viewmodels that use it.

PR #11093, opened 2023-01-29
chore: reorder Colors.Parse code to avoid dupe null checks

spouliot:

## PR Type

Input string is checked twice for null.

There was also a nullability warning in the method.

> /xxxsrc/Uno.UWP/UI/Colors.cs(57,39): warning CS8605: Unboxing a possibly null value. [/xxx/benchmarks/Colors_FromInteger/Colors_FromInteger.csproj]

## What is the new behavior?

Single null check.
Fixed warning.

Performance change is unmeasurable, but it’s less code :-)

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UWP/UI/Colors.cs
+++ b/src/Uno.UWP/UI/Colors.cs
@@ -14,7 +14,7 @@ namespace Windows.UI
 #endif
 	static partial class Colors
 	{
-		private static Dictionary<string, Color> _colorMap = new Dictionary<string, Color>(StringComparer.OrdinalIgnoreCase);
+		private static readonly Dictionary<string, Color?> _colorMap = new(StringComparer.OrdinalIgnoreCase);
 
 		public static Color FromARGB(byte a, byte r, byte g, byte b) => new(a, r, g, b);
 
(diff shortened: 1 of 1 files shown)


src/Uno.UWP/UI/Colors.cs
```diff
@@ -27,23 +27,21 @@ static partial class Colors
 		/// <returns></returns>
 		public static Color Parse(string colorCode)
 		{
-			if (colorCode?.StartsWith("#", StringComparison.Ordinal) ?? false)
+			if (!string.IsNullOrWhiteSpace(colorCode))
```
Youssef1313, line 30: Passing a whitespace here is very uncommon and will eventually fail too.

```suggestion
			if (!string.IsNullOrEmpty(colorCode))
```

src/Uno.UWP/UI/Colors.cs
```diff
@@ -27,23 +27,21 @@ static partial class Colors
...
+				else
+				{
+					if (!_colorMap.TryGetValue(colorCode, out var color))
 					{
 						var property = typeof(Colors).GetProperty(colorCode);
 
 						if (property != null)
 						{
-							_colorMap[colorCode] = color = (Color)property.GetValue(null);
+							_colorMap[colorCode] = color = (Color)property.GetValue(null)!;
```
Youssef1313, line 44: Can we get improved performance by hard-coding the dictionary and avoid reflection?

We can add a reflection-based unit test to ensure the dictionary is correct.

src/Uno.UWP/UI/Colors.cs
spouliot: Right now whitespace(s) would be treated the same as null or empty.

What input did you have in mind ?

src/Uno.UWP/UI/Colors.cs
spouliot: Yes, but that's for a different PR :)

The current code seems to have a potential issue with the linker, e.g. a property could be removed (not directly referenced) and that would change how a color is accepted or not. That's tricky to debug... so it might be better to refactor the properties code too.

src/Uno.UWP/UI/Colors.cs
Youssef1313: Changing `IsNullOrWhiteSpace` to `IsNullOrEmpty` will only change the exception message produced for a whitespace input. I think we don't care right?

The common case for this call is the success case, so we want that as fast as possible, and `IsNullOrEmpty` should be faster than `IsNullOrWhiteSpace`, and we will still throw an exception.

src/Uno.UWP/UI/Colors.cs
jeromelaban: Good catch. Indeed the current code will definitely fail if the linker removes members, and definitely, using an hardcoded table will help performance for bound string lookups.

jeromelaban (approved):

src/Uno.UWP/UI/Colors.cs
spouliot: I moved the existing

> if (!string.IsNullOrWhiteSpace(colorCode))

code thinking this matched the Windows behaviour :-)

I'll double-check the behaviour on Windows and make sure Uno behave the same.

spouliot: The method is not exposed in UWP nor WinUI so I don't think the behaviour with whitespace would matter.

PR #11158, opened 2023-02-02
fix(templates): Fix Server project

### LINKED ISSUE #11154: [Template] New app from unoapp template includes template processing code
### Current behavior

In Program.cs in Server project, there is still processing code (eg <!--#if (WebAssembly) )
image

This prevents the solution from being built after being created

### Expected behavior

Solution should build successfully and there shouldn't be any template code left in the source code

Clean up existing templates
`dotnet new uninstall uno.projecttemplates.dotnet`
Install 4.7.37 templates 
`dotnet new install uno.projecttemplates.dotnet::4.7.37` 
Create new app
`dotnet new unoapp -o MyApp`

### Workaround

_No response_

### Works on UWP/WinUI

None

### Environment

_No response_

### NuGet package version(s)

_No response_

### Affected platforms

_No response_

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

Youssef1313:
I have a fix ready, but wanted to first make sure this will fail in CI.

GitHub Issue (If applicable): closes #11154

## PR Type

## PR Checklist

## Other information


jeromelaban (approved):

src/SolutionTemplate/UnoSolutionTemplate.WinUI.netcore/Server/Program.cs
```diff
@@ -20,10 +20,10 @@ public static void Main(string[] args)
 
 			app.UseAuthorization();
 
-$if$($UseWebAssembly$ == True)
+#if USE_WASM
 			app.UseUnoFrameworkFiles();
 			app.MapFallbackToFile("index.html");
-$endif$
+#endif
```
jeromelaban, line 26: Why is this an issue?

Youssef1313: It gets replaced with `<!-- -->` which is invalid.

jeromelaban: I see. There's a C# syntax for it, but it's an annoying one, that will do for now.

agneszitte-nventive (approved):

ajpinedam (approved):

src/SolutionTemplate/UnoSolutionTemplate.WinUI.netcore/Server/Program.cs
```diff
@@ -20,10 +20,10 @@ public static void Main(string[] args)
 
 			app.UseAuthorization();
 
-$if$($UseWebAssembly$ == True)
+#if USE_WASM
```
dansiegel, line 23: This change does not appear to be in line with the original issue. The problem is that there was an invalid template condition. where `<!--#if` should have just been `#if`. This PR seems to be introducing a completely unnecessary MSBuild compiler constant which has nothing to do with the issue.
image
This would make no sense to have here, and since the unoapp template actually removes the Server project if WASM isn't included I would assume here that the correct action is actually to remove the condition entirely. @jeromelaban what do you think?

dansiegel (changes requested):

src/SolutionTemplate/UnoSolutionTemplate.WinUI.netcore/Server/Program.cs
Youssef1313: I agree the new MSBuild property is unnecessary, it just seemed to be quite easier to do to fix the issue. I can try with another approach if @jeromelaban thinks it's worth it.

jeromelaban: The origin of the condition is that it's possible to create the server project standalone, and it's best if the project can compile by itself (for template build tests, for instance).

The other approach needs modifying the template converter to template `<!-- #if` to be a proper #if, with `server.cs` not having `//-:cnd:noEmit` at the top so it can be interpreted by `dotnet new` properly.

Youssef1313: @jeromelaban To make sure I understand, is there something I need to do?

jeromelaban: It would be best not to use the `DefineConstants` and use an actual `dotnet new` variable to conditionally include the code.

dansiegel: @jeromelaban just looking at the template.json and the behavior doesn't match for the Server project... you have the `hosted` symbol, but in the modifiers, the Server is excluded when `WebAssembly` is false. There is no modifier for when `hosted` is false.

image

src/SolutionTemplate/UnoSolutionTemplate.WinUI.netcore/Server/Program.cs
```diff
@@ -20,10 +20,10 @@ public static void Main(string[] args)
 
 			app.UseAuthorization();
 
-$if$($UseWebAssembly$ == True)
+#if (UseWebAssembly)
```
dansiegel, line 23: This needs to use the actual dotnet new symbol which is `WebAssembly`
```suggestion
//+:cnd:noEmit
#if (WebAssembly)
```

dansiegel (changes requested):

src/SolutionTemplate/UnoSolutionTemplate.WinUI.netcore/Server/Program.cs
Youssef1313: @dansiegel This didn't work (at least in VSIX)

image

It might be working from `dotnet new` but I haven't tested.

Youssef1313:
@dansiegel Can you take another look?

src/SolutionTemplate/UnoSolutionTemplate.WinUI.netcore/Server/Program.cs
dansiegel: yeah dotnet templates and VSIX have different ways they work so I'm not surprised

Youssef1313:
@jeromelaban Can we get this merged?

jeromelaban:
Yes, this is the correct change, I took a look at the generated template. Thanks!!

PR #11219, opened 2023-02-07
build: Enable shared compilation

Youssef1313:
The removal of `DependenciesInitializer` was necessary to enable shared compilation. Those `Assembly.Load` calls are not needed and generally they can cause hard to investigate issues, regardless of whether we enable shared compilation or not, we shouldn't touch anything related to assembly loading in analyzers/generators.

(Thanks @jaredpar!)

## PR Type

## PR Checklist

## Other information


jeromelaban:
Looks like `content of 'Uno.UI.SourceGenerators\Uno.UI.SourceGenerators.DependencyObject.DependencyObjectGenerator\OuterClass.Inner_d21c9c80c794838c736d3a5351b7c79f.cs' did not match. Diff shown with expected as baseline:`

is failing.

Youssef1313:
@jeromelaban This should hopefully be ready to merge.

src/SourceGenerators/SourceGeneratorHelpers/Helpers/DependenciesInitializer.cs
```diff
@@ -1,53 +0,0 @@
...
-using System.Linq;
-using System.Reflection;
-
-#if NETFRAMEWORK
-using Uno.SourceGeneration;
-#endif
-
-namespace Uno.UI.SourceGenerators
-{
-	internal class DependenciesInitializer
```
jaredpar, line 12: If anyone knows the history of why this was initially added I'd be interested to know. Not high priority or anything. Just interested in the history.

jaredpar (approved):

src/SourceGenerators/SourceGeneratorHelpers/Helpers/DependenciesInitializer.cs
jeromelaban: Thanks for the help you're providing @jaredpar!

I can shed some light on that. Uno was originally using [Uno.SourceGeneration](https://github.com/unoplatform/Uno.SourceGeneration), which was running generators in a separate process way before Roslyn had its own implementation of source generators.  

At the time, the Uno generators needed external dependencies such as `Mono.Cecil` and a few other external helper assemblies. Those assemblies needed to be loaded at the right time and place, and the `net461` or netcore loaders were not looking for the libraries properly. The solution was to use the generators' assembly to load all the sibling assemblies explicitly with this helper.

Moving forward to Roslyn's driver (still a few years back), the same assembly location issue happened with the dependent assemblies, so this custom loader was kept in place. Recently though, we've reduced the number of assemblies, but we still have a few needed but it seems that now the loader is properly reading the files so we can remove that helper.

That being said @Youssef1313, we likely still need the compatibility with `Uno.SourceGeneration` when building generators in the `net461` configuration, and removing this helper is likely to make the compilation fail (if it's not, it's either our dotnet template tests are not using it `Uno.SourceGeneration`, or that it really is not needed anymore). Let's add an explicit template test which sets `UnoUIUseRoslynSourceGenerators` to `false`, next to this one:
https://github.com/unoplatform/uno/blob/master/build/test-scripts/run-template-tests.ps1#L102-L105

Youssef1313: @jeromelaban The whole code is wrapped in `#if NETSTANDARD2_0`, so I don't think Uno-based generation is affected at all. Anything I'm missing?

jeromelaban: You're right, I misread this and probably remember an older piece of the implementation, my bad! Then this _may_ mean that Roslyn somehow changed behavior along the way so that it's not needed anymore to load dependent assemblies.

Either way, let's merge this as-is then, thanks @Youssef1313.

jeromelaban (approved):

PR #11284, opened 2023-02-12
perf: remove reflection usage when parsing color names

spouliot:
This makes the code to behave more correctly when linking, since the properties could not be present at runtime - making the color non-parsable.

Also, as a nice side effect, using a `switch` (i.e. code) removes the `.cctor` from type.

## PR Type

- Bugfix

If the linker is enabled then some colors might not be parsed correctly since they use reflection.

## What is the new behavior?

Reflection is not used. It's better for performance and linking won't have any potential effect at runtime.

## PR Checklist

## Other information


Youssef1313 (commented):
Can we have a reflection-based unit test to make sure all members in `Colors` are handled in `Colors.Parse` method?

src/Uno.UWP/UI/Colors.cs
```diff
@@ -35,21 +33,153 @@ public static Color Parse(string colorCode)
...
-						if (property != null)
-						{
-							_colorMap[colorCode] = color = (Color)property.GetValue(null)!;
-						}
-						else
-						{
-							throw new InvalidOperationException($"The color {colorCode} is unknown");
-						}
-					}
+						nameof(Transparent) => 0x00FFFFFF,
```
Youssef1313, line 38: The original behavior was that `colorCode` is case insensitive. Now it is case sensitive. Is that intended?

spouliot: good catch! I'll remote the `nameof` and compare with a lowercase string

spouliot:
> Can we have a reflection-based unit test to make sure all members in `Colors` are handled in `Colors.Parse` method?

`Parse` does not exists on WinUI, but I guess we can have Uno specific tests.

src/Uno.UWP/UI/Colors.cs
```diff
@@ -35,21 +33,153 @@ public static Color Parse(string colorCode)
...
+						nameof(DarkGray) => 0xFFA9A9A9,
+						nameof(DarkGreen) => 0xFF006400,
+						nameof(DarkKhaki) => 0xFFBDB76B,
+						nameof(DarkMagenta) => 0xFF8B008B,
+						nameof(DarkOliveGreen) => 0xFF556B2F,
+						nameof(DarkOrange) => 0xFFFF8C00,
+						nameof(DarkOrchid) => 0xFF9932CC,
+						nameof(DarkRed) => 0xFF8B0000,
+						nameof(DarkSalmon) => 0xFFE9967A,
+						nameof(DarkSeaGreen) => 0xFF8FBC8B,
```
Xiaoy312, line 72: [DarkSeaGreen](https://learn.microsoft.com/en-us/uwp/api/windows.ui.colors.darkseagreen?view=winrt-22621) is #FF8FBC8F.

you might want to codegen those lines with uwp just to be safe:
```cs
var asd = string.Join("\n", typeof(Colors)
	.GetProperties(System.Reflection.BindingFlags.Public | System.Reflection.BindingFlags.Static)
	.Where(x => x.PropertyType == typeof(Color))
	.OrderByDescending(x => x.Name == "Transparent")
	.Select(x => $"nameof({x.Name}) => {x.GetValue(null).ToString().Replace("#", "0x")},")
	//.Select(x => $"\"{x.Name.ToLower()}\" => {x.GetValue(null).ToString().Replace("#", "0x")},")
);
```

src/Uno.UWP/UI/Colors.cs
```diff
@@ -35,21 +33,153 @@ public static Color Parse(string colorCode)
...
+						nameof(Wheat) => 0xFFF5DEB3,
+						nameof(White) => 0xFFFFFFFF,
+						nameof(WhiteSmoke) => 0xFFF5F5F5,
+						nameof(Yellow) => 0xFFFFFF00,
+						nameof(YellowGreen) => 0xFF9ACD32,
+						_ => throw new InvalidOperationException($"The color {colorCode} is unknown")
+					};
 
-					return color;
+					return new Color(color);
```
Xiaoy312, line 182: couldnt you just return the static properties, like?:
```cs
return colorCode switch
{
	"transparent" => Colors.Transparent,
};
```
not sure if this is more or less performant~?

src/Uno.UWP/UI/Colors.cs
spouliot: Weird, they were generated from similar code - but from Uno, not UWP itself. I'll double check that, thanks!

src/Uno.UWP/UI/Colors.cs
spouliot: Performance wise it should be identical (inlined).

However calling the `.ctor` makes it possible for the linker to remove the properties (when they are not used elsewhere in the app). Using them her would ensure they are not removable in any case.

src/Uno.UWP/UI/Colors.cs
spouliot: https://github.com/unoplatform/uno/blob/89bda710bc575be97edf6b87b601b5ed070061c4/src/Uno.UWP/UI/Colors.cs#L456

ok, that's an Uno bug :)
I'll fix it inside this PR

src/Uno.UWP/UI/Colors.cs
```diff
@@ -33,149 +33,149 @@ public static Color Parse(string colorCode)
 				}
 				else
 				{
-					uint color = colorCode switch
+					uint color = colorCode.ToLowerInvariant() switch
```
Youssef1313, line 36: I wonder which is faster/better, this or having a series of `if`/`else if`/.. with `StringComparison.OrdinalIgnoreCase`?

src/Uno.UWP/UI/Colors.cs
Youssef1313: I'm really curios how you caught this @Xiaoy312

src/Uno.UWP/UI/Colors.cs
spouliot: In general the C# compiler is _smart_ enough to rewrite `switch` statements using `if/else` or using a collection.

In this case there might be a small hit since a new string will likely be allocated (unless it's already all lowercased). However that would not affect (much) the `Colors` properties since they are cached.

I'll try to benchmark this once all other feedback are fixed on the PR.

Youssef1313: @spouliot Note: newer version of the compiler (which we are not yet using) has introduced a better code gen for switch over strings, see https://github.com/dotnet/roslyn/pull/66081

Consider benchmarking against this version of the compiler (we will move to this version of the compiler eventually)

src/Uno.UWP/UI/Colors.cs
Xiaoy312: @Youssef1313 2am insanity. just picked a few randomly to sanity check.

jeromelaban (approved):

jeromelaban:
@spouliot Looks like there's a small difference:
image

spouliot:
@jeromelaban oops, I only fixed the parsing method, not the property. At least it proved the tests are complete enough :D

https://github.com/unoplatform/uno/blob/6251d058355ef0be856cfd140a302314e0fe16f5/src/Uno.UWP/UI/Colors.cs#L456

PR #11295, opened 2023-02-13
fix(templates): Fix XAML conditional for Wasm and Skia

### LINKED ISSUE #11282: The new templates broke conditional xaml
### Current behavior

The new templates use a class library, and when it's compiled for net7.0, we don't know if it's skia or wasm and conditional xaml for those don't work. However, conditional xaml should work on Android/iOS/macOS since they can be determined from `TargetFramework` (I have only tried Wasm and Skia though)

Conditional XAML is supposed to work per the value of `XamarinProjectType` which is set here:

https://github.com/unoplatform/uno/blob/a6eda7c5d1ab8d2b34cfc443fbd54e9432b8cf16/src/SourceGenerators/Uno.UI.SourceGenerators/Content/Uno.UI.SourceGenerators.props#L71-L72

The XAML file is processed in the classlib which doesn't know `UnoRuntimeIdentifier` (in fact, if the classlib is built for net7.0, the output is cached for both wasm and skia since both are net7.0)

### Expected behavior

Conditional xaml should work the same as before with the shared project.

_No response_

### Workaround

_No response_

### Works on UWP/WinUI

None

### Environment

_No response_

### NuGet package version(s)

_No response_

### Affected platforms

_No response_

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

I have tried a little trick but it **didn't work**, though I thought it's worth mentioning.

In `Skia.Gtk.csproj`, reference the classlib as follows:

(Note: The same will be done for all other projects - but I was testing only for Skia to get something working first)

```xml
  <ItemGroup>
    <ProjectReference Include="..\ProjectName\ProjectName.csproj">
      <SetTargetFramework>TargetFrameworks=$(TargetFramework)</SetTargetFramework>
      <SkipGetTargetFrameworkProperties>true</SkipGetTargetFrameworkProperties>
      <Properties>XamarinProjectType=$(XamarinProjectType)</Properties>
    </ProjectReference>
  </ItemGroup>
```

Then in the classlib, to avoid a `GetTargetPath` error from MSBuild, I removed all the `TargetFrameworks` properties and replaced it with just `<TargetFramework>net7.0</TargetFramework>`.

**This unfortunately didn't work.**

Youssef1313:
GitHub Issue (If applicable): closes #11282

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/SolutionTemplate/Uno.ProjectTemplates.Dotnet/Uno.ProjectTemplates.Dotnet.csproj
+++ b/src/SolutionTemplate/Uno.ProjectTemplates.Dotnet/Uno.ProjectTemplates.Dotnet.csproj
@@ -339,7 +339,7 @@
 							s = Regex.Replace(s, pattern, @"$1<!--#if ($2)" + Environment.NewLine + "$1", RegexOptions.Multiline);
 
 							s = s
-								.Replace("$endif$", "\n#endif -->");
+								.Replace("$endif$", "\r\n#endif -->");
 								;
 						}
 
--- a/src/SolutionTemplate/UnoSolutionTemplate.WinUI.netcore/Server/UnoQuickStart.Server.csproj
+++ b/src/SolutionTemplate/UnoSolutionTemplate.WinUI.netcore/Server/UnoQuickStart.Server.csproj
@@ -1,8 +1,10 @@
 <Project Sdk="Microsoft.NET.Sdk.Web">
 
 	<PropertyGroup>
-		<TargetFramework>$basetargetframework$</TargetFramework>
+		<TargetFramework>$basetargetframework$-wasm1.0</TargetFramework>
 		<Nullable>enable</Nullable>
+		<TargetPlatformSupported>true</TargetPlatformSupported>
+		<TargetPlatformVersion>1.0</TargetPlatformVersion>
 	</PropertyGroup>
 
 	$if$($UseWebAssembly$ == True)<ItemGroup>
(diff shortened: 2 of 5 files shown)


jeromelaban:
This is an interesting idea, but this is likely to cause issues at multiple levels, namely for nuget, but also for msbuild itself:

```
11:26:04:877	3>c:\vs\odd\MSBuild\Current\Bin\amd64\Microsoft.Common.CurrentVersion.targets(1830,5): error : Project '..\..\ClassLibrary2\ClassLibrary2.csproj' targets 'net7.0-wasm1.0;net7.0-skia1.0'. It cannot be referenced by a project that targets '.NETCoreApp,Version=v7.0'.
```

Youssef1313:
@jeromelaban I'm not sure how I can reproduce this error. Any steps to repro?

Youssef1313:
> This is an interesting idea, but this is likely to cause issues at multiple levels, namely for nuget, but also for msbuild itself:
> 
> ```
> 11:26:04:877	3>c:\vs\odd\MSBuild\Current\Bin\amd64\Microsoft.Common.CurrentVersion.targets(1830,5): error : Project '..\..\ClassLibrary2\ClassLibrary2.csproj' targets 'net7.0-wasm1.0;net7.0-skia1.0'. It cannot be referenced by a project that targets '.NETCoreApp,Version=v7.0'.
> ```

This error seems to indicate that `ClassLibrary2.csproj` doesn't have a `net7.0` target, while the template is "supposed" to have it. Am I missing something?

jeromelaban:
That was indeed the problem. This still needs a lot more testing with the many parts of the pipeline.

Youssef1313:
@jeromelaban Anything I need to do on my end?

jeromelaban:
Yes, we need to work more on the downstream effects of such a template change, particularly when dealing with nuget packages. Most of the changes you made should also be in uno's targets, not in the project templates.

jeromelaban:
This will be addressed implicitly by the inclusion of `net8.0-browser`.

Issue #11308, opened 2023-02-14
Dropdown elements font render incorrect of DataGridComboBoxColumn in DataGrid


zhongjohnson:
### Current behavior

The elements in the DataGridComboBoxColumn renders incorrect

![image](https://user-images.githubusercontent.com/80380574/218659024-d98e99d5-b5d1-4e01-bc8b-9ba655e06997.png)

### Expected behavior

Expecting it can use the same font of DataGrid and render correctly.

### How to reproduce it (as minimally and precisely as possible)

        <controls:DataGrid x:Name="dataGrid"
                           FontFamily="{x:Bind ViewModel.PreferredFontFamily}">
            <controls:DataGrid.Columns>
                <controls:DataGridTextColumn  />
                <controls:DataGridComboBoxColumn 
                                                 FontFamily="{x:Bind ViewModel.PreferredFontFamily}" />
                <controls:DataGridTextColumn  />
            </controls:DataGrid.Columns>
        </controls:DataGrid>

### Works on UWP/WinUI

None

### NuGet package version(s)

Uno.WinUI.Skia.Wpf 4.6.19

### Affected platforms

Skia (WPF)

### IDE

Visual Studio 2022

MartinZikmund:
May be duplicate of #10801

Youssef1313:
@MartinZikmund While fixing #10801 will make this issue goes away, we really should make sure if it's the right root cause.

I'd prefer that we add the test @jeromelaban tried to add in #11330 (commit https://github.com/unoplatform/uno/pull/11330/commits/ce3b30a2095d6467e861588693ad69dff1a2091b), and make sure it's passing on Windows, then make sure it's also passing in Uno.

MartinZikmund:
Duplicate of https://github.com/unoplatform/uno-private/issues/1365

PR #11330, opened 2023-02-15
fix(ComboBox): Adjust inherited properties propagation to item containers

jeromelaban:
GitHub Issue (If applicable): closes https://github.com/unoplatform/uno/issues/11308 

## PR Type

- Bugfix

## What is the new behavior?

Inherited properties are now propagated from the ComboBox instance to ComboBoxItem.

## PR Checklist

## Other information


src/Uno.UI/UI/Xaml/Controls/ComboBox/ComboBox.cs
```diff
@@ -232,6 +232,64 @@ private void OnHeaderTemplateChanged(DataTemplate oldHeaderTemplate, DataTemplat
...
+
+			base.ClearContainerForItemOverride(element, item);
+		}
+
+		/// <summary>
+		/// Explicitly propagate inherited properties values to the item containers, as they are not direct visual children of the ComboBox.
+		/// </summary>
+		private void SynchronizeContainerInheritedProperties(DependencyObject element, object item)
+		{
+			if (element is ContentControl cc && !IsItemItsOwnContainer(item))
```
Youssef1313, line 254: Is the `element is ContentControl cc` check needed?

Do we have a test case where `element` is not a `ContentControl`?

jeromelaban: If it's not a ContentControl (so not an `ComboBoxItem`), `IsItemItsOwnContainer` will be false. We already have test cases for non-ComboBoxItem items.

src/Uno.UI/UI/Xaml/Controls/ComboBox/ComboBox.cs
```diff
@@ -232,6 +232,64 @@ private void OnHeaderTemplateChanged(DataTemplate oldHeaderTemplate, DataTemplat
...
+			{
+				foreach (var comboBoxProperty in DependencyProperty.GetPropertiesForType(this.GetType()))
```
Youssef1313, line 256: This will get properties for user-defined control that inherits `ComboBox`, does this behavior match WinUI? Consider a case where the user-defined control shadows a property of `ComboBox`, should the user-defined property be set or the original property?

src/Uno.UI/UI/Xaml/Controls/ComboBox/ComboBox.cs
```diff
@@ -232,6 +232,64 @@ private void OnHeaderTemplateChanged(DataTemplate oldHeaderTemplate, DataTemplat
...
+		/// <summary>
+		/// Explicitly clear inherited properties values to the item containers, as they are not direct visual children of the ComboBox.
+		/// </summary>
+		private void ClearContainerInheritedProperties(DependencyObject element, object item)
+		{
+			if (element is ContentControl cc && !IsItemItsOwnContainer(item))
+			{
+				foreach (var comboBoxProperty in DependencyProperty.GetPropertiesForType(this.GetType()))
+				{
+					if (comboBoxProperty.GetMetadata(this.GetType()) is FrameworkPropertyMetadata fpm
```
Youssef1313, line 281: Same questions here as well

src/Uno.UI/UI/Xaml/Controls/ComboBox/ComboBox.cs
jeromelaban: User-defined properties cannot be inherited, and the instance is not built by the user. The only situation is that if the user changes the item container style, the CombBox value will take precedence. I don't know if that's what WinUI does at this point, but it's worth looking at it.

Currently the PR is in draft for reasons like these, as there may be side effects in tests.

src/Uno.UI/UI/Xaml/Controls/ComboBox/ComboBox.cs
Youssef1313: @jeromelaban Doesn't that mean `element is ContentControl` is unnecessary check (ie, `!IsItemItsOwnContainer(item)` is sufficient alone)?

src/Uno.UI/UI/Xaml/Controls/ComboBox/ComboBox.cs
```diff
@@ -34,7 +34,7 @@
 
 namespace Windows.UI.Xaml.Controls
 {
-	public partial class ComboBox : Selector
+	public partial class ComboBox : Selector, IThemeChangeAware
```
Youssef1313, line 37: In [this doc](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/themeresource-internals.md):

> Some non-FrameworkElement types may also need to react to the theme change. This is supported by the IThemeChangeAware internal interface

That seems to imply that this interface is only useful for non-FrameworkElement types, but `ComboBox` is FrameworkElement`

Is the doc outdated? or should things work in a different way?

jeromelaban: The doc is probably outdated. It used for all framework elements to react to theme changes, if needed.

src/Uno.UI/UI/Xaml/Controls/ComboBox/ComboBox.cs
```diff
@@ -232,6 +232,64 @@ private void OnHeaderTemplateChanged(DataTemplate oldHeaderTemplate, DataTemplat
...
+
+		protected override void ClearContainerForItemOverride(DependencyObject element, object item)
+		{
+			ClearContainerInheritedProperties(element, item);
+
+			base.ClearContainerForItemOverride(element, item);
+		}
+
+		/// <summary>
+		/// Explicitly propagate inherited properties values to the item containers, as they are not direct visual children of the ComboBox.
```
agneszitte-nventive, line 250:
```suggestion
		/// Explicitly propagate inherited property values to the item containers, as they are not direct visual children of the ComboBox.
```

src/Uno.UI/UI/Xaml/Controls/ComboBox/ComboBox.cs
```diff
@@ -232,6 +232,64 @@ private void OnHeaderTemplateChanged(DataTemplate oldHeaderTemplate, DataTemplat
...
+							// the ContentControl default style.
+							element.SetValue(containerProperty, GetValue(comboBoxProperty), DependencyPropertyValuePrecedences.Local);
+						}
+					}
+				}
+			}
+		}
+
+		/// <summary>
+		/// Explicitly clear inherited properties values to the item containers, as they are not direct visual children of the ComboBox.
```
agneszitte-nventive, line 273:
```suggestion
		/// Explicitly clear inherited property values to the item containers, as they are not direct visual children of the ComboBox.
```

PR #11416, opened 2023-02-22
refactor: Remove unneeded IsEnabledChangedEventArgs

Youssef1313:

## PR Type

## PR Checklist

## Other information


jeromelaban (approved):
It's a change from the original source, but since it's all internal it won't change much. Thanks!

MartinZikmund:
@Youssef1313 If possible, I would prefer to keep the type. Even though it is technically unnecessary, it will make porting from WinUI "harder" as you will always have to do that extra sanity check to ensure, that the method you are overriding is actually the one, that you intend to and not something "slightly different" which is often the case in WinUI C++... I know that this work was for nothing then, but it may really be helpful, to keep comparing WinUI sources to Uno sources as simple as possible (also so that we can later potentially automate it more easily)

Youssef1313:
So should we close this? @jeromelaban @MartinZikmund

src/Uno.UI/UI/Xaml/FrameworkElement.cs
```diff
@@ -740,10 +735,7 @@ private void OnIsEnabledChanged(DependencyPropertyChangedEventArgs args)
 		{
 			UpdateHitTest();
 
-			_isEnabledChangedEventArgs ??= new IsEnabledChangedEventArgs();
-			_isEnabledChangedEventArgs.SourceEvent = args;
-
-			OnIsEnabledChanged(_isEnabledChangedEventArgs);
```
Youssef1313, line 746: This part of the code is what led me to this refactoring. It was confusing why we set the needed info to a static `_isEnabledChangedEventArgs` to pass it when we can directly pass it and make the code more readable.

MartinZikmund: The field is not useful. I would suggest changing the `IsEnabledChangedEventArgs` to be a read-only `struct` and just create it and pass it forward. This way we will keep the type for porting parity and avoid potential weird issues coming from it being a static field

jeromelaban: Agreed. Let's keep the code unmodified for source parity, and change the type to a struct.

jeromelaban: @Youssef1313 if changing to a struct does not give much benefit, let's close this PR.

Youssef1313:
Let's close out until it's proven this is problematic to keep source parity for now.

PR #11471, opened 2023-02-26
perf: Faster symbol lookup

Youssef1313:
Two changes made:

1. Move getting symbols to XamlCodeGeneration instead of XamlFileGenerator. So we now have less calls.
2. Using `Lazy<T>` instead of `AsLockedMemoized`. Here is a benchmark for `AsLockedMemoized` vs `Lazy<T>`

```csharp
using System;
using System.Collections.Concurrent;
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Running;
using Microsoft.CodeAnalysis;
using Microsoft.CodeAnalysis.CSharp;

public class Program
{
    public static void Main(string[] args)
    {
        _ = BenchmarkRunner.Run(typeof(Program).Assembly);
    }
}

public static class Funcs
{
    public static Func<TParam, TResult> Create<TParam, TResult>(Func<TParam, TResult> function) {
        return function;
    }
}

[MemoryDiagnoser]
public class Benchs
{
    private Func<string, ITypeSymbol> _findTypeByFullName;
    private Compilation _compilation;
    private Lazy<INamedTypeSymbol> _lazy;


    private ITypeSymbol SourceFindTypeByFullName(string fullName)
    {
        var symbol = _compilation.GetTypeByMetadataName(fullName);

        return symbol;
    }

    [GlobalSetup]
    public void Setup()
    {
        _compilation = CSharpCompilation.Create(null, new[] { SyntaxFactory.ParseSyntaxTree("namespace A.B.C { public class D { } }") });
        _findTypeByFullName = Funcs.Create<string, ITypeSymbol>(SourceFindTypeByFullName).AsLockedMemoized();
        _lazy = new Lazy<INamedTypeSymbol>(() => _compilation.GetTypeByMetadataName("A.B.C.D"));
    }

    [Benchmark(Baseline = true)]
    public void Baseline()
    {
        _findTypeByFullName("A.B.C.D");
    }

    [Benchmark]
    public void UsingLazy()
    {
        _ = _lazy.Value;
    }
}

public static class Ext
{
    public static Func<TKey, TResult> AsLockedMemoized<TKey, TResult>(this Func<TKey, TResult> func)
    {
        var values = new ConcurrentDictionary<TKey, TResult>();

        return (key) => values.GetOrAdd(key, func);
    }
}

```

Result on my machine:

|    Method |      Mean |     Error |    StdDev | Ratio | Gen 0 | Gen 1 | Gen 2 | Allocated |
|---------- |----------:|----------:|----------:|------:|------:|------:|------:|----------:|
|  Baseline | 23.743 ns | 0.0698 ns | 0.0583 ns |  1.00 |     - |     - |     - |         - |
| UsingLazy |  3.529 ns | 0.0187 ns | 0.0156 ns |  0.15 |     - |     - |     - |         - |

image

### DIFF
--- a/src/SourceGenerators/SourceGeneratorHelpers/Helpers/RoslynMetadataHelper.cs
+++ b/src/SourceGenerators/SourceGeneratorHelpers/Helpers/RoslynMetadataHelper.cs
@@ -1,9 +1,8 @@
-﻿using System;
-using System.Collections.Generic;
+﻿#nullable enable
+
+using System;
 using System.Linq;
 using Microsoft.CodeAnalysis;
-using Microsoft.CodeAnalysis.CSharp;
-using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Uno.Extensions;
 
 #if NETFRAMEWORK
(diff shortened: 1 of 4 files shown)


Youssef1313:
@jeromelaban @spouliot I was expecting an improvement, but I'm surprised it is *so large* that I'm suspecting I might be benchmarking incorrectly? Can you review the benchmark please?

My benchmark doesn't put into account that some symbols might not be accessed, which will actually make `Lazy<T>` approach even MORE FASTER!

jeromelaban (approved):

spouliot (approved):
@Youssef1313 I'm not familiar with `AsLockedMemoized` and will have to look it up...

However you can expect `Lazy<T>` to have been quite optimized by MS (both code and JIT) since it's often used in perf critical code.

Outside the benchmark it's not clear how many `Lazy` are not being used (evaluated) inside the source generation. That's where any gains will be...

Youssef1313:
> I'm not familiar with `AsLockedMemoized` and will have to look it up...

Basically ConcurrentDictionary under the hoods

Youssef1313:
https://github.com/unoplatform/uno/blob/1a45d03fbefd938a0e880edcca99d895cf651586/src/Uno.Foundation/Uno.Core.Extensions/Uno.Core.Extensions/FuncMemoizeExtensions.cs#L231-L267

The relevant conditional code of the above snippet is that:

```csharp
 			var values = new ConcurrentDictionary<TKey, TResult>(); 
  
 			return (key) => values.GetOrAdd(key, func); 
```

jeromelaban:
The original code from this part did not have Lazy available (as far as I remember), and if can use Lazy now, way to go :)

PR #11499, opened 2023-02-27
chore: Simplify Xaml generator state

Youssef1313:
@jeromelaban Tests are passing locally for me. Let's see if the CI gets green.

If this change breaks something, we should make sure it's tested, and I'd love to know more about what exactly will break and why (I don't have much details on how the generator handles HotReload).

**Review with hidden whitespace for an easier diff**


jeromelaban:
We cannot remove these blocks, they are needed for hot reload, and are not tested locally yet (the generator framework re-creates the generators instances during unit tests).

PR #11560, opened 2023-03-04
fix: Properly check Lamp support

morning4coffe-dev:
GitHub Issue (If applicable): fixes https://github.com/unoplatform/Uno.Gallery/issues/646

## PR Type

Because the Lamp is not implemented on other platforms than Android and iOS, the GetDefaultAsync() does not return any value on those other platforms.

## What is the new behavior?
The GetDefaultAsync() now returns null on WASM and Skia.

## PR Checklist

## Other information


Youssef1313 (commented):
I'm not sure why this is the right change.

We have always been using the generated code for unimplemented APIs.

Consuming code should have `#if __ANDROID__ || __IOS__` before using the API.

Youssef1313 (commented):
A `try` with `catch (NotImplementedException)` in the consuming code can be an option as well to be more future proof than a `#if` for specific platforms.

MartinZikmund:
@Youssef1313 in this specific case I think it is the correct approach - we have it for the other sensors as well - the method's purpose is to return the default instance or `null` if the sensor is not supported, which is the case here. It makes the consuming code cleaner and when in the future a target platform implements the sensor, the code will just start working there by updating Uno version, which is very nice 👍

MartinZikmund (approved):

Youssef1313:
Shpuld we at least still have not implemented attribute then?

MartinZikmund:
@Youssef1313 I think that would be more confusing - it is implemented - it just always gives you the information that this target does not have this sensor. E.g. none of the sensors are available on macOS because well... Apple - so it is logical that you always get `null` and you can rely on that fact that the remaining code that works with the instance where available is going to work without any platform specific code and `#ifs`. It is good because when porting existing app to Uno requires zero changes in this scenario. If it was marked as not implemented, user would be tempted to use `#if` because usually unimplemented methods throw in Uno. As mentioned, we have the same approach in the other sensors as well, so I think it is fine in this case 👍 .

Youssef1313 (approved):

jeromelaban (approved):

Youssef1313:
@morning4coffe-dev There are conflicts here

build/PackageDiffIgnore.xml
```diff
@@ -9384,7 +9384,11 @@
...
-			<Member fullName="System.Void Windows.Devices.Lights.Lamp..ctor()" reason="Not present in Windows" />
+			<Member fullName="System.Void Windows.UI.Xaml.Controls.TextBox.OnInputScopeChanged(Windows.UI.Xaml.DependencyPropertyChangedEventArgs e)" reason="Changed parameter type and made private." />
+			<Member fullName="System.Void Windows.UI.Xaml.Controls.TextBox.OnIsSpellCheckEnabledChanged(Windows.UI.Xaml.DependencyPropertyChangedEventArgs e)" reason="Changed parameter type and made private." />
+			<Member fullName="System.Void Windows.UI.Xaml.Controls.TextBox.OnIsTextPredictionEnabledChanged(Windows.UI.Xaml.DependencyPropertyChangedEventArgs e)" reason="Changed parameter type and made private." />
+			<Member fullName="System.Void Windows.UI.Xaml.Controls.TextBox.OnTextAlignmentChanged(Windows.UI.Xaml.DependencyPropertyChangedEventArgs e)" reason="Changed parameter type and made private." />
+      <Member fullName="System.Void Windows.Devices.Lights.Lamp..ctor()" reason="Not present in Windows" />
```
jeromelaban, line 9391: This looks incorrect or a merge duplicate. I think this is coming from your PR @MartinZikmund ?

Youssef1313: @jeromelaban It is coming from my PR, but I think you are looking only at the merge commit.

MartinZikmund: Yes, that's correct, that is only the merge commit

PR #11592, opened 2023-03-07
perf: More efficient FastConvert

Youssef1313:

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/DataBinding/BindingPropertyHelper.FastConvert.cs
+++ b/src/Uno.UI/DataBinding/BindingPropertyHelper.FastConvert.cs
@@ -364,54 +364,185 @@ namespace Uno.UI.DataBinding
 		{
 			if (outputType != typeof(Windows.UI.Xaml.Input.InputScope)) return false;
 
-			var nameValue = input.ToLowerInvariant().Trim() switch
+			var trimmed = input.AsSpan().Trim();
+			Windows.UI.Xaml.Input.InputScopeNameValue nameValue;
+			if (trimmed.Equals("0", StringComparison.OrdinalIgnoreCase) || trimmed.Equals("default", StringComparison.OrdinalIgnoreCase))
 			{
-				"0" or "default" => Windows.UI.Xaml.Input.InputScopeNameValue.Default,
-				"1" or "url" => Windows.UI.Xaml.Input.InputScopeNameValue.Url,
-				"5" or "emailsmtpaddress" => Windows.UI.Xaml.Input.InputScopeNameValue.EmailSmtpAddress,
-				"7" or "personalfullname" => Windows.UI.Xaml.Input.InputScopeNameValue.PersonalFullName,
-				"20" or "currencyamountandsymbol" => Windows.UI.Xaml.Input.InputScopeNameValue.CurrencyAmountAndSymbol,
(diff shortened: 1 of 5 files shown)


src/Uno.UI/UI/Xaml/Media/Animation/RepeatBehavior.cs
```diff
@@ -54,7 +54,7 @@ internal bool ShouldRepeat(TimeSpan elapsed, int count)
 		public override int GetHashCode()
 			=> Count.GetHashCode()
 				^ Duration.GetHashCode()
-				^ Type.GetHashCode();
+				^ ((int)Type).GetHashCode();
```
Youssef1313, line 57: Avoids boxing `Type`

src/Uno.UWP/UI/Input/PointerPointProperties.cs
```diff
@@ -153,7 +153,7 @@ public override string ToString()
 
 			// Misc
 			builder.Append('(');
-			builder.Append(PointerUpdateKind);
+			builder.Append(PointerUpdateKind.ToString());
```
Youssef1313, line 156: Avoid boxing `PointerUpdateKind`

src/Uno.UI/DataBinding/BindingPropertyHelper.FastConvert.cs
```diff
@@ -364,54 +364,185 @@ private static bool FastStringToInputScope(Type outputType, string input, ref ob
...
+			}
+			else if (trimmed.Equals("68", StringComparison.OrdinalIgnoreCase) || trimmed.Equals("chatwithoutemoji", StringComparison.OrdinalIgnoreCase))
+			{
+				nameValue = Windows.UI.Xaml.Input.InputScopeNameValue.ChatWithoutEmoji;
+			}
+			else
+			{
+				throw new InvalidOperationException($"Failed to create a '{outputType.FullName}' from the text '{input}'.");
+			}
+
```
spouliot, line 545: Do you have a benchmark that shows the gains ? like the old version, this version and a mix of using `Span` and `switch` ?

Nice things about `switch` is that 

* it's much easier to read the code

* the compiler (current and future) can optimize the code - having a benchmark allow anyone to compare numbers again to see if the extra code is still worth it.

jeromelaban: I guess that it may only reduce allocations, and only if tolowerinvariant reallocates even if the string is the same.

spouliot: Yes, it should reduce allocations (and `ToLowerInvariant` should not allocate anything on a lowercased string).

However a smart compiler [1] should be able to create a collision-free, case-insensitive hash code (and without any allocation) and use this for a faster switch. Right now the last `if ` conditions will be much slower than the first ones. That would not matter for smaller sets of values but this one is rather large.

[1] Roslyn might not be there yet and/or we could do this manually but probably not worth the dev time ;-)

Anyway the `switch` code is more optimization friendly (today or in the future) than a series of `if ... else if ...` and having a _canary_ benchmark could alert us if the situation change :-)

Youssef1313: Okay, let me close this out.

PR #11632, opened 2023-03-10
fix(iOS): Monochromatic Image

ajpinedam:
GitHub Issue (If applicable): closes https://github.com/unoplatform/nventive-private/issues/444

## PR Type

- Bugfix

Bitmap icon with some images are being cutoff when displayed

## PR Checklist

## Other information


Youssef1313 (commented):
Does this have the same semantics? I think it is more understandable that way.

```csharp
var index = 0;
for (int y = 0; y < height; y++)
{
	for (int x = 0; x < width; x++)
	{
		outputData[index++] = foreground.R;
		outputData[index++] = foreground.G;
		outputData[index++] = foreground.B;
		var sourceAlpha = rawData[index];
		outputData[index++] = sourceAlpha;
	}
}
```

jeromelaban:
@ajpinedam you should also be able to create a runtime tests similar to this one: 

https://github.com/unoplatform/uno/blob/3810a87698c4753133dba150d7f4ca8cbde75453/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_Border.cs#L241, make sure to add the IsTypePresent check for platforms that don't support it.

ajpinedam:
> Does this have the same semantics? I think it is more understandable that way.
> 
> ```cs
> var index = 0;
> for (int y = 0; y < height; y++)
> {
> 	for (int x = 0; x < width; x++)
> 	{
> 		outputData[index++] = foreground.R;
> 		outputData[index++] = foreground.G;
> 		outputData[index++] = foreground.B;
> 		var sourceAlpha = rawData[index];
> 		outputData[index++] = sourceAlpha;
> 	}
> }
> ```

@Youssef1313  Thanks for the suggestion.

I don't believe this is clearer since it will be very position dependent. In the current way, it is clear that we are accessing.

`index + n` position where **N** can be a value from 1 to 3, we can see that very easily, but I feel that when using `outputData[index++] = foreground.G;` anyone reading the code will need actually to think/calculate what value index was in the previous line, making it more error-prone.

spouliot (approved):

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_Icon.cs
```diff
@@ -0,0 +1,98 @@
...
+			TestServices.WindowHelper.WindowContent = parent;
+			await WindowHelper.WaitForLoaded(parent);
+
+			var snapshot = await TakeScreenshot(parent);
+
+			var sample = SUT.GetRelativeCoords(parent);
+			var centerX = sample.X + sample.Width / 2;
+			var centerY = sample.Y + sample.Height / 2;
+
+			ImageAssert.HasPixels(
```
ajpinedam, line 59: @jeromelaban this assert is failing on the first run. When run locally it could fail on the first run and running a second time the test passes.

ajpinedam: It seems the screenshot is being taken before the image finish loading. How can I wait till the image is loaded to take the screenshot?

1E05A83F-595D-4122-ABA5-431DC366A765

jeromelaban: You should be waiting for the image by either the ImageLoaded event, or by waiting (using `.WaitFor(() => )`) for the DesiredSize property to change to an acceptable value.

ajpinedam: @jeromelaban  The `BitmapIcon` has no public event we can use for this. We could instead test the Image, Set the Monochrome, and we should get the same behaviour and be able to test it, but then we would need to skip the test for `NET461`

jeromelaban: That's fine for net46, it's not working in this context.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_Image.cs
```diff
@@ -301,6 +301,76 @@ public async Task When_Image_Source_Nullify()
 				ExpectedPixels.At(centerX, centerY).Named("center without image").Pixel(Colors.White));
 		}
 
+		[TestMethod]
+		[RunsOnUIThread]
+#if NET461 || __MACOS__
```
ajpinedam, line 306: @jeromelaban  this doesn't seem to be working. How can I  skip this test for NET461 since the `MonochromaticColor` Property for the image is not available there.

Youssef1313: @ajpinedam The CI failure is not for NET461. It's for uap10.0.18362.

You need to wrap the whole test method with `#if !WINDOWS_UWP`

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_Image.cs
```diff
@@ -301,6 +301,76 @@ public async Task When_Image_Source_Nullify()
...
+					.Named("bottom-left")
+					.WithPixelTolerance(1, 1)
+					.Pixel(red),
+				ExpectedPixels
+					.At(sample.X + sample.Width - 1f, sample.Y + sample.Height - 1f)
+					.Named("bottom-right")
+					.WithPixelTolerance(1, 1)
+					.Pixel(red));
+		}
+
```
Youssef1313, line 373:
```suggestion
#endif

```

jeromelaban (approved):

PR #11679, opened 2023-03-14
perf: TextBlock performance improvements

ebariche:


src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.cs
```diff
@@ -36,6 +36,7 @@ public partial class TextBlock : DependencyObject
 		private string _inlinesText; // Text derived from the content of Inlines
 		private readonly SerialDisposable _foregroundChanged = new SerialDisposable();
 
+		private Run _reusableRun;
```
Youssef1313, line 39: I'm not sure if this is problematic.

```csharp
var textBlock = new TextBlock();
textBlock.Text = "Old value";
var run = (Run)textBlock.Inlines[0];
textBlock.Text = "New value";
// run.Text will be "New value", which is not expected.
```

jeromelaban: It's expected to fail, in the same way the previous implementation was. We need to change this behavior, but it will not be in this PR :)

jeromelaban (approved):

jeromelaban:
@mergifyio backport release/stable/4.8

PR #11688, opened 2023-03-14
fix(templates): Specify `AndroidManifest` correctly so that design-time builds don't fail

### LINKED ISSUE #11641: Central Package Management does not work with Mobile project
### Current behavior

When Central package management is used, the mobile project does not compile due to:

```
Projects that use central package version management should not define the version on the PackageReference items but on the PackageVersion items:
 Microsoft.Windows.SDK.BuildTools;Uno.Core.Extensions.Logging.Singleton;Microsoft.WindowsAppSDK;Uno.WinUI;Microsoft.Extensions.Logging.Console.
```

Even though the `PackageReferences` definitely don't have a `Version` attribute.

### Expected behavior

Should compile

- Clone https://github.com/MartinZikmund/uno-11641-repro
- Try to build Mobile project

### Workaround

Not using Central Package Management

### Works on UWP/WinUI

None

### Environment

_No response_

### NuGet package version(s)

_No response_

### Affected platforms

_No response_

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

Youssef1313:
GitHub Issue (If applicable): closes #11641

For non-design-time builds, AndroidManifest is correctly set from here:

https://github.com/unoplatform/uno/blob/560a44980945eb8343025832fe0ea23941c99435/build/uno.winui.single-project.targets#L58

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/SolutionTemplate/UnoSolutionTemplate.WinUI.netcore/Mobile/UnoQuickStart.Mobile.csproj
+++ b/src/SolutionTemplate/UnoSolutionTemplate.WinUI.netcore/Mobile/UnoQuickStart.Mobile.csproj
@@ -26,6 +26,7 @@
 		<ApplicationDisplayVersion>1.0</ApplicationDisplayVersion>
 		<ApplicationVersion>1</ApplicationVersion>
 
+		<AndroidManifest>Android\AndroidManifest.xml</AndroidManifest>
 		<IsUnoHead>true</IsUnoHead>
 
 		<SupportedOSPlatformVersion Condition="'$(TargetFramework)' == '$basetargetframework$-ios'">14.2</SupportedOSPlatformVersion>


src/SolutionTemplate/UnoSolutionTemplate.WinUI.netcore/Mobile/UnoQuickStart.Mobile.csproj
```diff
@@ -26,6 +26,7 @@
 		<ApplicationDisplayVersion>1.0</ApplicationDisplayVersion>
 		<ApplicationVersion>1</ApplicationVersion>
 
+		<AndroidManifest>Android\AndroidManifest.xml</AndroidManifest>
```
jeromelaban, line 29: This is a symptom of another problem, generally caused by nuget packages not being restored properly.

Youssef1313: @jeromelaban NuGet needs a passing design-time build in order to restore, and the error about AndroidManifest happens in design-time builds by a target that is run before anything from NuGet runs. So, AndroidManifest needs to be set early.

MartinZikmund: I assume same adjustment will be needed in the new templates as well (in Uno Extensions)

MartinZikmund:
@jeromelaban for reference for the reason for this change - https://github.com/NuGet/Home/issues/12480

jeromelaban:
The fix here is a workaround for the fact that we're not using an SDK (which can be executed prior to any design-time build). SDKExtras uses the `sdk/XX.XX` syntax, but this has its own set of issues (such as freezing the IDE because of dispatcher-bound external downloads).

This is not an ideal solution, but as nuget relies on design-time builds to work, this is the only way to work around it at this time, indeed.

MartinZikmund (approved):

PR #11706, opened 2023-03-15
Pointers issue

Youssef1313:

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/SamplesApp/SamplesApp.UITests/Windows_UI_Xaml_Media_Animation/TranslateTransformClickability.cs
+++ b/src/SamplesApp/SamplesApp.UITests/Windows_UI_Xaml_Media_Animation/TranslateTransformClickability.cs
@@ -9,7 +9,7 @@ using SamplesApp.UITests.TestFramework;
 namespace SamplesApp.UITests.Windows_UI_Xaml_Media_Animation
 {
 	[TestFixture]
-	public class TranslateTransformClickability : SampleControlUITestBase
+	public partial class TranslateTransformClickability : SampleControlUITestBase
 	{
 		[Test]
 		[AutoRetry]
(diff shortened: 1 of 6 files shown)


dr1rrb (commented):
I think that we should go that way.

For me the issue is that the native `dispatchTouchEvent` (and `dispatchGenericMotionEvent`) (invoked [here](https://github.com/unoplatform/uno/blob/master/src/Uno.UI.BindingHelper.Android/Uno/UI/UnoMotionHelper.java#L191)) in the `ViewGroup` just does not support the case you are trying to fix, i.e. a child is out of the bounds of its parent.

If we look in the implementation, it validates [if coordinates are above a child before dispatch the event to it](https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/view/ViewGroup.java;l=2731;drc=7884e551eb6f054c86bf7dbf138a2351104e2414;bpv=0;bpt=1?q=ViewGroup&ss=android%2Fplatform%2Fsuperproject but the [`isTransformedTouchPointInView`](https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/view/ViewGroup.java;drc=7884e551eb6f054c86bf7dbf138a2351104e2414;bpv=0;bpt=1;l=3048) ignores your case.

The only way to fix this is to manually dispatch events instead of relaying on the platform ... which is almost what you did by forcing almost all events to go through our  `dispatchStaticTransformMotionEvent`. The issue for me is that this is known to have some limitation and mostly relies on the fact that the platform is robust enough to fix the invalid flags we are providing ... as long as we are doing it only for few controls / layers.

And instead of re-implementing such custom dispatch in java with all the small tricks that implies, my preference would definitely be to re-use the dispatch we did for skia (which is already available for Android as we are using it also for Drag&Drop), and go back to native only when we reach native only elements (this is the piece that is missing so far). This would also probably more performant as we would significantly reduce the quantity of interop.

Youssef1313:
Replacing by https://github.com/unoplatform/uno/pull/11737

Issue #7393, opened 2021-10-27
[Android/iOS] Buttons that are translated out of their parent Panel are not clickable


kazo0:
### Current behavior

```xml
<Grid Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">
	<StackPanel Background="Red" VerticalAlignment="Center" HorizontalAlignment="Center">
		<Button Content="Test1">
			<Button.RenderTransform>
					<TranslateTransform Y="-32" />
				</Button.RenderTransform>
		</Button>
		<Button Content="Test2" />

		<Button Content="Test3" />
	</StackPanel>
</Grid>
```

![image](https://user-images.githubusercontent.com/4793020/139148798-44dd788b-8b2d-459c-a178-8ef1906176d3.png)

Given the above XAML, the "Test1" button will be translated outside of the StackPanel and will no longer receive pointer events

### Expected behavior

Should behave like Windows and still be clickable

### Works on UWP/WinUI

Yes

### Environment

Uno.UI / Uno.UI.WebAssembly / Uno.UI.Skia

### Affected platforms

iOS, Android

MartinZikmund:
No longer relevant for Skia targets, where it works correctly

PR #11737, opened 2023-03-17
fix(android): Fix clicking on element out of parent bounds

Youssef1313:
GitHub Issue (If applicable): Android part of https://github.com/unoplatform/uno/issues/7393

## PR Type

## PR Checklist

## Other information


src/Uno.UWP/UI/Core/CoreWindow.Android.cs
```diff
@@ -0,0 +1,12 @@
...
+using Windows.Devices.Input;
+
+namespace Windows.UI.Core;
+
+internal sealed partial class CoreWindowExtension : ICoreWindowExtension
+{
+	public CoreCursor PointerCursor { get => throw new NotImplementedException(); set => throw new NotImplementedException(); }
+
+	public void ReleasePointerCapture(PointerIdentifier pointer) => throw new NotImplementedException();
+	public void SetPointerCapture(PointerIdentifier pointer) => throw new NotImplementedException();
```
dr1rrb, line 11: I think that Android allows capture, so it's probably just a "TODO", but in all cases you should not throw an exception here. It's used by a lot of controls (like `ButtonBase`), throwing here will break internal state.

src/Uno.UWP/UI/Core/PointerEventArgs.cs
```diff
@@ -12,6 +12,16 @@ internal PointerEventArgs(PointerPoint currentPoint, VirtualKeyModifiers keyModi
...
 		}
 
+#if __ANDROID__
+		internal PointerEventArgs(PointerPoint currentPoint, VirtualKeyModifiers keyModifiers, Android.Views.MotionEvent motionEvent)
+			: this(currentPoint, keyModifiers)
+		{
+			MotionEvent = motionEvent;
+		}
+
+		public Android.Views.MotionEvent MotionEvent { get; }
```
dr1rrb, line 22:
```suggestion
		internal Android.Views.MotionEvent MotionEvent { get; }
```

src/Uno.UI/UI/Xaml/UIElement.Pointers.Managed.cs
```diff
@@ -71,7 +71,7 @@ private object CoerceHitTestVisibility(object baseValue)
 
 			// If we're not locally hit-test visible, visible, or enabled, we should be collapsed. Our children will be collapsed as well.
 			if (
-#if !__MACOS__
+#if !__MACOS__ && !__ANDROID__
```
dr1rrb, line 74: If you have to add this I think this is because this `HitTestVisibility` property coercing method is not re-evaluated on load. An unloaded element should be considered as `Collapsed`.

src/Uno.UI/UI/Xaml/UIElement.Pointers.Android.cs
```diff
@@ -43,125 +50,221 @@ partial class UIElement
 			}
 		}
 
-		protected sealed override bool OnNativeMotionEvent(MotionEvent nativeEvent, View originalSource, bool isInView)
+		protected sealed override bool OnNativeMotionEvent(MotionEvent nativeEvent)
```
dr1rrb, line 53: If I'm right currently there 2 cases where this method can be invoked now:
1. On the `VisualRoot` where you have to route the event to the `ICoreWIndowEvents` like you did here. In that case the `ownerEvents` will determine the `OriginalSource`, and start bubbling in managed code from this element. 
I would move this case to the [`dispatchTouchEvent`](https://developer.android.com/reference/android/view/View#dispatchTouchEvent(android.view.MotionEvent)) (and `dispatchGenericMotionEvent`) before Android even try to determine the target element.
3. When transitioning back from native bubbling to managed (i.e. when bubbling reached a native only element, we stopped managed bubbling and invoked something like `OnTouchEvent` on its parent to initiate / simulate native bubbling). 
In that case we must not forward to the `ownerEvents` but only resume managed bubbling (i.e. invoke `OnPointerXXX` methods).
Also, we should not re-create a new `PointerEventArgs` but re-use the already existing `PointerROUTEDEventsArgs` with the right `OriginalSource`.

Also I noticed that you remove the ability to raise multiple pointer events from a single `MotionEvent`. I think this is still relevant, especially for multi-touch (on Android a `MotionEvent` can contains info on multiple pointers at once while on Windows we will get invoked multiple times for each pointer).

src/Uno.UI/UI/Xaml/UIElement.Pointers.Managed.cs
Youssef1313: I added this because it seems there is no `UIElement.IsLoaded` on Android at all.

src/Uno.UI/UI/Xaml/UIElement.Pointers.cs
```diff
@@ -768,6 +769,19 @@ internal void RaiseDrop(global::Windows.UI.Xaml.DragEventArgs args)
...
+
+#if __ANDROID__
+			if (this is not RootVisual &&
+				this.Parent is not UIElement &&
+				this.Parent is Android.Views.View viewParent &&
+				ptArgs.MotionEvent is { } motionEvent
+				)
+			{
+				// bubblingMode = BubblingMode.NoBubbling;
+				viewParent.OnTouchEvent(motionEvent);
```
dr1rrb, line 781: I don't  remember if it's required or not, but you might have to adjust coordinates on `motionEvent` before in order to make coordinates relative to the `viewParent` element. (You can follow what's done in native : clone it using the right [`obtain`](https://developer.android.com/reference/android/view/MotionEvent#obtain(android.view.MotionEvent)), then [`offsetLocation`](https://developer.android.com/reference/android/view/MotionEvent#offsetLocation(float,%20float)) and finally [`recycle`](https://developer.android.com/reference/android/view/MotionEvent#recycle())) it.

Youssef1313:
This is a stale PR. Closing out. We'll need to move to managed pointers on Android

Issue #11769, opened 2023-03-22
Remove all JSInvoke uses for CSP compatibility


jeromelaban:

ebariche:
InvokeJS (Count = 88)
===============
[ ] AddIns\Uno.UI.Lottie\LottieVisualSourceBase.wasm.cs(315)
    private static string InvokeJs(string[] js) => WebAssemblyRuntime.InvokeJS(string.Concat(js));
[ ] Uno.Foundation.Runtime.WebAssembly\Interop\JSObjectMetadataProvider.wasm.cs(48)
    WebAssemblyRuntime.InvokeJS(prototype);
[ ] Uno.Foundation.Runtime.WebAssembly\Interop\JSObjectMetadataProvider.wasm.cs(53)
    WebAssemblyRuntime.InvokeJS($"{_type.FullName}.createInstance(\"{managedHandle}\", \"{id}\")");
[ ] Uno.Foundation.Runtime.WebAssembly\Interop\JSObjectMetadataProvider.wasm.cs(64)
    => WebAssemblyRuntime.InvokeJS($"{_type.FullName}.destroyInstance(\"{managedHandle}\", \"{jsHandle}\")");
[ ] Uno.Foundation.Runtime.WebAssembly\Interop\Runtime.wasm.cs(328)
    return InvokeJS(command);
[ ] Uno.UI.Runtime.WebAssembly\WasmHttpHandler.cs(62)
    WebAssemblyRuntime.InvokeJS(command);
[ ] Uno.UI.Runtime.WebAssembly\Web\Http\CookieManager.cs(56)
    WebAssemblyRuntime.InvokeJS(jsInvoke);
[ ] Uno.UI.Runtime.WebAssembly\Web\Http\CookieManager.cs(105)
    var cookies = WebAssemblyRuntime.InvokeJS("document.cookie");
[ ] Uno.UI.Runtime.WebAssembly\Xaml\UIElementWasmExtensions.cs(130)
    return WebAssemblyRuntime.InvokeJS(js);
[ ] Uno.UI\UI\Xaml\Controls\Image\Image.wasm.cs(112)
    WebAssemblyRuntime.InvokeJS("Uno.UI.WindowManager.current.setImageAsMonochrome(" + _htmlImage.HtmlId + ", \"" + 
    img.Value + "\", \"" + MonochromeColor.Value.ToHexString() + "\");");
[...]

PR #11782, opened 2023-03-23
chore: Remove WPF hosting mode

ebariche:
Issue: unoplatform/uno#11769

## PR Type

- Other... Please describe: Remove broken feature

## What is the new behavior?

WPF hosting mode was used to help debug WASM apps. We're removing it because WASM debugging has improved significantly since then, the feature has been broken for a while and it is getting in the way of supporting CSP (https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP).


jeromelaban (approved):

pkar70:
By the way, why most of PRs has no any comments, just code? No information why it is needed, etc.?

PR #11912, opened 2023-04-05
fix: XAML Hot reload mishandling empty paths

### LINKED ISSUE #11911: XAML Hot Reload sometimes fails due to empty string for FileSystemWatcher
### Current behavior

"Sometimes" (it happened to me once and I don't know how/why) the RemoteControl Source Generator is creating a path that's just an empty string and passing this to the FileSystemWatcher to watch for XAML file changes.

This causes a silent  error and leaves the RemoteControl in a broken state so that HotReload (or anything else using the RemoteControl) fails to work.

### Expected behavior

Not create empty paths where there shouldn't be any.
and
Not fail silently in a way that's hard to understand what went wrong and why.

Private repo can be shared if necessary - I don't know how to forcibly recreate this issue :(

### Workaround

I can't find a workaround 

### Works on UWP/WinUI

None

### Environment

_No response_

### NuGet package version(s)

4.9.0-dev.362

### Affected platforms

WebAssembly, Skia (WPF)

### IDE

Visual Studio 2022

### IDE version

17.4.4

### Relevant plugins

none

_No response_

mrlacey:
GitHub Issue (If applicable): closes #11911

## PR Type

- Bugfix

Empty paths cause RemoteControl to fail silently

## What is the new behavior?

Don't pass Empty strings to RemoteControl to watch for file changes

## PR Checklist

## Other information

I can't find any tests for this area fo the product and don't know where/how some should be added.

### DIFF
--- a/src/SourceGenerators/Uno.UI.SourceGenerators/RemoteControl/RemoteControlGenerator.cs
+++ b/src/SourceGenerators/Uno.UI.SourceGenerators/RemoteControl/RemoteControlGenerator.cs
@@ -119,8 +119,8 @@ namespace Uno.UI.SourceGenerators.RemoteControl
 		{
 			var xamlPaths = EnumerateLocalSearchPaths(context);
 
-            if (xamlPaths.Any())
-            {
+			if (xamlPaths.Any())
+			{
 				var distictPaths = string.Join(Path.PathSeparator.ToString(), xamlPaths);
 
 				sb.AppendLineIndented($"""
--- a/src/SourceGenerators/Uno.UI.SourceGenerators/RemoteControl/RemoteControlGenerator.cs
+++ b/src/SourceGenerators/Uno.UI.SourceGenerators/RemoteControl/RemoteControlGenerator.cs
@@ -118,14 +118,18 @@ namespace Uno.UI.SourceGenerators.RemoteControl
 		private static void BuildLibrarySearchPath(GeneratorExecutionContext context, IndentedStringBuilder sb)
 		{
 			var xamlPaths = EnumerateLocalSearchPaths(context);
-			var distictPaths = string.Join(Path.PathSeparator.ToString(), xamlPaths);
 
-			sb.AppendLineIndented($"""
-				[assembly: global::System.Reflection.AssemblyMetadata(
-					"{LibraryXamlSearchPathAssemblyMetadata}",
-					@"{distictPaths}"
-				)]
-				""");
+            if (xamlPaths.Any())
+            {
+				var distictPaths = string.Join(Path.PathSeparator.ToString(), xamlPaths);
+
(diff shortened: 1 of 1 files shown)


jeromelaban (approved):

src/SourceGenerators/Uno.UI.SourceGenerators/RemoteControl/RemoteControlGenerator.cs
```diff
@@ -176,7 +176,7 @@ IEnumerable<string> BuildSearchPath(string s)
 			var xamlPaths = from item in sources.SelectMany(BuildSearchPath)
 							select Path.GetDirectoryName(item);
 
-			return xamlPaths.Distinct();
+			return xamlPaths.Distinct().Where(x => !string.IsNullOrEmpty(x));
```
Youssef1313, line 179: Looks more of a workaround? We shouldn't be getting null/empty path I think?

mrlacey: Yep. If you can provide advice on how to tell why empty strings are appearing (& how to prevent them) I'll happily fix them there too.

Youssef1313: I don't, but wanted to make sure I understand the change. In case I got to make changes to this generator, I'll remember to check that and let you know

mrlacey: Having done some digging - here's what's happening.
I have a library that references Uno.UI.RemoteControl but contains no XAML files. When the library is built, [the Source generators recognize this as a "ValidPlatform"](https://github.com/unoplatform/uno/blob/master/src/SourceGenerators/Uno.UI.SourceGenerators/RemoteControl/RemoteControlGenerator.cs#L49) and so adds a `XamlSearchPath` attribute to the library even though there are not XAML files.
Then, when the application that references the library is built, the source generators for the app find the (empty string) attribute and adds that to the list to pass to the RemoteControlServer to watch as part of XamlHotReload.

I've now added some extra checks to avoid the empty strings being added to attributes.
This seemed safer than trying to update the existing logic to handle my case for 2 reasons:
1 - With no test coverage for this part of the code, I wanted to do the minimum I could that would mitigate the chance of breaking something while addressing the bug.
2 - This avoids any other (potential, future) scenario that might cause empty strings from ending up in that list from causing problems.

jeromelaban: It's indeed appropriate to have a failsafe at that location.

Issue #11610, opened 2021-02-05
[macOS/Skia.Gtk] [menu] Multiple items remain in focused state when scrolling in the left menu


DennyPerez18:
## Current behavior
Multiple items remain in selected state when scrolling in the left menu
![Screen Shot 2021-02-04 at 5 19 49 PM](https://user-images.githubusercontent.com/38770890/107043032-6f963a80-6790-11eb-801e-c73acfd281ef.png)

## Expected behavior
A focused state should be displayed only when cursor focus on the item

## How to reproduce it (as minimally and precisely as possible)
1. From the left menu scroll to the bottom on the list
2. Scroll to the top
3. Notice some elements remain in a focused state

## Environment

Nuget Package:

Package Version(s):

Affected platform(s):

- [ ] iOS
- [x] macOS
- [ ] Android
- [ ] WebAssembly
- [ ] WebAssembly renders for Xamarin.Forms
- [ ] Windows
- [ ] Build tasks

Visual Studio:

- [ ] 2017 (version: )
- [ ] 2019 (version: )
- [x] for Mac (version: )

Relevant plugins:

- [ ] Resharper (version: )

brswan:
@Xiaoy312 Created issue from the link you provided in the comment.

Issue: https://github.com/unoplatform/Uno.Gallery/issues/658

MartinZikmund:
@sakshi173 can you check this in the latest canaries after the fix by @dr1rrb in #11966?

dr1rrb:
FYI: I've tested the gallery for another bug, and I was no longer able to repro

jeromelaban:
@sakshi173 could you take a look again ? Thanks!

PR #11966, opened 2023-04-13
fix(pointers): Fix invalid pointer state is visual tree altered during dispatch

dr1rrb:
linked/close? https://github.com/unoplatform/uno/issues/11612
linked/close? https://github.com/unoplatform/uno/issues/11610

## Bugfix
Fix invalid pointer state is visual tree altered during dispatch

On native pointer move we do generate multiple managed events (Exit / enter / move). If the visual is modified by the exit, we will still send the enter event even if the visual tree has been modified and the target is no longer valid.

## What is the new behavior?
We make sure to re-compute the target top most element (a.k.a. `OriginalSource`) before dispatching events, but only if we detect a visual tree changes that could drive to an invalid state (to preserve perf).

## PR Checklist


src/Uno.UI/UI/Xaml/Controls/Primitives/SelectorItem.cs
```diff
@@ -180,7 +181,7 @@ private void UpdateCommonStates(ManipulationUpdateKind manipulationUpdate = Mani
 				pause = false;
 			}
 
-			if (delay < TimeSpan.Zero)
+			if (delay <= TimeSpan.Zero)
```
jeromelaban, line 184: Why is it needed?

src/Uno.UI/UI/Xaml/HitTestability.cs
```diff
@@ -30,4 +30,6 @@ internal enum HitTestability
 		/// </summary>
 		Visible,
 	}
+
+	internal record struct StalePredicate(Predicate<UIElement> Method, string Name);
```
jeromelaban, line 34: As it's internal, it would be better to avoid using a generic delegate type. You can create a delegate type specialized for this case.

src/Uno.UI/UI/Xaml/Internal/PointerCapture.cs
```diff
@@ -0,0 +1,346 @@
...
+		}
+		else
+		{
+			cloneOfAllCaptures = default;
+			return false;
+		}
+	}
+
+	private UIElement _nativeCaptureElement;
+	private readonly Dictionary<UIElement, PointerCaptureTarget> _targets = new Dictionary<UIElement, PointerCaptureTarget>(2);
```
jeromelaban, line 59: Can this leak somehow?

src/Uno.UI/UI/Xaml/Internal/PointerCapture.cs
```diff
@@ -0,0 +1,346 @@
...
+	internal PointerCaptureResult TryAddTarget(UIElement element, PointerCaptureKind kind, PointerRoutedEventArgs relatedArgs = null)
+	{
+		global::System.Diagnostics.Debug.Assert(
+			kind == PointerCaptureKind.Explicit || kind == PointerCaptureKind.Implicit,
+			"The initial capture kind must be Explicit **OR** Implicit.");
+
+		if (this.Log().IsEnabled(LogLevel.Information))
+		{
+			this.Log().Info($"{element.GetDebugName()}: Capturing ({kind}) pointer {Pointer}");
+		}
```
jeromelaban, line 102: Debug?

src/Uno.UI/UI/Xaml/Internal/PointerCapture.cs
```diff
@@ -0,0 +1,346 @@
...
+		target.LastDispatched = args;
+		if (MostRecentDispatchedEventFrameId < args.FrameId)
+		{
+			MostRecentDispatchedEventFrameId = args.FrameId;
+		}
+	}
+
+	private void EnsureEffectiveCaptureState()
+	{
+		if (_targets.Any())
```
jeromelaban, line 268: It's probably best to avoid .Any() (it uses casts internally)

src/Uno.UI/UI/Xaml/Internal/PointerCapture.cs
```diff
@@ -0,0 +1,346 @@
...
+	[MethodImpl(MethodImplOptions.AggressiveInlining)]
+	private void ReleasePointerNative()
+	{
+		try
+		{
+			_nativeCaptureElement.ReleasePointerNative(Pointer);
+		}
+		catch (Exception e)
+		{
+			this.Log().Error($"Failed to release native capture of {Pointer}", e);
```
jeromelaban, line 325: If the release fails, can we end up in an unrecoverable state?

src/Uno.UI/UI/Xaml/Media/VisualTreeHelper.cs
```diff
@@ -485,29 +494,65 @@ static VisualTreeHelper()
...
 							{
-								stale = SearchDownForStaleBranch(child.Current!, isChildStale);
+								TRACE($"- Is {isChildStale.Value.Name}");
+
+								stale = SearchDownForStaleBranch(child.Current!, isChildStale.Value);
+
+#if TRACE_HIT_TESTING
+								while (child.MoveNext())
+								{
+									using var ___ = SET_TRACE_SUBJECT(child.Current);
```
jeromelaban, line 528: One should be enough?

```suggestion
									using var _ = SET_TRACE_SUBJECT(child.Current);
```

src/Uno.UI/UI/Xaml/DragDrop/DropUITarget.cs
```diff
@@ -131,7 +131,7 @@ public IAsyncOperation<DataPackageOperation> DropAsync(CoreDragInfo dragInfo)
 				dragInfo.Position,
 				Window.Current.RootElement.XamlRoot, //TODO: Choose proper XamlRoot https://github.com/unoplatform/uno/issues/8978
 				getTestability: GetDropHitTestability,
-				isStale: elt => elt.IsDragOver(dragInfo.SourceId));
+				isStale: new (elt => elt.IsDragOver(dragInfo.SourceId), "IsDragOver"));
```
Youssef1313, line 134: I usually prefer explicit type when type isn't clear just by looking.

```suggestion
				isStale: new StalePredicate(elt => elt.IsDragOver(dragInfo.SourceId), "IsDragOver"));
```

src/Uno.UI/UI/Xaml/Internal/InputManager.Pointers.Managed.cs
```diff
@@ -368,7 +364,7 @@ private void ReleaseCaptures(PointerRoutedEventArgs routedArgs)
 		{
 			if (PointerCapture.TryGet(routedArgs.Pointer, out var capture))
 			{
-				foreach (var target in capture.Targets)
+				foreach (var target in capture.Targets.ToList())
```
Youssef1313, line 367: Why?

jeromelaban (commented):
Is testing possible? (even partial)

src/Uno.UI/UI/Xaml/Internal/PointerCapture.cs
```diff
@@ -0,0 +1,346 @@
+﻿using System;
```
Youssef1313, line 1: `#nullable enable`

src/Uno.UI/UI/Xaml/Internal/PointerCapture.cs
```diff
@@ -0,0 +1,346 @@
...
+		=> _actives.TryGetValue(pointer, out capture);
+
+	public static bool TryGet(Pointer pointer, out PointerCapture capture)
+		=> _actives.TryGetValue(pointer.UniqueId, out capture);
+
+	public static bool Any(out List<PointerCapture> cloneOfAllCaptures)
+	{
+		if (_actives.Any())
+		{
+			cloneOfAllCaptures = _actives.Values.ToList();
```
Youssef1313, line 48: Do we really need `List<T>`? Is `Values` (`ICollection<T>`) not sufficient?

src/Uno.UI/UI/Xaml/Internal/PointerCapture.cs
```diff
@@ -0,0 +1,346 @@
...
+
+	internal static bool TryGet(PointerIdentifier pointer, out PointerCapture capture)
+		=> _actives.TryGetValue(pointer, out capture);
+
+	public static bool TryGet(Pointer pointer, out PointerCapture capture)
+		=> _actives.TryGetValue(pointer.UniqueId, out capture);
+
+	public static bool Any(out List<PointerCapture> cloneOfAllCaptures)
+	{
+		if (_actives.Any())
```
Youssef1313, line 46:
```suggestion
		if (_actives.Count > 0)
```

src/Uno.UI/UI/Xaml/Internal/PointerCapture.cs
```diff
@@ -0,0 +1,346 @@
...
+
+	internal IEnumerable<PointerCaptureTarget> GetTargets(PointerCaptureKind kinds)
+		=> _targets
+			.Values
+			.Where(target => (target.Kind & kinds) != PointerCaptureKind.None);
+
+	internal PointerCaptureResult TryAddTarget(UIElement element, PointerCaptureKind kind, PointerRoutedEventArgs relatedArgs = null)
+	{
+		global::System.Diagnostics.Debug.Assert(
+			kind == PointerCaptureKind.Explicit || kind == PointerCaptureKind.Implicit,
```
Youssef1313, line 96:
```suggestion
			kind is PointerCaptureKind.Explicit or PointerCaptureKind.Implicit,
```

src/Uno.UI/UI/Xaml/Media/VisualTreeHelper.cs
Youssef1313: @jeromelaban It is not. This is currently considered as a regular variable by the compiler, not a discard.

There is an open proposal somewhere in dotnet/csharplang for that.

src/Uno.UI/UI/Xaml/Internal/InputManager.Pointers.Managed.cs
```diff
@@ -289,24 +279,30 @@ internal void OnPointerMoved(Windows.UI.Core.PointerEventArgs args)
...
 			}
 
 			// Second (try to) raise the PointerEnter on the OriginalSource
 			// Note: This won't do anything if already over.
-			routedArgs.Handled = false;
-			Raise(Enter, originalSource, routedArgs);
+			if (Raise(Enter, originalSource, routedArgs) is { VisualTreeAltered: true })
+			{
+				// The visual tree have been modified in a way that requires to perform a new hit test.
+				originalSource = HitTest(args).element ?? Windows.UI.Xaml.Window.Current.Content;
```
Youssef1313, line 302: Quite concerned about performance. I see that `VisualTreeAltered` is added to reduce the frequency of having to re-call `HitTest`, but still not sure if we will still be calling HitTest too much and if it will have a negative impact.

src/Uno.UI/UI/Xaml/Internal/InputManager.Pointers.Managed.cs
dr1rrb: Because the `ReleaseointerCapture` below is expected to modify the `capture.Targets` list, so we need to clone it before.

src/Uno.UI/UI/Xaml/Internal/InputManager.Pointers.Managed.cs
dr1rrb: Yes it will have a negative perf impact, that's why I added that flag that should be `true` in few rare cases. And actually it does not cover all cases (for instance if a `RenderTransform` is updated by the enter we will stiff dispatch the move to it), we are only covering cases that could drive to be stuck into an invalid state (pressed while element is no longer in the visual tree), other cases will just be recovered on next pointer event.

src/Uno.UI/UI/Xaml/Internal/PointerCapture.cs
dr1rrb: (Fix coming ... and found 3 places in the class of a possible nuff ref 😨)

src/Uno.UI/UI/Xaml/Internal/PointerCapture.cs
dr1rrb: Theoretically yes, but there is a usage in the code where we are using the indexer on `cloneOfAllCaptures`. As it's internal code, I think it's better to give "full" access on it to avoid cloning the list again and again.

src/Uno.UI/UI/Xaml/Internal/PointerCapture.cs
dr1rrb: No: the captures are tracked and released on pointer up or if the element is unloaded. If this leaks it means that we have a bigger problem :)

src/Uno.UI/UI/Xaml/Internal/PointerCapture.cs
```diff
@@ -0,0 +1,346 @@
...
+		}
+	}
+
+	private UIElement _nativeCaptureElement;
+	private readonly Dictionary<UIElement, PointerCaptureTarget> _targets = new Dictionary<UIElement, PointerCaptureTarget>(2);
+
+	private PointerCapture(Pointer pointer)
+	{
+		Pointer = pointer;
+	}
```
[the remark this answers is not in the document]
dr1rrb: Yes, but members above are `static`, I usually keep `static` members at the top of the `class` ... but indeed according to the SA1204 I should not (static members should be only on top a given group). But to be honest we don't enforce such rules in the uno project, we usually prefer to order member per functionality than by access modifier.

src/Uno.UI/UI/Xaml/Internal/PointerCapture.cs
dr1rrb: Yes and no : depends of the platform, but I would assume it would be only until the pointer is being released. Anyway this will impact only capture for which pointer goes outside of the bounds of the app.

src/Uno.UI/UI/Xaml/Controls/Primitives/SelectorItem.cs
dr1rrb: It was just an invalid comparison I found (as I was forcing the delay to 0 to cause my bug :) )

dr1rrb:
> Is testing possible? (even partial)

Was working on it, it's tested now :)

src/Uno.Foundation/Extensibility/ApiExtensibility.cs
```diff
@@ -61,7 +65,8 @@ public static void Register<TOwner>(Type type, Func<TOwner, object> builder)
 	/// <param name="owner">An optional owner to be passed to the extension constructor</param>
 	/// <param name="instance">The instance if the creation was successful</param>
 	/// <returns>True if the creation suceeded, otherwise False.</returns>
```
agneszitte, line 67:
```suggestion
	/// <param name="owner">An optional owner to be passed to the extension constructor</param>
	/// <param name="instance">The instance if the creation was successful</param>
	/// <returns>True if the creation succeeded, otherwise False.</returns>
```

src/Uno.UI.RuntimeTests/Helpers/UITestHelper.cs
```diff
@@ -0,0 +1,200 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using Windows.Foundation;
+using Windows.UI.Input.Preview.Injection;
+using Windows.UI.Xaml;
+using Private.Infrastructure;
+
+namespace Uno.UI.RuntimeTests.Tests.Uno_UI_Xaml_Core;
+
+// Note: This files contains bunch of helpers that are expected to be moved to the test engine among the pointer injection work
```
agneszitte, line 14:
```suggestion
// Note: This file contains a bunch of helpers that are expected to be moved to the test engine among the pointer injection work
```

src/Uno.UI/UI/Xaml/Internal/InputManager.Pointers.Managed.cs
```diff
@@ -163,7 +154,7 @@ internal void OnPointerExited(Windows.UI.Core.PointerEventArgs args)
 		{
 			// This is how UWP behaves: when out of the bounds of the Window, the root element is use.
```
agneszitte, line 155:
```suggestion
			// This is how UWP behaves: when out of the bounds of the Window, the root element is used.
```

src/Uno.UI/UI/Xaml/Internal/InputManager.Pointers.Managed.cs
```diff
@@ -289,24 +279,30 @@ internal void OnPointerMoved(Windows.UI.Core.PointerEventArgs args)
...
 
 			var routedArgs = new PointerRoutedEventArgs(args, originalSource);
 
 			// First raise the PointerExited events on the stale branch
 			if (staleBranch.HasValue)
 			{
-				Raise(Leave, staleBranch.Value, routedArgs);
+				if (Raise(Leave, staleBranch.Value, routedArgs) is { VisualTreeAltered: true })
+				{
+					// The visual tree have been modified in a way that requires to perform a new hit test.
```
agneszitte, line 292:
```suggestion
					// The visual tree has been modified in a way that requires performing a new hit test.
```

src/Uno.UI/UI/Xaml/Internal/InputManager.Pointers.Managed.cs
```diff
@@ -289,24 +279,30 @@ internal void OnPointerMoved(Windows.UI.Core.PointerEventArgs args)
...
+				}
 			}
 
 			// Second (try to) raise the PointerEnter on the OriginalSource
 			// Note: This won't do anything if already over.
-			routedArgs.Handled = false;
-			Raise(Enter, originalSource, routedArgs);
+			if (Raise(Enter, originalSource, routedArgs) is { VisualTreeAltered: true })
+			{
+				// The visual tree have been modified in a way that requires to perform a new hit test.
```
agneszitte, line 301:
```suggestion
				// The visual tree has been modified in a way that requires performing a new hit test.
```

src/Uno.UI/UI/Xaml/Internal/PointerCapture.cs
```diff
@@ -0,0 +1,362 @@
...
+
+/*
+ *	This file is intrinsically linked to the UIElement and uses some of its internal APIs:
+ *		internal List<Pointer> PointerCapturesBackingField;
+ *		internal partial void CapturePointerNative(Pointer pointer);
+ *		internal partial void ReleasePointerNative(Pointer pointer);
+ */
+
+/// <summary>
+/// This is an helper class that use to manage the captures for a given pointer.
```
agneszitte, line 25:
```suggestion
/// This is a helper class that uses to manage the captures for a given pointer.
```

src/Uno.UI/UI/Xaml/Internal/PointerCapture.cs
```diff
@@ -0,0 +1,362 @@
...
+		Pointer = pointer;
+	}
+
+	/// <summary>
+	/// The captured pointer
+	/// </summary>
+	public Pointer Pointer { get; }
+
+	/// <summary>
+	/// Gets the <see cref="PointerRoutedEventArgs.FrameId"/> of the last args that has been handled by this capture
```
agneszitte, line 74:
```suggestion
	/// Gets the <see cref="PointerRoutedEventArgs.FrameId"/> of the last args that have been handled by this capture
```

src/Uno.UI/UI/Xaml/Internal/PointerCapture.cs
```diff
@@ -0,0 +1,362 @@
...
+	/// </summary>
+	public long MostRecentDispatchedEventFrameId { get; private set; }
+
+	/// <summary>
+	/// Determines if this capture was made only for implicit kind
+	/// (So we should not use it to filter out some event on other controls)
```
agneszitte, line 80:
```suggestion
	/// Determines if this capture was made only for an implicit kind
	/// (So we should not use it to filter out some events on other controls)
```

src/Uno.UI/UI/Xaml/Internal/PointerCapture.cs
```diff
@@ -0,0 +1,362 @@
...
+		// Validate / update the state of this capture
+		EnsureEffectiveCaptureState();
+	}
+
+	/// <summary>
+	/// Validate if the provided routed event args are relevant for the given element according to the active captures
+	/// </summary>
+	/// <param name="element">The target element for which the args are validated</param>
+	/// <param name="args">The pending pointer event args that is under test</param>
+	/// <param name="autoRelease">A flag that allows to automatically release any pending out-dated capture (for PointerDown only)</param>
```
agneszitte, line 219:
```suggestion
	/// <param name="autoRelease">A flag that allows releasing automatically any pending outdated capture (for PointerDown only)</param>
```

src/Uno.UI/UI/Xaml/Internal/PointerCapture.cs
```diff
@@ -0,0 +1,362 @@
...
+		{
+			// If the capture is implicit, we should not filter out events for children elements.
+
+			return true;
+		}
+		else
+		{
+			// We should dispatch the event only if the control which has captured the pointer has already dispatched the event
+			// (Which actually means that the current control is a parent of the control which has captured the pointer)
+			// Remarks: This is not enough to determine parent-child relationship when we dispatch multiple events base on the same native event,
```
agneszitte, line 250:
```suggestion
			// Remarks: This is not enough to determine the parent-child relationship when we dispatch multiple events based on the same native event,
```

src/Uno.UI/UI/Xaml/Internal/PointerCapture.cs
```diff
@@ -0,0 +1,362 @@
...
+			//			(as they will all have the same FrameId), however in that case we dispatch events layer per layer
```
agneszitte, line 251:
```suggestion
			//			(as they will all have the same FrameId), however, in that case, we dispatch events layer per layer
```

src/Uno.UI/UI/Xaml/Internal/PointerCaptureTarget.cs
```diff
@@ -0,0 +1,60 @@
...
+		Kind = kind;
+	}
+
+	/// <summary>
+	/// The target element to which event args should be forwarded
+	/// </summary>
+	public UIElement Element { get; }
+
+	/// <summary>
+	/// The element to used for the native capture
```
agneszitte, line 24:
```suggestion
	/// The element to be used for the native capture
```

src/Uno.UI/UI/Xaml/Internal/PointerCaptureTarget.cs
```diff
@@ -0,0 +1,60 @@
...
+	/// </summary>
+	/// <remarks>
+	/// On WASM this might be different than the <see cref="Element"/>:
+	/// In case of implicit capture, the element used for the capture will prevent any pointer event on sub element
+	/// (sub element will actually get a pointer 'leave' on capture, and a 'enter' on capture release).
+	/// So instead of capturing using the actual element, we use the 'OriginalSource' of the 'relatedArgs',
+	/// so event will still be sent to sub elements and we will then filter them out if needed.
```
agneszitte, line 31:
```suggestion
	/// On WASM this might be different than the <see cref="Element"/>:
	/// In the case of implicit capture, the element used for the capture will prevent any pointer event on the sub-element
	/// (sub-element will actually get a pointer 'leave' on capture, and an 'enter' on capture release).
	/// So instead of capturing using the actual element, we use the 'OriginalSource' of the 'relatedArgs',
	/// so an event will still be sent to sub-elements and we will then filter them out if needed.
```

src/Uno.UI/UI/Xaml/Internal/PointerCaptureTarget.cs
```diff
@@ -0,0 +1,60 @@
...
+	/// </remarks>
+	public UIElement NativeCaptureElement { get; set; }
+
+	/// <summary>
+	/// Gets the current capture kinds that was enabled on the target
```
agneszitte, line 36:
```suggestion
	/// Gets the current capture kinds that were enabled on the target
```

src/Uno.UI/UI/Xaml/Internal/PointerCaptureTarget.cs
```diff
@@ -0,0 +1,60 @@
...
+	internal PointerCaptureKind Kind { get; set; }
+
+	/// <summary>
+	/// Determines if the <see cref="Element"/> is in the native bubbling tree.
+	/// If so we could rely on standard events bubbling to reach it.
+	/// Otherwise this means that we have to bubble the event in managed only.
+	///
+	/// This makes sense only for platform that has "implicit capture"
+	/// (i.e. all pointers events are sent to the element on which the pointer pressed
+	/// occured at the beginning of the gesture). This is the case on iOS and Android.
```
agneszitte, line 47:
```suggestion
	/// Determines if the <see cref="Element"/> is in the native bubbling tree.
	/// If so we could rely on standard events bubbling to reach it.
	/// Otherwise this means that we have to bubble the event in managed only.
	///
	/// This makes sense only for a platform that has "implicit capture"
	/// (i.e. all pointers events are sent to the element on which the pointer pressed
	/// occurred at the beginning of the gesture). This is the case on iOS and Android.
```

src/Uno.UI/UI/Xaml/Internal/PointerCaptureTarget.cs
```diff
@@ -0,0 +1,60 @@
...
+	/// </summary>
+	public bool? IsInNativeBubblingTree { get; set; }
+
+	/// <summary>
+	/// Gets the last event dispatched by the <see cref="Element"/>.
+	/// In case of native bubbling (cf. <see cref="IsInNativeBubblingTree"/>),
+	/// this helps to determine that an event was already dispatched by the Owner:
+	/// if a UIElement is receiving and event with the same timestamp, it means that the element
```
agneszitte, line 55:
```suggestion
	/// if a UIElement is receiving an event with the same timestamp, it means that the element
```

src/Uno.UI/UI/Xaml/Internal/RootVisual.cs
```diff
@@ -167,7 +167,8 @@ internal void ProcessPointerUp(PointerRoutedEventArgs args, bool isAfterHandledU
 			}
 
 #if __ANDROID__ || __IOS__ // Not needed on WASM as we do have native support of the exit event
-			// On Android and iOS we use the RootVisual to raise the UWP only exit event (in managed only)
+			// On iOS we use the RootVisual to raise the UWP only exit event (in managed only)
```
agneszitte, line 170:
```suggestion
			// On iOS we use the RootVisual to raise the UWP-only exit event (in managed only)
```

src/Uno.UI/UI/Xaml/UIElement.Pointers.cs
```diff
@@ -216,24 +217,55 @@ private void InitializePointers()
...
+			/// <summary>
+			/// Indicates that the visual tree has been modified in a way that the input manager must perform a complete hit testing sequence before dispatching a new event.
+			/// </summary>
+			/// <remarks>
+			/// This is designed for the case where for a single native pointer event, we are dispatching multiple managed events (e.g. managed Enter/Exit when we get only a native Move)
+			/// for all other cases **a full hit test must be perform**.
+			/// This means that we must not "capture"/cache the current top-most-element (a.k.a. OriginalSource) and try to update it on next event
+			/// as this flag does not take in consideration RenderTransform and other layout modification that does not alter the state of the pointer.
+			/// </remarks>
+			/// <remarks>This is used only for managed dispatch.</remarks>
```
agneszitte, line 265:
```suggestion
			/// Indicates that the visual tree has been modified in a way that the input manager must perform a complete hit-testing sequence before dispatching a new event.
			/// </summary>
			/// <remarks>
			/// This is designed for the case where for a single native pointer event, we are dispatching multiple managed events (e.g. managed Enter/Exit when we get only a native Move)
			/// for all other cases **a full hit test must be performed**.
			/// This means that we must not "capture"/cache the current top-most-element (a.k.a. OriginalSource) and try to update it on the next event
			/// as this flag does not take into consideration RenderTransform and other layout modification that does not alter the state of the pointer.
			/// </remarks>
			/// <remarks>This is used only for managed dispatch.</remarks>
```

src/Uno.UI/UI/Xaml/UIElement.Pointers.cs
```diff
@@ -216,24 +217,55 @@ private void InitializePointers()
...
+			public bool VisualTreeAltered { get; set; }
+		}
+
 		/// <summary>
 		/// Indicates if this element or one of its child might be target pointer pointer events.
 		/// Be aware this doesn't means that the element itself can be actually touched by user,
```
agneszitte, line 271:
```suggestion
		/// Indicates if this element or one of its children might be target pointer events.
		/// Be aware this doesn't means that the element itself can be actually touched by the user,
```

src/Uno.UI/UI/Xaml/UIElement.cs
```diff
@@ -910,6 +910,45 @@ public void InvalidateArrange()
 		}
 #endif
 
+		/// <summary>
+		/// This method has to be invoked for element that are going to be recycled WITHOUT necessarily being unloaded / loaded.
+		/// For instance, this is is not expected to be invoked for elements recycled by the the template pool as they are always unloaded.
+		/// The main use case is for ListView and is expected to be invoked by the ListView.CleanUpContainer.
```
agneszitte, line 916:
```suggestion
		/// This method has to be invoked for elements that are going to be recycled WITHOUT necessarily being unloaded / loaded.
		/// For instance, this is not expected to be invoked for elements recycled by the template pool as they are always unloaded.
		/// The main use case is for ListView and is expected to be invoked by ListView.CleanUpContainer.
```

src/Uno.UI/UI/Xaml/UIElement.cs
```diff
@@ -910,6 +910,45 @@ public void InvalidateArrange()
...
+				}
+			}
+		}
+
+		/// <summary>
+		/// This method has to be invoked on element that are going to be recycled WITHOUT necessarily being unloaded / loaded.
+		/// For instance, this is is not expected to be invoked for elements recycled by the the template pool as they are always unloaded.
+		/// The main use case is for ListView and is expected to be invoked by the ListView.CleanUpContainer.
+		/// </summary>
+		/// <remarks>This will walk the tree down to invoke this on all children!</remarks>
```
agneszitte, line 939:
```suggestion
		/// This method has to be invoked on elements that are going to be recycled WITHOUT necessarily being unloaded / loaded.
		/// For instance, this is not expected to be invoked for elements recycled by the template pool as they are always unloaded.
		/// The main use case is for ListView and is expected to be invoked by ListView.CleanUpContainer.
		/// </summary>
		/// <remarks>This will walk the tree down to invoke this on all children!</remarks>
```

jeromelaban (approved):

src/Uno.UI/UI/Xaml/Controls/ListViewBase/VirtualizingPanelLayout.Android.cs
```diff
@@ -1302,6 +1304,8 @@ private void UpdateLayout(GeneratorDirection direction, int availableExtent, int
 				AssertValidState();
 			}
 
+			XamlParent?.NativePanel.StopDetachedViewTrackingAndNotifyPendingAsRecycled();
```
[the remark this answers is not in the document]
dr1rrb: The worst case is a leak of few controls until the next measure phase so I estimated that it would not worth the cost of a try/finally ^^

PR #11973, opened 2023-04-14
fix: Adjust for missing props file

jeromelaban:
## PR Type

- Bugfix

## What is the new behavior?

Adjust for incorrect nuspec inclusion of the .targets props

### DIFF
--- a/build/Uno.WinUI.nuspec
+++ b/build/Uno.WinUI.nuspec
@@ -894,6 +894,21 @@
 	<file src="Uno.WinUI.targets" target="buildTransitive\net7.0-macos" />
 	<file src="Uno.WinUI.targets" target="buildTransitive\net5.0-windows" />
 
+	<file src="Uno.WinUI.props" target="buildTransitive\MonoAndroid" />
+	<file src="Uno.WinUI.props" target="buildTransitive\xamarinios10" />
+	<file src="Uno.WinUI.props" target="buildTransitive\xamarinmac20" />
+	<file src="Uno.WinUI.props" target="buildTransitive\netstandard2.0" />
+	<file src="Uno.WinUI.props" target="buildTransitive\net7.0" />
+	<file src="Uno.WinUI.props" target="buildTransitive\net6.0-android30.0" />
+	<file src="Uno.WinUI.props" target="buildTransitive\net6.0-ios" />
+	<file src="Uno.WinUI.props" target="buildTransitive\net6.0-maccatalyst" />
+	<file src="Uno.WinUI.props" target="buildTransitive\net6.0-macos" />
+	<file src="Uno.WinUI.props" target="buildTransitive\net7.0-android30.0" />
+	<file src="Uno.WinUI.props" target="buildTransitive\net7.0-ios" />
+	<file src="Uno.WinUI.props" target="buildTransitive\net7.0-maccatalyst" />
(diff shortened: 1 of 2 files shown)


build/Uno.WinUI.nuspec
```diff
@@ -894,6 +894,21 @@
...
+	<file src="Uno.WinUI.props" target="buildTransitive\net7.0" />
+	<file src="Uno.WinUI.props" target="buildTransitive\net6.0-android30.0" />
+	<file src="Uno.WinUI.props" target="buildTransitive\net6.0-ios" />
+	<file src="Uno.WinUI.props" target="buildTransitive\net6.0-maccatalyst" />
+	<file src="Uno.WinUI.props" target="buildTransitive\net6.0-macos" />
+	<file src="Uno.WinUI.props" target="buildTransitive\net7.0-android30.0" />
+	<file src="Uno.WinUI.props" target="buildTransitive\net7.0-ios" />
+	<file src="Uno.WinUI.props" target="buildTransitive\net7.0-maccatalyst" />
+	<file src="Uno.WinUI.props" target="buildTransitive\net7.0-macos" />
+	<file src="Uno.WinUI.props" target="buildTransitive\net5.0-windows" />
```
Youssef1313, line 910: Is it possible to only include it in the top-level `buildTransitive`? Would NuGet pick it up as "applicable for all TFMs"?

build/Uno.WinUI.nuspec
```diff
@@ -894,6 +894,21 @@
 	<file src="Uno.WinUI.targets" target="buildTransitive\net7.0-macos" />
 	<file src="Uno.WinUI.targets" target="buildTransitive\net5.0-windows" />
 
+	<file src="Uno.WinUI.props" target="buildTransitive\MonoAndroid" />
```
Youssef1313, line 897: The only thing this props file does is:

https://github.com/unoplatform/uno/blob/d077eaf8b9337c55c0fd9e912e65f99088a3ab9d/build/uno.winui.common.props#L4-L10

1. `UnoForceSingleTFM` doesn't seem to be set to true anywhere, so this effectively does nothing
2. There is no `uno.vscode.targets`

I think we can just delete the props file as dead.

build/Uno.WinUI.nuspec
jeromelaban: Yes, but not in this PR. We'll adjust that for net8.

build/Uno.WinUI.nuspec
jeromelaban: This file is now required by the vscode extension, both the property and the props file are defined there as well. (the name was incorrect, I fixed it).

build/uno.winui.common.props
```diff
@@ -2,10 +2,10 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
 	<!-- Workaround for OmniSharp single-TargetFramework use -->
-	<Import Project="$([MSBuild]::GetPathOfFileAbove('uno.vscode.targets', '$(MSBuildThisFileDirectory)../.vscode/'))" 
+	<Import Project="$([MSBuild]::GetPathOfFileAbove('uno.vscode.props', '$(MSBuildThisFileDirectory)../.vscode/'))"
```
Youssef1313, line 5: I don't see where `uno.vscode.props` is. Can you link please?

spouliot (approved):

Issue #11976, opened 2023-04-14
Revise `UNO_HAS_UIELEMENT_IMPLICIT_PINNING` usage in XAML generator


Youssef1313:
@jeromelaban Something is wrong here.

`UNO_HAS_UIELEMENT_IMPLICIT_PINNING` is only defined internally (in Uno.UI solution).

In user apps, we will always be hitting the `#else` part of the code.

I don't have enough background here, but if usage of WeakReference is required on some platforms, maybe we should propagate `UNO_HAS_UIELEMENT_IMPLICIT_PINNING` to user apps?

if it's not required anymore, then we should cleanup the generated code.

_Originally posted by @Youssef1313 in https://github.com/unoplatform/uno/pull/11854#discussion_r1155102346_

Youssef1313:
Original fix was in https://github.com/unoplatform/uno/pull/3936

and it was reverted (possibly unintentionally) https://github.com/unoplatform/uno/pull/3974/files#diff-70bcc603cf98904b7de6118e7f166581de45b6911245b3ac7ba397096c1923b0L29

PR #3936, opened 2020-09-02
fix(xbind): [iOS/macOS] Fix memory leak when using x:Bind

jeromelaban:
GitHub Issue (If applicable): fixes https://github.com/unoplatform/uno/issues/3914

## PR Type

What kind of change does this PR introduce?
- Bugfix

## What is the new behavior?

Using x:Bind on iOS/macOS will not keep UIElement instances alive.

## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)
- [x] [Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [ ] Validated PR `Screenshots Compare Test Run` results.
- [x] Contains **NO** breaking changes
- [x] Associated with an issue (GitHub or internal) and uses the [automatic close keywords](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue).
- [x ] Commits must be following the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) specification.

## Other information

Internal Issue (If applicable):

PR #12038, opened 2023-04-22
chore: leak test

Youssef1313:
Related to #11976. Let's see if any test will fail.

## PR Type

## PR Checklist

## Other information


Youssef1313 (commented):
@jeromelaban I was expecting that the test you added in https://github.com/unoplatform/uno/pull/3936 will fail with this change, but it didn't.

jeromelaban:
If it's not failing, there's a missing test. Apple platforms have implicit pinning of native-inheriting references, causing a circular reference that the .NET GC is not able to handle (and neither does Apple's ref counting).

Youssef1313:
@jeromelaban Looking into this again, the circular reference here is between one type that is UIElement (NSObject), and one type that is just System.Object (no native counterpart).

So, I think as long as the circular reference is not at native side, we should be good and not leaking.

Youssef1313:
@jeromelaban It also appears that there is a test added with the initial introduction of this specific code:

https://github.com/unoplatform/uno/blob/0effad58fd8a6c6e75380f0496c779f654275fc1/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_FrameworkElement_And_Leak.cs#L45-L46

PR #12056, opened 2023-04-24
Ensure callbacks added in callback handlers are not processed in the same run

MartinZikmund:

## PR Type

- Bugfix

In `DependencyObjectStore` we are going through callbacks via `for`. This can be a problem if any of the handlers, e.g. a callback modifies the list while it is being processed - as it can remove itself (moving the next callback, so it is then accidentally skipped), or add a new one which will then be processed immediately - causing potentially stack overflow.

## What is the new behavior?

Use a local copy of the list instead.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/Mock/Border.net.cs
+++ b/src/Uno.UI/Mock/Border.net.cs
@@ -18,7 +18,7 @@ namespace Windows.UI.Xaml.Controls
 				RemoveChild(previousValue);
 			}
 
-			if (newValue != null)
+			if (newValue is not null)
 			{
 				AddChild(newValue);
 			}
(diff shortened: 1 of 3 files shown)


src/Uno.UI/Mock/Border.net.cs
```diff
@@ -18,7 +18,10 @@ public override IEnumerable<UIElement> GetChildren()
 				RemoveChild(previousValue);
 			}
 
-			AddChild(newValue);
+			if (newValue != null)
```
jeromelaban, line 21:
```suggestion
			if (newValue is not null)
```

jeromelaban (approved):

PR #12066, opened 2023-04-25
fix: Ensure Commands are executed when RadioButton is toggled through AutomationPeer

### LINKED ISSUE #10557: [iOS] ButtonBase commands not working with VoiceOver turned on
### Current behavior

When voiceover is turned on on an iOS device, the Command on a RadioButton control will not execute.

### Expected behavior

When voiceover is turned on on an iOS device, the Command on a RadioButton control would execute.

- Create Uno app or use UnoAppTemplate
- Anywhere, put a RadioButton control and bind to it's Command property a piece of code
- Turn on voiceOver on your device
- In the app, focus the radio button and double-tap it to select it
- Notice the command is not executed 

### Workaround

None found

### Works on UWP/WinUI

_No response_

### Environment

### NuGet package version(s)

Uno.UI 4.4.20

### Affected platforms

iOS

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

There was[ a similar issue](https://github.com/unoplatform/nventive-private/issues/231) created about a year ago where a fix had been merged into Master, but looking into it now it seems like the fix is gone and the issue is still occuring.
[UnoApplicationTemplate-main.zip](https://github.com/unoplatform/uno/files/10085349/UnoApplicationTemplate-main.zip)

kazo0:
closes #10557 

## PR Type

- Bugfix

### DIFF
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_RadioButton.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_RadioButton.cs
@@ -385,10 +385,9 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 		}
 	}
 
-#pragma warning disable CS0067
 	public class TestCommand : ICommand
 	{
-		public event EventHandler CanExecuteChanged;
+		public event EventHandler CanExecuteChanged { add { } remove { } }
 		public event EventHandler CommandFired;
 
 		public TestCommand()
@@ -402,5 +401,4 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 
 		public void Execute(object o) => CommandFired.Invoke(this, null);
 	}
-#pragma warning restore CS0067
 }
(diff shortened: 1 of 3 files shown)


src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_RadioButton.cs
```diff
@@ -341,5 +345,62 @@ public async Task When_GroupName_Default_Two_Containers()
...
+		}
+
+		public bool CanExecute(object o)
+		{
+			return true;
+		}
+
+		public void Execute(object o) => CommandFired.Invoke(this, null);
 	}
+#pragma warning restore CS0067
```
Youssef1313, line 405:
```suggestion
	public class TestCommand : ICommand
	{
		public event EventHandler CanExecuteChanged { add { } remove { } }
		public event EventHandler CommandFired;

		public TestCommand()
		{
		}

		public bool CanExecute(object o)
		{
			return true;
		}

		public void Execute(object o) => CommandFired.Invoke(this, null);
	}
```

agneszitte (approved):

PR #12084, opened 2023-04-26
feat: Remove the Edge Visual Search from WASM

### LINKED ISSUE #11783: Make it possible to remove Edge's Visual Search overlay on images

https://stackoverflow.com/questions/71731966/how-do-i-prevent-edge-from-showing-the-visual-search-icon-when-users-hover-over

## Why is this needed:

It can break the design, e.g. having buttons with an image as content

## For which Platform:

- [x] WebAssembly
- [ ] Android
- [ ] iOS
- [ ] macOS (AppKit)
- [ ] Mac Catalyst
- [ ] Skia
  - [ ] WPF
  - [ ] GTK (Linux)
  - [ ] Linux Framebuffer
  - [ ] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

Darsh0307:
GitHub Issue (If applicable): closes https://github.com/unoplatform/uno/issues/11783

## PR Type

 Code style update (formatting)

It is showing us a Visual Search in the edge.

## What is the new behavior?

Now there will be no more visual search on the wasm

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/Image/HtmlImage.wasm.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Image/HtmlImage.wasm.cs
@@ -8,5 +8,7 @@ public class HtmlImage : UIElement
 	{
 		// Avoid the "drag effect" which is set by default in browsers
 		SetAttribute("draggable", "false");
+		HitTestVisibility = HitTestability.Invisible;
+		IsHitTestVisible = false;
 	}
 }
--- a/src/Uno.UI/WasmCSS/Uno.UI.css
+++ b/src/Uno.UI/WasmCSS/Uno.UI.css
@@ -174,6 +174,7 @@ embed.uno-frameworkelement.uno-unarranged {
   user-select: none;
   -webkit-user-drag: none;
   user-drag: none;
+  pointer-events: none;
 }
 
 .uno-scrollcontentpresenter {
--- a/src/Uno.UI/WasmCSS/Uno.UI.css
+++ b/src/Uno.UI/WasmCSS/Uno.UI.css
@@ -174,7 +174,6 @@ embed.uno-frameworkelement.uno-unarranged {
   user-select: none;
   -webkit-user-drag: none;
   user-drag: none;
-  pointer-events: none;
 }
 
 .uno-scrollcontentpresenter {


src/Uno.UI/WasmCSS/Uno.UI.css
```diff
@@ -174,6 +174,7 @@ embed.uno-frameworkelement.uno-unarranged {
   user-select: none;
   -webkit-user-drag: none;
   user-drag: none;
+  pointer-events: none;
```
Youssef1313, line 177: This is already applied to `.uno-uielement`:

https://github.com/unoplatform/uno/blob/337801acc17433fb31195ed925ef922a6c16a376/src/Uno.UI/WasmCSS/Uno.UI.css#L28-L35

Shouldn't HtmlImage already be a UIElement and already have `pointer-events: none`? Am I missing anything?

MartinZikmund (approved):

src/Uno.UI/UI/Xaml/Controls/Image/HtmlImage.wasm.cs
```diff
@@ -8,5 +8,7 @@ public HtmlImage() : base("img")
 	{
 		// Avoid the "drag effect" which is set by default in browsers
 		SetAttribute("draggable", "false");
+		HitTestVisibility = HitTestability.Invisible;
+		IsHitTestVisible = false;
```
Youssef1313, line 12: @dr1rrb Any pointer-related concerns?

Youssef1313: I don't know if we have such a test, but best to have a test for `Tapped` event specifically for `Image`

MartinZikmund: I did test the pointer events specifically, but yes, will also verify with a test

dr1rrb: Yes I think that this will break the `Tapped` event and all other pointer events, including manipulation events (which is commonly use to zoom on images for instance).

I tried on my computer, setting the `pointer-events: none;` doesn't seems to prevent that menu. I pretty sure that tehre a better way to fix it, for instance I noticed that images in bing.com does not have that menu and pointer events are not disabled. (Removed all layer above the images and still no menu, without having bing in the exemption list of my browser).

BTW changing the `IsHitTestVisible` will impact the "public API" of the image, the `HitTestVisibility = HitTestability.Invisible;` should be sufficient and won't impact application code.

dr1rrb (changes requested):
This will probably break pointer events, we need to add test for that ... and probably find a better approach.

src/Uno.UI/UI/Xaml/Controls/Image/HtmlImage.wasm.cs
MartinZikmund: @dr1rrb I have tested the `Tapped` event, and having it on `Image` worked (the hit test changes are on `HtmlImage`). Overall though, the menu does not show reliably and it seems the browser has some non-specific exemptions to it. However, using `pointer-events: none` seems to be the only reliable way to hide it always (and that is what people on Stack Overflow ended up with too)

Regarding `IsHitTestVisible` - the change is on `HtmlImage` only, so it shouldn't matter much, no? As this is internal implementation detail of `Image` on WASM anyway?

PR #12108, opened 2023-04-28
chore: Reduce file name length to avoid risk of too long filename on checkout

dr1rrb:
## Other
Reduce file name length to avoid risk of too long filename on checkout

Cannot checkout on windows unless in a very short path or having in `.gitconfig` :

```
[core]
    longpaths = true
```

## What is the new behavior?
Reduce the risk of issue

## PR Checklist


src/SourceGenerators/Uno.UI.SourceGenerators.netcore.Tests/XamlCodeGeneratorTests/Given_MvvmGeneratedMembers.cs
```diff
@@ -120,7 +119,7 @@ public async Task When_ObservableProperty_AttributeDoesNotExists()
 			</Page>
 			""");
 
-		var test = new Test(xamlFile)
+		var test = new MvvmTest(xamlFile, "WOPADNE")
```
Youssef1313, line 122: Not a fan tbh :'(

Youssef1313 (approved):
Approving, though the unreadable short names are unfortunate.

Youssef1313 (commented):
@dr1rrb Can we instead go from full hash to only the first 5 characters or so?

That would be done here:

https://github.com/unoplatform/uno/blob/08e81f34ec1cc503d46320fb6c6b5fd13ae9444e/src/SourceGenerators/SourceGeneratorHelpers/Helpers/HashBuilder.cs#L28-L47

I'd prefer that more than the current approach.

- Hash 32 chars -> 5 chars: Saves 27 characters
- `TestOutput` -> `Out`: 7 characters save

dr1rrb:
> Approving, though the unreadable short names are unfortunate.

Agree, it's the best approach I found to quickly and significantly reduce the file path.

> @dr1rrb Can we instead go from full hash to only the first 5 characters or so?
> 

We could, but it would require a bigger refactor and tests, while this PR is just files renaming in a test project for a quick patch.

global.json
```diff
@@ -1,12 +1,4 @@
 {
-  "sdk": {
-    "version": "7.0.102",
-    "allowPrerelease": true,
-    "rollForward": "latestPatch"
-  },
-  "tools": {
-    "dotnet": "7.0.102"
-  },
```
Youssef1313, line 9: These shouldn't be removed.

dr1rrb: Oups, fixed (even if fixing the specific version is incredibly annoying, has to change it on each checkout :/)

Youssef1313:
> We could, but it would require a bigger refactor and tests, while this PR is just files renaming in a test project for a quick patch.

I thought it should be a smaller change?

dr1rrb:
> > We could, but it would require a bigger refactor and tests, while this PR is just files renaming in a test project for a quick patch.
> 
> I thought it should be a smaller change?

Smaller change, but bigger impact

PR #12309, opened 2023-05-15
fix: [Android] fix android transformtovisual

kazo0:
## PR Type

- Bugfix

TransformToVisual was returning results with improper xy coords


Youssef1313 (commented):
Could we have a test?

kazo0:
> Could we have a test?

@Youssef1313 yeah working on it now, having trouble reproducing without using the android system status bar as an offset, @dr1rrb can you think of an easy repro for this? Right now I have a test with a visual structur of Grid > LinearLayout (with native padding) > Border and I am trying to do Border.TransformToVisual(null).TransformPoint(0, 0) but it is still giving me the proper values. Might be misunderstanding the issue

kazo0:
@dr1rrb @Youssef1313 

ok test added 🖖

Youssef1313 (approved):

PR #12321, opened 2023-05-17
fix: Read PrimaryLanguageOverride from an app-specific key

### LINKED ISSUE #12320: PrimaryLanguageOverride setting shared between skia applications
### Current behavior

Create two different applications
Set PrimaryLanguageOverride in one application
Run other application and not PrimaryLanguageOverride is also set.

### Expected behavior

Applications maintain separate values for PrimaryLanguageOverride

_No response_

### Workaround

_No response_

### Works on UWP/WinUI

None

### Environment

_No response_

### NuGet package version(s)

_No response_

### Affected platforms

_No response_

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

Currently skia applications use a common location for all settings. 
This will be resolved by [this PR](https://github.com/unoplatform/uno/pull/12314) but as this is breaking changes, will be 5.0 timeline
Fix for PrimaryLanguageOverride would be to use an app specific key for looking up the settings value.
Furthermore this would also provide separation for local debugging wasm applications. Currently if you create and debug two wasm applications that happen to use the same port (eg https://localhost:5000), they will access the same settings and thus the same value for PrimaryLanguageOverride. Even in 5.0 timeline when the fix for skia appdata is merged, we should keep the app specific key for PrimaryLanguageOverride to avoid confusion when debugging wasm applications.

Youssef1313:
GitHub Issue (If applicable): closes #12320

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Application.cs
+++ b/src/Uno.UI/UI/Xaml/Application.cs
@@ -231,8 +231,8 @@ namespace Windows.UI.Xaml
 
 		public static void Start(global::Windows.UI.Xaml.ApplicationInitializationCallback callback)
 		{
-			ApplicationLanguages.ApplyCulture();
 			StartPartial(callback);
+			ApplicationLanguages.ApplyCulture();
 		}
 
 		static partial void StartPartial(ApplicationInitializationCallback callback);
--- a/src/Uno.UWP/Globalization/ApplicationLanguages.cs
+++ b/src/Uno.UWP/Globalization/ApplicationLanguages.cs
@@ -82,7 +82,7 @@ namespace Windows.Globalization
 				throw new InvalidOperationException("ApplicationLanguages is being accessed too early before an instance of Application was created.");
 			}
 
-			return $"PrimaryLanguageOverrideSettingKey.{entryAssembly.GetName().Name}";
+			return $"{PrimaryLanguageOverrideSettingKey}.{entryAssembly.GetName().Name}";
 		}
 #endif
 
--- a/src/Uno.UWP/Globalization/ApplicationLanguages.cs
+++ b/src/Uno.UWP/Globalization/ApplicationLanguages.cs
@@ -62,6 +62,7 @@ namespace Windows.Globalization
 			ApplyLanguages();
 		}
 
+#if !NET461
 		private static bool TryGetSettingFromKey(string key, out string value)
 		{
 			if (ApplicationData.Current.LocalSettings.Values.TryGetValue(key, out var savedValue))
@@ -83,6 +84,7 @@ namespace Windows.Globalization
 
 			return $"PrimaryLanguageOverrideSettingKey.{entryAssembly.GetName().Name}";
 		}
+#endif
 
 		internal static void ApplyCulture()
 		{
--- a/src/Uno.UWP/Globalization/ApplicationLanguages.cs
+++ b/src/Uno.UWP/Globalization/ApplicationLanguages.cs
@@ -76,7 +76,12 @@ namespace Windows.Globalization
 
 		private static string GetAppSpecificSettingKey()
 		{
-			return $"PrimaryLanguageOverrideSettingKey.{Package.EntryAssembly.GetName().Name}";
+			if (Package.EntryAssembly is not { } entryAssembly)
+			{
+				throw new InvalidOperationException("ApplicationLanguages is being accessed too early before an instance of Application was created.");
+			}
+
+			return $"PrimaryLanguageOverrideSettingKey.{entryAssembly.GetName().Name}";
 		}
 
 		internal static void ApplyCulture()
--- a/src/Uno.UWP/Globalization/ApplicationLanguages.cs
+++ b/src/Uno.UWP/Globalization/ApplicationLanguages.cs
@@ -75,7 +75,7 @@ namespace Windows.Globalization
 			return false;
 		}
 
-		private static string GetAppSpecificSettingKey()
+		internal static string GetAppSpecificSettingKey()
 		{
 			if (Package.EntryAssembly is not { } entryAssembly)
 			{
--- a/src/Uno.UWP/Globalization/ApplicationLanguages.cs
+++ b/src/Uno.UWP/Globalization/ApplicationLanguages.cs
@@ -4,8 +4,8 @@ using System.Linq;
 using System.Text.RegularExpressions;
 using Windows.Storage;
 using Uno.Foundation.Logging;
-using Uno.UI;
 using Uno;
+using Windows.ApplicationModel;
 
 namespace Windows.Globalization
 {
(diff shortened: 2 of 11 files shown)


src/Uno.UI/UI/Xaml/Application.cs
```diff
@@ -231,8 +231,8 @@ public void Exit()
 
 		public static void Start(global::Windows.UI.Xaml.ApplicationInitializationCallback callback)
 		{
-			ApplicationLanguages.ApplyCulture();
 			StartPartial(callback);
```
Youssef1313, line 234: About the ordering change:

`StartPartial` is what creates an `Application` instance. So we want that to run first so that we can find the EntryAssembly set in Application's constructor.

src/Uno.UWP/Globalization/ApplicationLanguages.cs
```diff
@@ -20,16 +20,64 @@ public static partial class ApplicationLanguages
...
+				return value is not null;
+			}
+
+			value = null;
+			return false;
+		}
+
+		private static string GetAppSpecificSettingKey()
+		{
+			return $"PrimaryLanguageOverrideSettingKey.{Package.EntryAssembly.GetName().Name}";
```
Youssef1313, line 78: If someone tried to access ApplicationLanguages too early, before an Application instance is created. This is going to blow up.

@MartinZikmund Thoughts? Is it a valid use case to access ApplicationLanguages before Application instance is created? Not sure how to best approach this.

MartinZikmund: Technically it should never happen - I imagine the earliest usage in the `App` constructor, which is after `Current` was already set in the `Application` base constructor. I would suggest throwing a clear exception here (something like you are trying to use `ApplicationLanguages` too early in tyhe application lifecycle).

MartinZikmund (approved):

MartinZikmund:
I see that the lifecycle changes you are hitting here are the same ones I am seeing in my PR, so I will postpone my PR until this one is merged and also included in breaking changes branch, as there will be conflicts.

src/Uno.UWP/Globalization/ApplicationLanguages.cs
Youssef1313: @MartinZikmund One could attempt to access it before this line, for example:

https://github.com/unoplatform/uno.templates/blob/de24edc2b0f8cbbd14a988fb5d93ed21b0a23288/src/Uno.Templates/content/unoapp/MyExtensionsApp._1.Skia.Gtk/Program.cs#L19

I'll throw an exception.

MartinZikmund: Actually, based on my tests, it is possible to use even the `Package.Current` API in WinUI before the app starts:

```csharp
static void Main(string[] args)
{
    XamlCheckProcessRequirements();

    global::WinRT.ComWrappersSupport.InitializeComWrappers();
    var name = Package.Current.Id.Name;
    global::Microsoft.UI.Xaml.Application.Start((p) => {
        var context = new global::Microsoft.UI.Dispatching.DispatcherQueueSynchronizationContext(global::Microsoft.UI.Dispatching.DispatcherQueue.GetForCurrentThread());
        global::System.Threading.SynchronizationContext.SetSynchronizationContext(context);
        new App();
    });
}
```

So I would vote for making the `SetEntryAssembly` part of the host initialization sequence somehow. Let's discuss later

MartinZikmund (approved):

src/Uno.UWP/Globalization/ApplicationLanguages.cs
```diff
@@ -20,16 +21,71 @@ public static partial class ApplicationLanguages
...
+		}
+
+		private static string GetAppSpecificSettingKey()
+		{
+			if (Package.EntryAssembly is not { } entryAssembly)
+			{
+				throw new InvalidOperationException("ApplicationLanguages is being accessed too early before an instance of Application was created.");
+			}
+
+			return $"PrimaryLanguageOverrideSettingKey.{entryAssembly.GetName().Name}";
```
jeromelaban, line 85: I don't believe this should be the fix. The origin of the issue is likely because we're not storing app settings in a app identity based location, but rather a common location that is incorrectly shared.

nickrandolph: Right but how do we fix this in 4.x without it being a breaking change (ie what @MartinZikmund is working on).
My additional comment would be that this fix would limit overlap of wasm apps when debugging - currently if you use the same port (the template defaults to a fixed post to allow ui tests to "just work") you'll see PrimaryLanguageOverride from other apps you debugged with the same port. Adding the app name would limit this overlap

jeromelaban: Those are two different issues then. If this fix is for wasm, we're talking about a debug issue (which I don't believe needs fixing), not a production issue. The issue for Skia is different, where settings are placed at the wrong location.

In general linking the settings to the assembly name will cause issues, as if an app is logically deployed at the same location, but somehow changes its internal structure, the settings will be lost.

src/Uno.UWP/Globalization/ApplicationLanguages.cs
Youssef1313:
```
Failed: VerifyMultipleEraCalendar() [System.TypeInitializationException] 
 System.TypeInitializationException: TypeInitialization_Type, Windows.UI.Xaml.Controls.CalendarView
 ---> System.TypeInitializationException: TypeInitialization_Type, Windows.Globalization.ApplicationLanguages
 ---> System.InvalidOperationException: ApplicationLanguages is being accessed too early before an instance of Application was created.
   at Windows.Globalization.ApplicationLanguages.GetAppSpecificSettingKey()
   at Windows.Globalization.ApplicationLanguages..cctor()
   Exception_EndOfInnerExceptionStack
   at Windows.Globalization.Calendar..ctor()
   at Windows.UI.Xaml.Controls.CalendarView.GetOrCreateGregorianCalendar()
   at Windows.UI.Xaml.Controls.CalendarView.GetDefaultMaxDate()
   at Windows.UI.Xaml.Controls.CalendarView..cctor()
   Exception_EndOfInnerExceptionStack
   at Private.Infrastructure.CalendarHelper.CalendarViewHelper.<GetCalendarView>b__1_0()
   at Private.Infrastructure.TestServices.RunOnUIThread(Action action)
   at Private.Infrastructure.CalendarHelper.CalendarViewHelper.GetCalendarView()
   at Windows.UI.Xaml.Tests.Enterprise.CalendarViewIntegrationTests.VerifyMultipleEraCalendar()
   at Private.Infrastructure.CalendarHelper.CalendarViewHelper.<GetCalendarView>b__1_0()
   at Private.Infrastructure.TestServices.RunOnUIThread(Action action)
   at Private.Infrastructure.CalendarHelper.CalendarViewHelper.GetCalendarView()
   at Windows.UI.Xaml.Tests.Enterprise.CalendarViewIntegrationTests.VerifyMultipleEraCalendar()
   at Private.Infrastructure.CalendarHelper.CalendarViewHelper.<GetCalendarView>b__1_0()
   at Private.Infrastructure.TestServices.RunOnUIThread(Action action)
   at Private.Infrastructure.CalendarHelper.CalendarViewHelper.GetCalendarView()
   at Windows.UI.Xaml.Tests.Enterprise.CalendarViewIntegrationTests.VerifyMultipleEraCalendar()
   at Uno.UI.Samples.Tests.UnitTestsControl.<>c__DisplayClass63_1.<<ExecuteTestsForInstance>g__InvokeTestMethod|2>d.MoveNext()
```

😕

Youssef1313: I wasn't able to figure out what is happening here. It is Wasm-specific

src/Uno.UWP/Globalization/ApplicationLanguages.cs
jeromelaban: That's indeed interesting. Per [the documentation](https://www.w3schools.com/html/html5_webstorage.asp): _Web storage is per origin (per domain and protocol). All pages, from one origin, can store and access the same data._

I don't think that we should be discriminating based on the assembly name, though. We're currently not using the app manifest for wasm, but it feels like we should start doing so, so we can pick up the app identity and use it as a based for storing settings.

This is likely something we should be explaining in the storage docs as well, as files from one app may interfere in another app from the same origin.

/cc @MartinZikmund

Youssef1313:
Closing as the linked issue got closed

PR #12428, opened 2023-05-26
fix(skia): Fix element not clipped when render transformed

dr1rrb:
## Bugfix
Element not clipped when render transformed

When you apply a `RenderTransform` on an element, the clipping is applied in transformed coordinate space (a.k.a. "element's rendering coordinate space") resulting in an invalidate rendering.

## What is the new behavior?
The clipping is now applied in the "element's XAML coordinate space".

## PR Checklist

### DIFF
--- a/src/Uno.UI.Composition/Composition/Compositor.skia.cs
+++ b/src/Uno.UI.Composition/Composition/Compositor.skia.cs
@@ -81,56 +81,46 @@ namespace Windows.UI.Composition
 
 		internal void RenderVisual(SKSurface surface, Visual visual)
 		{
-			if (visual.Opacity != 0 && visual.IsVisible)
+			if (visual is { Opacity: 0 } or { IsVisible: false })
 			{
-				if (visual.ShadowState is { } shadow)
-				{
-					surface.Canvas.SaveLayer(shadow.Paint);
-				}
-				else
-				{
-					surface.Canvas.Save();
-				}
-
-				var visualMatrix = surface.Canvas.TotalMatrix;
-
-				visualMatrix = visualMatrix.PreConcat(SKMatrix.CreateTranslation(visual.Offset.X, visual.Offset.Y));
-				visualMatrix = visualMatrix.PreConcat(SKMatrix.CreateTranslation(visual.AnchorPoint.X, visual.AnchorPoint.Y));
-
-				if (visual.RotationAngleInDegrees != 0)
-				{
-					visualMatrix = visualMatrix.PreConcat(SKMatrix.CreateRotationDegrees(visual.RotationAngleInDegrees, visual.CenterPoint.X, visual.CenterPoint.Y));
-				}
-
-				if (visual.TransformMatrix != Matrix4x4.Identity)
-				{
-					visualMatrix = visualMatrix.PreConcat(visual.TransformMatrix.ToSKMatrix44().Matrix);
-				}
-
-				surface.Canvas.SetMatrix(visualMatrix);
-
-				ApplyClip(surface, visual);
-
-				using var opacityDisposable = PushOpacity(visual.Opacity);
-
-				visual.Render(surface);
-
-
-				surface.Canvas.Restore();
+				return;
 			}
+
+			if (visual.ShadowState is { } shadow)
+			{
+				surface.Canvas.SaveLayer(shadow.Paint);
+			}
+			else
+			{
+				surface.Canvas.Save();
+			}
+
+			// Set the position of the visual on the canvas (i.e. change coordinates system to the "XAML element" one)
+			surface.Canvas.Translate(visual.Offset.X + visual.AnchorPoint.X, visual.Offset.Y + visual.AnchorPoint.Y);
+
+			// Apply the clipping. This is relative to the visual's coordinate system, before any rendering transformation is applied.
+			ApplyClip(surface, visual);
+
+			// Applied rending transformation matrix (i.e. change coordinates system to the "rendering" one)
+			var transform = GetTransform(visual); 
+			if (!transform.IsIdentity)
+			{
+				var skTransform = transform.ToSKMatrix();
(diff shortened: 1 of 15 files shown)


jeromelaban (approved):

ajpinedam (approved):
🚀

src/Uno.UI.Composition/Composition/InsetClip.skia.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace Windows.UI.Composition;
+
+partial class InsetClip
+{
+	internal SKRect SKRect => new()
+	{
+		Top = TopInset - 1,
+		Bottom = BottomInset + 1,
+		Left = LeftInset - 1,
+		Right = RightInset + 1
```
carldebilly, line 15: Those `1` and `-1` are very strange.

src/Uno.UI.Composition/Composition/Compositor.skia.cs
```diff
@@ -81,56 +81,46 @@ internal void Render(SKSurface surface)
 
 		internal void RenderVisual(SKSurface surface, Visual visual)
 		{
-			if (visual.Opacity != 0 && visual.IsVisible)
+			if (visual is { Opacity: 0 } or { IsVisible: false })
```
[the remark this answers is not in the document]
carldebilly: @Youssef1313 
`¬(A ∧ B) ≡ (¬A ∨ ¬B)` / `¬(A ∨ B) ≡ (¬A ∧ ¬B)` :-)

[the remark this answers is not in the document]
Youssef1313: @carldebilly Yeah, I mean I won't use pattern matching :)
I don't see improved readability with `is` and `or` and braces compared with `visual.Opacity == 0 || !visual.IsVisible`. Also the latter doesn't have unnecessary null checks.

src/Uno.UWP/UI/Input/GestureRecognizer.Manipulation.cs
```diff
@@ -705,7 +705,7 @@ private void UpdateComputedValues()
 						var p2 = _pointer2.RawPosition;
 
 						Center = new Point((p1.X + p2.X) / 2, (p1.Y + p2.Y) / 2);
-						Distance = Vector2.Distance(p1.ToVector(), p2.ToVector());
```
jeromelaban, line 708: Do we have pointer tests to account for this change?

src/Uno.UI.Composition/Composition/InsetClip.skia.cs
dr1rrb: Indeed ... this is just an extract of what was previously done directly in the `Compositor`, so I didn't review it and will not in the current PR.

src/Uno.UI.Composition/Composition/Compositor.skia.cs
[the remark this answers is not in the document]
dr1rrb: I personally find it more readable with pattern matching, especially since it's only "positive" conditions. It indeed adds a null-check, but it will only make the code more robust, the cost of it is negligible.

src/Uno.UWP/UI/Input/GestureRecognizer.Manipulation.cs
dr1rrb: Yes we do have (Unit) tests for the `GestureRecognizer`. This is however just a rename of the method to remove duplicate.

src/Uno.UI/UI/Xaml/Controls/TextBlock/TextVisual.skia.cs
```diff
@@ -12,23 +12,27 @@
...
+using Uno.UI.Composition;
 
 #nullable enable
 
 namespace Windows.UI.Composition
 {
 	internal class TextVisual : Visual
 	{
-		private readonly TextBlock _owner;
+		private readonly WeakReference<TextBlock> _owner;
```
Youssef1313, line 23: Shouldn't `TextVisual` and `TextBlock` has the same life-time?

If I understand correctly, the need for `WeakReference` indicates that `TextVisual` lives longer than `TextBlock`, and we want the `TextBlock` to be GC'ed. But how can `TextVisual` live longer than `TextBlock`?

dr1rrb: It does not really lives longer : one of the parents of the `TextVisual` will be un-referenced so it's going to be collected at some point. But as now the `VisualCollection` properly sets `Visual.Parent` (and clear it on remove) we have 2 circular references: between visual and its parent and between the `TextBlock` and its `TextVisual`. The GC seems to not be able to resolve them.

carldebilly (approved):

MartinZikmund (approved):

src/Uno.UI.Composition/Composition/VisualCollection.cs
```diff
@@ -30,7 +25,7 @@ public void InsertAbove(Visual newChild, Visual sibling)
 		{
 			var index = _visuals.IndexOf(sibling);
 			_visuals.Insert(index, newChild);
-
+			newChild.Parent = _owner;
```
jeromelaban, line 28: I wonder if we should be enhancing our leak testing to validate for lingering visuals (as we're making a bi-directional link with this change).

dr1rrb: Our leak testing has already catch an issue with this. We could definitely enhance tests for this, but apparently it's already robust enough to catch leak on visuals.

MartinZikmund:
@dr1rrb should this be backported to 4x?

dr1rrb:
> @dr1rrb should this be backported to 4x?

No, impossible :) And anyway we need to give us time to validate those core changes internally before.

PR #9355, opened 2022-07-25
fix: AutoSuggestBox.QuerySubmitted ChosenSuggestion should be null when selection does not match

aesteves900:
GitHub Issue (If applicable): closes #

## PR Type

What kind of change does this PR introduce?

## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)
- [ ] [Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [ ] Validated PR `Screenshots Compare Test Run` results.
- [ ] Contains **NO** breaking changes
- [ ] Associated with an issue (GitHub or internal) and uses the [automatic close keywords](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue).
- [ ] Commits must be following the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) specification.

## Other information

Internal Issue (If applicable):

PR #1069, opened 2019-06-13
AutoSuggestBox works on Android

davidjohnoliver:
- fix hangs, crashes
 - ensure user can keep typing after suggestions appear. This is still somewhat limited relative to UWP, eg trying to change the position of the cursor will close the suggestions dropdown.

GitHub Issue (If applicable): fixes #1012 

## PR Type
What kind of change does this PR introduce?

Bugfix

## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Tested code with current [supported SDKs](../README.md#supported)
- [ ] Docs have been added/updated which fit [documentation template](https://github.com/nventive/Uno/blob/master/doc/.feature-template.md). (for bug fixes / features)
- [ ] [Unit Tests and/or UI Tests](doc/articles/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [ ] [Wasm UI Tests](doc/articles/working-with-the-samples-apps.md#running-the-webassembly-ui-tests-snapshots) are not showing unexpected any differences
- [x] Contains **NO** breaking changes
- [x] Updated the [Release Notes](https://github.com/nventive/Uno/tree/master/doc/ReleaseNotes)
- [x] Associated with an issue (GitHub or internal)

## Other information

Internal Issue (If applicable):

PR #12431, opened 2023-05-26
fix(AutoSuggestBox): fix the issue that e.ChosenSuggestion is incorrect

TopProgrammer77:

## PR Type

- Bugfix

My behavior to detect the fix
1. Type Text "abc" on TextBox of AutoSuggestBox
2. Press Enter Key. QuerySubmitted handler will be called.
 
e. ChosenSuggestion will be "abc", the string value 
## What is the new behavior?

e.ChosenSuggestion have to be null.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/AutoSuggestBox/AutoSuggestBox.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/AutoSuggestBox/AutoSuggestBox.cs
@@ -369,9 +369,9 @@ namespace Windows.UI.Xaml.Controls
 
 		private void SubmitSearch(object item)
 		{
-			var finalResult = item ?? GetObjectText(Text);
+			var finalResult = item;
 
-			QuerySubmitted?.Invoke(this, new AutoSuggestBoxQuerySubmittedEventArgs(finalResult is "" ? null : finalResult, userInput));
+			QuerySubmitted?.Invoke(this, new AutoSuggestBoxQuerySubmittedEventArgs(finalResult, userInput));
 
 			IsSuggestionListOpen = false;
 		}
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_AutoSuggestBox.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_AutoSuggestBox.cs
@@ -211,6 +211,25 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 			await WindowHelper.WaitForIdle();
 			Assert.AreEqual("b2", SUT.Text);
 		}
+
+		[TestMethod]
+		public async Task When_Submitting_After_Typing_Text()
+		{
+			var SUT = new AutoSuggestBox();
+			SUT.QuerySubmitted += (s, e) =>
+			{
+				Assert.IsNull(e.ChosenSuggestion);
+			};
+			WindowHelper.WindowContent = SUT;
+			await WindowHelper.WaitForIdle();
+			SUT.ItemsSource = new List<string>() { "ab", "abc", "abcde" };
+			await WindowHelper.WaitForIdle();
+			SUT.Focus(FocusState.Programmatic);
+			SUT.Text = "abc";
+			await WindowHelper.WaitForIdle();
+			SUT.Focus(FocusState.Programmatic);
+			await WindowHelper.WaitForIdle();
+		}
 #endif
 
 		[TestMethod]


Youssef1313:
It looks like this supersedes https://github.com/unoplatform/uno/pull/9355?

@kazo0 Since you reviewed the other PR, could you take a look at this one?

Youssef1313 (commented):
I think we need to explicitly pass null here:

https://github.com/unoplatform/uno/blob/9d6de4b11df69c75a2038e31021565971424bf2f/src/Uno.UI/UI/Xaml/Controls/AutoSuggestBox/AutoSuggestBox.cs#L388

src/Uno.UI/UI/Xaml/Controls/AutoSuggestBox/AutoSuggestBox.cs
```diff
@@ -369,9 +369,9 @@ private void OnQueryButtonClick(object sender, RoutedEventArgs e)
 
 		private void SubmitSearch(object item)
 		{
-			var finalResult = item ?? GetObjectText(Text);
+			var finalResult = item;
 
-			QuerySubmitted?.Invoke(this, new AutoSuggestBoxQuerySubmittedEventArgs(finalResult is "" ? null : finalResult, userInput));
+			QuerySubmitted?.Invoke(this, new AutoSuggestBoxQuerySubmittedEventArgs(finalResult, userInput));
```
Youssef1313, line 374: There is no point in assigning `item` to `finalResult`, you can use `item` directly.

Youssef1313: The `userInput` argument is quite concerning as well. Should it be `Text`? @Xiaoy312 @kazo0

TopProgrammer77:
@Youssef1313 , I think I and @kazo0  mentioned the same issue.
But in my opinion, this https://github.com/unoplatform/uno/pull/9355/commits/9908c9cae502cc088899869ffbdbb977627e8fa9#diff-de2b6bd5a0b6291648b873f3002bbd0714881f89f15e0a4075033c7bc1e9e8baR332 has some mistake.
For example, when ItemsSource = { "a1", "a2" } and I type "a1" and then press Enter, ChosenSuggestion should be null, not "a1".
Because we didn't select the item "a1"

Youssef1313:
> @Youssef1313 , I think I and @kazo0 mentioned the same issue. But in my opinion, this [9908c9c#diff-de2b6bd5a0b6291648b873f3002bbd0714881f89f15e0a4075033c7bc1e9e8baR332](https://github.com/unoplatform/uno/commit/9908c9cae502cc088899869ffbdbb977627e8fa9#diff-de2b6bd5a0b6291648b873f3002bbd0714881f89f15e0a4075033c7bc1e9e8baR332) has some mistake. For example, when ItemsSource = { "a1", "a2" } and I type "a1" and then press Enter, ChosenSuggestion should be null, not "a1". Because we didn't select the item "a1"

Yes, but shouldn't we pass null here to get this correct behavior?

https://github.com/unoplatform/uno/blob/9d6de4b11df69c75a2038e31021565971424bf2f/src/Uno.UI/UI/Xaml/Controls/AutoSuggestBox/AutoSuggestBox.cs#L388

TopProgrammer77:
> but shouldn't we pass null here to get this correct behavior?

If the parameter of SubmitSearch is null, ChosenSuggestion will be null. So It will be correct.

TopProgrammer77:
@Youssef1313 , you mean _suggestionsList.SelectedItem can't be null?

TopProgrammer77:
> > @Youssef1313 , I think I and @kazo0 mentioned the same issue. But in my opinion, this [9908c9c#diff-de2b6bd5a0b6291648b873f3002bbd0714881f89f15e0a4075033c7bc1e9e8baR332](https://github.com/unoplatform/uno/commit/9908c9cae502cc088899869ffbdbb977627e8fa9#diff-de2b6bd5a0b6291648b873f3002bbd0714881f89f15e0a4075033c7bc1e9e8baR332) has some mistake. For example, when ItemsSource = { "a1", "a2" } and I type "a1" and then press Enter, ChosenSuggestion should be null, not "a1". Because we didn't select the item "a1"
> 
> Yes, but shouldn't we pass null here to get this correct behavior?
> 
> https://github.com/unoplatform/uno/blob/9d6de4b11df69c75a2038e31021565971424bf2f/src/Uno.UI/UI/Xaml/Controls/AutoSuggestBox/AutoSuggestBox.cs#L388

Yes, you are right, When query button is clicked, ChosenSuggestion shoud be null.
i.e. SubmitSearch(null);

src/Uno.UI/UI/Xaml/Controls/AutoSuggestBox/AutoSuggestBox.cs
TopProgrammer77: @Youssef1313 , In all cases, the event args QueryText property contains the text from the text box.
So in my opinion, userInput is always _textBox.Text

Youssef1313: > So in my opinion, userInput is always _textBox.Text

Is that *guaranteed*? If yes, why not be on the safe side and use `_textBox.Text` as it's more clear?

TopProgrammer77: > > So in my opinion, userInput is always _textBox.Text
> 
> Is that _guaranteed_? If yes, why not be on the safe side and use `_textBox.Text` as it's more clear?

Oh, I was wrong. when we type the text with keyborar to TextBox, userInput refers to _textBox.Text. Otherwise( like prgrammatically change of textbox.text), userInput isn't same with _textBox.Text.

TopProgrammer77: I think the second parameter should be _textBox.Text. userInput is wrong.
AutoSuggestBoxQuerySubmittedEventArgs.QueryText is clearly _textBox.Text,
Thank you, @Youssef1313

TopProgrammer77: How about giving Text ( i.e. this.Text) as QuerySubmit()?

Youssef1313: I'd prefer `_textBox.Text` since that is what the [documentation](https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.autosuggestbox?view=winrt-22621#query-submitted) explicitly says.

> In all cases, the event args [QueryText](https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.autosuggestboxquerysubmittedeventargs.querytext?view=winrt-22621#windows-ui-xaml-controls-autosuggestboxquerysubmittedeventargs-querytext) property contains the text from the text box.

TopProgrammer77: @Youssef1313 ,By testing, `this.Text` is wrong! Because this.Text is set after calling the QuerSubmitted handler.

TopProgrammer77: https://github.com/unoplatform/uno/blob/9d6de4b11df69c75a2038e31021565971424bf2f/src/Uno.UI/UI/Xaml/Controls/AutoSuggestBox/AutoSuggestBox.cs#L349-L358
@Youssef1313 , I think it should be fixed.
And I don't know well if _textBox.Text is right.

TopProgrammer77: By calling ChoseItem before SubmitSearch, the this.Text is already set before calling QuerySubmitted.

src/Uno.UI/UI/Xaml/Controls/AutoSuggestBox/AutoSuggestBox.cs
```diff
@@ -364,16 +360,15 @@ private void OnQueryButtonClick(object sender, RoutedEventArgs e)
...
 
 		private void SubmitSearch(object item)
 		{
-			var finalResult = item ?? GetObjectText(Text);
-
-			QuerySubmitted?.Invoke(this, new AutoSuggestBoxQuerySubmittedEventArgs(finalResult is "" ? null : finalResult, userInput));
+			QuerySubmitted?.Invoke(this, new AutoSuggestBoxQuerySubmittedEventArgs(item, _textBox.Text));
 
 			IsSuggestionListOpen = false;
+			_suggestionsList.SelectedItem = null;
```
Youssef1313, line 371: Why is this line added?

Youssef1313 (commented):
This code path might need some tweaking:

https://github.com/unoplatform/uno/blob/5a502acf4e125a485edc492aad8805da9a6ebb2f/src/Uno.UI/UI/Xaml/Controls/AutoSuggestBox/AutoSuggestBox.cs#L383

I don't think we should use `_suggestionsList.SelectedItem` *if* the suggestion list is closed. So maybe `SubmitSearch(IsSuggestionListOpen ? _suggestionsList.SelectedItem : null);`

TopProgrammer77:
@Youssef1313 , How can I fix this https://github.com/unoplatform/uno/pull/12431/checks?check_run_id=13838800371

Youssef1313:
I restarted the build and it's green now.

TopProgrammer77:
> I restarted the build and it's green now.

Thank you, @Youssef1313 . Could you approve this PR?

src/Uno.UI/UI/Xaml/Controls/AutoSuggestBox/AutoSuggestBox.cs
```diff
@@ -94,13 +94,9 @@ private void OnItemsChanged(IObservableVector<object> sender, IVectorChangedEven
 
 		protected override void OnItemsSourceChanged(DependencyPropertyChangedEventArgs e)
 		{
-			// Calling this method before base.OnItemsSourceChanged() ensures that, in the case of an ObservableCollection, the list
-			// subscribes to CollectionChanged before AutoSuggestBox does. This is important for Android because the list needs to
-			// notify RecyclerView of collection changes before UpdateSuggestionList() measures it, otherwise we get errors like
-			// "Inconsistency detected. Invalid view holder adapter position"
-			UpdateSuggestionList();
-
 			base.OnItemsSourceChanged(e);
```
jeromelaban, line 97: Changing this will likely break android, which was fixed at that time: https://github.com/unoplatform/uno/pull/1069. You can validate the behavior of this particular change with the android binary and the test provided here:

https://github.com/unoplatform/uno/blob/eee3d255b2ccac1adc311c61d4d59fc2ee1b2df7/src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/AutoSuggestBoxTests/BasicAutoSuggestBox.xaml#L1

TopProgrammer77: Currently, I don't have the environment that I can test with Android. How about adding one more after base.OnItemsSourceChanged(e)?

```
UpdateSuggestionList();
base.OnItemsSourceChanged(e);
UpdateSuggestionsList();
```
Anyway, UpdateSuggestionsList() have to be after `base.OnItemSourceChanged(e);`

TopProgrammer77: When I build SamplesApp.Droid, I have some error related with Xamarainmac20.
How can I fix it? Can I remove the Xamarainmac20 in TargetPlatform of all the .csproj to fix?

Youssef1313: @TopProgrammer77 Which solution filter you are dealing with? What UnoTargetFrameworkOverride value are you using? and what is the error?

TopProgrammer77: @Youssef1313 ,Where to set the UnoTargetFrameworkOverride?

TopProgrammer77: > Changing this will likely break android, which was fixed at that time: #1069. You can validate the behavior of this particular change with the android binary and the test provided here:
> 
> https://github.com/unoplatform/uno/blob/eee3d255b2ccac1adc311c61d4d59fc2ee1b2df7/src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/AutoSuggestBoxTests/BasicAutoSuggestBox.xaml#L1

@jeromelaban , I tested in Android and the error was not detected. The error is resolved by @Xiaoy312 in https://github.com/unoplatform/uno/blob/3be20e489d67ed2c024163173f492ccb82566fbc/src/Uno.UI/UI/Xaml/Controls/ItemsControl/ItemsControl.cs#L688

TopProgrammer77: The android issue was related with https://github.com/unoplatform/uno/pull/12158. By merging this [PR](https://github.com/unoplatform/uno/pull/12158), it was resolved. So I think it is right to call `UpdateSuggestionList()` after `base.OnItemsSourceChanged(e)`.

MartinZikmund (approved):
As the related Android issue appears fixed, it seems this change is ok

TopProgrammer77:
Thank you so much. @MartinZikmund

DierkDroth:
@MartinZikmund would it be possible to get this PR merged? I would be happy to run some tests as a new build would be available.

Youssef1313 (approved):
LGTM, assuming the change to call `UpdateSuggestionList` later was well-tested on Android.

src/Uno.UI/UI/Xaml/Controls/AutoSuggestBox/AutoSuggestBox.cs
jeromelaban: It's generally best to avoid resolving comments until the maintainers have actually provided the resolution.

I've been able to test the changes on a device, and it is working properly. Thanks for the research!

DierkDroth:
@TopProgrammer77 @Youssef1313 @jeromelaban unfortunately this fix breaks the AutoSuggestBox completely. It possibly has fixed the orginal issue but now an issue is exposed (it might be a different issue, not sure) which makes the AutoSuggestBox completely unusable in my project and actually prevents me from upgrading to this build :-(

Anyway, after much hair-pulling I was able to isolate the cause of trouble. Please see attached simple repo, run it on WASM and check out the comments I left in sources. 

Is there any chance to get this fixed? As mentioned this is road block which prevents me upgrading UNO at all. Thanks in advance

[AutoSuggestBox5.zip](https://github.com/unoplatform/uno/files/11702283/AutoSuggestBox5.zip)

Youssef1313:
@DierkDroth Can you open an issue to better track this?

DierkDroth:
@Youssef1313 to clarify. So you're sure it's a different issue and not related? #12583

Issue #8996, opened 2022-06-08
[Gallery][Android] Commandbar- For 'Material Page' Close button is missing in light theme.


sakshi173:
### Current behavior

![Close button issue](https://user-images.githubusercontent.com/74563602/172694442-7682a158-b797-430c-9901-95a5721b44ce.png)

### Expected behavior

![Close button](https://user-images.githubusercontent.com/74563602/172694469-69ed6c97-6524-428a-84c2-f9775f39618f.png)

### How to reproduce it (as minimally and precisely as possible)

1. Open Uno Gallery APP
2. Tap Burger Menu on the left top right.
3. Click on the Light/Dark theme change toggle button.
4. Notice that the app is displaying in the Light theme.
5. Tap "CommandBar" from the "Component" list.
6. Tap on the 'Show Sample' button.
7. Observe that the Close button is missing on the top header.

### Affected platforms

Android

sakshi173:
@jeromelaban ,This issue is still present, please look into it.

Youssef1313:
This might get fixed by. https://github.com/unoplatform/uno/pull/12691

agneszitte:
@sakshi173 now that https://github.com/unoplatform/uno/pull/12691 is merged, can you confirm if this is still happening on latest canary build please

sakshi173:
@agneszitte , I retested but this issue is still present.

MartinZikmund:
Might still be relevant to Skia targets in general

MartinZikmund:
# Repro generator

This issue was classified as not reproducible with a code sample or test.

**Reason:** Native platform-specific (iOS/Android/SkiaGtk) CommandBar styling issue in Gallery.

PR #12503, opened 2023-06-02
fix: AppBarButton icon color for native renderers

kazo0:
related https://github.com/unoplatform/nventive-private/issues/487
related https://github.com/unoplatform/uno/issues/8996

## PR Type

- Bugfix

Setting the Foreground property of a BitmapIcon has no effect when being used within the native renderers for CommandBar on iOS/Android

## What is the new behavior?

Foreground on BitmapIcon is respected and takes precedence over the Foreground of the AppBarButton

### DIFF
--- a/src/Uno.UI/Controls/CommandBar/AppBarButtonExtensions.cs
+++ b/src/Uno.UI/Controls/CommandBar/AppBarButtonExtensions.cs
@@ -1,0 +1,36 @@
+﻿#nullable enable
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Windows.UI;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Controls;
+using Windows.UI.Xaml.Media;
+
+namespace Uno.UI
+{
+	internal static class AppBarButtonExtensions
+	{
+		public static bool TryGetIconColor(this AppBarButton appBarButton, out Color iconColor)
+		{
+			iconColor = default;
+
+			if (appBarButton.Icon?.ReadLocalValue(IconElement.ForegroundProperty) != DependencyProperty.UnsetValue &&
+				Brush.TryGetColorWithOpacity(appBarButton.Icon?.Foreground, out var iconForeground))
+			{
+				iconColor = iconForeground;
+				return true;
+			}
+
+			if (Brush.TryGetColorWithOpacity(appBarButton.Foreground, out var buttonForeground))
+			{
+				iconColor = buttonForeground;
+				return true;
+			}
+
+			return false;
+		}
+	}
+}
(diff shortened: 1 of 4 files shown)


ajpinedam (approved):
Nice!

agneszitte (approved):

src/Uno.UI/Controls/CommandBar/AppBarButtonExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+			{
+				iconColor = buttonForeground;
+				return true;
+			}
+
+			return false;
+		}
+	}
+}
+#nullable disable
```
Youssef1313, line 37: This doesn't have any effect at the end of the file

src/Uno.UI/Controls/CommandBar/AppBarButtonRenderer.Android.cs
```diff
@@ -152,14 +152,13 @@ protected override void Render()
...
-			var foregroundOpacity = foreground?.Opacity ?? 0;
+			var foregroundOpacity = 0d;
 			if (native.Icon != null)
 			{
-				if (foreground != null)
+				if (element.TryGetIconColor(out var iconColor))
 				{
-					DrawableCompat.SetTint(native.Icon, (Android.Graphics.Color)foreground.Color);
+					foregroundOpacity = iconColor.A / 255d;
+					DrawableCompat.SetTint(native.Icon, (Android.Graphics.Color)iconColor);
```
jeromelaban, line 161: We should be able to UI tests those changes, right? Runtime should be ok now as well.

src/Uno.UI/Controls/CommandBar/AppBarButtonRenderer.Android.cs
```diff
@@ -98,6 +96,21 @@ protected override void Render()
 						native.SetTitle(null);
 						break;
 
+					case FontIcon: // not supported
+					case PathIcon: // not supported
+					case SymbolIcon: // not supported
```
Youssef1313, line 101: This is reverting https://github.com/unoplatform/uno/pull/12691

kazo0: Doesn't the effectively achieve the same thing? And it also adds back to the warning log for when an unsupported IconElement is used?

Youssef1313: The thing is that IconElement is now supported. They work through the `case FrameworkElement`, that's why I merged the two switch statements using `iconOrContent`

kazo0: Omg you're right! Ok so it just uses the wrapper, oh wow I can't believe I didn't realize we could do that!

Alright I will keep this and also make a PR in Toolkit for NavigationBar also handles this!

jeromelaban:
@kazo0 is this PR still needed?

PR #12559, opened 2023-06-06
feat(scp): Add native support for reversed mouse wheel

dr1rrb:
## Feature
Add native support for reversed mouse wheel

There is no way to properly implement reversed mouse wheel scrolling on wasm

## What is the new behavior?
Uno offers flag to support this.

## PR Checklist


src/Uno.UI.Runtime.Skia.Gtk/GtkCorePointerInputSource.cs
```diff
@@ -411,7 +412,7 @@ private void UseDevice(PointerPoint pointer, Gdk.Device device)
 			case EventType.Scroll when EventHelper.GetScrollDeltas(evt!, out var scrollX, out var scrollY):
 				var isHorizontal = scrollY == 0;
 				properties.IsHorizontalMouseWheel = isHorizontal;
-				properties.MouseWheelDelta = (int)(isHorizontal ? scrollX : scrollY);
+				properties.MouseWheelDelta = (int)(isHorizontal ? scrollX : -scrollY) * ScrollContentPresenter.ScrollViewerDefaultMouseWheelDelta;
```
Youssef1313, line 415: Is the cast intended before the multiplication? or should it be after?

src/Uno.UI.Runtime.Skia.Linux.FrameBuffer/FrameBufferPointerInputSource.Mouse.cs
```diff
@@ -52,13 +53,13 @@ public void ProcessMouseEvent(IntPtr rawEvent, libinput_event_type type)
...
 				properties.IsHorizontalMouseWheel = true;
-				properties.MouseWheelDelta = (int)GetAxisValue(libinput_pointer_axis.ScrollHorizontal);
+				properties.MouseWheelDelta = (int)GetAxisValue(libinput_pointer_axis.ScrollHorizontal) * ScrollContentPresenter.ScrollViewerDefaultMouseWheelDelta;
 				raisePointerEvent = RaisePointerWheelChanged;
 			}
 			else if (libinput_event_pointer_has_axis(rawPointerEvent, libinput_pointer_axis.ScrollVertical) != 0)
 			{
 				properties.IsHorizontalMouseWheel = false;
-				properties.MouseWheelDelta = (int)GetAxisValue(libinput_pointer_axis.ScrollVertical);
+				properties.MouseWheelDelta = (int)GetAxisValue(libinput_pointer_axis.ScrollVertical) * ScrollContentPresenter.ScrollViewerDefaultMouseWheelDelta;
```
Youssef1313, line 62: Same question here

src/Uno.UI.Runtime.Skia.Gtk/GtkCorePointerInputSource.cs
dr1rrb: Not really, will change it. Anyway from my tests we are getting only `1` or `-1` so it should not change anything.

src/Uno.UI/UI/Xaml/Input/PointerRoutedEventArgs.wasm.cs
```diff
@@ -15,7 +16,7 @@
 
 namespace Windows.UI.Xaml.Input
 {
-	partial class PointerRoutedEventArgs
+	partial class PointerRoutedEventArgs : IHtmlHandleableRoutedEventArgs
```
jeromelaban, line 19: Maybe:
```suggestion
	partial class PointerRoutedEventArgs : ICanHandleHtmlRoutedEventArgs
```

src/Uno.UI/UI/Xaml/UIElement.Pointers.wasm.cs
```diff
@@ -168,69 +168,96 @@ public static void OnNativePointerEvent()
...
-						// If so, we stop bubbling since our parent will do the same!
-						var ptArgs = ToPointerArgs(element, args);
-						stopPropagation = element.IsOver(ptArgs.Pointer) || element.OnNativePointerEnter(ptArgs);
-						break;
+						result.Add(ptArgs, element.OnNativePointerEnter(ptArgs));
 					}
+					break;
+				}
 
 				case NativePointerEvent.pointerout: // No needs to check IsOver (leaving vs. bubbling), it's already done in native code
```
jeromelaban, line 187:
```suggestion
				case NativePointerEvent.pointerout: // No need to check IsOver (leaving vs. bubbling), it's already done in native code
```

src/Uno.UI.Runtime.Skia.Gtk/GtkCorePointerInputSource.cs
jeromelaban, line 1: Can this be tested with the pointers emulation?

src/Uno.UI/UI/Xaml/Input/PointerRoutedEventArgs.wasm.cs
dr1rrb: This is to match the already existing [`IHandleableRoutedEventArgs`](https://github.com/unoplatform/uno/blob/master/src/Uno.UI/UI/Xaml/Input/Internal/IHandleableRoutedEventArgs.cs)

src/Uno.UI.Runtime.Skia.Gtk/GtkCorePointerInputSource.cs
```diff
@@ -411,7 +412,7 @@ private void UseDevice(PointerPoint pointer, Gdk.Device device)
 			case EventType.Scroll when EventHelper.GetScrollDeltas(evt!, out var scrollX, out var scrollY):
 				var isHorizontal = scrollY == 0;
 				properties.IsHorizontalMouseWheel = isHorizontal;
-				properties.MouseWheelDelta = (int)(isHorizontal ? scrollX : scrollY);
+				properties.MouseWheelDelta = (int)((isHorizontal ? scrollX : -scrollY) * ScrollContentPresenter.ScrollViewerDefaultMouseWheelDelta);
```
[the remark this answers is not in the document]
dr1rrb: Yes, I went through all platform to ensure that we have same direction and (approximative) values.

For reference, on windows 1 step on my mouse gives :
up => -120
down => +120
right => +120
left => -120
(Windows 10 RDP to Win 11 ... so without the incredibly annoying reversed scroll of Win 11 - not opinionated at all)

src/Uno.UI.Runtime.Skia.Gtk/GtkCorePointerInputSource.cs
dr1rrb: Not this file explicitly. We could probably tests the rest of the PR, but wheel it's not yet supported, it would require some wire-up ...

src/Uno.UI/UI/Xaml/UIElement.Pointers.wasm.cs
```diff
@@ -143,9 +143,10 @@ public partial class UIElement : DependencyObject
 #if NET7_0_OR_GREATER
 	[JSExport]
 #endif
+#pragma warning disable IDE0055 // Fix formatting
```
Youssef1313, line 146: The reason you are seeing this is because the BlockSyntax inside the "case" is not indented. This behavior can be controlled via `csharp_indent_case_contents_when_block` in `.editorconfig` (the property defaults to true). The following is example from the docs that shows how this option controls formatting.

```
// csharp_indent_case_contents_when_block = true
case 0:
    {
        Console.WriteLine("Hello");
        break;
    }

// csharp_indent_case_contents_when_block = false
case 0:
{
    Console.WriteLine("Hello");
    break;
}
```

If the desired formatting is to not indent the block, then we should adjust the property in .editorconfig rather than disabling the analyzer with `#pragma`.

dr1rrb: Set this to `csharp_indent_case_contents_when_block = false:suggestion` so it won't force one over the other.

Youssef1313: That *will* still force the "false".

If this produced too many violations in CI, we may keep the consistency for now and indent the block, then adjust for the whole codebase in a follow-up PR.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_ScrollViewer.cs
```diff
@@ -751,5 +752,52 @@ public async Task When_TouchTap_Then_NestedElementReceivePointerEvents()
...
+
+		[TestMethod]
+		[RunsOnUIThread]
+#if !__SKIA__
+		[Ignore("Pointer injection supported only on skia for now.")]
+#endif
+		public async Task When_ReversedMouseWheel_Then_ScrollInReversedDirection()
+		{
+#if WINDOWS_UWP
+			Assert.Inconclusive("Mouse pointer helper not supported on UWP.");
```
Youssef1313, line 764: Consider adding `await Task.CompletedTask;` for WINDOWS_UWP to avoid disabling CS1998

dr1rrb: As far as I know the CS1998 is irrelevant in (runtime) tests (and actually in most code as it's almost only to ensure to not add overhead of the async/await state machine when not needed). The relevant one is CS4014 as it will ensure that a `Task` returned by a method is being awaited.

Adding `await Task.CompletedTask;` is only a trick to hide the rule without adding a proper ignore, and I suspect it will even be less performant (as it will generate the state machine for the `CompletedTask` it would not with the ignore). It also unnecessarily complexity the source code.

src/Uno.UI/UI/Xaml/UIElement.Pointers.wasm.cs
dr1rrb: It works on my computer ... let see on the CI :)

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_ScrollViewer.cs
Youssef1313: Technically, they are functionally different. While the difference isn't going to be noticeable for this specific case, it might be for other cases.

```csharp
using System;
using System.Threading.Tasks;

_ = new C().M();
Console.WriteLine("Exception is swallowed"); // if M is async, this is printed. Otherwise, the exception is thrown. This happens only when the returned task isn't awaited.

public class C
{
    public
        async
        Task<C> M()
    {
        throw new Exception();
    }
}
```

Youssef1313: Oh, my suggestion to add `await Task.CompletedTask;` will actually keep the `async` which will still swallow expcetions for non-awaited returned tasks.

Youssef1313: No strong opinion given this is a test code, but I'd personally add `async` modifier conditionally (rather than `await Task.CompletedTask;` which is pointless), or at least suppress for this specific case with `#pragma warning disable`.

My concern with globally disabling the warning for runtime tests is cases where exceptions are unexpectedly swallowed (but I also see it not being a common mistake we could do)

dr1rrb: Yes the case `_ = new C().M();` is why I said "almost", but here the difference in behavior is mainly because you also added the "ignore CS4014 trick" by adding the `_ =`, if you do the standard `await new C().M();` the exception is being raised properly.

Adding conditionally the `async` is also another option, but again adds noise in the source code. Another option would also to just put the whole test in the 1 #if !WINDOWS_UWP` but this as the disadvantage to hide tests that are not running on a specific platform.

As it's for **multi-targeted** tests, where it's quite common to have platform specific code like this, I would rather rely on the CS4014 than on the CS1998 and ease usage with a global ignore like this (in this project only of course!). Any opinion @jeromelaban ?

Issue #12454, opened 2023-05-29
Image Source to null leads to an undefined behaviour


jp-weber:
### Current behavior

If there is currently no value for the image source in a ListView or an ItemsRepeater of a DataTemplate, or if it is deleted, this leads to an exception being thrown and often an image from the cache is used as the image source.

For example the purple items have an image source of null (red=> pride flag, green => uno logo):
<img src="https://github.com/unoplatform/uno/assets/10946850/d9eac779-0fe1-455a-8438-3a8a32ee5b68" alt="Screenshot_20230528-193005" width="470" />

The problem also existed at MAUI and was solved recently:
https://github.com/dotnet/maui/issues/8787
https://github.com/dotnet/maui/pull/13614

```
[DOTNET] [41m[30mfail[39m[22m[49m: Microsoft.UI.Xaml.Data.BindingExpression[0]
[DOTNET]       Failed to apply binding to property [DependencyPropertyDetails(UriSource)] on [Microsoft.UI.Xaml.Media.Imaging.SvgImageSource] (Object reference not set to an instance of an object.)
[DOTNET]       System.NullReferenceException: Object reference not set to an instance of an object.
[DOTNET]          at Microsoft.UI.Xaml.Media.ImageSource.InitFromUri(Uri uri) in C:\a\1\s\src\Uno.UI\UI\Xaml\Media\ImageSource.cs:line 116
[DOTNET]          at Microsoft.UI.Xaml.Media.Imaging.SvgImageSource.OnUriSourceChanged(DependencyPropertyChangedEventArgs e) in C:\a\1\s\src\Uno.UI\UI\Xaml\Media\Imaging\SvgImageSource.cs:line 59
[DOTNET]          at Microsoft.UI.Xaml.Media.Imaging.SvgImageSource.<>c.<.cctor>b__63_0(DependencyObject s, DependencyPropertyChangedEventArgs e) in C:\a\1\s\src\Uno.UI\UI\Xaml\Media\Imaging\SvgImageSource.Properties.cs:line 26
[DOTNET]          at Microsoft.UI.Xaml.PropertyMetadata.RaisePropertyChanged(DependencyObject source, DependencyPropertyChangedEventArgs e) in C:\a\1\s\src\Uno.UI\UI\Xaml\PropertyMetadata.cs:line 187
[...]

PR #12560, opened 2023-06-07
fix(Image): SvgImageLoader crash on binding

ajpinedam:
GitHub Issue (If applicable): closes https://github.com/unoplatform/uno/issues/12454

## PR Type

- Bugfix

SVGImageSource, when set to null, there's an error, and it shows cached images.

## PR Checklist

## Other information


src/Uno.UI/UI/Xaml/Media/Imaging/SvgImageSource.cs
```diff
@@ -56,7 +56,12 @@ private void OnUriSourceChanged(DependencyPropertyChangedEventArgs e)
 		{
 			UnloadImageData();
 		}
-		InitFromUri(e.NewValue as Uri);
+
+		if (e.NewValue is Uri uri)
+		{
+			InitFromUri(uri);
+		}
```
Youssef1313, line 63: If we are handling `null` in `InitFromUri`, then we don't need the if condition here

jeromelaban: We still need the cast, that's fine if we check for null as well.

src/Uno.UI/UI/Xaml/Media/Imaging/SvgImageSource.cs
```diff
@@ -56,7 +56,12 @@ private void OnUriSourceChanged(DependencyPropertyChangedEventArgs e)
 		{
 			UnloadImageData();
 		}
-		InitFromUri(e.NewValue as Uri);
+
+		if (e.NewValue is Uri uri)
+		{
+			InitFromUri(uri);
```
Youssef1313, line 62: We should be consistent here and in:

https://github.com/unoplatform/uno/blob/aa761a554d8dcd6d4bdbc0c85f6df3d300585cc0/src/Uno.UI/UI/Xaml/Media/Imaging/BitmapImage.cs#L35

But couple of points:

- Aside from null annotations (which is done in the other PR), we should define whether `InitFromUri` expects null or not.
- Currently, you moved few lines in `InitFromUri`, but this move is redundant if we will not be passing null.
- What is the impact of clearing `FilePath` and `AbsoluteUri` in `InitFromUri`? Is the effect visible?

I think we only need one of the two changes being done here.

ajpinedam: `InitFromUri(Uri)` should not allow `null`s as part of the parameters, but unfortunately, this is not something we can fully enforce on the Source Code until we have 100% null enable capabilities. 

"I think we only need one of the two changes being done here. Either guard InitFromUri calls against nulls (ie, pass it non-null), or let InitFromUri itself handle nulls."

For this fix, I do agree that making `InitFromUri()` handle the `null` is more than sufficient, I will remove it from here, but the other PR [null enabling ImageSource](https://github.com/unoplatform/uno/pull/12562) will force us to add this null validation because of the casting.

cc: @Youssef1313

ajpinedam: "Currently, you moved few lines in InitFromUri, but this move is redundant if we will not be passing null."
Until we are 100% null-enabled, nothing prevents this method from being called with a null value from another place

Youssef1313 (approved):

jeromelaban:
@ajpinedam You'll need to raise `Assert.Inconclusive()` for the platforms that don't support SVG loading. You should be able to use `ApiInformation.IsTypePresent` (or similar) for that purpose.

jeromelaban (approved):

PR #12707, opened 2023-06-26
Extend RemoteControlServer functionality

mrlacey:
GitHub Issue (If applicable): none

## PR Type

- Other... Please describe: Enhancing existing functionality

Current behavior only allows for a single processor to be registered for each scope.
This prevents the implementation of functionality that relies on multiple RCClients being registered with the same scope or using an RCServer to communicate between clients.

## What is the new behavior?

This PR adds the ability to:

- Load multiple processors for each scope
- Unregister a previously registered processor

Because a client may not unregister a processor, the RCServer will only load a single instance of a processor (based on the full name) for each scope.
When attempting to register a processor, any existing versions of that processor will first be unloaded, and a new instance will replace the old one.

An awareness of the above is important as it means that multiple identical clients (multiple instances of the same thing) would not be supported. **Could this be an issue? - Feedback on this is desired?**

The implementation also uses a separate static class to allow messages received by one processor to be sent to other registered processors.

### <samp>🤖 Generated by Copilot at 5b1ad0d</samp>

This pull request enhances the `RemoteControlServer` class to handle multiple and dynamic processor assemblies. It introduces a new message type, `ProcessorRemoval`, to signal when a processor is unloaded and should be removed from the pool. It also improves the logging and dependency resolution of the processor assemblies.

## PR Checklist

## Other information


src/Uno.UI.RemoteControl.Host/RemoteControlServer.cs
```diff
@@ -35,16 +35,43 @@ public RemoteControlServer(IConfiguration configuration)
...
-					return context.LoadFromAssemblyPath(Path.Combine(Path.GetDirectoryName(_loopAssembly.Location)!, assemblyName.Name + ".dll"));
-				}
-				else
-				{
-					return context.LoadFromAssemblyName(assemblyName);
+					try
+					{
+						var loc = _loopAssembly.Location;
+
+						if (!string.IsNullOrWhiteSpace(loc))
```
nickrandolph, line 49: Each of these "if" conditions need an else, if only for outputting a meaningful log.
Also, if the processor doesn't load, what's the fallback - is there a way the user can fix this?

src/Uno.UI.RemoteControl.Host/RemoteControlServer.cs
```diff
@@ -35,16 +35,43 @@ public RemoteControlServer(IConfiguration configuration)
...
+									return context.LoadFromAssemblyPath(relPath);
+								}
+							}
+						}
+					}
+					catch (Exception exc)
+					{
+						if (this.Log().IsEnabled(LogLevel.Error))
+						{
+							System.Diagnostics.Debug.WriteLine(exc);
```
nickrandolph, line 68: Remove debug.writeline

src/Uno.UI.RemoteControl.Host/RemoteControlServer.cs
```diff
@@ -2,22 +2,22 @@
...
 using Uno.UI.RemoteControl.Messages;
-using System.Runtime.Loader;
-using Microsoft.Extensions.Configuration;
 
 namespace Uno.UI.RemoteControl.Host
 {
 	internal class RemoteControlServer : IRemoteControlServer, IDisposable
 	{
-		private readonly Dictionary<string, IServerProcessor> _processors = new Dictionary<string, IServerProcessor>();
+		////private readonly Dictionary<string, IServerProcessor> _processors = new();
```
nickrandolph, line 20: To avoid the static instance, can't we just have an instance of the SharedProcessorPool tracked here, instead of the Dictionary?

src/Uno.UI.RemoteControl.Host/RemoteControlServer.cs
```diff
@@ -56,9 +83,26 @@ public RemoteControlServer(IConfiguration configuration)
 		string IRemoteControlServer.GetServerConfiguration(string key)
 			=> _configuration[key] ?? "";
 
-		private void RegisterProcessor(IServerProcessor hotReloadProcessor)
+		private void RegisterProcessor(IServerProcessor processor)
 		{
-			_processors[hotReloadProcessor.Scope] = hotReloadProcessor;
+			if (SharedProcessorPool.Instance.Processors.ContainsKey(processor.Scope))
```
nickrandolph, line 88: Use TryGetValue instead of ContainsKey, to avoid two lookups?

src/Uno.UI.RemoteControl.Host/RemoteControlServer.cs
```diff
@@ -56,9 +83,26 @@ public RemoteControlServer(IConfiguration configuration)
...
+			{
+				// Remove any exising processor instances from the same source
+				foreach (var registeredProcessor in SharedProcessorPool.Instance.Processors[processor.Scope])
+				{
+					if (registeredProcessor.GetType().FullName == processor.GetType().FullName)
```
nickrandolph, line 93: Is it possible for this type to not be the actual same type? If it is, we should add a comment. If it's not possible, just compare the actual types

src/Uno.UI.RemoteControl.Host/RemoteControlServer.cs
```diff
@@ -207,12 +259,26 @@ private void ProcessDiscoveryFrame(Frame frame)
...
 					{
-						this.Log().LogDebug($"Failed to create instace of server processor in  {asm} : {exc}");
+						this.Log().LogDebug($"Failed to create instance of server processor in  {asm} : {exc}");
 					}
 				}
 			}
 		}
 
+		private void ProcessRemovalFrame(Frame frame)
+		{
```
nickrandolph, line 269: We should log trace or debug messages

src/Uno.UI.RemoteControl.Host/RemoteControlServer.cs
```diff
@@ -207,12 +259,26 @@ private void ProcessDiscoveryFrame(Frame frame)
 				{
 					if (this.Log().IsEnabled(LogLevel.Debug))
 					{
-						this.Log().LogDebug($"Failed to create instace of server processor in  {asm} : {exc}");
+						this.Log().LogDebug($"Failed to create instance of server processor in  {asm} : {exc}");
```
nickrandolph, line 262: It seems weird that we're logging debug messages here, when previously we were logging error message when the assembly couldn't be loaded

src/Uno.UI.RemoteControl/Messages/ProcessorRemoval.cs
```diff
@@ -0,0 +1,21 @@
...
+	{
+		public const string Name = nameof(ProcessorRemoval);
+
+		public ProcessorRemoval(string path, string processorScope)
+		{
+			Path = path;
+			ProcessorScope = processorScope;
+		}
+
+		public string Scope => "RemoteControlServer";
```
nickrandolph, line 13: Can scopes/messages that are shared between client and server be in a single file that's shared between the projects?

mrlacey: > Can scopes/messages that are shared between client and server be in a single file that's shared between the projects?

Do you mean the constant string value(s)?

src/Uno.UI.RemoteControl.Host/RemoteControlServer.cs
mrlacey: Because of the way they're created (via reflection), they _may_ be treated as different types. This seemed the simplest way to determine if we can treat them as the same.

src/Uno.UI.RemoteControl.Host/RemoteControlServer.cs
mrlacey: I'm not sure on the rules/guidance on when each should be used. This is likely only to be necessary/useful during development but have standardized with elsewhere.

src/Uno.UI.RemoteControl.Host/RemoteControlServer.cs
mrlacey: added logging in an else clause

mrlacey: If a processor doesn't load it's likely down to be trying to use something unsupported or a processor that isn't packed/distributed correctly. 
In either case, the fix isn't going to be straight forward and not something an end developer can deal with beyond reporting with full logs.

src/Uno.UI.RemoteControl.Host/RemoteControlServer.cs
mrlacey: moved to the static field version

mrlacey: The point of making it static was to enable a shared processor pool so that it's possible to have messages sent from one client relayed to others.

In terms of race conditions, I cannot foresee any situations where a message not being relayed to all clients is a problem, as this should only be an issue when connections are first being made, and this is when important messages are least likely to be sent. Any subsequent messages can always be resent if obviously not received by another client.

PR #12742, opened 2023-06-28
Update platform-specific-csharp.md to use the new template structure

ramezgerges:
## PR Type

### <samp>🤖 Generated by Copilot at 18b313f</samp>

This pull request fixes a documentation error in the platform-specific C# article. It corrects the file path for the `NativeWrapperControl.cs` file that demonstrates how to use native views in a cross-platform app.

## Other information
The instructions should refer to the class library, not the (outdated) shared project.

### DIFF
--- a/doc/articles/platform-specific-csharp.md
+++ b/doc/articles/platform-specific-csharp.md
@@ -12,12 +12,11 @@ This guide covers multiple approaches to managing per-platform code in C#. See [
 
 There are two ways to restrict code or XAML markup to be used only on a specific platform:
 
-* Use [conditionals](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives/preprocessor-if) within a shared file
+* Use [conditionals](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives/preprocessor-if) in a source file
 * Place the code in a separate file which is only included in the desired platform.
  
 The structure of an Uno app created with the default Visual Studio template is [explained in more detail here](uno-app-solution-structure.md).
- 
- 
+
  ## `#if` conditionals
  
  The most basic means of authoring platform-specific code is to use `#if` conditionals:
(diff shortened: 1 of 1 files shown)


Youssef1313 (changes requested):
I don't think the documented approach will work with the new project structure.

Previously, when we used shared projects, the code file in `PROJECTNAME.Shared` and the one in PROJECTNAME.SpecificPlatform` were to be compiled together in the same compilation, producing a single asssembly.

With the new project structure, the `PROJECTNAME.csproj` is a separate assembly, so this will be treated as a completely different class.

@jeromelaban to confirm, and to advise on how to move forward. My suggestion would be to delete the "Partial class definitions" approach in this doc page.

doc/articles/platform-specific-csharp.md
```diff
@@ -68,7 +68,7 @@ Heavy usage of `#if` conditionals makes code hard to read and comprehend. A bett
 
 ### A simple example
 
-Shared code in `PROJECTNAME.Shared/NativeWrapperControl.cs`:
+Shared code in `PROJECTNAME/NativeWrapperControl.cs`:
```
jeromelaban, line 71: > Previously, when we used shared projects, the code file in `PROJECTNAME.Shared` and the one in PROJECTNAME.SpecificPlatform` were to be compiled together in the same compilation, producing a single asssembly.

Shared projects created one binary per head, and we now have one binary per target framework for the new library project.

> @jeromelaban to confirm, and to advise on how to move forward. My suggestion would be to delete the "Partial class definitions" approach in this doc page.

Partial definitions are still valid, even if's not yet possible to make a distinction between wasm and skia. It will be possible in the future, likely in the next net8 preview with the introduction of `net8-browser`. 

We actually need to modify the next block to be:

---

Platform-specific code in `PROJECTNAME/NativeWrapperControl.Android.cs`:
```csharp
#if __ANDROID__
public partial class NativeWrapperControl : Control {
...
		private View CreateNativeView() {
			... //Android-specific code
```

Youssef1313 (approved):
Approving, assuming `PROJECTNAME.Droid` is updated to `PROJECTNAME`

doc/articles/platform-specific-csharp.md
ramezgerges: The article considers conditionals in a shared file and placing the code in the desired platform head as 2 separate approaches. As far as I currently understand it, partial definitions only work in the class library and should also be guarded by if conditionals, so these are similar to the first approach. Since the class library is compiled once per target framework (separate from each head), placing "the code in a file which is only included in the desired platform head" doesn't work anymore.

I've made more additions to fix/clarify the distinction.

doc/articles/platform-specific-csharp.md
```diff
@@ -13,7 +13,7 @@ This guide covers multiple approaches to managing per-platform code in C#. See [
 There are two ways to restrict code or XAML markup to be used only on a specific platform:
 
 * Use [conditionals](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives/preprocessor-if) within a shared file
-* Place the code in a file which is only included in the desired platform head.
+* Place the code in a separate file which is only included in the desired platform.
```
Youssef1313, line 16: The two approaches are almost the same. Both uses `#if` for files in the class library project.

But LGTM given the correctness of the change.

Youssef1313 (approved):

jeromelaban (approved):

PR #12760, opened 2023-06-30
Add Source location information to non-FrameworkElements

mrlacey:
GitHub Issue (If applicable): 

## PR Type

Other: Extending existing debug capabilities

DebugParseContext is not available for XAML elements that don't inherit from FrameworkElement. But there are scenarios where this information would be useful for other types.

## What is the new behavior?

Adding equivalent information for other types.

### <samp>🤖 Generated by Copilot at 10693b7</samp>

Improve XAML debugging for hot reload and source generators. Set `OriginalSourceLocation` for some non-`FrameworkElement` XAML elements in `XamlFileGenerator.cs`.

## PR Checklist

## Other information

As a draft for early visibility.
Need to consider the formatting of filename, line number, and character offset/position....

### DIFF
--- a/src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.cs
+++ b/src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.cs
@@ -2675,17 +2675,17 @@ namespace Uno.UI.SourceGenerators.XamlGenerator
 
 			foreach (var resource in (resourcesRoot?.Objects).Safe())
 			{
-				if (_isDebug)
-				{
-					// Track source location of resources by key as they may be lazily initialized
-					writer.AppendLineIndented($"global::Uno.UI.Helpers.MarkupHelper.SetElementProperty(\"RESOURCE::{resource.Members.FirstOrDefault(m => m.Member.Name == "Key").Value}\", \"OriginalSourceLocation\", \"file:///{_fileDefinition.FilePath.Replace("\\", "/")}#L{resource.LineNumber}:{resource.LinePosition}\");");
-				}
-
 				TryAnnotateWithGeneratorSource(writer, suffix: "PerKey");
 				var key = GetDictionaryResourceKey(resource, out var name);
 
 				if (key != null)
 				{
+					if (_isDebug)
+					{
+						// Track source location of resources by key as they may be lazily initialized
+						writer.AppendLineIndented($"global::Uno.UI.Helpers.MarkupHelper.SetElementProperty(\"RESOURCE::{key}\", \"OriginalSourceLocation\", \"file:///{_fileDefinition.FilePath.Replace("\\", "/")}#L{resource.LineNumber}:{resource.LinePosition}\");");
+					}
+
 					var wrappedKey = key;
 					if (!key.StartsWith("typeof(", StringComparison.InvariantCulture))
 					{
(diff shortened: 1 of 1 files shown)


src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.cs
```diff
@@ -3541,6 +3546,13 @@ void BuildCustomMarkupExtensionPropertyValue(IIndentedStringBuilder writer, Xaml
 			}
 		}
 
+		private bool IsNotFrameworkElementButNeedsSourceLocation(XamlObjectDefinition objectDefinition)
+		{
+			return objectDefinition.Type.Name == "VisualState"
+				|| objectDefinition.Type.Name == "AdaptiveTrigger"
+				|| objectDefinition.Type.Name == "StateTrigger";
+		}
```
Youssef1313, line 3554: nit: Make static and reduce repeating `objectDefinition.Type.Name`

```suggestion
		private static bool IsNotFrameworkElementButNeedsSourceLocation(XamlObjectDefinition objectDefinition)
			=> objectDefinition.Type.Name is "VisualState" or "AdaptiveTrigger" or "StateTrigger";
```

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.cs
```diff
@@ -3541,6 +3555,9 @@ void BuildCustomMarkupExtensionPropertyValue(IIndentedStringBuilder writer, Xaml
 			}
 		}
 
+		private static bool IsNotFrameworkElementButNeedsSourceLocation(XamlObjectDefinition objectDefinition)
+			=> objectDefinition.Type.Name is "VisualState" or "AdaptiveTrigger" or "StateTrigger";
```
jeromelaban, line 3559: We probably should be using lazy loaded symbols instead, like those, for performance reasons but also for exact type matching: 

https://github.com/unoplatform/uno/blob/1ddae936617e55845e1115482c057df92634a573/src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlCodeGeneration.cs#L260

jeromelaban (approved):

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.cs
mrlacey: > We probably should be using lazy loaded symbols instead, like those, for performance reasons but also for exact type matching:

Is this a general comment or a specific request for a change?
Is the performance issue about comparing with strings? There are equivalent checks elsewhere in the file, should they be changed too?
Are there examples of other places which already do a better/preferred way of handling this?
How to measure the perf impact of this, and any change?

carldebilly (approved):

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.cs
jeromelaban: Actually, I thought `.Type` was a roslyn symbol, but it's not in this case, since it's a string coming from the XAML parser. Let's keep it this way for now, apologies for the confusion.

PR #12763, opened 2023-07-01
fix(textbox): Fix TextBox keyboard navigation only working vertically

### LINKED ISSUE #11730: XYFocusKeyboardNavigation not working horizontaly on TextBoxes
### Current behavior

When using Skia with XYFocusKeyboardNavigation enabled, when TextBox is focused, left and right arrows are ignored (only moving cursor).

### Expected behavior

Arrows should move focus to next element if cursor is at the end of text.

https://playground.platform.uno/#001b1e32

### Workaround

_No response_

### Works on UWP/WinUI

Yes

### Environment

_No response_

### NuGet package version(s)

_No response_

### Affected platforms

WebAssembly, Skia (GTK on Linux/macOS/Windows)

### IDE

Visual Studio 2022

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

ramezgerges:
Now works in both directions (with quirks). Also added a new sample to showcase this.

GitHub Issue (If applicable): closes #11730

## PR Type

- Bugfix

Details are in the related issue.

## What is the new behavior?
Keyboard navigation should work horizontally as well. Currently, it isn't fully working. The navigation works, but it triggers 1 character early. Specifically, if the cursor is currently on the character before last in a textbox and the right arrow is pressed, the focus is moved to the neighbouring textbox (instead of the last character in the original textbox). A similar situation happens when going left as well. 

### <samp>🤖 Generated by Copilot at 2645857</samp>

This pull request adds a new UI test for text box keyboard navigation using `XYFocusKeyboardNavigation`, a property that enables the arrow keys to move the focus between elements in a grid. It also modifies the `TextBox` class to support this behavior by letting the grid handle the left and right arrow keys when the text box has a selection or a caret position at the edge. The changes affect the files `TextBox_Keyboard_Navigation.xaml`, `TextBox_Keyboard_Navigation.xaml.cs`, `TextBox.cs`, and `UITests.Shared.projitems`.

## PR Checklist

### DIFF
--- a/src/SamplesApp/UITests.Shared/UITests.Shared.projitems
+++ b/src/SamplesApp/UITests.Shared/UITests.Shared.projitems
@@ -2418,6 +2418,10 @@
       <SubType>Designer</SubType>
       <Generator>MSBuild:Compile</Generator>
     </Page>
+    <Page Include="$(MSBuildThisFileDirectory)Windows_UI_Xaml_Controls\TextBox\TextBox_Keyboard_Navigation.xaml">
+      <SubType>Designer</SubType>
+      <Generator>MSBuild:Compile</Generator>
+    </Page>
     <Page Include="$(MSBuildThisFileDirectory)Windows_UI_Xaml_Controls\TextBox\TextBox_IsTabStop.xaml">
       <SubType>Designer</SubType>
       <Generator>MSBuild:Compile</Generator>
@@ -5732,6 +5736,9 @@
     <Compile Include="$(MSBuildThisFileDirectory)Windows_UI_Xaml_Controls\ListView\ListView_Snap_Rubberband.xaml.cs">
       <DependentUpon>ListView_Snap_Rubberband.xaml</DependentUpon>
     </Compile>
+    <Compile Include="$(MSBuildThisFileDirectory)Windows_UI_Xaml_Controls\TextBox\TextBox_Keyboard_Navigation.xaml.cs">
+      <DependentUpon>TextBox_Keyboard_Navigation.xaml</DependentUpon>
+    </Compile>
     <Compile Include="$(MSBuildThisFileDirectory)Windows_UI_Xaml_Controls\TextBox\TextBox_PlaceholderForeground.xaml.cs">
       <DependentUpon>TextBox_PlaceholderForeground.xaml</DependentUpon>
     </Compile>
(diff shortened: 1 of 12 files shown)


ramezgerges:
It appears that this issue/PR is connected to a more general need to match the general behaviour of cursor movement inside a cursor as well as keyboard navigation between Uno and UWP. Vertical navigation not working is apparently no the only problem.

Here is a somewhat detailed description of how textboxes work on UWP with regards to keyboard navigation:

* Horizontally:
  1. The textbox handles the left/right keys if the cursor is moving inside the text. At the edges (beginning and end), the KeyDown event isn't handled in the textbox, and bubbles up normally. It's then the outer element's job to decide what to do with that event. If that is a Grid (as in the linked issue and the added sample in this PR), the focus should move to the previous/next element in the grid.
  2. This is still the case if there is nothing to the right/left of the textbox. This clearly makes sense, as the logic for changing focus is not within the textbox.
  3. When a section of text is selected in a textbox, going left/right moves the cursor to the beginning/end of the selection and unselects the section (this is standard behaviour in virtually any textbox in existence).
  4. When a textbox loses focus, it remembers its cursor position (`SelectionStart`) and selection (`SelectionLength`), although the selection indicator (blue background) and cursor are gone. When the textbox is focused again, they come back as they were.
  5. Holding shift and moving left/right to select text in a textbox works close to how moving left/right (without shift) works. If at the edge, behaviour with/without shift is the same.

* Vertically:
  1. Going up works as expected and always bubbles up to the outer element.
  2. Going down is *extremely* weird. First off, If the textbox isn't empty and the cursor is at the beginning, the event is handled in the textbox and doesn't bubble up.
  3. If the cursor is anywhere but the beginning or the end, the cursor moves to the end and is handled.
  4. If some text was selected, it's unselected, the cursor moves to the "end" of the selection and the event is handled. Here, I mean the direction the selection was moving towards. I.e. if one was selecting to the left (e.g. holding shift + left), pressing down would unselect the selection and move the cursor to the left end and vice versa. To make it more complicated, selections using a mouse are less predictable and follow a different logic that I haven't focused on for now.



Currently, this behaviour is not matched closely at all on either Skia.Wpf or Skia.Gtk (I haven't looked into Wasm yet, but I assume the issue won't be as bad). Even more annoyingly, a lot of the textbox manipulation happens natively in Skia, so intercepting selection/cursor changes can be problematic.

MartinZikmund (approved):
Seems good!

PR #12808, opened 2023-07-06
feat: Adding ElementMetadataUpdateHandlerAttribute

### LINKED ISSUE #12807: [Hot Reload] Expose way for internal and third party controls to intercept reloading of types during hot reload
Similar to the way MetadataUpdateHandlerAttribute is used to identify handlers for type updates, assembly level attributes can be defined for handling when the type has been updated.

### ElementUpdateHandlerAttribute 
**ElementUpdateHandlerAttribute(Type elementType, Type handlerType)**
_elementType_ - this is the type (that should inherit from FrameworkElement) that the handler is being registered for
_handlerType_ - the type of the handler to be registered. If a handler is registered for multiple elementTypes, only a single instance of the handler will be created and reused across all elementTypes that it's registered for

eg this registers the FrameUpdateHandler type as the handler for Frame
`[assembly: ElementUpdateHandler(typeof(Frame), typeof(FrameUpdateHandler))]
`

### Handler Methods (at least one should be implemented)
All methods should be static (no state!) methods (at least one) with void return type that the handler can implement

**BeforeVisualTreeUpdate(Type[] updatedTypes)** 
This will get invoked whenever UpdateApplication is invoked but before any updates are applied to the visual tree. This is only invoked once per UpdateApplication, irrespective of the number of types the handler is registered for

**AfterVisualTreeUpdate(Type[] updatedTypes)**
This will get invoked whenever UpdateApplication is invoked after all updates have been applied to the visual tree. This is only invoked once per UpdateApplication, irrespective of the number of types the handler is registered for

**ElementUpdate(FrameowkrElement element, Type[] updatedTypes)**
This is invoked when a specific element is found in the tree. This would be useful if the element holds references to controls that aren't in the visual tree and need to be updated (eg pages in the backstack of a frame)

**BeforeElementReplaced(FrameworkElement oldElement, FrameworkElement newElement, Type[] updatedTypes)**
This is invoked whenever UpdateApplication is invoked before an element is replaced in the visual three. This is invoked for each element in the visual tree that matches a type that has been updated. The oldElement is attached to the visual tree and existing datacontext. The newElement is not attached to the visual tree and won't yet have a data context

**AfterElementReplaced(FrameworkElement oldElement, FrameworkElement newElement, Type[] updatedTypes)**
This is invoked whenever UpdateApplication is invoked after an element is replaced in the visual three. This is invoked for each element in the visual tree that matches a type that has been updated. The oldElement is no longer attached to the visual tree and datacontext will be null. The newElement is attached to the visual tree and will have data context update, either inherited from parent or copies from the oldElement.

## Why is this needed:

Handlers can be registered either by uno itself or by third party packages (including toolkit and extensions) in order to customise the behaviour of hot reload for specific types. For example Frame will need to have a handler to make sure pages on the backstack are appropriately updated

nickrandolph:
GitHub Issue (If applicable): closes #12807

## PR Type

- Feature

No support for intercepting the visual tree updates as part of hot reload

## What is the new behavior?

see #12807 

### <samp>🤖 Generated by Copilot at 87cc951</samp>

This pull request enhances the hot reload feature for UI elements by introducing a new class `ElementUpdateAgent` and a new attribute `ElementMetadataUpdateHandlerAttribute`. These allow defining custom logic for updating and replacing elements in the visual tree when metadata changes, using handler actions that are loaded from the assemblies. The pull request also refactors and simplifies some existing methods for enumerating and updating elements in the `ClientHotReloadProcessor` class.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI.RemoteControl/HotReload/MetadataUpdater/ElementUpdaterAgent.cs
+++ b/src/Uno.UI.RemoteControl/HotReload/MetadataUpdater/ElementUpdaterAgent.cs
@@ -1,0 +1,243 @@
+﻿#if NET6_0_OR_GREATER || __WASM__ || __SKIA__
+#nullable enable
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using System.Reflection;
+using System.Reflection.Metadata;
+using Windows.UI.Xaml;
+using Uno;
+
+
+namespace Uno.UI.RemoteControl.HotReload.MetadataUpdater;
+
+internal sealed class ElementUpdateAgent : IDisposable
+{
+	/// Flags for hot reload handler Types like MVC's HotReloadService.
+	private const DynamicallyAccessedMemberTypes HotReloadHandlerLinkerFlags = DynamicallyAccessedMemberTypes.PublicMethods | DynamicallyAccessedMemberTypes.NonPublicMethods;
+
+	private readonly Action<string> _log;
+	private readonly AssemblyLoadEventHandler _assemblyLoad;
+	private readonly ConcurrentDictionary<Type, ElementUpdateHandlerActions> _elementHandlerActions = new();
+
+	internal const string MetadataUpdaterType = "System.Reflection.Metadata.MetadataUpdater";
+
+	public ElementUpdateAgent(Action<string> log)
+	{
+		_log = log;
+		_assemblyLoad = OnAssemblyLoad;
+		AppDomain.CurrentDomain.AssemblyLoad += _assemblyLoad;
+		LoadElementUpdateHandlerActions();
+	}
+
+	public ImmutableDictionary<Type, ElementUpdateHandlerActions> ElementHandlerActions => _elementHandlerActions.ToImmutableDictionary();
+
+	private void OnAssemblyLoad(object? _, AssemblyLoadEventArgs __)
+	{
+		// This should only be invoked on the (rare) occasion that assemblies
+		// haven't been loaded when the agent is initialized. Since the agent 
+		// is initialized when the first hot reload message is received, this
+		// should mean that all assemblies are already loaded.
+		// For this reason, we don't worry about incrementally loading handlers
+		// we just reload from all assemblies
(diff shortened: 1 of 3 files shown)


src/Uno.UI.RemoteControl/HotReload/MetadataUpdater/ElementUpdaterAgent.cs
```diff
@@ -0,0 +1,284 @@
+﻿// The structure of the ElementUpdaterAgent has been kept similar to the HotReloadAgent, 
+// which is based on the implementation in https://github.com/dotnet/aspnetcore/blob/26e3dfc7f3f3a91ba445ec0f8b1598d12542fb9f/src/Components/WebAssembly/WebAssembly/src/HotReload/HotReloadAgent.cs
+
+
+
+#if NET6_0_OR_GREATER || __WASM__ || __SKIA__
```
Youssef1313, line 6: I think this is always true?

src/Uno.UI.RemoteControl/HotReload/MetadataUpdater/ElementUpdaterAgent.cs
```diff
@@ -0,0 +1,284 @@
...
+		/// after an element is replaced in the visual three. 
+		/// This is invoked for each element in the visual tree that 
+		/// matches a type that has been updated. 
+		/// The oldElement is no longer attached to the visual tree and datacontext will be null. 
+		/// The newElement is attached to the visual tree and will have data context update, either inherited from parent or copies from the oldElement.
+		/// </summary>
+		public Action<FrameworkElement, FrameworkElement, Type[]?> AfterElementReplaced { get; set; } = (_, _, _) => { };
+	}
+
+#if NET6_0_OR_GREATER
```
Youssef1313, line 100: Always true?

Youssef1313 (approved):

src/Uno.UI.RemoteControl/HotReload/MetadataUpdater/ElementUpdaterAgent.cs
nickrandolph: I'll remove these from this agent and the hotreloadagent in next PR

Issue #12485, opened 2023-06-01
[Gallery][Wasm/iOS/Android]Checkbox(Material)- Box for all checkbox are invisible.


sakshi173:
### Current behavior

![Checkbox issue](https://github.com/unoplatform/uno/assets/74563602/082291e7-9dc0-4e45-90d5-1ea0c40ee1f7)

### Expected behavior

Checkbo should display.

### How to reproduce it (as minimally and precisely as possible)

1. Launch the Uno.Gallery app.
2. Click on “Checkbox” from the "UI Components" list.
3. Go to the 'material' section.
4. Notice that Box for all checkbox option are invisible.

### Works on UWP/WinUI

None

### Affected platforms

WebAssembly, Android, iOS

sakshi173:
I verified this issue is fixed due to that closing it.

sakshi173:
This issue is happening again, due to that reopening it.
![Checkbo issue new](https://github.com/unoplatform/uno/assets/74563602/662e10a7-3880-4aa0-843d-772de5d2b6be)

jeromelaban:
@Marc-Antoine-Soucy could you point to the PR you made to originally fix this issue?

Marc-Antoine-Soucy:
I started investigating, did other things and then @sakshi173 closed it. I didnt do any pr. I think it was ok when I checked it at the time.

jeromelaban:
This issue was introduced by https://github.com/unoplatform/Uno.Themes/pull/1003, but is likely creating revealing a resource application issue, similar to https://github.com/unoplatform/uno/issues/12706. It is likely that the resource application is done too early, causing the resolution to use app-wide resources instead of parent resolved ones.

PR #12821, opened 2023-07-07
fix(button): fix button ignoring ButtonForeground/PointerOver values

### LINKED ISSUE #12706: [Wasm] ButtonForegroundPointerOver not respected
### Current behavior

On Windows The button below changes text color from yellow to red on mouse over as expected. On Wasm, however, red ButtonForegroundPointerOver color is ignored and the text appears white. 
```
        <Button>
            <Button.Resources>
                <ResourceDictionary>
                    <ResourceDictionary.ThemeDictionaries>
                        <ResourceDictionary x:Key="Dark">
                            <SolidColorBrush x:Key="ButtonForeground" Color="Yellow"/>
                            <SolidColorBrush x:Key="ButtonForegroundPointerOver" Color="Red"/>
                        </ResourceDictionary>
                    </ResourceDictionary.ThemeDictionaries>
                </ResourceDictionary>
            </Button.Resources>
            Click me
        </Button>
```

ButtonForeground is respected on both platforms. 

### Expected behavior

Overriding ButtonForegroundPointerOver (and other  colors https://github.com/microsoft/microsoft-ui-xaml/blob/main/dev/CommonStyles/Button_themeresources.xaml) should work on all platforms.

Create a blank uno app. Place the above code inside the mainpage.xaml.

### Workaround

None found

### Works on UWP/WinUI

Yes

### Environment

### NuGet package version(s)

4.9.20

### Affected platforms

WebAssembly

### IDE

Visual Studio 2022

### IDE version

17.6.4

### Relevant plugins

_No response_

_No response_

ramezgerges:
GitHub Issue (If applicable): closes unoplatform/uno#12706, closes https://github.com/unoplatform/Uno.Themes/issues/1059

## PR Type

- Bugfix

### <samp>🤖 Generated by Copilot at 8f53a7e</samp>

This pull request updates the button styles and resources to match the latest WinUI 2.6 design, adds a sample button that demonstrates how to use theme resources for hover effects, and fixes a bug with the menu flyout subitems. It modifies `Generic.xaml`, `Buttons.xaml`, and `Flyout_TemplatedParent.xaml` files.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
@@ -433,6 +433,7 @@ namespace Windows.UI.Xaml.Controls
 			SetUpdateControlTemplate();
 
 			base.OnLoaded();
+
 		}
 
 		/// <summary>
--- a/src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
@@ -433,11 +433,6 @@ namespace Windows.UI.Xaml.Controls
 			SetUpdateControlTemplate();
 
 			base.OnLoaded();
-
-			foreach (var visualStateGroup in VisualStateManager.GetVisualStateGroups(GetTemplateRoot()))
-			{
-				visualStateGroup.UpdateResourceBindings();
-			}
 		}
 
 		/// <summary>
--- a/src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
@@ -436,10 +436,6 @@ namespace Windows.UI.Xaml.Controls
 
 			foreach (var visualStateGroup in VisualStateManager.GetVisualStateGroups(GetTemplateRoot()))
 			{
-				foreach (var visualState in visualStateGroup.States)
-				{
-					visualState.EnsureMaterialized();
-				}
 				visualStateGroup.UpdateResourceBindings();
 			}
 		}
--- a/src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
@@ -434,6 +434,14 @@ namespace Windows.UI.Xaml.Controls
 
 			base.OnLoaded();
 
+			foreach (var visualStateGroup in VisualStateManager.GetVisualStateGroups(GetTemplateRoot()))
+			{
+				foreach (var visualState in visualStateGroup.States)
+				{
+					visualState.EnsureMaterialized();
+				}
+				visualStateGroup.UpdateResourceBindings();
+			}
 		}
 
 		/// <summary>
--- a/src/Uno.UI/UI/Xaml/Style/Generic/Generic.xaml
+++ b/src/Uno.UI/UI/Xaml/Style/Generic/Generic.xaml
@@ -2805,38 +2805,46 @@
 	<Style x:Key="XamlDefaultButton"
 				   TargetType="Button">
         <Setter Property="Background" Value="{ThemeResource ButtonBackground}" />
+        <Setter Property="BackgroundSizing" Value="OuterBorderEdge" />
         <Setter Property="Foreground" Value="{ThemeResource ButtonForeground}" />
         <Setter Property="BorderBrush" Value="{ThemeResource ButtonBorderBrush}" />
-		<Setter Property="BorderThickness"
-				Value="{ThemeResource ButtonBorderThemeThickness}" />
-		<Setter Property="Padding"
-				Value="8,4,8,4" />
-		<Setter Property="HorizontalAlignment"
-				Value="Left" />
-		<Setter Property="VerticalAlignment"
-				Value="Center" />
-		<Setter Property="FontFamily"
-				Value="{ThemeResource ContentControlThemeFontFamily}" />
-		<Setter Property="FontWeight"
-				Value="Normal" />
-		<Setter Property="FontSize"
-				Value="{ThemeResource ControlContentThemeFontSize}" />
-		<Setter Property="UseSystemFocusVisuals"
-				Value="True" />
-		<Setter Property="Template">
-			<Setter.Value>
-				<ControlTemplate TargetType="Button">
-					<Grid x:Name="RootGrid"
-						  Background="{TemplateBinding Background}"
-						  CornerRadius="{TemplateBinding CornerRadius}">
-						<VisualStateManager.VisualStateGroups>
-							<VisualStateGroup x:Name="CommonStates">
-								<VisualState x:Name="Normal">
-									<!--<Storyboard>
-										<PointerUpThemeAnimation Storyboard.TargetName="RootGrid" />
-									</Storyboard>-->
-								</VisualState>
-								<VisualState x:Name="PointerOver">
+        <Setter Property="BorderThickness" Value="{ThemeResource ButtonBorderThemeThickness}" />
+        <Setter Property="Padding" Value="{StaticResource ButtonPadding}" />
+        <Setter Property="HorizontalAlignment" Value="Left" />
+        <Setter Property="VerticalAlignment" Value="Center" />
+        <Setter Property="FontFamily" Value="{ThemeResource ContentControlThemeFontFamily}" />
+        <Setter Property="FontWeight" Value="Normal" />
+        <Setter Property="FontSize" Value="{ThemeResource ControlContentThemeFontSize}" />
+        <Setter Property="UseSystemFocusVisuals" Value="{StaticResource UseSystemFocusVisuals}" />
+        <Setter Property="FocusVisualMargin" Value="-3" />
+        <Setter Property="Template">
+            <Setter.Value>
+                <ControlTemplate TargetType="Button">
+                    <ContentPresenter x:Name="ContentPresenter"
+              Background="{TemplateBinding Background}"
+              BackgroundSizing="{TemplateBinding BackgroundSizing}"
+              BorderBrush="{TemplateBinding BorderBrush}"
+              BorderThickness="{TemplateBinding BorderThickness}"
+              Content="{TemplateBinding Content}"
+              ContentTemplate="{TemplateBinding ContentTemplate}"
+              ContentTransitions="{TemplateBinding ContentTransitions}"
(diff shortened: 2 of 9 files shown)


ramezgerges:
The fix almost works. When on FluentStyle, ButtonForeground works correctly, but ButtonForegroundPointerOver only works when I set the sample to dark mode and then back to light mode. I suspect it have something to do with the initial setup not reading the value.

src/Uno.UI/UI/Xaml/Style/Generic/Generic.xaml
```diff
@@ -2804,12 +2804,9 @@
...
 				   TargetType="Button">
-		<Setter Property="Background"
-				Value="{ThemeResource SystemControlBackgroundBaseLowBrush}" />
-		<Setter Property="Foreground"
-				Value="{ThemeResource SystemControlForegroundBaseHighBrush}" />
-		<Setter Property="BorderBrush"
-				Value="{ThemeResource SystemControlForegroundTransparentBrush}" />
+        <Setter Property="Background" Value="{ThemeResource ButtonBackground}" />
+        <Setter Property="Foreground" Value="{ThemeResource ButtonForeground}" />
+        <Setter Property="BorderBrush" Value="{ThemeResource ButtonBorderBrush}" />
```
Youssef1313, line 2809: Are the following defined somewhere?

```xml
            <!-- Resources for Windows.UI.Xaml.Controls.Button -->
            <StaticResource x:Key="ButtonBackground" ResourceKey="SystemControlBackgroundBaseLowBrush" />
            <StaticResource x:Key="ButtonBackgroundPointerOver" ResourceKey="SystemControlBackgroundBaseLowBrush" />
            <StaticResource x:Key="ButtonBackgroundPressed" ResourceKey="SystemControlBackgroundBaseMediumLowBrush" />
            <StaticResource x:Key="ButtonBackgroundDisabled" ResourceKey="SystemControlBackgroundBaseLowBrush" />
            <StaticResource x:Key="ButtonForeground" ResourceKey="SystemControlForegroundBaseHighBrush" />
            <StaticResource x:Key="ButtonForegroundPointerOver" ResourceKey="SystemControlHighlightBaseHighBrush" />
            <StaticResource x:Key="ButtonForegroundPressed" ResourceKey="SystemControlHighlightBaseHighBrush" />
            <StaticResource x:Key="ButtonForegroundDisabled" ResourceKey="SystemControlDisabledBaseMediumLowBrush" />
            <StaticResource x:Key="ButtonBorderBrush" ResourceKey="SystemControlForegroundTransparentBrush" />
            <StaticResource x:Key="ButtonBorderBrushPointerOver" ResourceKey="SystemControlHighlightBaseMediumLowBrush" />
            <StaticResource x:Key="ButtonBorderBrushPressed" ResourceKey="SystemControlHighlightTransparentBrush" />
            <StaticResource x:Key="ButtonBorderBrushDisabled" ResourceKey="SystemControlDisabledTransparentBrush" />
```

if not, do we need to define them?

src/Uno.UI/UI/Xaml/Style/Generic/Generic.xaml
```diff
@@ -2840,20 +2837,19 @@
...
+                                            <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource ButtonBackgroundPointerOver}" />
+                                        </ObjectAnimationUsingKeyFrames>
+                                        <ObjectAnimationUsingKeyFrames Storyboard.TargetName="ContentPresenter" Storyboard.TargetProperty="BorderBrush">
+                                            <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource ButtonBorderBrushPointerOver}" />
+                                        </ObjectAnimationUsingKeyFrames>
+                                        <ObjectAnimationUsingKeyFrames Storyboard.TargetName="ContentPresenter" Storyboard.TargetProperty="Foreground">
+                                            <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource ButtonForegroundPointerOver}" />
+                                        </ObjectAnimationUsingKeyFrames>
+                                        <PointerUpThemeAnimation Storyboard.TargetName="ContentPresenter" />
```
Youssef1313, line 2850: This probably was commented out for a reason. @MartinZikmund do you know?

src/Uno.UI/UI/Xaml/Style/Generic/Generic.xaml
```diff
@@ -2840,20 +2837,19 @@
 									</Storyboard>-->
 								</VisualState>
 								<VisualState x:Name="PointerOver">
```
Youssef1313, line 2839: It seems you're fixing only PointerOver, but other visual states appear to be broken as well

src/Uno.UI/UI/Xaml/Style/Generic/Generic.xaml
ramezgerges: I discussed this with Martin. As a general rule, we tend to only move over properties from UWP as necessary. In this case, I've copied only what's necessary for the linked issue, but common sense would indicate that it's more reasonable to copy that entire block for the same reasons. @MartinZikmund should probably weigh in on this.

ramezgerges:
I'm not sure if the fix I just pushed is correct or if it's just a cover up for a deeper problem. As far as I've investigated, I think there might a problem with the template being read and setup AFTER the resource bindings are updated. So what've done is wait for the loading to end, and then update the bindings. There is probably another problem with "materializing" the lazy storyboards in VisualStates (although my understanding is very limited on that front).

src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
```diff
@@ -434,6 +434,14 @@ private protected override void OnLoaded()
...
 			base.OnLoaded();
 
+			foreach (var visualStateGroup in VisualStateManager.GetVisualStateGroups(GetTemplateRoot()))
+			{
+				foreach (var visualState in visualStateGroup.States)
+				{
+					visualState.EnsureMaterialized();
+				}
+				visualStateGroup.UpdateResourceBindings();
+			}
```
dr1rrb, line 444: Why are we doing this? Looks like we are explicitly disabling a perf optimization here, if it cause a bug we should either fix it, either just remove that optimization.

cc @jeromelaban (apparently you did it).

src/Uno.UI/UI/Xaml/Style/Generic/Generic.xaml
dr1rrb: Because it's not implemented:
image

Un-commenting it will cause lot of useless messages in the log

```suggestion
                                        <!--<PointerUpThemeAnimation Storyboard.TargetName="ContentPresenter" />-->
```

src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
dr1rrb: Looking at the code, the `EnsureMaterialized` is invoked when  you access to the `StoryBoard` or the `Setters` of the `VisualState`. You should probably be able to move the `UpdateResourceBindings` directly in that method.

ramezgerges: Removing the edits in `Control.cs` and adding a call in `EnsureMaterialized` to `UpdateReasourceBindings` somehow works (even though I've tried some variations of that before). To make it more perplexing, after making that adjustment, this doesn't work:

```csharp
private protected override void OnLoaded()
		{
			SetUpdateControlTemplate();

			base.OnLoaded();

			foreach (var visualStateGroup in VisualStateManager.GetVisualStateGroups(GetTemplateRoot()))
			{
				foreach (var visualState in visualStateGroup.States)
				{
					visualState.EnsureMaterialized();
				}
				// visualStateGroup.UpdateResourceBindings();
			}
		}
```

This also doesn't work:
```csharp
private protected override void OnLoaded()
		{
			SetUpdateControlTemplate();

			base.OnLoaded();

			foreach (var visualStateGroup in VisualStateManager.GetVisualStateGroups(GetTemplateRoot()))
			{
				// foreach (var visualState in visualStateGroup.States)
				// {
				// 	visualState.EnsureMaterialized();
				// }
				visualStateGroup.UpdateResourceBindings();
			}
		}
```

Meaning, redundant calls to ` VisualStateGroup.UpdateResourceBindings()` or `VisualState.EnsureMaterialized` somehow break things (even though one would think they're idemponent). BUT, making the redundant calls together works (!!!). I'm not sure what's causing this. I went quite deep into the call stack, but nothing looks alarming.

ramezgerges: I'm not sure what happened, but I've tested the previous two snippets again, and the second one works, so the behaviour is less crazy than what I previously said. I'll test one more time later and see if everything makes sense.

src/Uno.UI/UI/Xaml/Style/Generic/Generic.xaml
MartinZikmund: In this case you also need to port the resources the style references, as it is very likely they are also missing from our current styles in Uno.UI. These will probably be in `SystemResources.xaml` and there make sure you get those for all themes (Dark/Default, Light and HighContrast)

ramezgerges: I found the resources present in SystemResources.xaml and also in fluent v2. As for the properties in the style itself, I've just updated the entirety of the default Button style to match Windows. A lot of the differences were due to limitations that are no longer present. I've tested these changes minimally, but obviously I couldn't test every single line change. @MartinZikmund, could you make sure everything makes sense now?

jeromelaban: I've adjust the themes updates to avoid unnecessary resources adjustments (they fail some runtime tests), and mention that the Sample needs to be using fluent styles to function properly.

There are also two additional runtime test that validate storyboard and setters override.

jeromelaban: The change does not fix https://github.com/unoplatform/uno/issues/12485, will look further.

jeromelaban: Ok so the good news is that the fix works properly, but it's rather the playground and the gallery that are not working well when https://github.com/unoplatform/Uno.Themes/pull/1003 is applied. Let's follow-up in https://github.com/unoplatform/Uno.Themes/issues/1059.

jeromelaban (approved):

PR #12884, opened 2023-07-13
fix: property inheritance for Control.IsEnabledProperty

### LINKED ISSUE #12758: [iOS, Droid] Enabled Buttons inside a Disabled control are not disabled
### Current behavior

UWP : 
image

ios: 
image
droid: 
image

### Expected behavior

The behaviour on uwp should be replicated on every platform. The buttons shouldn't be able to override their disabled state on mobile.

1. Download this sample
[SampleUnoDisabledButtonBug.zip](https://github.com/unoplatform/uno/files/11909270/SampleUnoDisabledButtonBug.zip)
2. Deploy it on windows first
3. Deploy it on either mobile platforms

### Workaround

Maybe disabling manually the buttons via a view model when the whole control is disabled.

### Works on UWP/WinUI

Yes

### Environment

### NuGet package version(s)

4.8.15

### Affected platforms

Android, iOS, Skia (WPF)

### IDE

Visual Studio 2019

### IDE version

17.6.2

### Relevant plugins

_No response_

_No response_

ramezgerges:
GitHub Issue (If applicable): closes #12758

## PR Type

- Bugfix

### <samp>🤖 Generated by Copilot at ea88803</samp>

This pull request adds a new UI test for the `IsEnabled` property of buttons inside content controls, and fixes a bug in the `DependencyObjectStore` class that prevented the correct coercion of the property. It modifies the `UITests.Shared` project and the `DependencyObjectStore` class, and adds new XAML and code-behind files for the test page.

## PR Checklist

## Other information

### DIFF
--- a/src/SourceGenerators/Uno.UI.SourceGenerators.Internal/DependencyObject/DependencyPropertyGenerator.cs
+++ b/src/SourceGenerators/Uno.UI.SourceGenerators.Internal/DependencyObject/DependencyPropertyGenerator.cs
@@ -467,7 +467,7 @@ namespace Uno.UI.SourceGenerators.DependencyObject
 				{
 					builder.AppendLineIndented($"\t\t, coerceValueCallback: (instance, baseValue, precedence) => (({containingTypeName})instance).Coerce{propertyName}(baseValue, precedence)");
 				}
-				else // data.CoerceCallbackMethod.Parameters.Length == 1
+				else
 				{
 					builder.AppendLineIndented($"\t\t, coerceValueCallback: (instance, baseValue, precedence) => (({containingTypeName})instance).Coerce{propertyName}(baseValue)");
 				}
--- a/src/SourceGenerators/Uno.UI.SourceGenerators.Internal/DependencyObject/DependencyPropertyGenerator.cs
+++ b/src/SourceGenerators/Uno.UI.SourceGenerators.Internal/DependencyObject/DependencyPropertyGenerator.cs
@@ -110,7 +110,7 @@ namespace Uno.UI.SourceGenerators.DependencyObject
 			public string PropertyName { get; }
 			public bool HasPropertySuffix { get; }
 
-			public bool HasCoerceCallback { get; }
+			public IMethodSymbol? CoerceCallbackMethod { get; }
 
 			public bool IsCallbackWithDPChangedArgs { get; }
 			public bool IsCallbackWithDPChangedArgsOnly { get; }
@@ -143,7 +143,7 @@ namespace Uno.UI.SourceGenerators.DependencyObject
 				ContainingTypeFullyQualifiedName = dpSymbol.ContainingType.ToDisplayString(SymbolDisplayFormat.FullyQualifiedFormat);
 				ContainingTypeHintName = HashBuilder.BuildIDFromSymbol(dpSymbol.ContainingType);
 				ContainingTypeHasGetDefaultValueMethod = dpSymbol.ContainingType.GetFirstMethodWithName($"Get{PropertyName}DefaultValue") is not null;
-				HasCoerceCallback = dpSymbol.ContainingType.GetFirstMethodWithName("Coerce" + PropertyName) is not null;
+				CoerceCallbackMethod = dpSymbol.ContainingType.GetFirstMethodWithName("Coerce" + PropertyName);
 
 				MemberSymbolNodeContent = SymbolToNodeString(dpSymbol);
 
@@ -363,9 +363,16 @@ namespace Uno.UI.SourceGenerators.DependencyObject
 				}
 			}
 
-			if (coerceCallback || data.HasCoerceCallback)
+			if (coerceCallback)
 			{
-				builder.AppendLineIndented($"\t\t, coerceValueCallback: (instance, baseValue) => Coerce{propertyName}(instance, ({propertyTypeName})baseValue)");
+				if (data.CoerceCallbackMethod is { } m && m.Parameters.Length == 2)
+				{
+					builder.AppendLineIndented($"\t\t, coerceValueCallback: (instance, baseValue, precedence) => Coerce{propertyName}(instance, ({propertyTypeName})baseValue, precedence)");
+				}
+				else // data.CoerceCallbackMethod.Parameters.Length == 1
+				{
+					builder.AppendLineIndented($"\t\t, coerceValueCallback: (instance, baseValue, precedence) => Coerce{propertyName}(instance, ({propertyTypeName})baseValue)");
+				}
 			}
 
 			if (data.IsCallbackWithDPChangedArgs)
@@ -454,9 +461,16 @@ namespace Uno.UI.SourceGenerators.DependencyObject
 				builder.AppendLineIndented($"\t\t, backingFieldUpdateCallback: On{propertyName}BackingFieldUpdate");
 			}
 
-			if (coerceCallback || data.HasCoerceCallback)
+			if (coerceCallback)
 			{
-				builder.AppendLineIndented($"\t\t, coerceValueCallback: (instance, baseValue) => (({containingTypeName})instance).Coerce{propertyName}(baseValue)");
+				if (data.CoerceCallbackMethod is { } m && m.Parameters.Length == 2)
+				{
+					builder.AppendLineIndented($"\t\t, coerceValueCallback: (instance, baseValue, precedence) => (({containingTypeName})instance).Coerce{propertyName}(baseValue, precedence)");
+				}
+				else // data.CoerceCallbackMethod.Parameters.Length == 1
+				{
+					builder.AppendLineIndented($"\t\t, coerceValueCallback: (instance, baseValue, precedence) => (({containingTypeName})instance).Coerce{propertyName}(baseValue)");
+				}
 			}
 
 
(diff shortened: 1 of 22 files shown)


ramezgerges:
The problem stems from the fact that some properties such as IsEnabled, IsVisible, etc. in parents force their values on children but only in certain situations. For example, here if a parent is not IsEnabled, the children must be not IsEnabled as well, but if IsEnabled isn't true, the child is free to set its own value. The fix in this PR isn't merge-ready obviously, but is meant to showcase a way to sidestep the problem for IsEnabledProperty only. A more reasonable approach would be to give a special marker to properties that exhibit this behaviour and make the handling logic general to all such properties.

src/Uno.UI/UI/Xaml/PropertyMetadata.cs
```diff
@@ -42,6 +42,14 @@ public partial class PropertyMetadata
 		/// </summary>
 		internal bool CoerceWhenUnchanged { get; set; } = true;
 
+		/// <summary>
+		/// Normally, when a property is updated and the coerced value is equal to
+		/// the value being updated to, the coerced value is thrown away. <seealso cref="DependencyObjectStore.ApplyCoercion"/>
+		/// This property is added specifically for Control.IsEnabledProperty in order
+		/// to override this behaviour.
+		/// </summary>
+		internal bool KeepCoercedWhenEquals { get; set; }
```
jeromelaban, line 51: I'd tend to think that moving this property to be a metadata option for FrameworkMetadata would be better:
https://github.com/unoplatform/uno/blob/86522f24d56fe39470def506e13fe6a7d85e352b/src/Uno.UI/UI/Xaml/FrameworkPropertyMetadataOptions.cs

Both CoerceWhenUnchanged and KeepCoercedWhenEquals are internal properties, which makes then appropriate to be in `FrameworkPropertyMetadataOptions`. It would also consume less memory (bit instead of a flag).

Can you try moving both as flags?

dr1rrb (approved):

src/Uno.UI/UI/Xaml/DependencyObjectStore.cs
```diff
@@ -636,7 +636,8 @@ private void ApplyCoercion(DependencyObject actualInstanceAlias, DependencyPrope
 				return;
 			}
 
-			if (!propertyDetails.Metadata.CoerceWhenUnchanged && Equals(previousValue, baseValue))
+			if (Equals(previousValue, baseValue) &&
+				propertyDetails.Metadata is FrameworkPropertyMetadata fpmd && (fpmd.Options & FrameworkPropertyMetadataOptions.CoerceWhenUnchanged) == 0)
```
jeromelaban, line 640: @ramezgerges Let's use extensions methods on Options instead of doing bitwise logic directly here (there are other methods doing this for other flags already)

src/Uno.UI/UI/Xaml/DependencyObjectStore.cs
```diff
@@ -652,7 +653,8 @@ private void ApplyCoercion(DependencyObject actualInstanceAlias, DependencyPrope
 				// Source: https://msdn.microsoft.com/en-us/library/ms745795%28v=vs.110%29.aspx?f=255&MSPPError=-2147217396
 				SetValueInternal(previousValue, DependencyPropertyValuePrecedences.Coercion, propertyDetails);
 			}
-			else if (!Equals(coercedValue, baseValue))
+			else if (!Equals(coercedValue, baseValue) ||
```
jeromelaban, line 656: Same here

src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
```diff
@@ -109,8 +111,33 @@ private protected virtual void ChangeVisualState(bool useTransitions)
...
+		private bool GetIsEnabledValue()
+		{
+			if (!_IsEnabledPropertyBackingFieldSet)
+			{
+				_IsEnabledPropertyBackingField = (bool)GetValue(IsEnabledProperty);
+				_IsEnabledPropertyBackingFieldSet = true;
+			}
+			return _IsEnabledPropertyBackingField;
+		}
+		private void SetIsEnabledValue(bool value) => SetValue(IsEnabledProperty, value);
```
dr1rrb, line 140: As it's now a flag on the enum, we no longer have to disable the generation for that property
```suggestion
		[GeneratedDependencyProperty(DefaultValue = true, ChangedCallback = true, CoerceCallback = true, Options = FrameworkPropertyMetadataOptions.Inherits | FrameworkPropertyMetadataOptions.KeepCoercedWhenEquals)]
		public static DependencyProperty IsEnabledProperty { get; } = CreateIsEnabledProperty();
```

src/Uno.UI/UI/Xaml/DependencyObjectStore.cs
dr1rrb, line 640:
```suggestion
				propertyDetails.Metadata is FrameworkPropertyMetadata fpmd &&
				!fpmd.Options.HasFlag(FrameworkPropertyMetadataOptions.CoerceWhenUnchanged))
```

src/SourceGenerators/Uno.UI.SourceGenerators.Internal/DependencyObject/DependencyPropertyGenerator.cs
```diff
@@ -454,9 +461,16 @@ private static void GenerateProperty(IndentedStringBuilder builder, GenerationCa
...
 
-			if (coerceCallback || data.HasCoerceCallback)
+			if (coerceCallback || data.CoerceCallbackMethodParameterLength is not null)
 			{
-				builder.AppendLineIndented($"\t\t, coerceValueCallback: (instance, baseValue) => (({containingTypeName})instance).Coerce{propertyName}(baseValue)");
+				if (data.CoerceCallbackMethodParameterLength is 2)
+				{
+					builder.AppendLineIndented($"\t\t, coerceValueCallback: (instance, baseValue, precedence) => (({containingTypeName})instance).Coerce{propertyName}(baseValue, precedence)");
+				}
+				else // data.CoerceCallbackMethod.Parameters.Length == 1
```
Youssef1313, line 470:
```suggestion
				else
```

src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
```diff
@@ -166,9 +166,29 @@ private protected void SuppressIsEnabled(bool suppress)
...
+			if (_suppressIsEnabled)
+			{
+				return false;
+			}
+
+			// The baseValue hasn't been set yet, so we need to make sure we're not
+			// reading soon-to-be-outdated values
+			var localValue = precedence < DependencyPropertyValuePrecedences.Inheritance ? // < actually means higher precedence
+				baseValue :
+				((IDependencyObjectStoreProvider)this).Store.GetPropertyDetails(IsEnabledProperty).GetValueUnderPrecedence(DependencyPropertyValuePrecedences.Coercion).value;
```
Youssef1313, line 180: `localValue` is evaluated too early. It could get evaluated without actually being used. We might also want to avoid calling `GetPropertyDetails` twice (and also avoid calling it unnecessarily)

src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
```diff
@@ -173,22 +173,28 @@ private protected virtual object CoerceIsEnabled(object baseValue, DependencyPro
...
-				baseValue :
-				((IDependencyObjectStoreProvider)this).Store.GetPropertyDetails(IsEnabledProperty).GetValue(DependencyPropertyValuePrecedences.Inheritance);
+
+			var parentValue = baseValue;
+			DependencyPropertyDetails propertyDetails = null;
+
+			if (precedence != DependencyPropertyValuePrecedences.Inheritance)
+			{
+				propertyDetails = ((IDependencyObjectStoreProvider)this).Store.GetPropertyDetails(IsEnabledProperty);
+				parentValue = propertyDetails.GetValue(DependencyPropertyValuePrecedences.Inheritance);
```
jeromelaban, line 185: You should be able to use GetValue with the specific precedence parameter to get the value at the precedence that you need. It's best to avoid using property details outside DependencyObjectStore.

ramezgerges: GetValue internally gets the property details anyway. I'm just shortcutting having to go through multiple frames just to get the property details again later. I don't mind it either way, so just confirm again if you want me to swap out the `GetPropertyDetails` calls.

Youssef1313: I think the extra frames aren't an issue.

jeromelaban: Indeed, the extra frames are not an issue. They may get inlined as needed, but it's generally not a concern.

src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
```diff
@@ -166,9 +166,30 @@ private protected void SuppressIsEnabled(bool suppress)
...
+
+			// The baseValue hasn't been set inside PropertyDetails yet, so we need to make sure we're not
+			// reading soon-to-be-outdated values
+
+			var parentValue = precedence == DependencyPropertyValuePrecedences.Inheritance ?
+				baseValue :
+				this.GetValue(IsEnabledProperty, DependencyPropertyValuePrecedences.Inheritance);
+
+			// If the parent is disabled, this control must be disabled as well
+			if (parentValue != DependencyProperty.UnsetValue && !(bool)parentValue!)
```
jeromelaban, line 184: Let's not assume that parentValue is a boolean, it may happen as a result of invalid binding by the user. Let's use `parentValue as bool parentValueAsBool`.

MartinZikmund (approved):
Seems good to me!

jeromelaban (approved):

jeromelaban:
🚀🚀

PR #12969, opened 2023-07-21
fix: Avoid raising `KeyboardHelper` events on non-UI thread

MartinZikmund:

## PR Type

- Bugfix

`KeyboardHelper` is injecting keyboard events during runtime tests, but it is not always happening on the UI thread. This sometimes causes race condition where the visual collection changes while it is being rendered - thus causing runtime tests to fail.

## What is the new behavior?

This is not happening anymore.

### <samp>🤖 Generated by Copilot at a7184a7</samp>

Refactored `PressKeySequence` method to use element's dispatcher thread for key events. This improves the reliability of the integration tests that use this method.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI.RuntimeTests/IntegrationTests/common/TestServices.KeyboardHelper.cs
+++ b/src/Uno.UI.RuntimeTests/IntegrationTests/common/TestServices.KeyboardHelper.cs
@@ -204,7 +204,15 @@ namespace Private.Infrastructure
 
 				async Task RaiseOnElementDispatcherAsync(UIElement element, RoutedEvent routedEvent, RoutedEventArgs args)
 				{
-					if (element.Dispatcher.HasThreadAccess)
+					bool raiseSynchronously = element.Dispatcher.HasThreadAccess;
+#if __WASM__
+					if (!Uno.UI.Dispatching.CoreDispatcher.IsThreadingSupported)
+					{
+						raiseSynchronously = true;
+					}
+#endif
+
+					if (raiseSynchronously)
 					{
 						element.SafeRaiseEvent(routedEvent, args);
 					}
--- a/src/Uno.UI.RuntimeTests/IntegrationTests/common/TestServices.KeyboardHelper.cs
+++ b/src/Uno.UI.RuntimeTests/IntegrationTests/common/TestServices.KeyboardHelper.cs
@@ -1,7 +1,9 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Globalization;
+using System.Threading.Tasks;
 using Windows.System;
+using Windows.UI.Core;
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Input;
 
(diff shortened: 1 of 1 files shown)


Youssef1313 (approved):
Does https://github.com/unoplatform/uno/blob/42915026055520535d5f0e8566a3f4efcca95466/src/Uno.UI.RuntimeTests/IntegrationTests/common/TestServices.InputHelper.cs#L46-L52 suffer a similar issue?

ajpinedam (approved):
Nice!

src/Uno.UI.RuntimeTests/IntegrationTests/common/TestServices.KeyboardHelper.cs
```diff
@@ -118,7 +120,7 @@ public static class KeyboardHelper
 			};
 
 
-			public static void PressKeySequence(string keys, UIElement element = null)
+			public static async void PressKeySequence(string keys, UIElement element = null)
```
Youssef1313, line 123: Oh, callers should probably be updated to `await` the call then. Consider also renaming to `PressKeySequenceAsync`

```suggestion
			public static async Task PressKeySequence(string keys, UIElement element = null)
```

jeromelaban, line 123: Should we dispatch the whole method on the dispatcher? That way we won't need to handle exceptions that may push the app to fail when raised on the thread pool.

MartinZikmund: I don't want to change the signature to stay compatible with WinUI tests, which have the same naming. They are also not awaiting, but they do expect the event to be raised asynchronously

MartinZikmund: I don't want to change the signature to stay compatible with WinUI tests, which have the same naming. They are also not awaiting, but they do expect the event to be raised asynchronously

Youssef1313: Currently this is fire-and-forget, so it should still be problematic

MartinZikmund: I the same thing is happening in case of WinUI though - they also raise the event asynchronously in a "fire and forget" fashion and then just do some "waiting" action afterwards to know when the event was actually processed. The simple solution would be to adjust the API to be fully async, which is doable,, I was just trying to avoid it to keep the "portability" easier

Youssef1313: Then I would make this fire-and-forget from the beginning, ie, make `RaiseOnElementDispatcherAsync` non-async, and use `_ = element.Dispatcher.RunAsync(...)` instead of `await element.Dispatcher.RunAsync(...)`

jeromelaban:
@MartinZikmund could you resolve the conflict?

src/Uno.UI.RuntimeTests/IntegrationTests/common/TestServices.KeyboardHelper.cs
MartinZikmund: Had to keep the original approach, because otherwise we could theoretically schedule the events out of order, breaking the goal of pressing the given keys in a specific sequence...

PR #12987, opened 2023-07-24
fix: [Hot Reload] Update new instance with property values from existing instance

### LINKED ISSUE #12986: [Hot Reload] Property values are lost when new instance of element created
### Current behavior

Application includes MyUserControl
MainPage includes MyUserControl with Height=100
Run application
Modify the MyUserControl -> this will cause a new instance of MyUserControl#1 to be added to the visual tree
MyUserControl#1 instance doesn't have Height set

### Expected behavior

MyUserControl#1.Height=100

_No response_

### Workaround

_No response_

### Works on UWP/WinUI

None

### Environment

_No response_

### NuGet package version(s)

_No response_

### Affected platforms

_No response_

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

nickrandolph:
GitHub Issue (If applicable): closes #12986 

## PR Type

- Bugfix

New instance doesn't have any properties set

## What is the new behavior?

New instance has same property values as the old instance

### <samp>🤖 Generated by Copilot at 46278d4</samp>

This pull request enhances the hot reload feature for XAML views by reducing unnecessary updates, preserving dependency property values, and clearing DataContext values when replacing pages. These changes improve the performance, usability, and consistency of the hot reload experience.

## PR Checklist

- [N/A] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)
- [X] Validated PR `Screenshots Compare Test Run` results.

## Other information

### DIFF
--- a/src/Uno.UI.RemoteControl/HotReload/ClientHotReloadProcessor.Xaml.cs
+++ b/src/Uno.UI.RemoteControl/HotReload/ClientHotReloadProcessor.Xaml.cs
@@ -137,15 +137,19 @@ namespace Uno.UI.RemoteControl.HotReload
 				return;
 			}
 
-			if (newView.DataContext is null
-				&& oldView.DataContext is not null)
+			var commonProperties = from oldProp in DependencyProperty.GetPropertiesForType(oldView.GetType())
+								   from newProp in DependencyProperty.GetPropertiesForType(newView.GetType())
+								   where oldProp.Name == newProp.Name
+								   let oldValue = oldView.GetValue(oldProp)
+								   let newValue = newView.GetValue(newProp)
+								   where oldValue is not null && newValue is null
+								   select (oldProp, oldValue, newProp);
+			foreach (var p in commonProperties)
 			{
-				// If the DataContext is not provided by the page itself, it may
-				// have been provided by an external actor. Copy the value as is
-				// in the DataContext of the new element.
-
-				newView.DataContext = oldView.DataContext;
+				newView.SetValue(p.newProp, p.oldValue);
 			}
+
+
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


src/Uno.UI.RemoteControl/HotReload/ClientHotReloadProcessor.Xaml.cs
```diff
@@ -137,15 +124,19 @@ private static void PropagateProperties(FrameworkElement oldView, FrameworkEleme
...
+								   where oldValue is not null && newValue is null
+								   select (oldProp, oldValue, newProp);
+			foreach (var p in commonProperties)
 			{
-				// If the DataContext is not provided by the page itself, it may
-				// have been provided by an external actor. Copy the value as is
-				// in the DataContext of the new element.
-
-				newView.DataContext = oldView.DataContext;
+				newView.SetValue(p.newProp, p.oldValue);
```
jeromelaban, line 136: @nickrandolph This is likely to cause trouble with properties value precedence, where GetValue reads from any precedence, but SetValue set only for the local precedence. What's the scenario that is being addressed?

nickrandolph: ok, so would be better to constraing GetValue to local precedence then? Essentially we want to make sure any local values that would have been set on the xaml element are set on the new instance

jeromelaban: The intent here is to re-apply local values only, so there's a GetValue with precedence. If the value is not `Unset` then it can be applied to the new instance.

nickrandolph: @jeromelaban can you confirm [my change](https://github.com/unoplatform/uno/blob/93084456efaa6b7df81f35c63a00029bafe4fcb3/src/Uno.UI.RemoteControl/HotReload/ClientHotReloadProcessor.Xaml.cs#L130) is what makes sense here?

jeromelaban: @nickrandolph You'll need to add a condition like `oldValue != DependencyProperty.UnsetValue` as well, because getting the local precedence will only return `null` if the `oldValue` was explicitly locally set to `null`. I actually mislead you on the `GetValue` call, reading again the code. [`oldView.ReadLocalValue()`](https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.dependencyobject.readlocalvalue?view=winrt-22621#returns) is the right call to only get the local value (regardless of any other precedence).

nickrandolph:
replaced by https://github.com/unoplatform/uno/pull/13860

PR #12995, opened 2023-07-24
fix(combobox): fix binding to an old DataSource when both ItemsSource and SelectedItem are bound

### LINKED ISSUE #11956: Bindings invoke disposed ViewModels even though the DataContext was set to null beforehand
### Context

Our client complains about (non-fatal) exceptions (thrown from ViewModels) being reported in their diagnostics dashboard.
After some investigation and some attempted fixes, I was able to resolve the issues for the UWP version of the app. However, the mobile version still have the problem.

### Current behavior

Basically, the `Uno.UI.DataBinding.BindingPropertyHelper` invokes properties on some disposed ViewModels, which leads to exceptions.

We tried to fix this by ensuring that we **set the `DataContext` of pages to `null` before disposing it** (the ViewModel).
This seems to work fine on UWP, but not on mobile.

### Expected behavior

Mobile behaves like Windows.

_No response_

### Workaround

_No response_

### Works on UWP/WinUI

Yes

### Environment

### NuGet package version(s)

4.6.44

### Affected platforms

Android, iOS

### IDE

Visual Studio 2022

### IDE version

17.3.6

### Relevant plugins

_No response_

_No response_

ramezgerges:
GitHub Issue (If applicable): closes #11956

## PR Type

### <samp>🤖 Generated by Copilot at 87d757c</samp>

This pull request enhances the `ImmutableList<T>` and `DependencyProperty` classes, and adds a new feature for synchronizing the selection of an items control with the current item of a collection view. It modifies the `ImmutableList<T>.Insert` method, the `DependencyProperty.Register` method, and the `Selector.IsSynchronizedWithCurrentItemProperty` field. It also fixes a bug in the `ImmutableList<T>` array copying logic and adds a custom comparer class for dependency property bindings.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_ComboBox.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_ComboBox.cs
@@ -569,11 +569,8 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 				dc.Dispose();
 				SUT.DataContext = null;
 
-				// On windows, the old DataContext's getter is not accessed, but surprisingly,
-				// the old setter is accessed
 				await WindowHelper.WaitForIdle();
 				Assert.AreEqual(dc.GetterAccessedAfterDispose, false);
-				Assert.AreEqual(dc.SetterAccessedAfterDispose, true);
 			}
 			finally
 			{
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_ComboBox.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_ComboBox.cs
@@ -569,12 +569,11 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 				dc.Dispose();
 				SUT.DataContext = null;
 
-				// ItemsSource's binding internally updates SelectedItem so this test is
-				// really about SelectedItem's binding updating first, otherwise SelectedItem
-				// will access DisposableDataContext.Item (because two-way binding) after
-				// it's disposed and throw an exception
+				// On windows, the old DataContext's getter is not accessed, but surprisingly,
+				// the old setter is accessed
 				await WindowHelper.WaitForIdle();
-				Assert.AreEqual(dc.AccessedAfterDispose, false);
+				Assert.AreEqual(dc.GetterAccessedAfterDispose, false);
+				Assert.AreEqual(dc.SetterAccessedAfterDispose, true);
 			}
 			finally
 			{
@@ -977,7 +976,8 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 
 			private int _item;
 			private bool _isDisposed;
-			public bool AccessedAfterDispose { get; private set; }
+			public bool GetterAccessedAfterDispose { get; private set; }
+			public bool SetterAccessedAfterDispose { get; private set; }
 
 			public int Item
 			{
@@ -985,7 +985,7 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 				{
 					if (_isDisposed)
 					{
-						AccessedAfterDispose = true;
+						GetterAccessedAfterDispose = true;
 					}
 
 					return _item;
@@ -994,7 +994,7 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 				{
 					if (_isDisposed)
 					{
-						AccessedAfterDispose = true;
+						SetterAccessedAfterDispose = true;
 					}
 
 					_item = value;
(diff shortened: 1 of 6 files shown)


src/Uno.UI/UI/Xaml/DependencyProperty.cs
```diff
@@ -70,6 +71,13 @@ private DependencyProperty(string name, Type propertyType, Type ownerType, Prope
 
 			// Improve the performance of the hash code by
 			CachedHashCode = _name.GetHashCode() ^ ownerType.GetHashCode();
+
+			DependentProperties = dependentProperties.ToImmutableArray();
```
Youssef1313, line 75: Why not make it an immutable array in the first place and avoid allocating both a list an array?

src/Uno.UI/UI/Xaml/DependencyPropertyDetailsCollection.Bindings.cs
```diff
@@ -219,5 +226,27 @@ internal BindingExpression GetBindingExpression(DependencyProperty dependencyPro
 
 			return null;
 		}
+
+		class BindingExpressionComparer : IComparer<BindingExpression>
```
Youssef1313, line 230: This should be singleton.
Also, specify the default access modifier (`private`) and make it sealed as well.

src/Uno.UI/UI/Xaml/DependencyPropertyDetailsCollection.Bindings.cs
```diff
@@ -41,6 +42,7 @@ internal void ApplyTemplatedParent(FrameworkElement templatedParent)
 		/// </summary>
 		public void ApplyDataContext(object dataContext)
 		{
+			_bindings = new (_bindings.OrderBy(x => x, new BindingExpressionComparer()));
```
Youssef1313, line 45: Any performance concerns?

cc @jeromelaban @ebariche

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_ComboBox.cs
```diff
@@ -931,6 +965,45 @@ protected override void OnCollectionChanged(NotifyCollectionChangedEventArgs e)
...
+					return _item;
+				}
+				set
+				{
+					if (_isDisposed)
+					{
+						Assert.Fail();
+					}
+
+					_item = value;
```
Youssef1313, line 998: The exception thrown by `Assert.Fail` will get caught here:

https://github.com/unoplatform/uno/blob/99a8c21d1bdae50965420accd2728991eea111e9/src/Uno.UI/DataBinding/BindingPath.cs#L865

Instead of throwing here, introduce a bool property to `DisposableDataContext` and set it to true in this code path, then assert the property value

Youssef1313: Actually you can just expose `_isDisposed`

src/Uno.UI/UI/Xaml/DependencyPropertyDetailsCollection.Bindings.cs
ramezgerges: This is actually a mistake. I noticed the performance implications and decided to insert in the correct place inside of sorting, but I forgot to remove that line. Adding already creates a new array and copies everything over, so inserting at a certain index and appending should be about the same cost.

src/Uno.UI/DataBinding/BindingPath.cs
```diff
@@ -194,9 +194,7 @@ public string Path
 			set
 			{
 				if (!_disposed
-					&& _value != null
```
jeromelaban, line 197: This is likely going to have an impact on animations. Let's see what the CI says.

ramezgerges: Indeed. We need to make the AreDifferent check for a reason that I so far elude me, but I guess we can use the old Value, but I'm not sure if there are specific cases that should actually call the getter instead of using the old value. Let's see what tests fail after the small change I just did.

src/Uno.UI/DataBinding/BindingPath.cs
```diff
@@ -827,14 +818,6 @@ private void BuildValueGetter()
 				}
 			}
 
-			private void BuildPrecedenceSpecificValueGetter()
-			{
-				if (_precedenceSpecificGetter == null && _dataContextType != null)
-				{
-					_precedenceSpecificGetter = BindingPropertyHelper.GetValueGetter(_dataContextType, PropertyName, _precedence, _allowPrivateMembers);
```
Youssef1313, line 834: This code is probably wrong. I think it should have been `BindingPropertyHelper.GetPrecedenceSpecificValueGetter` but I have no idea if that's the root cause.

jeromelaban: This is indeed interesting. Looking at https://github.com/unoplatform/uno/commit/5fe335c4d25aad59b3c6c76cbb77f945efe5fd5c, this was originally the case, and there this test:

https://github.com/unoplatform/uno/blob/ad7fe3d6e94eaee53242490230a9202efe4abb4c/src/Uno.UI.Tests/BinderTests/Given_Binder.cs#L616

was added at that time and that specifying a precedence to `GetValueGetter` was enough to make it precedence specific.

PR #13031, opened 2023-07-27
fix: Fix toggle switch knob translation

### LINKED ISSUE #12643: [GTK][WASM] Fluent ToggleSwitch animation when pressed isn't correct
### Current behavior

Thumb moving to left side before releasing click, if `IsOn` is set to `true`:

GTK / WASM 

https://github.com/unoplatform/uno/assets/114886335/8031d46a-ab0d-4dbb-833d-b423018e1463

### Expected behavior

The thumb should only move when the click is released as it is on Android, for example:

https://github.com/unoplatform/uno/assets/114886335/ab884a63-919e-4cd5-9b5c-9a84e640438b

Using Uno.UI 4.10.0-dev.337, add a ToggleSwitch control with `DefaultToggleSwitchStyle` style, set `IsOn` to `True` and run the app.

### Workaround

_No response_

### Works on UWP/WinUI

Yes

### Environment

_No response_

### NuGet package version(s)

_No response_

### Affected platforms

WebAssembly, Skia (GTK on Linux/macOS/Windows)

### IDE

Visual Studio 2022

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

Youssef1313:
GitHub Issue (If applicable): closes #12643

## PR Type

- Bugfix

When a ToggleSwitch has `IsOn` set to true, pressing on it (without releasing) shows the knob shifted to the left by 24.

On the ToggleSwitch-level, things go wrong here:

https://github.com/unoplatform/uno/blob/33b35eb25900ef461a2b98253255e5dc47acc8a5/src/Uno.UI/UI/Xaml/Controls/ToggleSwitch/ToggleSwitch.mux.cs#L586

Specifically, `_maxKnobTranslation` and `_minKnobTranslation` end up having the wrong value. The problem is specifically on line 617. The ToggleSwitch is in "On" state, and so _spKnobTransform.X is supposed to be 24, but it reads 0. The root cause is how the animations work.

The correct sequence is:

1. visual states are updated in OnApplyTemplate
1. We go to "On" state
1. "On" state should set the translation to 24
1. During arrange, SizeChangedHandler is invoked reading 24

What happens in reality is:

1. visual states are updated in OnApplyTemplate
1. We go to "On" state
1. Setting the translation to 24 is dispatched
1. During arrange, SizeChangedHandler is invoked reading 0
1. the translation is actually set to 24, but it's too late now.

## What is the new behavior?

When an animation duration is zero, we execute it directly without dispatching

### <samp>🤖 Generated by Copilot at 0b1215b</samp>

This pull request improves the performance and correctness of zero duration animations and simplifies the logic of the `Duration` struct. It changes the `Play` method of the `Timeline` class and the `HasTimeSpan` property of the `Duration` struct in the files `Timeline.animation.cs` and `Duration.cs`.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Duration.cs
+++ b/src/Uno.UI/UI/Xaml/Duration.cs
@@ -28,8 +28,7 @@ namespace Windows.UI.Xaml
 		{
 			get
 			{
-				return this.Type == DurationType.TimeSpan &&
-					this.TimeSpan.CompareTo(TimeSpan.Zero) > 0;
+				return this.Type == DurationType.TimeSpan;
 			}
 		}
 
--- a/src/Uno.UI/UI/Xaml/Media/Animation/Timeline.animation.cs
+++ b/src/Uno.UI/UI/Xaml/Media/Animation/Timeline.animation.cs
@@ -283,22 +283,32 @@ namespace Windows.UI.Xaml.Media.Animation
 			private void Play()
 			{
 				_animator?.Dispose();
-				InitializeAnimator(); // Create the animator
 
-				if (!EnableDependentAnimation && _owner.GetIsDependantAnimation())
-				{ // Don't start the animator its a dependent animation
-					return;
+				if (Duration.HasTimeSpan && Duration.TimeSpan == TimeSpan.Zero)
+				{
+					State = TimelineState.Active;
+					SetValue(ComputeToValue());
+					State = TimelineState.Stopped;
 				}
+				else
+				{
+					InitializeAnimator(); // Create the animator
 
-				UseHardware();//Ensure that the GPU is used for animations
+					if (!EnableDependentAnimation && _owner.GetIsDependantAnimation())
+					{ // Don't start the animator its a dependent animation
+						return;
+					}
 
-				if (BeginTime.HasValue)
-				{ // Set the start delay
-					_animator.StartDelay = (long)BeginTime.Value.TotalMilliseconds;
+					UseHardware();//Ensure that the GPU is used for animations
+
+					if (BeginTime.HasValue)
+					{ // Set the start delay
+						_animator.StartDelay = (long)BeginTime.Value.TotalMilliseconds;
+					}
+
+					_animator.Start();
+					State = TimelineState.Active;
 				}
-
-				_animator.Start();
-				State = TimelineState.Active;
 
 #if __IOS__
 				// On iOS, animations started while the app is in the background will lead to properties in an incoherent state (native static
--- a/src/Uno.UI/UI/Xaml/Media/Animation/Timeline.animation.cs
+++ b/src/Uno.UI/UI/Xaml/Media/Animation/Timeline.animation.cs
@@ -284,31 +284,22 @@ namespace Windows.UI.Xaml.Media.Animation
 			{
 				_animator?.Dispose();
 
-				if (Duration.HasTimeSpan && Duration.TimeSpan == TimeSpan.Zero)
-				{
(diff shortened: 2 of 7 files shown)


src/Uno.UI/UI/Xaml/Media/Animation/Timeline.animation.cs
```diff
@@ -283,22 +283,32 @@ private void OnAnimatorAnimationEnd(object sender, EventArgs e)
 			private void Play()
 			{
 				_animator?.Dispose();
```
Youssef1313, line 285: @dr1rrb Could this dispose twice now? Maybe I should set `_animator` to null in the new zero duration condition?

Youssef1313:
@dr1rrb I just found that we have an `ImmediateAnimator` implementation. Is it better to use it? I think it might not make any difference (other than an extra little non-important allocation with `ImmediateAnimator`)?

carldebilly (approved):

src/Uno.UI/UI/Xaml/Duration.cs
```diff
@@ -28,8 +28,7 @@ public bool HasTimeSpan
 		{
 			get
 			{
-				return this.Type == DurationType.TimeSpan &&
-					this.TimeSpan.CompareTo(TimeSpan.Zero) > 0;
+				return this.Type == DurationType.TimeSpan;
```
dr1rrb, line 31: This modifies the public behavior, is this is valid?

src/Uno.UI/UI/Xaml/Media/Animation/Timeline.animation.cs
dr1rrb: Don't know if it would have a bad effect, but yes for safety I would reset it to `null`, however I would do it here right after the `Dispose()` (and also in the `OnEnd` to be consistent ^^)

dr1rrb (approved):

src/Uno.UI/UI/Xaml/Duration.cs
Youssef1313: @dr1rrb That appears to match Windows from my testing.

dr1rrb:
> @dr1rrb I just found that we have an `ImmediateAnimator` implementation. Is it better to use it? I think it might not make any difference (other than an extra little non-important allocation with `ImmediateAnimator`)?

Actually I think that there is major difference the `ImmediateAnimator` will raise the end event while you won't  https://github.com/unoplatform/uno/blob/master/src/Uno.UI/UI/Xaml/Media/Animation/Animators/ImmediateAnimator.cs#L26. We need to validate the behavior of windows. If the event is raised, I would then use that `ImmediateAnimator` as it would have catch this potential issue.

Youssef1313:
There is a test failure (`When_Rectangle_Fill_Animated`) which appears to be actually expecting the animation end to be invoked (i.e, it waits for the "Completed" event to be fired)

src/Uno.UI/UI/Xaml/Media/Animation/Timeline.animation.cs
Youssef1313: This shouldn't be a concern with the last commit since `InitializeAnimator` will always assign a new value to `_animator`

ajpinedam (approved):

Xiaoy312 (approved):

dr1rrb (approved):

MartinZikmund:
@mergifyio backport legacy/4x

PR #13037, opened 2023-07-27
fix(menuflyout): properly change the DataContext of a MenuFlyout and MenuFlyoutPresenter during opening

### LINKED ISSUE #12490: `MenuFlyout.DataContext` is not propagated to child items when flyout is closed
### Current behavior
MenuFlyout does propagate DataContext change to child elements only when flyout is opened.

In Opening event to correctly propagate DataContext you have to set it null before setting actual value
```
private void Flyout_Opening(object sender, object e)
        {
            if (sender is MenuFlyout flyout && flyout.Target is FrameworkElement elem)
            {
#if HAS_UNO
                // Propagation only when null
                flyout.DataContext = null; 
                flyout.DataContext = elem.DataContext;
#endif
            }
        }
```

But when it's opened it's propagating DataContext well
```
private void Flyout_Opened(object sender, object e)
        {
            if (sender is MenuFlyout flyout && flyout.Target is FrameworkElement elem)
            {
#if HAS_UNO
                flyout.DataContext = elem.DataContext;
#endif
            }
        }
```

uno

Related to https://github.com/unoplatform/uno/issues/11599

### Expected behavior
WinUI does propagate DataContext to child items of MenuFlyout at any state

winui

Repro
[FlyoutUnoAppIssues.zip](https://github.com/unoplatform/uno/files/11624502/FlyoutUnoAppIssues.zip)

### Workaround
Use conditional syntax and set DataContext to null explicity

### Works on UWP/WinUI

Yes

### Environment

_No response_

### NuGet package version(s)

Uno 4.10.0-dev.111

### Affected platforms

WebAssembly, Android, iOS, macOS (AppKit), Mac Catalyst, Skia (WPF), Skia (GTK on Linux/macOS/Windows), Skia (Linux Framebuffer), Skia (Tizen)

### IDE

Visual Studio 2022

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

### LINKED ISSUE #12494: `FlyoutBase.DataContext` should be inherited from an element is shown at
### Current behavior

Investigating https://github.com/unoplatform/uno/issues/12490 made me confusing

Having single shared flyout across elements only last element set to Flyout property of a button inherits DataContext

```
<Page.Resources>
        <local:StringFormatConverter x:Key="StringFormatConverter" />

        <Flyout x:Key="flyout" x:Name="flyout">
            <TextBlock Text="{Binding Text, Converter={StaticResource StringFormatConverter}, ConverterParameter='DataContext: {0}'}" />
        </Flyout>
    </Page.Resources>

    <StackPanel
        HorizontalAlignment="Center"
        VerticalAlignment="Center"
        Spacing="12">
        <Button
            x:Name="btn1"
            Width="220"
            Content="{Binding Text, Converter={StaticResource StringFormatConverter}, ConverterParameter='Button 1 DataContext: {0}'}"
            Flyout="{StaticResource flyout}" />
        <Button
            x:Name="btn2"
            Width="220"
            Content="{Binding Text, Converter={StaticResource StringFormatConverter}, ConverterParameter='Button 2 DataContext: {0}'}"
            Flyout="{StaticResource flyout}" />
        <Button
            x:Name="btn3"
            Width="220"
            Content="{Binding Text, Converter={StaticResource StringFormatConverter}, ConverterParameter='Button 3 DataContext: {0}'}"
            Flyout="{StaticResource flyout}" />
    </StackPanel>
```

uno

### Expected behavior

Flyout should inherit DataContext from the target element is shown at which is the case for WinUI

winui

Repro:
[FlyoutDataContextUnoAppIssues.zip](https://github.com/unoplatform/uno/files/11626139/FlyoutDataContextUnoAppIssues.zip)

### Workaround

Manually set DataContext of a target element for example in open events but keep in mind https://github.com/unoplatform/uno/issues/12490

### Works on UWP/WinUI

Yes

### Environment

### NuGet package version(s)

Uno 4.10.0-dev.111

### Affected platforms

WebAssembly, Android, iOS, macOS (AppKit), Mac Catalyst, Skia (WPF), Skia (GTK on Linux/macOS/Windows), Skia (Linux Framebuffer), Skia (Tizen)

### IDE

Visual Studio 2022

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

ramezgerges:
GitHub Issue (If applicable): closes #12490, closes #12494

## PR Type

- Bugfix

### <samp>🤖 Generated by Copilot at c2917c2</samp>

This pull request adds a new precedence level for the `DataContext` property of the `Popup` child element, to enable data context propagation and avoid conflicts with the local value. It also simplifies the logic for checking the local value in the `Popup` class and updates the `DependencyPropertyValuePrecedences` enum accordingly.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/Popup/Popup.Base.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Popup/Popup.Base.cs
@@ -125,12 +125,15 @@ public partial class Popup : FrameworkElement, IPopup
 
 	partial void OnChildChangedPartial(UIElement oldChild, UIElement newChild)
 	{
-		if (oldChild is IDependencyObjectStoreProvider provider && !_childHasOwnDataContext)
+		if (oldChild is IDependencyObjectStoreProvider provider)
 		{
-			provider.Store.ClearValue(provider.Store.DataContextProperty, DependencyPropertyValuePrecedences.Local);
-			provider.Store.ClearValue(provider.Store.TemplatedParentProperty, DependencyPropertyValuePrecedences.Local);
-			provider.Store.ClearValue(AllowFocusOnInteractionProperty, DependencyPropertyValuePrecedences.Local);
-			provider.Store.ClearValue(AllowFocusWhenDisabledProperty, DependencyPropertyValuePrecedences.Local);
+			provider.Store.ClearValue(provider.Store.DataContextProperty, DependencyPropertyValuePrecedences.Popup);
+			if (!_childHasOwnDataContext)
+			{
+				provider.Store.ClearValue(provider.Store.TemplatedParentProperty, DependencyPropertyValuePrecedences.Local);
+				provider.Store.ClearValue(AllowFocusOnInteractionProperty, DependencyPropertyValuePrecedences.Local);
+				provider.Store.ClearValue(AllowFocusWhenDisabledProperty, DependencyPropertyValuePrecedences.Local);
+			}
 		}
 
 		UpdateDataContext(null);
@@ -154,26 +157,23 @@ public partial class Popup : FrameworkElement, IPopup
 
 	private void UpdateDataContext(DependencyPropertyChangedEventArgs e)
 	{
-		_childHasOwnDataContext = false;
 		if (PropagatesDataContextToChild && Child is IDependencyObjectStoreProvider provider)
 		{
-			var dataContextProperty = provider.Store.ReadLocalValue(provider.Store.DataContextProperty);
+			var childLocalValue = provider.Store.ReadLocalValue(provider.Store.DataContextProperty);
+			_childHasOwnDataContext = childLocalValue != null && childLocalValue != DependencyProperty.UnsetValue;
 
-			var shouldClearValue = e != null && e.NewValue == null && dataContextProperty == e.OldValue;
+			var childPopupValue = provider.Store.GetValue(provider.Store.DataContextProperty, DependencyPropertyValuePrecedences.Popup);
+
+			var shouldClearValue = e != null && e.NewValue == null && childPopupValue == e.OldValue;
 			if (shouldClearValue)
 			{
 				//In this case we are clearing the DataContext that was previously set by the Popup
 				//This usually occurs when the owner of the Popup is removed from the Visual Tree
-				provider.Store.ClearValue(provider.Store.DataContextProperty, DependencyPropertyValuePrecedences.Local);
-			}
-			else if (dataContextProperty != null && dataContextProperty != DependencyProperty.UnsetValue)
-			{
-				// Child already has locally set DataContext, we shouldn't overwrite it.
-				_childHasOwnDataContext = true;
+				provider.Store.ClearValue(provider.Store.DataContextProperty, DependencyPropertyValuePrecedences.Popup);
 			}
 			else
 			{
-				provider.Store.SetValue(provider.Store.DataContextProperty, this.DataContext, DependencyPropertyValuePrecedences.Local);
+				provider.Store.SetValue(provider.Store.DataContextProperty, this.DataContext, DependencyPropertyValuePrecedences.Popup);
 			}
 		}
 	}
--- a/src/Uno.UI/UI/Xaml/Controls/Popup/Popup.Base.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Popup/Popup.Base.cs
@@ -23,8 +23,6 @@ public partial class Popup : FrameworkElement, IPopup
 	private FocusState _lastFocusState = FocusState.Unfocused;
 	private IDisposable _openPopupRegistration;
 
-	private bool _childHasOwnDataContext;
-
 	public event EventHandler<object> Closed;
 	public event EventHandler<object> Opened;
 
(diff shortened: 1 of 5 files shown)


ramezgerges:
Copy from internal discussion:

It seems that the problem lies in the way Popup propagates DataContext updates to its Child. The DataContext of the Child is not updated if it's already locally set on the Child, which looks innocent at first, but actually, the Popup sets the DataContext of the Child at the Local precedence (also shown above), so if the DataContext of the Popup changes, it won't update its Child's DataContext, despite the fact that the Popup itself set that Local value for the child. There might be a need to introduce a new precedence value that's weaker than Local to distinguish between actually Local values and values set by the Popup.

ramezgerges:
I'm having trouble getting a runtime test to work. manually calling flyout.Open doesn't actually open the flyout for some reason.

ramezgerges:
I looked into the button, and IIUC the button calls flyout.ShowAt(this), but a call to flyout.Open() itself won't position the flyout next to the button.

src/Uno.UI/UI/Xaml/Controls/Popup/Popup.Base.cs
```diff
@@ -154,26 +157,23 @@ protected internal override void OnTemplatedParentChanged(DependencyPropertyChan
...
-			else if (dataContextProperty != null && dataContextProperty != DependencyProperty.UnsetValue)
-			{
-				// Child already has locally set DataContext, we shouldn't overwrite it.
-				_childHasOwnDataContext = true;
+				provider.Store.ClearValue(provider.Store.DataContextProperty, DependencyPropertyValuePrecedences.Popup);
 			}
 			else
 			{
-				provider.Store.SetValue(provider.Store.DataContextProperty, this.DataContext, DependencyPropertyValuePrecedences.Local);
+				provider.Store.SetValue(provider.Store.DataContextProperty, this.DataContext, DependencyPropertyValuePrecedences.Popup);
```
Youssef1313, line 176: I'm wondering if adding `FrameworkPropertyMetadataOptions.LogicalChild` to `Popup.ChildProperty` would make the inheritance precedence work seamlessly?

ramezgerges: Didn't work. The flyout becomes RootVisual's child later, so it inherits a null datacontext. Further attempts at playing with `ValueDoesNotInheritDataContext` also didn't work

src/Uno.UI/UI/Xaml/DependencyPropertyValuePrecedences.cs
```diff
@@ -27,6 +27,11 @@ public enum DependencyPropertyValuePrecedences : int
 		/// </summary>
 		Local,
 
+		/// <summary>
+		/// Values set by a Popup to its child
+		/// </summary>
+		Popup,
```
MartinZikmund, line 33: I think we can't include the PR as is in 4x, as this is a breaking change (as it changes a public enum and also moves all the subsequent members +1 in value. What do you think @Youssef1313 ?

Youssef1313: Yup it will be very risky to backport.

Youssef1313: Well, actually while public, I don't see how it can be used by external code, but I can't guarantee that 😄 

So let's stay on the safe side

jeromelaban: I don't think we should be adding a new precedence level for this particular scenario. It has a wide performance impact for a very specific use-case. Maybe there's another condition we can use to avoid this particular issue.

src/Uno.UI/UI/Xaml/Controls/Popup/Popup.Base.cs
jeromelaban, line 176: When the issue happens, at which precedence is the offending datacontext being set? In some cases, the datacontext may be incorrectly set at the style level, because of the popup being created during a style for instance.

ramezgerges: I'm not sure what you mean by the "offending" data context. The problem is that the popup sets the DataContext for its logical child at the Local level, which causes an ambiguity between the DataContext set by the popup and an actual Local Datacontext. The other obvious option is to set the DataContext at the inheritance level, except that the the logical child's visual parent (mostly PopupRoot) will override the Inheritance DataContext (to null).

jeromelaban: Then we may need to break that link from the `PopupRoot`, and prevent the DataContext to change. Depending on how the parent is set, we may able to use this:
https://github.com/unoplatform/uno/blob/06ebb2b56d9df2fb0b35be688fd9900483efd870/src/Uno.UI/UI/Xaml/FrameworkPropertyMetadataOptions.cs#L35

ramezgerges: That didn't work out, I'm not exactly sure why. @Youssef1313 had a clever idea to set the DataContext on `PopupPanel` instead of the child itself (and then the child can inherit those values). `PopupPanel` is internal, so we don't need to adhere to what happens on WinUI. I'm not sure if this has any unintended side effects. Let's see what CI says.

jeromelaban: Looks like it did help! Very nice.

jeromelaban (approved):

PR #13088, opened 2023-08-02
Allow RCServer to load processors in specific Assem…

mrlacey:
…blyLoadContext

GitHub Issue (If applicable): _none_

## PR Type

- Refactoring

Each RemoteControlServer instance uses a separate AssemblyLoadContext.

## What is the new behavior?

`AssemblyLoadContext`s are now pooled and grouped by `appInstanceId`.
This allows for the registering of multiple Processors with the same Id which can access the same static instances of types.

## Copilot Summary

copilot:summary

## PR Checklist

-  Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)
-  [Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
-  Validated PR `Screenshots Compare Test Run` results.
-  Associated with an issue (GitHub or internal) and uses the [automatic close keywords](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue).

## Other information

Created as an alternative to and in response to feedback on #12707

### DIFF
--- a/src/Uno.UI.RemoteControl.Host/RemoteControlServer.cs
+++ b/src/Uno.UI.RemoteControl.Host/RemoteControlServer.cs
@@ -12,40 +12,24 @@ using Uno.Extensions;
 using Uno.UI.RemoteControl.Messages;
 using System.Runtime.Loader;
 using Microsoft.Extensions.Configuration;
+using System.Reflection;
+using System.Runtime.Remoting.Messaging;
 
 namespace Uno.UI.RemoteControl.Host
 {
 	internal class RemoteControlServer : IRemoteControlServer, IDisposable
 	{
-		private readonly Dictionary<string, IServerProcessor> _processors = new Dictionary<string, IServerProcessor>();
+		private readonly Dictionary<string, IServerProcessor> _processors = new();
 
 		private System.Reflection.Assembly? _loopAssembly;
 		private WebSocket? _socket;
+		private string? _appInstanceId;
 		private readonly IConfiguration _configuration;
-		private readonly AssemblyLoadContext _loadContext;
+		private readonly static Dictionary<string, AssemblyLoadContext> _loadContexts = new();
 
 		public RemoteControlServer(IConfiguration configuration)
 		{
 			_configuration = configuration;
-			_loadContext = new AssemblyLoadContext(null, isCollectible: true);
-			_loadContext.Unloading += (e) =>
-			{
-				if (this.Log().IsEnabled(LogLevel.Debug))
-				{
-					this.Log().LogDebug("Unloading assembly context");
-				}
-			};
-			_loadContext.Resolving += (context, assemblyName) =>
-			{
-				if (_loopAssembly is not null)
-				{
-					return context.LoadFromAssemblyPath(Path.Combine(Path.GetDirectoryName(_loopAssembly.Location)!, assemblyName.Name + ".dll"));
-				}
-				else
-				{
-					return context.LoadFromAssemblyName(assemblyName);
-				}
-			};
 
 			if (this.Log().IsEnabled(LogLevel.Debug))
 			{
@@ -56,10 +40,74 @@ namespace Uno.UI.RemoteControl.Host
 		string IRemoteControlServer.GetServerConfiguration(string key)
 			=> _configuration[key] ?? "";
 
-		private void RegisterProcessor(IServerProcessor hotReloadProcessor)
+		private void EnsureLoadContextForApplicationId(string applicationId)
 		{
-			_processors[hotReloadProcessor.Scope] = hotReloadProcessor;
+			if (_loadContexts.ContainsKey(applicationId))
+			{
+				return;
+			}
+
+			var loadContext = new AssemblyLoadContext(null, isCollectible: true);
+			loadContext.Unloading += (e) =>
+			{
+				if (this.Log().IsEnabled(LogLevel.Debug))
+				{
+					this.Log().LogDebug("Unloading assembly context");
+				}
+			};
+
+			// Add custom resolving so we can find dependencies even when the processor assembly
+			// is built for a different .net version than the host process.
+			loadContext.Resolving += (context, assemblyName) =>
+			{
+				if (_loopAssembly is not null)
+				{
+					try
+					{
+						var loc = _loopAssembly.Location;
+						if (!string.IsNullOrWhiteSpace(loc))
+						{
+							var dir = Path.GetDirectoryName(loc);
+							if (!string.IsNullOrEmpty(dir))
+							{
+								var relPath = Path.Combine(dir, assemblyName.Name + ".dll");
+								if (File.Exists(relPath))
+								{
+									return context.LoadFromAssemblyPath(relPath);
+								}
+							}
+						}
+						else
+						{
+							if (this.Log().IsEnabled(LogLevel.Debug))
+							{
+								this.Log().LogDebug("Failed for identify location of dependency: {assemblyName}", assemblyName);
+							}
+						}
+					}
+					catch (Exception exc)
+					{
+						if (this.Log().IsEnabled(LogLevel.Error))
+						{
+							this.Log().LogError(exc, "Failed for load dependency: {assemblyName}", assemblyName);
+						}
+					}
+				}
+				return context.LoadFromAssemblyName(assemblyName);
+			};
+
+			_loadContexts.Add(applicationId, loadContext);
 		}
+
+		private AssemblyLoadContext GetAssemblyLoadContext(string applicationId)
+		{
+			EnsureLoadContextForApplicationId(applicationId);
+
+			return _loadContexts[applicationId];
+		}
+
+		private void RegisterProcessor(IServerProcessor hotReloadProcessor)
+			=> _processors[hotReloadProcessor.Scope] = hotReloadProcessor;
 
 		public async Task Run(WebSocket socket, CancellationToken ct)
 		{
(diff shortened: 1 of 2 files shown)


src/Uno.UI.RemoteControl.Host/RemoteControlServer.cs
```diff
@@ -12,55 +12,103 @@
...
 	{
-		private readonly Dictionary<string, IServerProcessor> _processors = new Dictionary<string, IServerProcessor>();
+		private readonly Dictionary<string, IServerProcessor> _processors = new();
 
 		private System.Reflection.Assembly? _loopAssembly;
 		private WebSocket? _socket;
+		private string? _appInstanceId;
 		private readonly IConfiguration _configuration;
-		private readonly AssemblyLoadContext _loadContext;
+		private readonly static Dictionary<string, AssemblyLoadContext> _loadContexts = new();
```
nickrandolph, line 28: Can we move this to the top of the class to separate it away from the instance fields?

src/Uno.UI.RemoteControl.Host/RemoteControlServer.cs
```diff
@@ -12,55 +12,103 @@
...
-		string IRemoteControlServer.GetServerConfiguration(string key)
-			=> _configuration[key] ?? "";
-
-		private void RegisterProcessor(IServerProcessor hotReloadProcessor)
+		private AssemblyLoadContext GetAssemblyLoadContext(string applicationId)
 		{
-			_processors[hotReloadProcessor.Scope] = hotReloadProcessor;
+			EnsureLoadContextForApplicationId(applicationId);
+
+			return _loadContexts[applicationId];
```
nickrandolph, line 106: Prevent multiple lookups on _loadContexts by returning the context from EnsureLoadContextForApplicaitonId

src/Uno.UI.RemoteControl.Host/RemoteControlServer.cs
```diff
@@ -12,55 +12,103 @@
...
+						var loc = _loopAssembly.Location;
+						if (!string.IsNullOrWhiteSpace(loc))
+						{
+							var dir = Path.GetDirectoryName(loc);
+							if (!string.IsNullOrEmpty(dir))
+							{
+								var relPath = Path.Combine(dir, assemblyName.Name + ".dll");
+								if (File.Exists(relPath))
+								{
+									return context.LoadFromAssemblyPath(relPath);
```
nickrandolph, line 76: It would be worth adding a Trace in here to log out the path to the assembly that's been loaded

src/Uno.UI.RemoteControl.Host/RemoteControlServer.cs
```diff
@@ -12,55 +12,103 @@
...
+			=> _configuration[key] ?? "";
+
+		private void EnsureLoadContextForApplicationId(string applicationId)
+		{
+			if (_loadContexts.ContainsKey(applicationId))
+			{
+				return;
+			}
+
+			var loadContext = new AssemblyLoadContext(null, isCollectible: true);
```
nickrandolph, line 50: We might as well set the name of the AssemblyLoadContext to be the applicationId

src/Uno.UI.RemoteControl.Host/RemoteControlServer.cs
```diff
@@ -152,19 +205,19 @@ private void ProcessDiscoveryFrame(Frame frame)
...
 					if (this.Log().IsEnabled(LogLevel.Debug))
 					{
-						this.Log().LogDebug($"Discovery: Loading {file}");
+						this.Log().LogDebug("Discovery: Loading {File}", file);
 					}
 
 					try
 					{
-						assemblies.Add(_loadContext.LoadFromAssemblyPath(file));
+						assemblies.Add(asmblyLoadContext.LoadFromAssemblyPath(file));
```
jeromelaban, line 213:
```suggestion
						assemblies.Add(assemblyLoadContext.LoadFromAssemblyPath(file));
```

src/Uno.UI.RemoteControl.Host/RemoteControlServer.cs
```diff
@@ -152,19 +205,19 @@ private void ProcessDiscoveryFrame(Frame frame)
...
 					}
 					catch (Exception exc)
 					{
 						// With additional processors there may be duplicates of assemblies already loaded
 						if (this.Log().IsEnabled(LogLevel.Debug))
 						{
-							this.Log().LogDebug($"Failed to load assembly {file} : {exc}");
+							this.Log().LogDebug("Failed to load assembly {File} : {Exc}", file, exc);
```
jeromelaban, line 220: Did you want to use indexed formatting? Or is there something with the interpolation that does not work?

src/Uno.UI.RemoteControl.Host/RemoteControlServer.cs
```diff
@@ -240,6 +299,24 @@ public void Dispose()
...
 			}
+
+			// Unload any AssemblyLoadContexts not being used by any current connection
+			foreach (var appId in _appInstanceIds)
+			{
+				if (_loadContexts.TryGetValue(appId, out var lc))
+				{
+					if (--lc.Count == 0)
+					{
+						_loadContexts[appId].Context.Unload();
```
jeromelaban, line 310: We may want to put this around a try/catch, just to be safe.

src/Uno.UI.RemoteControl.Host/RemoteControlServer.cs
```diff
@@ -2,64 +2,110 @@
...
 	internal class RemoteControlServer : IRemoteControlServer, IDisposable
 	{
-		private readonly Dictionary<string, IServerProcessor> _processors = new Dictionary<string, IServerProcessor>();
+		private readonly static Dictionary<string, (AssemblyLoadContext Context, int Count)> _loadContexts = new();
+		private readonly Dictionary<string, IServerProcessor> _processors = new();
 
-		private System.Reflection.Assembly? _loopAssembly;
+		private string? _resolveAssemblyLocation;
 		private WebSocket? _socket;
+		private readonly List<string> _appInstanceIds = new();
```
jeromelaban, line 26: Can't we use `_loadContexts` to get the list of appInstanceIds? This would also avoid to have to synchronize both collections (it's enumerated from multiple threads in Dispose)

src/Uno.UI.RemoteControl.Host/RemoteControlServer.cs
```diff
@@ -240,6 +299,24 @@ public void Dispose()
 			{
 				processor.Value.Dispose();
 			}
+
+			// Unload any AssemblyLoadContexts not being used by any current connection
+			foreach (var appId in _appInstanceIds)
```
jeromelaban, line 304: We're enumerating a shared resource, we should move the `_appInstanceIds` to a ConcurrentDictionary.

src/Uno.UI.RemoteControl.Host/RemoteControlServer.cs
mrlacey: One of the analyzers was complaining about using interpolation (I think because it's slower).

src/Uno.UI.RemoteControl.Host/RemoteControlServer.cs
mrlacey: the _appInstanceIds are specific to the RCServer instance. These are tracked separately so we know what to decrement and possibly remove from the (shared) _loadContexts.

src/Uno.UI.RemoteControl.Host/RemoteControlServer.cs
mrlacey: the _appInstanceIds are unique to the Server instance. It's the _loadContext that's shared so I'll make that concurrent.

I'm also open to renaming these variables if you have better suggestions.

src/Uno.UI.RemoteControl.Host/RemoteControlServer.cs
jeromelaban: I'm interested to know which one, it should be fine to use.

mrlacey: https://learn.microsoft.com/en-gb/dotnet/fundamentals/code-analysis/quality-rules/ca2254

src/Uno.UI.RemoteControl.Host/RemoteControlServer.cs
```diff
@@ -240,6 +312,47 @@ public void Dispose()
...
+			{
+				if (_loadContexts.TryGetValue(appId, out var lc))
+				{
+					if (lc.Count > 1)
+					{
+						_loadContexts.TryUpdate(appId, (lc.Context, lc.Count - 1), (lc.Context, lc.Count));
```
jeromelaban, line 323: You'll probably need to check for the return value as the update may fail.

src/Uno.UI.RemoteControl.Host/RemoteControlServer.cs
```diff
@@ -44,7 +44,13 @@ private AssemblyLoadContext GetAssemblyLoadContext(string applicationId)
...
 			if (_loadContexts.TryGetValue(applicationId, out var lc))
 			{
-				_loadContexts.TryUpdate(applicationId, (lc.Context, lc.Count + 1), (lc.Context, lc.Count));
+				if (!_loadContexts.TryUpdate(applicationId, (lc.Context, lc.Count + 1), (lc.Context, lc.Count)))
+				{
+					if (this.Log().IsEnabled(LogLevel.Debug))
+					{
+						this.Log().LogDebug($"Failed to increment the count of load context connections for '{applicationId}'");
+					}
+				}
```
jeromelaban, line 53: This may cause memory leaks in case of concurrency issues. 

To fix this, we either need to either:
- move to optimistic retries to redo the `TryGetValue` to get an updated value until the `TryUpdate` works (for this location and the one below)
- use a synchronization root with a simple `private object _loadContextGate = new()` and move to a `Dictionary<,>` instead of a `ConcurrentDictionary<,>` then use `lock()` statements where appropriate.

Either implementation can work, but I'd lean over to the lock for now to make the logic simpler (as long as there is only  
`_loadContext` to synchronize)

jeromelaban (approved):

jeromelaban:
@mrlacey The build is failing for an unrelated failure, which will be mitigated by https://github.com/unoplatform/uno/pull/13372

PR #13101, opened 2023-08-04
perf: Avoid HashtableEx for long-living collections

Youssef1313:

## PR Type

### <samp>🤖 Generated by Copilot at 8df4540</samp>

This pull request optimizes and simplifies the code related to dependency properties in the Uno.UI project. It replaces the `HashtableEx` class with the `Dictionary` class in several places, improving performance and memory usage. It also enhances the `FastTypeComparer` class by using a generic interface and a nullability attribute. It deletes three unused files that contained obsolete code.

## PR Checklist

## Other information


src/Uno.UI/UI/Xaml/DependencyProperty.cs
```diff
@@ -28,16 +29,16 @@ public sealed partial class DependencyProperty
...
 		/// A static <see cref="PropertyCacheEntry"/> used for lookups and avoid creating new instances. This assumes that uses are non-reentrant.
 		/// </summary>
 		private readonly static PropertyCacheEntry _searchPropertyCacheEntry = new();
 
 
-		private readonly static FrameworkPropertiesForTypeDictionary _getFrameworkPropertiesForType = new FrameworkPropertiesForTypeDictionary();
+		private readonly static Dictionary<Uno.CachedTuple<Type, Windows.UI.Xaml.FrameworkPropertyMetadataOptions>, DependencyProperty[]> _getFrameworkPropertiesForType = new();
 
 		private readonly PropertyMetadata _ownerTypeMetadata; // For perf consideration, we keep direct ref the metadata for the owner type
 		private readonly PropertyMetadataDictionary _metadata = new PropertyMetadataDictionary();
```
Youssef1313, line 44: This is quite interesting to optimize.

In most cases, `_metadata` is going to be single element only. It only starts to have multiple elements when `OverrideMetadata` is called, which is not the common case.

This field should potentially be lazy, and when null we only consider `_ownerTypeMetadata`.

jeromelaban:
I don't think we'll be implementing this as-is, that is until Wasm gets better at handling generics that are in the critical path. We may need to alter the way these collections are implemented based on per-platform characteristics, but we're not there yet.

jeromelaban:
@Youssef1313 I'll close this PR for now. We'll revisit when generics will get better, maybe net9?

PR #13124, opened 2023-08-04
Allow the user to disable recycling behavior of controls

### LINKED ISSUE #13125: Template-recycled `TextBox` can reset `TwoWay` data-binding source
## For which Platform:

- [ ] WebAssembly
- [ ] Android
- [ ] iOS
- [ ] macOS (AppKit)
- [ ] Mac Catalyst
- [ ] Skia
  - [ ] WPF
  - [ ] GTK (Linux)
  - [ ] Linux Framebuffer
  - [ ] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

MartinZikmund:
GitHub Issue (If applicable): closes #13125

## PR Type

- Bugfix
- Feature

`TextBox`, `ToggleSwitch` and `ToggleButton` always change their state when recycled, which can cause issues in some scenarios.

## What is the new behavior?

Recycling no longer happening if user sets `IsResetOnTemplateRecycle` to `false` on the control.

### <samp>🤖 Generated by Copilot at a7d8c6e</samp>

No summary available (Limit exceeded: required to process 59554 tokens, but only 50000 are allowed per call)

## PR Checklist

## Other information


src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.cs
```diff
@@ -1088,10 +1088,20 @@ internal void OnSelectionChanged()
 
 		public void OnTemplateRecycled()
 		{
+			if (!IsResetOnTemplateRecycle)
+			{
+				return;
+			}
```
Youssef1313, line 1094: I very much think this is going to lead to even more problematic behaviors.

Youssef1313 (commented):
I'd like us to remove template pooling completely, evaluate the performance, and then find a different way to solve the performance issues that pooling aims to solve.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.cs
```diff
@@ -677,5 +677,26 @@ public async Task When_ReadOnly_Toggled_Repeatedly()
...
+			textBox.OnTemplateRecycled();
+
+			Assert.IsTrue(string.IsNullOrEmpty(textBox.Text));
+
+			textBox.IsTemplateRecyclingEnabled = false;
+
+			textBox.Text = "Hello";
+			textBox.OnTemplateRecycled();
+
+			Assert.AreEqual(text, textBox.Text);
```
Youssef1313, line 698: This is not the original issue right?

I mean it's totally okay for textBox.Text to be empty here, but **the way** the text changed has unintended side effects, is that right? If so, do we have a way to reproduce the unintended side effects in the test?

At the very least, the test needs an explanation comment.

What I would like to do is have a new way of changing property value that doesn't push two way bindings.

That would probably be a boolean property on `IFrameworkTemplatePoolAware` that gets set and unset before and after [this line](https://github.com/unoplatform/uno/blob/79887869b418b12d7fcb17730a27af753b9adb7c/src/Uno.UI/UI/Xaml/FrameworkTemplatePool.cs#L349C26-L349C42).

Then somewhere in the binding engine have check along the lines of:

```csharp
if (ActualInstance is IFrameworkTemplatePoolAware templateAwareElement && templateAwareElement.IsRecycling)
{
    // skip pushing bindings when properties change.
}
```

ajpinedam (approved):

PR #13214, opened 2023-08-11
fix(commandbar): Fix rendering issues with commandbar

### LINKED ISSUE #13213: `CommandBar` Popup renders incorrectly on initial load, and has wrong vertical offset on all loads
### Current behavior

Normally, CommandBars have a "SecondaryCommands" button that expands the bar and opens a popup with a list of secondary commands when pressed, like this
2d04ac44-d20e-4932-983a-105de7a1c511
What actually happens on Uno is this
29ce9adf-1e76-4d16-aef8-47051d77b624

The popup is opened where a collapsed CommandBar would be, not where an expanded one is.

Also, on the initial render, the Popup doesn't take its full correct width, but is only wide enough to contain the words "Like" and "Dislike", without the right padding present in the above screenshots

### Expected behavior

2d04ac44-d20e-4932-983a-105de7a1c511

<Grid Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">
		<Grid.RowDefinitions>
			<RowDefinition Height="Auto"/>
			<RowDefinition Height="*"/>
			<RowDefinition Height="Auto"/>
		</Grid.RowDefinitions>

		<TextBlock Text="Hello, world !" Margin="20" FontSize="30" />
		
		<CommandBar Grid.Row="2">
			<AppBarToggleButton Icon="Shuffle" Label="Shuffle" Click="AppBarButton_Click" />
			<AppBarToggleButton Icon="RepeatAll" Label="Repeat" Click="AppBarButton_Click"/>
			<AppBarSeparator/>
			<AppBarButton Icon="Back" Label="Back" Click="AppBarButton_Click"/>
			<AppBarButton Icon="Stop" Label="Stop" Click="AppBarButton_Click"/>
			<AppBarButton Icon="Play" Label="Play" Click="AppBarButton_Click"/>
			<AppBarButton Icon="Forward" Label="Forward" Click="AppBarButton_Click"/>

			<CommandBar.SecondaryCommands>
				<AppBarButton Label="Like" Click="AppBarButton_Click"/>
				<AppBarButton Label="Dislike" Click="AppBarButton_Click"/>
			</CommandBar.SecondaryCommands>

			<CommandBar.Content>
				<TextBlock Text="Now playing..." Margin="12,14"/>
			</CommandBar.Content>
		</CommandBar>

	</Grid>

### Workaround

_No response_

### Works on UWP/WinUI

None

### Environment

_No response_

### NuGet package version(s)

_No response_

### Affected platforms

_No response_

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

ramezgerges:
GitHub Issue (If applicable): closes #13213

## PR Type

- Bugfix

### <samp>🤖 Generated by Copilot at 275d34b</samp>

This pull request improves the layout and positioning of the `CommandBar` and `Popup` controls. It fixes a bug in the `CommandBar` overflow popup and adds a feature to the `Popup` custom layouter.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/CommandBar/CommandBar.Partial.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/CommandBar/CommandBar.Partial.cs
@@ -261,6 +261,12 @@ namespace Windows.UI.Xaml.Controls
 
 			base.OnApplyTemplate();
 
+			// UNO Specific
+			if (m_tpContentRoot is { })
+			{
+				m_tpContentRoot.RenderTransform.Changed += RenderTransformOnChanged;
+			}
+
 			GetTemplatePart("PrimaryItemsControl", out m_tpPrimaryItemsControlPart);
 			GetTemplatePart("SecondaryItemsControl", out m_tpSecondaryItemsControlPart);
 
@@ -359,6 +365,16 @@ namespace Windows.UI.Xaml.Controls
 		}
 
 		#region Uno Only
+
+		private void RenderTransformOnChanged(object? sender, EventArgs args)
+		{
+			// The popup's PopupPanel depends on the RenderTransform of ContentRoot.
+			// When RenderTransform is changed, only the rendering is updated, without
+			// invalidating Arrange. However, the PopupPanel updates its anchor position in
+			// ArrangeOverride, so we need to add this hack to forcibly invalidate its
+			// Arrange to recalculate its position after the RenderTransform of ContentRoot is updated.
+			m_tpOverflowPopup?.PopupPanel.InvalidateArrange();
+		}
 
 		private void OnOverflowPopupClosed(object? sender, object e)
 		{
--- a/src/Uno.UI/UI/Xaml/Controls/CommandBar/CommandBar.Partial.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/CommandBar/CommandBar.Partial.cs
@@ -71,6 +71,9 @@ namespace Windows.UI.Xaml.Controls
 		private readonly SerialDisposable m_overflowPopupClosedEventHandler = new SerialDisposable();
 		private readonly SerialDisposable m_overflowPresenterItemsPresenterKeyDownEventHandler = new SerialDisposable();
 
+		// Uno specific
+		private readonly SerialDisposable m_contentRootRenderTransformChangedEventHandler = new SerialDisposable();
+
 		private readonly SerialDisposable m_accessKeyInvokedEventHandler = new SerialDisposable();
 		private readonly SerialDisposable m_overflowPopupOpenedEventHandler = new SerialDisposable();
 
@@ -140,6 +143,8 @@ namespace Windows.UI.Xaml.Controls
 			m_overflowContentSizeChangedEventHandler.Disposable = null;
 			m_overflowPopupClosedEventHandler.Disposable = null;
 			m_overflowPresenterItemsPresenterKeyDownEventHandler.Disposable = null;
+
+			m_contentRootRenderTransformChangedEventHandler.Disposable = null;
 
 			m_accessKeyInvokedEventHandler.Disposable = null;
 			m_overflowPopupOpenedEventHandler.Disposable = null;
@@ -240,6 +245,7 @@ namespace Windows.UI.Xaml.Controls
 			if (m_tpOverflowContentRoot is { })
 			{
 				m_overflowContentSizeChangedEventHandler.Disposable = null;
+				m_contentRootRenderTransformChangedEventHandler.Disposable = null;
 			}
 
 			if (m_tpOverflowPresenterItemsPresenter is { })
@@ -265,6 +271,7 @@ namespace Windows.UI.Xaml.Controls
 			if (m_tpContentRoot is { })
 			{
 				m_tpContentRoot.RenderTransform.Changed += RenderTransformOnChanged;
+				m_contentRootRenderTransformChangedEventHandler.Disposable = Disposable.Create(() => m_tpContentRoot.RenderTransform.Changed -= RenderTransformOnChanged);
 			}
 
 			GetTemplatePart("PrimaryItemsControl", out m_tpPrimaryItemsControlPart);
--- a/src/Uno.UI/UI/Xaml/Controls/Popup/PopupPanel.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Popup/PopupPanel.cs
@@ -97,10 +97,7 @@ internal partial class PopupPanel : Panel
 	{
 		// Note: Here finalSize is expected to be the be the size of the window
 
-		var child = this.GetChildren().FirstOrDefault();
-
-		// the Child could have been measured without PopupPanel itself being measured, so _lastMeasuredSize could be outdated
-		var size = Popup.CustomLayouter is { } || child is null ? _lastMeasuredSize : LayoutInformation.GetDesiredSize(child);
+		var size = _lastMeasuredSize;
 
 		// Usually this check is achieved by the parent, but as this Panel
 		// is injected at the root (it's a subView of the Window), we make sure
@@ -111,6 +108,7 @@ internal partial class PopupPanel : Panel
 			size = finalSize = default;
 		}
 
+		var child = this.GetChildren().FirstOrDefault();
 		if (child == null)
 		{
 			return finalSize;
--- a/src/Uno.UI/UI/Xaml/Controls/Popup/PopupPanel.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Popup/PopupPanel.cs
@@ -97,7 +97,7 @@ internal partial class PopupPanel : Panel
 	{
 		// Note: Here finalSize is expected to be the be the size of the window
 
-		var size = _lastMeasuredSize;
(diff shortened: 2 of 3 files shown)


src/Uno.UI/UI/Xaml/Controls/CommandBar/CommandBar.Partial.cs
```diff
@@ -261,6 +261,12 @@ protected override void OnApplyTemplate()
 
 			base.OnApplyTemplate();
 
+			// UNO Specific
+			if (m_tpContentRoot is { })
+			{
+				m_tpContentRoot.RenderTransform.Changed += RenderTransformOnChanged;
```
Youssef1313, line 267: This will leak.

src/Uno.UI/UI/Xaml/Controls/CommandBar/CommandBar.Partial.cs
```diff
@@ -261,6 +261,12 @@ protected override void OnApplyTemplate()
 
 			base.OnApplyTemplate();
 
+			// UNO Specific
```
Youssef1313, line 264: @MartinZikmund Do we need an issue to track down the underlying root cause and remove the Uno-specific code?

src/Uno.UI/UI/Xaml/Controls/Popup/PopupPanel.cs
```diff
@@ -97,7 +97,10 @@ protected override Size ArrangeOverride(Size finalSize)
 	{
 		// Note: Here finalSize is expected to be the be the size of the window
 
-		var size = _lastMeasuredSize;
+		var child = this.GetChildren().FirstOrDefault();
+
+		// the Child could have been measured without PopupPanel itself being measured, so _lastMeasuredSize could be outdated
+		var size = Popup.CustomLayouter is { } || child is null ? _lastMeasuredSize : LayoutInformation.GetDesiredSize(child);
```
Youssef1313, line 103: I'm worried about the performance characteristics of this change.
Are we able to detect exactly when this measurement becomes outdated and invalidate the measurement only for this case?

ramezgerges: This shouldn't affect performance pretty much at all. The call to GetDesiredSize gets an already measured size, and doesn't do any remeasurements. Anyway, I'm working on a different solution

ramezgerges:
There seems to be a new problem with the CommandBar expanding down. The `LayoutRoot` Grid is not collapsing after it's opened.

There's an animation in the template that updates the Clip.Transform.Y of the `LayoutRoot` but that animation is never triggered. The problem is that there's a condition at
https://github.com/unoplatform/uno/blob/90cc85e16b8e3c9573c406f8c87acf9293a61a22/src/Uno.UI/UI/Xaml/Media/Animation/Timeline.cs#L406
that doesn't trigger the animation if the View is null, which is the case here since the Transform is affecting a Clip area, and not a view. It's not very clear atm why the condition is present.

dr1rrb:
> It's not very clear atm why the condition is present.

Because when animation was implemented few years ago, they assumed that an animation and transforms was tightly link to a view (and a single view) ... which is wrong (both can be re-used to animate/transform multiple dependency objects).

But here the only effect is that the animation is considered as "dependent" (understand "will invalidate the layout") ... which is indeed probably the case on uno, but not on windows (cf. https://learn.microsoft.com/en-us/windows/uwp/debug-test-perf/optimize-animations-and-media#use-independent-instead-of-dependent-animations). We should just match the windows behavior and allow animations on the `Clip` property even if not flagged with the [`AllowDependentAnimations`](https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.media.animation.timeline.allowdependentanimations?view=winrt-22621).

(Note: As a test to validate that it's  the only issue, you can change the template to set the `AllowDependentAnimations` flag to `true`)

src/Uno.UI/UI/Xaml/UIElement.Layout.crossruntime.cs
```diff
@@ -255,6 +257,13 @@ private void DoMeasure(Size availableSize)
 #endif
 					{
 						LayoutInformation.SetAvailableSize(this, availableSize);
+
+						if (previousDesiredSize != LayoutInformation.GetDesiredSize(this) &&
+							!IsMeasureDirtyPathDisabled &&
+							this.GetParent() is UIElement parent)
+						{
+							parent.InvalidateMeasure();
```
carldebilly, line 265: Adding explanations...

```suggestion
							// When the desired size is changing, it is important to make sure to redo the measurement
							// on the parent element. This case usually occurs if the .Measure() is invoked outside
							// the measure+arrange phases.

							parent.InvalidateMeasure();
```

carldebilly (approved):

ramezgerges:
@carldebilly @dr1rrb 

The previous fix had the problem of measuring elements multiple times. The new fix keeps state of which element(s) are currently being measured as part of the layout cycle and which are just getting measure due to any random call. Measure calls now only invalidate flags if they happen during a Parent's Measure call and that parent itself recursively satisfies this same condition. There are almost definitely some edge cases (although nothing that I can immediately think of), but I guess this should make sense for the overwhelming majority of cases.

src/Uno.UI/UI/Xaml/UIElement.Layout.crossruntime.cs
```diff
@@ -235,14 +243,29 @@ private void DoMeasure(Size availableSize)
...
+						if (partOfLayoutCycle)
+						{
+							foreach (var child in GetChildren())
+							{
+								child.CurrentlyBeingMeasuredByParentInLayoutCycle = true;
+								disposables.Add(Disposable.Create(() => child.CurrentlyBeingMeasuredByParentInLayoutCycle = false));
+							}
+						}
 						MeasureCore(availableSize);
+						disposables.ForEach(d => d.Dispose());
```
Youssef1313, line 268:
```suggestion
						UIElement[] children = null;
						// This checks makes sure that this is part of the layout cycle,
						// and not just a random call to a parent's Measure()
						if (partOfLayoutCycle)
						{
							children = GetChildren().ToArray();
							foreach (var child in children)
							{
								child.CurrentlyBeingMeasuredByParentInLayoutCycle = true;
							}
						}

						MeasureCore(availableSize);
						if (partOfLayoutCycle)
						{
							foreach (var child in children)
							{
								child.CurrentlyBeingMeasuredByParentInLayoutCycle = false;
							}
						}
```

src/Uno.UI/UI/Xaml/UIElement.Layout.crossruntime.cs
```diff
@@ -235,14 +243,29 @@ private void DoMeasure(Size availableSize)
...
 #if DEBUG
 					try
 #endif
 					{
+						var disposables = new List<IDisposable>();
+						// This checks makes sure that this is part of the layout cycle,
+						// and not just a random call to a parent's Measure()
+						if (partOfLayoutCycle)
+						{
+							foreach (var child in GetChildren())
```
dr1rrb, line 261: I'm really worried about the perf impact of this: Creating a new list on each measure, adding an enumeration of the `Children` collection and finally enumerating the newly created collection ... I would rather check in the `child.Measure` if the direct parent is `CurrentlyBeingMeasuredByParentInLayoutCycle`.

dr1rrb (commented):
CC @carldebilly

Youssef1313 (changes requested):
I don't think the measure changes are the correct changes.

I investigated this, and to me it's looking like an animations issue.

Doing the following change will make the CommanrBar expand properly

image

src/Uno.UI/UI/Xaml/Controls/CommandBar/CommandBar.Partial.cs
```diff
@@ -71,6 +71,9 @@ private enum OverflowInitialFocusItem
 		private readonly SerialDisposable m_overflowPopupClosedEventHandler = new SerialDisposable();
 		private readonly SerialDisposable m_overflowPresenterItemsPresenterKeyDownEventHandler = new SerialDisposable();
 
+		// Uno specific
+		private readonly SerialDisposable m_contentRootRenderTransformChangedEventHandler = new SerialDisposable();
```
Youssef1313, line 75: For now, these changes that are related to invalidating popup panel arrange can go into a separate PR. Also, I'd prefer us to open an issue to track removing this Uno-specific code

ramezgerges:
> I don't think the measure changes are the correct changes.
> 
> I investigated this, and to me it's looking like an animations issue.
> 
> Doing the following change will make the CommanrBar expand properly
> 
> image
> 
> We need to discuss this more.

No, that's completely unrelated. The measure changes are needed because the first time the popup is opened, it opens with an incorrect width. The clipping issue is not related and I don't think changing the template instead of fixing the root cause is a good idea.

Youssef1313:
> the first time the popup is opened, it opens with an incorrect width.

I can't reproduce this.

---

> The clipping issue is not related and I don't think changing the template instead of fixing the root cause is a good idea.

Per my findings, the clipping issue appears to be an animations bug. Changing the template is definitely not the correct fix, but it can serve as a temporary workaround if we couldn't fix the animations issue.

Youssef1313:
PR https://github.com/unoplatform/uno/pull/15620 combined with https://github.com/unoplatform/uno/pull/15619 should fix the issue.

PR #13056, opened 2023-07-30
fix(calendardatepicker): fix open on tap, close on escape

ramezgerges:
GitHub Issue (If applicable): closes #10165

## PR Type

What kind of change does this PR introduce?

### <samp>🤖 Generated by Copilot at e81f61d</samp>

This pull request enhances the `CalendarDatePicker` control with better integration tests and a tap-to-open feature. It also adds a keyboard shortcut to close flyouts and fixes a bug with routed events for popups.

## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)
- [ ] [Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [ ] Validated PR `Screenshots Compare Test Run` results.
- [ ] Contains **NO** breaking changes
- [ ] Associated with an issue (GitHub or internal) and uses the [automatic close keywords](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue).
- [ ] Commits must be following the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) specification.

## Other information

Internal Issue (If applicable):

PR #13241, opened 2023-08-14
CalendarDatePickerIntegrationTests followup

ramezgerges:
This is a followup to https://github.com/unoplatform/uno/pull/13056 and responds to some comments after the merge.

### DIFF
--- a/src/Uno.UI.RuntimeTests/IntegrationTests/common/TestServices.InputHelper.cs
+++ b/src/Uno.UI.RuntimeTests/IntegrationTests/common/TestServices.InputHelper.cs
@@ -48,7 +48,7 @@ namespace Private.Infrastructure
 				throw new System.NotImplementedException();
 			}
 
-			public static void Tap(UIElement element)
+			public static void Tap(UIElement element, uint waitBetweenPressRelease = 0)
 			{
 #if NETFX_CORE || __SKIA__
 				Finger finger = null;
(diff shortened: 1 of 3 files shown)


src/Uno.UI.RuntimeTests/IntegrationTests/common/TestServices.InputHelper.cs
```diff
@@ -60,10 +60,9 @@ public static void Tap(UIElement element)
 					finger.Press(center);
 				});
 
-#if NETFX_CORE
-				// On Windows, We need to wait a bit before releasing for the popup to open.
-				Thread.Sleep(600);
-#endif
+				// UNO specific - we sometimes need to explicitly wait between press and release, e.g. in CalendarDatePickerIntegrationTests
+				Thread.Sleep((int)waitBetweenPressRelease);
```
jeromelaban, line 64: I'm surprised that this has an effect on uno, since this code may be executed on the UI thread, but maybe not in this case?

ramezgerges: I guess I wrote it out poorly. What I meant was that the change was needed even though most of the code is ported from WinUI, so we probably shouldn't need to change it (although the change looks super random). In fact, the waiting parameter workaround is not used in Uno, but specificially for running the tests on Windows.

jeromelaban (approved):

Issue #12030, opened 2023-04-21
Investigate why CA1815 violations were not reported


Youssef1313:
### Current behavior

There was few CA1815 violations (see https://github.com/unoplatform/uno/pull/12029), but CI and local builds were green.

We already set that to error:

https://github.com/unoplatform/uno/blob/bdd49d9b5ec46c205c67b4903bfa9e5d99702627/.editorconfig#L310-L311

### Works on UWP/WinUI

None

Youssef1313:
@mavasani Sorry for the ping, looking at binlog, it looks like everything is right on our end:

![image](https://user-images.githubusercontent.com/31348972/233705513-f241cfce-ec27-4f5d-a82c-27f7d6ca236d.png)

Youssef1313:
IntelliSense shows the error correctly, but not on build.

![image](https://user-images.githubusercontent.com/31348972/233705746-d377b7ee-0fb2-4e4f-85eb-3fe15df68058.png)

Youssef1313:
@sharwell It's empty.

![image](https://user-images.githubusercontent.com/31348972/235284300-e00a6e4e-a3b8-4726-ab1d-43f3769652cf.png)

Yeah I'm planning to move to .globalconfig, but in this case .editorconfig is supposed to just work right?

Youssef1313:
@sharwell fyi this is still an issue with globalconfig and also updated NetAnalyzers package to `8.0.0-preview1.23165.1`

Youssef1313:
I was able to root this down.

Most classes here have an auto-generated partial. When it happens that this is the first location of a symbol, the diagnostic isn't reported because the analyzer is configured to not report on generated code.

MartinZikmund:
@Youssef1313 Weird behavior! It sounds like this is something that the .NET team should address though

Youssef1313:
Related Roslyn issue: https://github.com/dotnet/roslyn/issues/69543

MartinZikmund:
Still relevant

PR #12029, opened 2023-04-21
chore: fix CA1815 violations in Uno.UWP/Uno.Skia

ramezgerges:
Updated some structs to follow CA1815 [0].

[0] https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1815

PR #13264, opened 2023-08-15
Draft PR to see whether CA1815 are now reported in CI or not.

Youssef1313:
Reverts unoplatform/uno#12029

This PR should **NOT** be merged. The expectation is that it fails CI. If it didn't, then https://github.com/unoplatform/uno/issues/12030 is still an issue.

If it failed, then we can close https://github.com/unoplatform/uno/issues/12030


Youssef1313:
Diagnostic is still not reported..

PR #13288, opened 2023-08-16
ci: Adjust winui autoconvert for faster build and use of dotnet build

jeromelaban:
## PR Type

- Build or CI related changes

### <samp>🤖 Generated by Copilot at 5fde81a</samp>

This pull request enhances the `.github/workflows/winui-autoconvert.yml` workflow to make it faster, more reliable, and more flexible. It allows the workflow to run on pull requests, use the latest SDK, and avoid modifying non-master branches.


ajpinedam (approved):
There seems to be an error with Android, but seems looks ok

Issue #9289, opened 2022-07-15
[Skia] `TextBlock` `Hyperlinks` mouse over / pressed coloring and cursor


mikernet:
## What would you like to be added:

TextBlock on Skia currently does not handle mouse over tracking to allow for hyperlinks to change color and cursor to change to a hand when the mouse is over them. They also don't change color when pressed.

Note that https://github.com/unoplatform/uno/issues/9287 should be done first so that the mouse over / pressed colors can be correctly set from theme resources.

## Why is this needed:

User experience and parity with UWP.

## For which Platform:

- [ ] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] WebAssembly renderers for Xamarin.Forms
- [ ] macOS
- [X] Skia
  - [ ] WPF
  - [ ] GTK (Linux)
  - [ ] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

ramezgerges:
Should've been closed by #13371

PR #13371, opened 2023-08-23
fix(hyperlink): change color on hover/press

ramezgerges:
GitHub Issue (If applicable): #9289

## PR Type

- Bugfix

### <samp>🤖 Generated by Copilot at 4a4cb9d</samp>

This pull request adds hover and click effects for `Hyperlink` elements inside a `TextBlock`, using different foreground colors based on the pointer state and the theme. It also refactors some of the existing logic and adds new resources for the colors in `SystemResources.xaml`.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.cs
@@ -896,7 +896,7 @@ namespace Windows.UI.Xaml.Controls
 				return;
 			}
 
-			Debug.Assert(that._hyperlinkOver == null);
+			global::System.Diagnostics.Debug.Assert(that._hyperlinkOver == null);
 
 			var point = e.GetCurrentPoint(that);
 
@@ -913,7 +913,7 @@ namespace Windows.UI.Xaml.Controls
 				return;
 			}
 
-			Debug.Assert(that.FindHyperlinkAt(e.GetCurrentPoint(that).Position) == null);
+			global::System.Diagnostics.Debug.Assert(that.FindHyperlinkAt(e.GetCurrentPoint(that).Position) == null);
 
 			that._hyperlinkOver?.ReleasePointerOver(e.Pointer);
 			that._hyperlinkOver = null;
@@ -939,7 +939,7 @@ namespace Windows.UI.Xaml.Controls
 
 		private void UpdateHyperlinks()
 		{
-			Debug.Assert(_hyperlinkOver is null || _hyperlinks.Where(h => h.hyperlink == _hyperlinkOver).Count() == 1);
+			global::System.Diagnostics.Debug.Assert(_hyperlinkOver is null || _hyperlinks.Where(h => h.hyperlink == _hyperlinkOver).Count() == 1);
 
 			if (UseInlinesFastPath) // i.e. no Inlines
 			{
@@ -1021,7 +1021,7 @@ namespace Windows.UI.Xaml.Controls
 			{
 				var hasHyperlink = _hyperlinks.Count > 0;
 
-				Debug.Assert(!(!hasHyperlink && _hyperlinkOver is { }));
+				global::System.Diagnostics.Debug.Assert(!(!hasHyperlink && _hyperlinkOver is { }));
 
 				return hasHyperlink;
 			}
--- a/src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.cs
@@ -807,7 +807,7 @@ namespace Windows.UI.Xaml.Controls
 #else
 		private static readonly PointerEventHandler OnPointerPressed = (object sender, PointerRoutedEventArgs e) =>
 		{
-			if (!(sender is TextBlock that) || !that.HasHyperlink)
+			if (sender is not TextBlock { HasHyperlink: true } that)
 			{
 				return;
 			}
@@ -829,7 +829,7 @@ namespace Windows.UI.Xaml.Controls
 				return;
 			}
 
-			hyperlink.SetPointerPressed(e.Pointer, that.Resources.ThemeDictionaries);
+			hyperlink.SetPointerPressed(e.Pointer);
 			e.Handled = true;
 			that.CompleteGesture(); // Make sure to mute Tapped
 		};
@@ -869,7 +869,7 @@ namespace Windows.UI.Xaml.Controls
 
 		private static readonly PointerEventHandler OnPointerMoved = (sender, e) =>
 		{
-			if (!(sender is TextBlock that) || !that.HasHyperlink)
+			if (sender is not TextBlock { HasHyperlink: true } that)
 			{
 				return;
 			}
@@ -881,13 +881,13 @@ namespace Windows.UI.Xaml.Controls
 			{
 				that._hyperlinkOver?.ReleasePointerOver(e.Pointer);
 				that._hyperlinkOver = hyperlink;
-				hyperlink?.SetPointerOver(e.Pointer, that.Resources.ThemeDictionaries);
+				hyperlink?.SetPointerOver(e.Pointer);
 			}
 		};
 
 		private static readonly PointerEventHandler OnPointerEntered = (sender, e) =>
 		{
-			if (!(sender is TextBlock that) || !that.HasHyperlink)
+			if (sender is not TextBlock { HasHyperlink: true } that)
 			{
 				return;
 			}
@@ -899,12 +899,12 @@ namespace Windows.UI.Xaml.Controls
 			var hyperlink = that.FindHyperlinkAt(point.Position);
 
 			that._hyperlinkOver = hyperlink;
-			hyperlink?.SetPointerOver(e.Pointer, that.Resources.ThemeDictionaries);
+			hyperlink?.SetPointerOver(e.Pointer);
 		};
 
 		private static readonly PointerEventHandler OnPointerExit = (sender, e) =>
 		{
-			if (!(sender is TextBlock that) || !that.HasHyperlink)
+			if (sender is not TextBlock { HasHyperlink: true } that)
 			{
 				return;
 			}
@@ -935,6 +935,8 @@ namespace Windows.UI.Xaml.Controls
 
 		private void UpdateHyperlinks()
 		{
+			Debug.Assert(_hyperlinkOver is null || _hyperlinks.Where(h => h.hyperlink == _hyperlinkOver).Count() == 1);
+
 			if (UseInlinesFastPath) // i.e. no Inlines
 			{
 				if (HasHyperlink)
@@ -952,6 +954,7 @@ namespace Windows.UI.Xaml.Controls
 						hyperlink.hyperlink.AbortAllPointerState();
 					}
 
+					_hyperlinkOver = null;
 					_hyperlinks.Clear();
 				}
 
@@ -960,6 +963,7 @@ namespace Windows.UI.Xaml.Controls
 
 			var previousHasHyperlinks = HasHyperlink;
 			var previousHyperLinks = _hyperlinks.Select(h => h.hyperlink).ToList();
+			_hyperlinkOver = null;
 			_hyperlinks.Clear();
 
 			var start = 0;
@@ -1011,7 +1015,7 @@ namespace Windows.UI.Xaml.Controls
 		{
 			get
 			{
-				var hasHyperlink = _hyperlinks.Any();
+				var hasHyperlink = _hyperlinks.Count > 0;
 
 				Debug.Assert(!(!hasHyperlink && _hyperlinkOver is { }));
 
--- a/src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.cs
@@ -917,7 +917,7 @@ namespace Windows.UI.Xaml.Controls
 					// Make sure to clear the pressed state of removed hyperlinks
 					foreach (var hyperlink in _hyperlinks)
 					{
-						hyperlink.hyperlink.AbortAllPointerPressed();
+						hyperlink.hyperlink.AbortAllPointerState();
 					}
 
 					_hyperlinks.Clear();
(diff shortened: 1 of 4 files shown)


ramezgerges:
Windows has a weird thing where a hyperlink inside a textblock will still be hovered (and is clickable) even though the pointer is not directly over it. Basically, the last hyperlink in TextBlock.Inlines gets the mouse capture anywhere inside the TextBlock but not when the pointer is directly over any other run. I've skipped this since it gets into native rendering code and there are also inconsistencies that I couldn't find a pattern for.

Also, this PR doesn't deal with the pointer icon changing on hover.

src/Uno.UI/UI/Xaml/Documents/Hyperlink.cs
```diff
@@ -24,6 +25,8 @@ public sealed partial class Hyperlink : Span
 		private readonly IFocusable _focusableHelper;
 #endif
 
+		private const string HyperlinkForegroundPressedKey = "HyperlinkForegroundPressed";
+		private const string HyperlinkForegroundPointerOverKey = "HyperlinkForegroundPointerOver";
```
Youssef1313, line 29: These are only defined for Fluent v2. I wonder how this can work when using non-fluent. I see you already have the right test combination, but really can't see how this can work on non-fluent.

Youssef1313: Oh I see you defined them. Browser's CTRL+F failed me somehow

src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.cs
```diff
@@ -827,7 +829,7 @@ private void UpdateInlines(string text)
 				return;
 			}
 
-			hyperlink.SetPointerPressed(e.Pointer);
+			hyperlink.SetPointerPressed(e.Pointer, that.Resources.ThemeDictionaries);
```
Youssef1313, line 832: No need to pass this I think. You can probably use `Application.Current.Resources`

src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.cs
```diff
@@ -865,14 +867,66 @@ private void UpdateInlines(string text)
 			}
 		};
 
+		private static readonly PointerEventHandler OnPointerMoved = (sender, e) =>
+		{
+			if (!(sender is TextBlock that) || !that.HasHyperlink)
```
Youssef1313, line 872:
```suggestion
			if (sender is not TextBlock that || !that.HasHyperlink)
```

src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.cs
```diff
@@ -865,14 +867,66 @@ private void UpdateInlines(string text)
...
+			{
+				that._hyperlinkOver?.ReleasePointerOver(e.Pointer);
+				that._hyperlinkOver = hyperlink;
+				hyperlink?.SetPointerOver(e.Pointer, that.Resources.ThemeDictionaries);
+			}
+		};
+
+		private static readonly PointerEventHandler OnPointerEntered = (sender, e) =>
+		{
+			if (!(sender is TextBlock that) || !that.HasHyperlink)
```
Youssef1313, line 890:
```suggestion
			if (sender is not TextBlock that || !that.HasHyperlink)
```

src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.cs
```diff
@@ -865,14 +867,66 @@ private void UpdateInlines(string text)
...
+
+			var hyperlink = that.FindHyperlinkAt(point.Position);
+
+			that._hyperlinkOver = hyperlink;
+			hyperlink?.SetPointerOver(e.Pointer, that.Resources.ThemeDictionaries);
+		};
+
+		private static readonly PointerEventHandler OnPointerExit = (sender, e) =>
+		{
+			if (!(sender is TextBlock that) || !that.HasHyperlink)
```
Youssef1313, line 907:
```suggestion
			if (sender is not TextBlock that || !that.HasHyperlink)
```

src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.cs
```diff
@@ -934,17 +991,33 @@ private void UpdateHyperlinks()
...
 				RemoveHandler(PointerCaptureLostEvent, OnPointerCaptureLost);
 			}
 		}
 
-		private bool HasHyperlink => _hyperlinks.Any();
+		private bool HasHyperlink
+		{
+			get
+			{
+				var hasHyperlink = _hyperlinks.Any();
```
Youssef1313, line 1014: Since we are modifying these lines, let's use the more efficient Count property :)
```suggestion
				var hasHyperlink = _hyperlinks.Count;
```

src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.cs
```diff
@@ -934,17 +991,33 @@ private void UpdateHyperlinks()
...
+
+				Debug.Assert(!(!hasHyperlink && _hyperlinkOver is { }));
```
Youssef1313, line 1016: Does this assert hold if you had a hyperlink, on on pointer over you removed it from the TextBlock Inlines?

If the assert wouldn't hold in this case, I suggest deleting it completely.

src/Uno.UI/UI/Xaml/Documents/Hyperlink.cs
```diff
@@ -202,18 +230,28 @@ internal void OnClick()
...
 		{
-#if XAMARIN
-			var normalColor = Brush.GetColorWithOpacity(Foreground, Colors.Transparent).Value;
-			var pressedColor = Color.FromArgb((byte)(normalColor.A / 2), normalColor.R, normalColor.G, normalColor.B);
-			return new SolidColorBrush(pressedColor);
-#else
-			return null;
-#endif
+			if (_pressedState.pointer is { }
+				&& _pressedState.themes.UnoGetValueOrDefault(HyperlinkForegroundPressedKey,null) is Brush brush)
```
Youssef1313, line 238:
```suggestion
				&& _pressedState.themes.UnoGetValueOrDefault(HyperlinkForegroundPressedKey, null) is Brush brush)
```

src/Uno.UI/UI/Xaml/Documents/Hyperlink.cs
```diff
@@ -202,18 +230,28 @@ internal void OnClick()
...
+			{
+				this.SetValue(ForegroundProperty, brush, DependencyPropertyValuePrecedences.Animations);
+			}
+			else
+			{
+				if (_hoveredState.pointer is { }
+					&& _hoveredState.themes.UnoGetValueOrDefault(HyperlinkForegroundPointerOverKey,null) is Brush brush2)
```
Youssef1313, line 245:
```suggestion
					&& _hoveredState.themes.UnoGetValueOrDefault(HyperlinkForegroundPointerOverKey, null) is Brush brush2)
```

src/Uno.UI/UI/Xaml/Documents/Hyperlink.cs
```diff
@@ -202,18 +230,28 @@ internal void OnClick()
...
-			return null;
-#endif
+			if (_pressedState.pointer is { }
+				&& _pressedState.themes.UnoGetValueOrDefault(HyperlinkForegroundPressedKey,null) is Brush brush)
+			{
+				this.SetValue(ForegroundProperty, brush, DependencyPropertyValuePrecedences.Animations);
+			}
+			else
+			{
+				if (_hoveredState.pointer is { }
```
Youssef1313, line 244: This should probably be outside the `else`. That is, check for **both** `_pressedState` and `_hoveredState`. I *think* it might be possible to have both of them set (e.g, multi-pointers)?

I'm not very sure though, but it's safest to check for both anyways.

src/Uno.UI/UI/Xaml/Style/Generic/SystemResources.xaml
```diff
@@ -631,6 +631,8 @@
 			<SolidColorBrush x:Key="HyperlinkPointerOverForegroundThemeBrush" Color="#CC9C72FF" />
 			<SolidColorBrush x:Key="HyperlinkPressedForegroundThemeBrush" Color="#999C72FF" />
 			<StaticResource x:Key="HyperlinkForeground" ResourceKey="SystemControlHyperlinkTextBrush" />
+			<StaticResource x:Key="HyperlinkForegroundPointerOver" ResourceKey="SystemControlPageTextBaseMediumBrush" />
```
Youssef1313, line 634: Should this be `SystemControlHyperlinkBaseMediumBrush`?

src/Uno.UI/UI/Xaml/Style/Generic/SystemResources.xaml
```diff
@@ -631,6 +631,8 @@
...
+			<StaticResource x:Key="HyperlinkForegroundPressed" ResourceKey="SystemControlHighlightBaseMediumLowBrush" />
```
Youssef1313, line 635: Should this be `SystemControlHighlightBaseMediumLowBrush`?

src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.cs
```diff
@@ -887,12 +941,15 @@ private void UpdateHyperlinks()
...
 					RemoveHandler(PointerCaptureLostEvent, OnPointerCaptureLost);
 
 					// Make sure to clear the pressed state of removed hyperlinks
 					foreach (var hyperlink in _hyperlinks)
 					{
-						hyperlink.hyperlink.AbortAllPointerPressed();
+						hyperlink.hyperlink.AbortAllPointerState();
 					}
 
 					_hyperlinks.Clear();
```
Youssef1313, line 955: I think you might need to set `_hyperlinkOver` to null here?

src/Uno.UI/UI/Xaml/Documents/Hyperlink.cs
ramezgerges: Well, only one value at Precedence=Animations can be set at any time, so even if we have 2 pointers -one pressed and one hovering- we still have to pick 1. Pressed is stronger, so it takes priority.

src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.cs
ramezgerges: Yes, because removing the hyperlink triggers `InlineCollection.OnCollectionChanged` which triggers `TextBlock.InvalidateInlines` which triggers `TextBlock.UpdateHyperlinks` which clears the removes all removed hyperlinks and their state.

src/Uno.UI/UI/Xaml/Style/Generic/SystemResources.xaml
ramezgerges: Yes. My bad.

src/Uno.UI/UI/Xaml/Documents/Hyperlink.cs
```diff
@@ -148,22 +151,45 @@ private void OnUnderlineStyleChanged()
 
 		#endregion
 
+		#region Hover
+		private Pointer _hoveredState;
```
Youssef1313, line 155: nit: `_hoveredPointer` might be a better name

src/Uno.UI/UI/Xaml/Documents/Hyperlink.cs
```diff
@@ -148,22 +151,45 @@ private void OnUnderlineStyleChanged()
...
+			else
+			{
+				return false;
+			}
+		}
+		#endregion
+
 		#region Click
-		private Pointer _pressedPointer;
+		private Pointer _pressedState;
```
Youssef1313, line 178: The old name `_pressedPointer` should be good now that this no longer holds the resources.

jeromelaban:
@ramezgerges looks like there are still a few build errors.

ramezgerges:
https://github.com/unoplatform/uno/blob/f27d70315e2150275e3a40b81449ff9e95e02962/src/Uno.UI/UI/Xaml/Internal/InputManager.Pointers.Managed.cs#L224-L229

We seem to not release the pointer on PointerExited, which is not the case on Windows. On Windows, the moment you're not over an element -even if still pressing- you're no longer pressing the element (and the element is not capturing the pointer in general). This specifically breaks the case where (at least on Gtk) you press on an element, move out of it, alt-tab out, release the pointer, and alt-tab back in. In that case, the element will continue to be "pressed" when it shouldn't be. Another example is hyperlinks. Currently on Uno, if you press a hyperlink, move out of it, move back in, then release, the hyperlink will get clicked (this is also the case with buttons, etc). On Windows, the moment you step out of the hyperlink's bounds, the hyperlink reverts back to the "Normal" visual state and won't be clicked on release.

If we change this, would that break anything? I have a feeling drag and drop might break because of this.

Youssef1313:
@ramezgerges Is this ready for review?

Regarding your previous comment, we shouldn't change anything in the core pointers logic in this PR. This can be opened as a separate issue and be triaged and fixed separately.

dr1rrb:
> We seem to not release the pointer on PointerExited, which is not the case on Windows. On Windows, the moment you're not over an element -even if still pressing- you're no longer pressing the element (and the element is not capturing the pointer in general). This specifically breaks the case where (at least on Gtk) you press on an element, move out of it, alt-tab out, release the pointer, and alt-tab back in. In that case, the element will continue to be "pressed" when it shouldn't be. Another example is hyperlinks.

It's what I observed on windows : the `PointerReleased` event is raised only on elements that are directly under the mouse when you release the pointer, it is not raised on `PointerMoved` when you cross the bounds of the element. (Also it means that you can get a `PointerReleased` without having any `PointerPressed`.) **Internally** we are maintaining a `IsPressed` state which should (or not) be cleared when pointer goes out of the bounds of the element, but this state is almost not used since we refactored visual states to be implemented in each control like on WinUI (instead of relying on a common shared logic which was using that state).

> Currently on Uno, if you press a hyperlink, move out of it, move back in, then release, the hyperlink will get clicked (this is also the case with buttons, etc). On Windows, the moment you step out of the hyperlink's bounds, the hyperlink reverts back to the "Normal" visual state and won't be clicked on release.

Again this is the behavior I observed on UWP for all `Button`. I would be surprised if it has changed, and it's the main difference between `Button.Click` and `Button,.Tap` events.

Youssef1313:
I don't think anything pointer-related should be blocking this PR anyway.

Issue #11809, opened 2023-03-27
ListView Virtualization can cause items to be hidden forever if scrolling too rapidly on WASM


roxk:
### Current behavior

1. Bind a items source of a list view to an observation collection with 200 items.
2. Launch the app, observe the first item's binding.
3. Drag scrollbar to scroll to the bottom of the list. Then scroll back up.
4. Observe the first item in the list isn't the one observed at (2).

### Expected behavior

At (4), the first item is the one observed at (2).

### How to reproduce it (as minimally and precisely as possible)

Minimal repro:
https://github.com/roxk/UnoVirtualizationWasmBug

Video:
https://user-images.githubusercontent.com/16918354/227860838-022f44aa-9d09-43c7-975e-15e7ce27afda.mp4

### Workaround

Disable virtualization by using `StackPanel` as the `ItemsPanel`. But our app currently faces severe performance issues due to this workaround.

### Works on UWP/WinUI

Yes

### Environment

Uno.UI / Uno.UI.WebAssembly / Uno.UI.Skia

### NuGet package version(s)

```
    <PackageReference Include="Microsoft.Windows.Compatibility" Version="5.0.0" />
    <PackageReference Include="Microsoft.Extensions.Logging" Version="5.0.0" />
    <PackageReference Include="Uno.Extensions.Logging.WebAssembly.Console" Version="1.4.0" />
    <PackageReference Include="Uno.UI.WebAssembly" Version="4.8.15" />
    <PackageReference Include="Uno.UI.RemoteControl" Version="4.8.15" Condition="'$(Configuration)'=='Debug'" />
    <PackageReference Include="Uno.UI.Adapter.Microsoft.Extensions.Logging" Version="4.8.15" />
    <PackageReference Include="Uno.Wasm.Bootstrap" Version="7.0.11" />
    <PackageReference Include="Uno.Wasm.Bootstrap.DevServer" Version="7.0.11" />
```

### Affected platforms

WebAssembly

### IDE

Visual Studio 2022

### IDE version

17.5.3

roxk:
There is a similar iOS issue https://github.com/unoplatform/uno/issues/10273. Not sure if the fix is up/relevant to WASM. The issue said this is fixed in 4.6 but I can still repro in `Uno.UI.WebAssembly:4.8.15`.

roxk:
I can reproduce in `SampleApp.Wasm`'s `VirtualizingPanelAdapterIdCache` page. Looking into it.

PR #13409, opened 2023-08-25
fix(listview): miscellaneous fixes to recycled containers

### LINKED ISSUE #9415: [Skia|WASM] [ListView] Sometimes selection background is not set on selected item
### Current behavior

Tested a host a listview with random strings array as item source
Tried to select any item and some random times I get no visible selection highlight but event SelectionChanged is raised and SelectedItem property is changed
Often that happens when you scroll out of scope that items and get them back into the scope

### Expected behavior

Visible selection highlight on item

[UnoApp.zip](https://github.com/unoplatform/uno/files/9251934/UnoApp.zip)
Attached sample repro

### Workaround

-

### Works on UWP/WinUI

Yes

### Environment

### NuGet package version(s)

_No response_

### Affected platforms

WebAssembly, Skia (WPF), Skia (GTK on Linux/macOS/Windows)

### IDE

Visual Studio 2022

### IDE version

_No response_

### Relevant plugins

_No response_

Dunno if it the same behaviour in the mobile and MacOS platforms

ramezgerges:
GitHub Issue (If applicable): closes #9415

## PR Type

- Bugfix

### <samp>🤖 Generated by Copilot at 20350a1</samp>

Improve the handling of item container indices in `UIElement`. Add a comment to clarify the logic in `UIElement.cs`.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_ListViewBase.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_ListViewBase.cs
@@ -1607,8 +1607,8 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 			ImageAssert.HasColorAt(si, 70, 65, Colors.FromARGB("#66AEE7")); // selected
 
 			// check starting from below the second item that nothing looks selected or hovered
-			ImageAssert.DoesNotHaveColorInRectangle(si, new Rectangle(0, 110, si.Width, si.Height - 110), Colors.FromARGB("#66AEE7")); // clicked
-			ImageAssert.DoesNotHaveColorInRectangle(si, new Rectangle(0, 110, si.Width, si.Height - 110), Colors.FromARGB("#FFE6E6E6")); // hovered
+			ImageAssert.DoesNotHaveColorInRectangle(si, new Rectangle(100, 110, si.Width, si.Height - 110), Colors.FromARGB("#66AEE7")); // selected
+			ImageAssert.DoesNotHaveColorInRectangle(si, new Rectangle(100, 110, si.Width, si.Height - 110), Colors.FromARGB("#FFE6E6E6")); // hovered
 		}
 
 		[TestMethod]
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_ListViewBase.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_ListViewBase.cs
@@ -1299,6 +1299,8 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 		[RunsOnUIThread]
 #if __IOS__ || __ANDROID__
 		[Ignore("Disabled because of animated scrolling, even when explicitly requested.")]
+#elif __WASM__
+		[Ignore("Flaky in CI.")]
 #endif
 		public async Task When_Large_List_Scroll_To_End_Then_Back_Up_And_First_Item2()
 		{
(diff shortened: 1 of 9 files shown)


ramezgerges:
There is an additional problem with ListViews and virtualization, where if you scroll rapidly to the end of a long ListView and scroll back up, the first couple of items won't be shown. It's extremely inconsistent and I have to practically follow a very exact sequence to reproduce it. Looking into it now.

Also, I'm finding it rather difficult to write a test for this, since I basically need to test what I see on the screen, not any reference to anything that may or may not be recycled away/back.

src/Uno.UI/UI/Xaml/UIElement.cs
```diff
@@ -1043,6 +1043,12 @@ internal virtual void PrepareForRecycle()
 		{
 			ClearPointerStateOnRecycle();
 
+
+			// It's important to clear the index on the way out (recycle) and not wait to set it on the way in (reuse)
+			// because this property is used e.g. in ItemsControl.ContainerFromIndexInner which is used by e.g.
+			// ListView to get a clicked container.
+			ClearValue(ItemsControl.IndexForItemContainerProperty);
```
jeromelaban, line 1050: Very nice find! We have similar properties if I remember correctly, could you take a look and see if it needs to be done for others?

Youssef1313: This is scary in UIElement. I think `ClearValue` should throw if the actual instance type isn't the same as (or a derived from) the property owner type.

Better to override this in `ItemsControl`. In ItemsControl, there is `CleanUpContainer` method that I'm not sure whether it should be called as well or not. @jeromelaban Is `CleanUpContainer` what you're referring to?

ramezgerges: @Youssef1313  I'm pretty sure that's not true. We set this on ListViewItem anyway and not on ListView itself (also, it's not throwing for me). CleanUpContainer is called conditionally (I'm guessing for performance reasons) where it more thoroughly cleans up the container (e.g. remove the content). We need to clear IndexForItemContainerProperty even if we don't CleanUpContainer. This is the case for the other properties, but we use IndexForItemContainerProperty to get ContainerFromIndex, otherwise we could leave it stale like the other properties until it's reused (where it will get set to a new value anyway).

ramezgerges:
> There is an additional problem with ListViews and virtualization, where if you scroll rapidly to the end of a long ListView and scroll back up, the first couple of items won't be shown. It's extremely inconsistent and I have to practically follow a very exact sequence to reproduce it. Looking into it now.

I'm finding this extremely painful to debug. It reproduces so inconsistently that I have no idea what's causing it or what to break on. If I find a more consistent repro, I will get back to it.

src/Uno.UI/UI/Xaml/UIElement.cs
Youssef1313: Oh that property is attached, so indeed there is no concern with throwing.

> We need to clear IndexForItemContainerProperty even if we don't CleanUpContainer

I didn't mean to clear `IndexForItemContainerProperty` only in `CleanupContainer`. The question is about whether we need to call CleanupContainer in `PrepareForRecycle`

jeromelaban: > Better to override this in `ItemsControl`. In ItemsControl, there is `CleanUpContainer` method that I'm not sure whether it should be called as well or not. @jeromelaban Is `CleanUpContainer` what you're referring to?

Indeed, I misread the location of the change. This should be in `ItemsControl.CleanupContainer` or `ClearContainerForItemOverride`:

https://github.com/unoplatform/uno/blob/66b8c63dc66119c334a7b9a676ffbd9f7a7c9ff6/src/Uno.UI/UI/Xaml/Controls/ItemsControl/ItemsControl.cs#L1105

not in `UIElement`. We may call this conditionally (probably incorrectly), but it may be platform dependent. Which platforms do you see this issue happening?

ramezgerges: It seems like the conditional calling is indeed incorrect, as unconditionally calling `CleanUpContainer` also seems to fix a problem with scrolling that I noticed a couple of times where scrolling all the way down would clip the last 2 items. In that case, putting the `ClearValue` call inside `CleanUpContainer` makes more sense. Ideally, I would prefer if we knew exactly when we could get away with not calling `CleanUpContainer`, but due to the inconsistency of things, I guess we can just clean the container every time until further notice.

The issue seems to appear mainly on WASM and skia.

jeromelaban: Could you point to the conditional call you're seeing?

ramezgerges: `CleanUpContainer` is called if `clearContainer` is true in `RecycleViewForItem` (I've changed it in the last couple of commits, so you'll find it in the before side of the diff)

ramezgerges:
So, CI broke when I called `CleanUpContainer` unconditionally. The tests that were failing were testing how many times the datatemplate factory was called. Since we were always cleaning up the container, I'm assuming it have to recreate the date template everytime? Anyway, I reverted back the change but now `CleanUpContainer` is called in `UnfillLayout` as well, which both fixes the problem and still doesn't recreate the datatemplate more than it needs to.

Also, I've finally found a solution to a problem similar to #11809. Still can't programatically reproduce this, so I can't write tests. 

@jeromelaban is this mergable in this state if CI passes?

ramezgerges:
It seems like a test I added is failing on WASM. I tried locally and it passes, so it seems like it's a CI problem.

jeromelaban:
This is the failure that is observed:

https://dev.azure.com/uno-platform/Uno%20Platform/_TestManagement/Runs?runId=2493484&_a=resultSummary&resultId=100000

Is this what is working locally for you?

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_ListViewBase.cs
```diff
@@ -1293,6 +1295,318 @@ public async Task When_Large_List_Scroll_To_End_Then_Back_Up_And_First_Item()
...
+			mouse.MoveTo(new Point(bounds.GetMidX(), bounds.Top + 10));
+			mouse.Press();
+			mouse.Release();
+
+			await Task.Delay(200);
+			await WindowHelper.WaitForIdle();
+
+			// This problem is incredibly difficult to reproduce. These are the exact
+			// mouse wheel deltas that my trackpad produced when I reproduced this manually.
+			var deltasString = """
```
jeromelaban, line 1425: You can use a `new int[]{ -266, -393, ... }` instead, this will avoid typos and will be foldable in VS.

src/Uno.UI/UI/Xaml/Controls/ListViewBase/VirtualizingPanelLayout.managed.cs
```diff
@@ -514,7 +514,7 @@ void UnfillBackward()
 				while (firstMaterializedLine != null && GetMeasuredEnd(firstMaterializedLine.FirstView) < ExtendedViewportStart + extentAdjustment)
 				{
 					// Dematerialize lines that are entirely outside extended viewport
-					RecycleLine(firstMaterializedLine, clearContainer: false);
+					RecycleLine(firstMaterializedLine, clearContainer: true);
```
jeromelaban, line 517: I would have guess that other tests would have failed as a result, but that does not seem like it. Could you try and reproduce something for this? If containers are cleared, we should be getting cleared datacontext changes for generated containers on a large scroll.

ramezgerges: Indeed, there was a UI test failing because of this. I've converted the UI test to a runtime test (and kept the UI test because it also took screenshots and stuff) and moved the additional cleanup (of visual states and indices) to the lighter and always triggered PrepareForRecycle rather than the heavier and conditionally triggered `CleanUpContainerOverride` which also clears the DataContext etc. This way, we don't need to make any additional calls with `cleanContainer: true` and we still get the most necessary cleanups each time.

jeromelaban (approved):
Looking good, thanks!

PR #13489, opened 2023-09-02
refactor: Cleanup easing functions on Android

Youssef1313:

## PR Type

### <samp>🤖 Generated by Copilot at daa5995</samp>

Simplified the implementation of easing functions for Android animations by using a common base class. Removed redundant Android-specific files for each easing function.

## PR Checklist

## Other information


src/Uno.UI/UI/Xaml/Media/Animation/BackEase.Android.cs
```diff
@@ -1,26 +0,0 @@
...
-		internal override ITimeInterpolator CreateTimeInterpolator()
-		{
-			switch (this.EasingMode)
-			{
-				case EasingMode.EaseIn:
-					return new AnticipateInterpolator((float)this.Amplitude);
-				case EasingMode.EaseOut:
-					return new OvershootInterpolator((float)this.Amplitude);
-				case EasingMode.EaseInOut:
-					return new AnticipateOvershootInterpolator((float)this.Amplitude);
```
jeromelaban, line 20: I'm not sure to understand where that code lives after your changes. What replaces it?

Youssef1313: @jeromelaban In `EasingFunctionBase.Android.cs`, there is a new `AndroidTimeInterpolator` that is returned by `CreateTimeInterpolator`. This simplifies the code, and also ensures we provide a correct time interpolators. For example, in SineEase, the implementation of `CreateTimeInterpolator` was incorrect because there is no such built-in interpolator, but `AndroidTimeInterpolator` will provide the correct implementation for anything.

jeromelaban: Got it. I wonder, do you think it's testable?

Youssef1313: It might be quite tricky since animations are very sensitive to timing. Not sure if there is a good approach to test this. For the actual implementation of easing functions, they are tested here: https://github.com/unoplatform/uno/blob/master/src/Uno.UI.Tests/Windows_UI_Xaml_Media_Animation/Given_Ease.cs

It's just that we now delegate to our implementation instead of creating a "best-effort" built-in interpolator.

dr1rrb: IMO the current tests are sufficient to test the different functions. I would however try to add a test to confirm that the new `AndroidTimeInterpolator` is being used correctly by the platform.

~~My suggestion would be then to create an incredibly long animation and check that the animated object somehow respect the defined easing function (it's probably easier to validate that using easing only for `in` **or** `out`, so you can more easily detect significant changes at the beginning of the animation vs a linear movement). Also thanks to runtime-test, it should be much more reliable than current animation tests which are using UI tests.~~

The test would then consist only to validate that the `GetInterpolation` is being invoked as expected ^^

dr1rrb (approved):

jeromelaban (approved):
Nice work!

PR #13491, opened 2023-09-02
fix!: update SimpleOrientationSensor to match Windows API

ajpinedam:

## PR Type

- Bugfix

### <samp>🤖 Generated by Copilot at 9373635</samp>

This pull request adds support for the `SimpleOrientationSensor` class on iOS and improves its implementation on other platforms. It uses platform-specific APIs to get device motion data and fire the `OrientationChanged` event. It also enhances the performance and resource management of the sensor by using lazy initialization and partial methods.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UWP/Devices/Sensors/SimpleOrientationSensor.Android.cs
+++ b/src/Uno.UWP/Devices/Sensors/SimpleOrientationSensor.Android.cs
@@ -19,9 +19,28 @@ namespace Windows.Devices.Sensors
 {
 	public partial class SimpleOrientationSensor
 	{
+		private SimpleOrientationEventListener? _orientationListener;
+		private SettingsContentObserver? _contentObserver;
+		private SensorManager? _sensorManager;
+		private ISensorEventListener? _sensorEventListener;
+
+		// Threshold, in meters per second squared, closely equivalent to an angle of 25 degrees which correspond to the value when Android detect new screen orientation 
+		private const double _threshold = 4.55;
+		private const Android.Hardware.SensorType _gravitySensorType = Android.Hardware.SensorType.Gravity;
+
 		#region Static
 
 		private static Orientation _defaultDeviceOrientation;
+
+		private static partial SimpleOrientationSensor? TryCreateInstance()
+		{
+			if (Application.Context.GetSystemService(Context.SensorService) is not SensorManager)
+			{
+				return null;
+			}
+
+			return new SimpleOrientationSensor();
+		}
 
 		private static Orientation DefaultDeviceOrientation
 		{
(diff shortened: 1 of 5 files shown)


src/Uno.UWP/Devices/Sensors/SimpleOrientationSensor.Android.cs
```diff
@@ -107,6 +118,81 @@ private static Orientation DefaultDeviceOrientation
...
 		}
 
+		partial void StartListeningOrientationChanged()
+		{
+			var mainLooper = Looper.MainLooper;
+			var context = ContextHelper.Current;
+
+			// Thread pool is used to avoid startup
+			// cost of threads creation.
+			ThreadPool.QueueUserWorkItem(
```
MartinZikmund, line 128: What if the user starts and immediately stops? Then the queued work item would execute only after - which would mean the stop would have no effect and the sensor would start running anyway

src/Uno.UWP/Devices/Sensors/SimpleOrientationSensor.Android.cs
```diff
@@ -107,6 +118,81 @@ private static Orientation DefaultDeviceOrientation
...
+						// If the the device has a gyroscope we will use the SensorType.Gravity, if not we will use single angle orientation calculations instead
+						if (gravitySensor != null && useGravitySensor)
+						{
+							_sensorEventListener = new OrientationListener(this);
+							_sensorManager.RegisterListener(_sensorEventListener, _sensorManager.GetDefaultSensor(_gravitySensorType), SensorDelay.Normal);
+						}
+						else
+						{
+							_orientationListener = new SimpleOrientationEventListener(orientation => OnOrientationChanged(orientation));
+							_contentObserver = new SettingsContentObserver(new Handler(mainLooper!), () => OnIsAccelerometerRotationEnabledChanged(IsAccelerometerRotationEnabled));
```
MartinZikmund, line 150: Why is the main looper needed here?

ajpinedam: I just validated, and this logic was already there, this PR only moves it to a place where we can cancel the object initialization in case the sensor cannot start measuring allowing us to return null instead.

But looking at the code (and to answer your question): The  `ContentObserver` object init requires a Handler. It seems they wanted to use the MainThread looper to listen for these changes.

src/Uno.UWP/Devices/Sensors/SimpleOrientationSensor.Android.cs
ajpinedam: I see. Let me run a few manual tests and validate this. Will keep you posted

ajpinedam: Validated: The StartListening method is not public and is only available from the event registration. Having said so, the chances of a user starting and stopping listening immediately are very minimal (close to zero)

cc: @MartinZikmund

MartinZikmund:
@ajpinedam Can you please rebase and adjust the issues?

MartinZikmund (approved):

PR #13538, opened 2023-09-06
chore: Simplify PopupPanel

Youssef1313:
This code used to exist to fix a leak on iOS only. Popups have been refactored since then and I don't think there is a leak concern now.

## PR Type

### <samp>🤖 Generated by Copilot at 7ad2a20</samp>

Refactor `PopupPanel` to use a regular reference to its `Popup` and delegate child lifetime management to `PopupRoot`. This improves the memory efficiency and reliability of popup controls.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/Popup/PopupPanel.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Popup/PopupPanel.cs
@@ -24,7 +24,20 @@ namespace Windows.UI.Xaml.Controls.Primitives;
 
 internal partial class PopupPanel : Panel
 {
+#if UNO_HAS_UIELEMENT_IMPLICIT_PINNING
+	private WeakReference _popup;
+
+	public Popup Popup
+	{
+		get => _popup?.Target as Popup;
+		set
+		{
+			_popup = new WeakReference(value);
+		}
+	}
+#else
 	public Popup Popup { get; }
+#endif
 
 	public PopupPanel(Popup popup)
 	{
--- a/src/Uno.UI/UI/Xaml/Controls/Popup/PopupPanel.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Popup/PopupPanel.cs
@@ -32,7 +32,8 @@ internal partial class PopupPanel : Panel
 		get => _popup?.Target as Popup;
 		set
 		{
-			_popup = new WeakReference(value);
+			WeakReferencePool.ReturnWeakReference(this, _popup);
+			_popup = WeakReferencePool.RentWeakReference(this, value);
 		}
 	}
 #else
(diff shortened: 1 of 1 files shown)


ajpinedam (approved):
Is there a way to validate this is no longer an iOS issue?

MartinZikmund (approved):

src/Uno.UI/UI/Xaml/Controls/Popup/PopupPanel.cs
```diff
@@ -24,17 +24,7 @@ namespace Windows.UI.Xaml.Controls.Primitives;
...
-	public Popup Popup
-	{
-		get => _popup?.Target as Popup;
-		set
-		{
-			WeakReferencePool.ReturnWeakReference(this, _popup);
-			_popup = WeakReferencePool.RentWeakReference(this, value);
-		}
-	}
+	public Popup Popup { get; }
```
jeromelaban, line 27: @Youssef1313 This is likely going to cause memory leaks on iOS/Android. Why do you think this is needed? If this is for a performance issue, we can use the already existing compiler constant `UNO_HAS_UIELEMENT_IMPLICIT_PINNING` to handle this difference.

Youssef1313: @jeromelaban I don't see how this could leak. The Popup and PopupPanel seem to have the same life-time. So instead of exhausting the pool (because the popup seems to never be returned), we just store a strong reference.

The code has changed much since the use of this weak reference was added, and I suspect that this is no longer needed.

Youssef1313: @jeromelaban I moved to using `UNO_HAS_UIELEMENT_IMPLICIT_PINNING`.

I also moved to a simple `WeakReference` since renting from the pool without returning the instance back doesn't feel a good idea.

jeromelaban: It actually is a preferable to use the renting part, since DependencyObjects implement `IWeakReferenceProvider`, which automatically returns a self-managed weak reference from the popup:

See:

https://github.com/unoplatform/uno/blob/fd3e16196fdad21282e9a82ffdde44c7d9aaada0/src/Uno.UI/DataBinding/WeakReferencePool.cs#L60-L65

We could remove `WeakReferencePool.ReturnWeakReference(this, _popup);` call, which does not return anything in that context, though.

jeromelaban (approved):

jeromelaban:
@Youssef1313 

```
src\Uno.UI\UI\Xaml\Controls\Popup\PopupPanel.cs(35,48): 
Error CS1503: Argument 2: cannot convert from 'System.WeakReference' to 'Uno.UI.DataBinding.ManagedWeakReference'
```

PR #13664, opened 2023-09-16
fix(android): Invalidate measure when removing a managed view

### LINKED ISSUE #13539: [Android] ItemsControl not resizing properly

In this issue 
https://github.com/unoplatform/uno.toolkit.ui/issues/798
we found that in some case the ItemsControl was not resizing correctly.
When we switched to ItemsRepeater the situation normalized and the resize worked.

So the ShadowContainer, in that case, was correctly stretching when the content was changed to be larger but if the content was then shrunk, the ShadowContainer does not resize:

See Shadow Builder control inside of the Toolkit sample app:

shadowcontainerstretch

Or Attached to this Issue is a Project showing the case.

https://github.com/unoplatform/uno/assets/116665025/e1d2cfcb-99f7-40b7-8a55-fb1b39359311

Project

[UnoAppBlankSample.zip](https://github.com/unoplatform/uno/files/12541339/UnoAppBlankSample.zip)

### Why is this needed

we need to find out why the ItemsControl was not resized

### For which platform

Android

_No response_

Youssef1313:
GitHub Issue (If applicable): closes #13539

## PR Type

### <samp>🤖 Generated by Copilot at a2e5829</samp>

This pull request fixes a layout bug in `StackPanel` on Android by enhancing the `BindableView` and `UIElement` classes to handle child view changes more consistently. It also adds a unit test to verify the expected behavior of the `StackPanel` when adding or removing children.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_StackPanel.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_StackPanel.cs
@@ -58,10 +58,14 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 				}
 			};
 
+			int expectedMeasureAndArrangeCount = 0;
+
 			TestServices.WindowHelper.WindowContent = SUT;
 			await TestServices.WindowHelper.WaitForLoaded(SUT);
-			Assert.AreEqual(1, SUT.MeasureCount);
-			Assert.AreEqual(1, SUT.ArrangeCount);
+			await TestServices.WindowHelper.WaitForIdle();
+			expectedMeasureAndArrangeCount++;
+			Assert.AreEqual(expectedMeasureAndArrangeCount, SUT.MeasureCount);
+			Assert.AreEqual(expectedMeasureAndArrangeCount, SUT.ArrangeCount);
 			Assert.AreEqual(50, SUT.LastMeasureOverrideReturn.Height);
 			Assert.AreEqual(50, SUT.LastArrangeOverrideReturn.Height);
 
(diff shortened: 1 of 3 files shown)


src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_StackPanel.cs
```diff
@@ -19,8 +19,109 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
...
+			await TestServices.WindowHelper.WaitForRelayouted(SUT);
+			expectedMeasureAndArrangeCount++;
+			Assert.AreEqual(expectedMeasureAndArrangeCount, SUT.MeasureCount);
+			Assert.AreEqual(expectedMeasureAndArrangeCount, SUT.ArrangeCount);
+			Assert.AreEqual(50, SUT.LastMeasureOverrideReturn.Height);
+			Assert.AreEqual(50, SUT.LastArrangeOverrideReturn.Height);
+
+			SUT.Children.Remove(SUT.Children.Single());
+#if !__CROSSRUNTIME__
+			await TestServices.WindowHelper.WaitForRelayouted(SUT); // arrange is skipped (incorrectly) when finalRect is default, which is the case after removing the last child from StackPanel.
```
Youssef1313, line 94: This is caused by https://github.com/unoplatform/uno/blob/e774ca82a87f8361beab37b69765469b94ce6033/src/Uno.UI/UI/Xaml/UIElement.Layout.crossruntime.cs#L307-L319

This code is wrong and should be removed. If it causes any issues for Canvas, it should be re-investigated to understand the real root cause, or alternatively, special case this code path for Canvas specifically, but that sounds like a hack

Youssef1313: The `finalRect == default` existed before the Canvas-related fix. The Canvas-related fix (which is the check for `ICustomClippingElement`) was done after it because `finalRect == default` isn't correct in the first place.

The finalRect == `default` check was added as a performance optimization, but I don't see much value in it. I don't we will hit it that much, so it can safely be removed. Probably in a follow-up PR.

MartinZikmund (approved):

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_StackPanel.cs
MartinZikmund: Definitely worth doing, can you create an issue, so it is not forgotten?

jeromelaban (approved):

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_StackPanel.cs
Youssef1313: https://github.com/unoplatform/uno/issues/13672

Issue #10263, opened 2022-10-26
`CornerRadius` rendering does not take `BorderThickness` into account


MartinZikmund:
### Current behavior

On most targets `CornerRadius` does not adjust based on `BorderThickness`

### Expected behavior

Outer corner radius == `CornerRadius` + `BorderThickness` / 2
Inner corner radius == `CornerRadius` - `BorderThickness` / 2

### Affected platforms

WebAssembly, Android, iOS, macOS (AppKit), Mac Catalyst, Skia (WPF), Skia (GTK on Linux/macOS/Windows), Skia (Linux Framebuffer), Skia (Tizen)

Youssef1313:
Is this a duplicate of https://github.com/unoplatform/uno/issues/6891?

MartinZikmund:
From #6891 

### Current behavior

Edge case to fix on all platforms where Inner Radius doesn't match UWP when BorderThickness >= CornerRadius

![image](https://user-images.githubusercontent.com/16295702/131129906-f5d59954-3775-4f51-ab4f-62b489c63130.png)
![image](https://user-images.githubusercontent.com/16295702/131129985-2646db67-8a7b-401f-a4d9-d69e81abde2f.png)
![image](https://user-images.githubusercontent.com/16295702/131130024-7e623caa-52a3-454e-ae93-3a1f847a8981.png)
**_Android_**
![image](https://user-images.githubusercontent.com/16295702/131131644-99eb82f9-8cfa-41b0-9cb2-3131deb2cca2.png)


### Expected behavior

Inner Radius calculations should match UWP behaviors when BorderThickness >= CornerRadius

### How to reproduce it (as minimally and precisely as possible)

```
<StackPanel>
		<toolkit:ElevatedView Height="100"
							  Width="200"
							  Background="Yellow"
							  CornerRadius="15"
							  BorderBrush="#66121212"
							  BorderThickness="15"
							  Margin="30">
			<Grid Background="DeepPink"
				  CornerRadius="25">
				<TextBlock Text="Test with ElevatedView"
						   VerticalAlignment="Center"
						   HorizontalAlignment="Center" />
			</Grid>
		</toolkit:ElevatedView>

		<Grid Height="100"
			  Width="200"
			  Background="Orange"
			  CornerRadius="15"
			  BorderBrush="#66121212"
			  BorderThickness="15"
			  Margin="30">
			<Grid Background="Red"
				  CornerRadius="25">
				<TextBlock Text="Test with a Grid"
[...]

PR #13682, opened 2023-09-18
fix: Fix corner radius handling on Wasm

Youssef1313:
GitHub Issue (If applicable): Wasm part of #10263

## PR Type

### <samp>🤖 Generated by Copilot at 17282ef</samp>

This pull request improves the border rendering of various UI elements on the WASM platform, by adding support for rounded corners based on the `CornerRadius` property and updating the border layer with the correct properties when they change. It also fixes a typo in a sample and simplifies the API for setting the border layer in the `FrameworkElement` class.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/Border/BorderLayerRenderer.wasm.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Border/BorderLayerRenderer.wasm.cs
@@ -68,7 +68,7 @@ namespace Windows.UI.Xaml.Shapes
 							("border-width", borderWidth));
 						break;
 					case GradientBrush gradientBrush:
-						var border = gradientBrush.ToCssString(element.RenderSize); // TODO: Reevaluate when size is changing
+						var border = gradientBrush.ToCssString(element.RenderSize);
 						element.SetStyle(
 							("border-style", "solid"),
 							("border-color", ""),
@@ -77,7 +77,7 @@ namespace Windows.UI.Xaml.Shapes
 							("border-image-slice", "1"));
 						break;
 					case RadialGradientBrush radialGradientBrush:
-						var radialBorder = radialGradientBrush.ToCssString(element.RenderSize); // TODO: Reevaluate when size is changing
+						var radialBorder = radialGradientBrush.ToCssString(element.RenderSize);
 						element.SetStyle(
 							("border-style", "solid"),
 							("border-color", ""),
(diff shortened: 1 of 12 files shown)


src/Uno.UI/UI/Xaml/Controls/Border/BorderLayerRenderer.wasm.cs
```diff
@@ -119,6 +98,19 @@ public static void SetBorder(UIElement element, Thickness thickness, Brush brush
...
 			}
+
+			if (cornerRadius == CornerRadius.None)
+			{
+				element.ResetStyle("border-radius", "overflow");
+			}
+			else
+			{
+				var outer = cornerRadius.GetRadii(element.RenderSize, thickness).Outer;
+				var borderRadius = $"{outer.TopLeft.X}px {outer.TopRight.X}px {outer.BottomRight.X}px {outer.BottomLeft.X}px / {outer.TopLeft.Y}px {outer.TopRight.Y}px {outer.BottomRight.Y}px {outer.BottomLeft.Y}px";
```
MartinZikmund, line 109: Does this get recalculated when render size changes (render scale transform for example)? I remember I found the border layer renderer on WASM wasn't refreshed in such case, which meant the corner radius could get incorrect, but maybe it has changed since

Youssef1313: I think it might not work for this specific case. Do we want that in this PR or in a follow up?

MartinZikmund: I guess it is mostly an edge case, but needs to be addressed - but I am fine with separate PR

MartinZikmund (approved):

ajpinedam (approved):

src/Uno.UI/UI/Xaml/Controls/Border/BorderLayerRenderer.wasm.cs
Youssef1313: There is a test that is failing because of this actually. That is because the CornerRadius update happens before arrange, so RenderSize is zero at that time. So after arrange, the CornerRadius doesn't show up at all.

I'll address in this PR.

jeromelaban, line 109: Let's use an explicit javascript invocation instead of formatting on the C# side (which may also have failed because of non-invariant formatting), let's do the same with the reset above.

jeromelaban (approved):

src/Uno.UI/UI/Xaml/Controls/Border/BorderLayerRenderer.wasm.cs
Youssef1313: I made the change for corner radius specifically.
Not sure how much else we want to move?

src/Uno.UI/UI/Xaml/Controls/Border/BorderLayerRenderer.wasm.cs
```diff
@@ -119,6 +98,17 @@ public static void SetBorder(UIElement element, Thickness thickness, Brush brush
...
+
+			if (cornerRadius == CornerRadius.None)
+			{
+				element.ResetStyle("border-radius", "overflow");
+			}
+			else
+			{
+				var outer = cornerRadius.GetRadii(element.RenderSize, thickness).Outer;
+				WindowManagerInterop.SetCornerRadius(element.HtmlId, outer.TopLeft.X, outer.TopLeft.Y, outer.TopRight.X, outer.TopRight.Y, outer.BottomRight.X, outer.BottomRight.Y, outer.BottomLeft.X, outer.BottomLeft.Y);
+				element.SetStyle("overflow", "hidden"); // overflow: hidden is required here because the clipping can't do its job when it's non-rectangular.
```
MartinZikmund, line 110: Can this be done inside the `SetCornerRadius` as well? (if it makes sense)

jeromelaban: Indeed, that would be interesting.

Youssef1313: Done!

PR #13693, opened 2023-09-19
docs: Remove missing doc

nickrandolph:


MartinZikmund:
Why not remove it completely 🤔

nickrandolph:
> Why not remove it completely 🤔

🤷‍♂️ - wishful thinking that we'd have a document to go there as an overview for migration.

MartinZikmund (approved):

PR #13792, opened 2023-09-25
ci: Update scaled sets

jeromelaban:


build/ci/templates/nuget-cache.yml
```diff
@@ -6,10 +6,10 @@ steps:
...
     displayName: Create nuget packages cache folder
-
+  
   - script: |
-      compact /c "/s:${{ parameters.nugetPackages }}"
-
+      compact /c "${{ parameters.nugetPackages }}"
+  
     condition: and(succeeded(), eq( variables['Agent.OS'], 'Windows_NT' ))
     displayName: Compress nuget package cache folder
```
Youssef1313, line 14: How is the cache supposed to work? The `Cache` task below has a condition on `enable_package_cache` which is never set to true.

build/ci/templates/nuget-cache.yml
Youssef1313, line 1: Also the cache task key is dependent on OS, why is that needed? I thought NuGet cache could be reused across different OSes? It's just the nupkg being unzipped if I understand correctly?

build/ci/templates/nuget-cache.yml
```diff
@@ -6,10 +6,10 @@ steps:
...
```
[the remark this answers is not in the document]
jeromelaban: It's not particularly important at this time, because the nuget will handle it properly anyways.

build/ci/templates/nuget-cache.yml
jeromelaban: it's not exactly reusable because of casing, though I don't remember the specifics.

Issue #13844, opened 2023-09-27
TestTranslatedClick fails on faster build agents


jeromelaban:

MartinZikmund:
Relevant

PR #13845, opened 2023-09-27
Dev/jela/build adjust

jeromelaban:
## PR Type

- Bugfix

## What is the new behavior?

- Disable flaky wasm test (https://github.com/unoplatform/uno/issues/13844)
- Adjust generation for new pointer type

## Copilot Summary

copilot:summary

## PR Checklist

## Other information

### DIFF
--- a/src/SourceGenerators/Uno.UI.SourceGenerators.Internal/TSBindings/TSBindingsGenerator.cs
+++ b/src/SourceGenerators/Uno.UI.SourceGenerators.Internal/TSBindings/TSBindingsGenerator.cs
@@ -19,6 +19,7 @@ namespace Uno.UI.SourceGenerators.TSBindings
 		private string[] _sourceAssemblies;
 
 		private static INamedTypeSymbol _intPtrSymbol;
+		private static INamedTypeSymbol _uintPtrSymbol;
 		private static INamedTypeSymbol _structLayoutSymbol;
 		private static INamedTypeSymbol _interopMessageSymbol;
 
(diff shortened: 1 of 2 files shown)


src/SourceGenerators/Uno.UI.SourceGenerators.Internal/TSBindings/TSBindingsGenerator.cs
```diff
@@ -18,9 +18,10 @@ class TSBindingsGenerator : ISourceGenerator
 		private string _bindingsPaths;
 		private string[] _sourceAssemblies;
 
-		private static INamedTypeSymbol _intPtrSymbol;
-		private static INamedTypeSymbol _structLayoutSymbol;
-		private static INamedTypeSymbol _interopMessageSymbol;
+		private INamedTypeSymbol _intPtrSymbol;
+		private INamedTypeSymbol _uintPtrSymbol;
+		private INamedTypeSymbol _structLayoutSymbol;
+		private INamedTypeSymbol _interopMessageSymbol;
```
Youssef1313, line 24: The compiler is free to re-use generator instances:

https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.iincrementalgenerator?view=roslyn-dotnet-4.6.0

> The lifetime of a generator is controlled by the compiler. State should not be stored directly on the generator, as there is no guarantee that the same instance will be used on a subsequent generation pass.

I think this might still be problematic. It's probably best to pass these around.

For IntPtr, there is actually a SpecialType for it, so we might not need that symbol at all.

src/SourceGenerators/Uno.UI.SourceGenerators.Internal/TSBindings/TSBindingsGenerator.cs
```diff
@@ -375,6 +377,7 @@ private int GetNativeFieldSize(IFieldSymbol field)
 				field.Type.SpecialType is SpecialType.System_Int32 ||
 				field.Type.SpecialType is SpecialType.System_UInt32 ||
 				SymbolEqualityComparer.Default.Equals(field.Type, _intPtrSymbol) ||
+				SymbolEqualityComparer.Default.Equals(field.Type, _uintPtrSymbol) ||
```
Youssef1313, line 380: I think this could be reverted?

PR #13850, opened 2023-09-27
fix(scrollviewer): keyboard scrolling with arrow keys

### LINKED ISSUE #12964: ListView should scroll the outer ScrollViewer when keyboard navigation reaches the ListView start/end
### Current behavior

When moving up and down a ListView inside a ScrollViewer with a keyboard, the movement stop at the boundaries (start/end).

### Expected behavior

The key event should propagate to the outer ScrollViewer to let it scroll down when down arrow is pressed while the ListView is already at the last item. The same should happen in the opposite direction if up arrow is pressed. In the following video.

https://github.com/unoplatform/uno/assets/66218781/db5aef1f-f71f-4223-97d4-a7baba6b1150

<ScrollViewer>
		<StackPanel>
			<ListView ItemsSource="ABCDE"/>
			<ListView ItemsSource="ABCDE"/>
			<ListView ItemsSource="ABCDE"/>
			<ListView ItemsSource="ABCDE"/>
			<ListView ItemsSource="ABCDE"/>
			<ListView ItemsSource="ABCDE"/>
			<ListView ItemsSource="ABCDE"/>
			<ListView ItemsSource="ABCDE"/>
			<ListView ItemsSource="ABCDE"/>
			<ListView ItemsSource="ABCDE"/>
			<ListView ItemsSource="ABCDE"/>
			<ListView ItemsSource="ABCDE"/>
		</StackPanel>
	</ScrollViewer>

### Workaround

_No response_

### Works on UWP/WinUI

Yes

### Environment

_No response_

### NuGet package version(s)

_No response_

### Affected platforms

_No response_

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

### LINKED ISSUE #13849: `ScrollViewer` can't be scrolled with keyboard
### Current behavior

Nothing happens.

### Expected behavior

On WinUI, the ScrollViewer can be scrolled with arrow keys.

_No response_

### Workaround

_No response_

### Works on UWP/WinUI

None

### Environment

_No response_

### NuGet package version(s)

_No response_

### Affected platforms

_No response_

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

ramezgerges:
GitHub Issue (If applicable): closes #13849, closes #12964

## PR Type

- Bugfix

### <samp>🤖 Generated by Copilot at bc78713</samp>

This pull request adds keyboard scrolling support for the `ScrollViewer` control on all platforms. It modifies the `ContentControl` and `ScrollContentPresenter` classes to handle input events, and adds a new test method to verify the functionality. It affects the files `ContentControl.cs`, `ScrollContentPresenter.cs`, `ScrollContentPresenter.Managed.cs`, and `Given_ScrollViewer.cs`.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/ContentControl/ContentControl.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/ContentControl/ContentControl.cs
@@ -327,7 +327,7 @@ namespace Windows.UI.Xaml.Controls
 
 			if (args.GetCurrentPoint(this).Properties.IsLeftButtonPressed)
 			{
-				args.Handled = Focus(FocusState.Pointer);;
+				args.Handled = Focus(FocusState.Pointer);
 			}
 		}
 
--- a/src/Uno.UI/UI/Xaml/Controls/ContentControl/ContentControl.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/ContentControl/ContentControl.cs
@@ -11,6 +11,7 @@ using Windows.UI.Xaml.Media.Animation;
 using System.Collections;
 using System.Linq;
 using Windows.UI.Xaml.Data;
+using Windows.UI.Xaml.Input;
 using Windows.UI.Xaml.Markup;
 using Uno;
 
@@ -61,7 +62,6 @@ namespace Windows.UI.Xaml.Controls
 		public ContentControl()
 		{
 			DefaultStyleKey = typeof(ContentControl);
-
 			InitializePartial();
 		}
 
@@ -314,6 +314,20 @@ namespace Windows.UI.Xaml.Controls
 			if (IsContentPresenterBypassEnabled)
 			{
 				SetUpdateTemplate();
+			}
+		}
+
+		protected override void OnPointerPressed(PointerRoutedEventArgs args)
+		{
+			if (args.Handled)
+			{
+				base.OnPointerPressed(args);
+				return;
+			}
+
+			if (args.GetCurrentPoint(this).Properties.IsLeftButtonPressed)
+			{
+				args.Handled = Focus(FocusState.Pointer);;
 			}
 		}
 
--- a/src/Uno.UI/UI/Xaml/Controls/ContentControl/ContentControl.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/ContentControl/ContentControl.cs
@@ -11,7 +11,6 @@ using Windows.UI.Xaml.Media.Animation;
 using System.Collections;
 using System.Linq;
 using Windows.UI.Xaml.Data;
-using Windows.UI.Xaml.Input;
 using Windows.UI.Xaml.Markup;
 using Uno;
 
@@ -62,6 +61,7 @@ namespace Windows.UI.Xaml.Controls
 		public ContentControl()
 		{
 			DefaultStyleKey = typeof(ContentControl);
+
 			InitializePartial();
 		}
 
@@ -314,20 +314,6 @@ namespace Windows.UI.Xaml.Controls
 			if (IsContentPresenterBypassEnabled)
 			{
 				SetUpdateTemplate();
-			}
-		}
-
-		protected override void OnPointerPressed(PointerRoutedEventArgs args)
-		{
-			if (args.Handled)
-			{
-				base.OnPointerPressed(args);
-				return;
-			}
-
-			if (args.GetCurrentPoint(this).Properties.IsLeftButtonPressed)
-			{
-				args.Handled = Focus(FocusState.Pointer);
 			}
 		}
 
--- a/src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.cs
@@ -23,6 +23,7 @@ using Uno.Extensions;
 using Uno.Foundation.Logging;
 using Uno.UI.Extensions;
 using Windows.Foundation.Metadata;
+using Uno.UI.Xaml.Core;
 
 #if __ANDROID__
 using View = Android.Views.View;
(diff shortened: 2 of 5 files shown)


src/Uno.UI/UI/Xaml/Controls/ContentControl/ContentControl.cs
```diff
@@ -317,6 +317,20 @@ private protected override void OnLoaded()
...
+			if (args.Handled)
+			{
+				base.OnPointerPressed(args);
+				return;
+			}
+
+			if (args.GetCurrentPoint(this).Properties.IsLeftButtonPressed)
+			{
+				args.Handled = Focus(FocusState.Pointer);
+			}
```
jeromelaban, line 331: Why do we need this part? I would not think that a ContentControl can be focused? /cc @MartinZikmund

src/Uno.UI/UI/Xaml/Controls/ContentControl/ContentControl.cs
```diff
@@ -317,6 +317,20 @@ private protected override void OnLoaded()
 			}
 		}
 
+		protected override void OnPointerPressed(PointerRoutedEventArgs args)
+		{
+			if (args.Handled)
+			{
+				base.OnPointerPressed(args);
+				return;
+			}
```
dr1rrb, line 326: If the `args` is already flagged as `Handled` you should not get it here. It's a bug you noticed ? If yes please open a bug and link it here, otherwise please remove.

src/Uno.UI/UI/Xaml/Controls/ContentControl/ContentControl.cs
dr1rrb, line 331: Did you confirm that on Windows `ContentControl` flags the event has handled. This might have significant side effects, I would be surprised we did detect that behavior earlier.

Also @MartinZikmund might have an opinion about the focus management :)

src/Uno.UI/UI/Xaml/Controls/ContentControl/ContentControl.cs
ramezgerges: Not sure why I decided to add that, but yes you're right, not needed.

src/Uno.UI/UI/Xaml/Controls/ContentControl/ContentControl.cs
ramezgerges: Yes, a lot of ContentControls are focusable. I tested with a bunch of ContentControls to make sure. However, it turns out that that's not general to all ContentControls. ScrollViewer for instance is not focusable, but it tries to grab focus (so that whatever is inside ends up with the focus.) Further more, focus handling should be done inside each element's pointer callbacks, not at the ContentControl level. For example, ButtonBase handles it in OnPointerPressed, but SelectorItems and ScrollViewers do it in OnPointerReleased.
 
 Interestingly, when I was testing ScrollViewer focus in WinUI, a ScrollViewer was indeed being focused, But it turns out that's not the ScrollViewer I was testing, but an "invisible" root ScrollViewer that's used by WinUI windows (thanks to @MartinZikmund for the insight.)

ramezgerges: Upon further testing, it's at the ScrollViewer level, not ContentControl level, that the event gets handled.

src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.cs
```diff
@@ -1587,6 +1588,95 @@ private void ResetScrollIndicator(bool forced = false)
...
+					ScrollToVerticalOffset(newOffset);
+					args.Handled = !NumericExtensions.AreClose(oldVerticalOffset, Presenter.TargetVerticalOffset);
+				}
+
+				args.Handled |= key is VirtualKey.PageUp or VirtualKey.Down;
+			}
+
+			int GetDelta(double l)
+			{
+				var length = (int)Math.Max(0, Math.Round(l) - 16);
```
MartinZikmund, line 1657: Can you comment this code to explain what it does?

MartinZikmund (approved):

src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.cs
ramezgerges: This gets the delta that should be applied when arrow keys are pressed. I honestly have no idea what's the original logic, I just reverse-engineered the numbers until they matched precisely. I think the original code just has some weird rounding somewhere that makes the numbers this weird to calculate.

src/Uno.UI/UI/Xaml/Controls/ContentControl/ContentControl.cs
jeromelaban: Got it. Now that we have general access to winui, are you able to find a more definitive answer about it?

ramezgerges: ScrollViewer is the one that explicitly asks for focus (like the PR currently does).

src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.cs
jeromelaban: @ramezgerges could you add the comment in the code directly?

jeromelaban:
@ramezgerges now that we can test the sample app more easily, I tried looking at this sample:

image

It looks like the navigation works with the keys, which is good, but it seems that the scrollbars don't follow the actual position when done through keys. 

If it's an issue that was there previously, let's open an issue to track, if not, let's try to fix as well.

ramezgerges:
I just did some testing and in fact, the ScrollViewer is NOT supposed to scroll. If I remember correctly, this is due to some focus logic that I fixed in another PR a while ago. This is the behaviour on skia (good), but on WASM, it seems that the native scrolling still takes place (which is why the scrollbars aren't moving in sync). It will probably need a workaround like the one I did in #14134.

PR #13895, opened 2023-10-04
ci: Misc CI performance improvements

Youssef1313:

## PR Type

### <samp>🤖 Generated by Copilot at 5b022c5</samp>

This pull request optimizes the checkout steps of various jobs in the CI pipeline by using parameters such as `fetch-depth`, `refspec`, and `skipCheckout`. This reduces the amount of data transferred and improves the performance and reliability of the CI process. It also adds support for both WinUI and UWP XAML flavors for the Wasm UI tests by changing the artifact names and the download task.

## PR Checklist

## Other information

### DIFF
--- a/build/ci/.azure-devops-commitsar.yml
+++ b/build/ci/.azure-devops-commitsar.yml
@@ -11,6 +11,7 @@ jobs:
   steps:
   - checkout: self
     clean: true
+    fetchTags: false
 
   - bash: |
       curl -L -O https://github.com/aevea/commitsar/releases/download/v0.20.2/commitsar_0.20.2_linux_amd64.tar.gz
--- a/build/ci/.azure-devops-package-netcoremobile.yml
+++ b/build/ci/.azure-devops-package-netcoremobile.yml
@@ -52,6 +52,7 @@ jobs:
   steps:
   - checkout: self
     clean: true
+    fetchTags: false
 
   - template: templates/nuget-cache.yml
     parameters:
--- a/build/ci/.azure-devops-package-netcoremobile.yml
+++ b/build/ci/.azure-devops-package-netcoremobile.yml
@@ -52,7 +52,6 @@ jobs:
   steps:
   - checkout: self
     clean: true
-    fetchTags: false
 
   - template: templates/nuget-cache.yml
     parameters:
--- a/build/ci/.azure-devops-pipeline-validations.yml
+++ b/build/ci/.azure-devops-pipeline-validations.yml
@@ -22,7 +22,6 @@ jobs:
   steps:
   - checkout: self
     clean: true
-    fetchTags: false
 
   - template: templates/gitversion.yml
     parameters:
--- a/build/ci/.azure-devops-pipeline-validations.yml
+++ b/build/ci/.azure-devops-pipeline-validations.yml
@@ -22,6 +22,7 @@ jobs:
   steps:
   - checkout: self
     clean: true
+    fetchTags: false
 
   - template: templates/gitversion.yml
     parameters:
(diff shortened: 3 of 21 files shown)


build/ci/.azure-devops-commitsar.yml
```diff
@@ -11,6 +11,8 @@ jobs:
   steps:
   - checkout: self
     clean: true
+    fetchTags: false
+    fetchDepth: 1
```
Youssef1313, line 15: This one needs careful review. I'm not sure if commitsar needs the history?

build/ci/.azure-devops-wasm-uitests.yml
```diff
@@ -83,7 +91,7 @@ jobs:
 
   - task: DownloadBuildArtifacts@0
     inputs:
-        artifactName: wasm-uitest-binaries
+        artifactName: wasm-uitest-binaries-UWP
```
Youssef1313, line 94: Note: Few lines below we run `dotnet-serve` on the `site-net7.0-UWP` directory. So we previously were downloading both UWP and WinUI artifacts and then using only UWP artifacts. The separation allows for granular download, which should be faster.

build/ci/.azure-devops-uap.yml
```diff
@@ -20,6 +20,7 @@ jobs:
 
   steps:
   - checkout: self
+    clean: true
```
Youssef1313, line 23: @jeromelaban I see this job (and also many other jobs) are using `gitversion.yml`, probably unnecessarily. Do you know for which jobs we can remove the use of `gitversion.yml`? Is it any non-packaging job? The removal of `gitversion.yml` will allow us to do shallow fetch which will be faster.

build/ci/.azure-devops-package-netcoremobile.yml
```diff
@@ -52,6 +52,7 @@ jobs:
   steps:
   - checkout: self
     clean: true
+    fetchTags: false
```
jeromelaban, line 55: Let's keep the tags for all package builds, even if gitversion uses a cache at this time.

build/ci/.azure-devops-pipeline-validations.yml
```diff
@@ -22,6 +22,7 @@ jobs:
   steps:
   - checkout: self
     clean: true
+    fetchTags: false
```
jeromelaban, line 25: Tags are required by gitversion

build/ci/.azure-devops-uap.yml
jeromelaban: Non packaging jobs don't need gitversion (it's not supposed to be there, iirc). For the others, full clone and tags are required.

build/ci/.azure-devops-winui-convert.yml
```diff
@@ -24,12 +24,12 @@ jobs:
   steps:
   - checkout: self
     clean: true
+    fetchTags: false
+    fetchDepth: 1
```
jeromelaban, line 28: Versioning of sources may happen here as well, let's keep the full clone+tags

Youssef1313: @jeromelaban The jobs calling `.azure-devops-winui-convert.yml` should be doing the versioning if required. I think this job only affects source files. All binaries produced during this job are discarded in `Cleanup sources before publish` step.

Youssef1313:
@jeromelaban I'm not able to see why there are failures similar to this in CI:

```
MSBUILD : error MSB1009: Project file does not exist.
Switch: C:\a\1\s\src\Uno.UI-UnitTests-only.slnf
```

Are you able to see if there is something obvious I'm missing?

build/ci/.azure-devops-winui-convert.yml
jeromelaban: In this context, the versions we're looking for are the ones provided by informational version and related attributes. It may not be needed anymore to modify sources as it's msbuild that includes them. We just need to make sure that the output packages binaries continue to have these versions set.

PR #13898, opened 2023-10-04
ci: Avoid restoring the whole solution unnecessarily when running sync generator

Youssef1313:

## PR Type

### <samp>🤖 Generated by Copilot at 3d05cb8</samp>

Simplify nuget restore task for sync generator. Use solution filters that only include `Uno.UI` and its dependencies to speed up and stabilize the API synchronization with UWP.

## PR Checklist

## Other information

### DIFF
--- a/build/Uno.UI.Build.csproj
+++ b/build/Uno.UI.Build.csproj
@@ -234,8 +234,9 @@
 	<Target Name="BuildSyncGenerator">
 		<!-- Restore the nuget packages for packages-* solution filters -->
 		<!-- These filters contain Uno.UI project and its dependencies, which are what needs to be restored -->
+		<!-- In addition, the reference project needs to be restored as well. -->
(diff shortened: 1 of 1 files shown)


jeromelaban:
Restoring the whole solution is needed for Roslyn to function properly and generate the synced APIs.

jeromelaban:
@ramezgerges The restore list is already trimmed down, we could probably trim it down further, but I don't think it will yield important results.

Youssef1313:
> The restore list is already trimmed down, we could probably trim it down further, but I don't think it will yield important results.

Restore is taking about 1 minute 15 seconds, and is expected to shrink to ~40 seconds. The save here is half a minute and indeed probably might not be worth it.

Let's see if it gets green or near green with the last change. If not, I'll close as it won't be worth the time.

Youssef1313:
The conversion job is green. So hopefully everything else should get green as well.

jeromelaban:
It seems to reduce the restore time, let's merge :)

PR #13910, opened 2023-10-06
Fix `DoubleTapped` not fired properly when a child subscribes to `Tapped`

### LINKED ISSUE #13856: [NPE] When double click any file in the content section, window to save the file is not opening.
### Current behavior

https://github.com/unoplatform/uno/assets/74563602/408c8ef5-5b38-451d-8f59-8747fb6951a9

### Expected behavior

**WASM Prod**

https://github.com/unoplatform/uno/assets/74563602/bf63989a-849e-446b-b912-d97e0b46a0ec

1. Open NPE https://purple-pond-0688d230f.azurestaticapps.net/
2. Click on 'Open button for any .Net file.
3. Go to 'Content' section.
4. Click on 'lib' to open the dropdown.
5. Double click on the 'dll' or 'xml' file
6. Observe that window to save the file is not opening.

### Workaround

_No response_

### Works on UWP/WinUI

None

### Environment

_No response_

### NuGet package version(s)

_No response_

### Affected platforms

WebAssembly

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

Youssef1313:
GitHub Issue (If applicable): closes #13856

## PR Type

### <samp>🤖 Generated by Copilot at d8fd096</samp>

This pull request adds the `PointerId` property to various gesture event classes and passes it to the corresponding events to align with the UWP API and provide more information to the event handlers. It also simplifies and fixes the gesture prevention logic in the `Gesture` and `GestureRecognizer` classes and prevents gesture conflicts in the `UIElement` class.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/UIElement.Pointers.cs
+++ b/src/Uno.UI/UI/Xaml/UIElement.Pointers.cs
@@ -549,18 +549,24 @@ namespace Windows.UI.Xaml
 		partial void PrepareManagedGestureEventBubbling(RoutedEvent routedEvent, ref RoutedEventArgs args, ref BubblingMode bubblingMode)
 		{
 			// When we bubble a gesture event from a child, we make sure to abort any pending gesture/manipulation on the current element
-			if (routedEvent == HoldingEvent)
+			if (IsGestureRecognizerCreated)
 			{
-				if (IsGestureRecognizerCreated)
+				if (routedEvent == TappedEvent)
+				{
+					GestureRecognizer.PreventEvents(((TappedRoutedEventArgs)args).PointerId, GestureSettings.Tap);
+				}
+				else if (routedEvent == DoubleTappedEvent)
+				{
+					GestureRecognizer.PreventEvents(((DoubleTappedRoutedEventArgs)args).PointerId, GestureSettings.DoubleTap);
+				}
+				else if (routedEvent == RightTappedEvent)
+				{
+					GestureRecognizer.PreventEvents(((RightTappedRoutedEventArgs)args).PointerId, GestureSettings.RightTap);
+				}
+				else if (routedEvent == HoldingEvent)
 				{
 					GestureRecognizer.PreventEvents(((HoldingRoutedEventArgs)args).PointerId, GestureSettings.Hold);
 				}
-			}
-			else
-			{
-				// Note: Here we should prevent only the same gesture ... but actually currently supported gestures
-				// are mutually exclusive, so if a child element detected a gesture, it's safe to prevent all of them.
-				CompleteGesture(); // Make sure to set the flag _isGestureCompleted, so won't try to recognize double tap
 			}
 		}
 
--- a/src/Uno.UI/UI/Xaml/UIElement.Pointers.cs
+++ b/src/Uno.UI/UI/Xaml/UIElement.Pointers.cs
@@ -548,6 +548,14 @@ namespace Windows.UI.Xaml
 
 		partial void PrepareManagedGestureEventBubbling(RoutedEvent routedEvent, ref RoutedEventArgs args, ref BubblingMode bubblingMode)
 		{
+			if (routedEvent != HoldingEvent && FeatureConfiguration.UIElement.DisablePointersSpecificEventPrevention)
+			{
+				// If the feature flag is set, call CompleteGesture.
+				// This is known to not be correct, but it's there just in case the prevention logic caused regressions.
+				CompleteGesture();
+				return;
+			}
+
 			// When we bubble a gesture event from a child, we make sure to abort any pending gesture/manipulation on the current element
 			if (IsGestureRecognizerCreated)
 			{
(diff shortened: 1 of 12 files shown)


src/Uno.UI/UI/Xaml/UIElement.Pointers.cs
```diff
@@ -549,19 +549,25 @@ private void ToggleGesture(RoutedEvent routedEvent)
...
+				}
+				else if (routedEvent == DoubleTappedEvent)
+				{
+					GestureRecognizer.PreventEvents(((DoubleTappedRoutedEventArgs)args).PointerId, GestureSettings.DoubleTap);
+				}
+				else if (routedEvent == RightTappedEvent)
+				{
+					GestureRecognizer.PreventEvents(((RightTappedRoutedEventArgs)args).PointerId, GestureSettings.RightTap);
+				}
+				else if (routedEvent == HoldingEvent)
```
dr1rrb, line 566: We should add a feature flag for this, it might cause some regression and it's easy to add one here.

BTW: Great that you are able to disable only using the pointer ID 👍

src/Uno.UWP/UI/Input/GestureRecognizer.Gesture.cs
```diff
@@ -120,37 +120,9 @@ public void ProcessComplete()
...
-				Settings &= ~GestureSettings.RightTap;
-				if ((Settings & GestureSettingsHelper.SupportedGestures) == GestureSettings.None)
-				{
-					IsCompleted = true;
-				}
-			}
-
-			public void PreventHolding()
-			{
-				StopHoldingTimer();
```
dr1rrb, line 152: I think we should keep this

src/Uno.UWP/UI/Input/GestureRecognizer.cs
```diff
@@ -183,25 +183,8 @@ internal GestureSettings PreventEvents(uint pointerId, GestureSettings events)
...
-				{
-					gesture.PreventDoubleTap();
-				}
-
-				if (events.HasFlag(GestureSettings.Hold))
-				{
-					gesture.PreventHolding();
-				}
+				const GestureSettings PreventionMask = GestureSettings.Tap | GestureSettings.RightTap | GestureSettings.DoubleTap | GestureSettings.Hold;
+				gesture.PreventGestures(events & PreventionMask);
```
dr1rrb, line 187:
```suggestion
				gesture.PreventGestures(events & GestureSettingsHelper.SupportedGestures);
```

src/Uno.UWP/UI/Input/RightTappedEventArgs.cs
```diff
@@ -9,16 +9,19 @@ namespace Windows.UI.Input
 {
 	public partial class RightTappedEventArgs
 	{
-		internal RightTappedEventArgs(PointerDeviceType type, Point position)
+		internal RightTappedEventArgs(uint pointerId, PointerDeviceType type, Point position)
```
dr1rrb, line 12: Personal preference: usually I try to keep the properties -and their initialization- in the same order of the constructor parameters, I find it easier to read/maintain :)

src/Uno.UWP/UI/Input/GestureRecognizer.cs
Youssef1313: This set is different from SupportedGestures

src/Uno.UWP/UI/Input/RightTappedEventArgs.cs
Youssef1313: Agreed!

src/Uno.UWP/UI/Input/GestureRecognizer.Gesture.cs
Youssef1313: Nice catch. It's an unintended change.

src/Uno.UI/UI/Xaml/UIElement.Pointers.cs
dr1rrb: Sure we will have to and the flag would be configured to enable the new behavior by default to detect those. But providing a feature flag allows users to still upgrade uno without being blocked by a regression caused by this and reduces the pressure on our team to fix them.

Youssef1313: Ah, so it's a "just in case" there is a regression? Sounds good to me. Should we plan removing that flag in Uno 6 if no regressions were revealed?

src/Uno.UWP/UI/Input/GestureRecognizer.cs
dr1rrb: Yes it has the `HoldWithMouse` which is never enabled by `UIElement` (which uses this method). But basically I think that there is no reason to not allow the `Prevent<Events|Gestures>` to also disable that gesture and it would be more future proof to re-use that set than creating a specific one here (so when we will add support for the missing `CrossSlide` gesture, we will just have to review usages of the existing set).

Youssef1313: Great! Will make the change.

src/Uno.UI/UI/Xaml/UIElement.Pointers.cs
dr1rrb: Well currently we don't have any well formalized way to remove those flags ... I would say that the easiest would be to create an new issue :)

dr1rrb (approved):

PR #13916, opened 2023-10-06
docs: fix sample causing CS0267

mmarinchenko:
_[CS0267](https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs0267): The partial modifier can only appear immediately before 'class', 'record', 'struct', 'interface', or a method return type._


Youssef1313 (approved):
LGTM. Thanks @mmarinchenko

kazo0 (approved):

PR #13963, opened 2023-10-12
perf: Avoid InterpEntry for BindingPropertyHelper.GetPropertyType()

ebariche:
## PR Type

- Perf

### <samp>🤖 Generated by Copilot at db14c4e</samp>

This pull request enhances the data binding system by optimizing the `GetPropertyType` method for faster and safer access to property types. It also removes some unused or redundant namespaces from the `BindingPropertyHelper.cs` file.

## PR Checklist

### DIFF
--- a/src/Uno.UI/DataBinding/BindingPropertyHelper.cs
+++ b/src/Uno.UI/DataBinding/BindingPropertyHelper.cs
@@ -6,12 +6,12 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 
+using Uno.Collections;
 using Uno.Extensions;
 using Uno.Foundation.Logging;
 using System.Globalization;
 using System.Reflection;
-using Uno.UI.DataBinding;
-using Uno;
+
 using Windows.UI.Xaml;
 using System.Collections;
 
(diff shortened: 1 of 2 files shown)


src/Uno.UI/DataBinding/BindingPropertyHelper.cs
```diff
@@ -110,9 +110,9 @@ public static bool IsEvent(Type type, string property)
 
 		public static Type? GetPropertyType(Type type, string property, bool allowPrivateMembers)
 		{
-			var key = CachedTuple.Create(type, property, allowPrivateMembers);
+			var key = new GetPropertyTypeKey(type, property, allowPrivateMembers);
```
Youssef1313, line 113: Is this code path supposed to be executed only on the UI thread?
If so, I think we could change `GetPropertyTypeKey` to be mutable and have a static key, then we just mutate the existing key and use that for getting the value from the dictionary. This will make `GetPropertyType` not allocate a new `GetPropertyTypeKey` each time.

And also, if it's expected to be on UI thread only, I don't see why we need to lock below.

src/Uno.UI/DataBinding/BindingPropertyHelper.cs
```diff
@@ -1411,6 +1411,49 @@ private static object UnsetValueGetter(object unused)
...
+					{
+						return
+							k1._cachedHashcode == k2._cachedHashcode &&
+							k1._allowPrivateMembers == k2._allowPrivateMembers &&
+							k1._type == k2._type &&
+							k1._property == k2._property;
+
+					}
+
+					return false;
```
Youssef1313, line 1451:
```suggestion
					return x is null && y is null;
```

src/Uno.UI/DataBinding/BindingPropertyHelper.cs
jeromelaban: Yes, it's UI Thread only.

src/Uno.UI/DataBinding/BindingPropertyHelper.cs
```diff
@@ -1411,6 +1411,48 @@ private static object UnsetValueGetter(object unused)
 		/// <remarks>This method needs to be aligned with the symbols query in BindableTypeProvidersSourceGenerator.</remarks>
 		private static bool IsValidMetadataProviderType(Type type)
 			=> type.IsPublic && type.IsClass;
+
+		private class GetPropertyTypeKey
```
jeromelaban, line 1415: I think most of this should be a `readonly record struct` with only the hash caching part exposed.

Youssef1313: If we will follow https://github.com/unoplatform/uno/pull/13963#discussion_r1357153836 to reduce allocations of this class, then moving to a struct will re-introduce the allocations as boxing allocations since Hashtable deals with `object`s everywhere.

src/Uno.UI/DataBinding/BindingPropertyHelper.cs
ebariche: I was worried about multiple threads accessing this one, let's do it !

src/Uno.UI/DataBinding/BindingPropertyHelper.cs
ebariche: Also we wouldn't be able to mutate it.

src/Uno.UI/DataBinding/BindingPropertyHelper.cs
ebariche: We won't be inserting null keys so these branches can be avoided.

jeromelaban (approved):

PR #14057, opened 2023-10-22
perf: Optimize SVG array allocations

### LINKED ISSUE #14056: `SvgImageSource.ReadFromStreamAsync` could be optimized

Looking at the code:

```csharp
#if __ANDROID__ || __SKIA__
	private async Task<ImageData> ReadFromStreamAsync(Stream stream, CancellationToken ct)
	{
		if (stream.CanSeek && stream.Position != 0)
		{
			stream.Position = 0;
		}

		var memoryStream = new MemoryStream();
		await stream.CopyToAsync(memoryStream, 81920, ct);
		var data = memoryStream.ToArray();
		return ImageData.FromBytes(data);
	}
#endif
```

Two optimizations come to mind:

- Check if the given `stream` is already MemoryStream, and if so, skip creating a new MemoryStream and copying to it.
- Avoid the array allocation done by `memoryStream.ToArray()`. There are two ways that come in mind:
    - Instead of giving `byte[]` to ImageData, give it the MemoryStream. Note that:
        - the array is actually used to create a MemoryStream back again [here](https://github.com/unoplatform/uno/blob/2af7b1a6953cc501dd6b16c07c4510fd8a937828/src/AddIns/Uno.UI.Svg/SvgProvider.cs#L159)
        - Not sure if this optimization will not be safe if the given stream was already MemoryStream and we are reusing the same instance.
    - Use `TryGetBuffer`, which will reuse the internal array instead of creating a new array. I think this option makes more sense.

### Why is this needed

_No response_

### For which platform

_No response_

_No response_

Youssef1313:
GitHub Issue (If applicable): closes #14056

## PR Type

### <samp>🤖 Generated by Copilot at ee0ccf4</samp>

This pull request refactors the code related to loading and handling SVG data and image data in the Uno.UI project. It replaces the use of `byte[]` with `ArraySegment<byte>` in several classes and interfaces, such as `SvgProvider`, `ImageData`, and `ISvgProvider`, to improve memory efficiency and performance. It also modifies the `SvgImageSource` class to use the same buffer for the stream and the `SvgProvider`.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Media/Imaging/Svg/ISvgProvider.cs
+++ b/src/Uno.UI/UI/Xaml/Media/Imaging/Svg/ISvgProvider.cs
@@ -23,7 +23,7 @@ public interface ISvgProvider
 
 	event EventHandler? SourceLoaded;
 
-	Task<bool> TryLoadSvgDataAsync(byte[] imageData);
+	Task<bool> TryLoadSvgDataAsync(ArraySegment<byte> imageData);
 
 	void Unload();
 }
--- a/src/Uno.UI/UI/Xaml/Media/Imaging/SvgImageSource.cs
+++ b/src/Uno.UI/UI/Xaml/Media/Imaging/SvgImageSource.cs
@@ -144,10 +144,18 @@ public partial class SvgImageSource : ImageSource
 			stream.Position = 0;
 		}
 
-		var memoryStream = new MemoryStream();
-		await stream.CopyToAsync(memoryStream, 81920, ct);
-		var data = memoryStream.ToArray();
-		return ImageData.FromBytes(data);
+		if (stream is not MemoryStream memoryStream)
+		{
+			memoryStream = new MemoryStream();
+			await stream.CopyToAsync(memoryStream, 81920, ct);
+		}
+
+		if (!memoryStream.TryGetBuffer(out var buffer))
+		{
+			buffer = new ArraySegment<byte>(memoryStream.ToArray());
+		}
+
+		return ImageData.FromBytes(buffer);
 	}
 #endif
 
--- a/src/AddIns/Uno.UI.Svg/SvgProvider.cs
+++ b/src/AddIns/Uno.UI.Svg/SvgProvider.cs
@@ -98,7 +98,7 @@ public partial class SvgProvider : ISvgProvider
 #if !__NETSTD_REFERENCE__
 	async
 #endif
-	Task<bool> TryLoadSvgDataAsync(byte[] svgBytes)
+	Task<bool> TryLoadSvgDataAsync(ArraySegment<byte> svgBytes)
 	{
 #if __NETSTD_REFERENCE__
 		return Task.FromResult(false);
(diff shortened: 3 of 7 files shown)


src/Uno.UI/UI/Xaml/Media/Imaging/Svg/ISvgProvider.cs
```diff
@@ -23,7 +23,7 @@ public interface ISvgProvider
 
 	event EventHandler? SourceLoaded;
 
-	Task<bool> TryLoadSvgDataAsync(byte[] imageData);
+	Task<bool> TryLoadSvgDataAsync(ArraySegment<byte> imageData);
```
Youssef1313, line 26: **This is a breaking change** since `ISvgProvider` is public.

Does someone see a way to implement the optimization without breaking the API? :/

src/Uno.UI/UI/Xaml/Media/Imaging/SvgImageSource.cs
```diff
@@ -144,10 +144,18 @@ private async Task<ImageData> ReadFromStreamAsync(Stream stream, CancellationTok
...
+			memoryStream = new MemoryStream();
+			await stream.CopyToAsync(memoryStream, 81920, ct);
+		}
+
+		if (!memoryStream.TryGetBuffer(out var buffer))
+		{
+			buffer = new ArraySegment<byte>(memoryStream.ToArray());
+		}
+
+		return ImageData.FromBytes(buffer);
```
Youssef1313, line 158: The whole idea is in this block of code. Two optimizations are employed here:

- If given MemoryStream already, no need to create a new one and do the copying.
- Instead of calling ToArray which will create a new array, just get access to the internal array. Note that this has a behavior difference in that if the original array used to create the MemoryStream was mutated, that would now be reflected in ImageData were previously it wouldn't.

@MartinZikmund thoughts?

src/Uno.UI/UI/Xaml/Media/Imaging/Svg/ISvgProvider.cs
jeromelaban: I don't think there's a problem breaking this API, Uno is the only consumer. We may also put this API as internal (if we have not documented it)/

src/Uno.UI/UI/Xaml/Media/Imaging/SvgImageSource.cs
jeromelaban: It may be a problem to mutate what has been provided by the user, particularly if it does not reflect what WinUI does.

Youssef1313: Yeah I think we need to figure out what happens exactly in WinUI

src/Uno.UI/UI/Xaml/Media/Imaging/Svg/ISvgProvider.cs
MartinZikmund: We haven't, so can be made internal

PR #14066, opened 2023-10-23
ci: Fix ReferenceImplComparer to report accessible non-public members

Youssef1313:

## PR Type

### <samp>🤖 Generated by Copilot at 1049ea1</samp>

Improve reference assembly comparison using Mono.Cecil. Use helper methods to check accessibility of types and members in `Program.cs`.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.ReferenceImplComparer/Program.cs
+++ b/src/Uno.ReferenceImplComparer/Program.cs
@@ -61,7 +61,8 @@ namespace Uno.ReferenceImplComparer
 		private static bool IsAccessible(TypeDefinition type)
 		{
 			// https://github.com/jbevain/cecil/blob/56d4409b8a0165830565c6e3f96f41bead2c418b/Mono.Cecil/TypeAttributes.cs#L21-L26
-			return type.IsPublic || type.IsNestedPublic || type.IsNestedFamily || type.IsNestedFamilyOrAssembly;
+			return type.IsPublic ||
+				((type.IsNestedPublic || type.IsNestedFamily || type.IsNestedFamilyOrAssembly) && IsAccessible(type.DeclaringType));
 		}
 
 		private static bool IsAccessible(PropertyDefinition property)
--- a/src/Uno.ReferenceImplComparer/Program.cs
+++ b/src/Uno.ReferenceImplComparer/Program.cs
@@ -136,6 +136,18 @@ namespace Uno.ReferenceImplComparer
 				// 1. If the method is overrides. This is very problematic because in app code that is compiled against reference binaries, a base.XYZ call could refer to the wrong method.
 				// 2. Sometimes there is intent to introduce Skia-specific or Wasm-specific API. Doing so
 				//    under `#if __SKIA__` or `#if __WASM__` isn't enough because you can't call it when compiling against the reference binaries.
+
+				if (runtimeMember is MethodReference { Name: "Finalize" } finalizer && finalizer.Resolve().IsVirtual)
+				{
+					// Assumption is that it's okay for finalizers to be in runtime API but not reference API.
+					continue;
+				}
+				else if (runtimeMember is MethodReference @operator && @operator.Name is "op_Implicit" or "op_Explicit")
+				{
+					// Operators must be declared publicly.
+					continue;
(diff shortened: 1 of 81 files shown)


src/Uno.ReferenceImplComparer/Program.cs
```diff
@@ -46,13 +46,41 @@ static int Main(string[] args)
...
+
+		private static bool IsAccessible(TypeDefinition type)
+		{
+			// https://github.com/jbevain/cecil/blob/56d4409b8a0165830565c6e3f96f41bead2c418b/Mono.Cecil/TypeAttributes.cs#L21-L26
+			return type.IsPublic || type.IsNestedPublic || type.IsNestedFamily || type.IsNestedFamilyOrAssembly;
+		}
+
+		private static bool IsAccessible(PropertyDefinition property)
+		{
+			return property.GetMethod is not null && IsAccessible(property.GetMethod);
```
MartinZikmund, line 69: Should this also check for setters?

Youssef1313: Yeah I think it should, but I kept it the same for simplicity for now

Youssef1313: It should even check for the property accessibility itself.

Youssef1313: Somehow, PropertyDefinition doesn't seem to have `IsPublic` property and its friends. So keeping the original behavior for now to check the getter only which is 99.9999% has the same accessibility as the property itself.

src/Uno.UWP/Security/Authentication/Web/WebAuthenticationBrokerProvider.wasm.cs
```diff
@@ -11,7 +11,7 @@ namespace Uno.AuthenticationBroker
 {
 	partial class WebAuthenticationBrokerProvider
 	{
-		protected virtual string[] GetApplicationCustomSchemes()
```
Youssef1313, line 14: I'm assuming this is a safe change. Application code is compiled against the reference library which uses `IEnumerable<string>`, so my thought is that the return type change here can't be observed.

src/Uno.ReferenceImplComparer/Program.cs
Youssef1313: > I think it should even check for the property accessibility itself (not getters or setters). However, I never saw code where the getter doesn't match the property accessibility.

I take that back. We should check both.

src/Uno.UI/UI/Xaml/FrameworkElement.wasm.cs
```diff
@@ -443,15 +443,5 @@ private protected override void UpdateDOMProperties()
 				base.UpdateDOMProperties();
 			}
 		}
-
-		public override string ToString()
-		{
-			if (FeatureConfiguration.UIElement.RenderToStringWithId && !Name.IsNullOrEmpty())
-			{
-				return $"{base.ToString()}\"{Name}\"";
-			}
```
jeromelaban, line 452: That part was performance relate, iirc. Did you move it somewhere else?

Youssef1313: Yes, moved to `FrameworkElement.crossruntime.cs`

The problem with the current approach is that if in your app you did something like this:

```csharp
public class MyCustomFE : FrameworkElement
{
    // This DOES NOT call FrameworkElement.ToString. It calls object.ToString!
    // This is because the reference API doesn't have FrameworkElement.ToString so the compiler translates this as a call to object.ToString because that is what the compiler finds in compile-time.
    public override string ToString() => base.ToString();
}
```

src/Uno.UWP/Security/Authentication/Web/WebAuthenticationBrokerProvider.wasm.cs
jeromelaban: All good, it's fixing a bug that may have already been there.

jeromelaban (approved):

PR #14069, opened 2023-10-23
perf: Implement a FreezableList to use for DependencyObjectStore

Youssef1313:
image

**FreezableList is about 3x faster**

## PR Type

### <samp>🤖 Generated by Copilot at 6a662d9</samp>

This pull request introduces a new `FreezableList` class that improves the performance of dependency property collections in the `DependencyObjectStore` and `DependencyPropertyDetailsCollection` classes. It also adds a benchmark project to compare `FreezableList` and `ImmutableList`, and updates some nullable annotations and null checks in the code.

## PR Checklist

## Other information

### DIFF
--- a/src/SamplesApp/Benchmarks.Shared/Suite/Collections/FreezableListVsImmutableListBench.cs
+++ b/src/SamplesApp/Benchmarks.Shared/Suite/Collections/FreezableListVsImmutableListBench.cs
@@ -1,0 +1,32 @@
+#if !WINDOWS_UWP
+using BenchmarkDotNet.Attributes;
+using Uno.Collections;
+
+namespace SamplesApp.Benchmarks.Suite;
+
+[MemoryDiagnoser]
+public class FreezableListVsImmutableListBench
+{
+	[Benchmark(Baseline = true)]
+	public void ImmutableList()
+	{
+		var x = new ImmutableList<string>();
+		for (int i = 0; i < 10; i++)
+		{
+			x = x.Add(null);
+		}
+	}
+
+
+	[Benchmark]
+	public void FreezableList()
+	{
+		var x = new FreezableList<string>();
+		for (int i = 0; i < 10; i++)
+		{
+			x.Add(null);
+		}
+	}
+}
+
+#endif
--- a/src/Uno.UI/UI/Xaml/DependencyPropertyDetailsCollection.Bindings.cs
+++ b/src/Uno.UI/UI/Xaml/DependencyPropertyDetailsCollection.Bindings.cs
@@ -1,8 +1,6 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Text;
+﻿#nullable enable
+
 using Uno.Collections;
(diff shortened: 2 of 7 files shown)


src/SamplesApp/Benchmarks.Shared/Suite/Collections/FreezableListVsImmutableListBench.cs
```diff
@@ -0,0 +1,32 @@
+#if !WINDOWS_UWP
+using BenchmarkDotNet.Attributes;
+using Uno.Collections;
+
+namespace SamplesApp.Benchmarks.Suite;
+
+[MemoryDiagnoser]
```
Youssef1313, line 7: Note: The benchmark output didn't contain anything related to memory :/
Not sure why is that.

src/Uno.UI/UI/Xaml/DependencyPropertyDetailsCollection.Bindings.cs
```diff
@@ -84,11 +101,15 @@ internal void SuspendBindings()
...
 				_bindingsSuspended = true;
 
-				var bindings = _bindings.Data;
-
-				for (int i = 0; i < bindings.Length; i++)
+				var bindings = _bindings;
+				if (bindings is not null)
 				{
-					bindings[i].SuspendBinding();
+					bindings.Freeze();
```
jeromelaban, line 107: I like the idea, though it's not the same as having an immutable list, particularly if for some reason the property notification is reentrant. In such a case, handlers would be missed. Have you tried throwing of a list is already frozen and tried to be frozen again?

Youssef1313: @jeromelaban Would having a counter resolve your concern?
i.e, on Freeze we will increase the counter by one, and on Unfreeze we will decrease it, and we only push the queue to the list if the counter became zero?

i.e,

```
Count is 0 initially
Freeze: Count 1
Freeze: Count 2
Unfreeze: Count 1 // do nothing.
Unfreeze: Count 0 // push the queue to the actual list.
```

jeromelaban: That's where I'm not sure. In theory, the reentrant call (if that can happen) would invoke the new list, regardless of upper-stack freezing. That's where I'm asking if you throw when there's a reentrant call, does it actually happen with all the tests we have.

Youssef1313: @jeromelaban I added a `throw`. There are some test failures but they are not showing the exception, so I think the exception was caught somewhere and just causes the test to produce wrong results. I think I'll need to debug locally and see exactly what's happening. The hard part is understanding and finding a test case that will produce different results per the used approach.

MartinZikmund:
@ebariche Can you please check if this PR can be completed/reworked?

PR #14134, opened 2023-10-26
fix(scrollviewer): make scrolling cancelable when Handled = true on WASM

### LINKED ISSUE #10098: [WASM] Unable to disable Mouse Wheel Scrolling behaviour for ScrollViewer
### Current behavior

For some use cases it important (**mandatory**) to **disable** Scrolling behaviour with **Mouse Wheel**, but **leave** the opportunity to perform scrolling with **ScrollBar**. This is wery important for cases when custom scrolling behaviour supposed to be implemented. Such behaviour can be achieved if handle PointerWheelChanged event of nested controls and set Handled property, of PointerRoutedEventArgs, to True

```
<ScrollViewer>
    <Grid PointerWheelChanged="OnPointerWheelChanged" Padding="16 0" Background="Transparent">
        <ItemsRepeater PointerWheelChanged="OnPointerWheelChanged" ItemsSource="{x:Bind Items}">
            <ItemsRepeater.ItemTemplate>
                <DataTemplate x:DataType="x:String">
                    <TextBlock Text="{x:Bind}"/>
                </DataTemplate>
            </ItemsRepeater.ItemTemplate>
        </ItemsRepeater>
    </Grid>
</ScrollViewer>
```

```
private void OnPointerWheelChanged(object sender, PointerRoutedEventArgs e)
{
    e.Handled = true;
}
```

This code sucessfully disables scroll behaviour in UWP, but, has no effect in WASM project

If, possible, can you help with workaround for this problem? This is blocking issue for our team!

### Expected behavior

When PointerWheelChanged event handled in nested controls it blocks ScrollViewer scrolling with mouse wheel

1. Run UWP project
2. Scroll list with mouse wheel
3. Click "Disable/Enable Mouse weel scrolling"
4. Scroll list with mouse wheel
5. Run WASM project
6. repeat 2-4 steps

[MouseScrollDisable.zip](https://github.com/unoplatform/uno/files/9740794/MouseScrollDisable.zip)

### Workaround

_No response_

### Works on UWP/WinUI

Yes

### Environment

### NuGet package version(s)

_No response_

### Affected platforms

WebAssembly

### IDE

Visual Studio 2022

### IDE version

17.3.4

### Relevant plugins

_No response_

_No response_

ramezgerges:
GitHub Issue (If applicable): closes #10098

## PR Type

- Bugfix

## Copilot Summary

copilot:summary

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/ScrollContentPresenter/ScrollContentPresenter.wasm.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/ScrollContentPresenter/ScrollContentPresenter.wasm.cs
@@ -10,6 +10,7 @@ using Uno.Disposables;
 using System.Runtime.CompilerServices;
 using System.Text;
 using Windows.Foundation;
+using Windows.UI.Xaml.Input;
 using Uno.UI.Xaml;
 
 using Uno.UI.Extensions;
@@ -183,6 +184,17 @@ namespace Windows.UI.Xaml.Controls
 			base.OnLoaded();
 			RestoreScroll();
 			RegisterEventHandler("scroll", (EventHandler)OnScroll, GenericEventHandlers.RaiseEventHandler);
+
+			// a workaround to make scrolling cancelable on WASM
+			AddHandler(PointerWheelChangedEvent, new PointerEventHandler(CancelNativeScrollIfHandled), true);
+		}
+
+		private void CancelNativeScrollIfHandled(object _, PointerRoutedEventArgs args)
+		{
+			if (args.Handled)
+			{
+				((IHtmlHandleableRoutedEventArgs)args).HandledResult |= HtmlEventDispatchResult.PreventDefault;
+			}
 		}
 
 		private void RestoreScroll()
@@ -203,6 +215,8 @@ namespace Windows.UI.Xaml.Controls
 		{
 			base.OnUnloaded();
 			UnregisterEventHandler("scroll", (EventHandler)OnScroll, GenericEventHandlers.RaiseEventHandler);
+
+			RemoveHandler(PointerWheelChangedEvent, new PointerEventHandler(CancelNativeScrollIfHandled));
 
 			if (_rootEltUsedToProcessScrollTo is { } rootElt)
 			{


src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.wasm.cs
```diff
@@ -72,5 +74,29 @@ private void ChangeViewForOrientation(Orientation orientation, double scrollAdju
...
+		}
+
+		private protected override void OnUnloaded()
+		{
+			base.OnUnloaded();
+			RemoveHandler(KeyDownEvent, new KeyEventHandler(OnKeyDown));
+			RemoveHandler(PointerWheelChangedEvent, new KeyEventHandler(OnKeyDown));
+		}
+
+		private void OnKeyDown(object sender, KeyRoutedEventArgs args) => CancelNextNativeScroll = args.Handled;
```
dr1rrb, line 100: I'm wondering why this is in the `ScrollViewer` instead of the `ScrollContentPresenter`? I have the feeling that we could keep all scroll handling in the presenter.

src/Uno.UI/UI/Xaml/Controls/ScrollContentPresenter/ScrollContentPresenter.wasm.cs
```diff
@@ -320,8 +337,12 @@ private void TryProcessScrollTo()
 			}
 		}
 
-		private void OnScroll(object sender, EventArgs args)
+		private bool OnScroll(object sender, RoutedEventArgs routedEventArgs)
 		{
+			if (Scroller?.CancelNextNativeScroll ?? false)
+			{
+				return true;
+			}
```
dr1rrb, line 345: I've the feeling this is too late: this event is raised by the platform when it has scrolled, so we can update the managed properties and raise the event accordingly ... ignoring it here would just drive to be out-of-sync with the native scroll offsets?

src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.wasm.cs
ramezgerges: Initially, that was indeed how I implemented it, but AFAIK, our behaviour on the other platforms is that if args are Handled at the ScrollViewer level, we still don't scroll, so we need to wait until we go to the ScrollViewer to see if we handle it or not.

src/Uno.UI/UI/Xaml/Controls/ScrollContentPresenter/ScrollContentPresenter.wasm.cs
ramezgerges: MDN seems to agree with you, but I've tested this on WASM and it works (I've tested on Chromium). I can't find the link right now, but if I remember correctly, someone online also suggested this, so I guess we're fine?

src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.wasm.cs
dr1rrb: But we could move those event handlers in the SCP so all the logic will remain in the same class/file ?

src/Uno.UI/UI/Xaml/Controls/ScrollContentPresenter/ScrollContentPresenter.wasm.cs
dr1rrb: Ok, we need to add some details in comments then. I would also validate this at least on Firefox which is not chromium based.

src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.wasm.cs
ramezgerges: I guess we could subscribe to the ScrollViewer events from the SCP, but it would look awkward. It makes more sense to me for the ScrollViewer to handle its own `CancelNextNativeScroll` state and for the SCP to just read from it, but if you think moving everything to SCP makes more sense, we can go with that too.

src/Uno.UI/UI/Xaml/Controls/ScrollContentPresenter/ScrollContentPresenter.wasm.cs
ramezgerges: Just tested firefox, also works. Not sure what to add to the comments, we just `preventDefault` the event. If you have something more eloquent, feel free to add a comment there.

MartinZikmund (approved):

PR #14253, opened 2023-11-03
fix: Ensure `AffectsMeasure` and `AffectsArrange` are aligned with WinUI 3

### LINKED ISSUE #14254: `AffectsMeasure` and `AffectsArrange` are not always matching WinUI 3
### Current behavior

In some cases missing, some cases mixed up.

### Expected behavior

_No response_

_No response_

### Workaround

_No response_

### Works on UWP/WinUI

None

### Environment

_No response_

### NuGet package version(s)

_No response_

### Affected platforms

_No response_

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

MartinZikmund:
GitHub Issue (If applicable): closes https://github.com/unoplatform/uno/issues/14254

## PR Type

- Bugfix

### <samp>🤖 Generated by Copilot at 20f69b3</samp>

This pull request improves the layout performance and accuracy of various XAML controls by adjusting the metadata options of their dependency properties. It ensures that the controls update their layout when the properties that affect their size, position, or alignment change, and avoids unnecessary layout updates when the properties do not affect the layout. It also adds or modifies some properties to enable setting them in XAML and to trigger layout updates when they change. The changes affect the `Border`, `ContentControl`, `ContentPresenter`, `Control`, `DatePicker`, `Grid`, `Image`, `MediaPlayerElement`, `RelativePanel`, `SplitView`, `StackPanel`, `TextBlock`, `TextBox`, `ItemsControl`, `ItemsPresenter`, and `ScrollContentPresenter` classes, as well as the `DependencyPropertyMixins.tt` file.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
@@ -195,9 +195,15 @@ namespace Microsoft.UI.Xaml.Controls
 			set { SetValue(TemplateProperty, value); }
 		}
 
-		// Using a DependencyProperty as the backing store for Template.  This enables animation, styling, binding, etc...
 		public static DependencyProperty TemplateProperty { get; } =
-			DependencyProperty.Register("Template", typeof(ControlTemplate), typeof(Control), new FrameworkPropertyMetadata(null, FrameworkPropertyMetadataOptions.ValueDoesNotInheritDataContext, (s, e) => ((Control)s)?.OnTemplateChanged(e)));
+			DependencyProperty.Register(
+				nameof(Template), 
+				typeof(ControlTemplate), 
+				typeof(Control), 
+				new FrameworkPropertyMetadata(
+					null, 
+					FrameworkPropertyMetadataOptions.ValueDoesNotInheritDataContext | FrameworkPropertyMetadataOptions.AffectsMeasure,
+					(s, e) => ((Control)s)?.OnTemplateChanged(e)));
 
 		private void OnTemplateChanged(DependencyPropertyChangedEventArgs e)
 		{
@@ -640,7 +646,7 @@ namespace Microsoft.UI.Xaml.Controls
 				typeof(Control),
 				new FrameworkPropertyMetadata(
 					FontWeights.Normal,
-					FrameworkPropertyMetadataOptions.Inherits,
+					FrameworkPropertyMetadataOptions.Inherits | FrameworkPropertyMetadataOptions.AffectsMeasure,
 					(s, e) => ((Control)s)?.OnFontWeightChanged((FontWeight)e.OldValue, (FontWeight)e.NewValue)
 				)
 			);
@@ -662,7 +668,7 @@ namespace Microsoft.UI.Xaml.Controls
 				typeof(Control),
 				new FrameworkPropertyMetadata(
 					14.0,
-					FrameworkPropertyMetadataOptions.Inherits,
+					FrameworkPropertyMetadataOptions.Inherits | FrameworkPropertyMetadataOptions.AffectsMeasure,
 					(s, e) => ((Control)s)?.OnFontSizeChanged((double)e.OldValue, (double)e.NewValue)
 				)
 			);
@@ -684,7 +690,7 @@ namespace Microsoft.UI.Xaml.Controls
 				typeof(Control),
 				new FrameworkPropertyMetadata(
 					FontFamily.Default,
-					FrameworkPropertyMetadataOptions.Inherits,
+					FrameworkPropertyMetadataOptions.Inherits | FrameworkPropertyMetadataOptions.AffectsMeasure,
 					(s, e) => ((Control)s)?.OnFontFamilyChanged(e.OldValue as FontFamily, e.NewValue as FontFamily)
 				)
 			);
@@ -705,7 +711,7 @@ namespace Microsoft.UI.Xaml.Controls
 				typeof(Control),
 				new FrameworkPropertyMetadata(
 					FontStyle.Normal,
-					FrameworkPropertyMetadataOptions.Inherits,
+					FrameworkPropertyMetadataOptions.Inherits | FrameworkPropertyMetadataOptions.AffectsMeasure,
 					(s, e) => ((Control)s)?.OnFontStyleChanged((FontStyle)e.OldValue, (FontStyle)e.NewValue)
 				)
 			);
@@ -739,7 +745,7 @@ namespace Microsoft.UI.Xaml.Controls
 				typeof(Control),
 				new FrameworkPropertyMetadata(
 					Thickness.Empty,
-					FrameworkPropertyMetadataOptions.None,
+					FrameworkPropertyMetadataOptions.AffectsMeasure,
 					(s, e) => ((Control)s)?.OnPaddingChanged((Thickness)e.OldValue, (Thickness)e.NewValue)
 				)
 			);
@@ -762,7 +768,7 @@ namespace Microsoft.UI.Xaml.Controls
 				typeof(Control),
 				new FrameworkPropertyMetadata(
 					Thickness.Empty,
-					FrameworkPropertyMetadataOptions.None,
+					FrameworkPropertyMetadataOptions.AffectsMeasure,
 					(s, e) => ((Control)s)?.OnBorderThicknessChanged((Thickness)e.OldValue, (Thickness)e.NewValue)
 				)
 			);
--- a/src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
@@ -197,11 +197,11 @@ namespace Microsoft.UI.Xaml.Controls
 
 		public static DependencyProperty TemplateProperty { get; } =
 			DependencyProperty.Register(
-				nameof(Template), 
-				typeof(ControlTemplate), 
-				typeof(Control), 
+				nameof(Template),
+				typeof(ControlTemplate),
+				typeof(Control),
 				new FrameworkPropertyMetadata(
-					null, 
+					null,
 					FrameworkPropertyMetadataOptions.ValueDoesNotInheritDataContext | FrameworkPropertyMetadataOptions.AffectsMeasure,
 					(s, e) => ((Control)s)?.OnTemplateChanged(e)));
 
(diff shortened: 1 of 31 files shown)


src/Uno.UI/UI/Xaml/Shapes/Rectangle.cs
```diff
@@ -22,7 +22,7 @@ protected override Size MeasureOverride(Size availableSize)
 			typeof(Rectangle),
 			new FrameworkPropertyMetadata(
 				defaultValue: 0.0,
-				options: FrameworkPropertyMetadataOptions.AffectsArrange
+				options: FrameworkPropertyMetadataOptions.AffectsMeasure
```
Youssef1313, line 25: Hmm.. I don't think this should be affecting measure. But if that's what WinUI is doing let's follow that. I wouldn't expect this property to be changing dynamically that much in real scenarios, so doing the potentially unnecessary measure shouldn't be a problem.

src/Uno.UI/UI/Xaml/Shapes/Shape.cs
```diff
@@ -104,6 +104,7 @@ public double StrokeThickness
 			typeof(Shape),
 			new FrameworkPropertyMetadata(
 				defaultValue: 1.0d,
+				FrameworkPropertyMetadataOptions.AffectsMeasure,
```
Youssef1313, line 107: This was intentionally not added. In `OnStrokeThicknessChanged`, we only invalidate measure if stroke is not null.

Youssef1313: If we want to follow what WinUI is doing here, we should then delete the propertyChangedCallback.

@dr1rrb What are your thoughts?

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.cs
```diff
@@ -618,6 +618,7 @@ public TextWrapping TextWrapping
 				typeof(TextBox),
 				new FrameworkPropertyMetadata(
 					defaultValue: TextWrapping.NoWrap,
+					options: FrameworkPropertyMetadataOptions.AffectsMeasure,
```
Youssef1313, line 621: The code in `OnTextWrapipngChanged` will end up invalidating measure anyways. This probably applies to other properties for TextBox as well.

Not sure if that will end up invalidating measure twice which in turn will cause measuring to happen twice?

src/Uno.UI/UI/Xaml/Shapes/Rectangle.cs
MartinZikmund: It does according to this:

image

It is interesting though, maybe they do the calculations for it in measure

src/Uno.UI/UI/Xaml/Shapes/Shape.cs
MartinZikmund: Didn't catch that, good point! Can revert based on what you think

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.cs
MartinZikmund: Afaik it shouldn't, the invalidation is just scheduled, so it should not cause trouble

src/Uno.UI/UI/Xaml/Shapes/Shape.cs
Youssef1313: I have no strong opinion as long as we keep one of `AffectsMeasure` and `OnStrokeThicknessChanged`. Will let @dr1rrb weigh on which one we should use.

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.cs
jeromelaban: The only effect here is that `InvalidateMeasure` will be called multiple times. In general, we have debouncing conditions where we won't be traversing the tree when asking for invalidation multiple times, so it should not be an issue. That being said, we may want to look for duplicates, particularly if they're done in the property changed handlers because the DP did not have the flag set.

src/Uno.UI/UI/Xaml/FrameworkElement.cs
```diff
@@ -160,16 +160,12 @@ public FlowDirection FlowDirection
...
 #endif
-		[GeneratedDependencyProperty(DefaultValue = FlowDirection.LeftToRight, Options = FrameworkPropertyMetadataOptions.Inherits, ChangedCallback = true)]
-		public static DependencyProperty FlowDirectionProperty { get; } = CreateFlowDirectionProperty();
-
-		private void OnFlowDirectionChanged(FlowDirection oldValue, FlowDirection newValue)
-		{
+		[GeneratedDependencyProperty(DefaultValue = FlowDirection.LeftToRight, Options =
 #if SUPPORTS_RTL
-			this.InvalidateArrange();
-			VisualTreeHelper.GetParent(this)?.InvalidateArrange();
```
jeromelaban, line 170: This seems like a different type of change. In practice we should not have to invalidate the parent, but maybe the original commit will tell why it was there?

Youssef1313: @jeromelaban It didn't work for me without that when I originally wrote this, and I haven't put enough time to understand why when I found this quite hacky way.

I was discussing this specific change with Martin. All we need is to verify that RTL checkbox still works as expected. If it does, then the change should be okay.

src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
```diff
@@ -208,9 +208,15 @@ public ControlTemplate Template
...
-		// Using a DependencyProperty as the backing store for Template.  This enables animation, styling, binding, etc...
 		public static DependencyProperty TemplateProperty { get; } =
-			DependencyProperty.Register("Template", typeof(ControlTemplate), typeof(Control), new FrameworkPropertyMetadata(null, FrameworkPropertyMetadataOptions.ValueDoesNotInheritDataContext, (s, e) => ((Control)s)?.OnTemplateChanged(e)));
+			DependencyProperty.Register(
+				nameof(Template), 
+				typeof(ControlTemplate), 
+				typeof(Control), 
+				new FrameworkPropertyMetadata(
+					null, 
+					FrameworkPropertyMetadataOptions.ValueDoesNotInheritDataContext | FrameworkPropertyMetadataOptions.AffectsMeasure,
```
dr1rrb, line 218: I think that we should not change at large options like this ... I suspect that there are lot of case where the `InvalidateMeasure` was done in the property change callback. For instance this case you can see that the `INvalidateArrnge` is invoked here https://github.com/unoplatform/uno/pull/14253/files#diff-7e87918895f8b453fa65dab2ed76539d91c6f5c9f896ef2962aaadd7af2e121aR243 after some check, and noticeably a feature flag for that ...

src/Uno.UI/UI/Xaml/Shapes/Shape.cs
dr1rrb: Yes I think that yes we should not cause an invalidate measure when we can prevent them like in this case where the `Stroke is null` ... even if cleaner to just have the flag on all properties, it will have a negative perf impact so I would keep the `OnStrokeThicknessChanged` callback.

dr1rrb (commented):
Even if WinUI does invalidate the measure for each property, as it would have a negative perf impact and it's somehow an internal concern I think we should not apply this at large on almost all DP.

src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
MartinZikmund: Agreed, in such case I would still keep the flag there, commented out with an explainer why it is deliberately missing in our case, because it will make it easier to understand when we make changes in the future

src/Uno.UI/UI/Xaml/Shapes/Shape.cs
```diff
@@ -73,7 +73,7 @@ public Brush Stroke
 			typeof(Shape),
 			new FrameworkPropertyMetadata(
 				defaultValue: null,
-				options: FrameworkPropertyMetadataOptions.AffectsArrange,
+				options: FrameworkPropertyMetadataOptions.AffectsMeasure,
```
Youssef1313, line 76: `OnStrokeChanged` will invalidate measure only if necessary.

Youssef1313 (approved):

carldebilly (approved):
This will surely fix some obscur bugs ;-)

src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.cs
```diff
@@ -196,7 +196,7 @@ public FontStretch FontStretch
 			set => SetFontStretchValue(value);
 		}
 
-		[GeneratedDependencyProperty(ChangedCallbackName = nameof(OnFontStretchChanged), DefaultValue = FontStretch.Normal, Options = FrameworkPropertyMetadataOptions.Inherits)]
+		[GeneratedDependencyProperty(ChangedCallbackName = nameof(OnFontStretchChanged), DefaultValue = FontStretch.Normal, Options = FrameworkPropertyMetadataOptions.Inherits | FrameworkPropertyMetadataOptions.AffectsMeasure)]
```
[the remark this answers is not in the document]
MartinZikmund: That was removed among the changes in this PR:

image

And instead, I added explicit `InvalidateMeasure` in callsites that were not coming from `AffectsMeasure` property changes (which was just a few locations)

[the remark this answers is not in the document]
MartinZikmund: However, I guess that could be kept, as it does not result in duplicate measure anyway and there is no location that needs `InvalidateTextBlock` without a `InvalidateMeasure` call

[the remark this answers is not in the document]
Youssef1313: Ahh I see. Didn't notice it in the diff. I'll comment on that change separately

src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.cs
```diff
@@ -1228,7 +1225,11 @@ private Hyperlink FindHyperlinkAt(PointerRoutedEventArgs e)
 		}
 		#endregion
 
-		private void InvalidateTextBlock() => InvalidateTextBlockPartial();
+		private void InvalidateTextBlock()
```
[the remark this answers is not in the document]
MartinZikmund: I thought so too, but apparently on Android it the invalidation was needed - one of the tests started failing due to it missing - as the android native paint got invalidated only after the layout pass 🤔 . I could revert the last commit and adjust it to make the invalidation happen only on Android

PR #14316, opened 2023-11-07
chore: align Colorpicker with WinUI

ramezgerges:

## PR Type

### <samp>🤖 Generated by Copilot at 65cf571</samp>

This pull request updates and ports the ColorPicker control and related classes from the MUX repo to Uno, and fixes some bugs and accessibility issues. It also adds some helper methods and enums for color manipulation and display.

## PR Checklist

## Other information


src/Uno.UWP/UI/ctypes.cs
```diff
@@ -0,0 +1,171 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+// MUX Reference ctypes.h & ctypes.cpp, tag winui3/release/1.4.2
+
+using System;
+namespace Windows.UI;
+
+internal class Ctypes
```
Youssef1313, line 9:
```suggestion
internal static class Ctypes
```

src/Uno.UI/Microsoft/UI/Xaml/Controls/ColorPicker/ColorPickerSliderAutomationPeer.cs
```diff
@@ -91,51 +90,28 @@ public void RaisePropertyChangedEvent(Color oldColor, Color newColor, int oldVal
...
-					case ColorPickerHsvChannel.Value:
-						resourceStringWithoutName = ResourceAccessor.GetLocalizedStringResource(ResourceAccessor.SR_ValueStringValueSliderWithoutColorName);
-						break;
-					default:
-						return string.Empty;
-				}
 
-				return StringUtil.FormatString(
-					resourceStringWithoutName,
-					value);
```
jeromelaban, line 137: Seems like a big change, was it removed from the original source?

jeromelaban (approved):

src/Uno.UI/Microsoft/UI/Xaml/Controls/ColorPicker/ColorPickerSliderAutomationPeer.cs
ramezgerges: Yes, anything with `DownlevelHelper` has been removed in WinUI3.

robloo: It makes sense to remove some of these helpers as WinUI3 was a new jump in the minimum supported feature set. I ported the DownlevelHelper originally because it was used quite a bit in the code base. I also used it to disable .ToDisplayName(), etc. because the algorithm wasn't open sourced at the time to calculate that. I see you've added that algorithm now as well so there doesn't seem to be a reason to keep DownlevelHelper for color picker at least.

src/Uno.UI.RuntimeTests/MUX/Microsoft_UI_Xaml_Controls/ColorPicker/ColorPickerPage.xaml
```diff
@@ -0,0 +1,281 @@
+<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
+<local:TestPage
```
robloo, line 2: I believe most of the XAML here is copy/pasted from WinUI 3. Is it wise to change to tabs instead of spaces? When I originally ported the control I did NOT change the XAML at all so it could simply be copy/pasted to keep in sync without showing many spurious changes or require new indentation styles. Someday that sync could be automated even.

I've noticed for other WinUI 3 controls you seem to keep the WinUI 3 XAML indentation of two spaces as well.

src/Uno.UI/Microsoft/UI/Xaml/Controls/ColorPicker/ColorHelpers.cs
```diff
@@ -394,67 +410,120 @@ int sgn(T val)
 
 			bgraCheckeredPixelData.Capacity = width * height * 4;
 
-			//WorkItemHandler workItemHandler =
-			//(IAsyncAction workItem) =>
-			await Task.Run(() =>
-			{
-				for (int y = 0; y < height; y++)
+			WorkItemHandler workItemHandler =
```
robloo, line 413: I intentionally ported this portion to the modern async/await pattern in C#. UWP/WinUI in C++ obviously doesn't have that pattern available.

I suppose if Uno now fully supports what was done in UWP here the code is ok. But then again I wonder if this is one of the places it is better to do what C# does considering the code is now C# instead of C++.

src/Uno.UI/Microsoft/UI/Xaml/Controls/ColorPicker/ColorHelpers.cs
```diff
@@ -394,67 +410,120 @@ int sgn(T val)
...
-
-			//	asyncActionToAssign = null;
+			if (asyncActionToAssign is { })
+			{
+				asyncActionToAssign.Cancel();
+			}
 
-			// Uno Doc: Assumed normal priority is acceptable
-			await dispatcherHelper.RunAsync(CoreDispatcherPriority.Normal, () =>
+			// Uno Doc: ThreadPool is not implemented, so we work around it
```
robloo, line 455: Oh, ok, you ran into unsupported things here and did Task.

src/Uno.UI/Microsoft/UI/Xaml/Controls/ColorPicker/ColorHelpers.cs
```diff
@@ -394,67 +410,120 @@ int sgn(T val)
...
 				completedFunction?.Invoke(checkeredBackgroundBitmap);
 			});
-			//});
+		}
+
+		// Copied from WindowsCommunityToolkit/Microsoft.Toolkit.Uwp/Extensions/DispatcherQueueExtensions.cs
+		// Licensed to the .NET Foundation under one or more agreements.
+		// The .NET Foundation licenses this file to you under the MIT license.
+		// See the LICENSE file in the project root for more information.
+		private static Task EnqueueAsync(this DispatcherQueue dispatcher, Action function, DispatcherQueuePriority priority = DispatcherQueuePriority.Normal)
```
robloo, line 489: So this is brand-new and doesn't exist in WinUI it seems. Not sure that is the right approach then. I didn't need to do this in the original port and just using async/await and Task was fine. Calculating a bitmap in a byte array can be done on any thread.

src/Uno.UI/Microsoft/UI/Xaml/Controls/ColorPicker/ColorHelpers.cs
```diff
@@ -486,27 +555,23 @@ int sgn(T val)
 			return bitmap;
 		}
 
-		// UNO TODO: This method is only used when drawing the color spectrum using WinUI composition API's which are currently unsupported.
-		// This method is only partially ported to C#.
-		/*public static LoadedImageSurface CreateSurfaceFromPixelData(
+		public static LoadedImageSurface CreateSurfaceFromPixelData(
```
robloo, line 558: Keep in mind the ColorPicker has two spectrum rendering paths:

 1. As writable bitmaps normally
 2. Using a special composition brush for performance on supported systems

Since I highly doubt Uno now supports composition brushes and the composition system for rendering the spectrum, you probably should keep some comments here. Porting this method over to C# more fully is also probably fine for future-proofing the code but again without composition brushes is unnecessary.

src/Uno.UI/Microsoft/UI/Xaml/Controls/ColorPicker/ColorPicker.cs
```diff
@@ -1505,5 +1485,8 @@ private Color GetCheckerColor()
 
 			return checkerColor;
 		}
+
+		// Uno Doc: There is no drop-in C# equivalent for the C++'s '_wtoi' function
+		private int _wtoi(string text) => int.Parse(text, CultureInfo.InvariantCulture);
```
robloo, line 1490: Good idea to just re-implement a version of this

src/Uno.UI/Microsoft/UI/Xaml/Controls/ColorPicker/ColorHelpers.cs
```diff
@@ -1,14 +1,30 @@
...
+using Windows.System;
 using Windows.System.Threading;
 using Windows.UI;
-using Windows.UI.Core;
+using Windows.UI.Xaml.Media;
 using Windows.UI.Xaml.Media.Imaging;
+using Uno.Extensions;
+using Uno.UI.Dispatching;
+using Uno.UI.Helpers.WinUI;
+using size_t = System.UInt64; // Uno Doc: size_t is 8 bytes on mosts modern machines
```
robloo, line 21: Defining and using a new `size_t` here IMO is questionable. When converting from C++ to C# this no longer makes sense and just hurts readability of the code. I had converted all cases to use regular C# types and it was working fine.

src/Uno.UWP/UI/ColorHelper.cs
```diff
@@ -1,20 +1,44 @@
...
+			// IFC_RETURN(ColorHelper_ToDisplayNameId(color, &colorNameResourceId));
+			// IFC_RETURN(DXamlCore::GetCurrent()->GetLocalizedResourceString(colorNameResourceId, returnValue));
+			//
+			// return S_OK;
+
+			return Enum.GetName(Enum.GetValues(typeof(KnownColors)).Cast<KnownColors>().MinBy(c => ColorDistance(new Color((uint)c), color))!);
+
+			static double ColorDistance(Color color1, Color color2)
+			{
+				// https://stackoverflow.com/a/8796867
```
robloo, line 27: Interesting, is this algorithm not open sourced in WinUI 3?

For Avalonia I tried two things:
 1) Using the 3D Euclidian distance (geometric distance) between RGB colors. That works OK.
 2) Finally decided on doing 3D Euclidian distance in HSV color space. There is rounding included.

If the algorithm you have here works ok, great. Otherwise, feel free to use some ideas from Avalonia as well (which also implements caching).

https://github.com/AvaloniaUI/Avalonia/blob/0567a6dbe7dc122ba90e019320708e0cbc51b635/src/Avalonia.Controls.ColorPicker/Helpers/ColorHelper.cs#L63

src/Uno.UWP/UI/StringConversions.cs
```diff
@@ -0,0 +1,58 @@
...
+// Partially ported
+// MUX Reference StringConversions.cpp, tag winui3/release/1.4.2
+
+using System;
+namespace Windows.UI;
+
+internal class StringConversions
+{
+	public static void
+		UnsignedFromHexString(
```
robloo, line 13: If I understand what this is doing, I really would NOT have ported it over to Uno. It is not code that is easy to follow. Using C#'s built-in parsing methods can do most of this in one line of code. Again, I would not try to copy verbatim C++ and Windows methods for stuff like this... that said, as it could be used elsewhere, I guess I see the point in implementing the API at least.

src/Uno.UWP/UI/ctypes.cs
robloo, line 9: Since this is internal (not part of public API) are you SURE this needs to be in Uno? This is duplicating things that .NET can already do I think. We shouldn't duplicate code like this just because it was needed in C++. C#/.NET should be used for low-level stuff like this now IMO.

src/Uno.UI/Microsoft/UI/Xaml/Controls/ColorPicker/ColorHelpers.cs
ramezgerges: Yeah, I found it questionable as well, but I guess it makes the code nicer to compare. No strong opinions here, could remove if it was found really ugly.

src/Uno.UI/Microsoft/UI/Xaml/Controls/ColorPicker/ColorHelpers.cs
ramezgerges: You're right. I stopped half-way through and decided it isn't worth it, but I kept the parts that I did port so that whoever needs to port the code later doesn't have to duplicate the work.

src/Uno.UI/Microsoft/UI/Xaml/Controls/ColorPicker/ColorHelpers.cs
ramezgerges: This is takes from the community toolkit and seems like a very reasonable extension of `TryEnqueue` that should probably have been present iniaitially. This is different from the WinUI source because C++ doesn't have async/await. so `TryEnqueue` works for everything. Probably fine anyway, but as usual I try to make the code as close to the original.

src/Uno.UI/Microsoft/UI/Xaml/Controls/ColorPicker/ColorHelpers.cs
ramezgerges: Yes, I eventually had to make a workaround like you did, but I tried to keep as much of the original source as possible. This will help if we need to compare it later.

src/Uno.UWP/UI/ColorHelper.cs
ramezgerges: I'm not sure if it's somewhere in the source, but I couldn't find it. Since the original implementation also uses localization and a bunch of other stuff that I couldn't port, I thought I might as well go with something simple for the distance calculation. I tested out and I think it works pretty well, but if we find it problematic later, I will definitely take a look at your implementation. Thanks. Caching seems like it would be a definite improvement, so I will add that.

src/Uno.UWP/UI/StringConversions.cs
ramezgerges: I agree, I only started porting it because the built-in methods don't have the exact semantics of the c++ implementation. But yes, I tried to find c# replacements whenever I could (like in `_wtoi`).

src/Uno.UWP/UI/ctypes.cs
ramezgerges: This is just to support `StringConversions` which, like you said, might be unnecessary. I think keeping it around wouldn't be a terrible idea as it could help keep the code looking similar to the original, but its usage should be heavily discouraged.

src/Uno.UI.RuntimeTests/MUX/Microsoft_UI_Xaml_Controls/ColorPicker/ColorPickerPage.xaml
Youssef1313: The problem is that VS will try to re-format when you paste code.

I faced this problem earlier when updating ColorPicker and at that time I added this `.editorconfig`: https://github.com/unoplatform/uno/blob/f3f7a51e2c09440e23d08247736067a4d2b6651c/src/Uno.UI/Microsoft/UI/Xaml/.editorconfig

I think we should add the same .editorconfig to `Uno.UI.RuntimeTests/MUX` directory

ramezgerges: Good call.

src/Uno.UI/Microsoft/UI/Xaml/Controls/ColorPicker/ColorHelpers.cs
robloo: This is a grey-area where to me it makes sense to stay with C#/.NET. Code simply doesn't exist to keep it 1-to-1. If you made the decision on the opposite side of that grey area I guess it's fine as long as it still works. It does make sense to keep compatibility as close as possible. I wouldn't have changed it if it was me though.

src/Uno.UI/Microsoft/UI/Xaml/Controls/ColorPicker/ColorHelpers.cs
robloo: Same comments as above. When functionality simply doesn't exist my instinct is to go with C#/.NET convention rather than try to shoe-horn in a **new** solution somewhere in-between C#/.NET and C++ all in an attempt to make the code "closer to the original". It seems better not to do new things here.

Someone else is going to have to make the final decision here.

src/Uno.UI/Microsoft/UI/Xaml/Controls/ColorPicker/ColorHelpers.cs
robloo: Can you add some of the removed comments back then?

```
// UNO TODO: This method is only used when drawing the color spectrum using WinUI composition API's which are currently unsupported.
// This method is only partially ported to C#.
```

src/Uno.UI/Microsoft/UI/Xaml/Controls/ColorPicker/ColorHelpers.cs
robloo: Let's let someone else break the tie then. As stated, I usually would default to C#/.NET convention in these cases (and did so in the original port).

src/Uno.UWP/UI/ColorHelper.cs
robloo: I'll dig around in WinUI 3 code when I have a chance. This is something that could be updated later in another PR as well. I was just curious what you might have found yourself.

src/Uno.UWP/UI/StringConversions.cs
robloo: I would keep the method `UnsignedFromHexString` and implement it using C#/.NET code.

This is a general problem in all the code ported so far. C++/WinUI has completely different culture/globalization mechanisms than .NET. However, it is better to use built-in .NET rules in cases like this. As a benefit, you also don't need to re-write methods like this -- just use .NET.

src/Uno.UWP/UI/ctypes.cs
robloo: In this case, if I was making the final review I would remove it completely. It is internal and not part of the UWP/WinUI contract. We should use C#/.NET built-in methods for things like this. It's going to be extremely difficult to maintain this type of code otherwise.

Issue #7271, opened 2021-10-04
[Skia] Restore Runtime Tests


jeromelaban:
### Current behavior

After the introduction of Skia runtime tests, the following runtime tests are failing:

- [ ] `ValidateLayoutRoundingForPixelDimensions`
- [ ] `CanReplaceSingleItem`
- [ ] `When_CloseLast_Then_FileDeleted`
- [ ] `Should_Close_Open_Popups`
- [ ] `When_TransformToVisual_WithMargin`
- [ ] `When_MinWidth_SmallerThan_AvailableSize`
- [ ] `TestVariousArrangedPosition`
- [ ] `When_ProgressRing_Visible`
- [ ] `When_Calling_Select_With_Negative_Values`
- [ ] `When_Calling_Select_With_In_Range_Values`
- [ ] `When_Calling_Select_With_Out_Of_Range_Length`
- [ ] `When_Calling_Select_With_Out_Of_Range_Start`
- [ ] `When_Nested_In_Native_View`

Those tests are now ignore in order to stabilize existing passing tests.

### Expected behavior

The tests pass.

### NuGet package version(s)

3.11-dev

### Affected platforms

Skia (GTK on Linux/macOS/Windows)

Youssef1313:
`When_Calling_Select_With_Negative_Values` this one is passing for me locally. The code for it is actually not platform-specific.

For other selection tests, `Select` in `TextBoxViewExtension` class seems to be missing `EnsureWidget` call. This was causing us to operate on a null `_currentInputWidget` and just returning without doing anything. BUT, even after adding the `EnsureWidget` call. The native APIs doesn't seem to be working. Will try to investigate further.

Youssef1313:
`ValidateLayoutRoundingForPixelDimensions` is passing locally for me.

`Should_Close_Open_Popups` is failing due to `RegisterOpenPopup` being called twice for the same popup. It has an `AddDistinct` method, but to me the `AddDistinct` will never work since we're comparing a newly instantiated `WeakReference`, not the popup itself. I'm not sure if using `WeakReferenceEqualityComparer` is the correct fix here.

Anyway, instead of completely disable this test, let's turn the check into `> 0` instead of `== 1` (only on Skia), until it's properly fixed.

jeromelaban:
We can enable those back one we find the proper fixes, the point of the associated PR is to have regression validation. 

For the ValidateLayoutRoundingForPixelDimensions is suspect this is related to DPI awareness.

[...]

PR #14368, opened 2023-11-09
fix: Workaround a lifecycle bug to allow ContentControl to be measured properly when not in visual tree

Youssef1313:
GitHub Issue (If applicable): Related #7271

## PR Type

### <samp>🤖 Generated by Copilot at 5b95944</samp>

This pull request fixes a unit test failure on Skia by overriding the `MeasureOverride` method of `ContentControl` to handle a null first child. This is a temporary workaround for a bug in the ContentControl measurement logic on Skia.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/ContentControl/ContentControl.skia.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/ContentControl/ContentControl.skia.cs
@@ -33,6 +33,7 @@ namespace Windows.UI.Xaml.Controls
 			// 1. FrameworkElement.MeasureCore calls 'InvokeApplyTemplate'
 			// 2. ContentControl overrides ApplyTemplate to do a bunch of stuff, including a call to CreateDefaultVisuals.
 			// 3. CreateDefaultVisuals adds the Content as Child if it's a UIElement.
+			// Tracking issue: https://github.com/unoplatform/uno/issues/190
 			var child = this.FindFirstChild() ?? (Content as UIElement);
 			return child != null ? MeasureElement(child, availableSize) : new Size(0, 0);
 		}
--- a/src/Uno.UI/UI/Xaml/Controls/ContentControl/ContentControl.skia.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/ContentControl/ContentControl.skia.cs
@@ -25,6 +25,16 @@ namespace Windows.UI.Xaml.Controls
 			RemoveChild(ContentTemplateRoot);
 		}
 
-		protected override Size MeasureOverride(Size availableSize) => base.MeasureOverride(availableSize);
+		protected override Size MeasureOverride(Size availableSize)
+		{
+			// This is the same as Control.MeasureOverride, except that we use Content if FindFirstChild is null.
+			// The fact that FindFirstChild is null is actually a BUG!
+			// The way this works in WinUI is:
+			// 1. FrameworkElement.MeasureCore calls 'InvokeApplyTemplate'
+			// 2. ContentControl overrides ApplyTemplate to do a bunch of stuff, including a call to CreateDefaultVisuals.
+			// 3. CreateDefaultVisuals adds the Content as Child if it's a UIElement.
+			var child = this.FindFirstChild() ?? (Content as UIElement);
+			return child != null ? MeasureElement(child, availableSize) : new Size(0, 0);
+		}
 	}
 }
--- a/src/Uno.UI/UI/Xaml/UIElement.Layout.crossruntime.cs
+++ b/src/Uno.UI/UI/Xaml/UIElement.Layout.crossruntime.cs
@@ -149,11 +149,6 @@ namespace Windows.UI.Xaml
 				return; // Only FrameworkElements are measurable
 			}
 
-			if (double.IsNaN(availableSize.Width) || double.IsNaN(availableSize.Height))
-			{
-				throw new InvalidOperationException($"Cannot measure [{GetType()}] with NaN");
-			}
-
 			if (Visibility == Visibility.Collapsed)
 			{
 				if (availableSize == LastAvailableSize)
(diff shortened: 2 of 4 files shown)


src/Uno.UI/UI/Xaml/Controls/ContentControl/ContentControl.skia.cs
```diff
@@ -25,6 +25,16 @@ public partial class ContentControl
 			RemoveChild(ContentTemplateRoot);
 		}
 
-		protected override Size MeasureOverride(Size availableSize) => base.MeasureOverride(availableSize);
+		protected override Size MeasureOverride(Size availableSize)
+		{
+			// This is the same as Control.MeasureOverride, except that we use Content if FindFirstChild is null.
```
MartinZikmund, line 30: Can you create an issue to track this issue specifically? Just so we can include it in lifecycle fixes backlog

MartinZikmund (approved):

src/Uno.UI/UI/Xaml/Controls/ContentControl/ContentControl.skia.cs
Youssef1313: @MartinZikmund I guess this is already https://github.com/unoplatform/uno/issues/190 ?

The issue mentions:

> On Windows the template of a control is being applied only when this control is in the visual tree and either we explicitly invoke the ApplyTemplate method, or we wait for the first measure pass.

The case I'm dealing with here is the "wait for the first measure pass". I'll add a link to the issue in code comments.

ajpinedam (approved):

src/Uno.UI/UI/Xaml/Controls/ContentControl/ContentControl.skia.cs
```diff
@@ -25,6 +25,17 @@ public partial class ContentControl
...
+			// The fact that FindFirstChild is null is actually a BUG!
+			// The way this works in WinUI is:
+			// 1. FrameworkElement.MeasureCore calls 'InvokeApplyTemplate'
+			// 2. ContentControl overrides ApplyTemplate to do a bunch of stuff, including a call to CreateDefaultVisuals.
+			// 3. CreateDefaultVisuals adds the Content as Child if it's a UIElement.
+			// Tracking issue: https://github.com/unoplatform/uno/issues/190
+			var child = this.FindFirstChild() ?? (Content as UIElement);
+			return child != null ? MeasureElement(child, availableSize) : new Size(0, 0);
```
jeromelaban, line 38: Looks like this is causing another issue:

https://dev.azure.com/uno-platform/Uno%20Platform/_TestManagement/Runs?runId=2583081&_a=resultSummary&resultId=101519

Could be that we were testing for an invalid behavior, though:
https://github.com/unoplatform/uno/blob/7a65f71b34f7bdb5fb7648f2423ed529c342ec50/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_FrameworkElement.cs#L266

Youssef1313: @jeromelaban This test also fails in WinUI. However, there is another NaN-measuring test that doesn't fail in WinUI. So I still need to figure that out 👀

Youssef1313: @jeromelaban The exception on WinUI is a weird "COMException". It happens when `Measure` is called from C# with NaN. However, if `Measure` is called with `NaN` somewhere in the C++ code, it is allowed.

I think it should be acceptable to modify this test and expect no exception be thrown?

To see this behavior on WinUI, use this code

```csharp
            var grid = new Grid();
            var SUT = new MyControl02();
            SUT.Content = new Grid();
            grid.Children.Add(SUT);
            SUT.BaseAvailableSize = new Size(double.NaN, double.NaN);
            grid.Measure(new Size(double.PositiveInfinity, double.PositiveInfinity));
```

and

```csharp
    public partial class MyControl02 : ContentControl
    {
        public List<Size> MeasureOverrides { get; } = new List<Size>();

        public Size? BaseAvailableSize;

        protected override Size MeasureOverride(Size availableSize)
        {
            MeasureOverrides.Add(availableSize);
            //(Content as UIElement).Measure(BaseAvailableSize ?? availableSize);
            //return (Content as UIElement).DesiredSize;
            return base.MeasureOverride(BaseAvailableSize ?? availableSize);
        }
    }
```

The above code will work without any crashes/exceptions.
However, `base.MeasureOverride` will end up calling `ContentControl.MeasureOverride` which in turn will call `Measure` with NaN.

Now, if you remove the `base.MeasureOverride` line and uncomment the two lines above it, it will crash.

image

jeromelaban: This was likely present to align with some original behavior. Did you take a look at why this test was originally added?

Otherwise, if adjusting the test matches more with what WinUI does, I'm good with it.

Youssef1313: @jeromelaban https://github.com/unoplatform/private/issues/93

src/Uno.UI/UI/Xaml/UIElement.Layout.crossruntime.cs
```diff
@@ -149,11 +149,6 @@ public void Measure(Size availableSize)
 				return; // Only FrameworkElements are measurable
 			}
 
-			if (double.IsNaN(availableSize.Width) || double.IsNaN(availableSize.Height))
-			{
-				throw new InvalidOperationException($"Cannot measure [{GetType()}] with NaN");
-			}
```
jeromelaban, line 155: That was originally placed because UWP refused to measure with NaN. Won't this hide other bugs?

Youssef1313: @jeromelaban When I tested in WinUi, it was interesting as I came across two cases:

1. C# code tries to measure with NaN. This fails with COMException.
2. C# code that causes an internal measure with NaN where the "Measure" call happens in the C++ code of WinUI. This does **not** fail.

We can't replicate that behavior, and the fix for ContentControl revealed the case "2" where it works in WinUI and throws in our case.

Youssef1313:
Making draft. It seems this can get fixed with https://github.com/unoplatform/uno/pull/14528.

We'll see after https://github.com/unoplatform/uno/pull/14528 is completed, then we can either close or merge this.

Youssef1313:
Fixed in lifecycle already :)

Issue #9417, opened 2022-08-03
[Skia] Skia-based TextBox implementation


mikernet:
## What would you like to be added:

A Skia-based TextBox implementation.

## Why is this needed:

Current implementation shows a TextBlock and overlays a native platform TextBox on focus which has many drawbacks.

## For which Platform:

- [ ] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] WebAssembly renderers for Xamarin.Forms
- [ ] macOS
- [ ] Skia
  - [ ] WPF
  - [ ] GTK (Linux)
  - [ ] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

## Anything else we need to know?

MartinZikmund:
# Repro generator

This issue was classified as not reproducible with a code sample or test.

**Reason:** Enhancement for Skia TextBox implementation.

PR #14371, opened 2023-11-09
feat(textbox + textblock): TextBox Undo/Redo, TextBlock selection, rendering fixes and major cleanup

### LINKED ISSUE #14415: [Skia] `TextBox` Undo/redo support

ramezgerges:
GitHub Issue (If applicable): #9417, closes #14415

## PR Type

- Feature

### <samp>🤖 Generated by Copilot at 5684f14</samp>

This pull request enhances the `TextBox` control on various platforms by fixing bugs, improving performance, and adding password masking support. It modifies the `TextBox.cs` and `InlineCollection.skia.cs` files.

## PR Checklist

## Other information


src/Uno.UI.Runtime.Skia.Gtk/Input/GtkCorePointerInputSource.cs
```diff
@@ -486,7 +486,7 @@ private void UseDevice(PointerPoint pointer, Gdk.Device device)
 
 		var pointerPoint = new Windows.UI.Input.PointerPoint(
 			frameId: time,
-			timestamp: time,
+			timestamp: time * (ulong)TimeSpan.TicksPerMillisecond, // time is in ms, timestamp is in ticks
```
jeromelaban, line 489: @dr1rrb do you have an opinion on this?

src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.cs
```diff
@@ -46,6 +46,20 @@ public partial class TextBlock : DependencyObject
 
 		private Run _reusableRun;
 		private bool _skipInlinesChangedTextSetter;
+		private bool _subscribeToPointerEvents;
+		private bool _isPressed;
+		private (int start, int end) Selection
```
jeromelaban, line 51: `(int start, int end)` would benefit from being a record that can be referenced, such as `TextSpan` or similar.

src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.iOSmacOS.cs
```diff
@@ -21,5 +21,10 @@ private CGRect GetDrawRect(CGRect rect)
 
 			return rect;
 		}
+
+		public TextBlock()
+		{
+			_hyperlinks.CollectionChanged += HyperlinksOnCollectionChanged;
```
jeromelaban, line 27: Why does this need to be in every target? Is there something specific to a target that makes it not available to others?

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.cs
```diff
@@ -1214,7 +1215,15 @@ public void PasteFromClipboard()
 			_ = Dispatcher.RunAsync(CoreDispatcherPriority.High, async () =>
 			{
 				var content = Clipboard.GetContent();
-				var clipboardText = await content.GetTextAsync();
+				string clipboardText;
+				try
+				{
+					clipboardText = await content.GetTextAsync();
+				}
+				catch (InvalidOperationException)
```
jeromelaban, line 1223: Let's log here, even if in debug or trace levels.

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.cs
```diff
@@ -1228,7 +1237,13 @@ public void PasteFromClipboard()
 
 				PasteFromClipboardPartial(clipboardText, selectionStart, selectionLength, currentText);
 
+#if __SKIA__
+				_suppressCurrentlyTyping = true;
+#endif
 				Text = currentText;
+#if __SKIA__
+				_suppressCurrentlyTyping = false;
+#endif
```
jeromelaban, line 1246: This would need to be wrapped in a try/finally, in case setting the text fails for any reason.

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.cs
```diff
@@ -1255,7 +1270,13 @@ public void CutSelectionToClipboard()
 		{
 			CopySelectionToClipboard();
 			CutSelectionToClipboardPartial();
+#if __SKIA__
+			_suppressCurrentlyTyping = true;
+#endif
 			Text = Text.Remove(SelectionStart, SelectionLength);
+#if __SKIA__
+			_suppressCurrentlyTyping = false;
+#endif
```
jeromelaban, line 1279: Same here for the try/finally.

src/Uno.UWP/UI/Input/GestureRecognizer.cs
```diff
@@ -24,7 +24,7 @@ public partial class GestureRecognizer
 		internal const int TapMaxXDelta = 10;
 		internal const int TapMaxYDelta = 10;
 
-		internal const ulong MultiTapMaxDelayTicks = TimeSpan.TicksPerMillisecond * 1000;
+		internal const ulong MultiTapMaxDelayTicks = TimeSpan.TicksPerMillisecond * 500;
```
jeromelaban, line 27: @dr1rrb this also looks like something you'd be interested in.

src/Uno.UWP/UI/Input/Preview.Injection/InputInjector.cs
```diff
@@ -68,6 +68,8 @@ public void InjectTouchInput(IEnumerable<InjectedInputTouchInfo> input)
 		var touch = _touch!.Value.state;
 		foreach (var info in input)
 		{
+			touch.Timestamp = (ulong)DateTime.Now.Ticks; // Timestamp is stale
```
jeromelaban, line 71: @dr1rrb

src/Uno.UWP/UI/Input/Preview.Injection/InputInjector.cs
```diff
@@ -93,6 +95,7 @@ public void InjectMouseInput(IEnumerable<InjectedInputMouseInfo> input)
 	{
 		foreach (var info in input)
 		{
+			_mouse.Timestamp = (ulong)DateTime.Now.Ticks; // Timestamp is stale
```
jeromelaban, line 98: @dr1rrb

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.skia.cs
```diff
@@ -31,21 +31,34 @@ private enum ContextMenuItem
...
+
+	private bool _clearHistoryOnTextChanged = true;
+
+	private bool _currentlyTyping;
+	private bool _suppressCurrentlyTyping;
+	private (int selectionStart, int selectionLength, bool selectionEndsAtTheStart) _selectionWhenTypingStarted;
+	private string _textWhenTypingStarted;
+
+	private int _historyIndex;
+	private List<(TextBoxAction action, int selectionStart, int selectionLength, bool selectionEndsAtTheStart)> _history = new(); // the selection of an action is what was selected right before it happened. Might turn out to be unnecessary.
```
jeromelaban, line 57: Let's use a record instead of tuples here.

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.skia.cs
```diff
@@ -58,6 +71,74 @@ private enum ContextMenuItem
 
 	internal ContentControl ContentElement => _contentElement;
 
+	// We track what constitutes one typing "action" that can be undone/redone. The general gist is that
+	// any sequence of characters (with backspace allowed) without any navigation moves (pointer click, arrow keys, etc.)
+	// will be one "run"/"action". However, there are some arbitrary exceptions, so that is only a rule of thumb.
+	private bool CurrentlyTyping
```
jeromelaban, line 77: Let's make a method instead of a property, it will be more in-line with what the rest of the codebase uses for this, particularly because setting the value may end up not being done. Something like `TrySetCurrentlyTyping`.

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.skia.cs
```diff
@@ -58,6 +71,74 @@ private enum ContextMenuItem
...
+		DependencyProperty.Register(
+			nameof(CanRedo), typeof(bool),
+			typeof(TextBox),
+			new FrameworkPropertyMetadata(true));
+
+	public bool CanRedo
+	{
+		get => (bool)GetValue(CanRedoProperty);
+		private set => SetValue(CanRedoProperty, value);
+	}
```
jeromelaban, line 134: Let's use generated properties for this.

src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.iOSmacOS.cs
ramezgerges: HyperlinksOnCollectionChanged just checks whether we should be subscribing to pointer events (previously was inside `UpdateHyperlinks`). I've basically included it on all targets except WASM as it wasn't included there in the first place (WASM has its own `UpdateHyperlinks`). Since the constructor for each target is not shared, I've had to manually add it to each target.

src/Uno.UI.Runtime.Skia.Gtk/Input/GtkCorePointerInputSource.cs
dr1rrb: Yes I have an opinion : we should do the same fix with "FrameBuffer" (for mouse and touch) where the timestamp is in micro sec instead of ticks :) (According to the doc of `libinput_event_pointer_get_time_usec` and `libinput_event_touch_get_time_usec`)

BTW sounds unrelated to the current PR, should we move those fixes to another one?

src/Uno.UWP/UI/Input/Preview.Injection/InputInjector.cs
dr1rrb: This however is invalid, the `touch` represent the "current state" of the injected touch input, i.e. the info extracted from the last injected touch pointer. It will be updated on line 81 by the `touch.Update(args);`, you should not update it "before" raising the event.

Also for injected pointers, the time is virtual and controlled by the user using the [`TimeOffsetInMilliseconds` and `PerformanceCount`](https://github.com/unoplatform/uno/blob/master/src/Uno.UWP/UI/Input/Preview.Injection/InjectedInputPointerInfo.cs#L19). Using the "real time" is invalid.

Looking at what you did, I suspect the issue is here https://github.com/unoplatform/uno/blob/master/src/Uno.UWP/UI/Input/Preview.Injection/InjectedInputPointerInfo.cs#L84. I think that we should invert and use the `PerformanceCount` for the `FrameId` and the `TimeOffsetInMilliseconds * 1000` for the `Timestamp`.

src/Uno.UI.Runtime.Skia.Gtk/Input/GtkCorePointerInputSource.cs
dr1rrb: Edit: just searched the doc for the timestamp: "The time, relative to the system boot time, in microseconds." (https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.input.pointerpoint.timestamp?view=windows-app-sdk-1.4#microsoft-ui-input-pointerpoint-timestamp)

So it means that here you should do
```suggestion
			timestamp: time * 1000, // time is in ms, timestamp is in micro-seconds
```

But it also means that we need to fix `WPFCoreInputSource`, and the gesture recognizer assumes that the timestmp is in ticks (e.g. [for long press](https://github.com/unoplatform/uno/blob/master/src/Uno.UWP/UI/Input/GestureRecognizer.Gesture.cs#L392)) But then it's a complete different issue https://github.com/unoplatform/uno/issues/14535

dr1rrb: Note 2: as an intermediate solution (i.e. before we fix #14535) we should patch it by converting it to ticks on GTK and FrameBuffer, at least it will be consistent with other platforms and fix some interaction issues (hold and inertia for instance).

src/Uno.UI.Runtime.Skia.Wpf/Input/WpfKeyboardInputSource.cs
```diff
@@ -70,26 +72,32 @@ private void HostOnKeyUp(object sender, System.Windows.Input.KeyEventArgs args)
...
 		}
 		catch (Exception e)
 		{
 			Windows.UI.Xaml.Application.Current.RaiseRecoverableUnhandledException(e);
 		}
 	}
 
+	// WPF doesn't expose the scancode, but it exists internally
+	private static uint GetScanCode(KeyEventArgs args)
+		=> (uint)((int)(typeof(KeyEventArgs).GetProperty("ScanCode", BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance)?.GetValue(args) ?? 0));
```
lindexi, line 96: Thank you. Is it cheaper to use MapVirtualKey?

ramezgerges: I'm guessing a system call would be more expensive (at least theoretically, it would give the OS a reason to pre-empt the process), but the main reason is that I'm not sure the 2 approaches will always return the same thing. The docs say something about left/right-handedness and default values on failure, etc., so it's easier to just get what WPF gives us. We can re-evaluate later if this turns out to be a problem I guess.

jeromelaban, line 96: Let's make it clear when the property cant 'e found. It will othewise fail with a null ref without much details.

lindexi: @jeromelaban The normal keyboard event can always find this property. The PresentationCore(WPF) will create the KeyEventArgs with the ScanCode, and the WPF will use the lParam windows message or MapVirtualKey to get the ScanCode.

And the wpf will only use the ScanCode property in `KeyDown --> Alt Numpad` business. In other words, the ScanCode property work in a less important piece of business logic code.

When the property can not be found? 1. The keyboard event from windows message, but wpf call the MapVirtualKey with Win32Exception. 2. We fake the keyboard event.

lindexi: The code to get the ScanCode (only ways) : https://github.com/dotnet/wpf/blob/187be47c5554d4745e9b17385889a8086d6e4286/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/InterOp/HwndSource.cs#L2362

The code to use ScanCode (only ways) : https://github.com/dotnet/wpf/blob/187be47c5554d4745e9b17385889a8086d6e4286/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/TextCompositionManager.cs#L435

jeromelaban: The point here that [ScanCode](https://github.com/dotnet/wpf/blob/187be47c5554d4745e9b17385889a8086d6e4286/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/KeyEventArgs.cs#L206) is an internal property that may be changing names in the future. It's a best practice, when depending on internal properties, to validate that the property still exists and if not, to provide a proper error message.

lindexi: @jeromelaban Is the ScanCode here a necessary and important attribute in the UNO framework? If so, I'd be more than happy to submit an API change request in the WPF repository.

jeromelaban: According to @ramezgerges, it is important. Regardless, since we support .NET 7, we'll need to use reflection for the time being.

lindexi: > @jeromelaban Is the ScanCode here a necessary and important attribute in the UNO framework? If so, I'd be more than happy to submit an API change request in the WPF repository.

@jeromelaban I added here: https://github.com/dotnet/wpf/issues/8473

src/Uno.UI.Runtime.Skia.Wpf/Input/WpfKeyboardInputSource.cs
```diff
@@ -70,26 +72,43 @@ private void HostOnKeyUp(object sender, System.Windows.Input.KeyEventArgs args)
...
+	private static uint GetScanCode(KeyEventArgs args)
+	{
+		try
+		{
+			return (uint)((int)(typeof(KeyEventArgs).GetProperty("ScanCode", BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance)?.GetValue(args) ?? 0));
+		}
+		catch (Exception e)
+		{
+			Console.WriteLine("Couldn't get ScanCode from WPF KeyEventArgs.");
+			Console.WriteLine(e);
```
jeromelaban, line 104: Let's use LogError reporting. Also, let's not return a default value if the property is not available, it needs to fail explicitly.

MartinZikmund (approved):

src/Uno.UI.Runtime.Skia.Wpf/Input/WpfKeyboardInputSource.cs
```diff
@@ -70,26 +72,46 @@ private void HostOnKeyUp(object sender, System.Windows.Input.KeyEventArgs args)
...
 		}
 	}
 
+	// WPF doesn't expose the scancode, but it exists internally
+	private static uint GetScanCode(KeyEventArgs args)
+	{
+		try
+		{
+			return (uint)((int)(typeof(KeyEventArgs).GetProperty("ScanCode", BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance)?.GetValue(args) ?? 0));
+		}
```
jeromelaban, line 100: I would rather see:

```suggestion
		try
		{
		   if((typeof(KeyEventArgs).GetProperty("ScanCode", BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance) is {} propertyInfo)
		   {
		      return (uint)((int)propertyInfo.GetValue(args) ?? 0));
		   }
		   else
		   {
		      if (typeof(WpfKeyboardInputSource).Log().IsEnabled(LogLevel.Error))
			{
				typeof(WpfKeyboardInputSource).Log().LogError("Unable to get the ScanCode property from WPF. This likely means this WPF version is not compatible with Uno Platform, contact the developers for more information.", e);
			}
		   }
		}
```

src/Uno.UWP/UI/Input/GestureRecognizer.cs
jeromelaban: @ramezgerges why did you need to change this particular value?

jeromelaban: @dr1rrb explained that it's a valid coming from winui.

PR #14618, opened 2023-11-29
New docs styling

mcereda95:
## PR Type

The new style for docfx is fixed after the rebase

after the rebase, the new style was broken: navbar, sidebar and some fonts were no longer working

## What is the new behavior?

the issues have been fixed and the new design matches the figma file

### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at 9258a87</samp>

This pull request updates the documentation website of Uno Platform to improve its appearance and responsiveness on different devices. It modifies the styles and scripts of various components, such as the navbar, the side navigation, and the title section. It also switches to the `dart-sass` implementation of the `sass` module for better compatibility and performance.

## PR Checklist

## Other information
NA

https://dev.azure.com/nventive/UNO%20Marketing/_workitems/edit/280519


mcereda95:
I add a new commit because I noticed during the QA that navbar was scrollable in certain window size 
image

MartinZikmund (approved):

nickrandolph:
Not enough spacing between menu items
image

MartinZikmund:
Would it make sense to merge this as first step and do the next updates one by one? To me it feels this already is a huge step forward, even though I have quite many additional comments next, but this as the first step would already make big UX improvement

MartinZikmund:
The default view is very odd on mobile:

image

One would intuitively press the hamburger menu, which is instead the menu for the whole website. 

Then there is the block with just "Overview" there - user has no idea what it is referring to.

Then only below it is a "button" that opens the ToC - which does not immediately look clickable. Compare this to Auth0 docs for example:

image

They ditch the "common header" of the website in favor of purely docs specific header, where the hamburger menu **is** the TOC.

morning4coffe-dev:
I think we should not hide the "In this article" on mobile. I like how MS docs does this, they put it on the top of the article, so the user is able to quicky see what is the article about and jump where they want to go.
image

morning4coffe-dev:
> Not enough spacing between menu items image

I second this a lot. I know it won't be easy, but we should also support something like, for example, GitHub does. When you open it on a device with a detachable keyboard, they make the menu items bigger. I think this would also improve the navigation greatly on mobile.

morning4coffe-dev:
This looks a little weird on mobile, when it's too long. I think it should collapse the first few items 
image
I have something like this in mind:
image

morning4coffe-dev:
Corner radius on the bottom, but not on the top element 😁 
image

jeromelaban:
@mergifyio backport release/stable/5.0

PR #14630, opened 2023-11-30
fix: avoid creating disconnected native commandbars on load

kazo0:
## PR Type

- Bugfix

Avoid a crash when navigating from and then back to a page with a native CommandBar on it


jeromelaban:
@mergifyio backport release/stable/5.0 release/stable/4.10 legacy/4x

rfabre:
Is it possible to have a backport of this PR in Uno 4.4 ?

PR #14693, opened 2023-12-05
perf: use CompositionTarget to only render a specific target

### LINKED ISSUE #14345: [Skia] `CompositionTarget`-based rendering

Currently we have a single `Compositor` and each visual change will invalidate all rendering surfaces (even across multiple `XamlRoots`). However, we should instead invalidate and render just a single `CompositionTarget` which matches the current visual tree.

### Why is this needed

Rendering performance.

### For which platform

_No response_

ramezgerges:
GitHub Issue (If applicable): closes #14345

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Internal/ContentRoot.cs
+++ b/src/Uno.UI/UI/Xaml/Internal/ContentRoot.cs
@@ -8,10 +8,10 @@ using System;
 using Uno.UI.Xaml.Input;
 using Uno.UI.Xaml.Islands;
 using Windows.UI;
-using Microsoft.UI.Composition;
 using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Hosting;
 using Microsoft.UI.Xaml.Input;
+using Microsoft.UI.Xaml.Media;
 using static Microsoft/* UWP don't rename */.UI.Xaml.Controls._Tracing;
 
 /*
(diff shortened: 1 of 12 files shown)


src/Uno.UI/UI/Xaml/Internal/ContentRoot.cs
```diff
@@ -68,6 +69,10 @@ public ContentRoot(ContentRootType type, Color backgroundColor, UIElement? rootE
 			FocusAdapter = new FocusAdapter(this);
 			FocusManager.SetFocusObserver(new FocusObserver(this));
 
+			CompositionTarget = new CompositionTarget();
+			CompositionTarget.Root = VisualTree.RootElement.Visual;
```
MartinZikmund, line 73: To make this more concise, I think the `Root` setter could automatically set the `CompositionTarget` as well, thus the following line's logic would happen in the setter

MartinZikmund (approved):

jeromelaban:
@MartinZikmund I missed that it was targeting your branch. Merge as needed, though it will need to be built on your target branch since the CI does not cover such merge target validations,

src/Uno.UWP/Generated/3.0.0.0/Windows.UI.Composition/CompositionTarget.cs
```diff
@@ -8,25 +8,6 @@ namespace Windows.UI.Composition
 #endif
 	public partial class CompositionTarget : global::Windows.UI.Composition.CompositionObject
 	{
-#if __ANDROID__ || __IOS__ || IS_UNIT_TESTS || __WASM__ || __SKIA__ || __NETSTD_REFERENCE__ || __MACOS__
```
MartinZikmund, line 11: Undo this change and adjust the one in `Microsoft.UI.Xaml.Media` instead. Then the implementation will need to be inside Uno.UI project, so the `Visual` in `Uno.UI.Composition` will just use plain `object` to reference the instance (as composition project does not have reference to `Uno.UI`)

PR #14697, opened 2023-12-05
fix(TextBox): Fix the issue where the text in the TextBox is not selected when dragging the mouse on FireFox

### LINKED ISSUE #14313: The text is not selected when dragging the Text in the TextBox on FireFox
### Current behavior

For the text in the TextBox,
- when double-clicking on the text, text is selected. OK
- when pressing SHIFT+LEFT or RIGHT, text is selected. OK
- When dragging mouse on the text, text isn't selected. Wrong

This issue is occurred on FireFox

[1.zip](https://github.com/unoplatform/uno/files/13273770/1.zip)

### Expected behavior

_No response_

_No response_

### Workaround

_No response_

### Works on UWP/WinUI

None

### Environment

_No response_

### NuGet package version(s)

_No response_

### Affected platforms

_No response_

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

tprr7:
… WASM

GitHub Issue (If applicable): closes #14313 

## PR Type

- Bugfix
## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Internal/PointerCapture.wasm.cs
+++ b/src/Uno.UI/UI/Xaml/Internal/PointerCapture.wasm.cs
@@ -6,6 +6,8 @@ using Windows.Devices.Input;
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Input;
 using Uno.Foundation.Logging;
+using Windows.UI.Xaml.Controls;
+using NativeMethods = __Windows.Security.ExchangeActiveSyncProvisioning.EasClientDeviceInformation.NativeMethods;
 
 namespace Uno.UI.Xaml.Core;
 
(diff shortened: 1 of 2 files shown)


src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.cs
```diff
@@ -966,7 +966,10 @@ protected override void OnPointerPressed(PointerRoutedEventArgs args)
 
 			if (ShouldFocusOnPointerPressed(args)
 				// UWP Captures pointer is not Touch
-				&& CapturePointer(args.Pointer))
+#if !__WASM__
```
dr1rrb, line 969: For me it sounds like if we where just hiding a bug here. The comment mention that the `TextBox` is capturing the pointer here, changing this behavior for all browsers is probably problematic.

I assume the issue is that as we capture the pointer, FF is no longer able to select the text. If so, I think that we should do "managed only capture" instead. (It's not supported yet, but it's only about not calling the [`CaptureNative`](https://github.com/unoplatform/uno/blob/master/src/Uno.UI/UI/Xaml/Internal/PointerCapture.wasm.cs#L14) in that particular case, and probably only on FF)

Also I think that this code has not been changed since a while and I suspect that the selection issue is a regression. It would be nice to determine what change introduced that regression to better understand how we could fix it.

DierkDroth: @dr1rrb thanks for your feedback.

Obviously it would be pretty hard - and most likely a waste of effort  - to ask an external contributor like @TopProgrammer77 to dig through the history of UNO changes in order to understand what potentially might have gone wrong in the past. 

However, it could make sense - forward thinking  - to fix the issue based one the `CaptureNative `approach you suggested. I tasked @TopProgrammer77 accordingly ...

PR #14878, opened 2023-12-19
```skia``` RichTextBlock paragraph implementation

### PR DESCRIPTION
**Pull request by limeniye**
GitHub Issue: https://github.com/unoplatform/uno/issues/81

<h3>PR Status</h3>

Still under development.

<h3>The goal of PR</h3>

Implementation of Paragraph functionality for RichTextBlock.

<h3>Changelog</h3>

[IBlock](https://github.com/unoplatform/uno/blob/5c4648510710c4500d449efcad6116679ca536fc/src/Uno.UI/UI/Xaml/Documents/TextFormatting/IBlock.skia.cs#L14) 
interface has been replaced to ISegmentsElement.
This is necessary for the following reasons:
- The IBlock interface differs from the IBlock implementation of the Block type: it has redundant properties and the absence of some that are present in the original IBlock implementation;
- IBlock performs a different functionality and its name confuses with the name of the Block type implementation;
- TextBlock's implementation of the IBlock interface makes some sense, but TextBlock is actually just a conditional block element, not an actual one, which is a bit confusing.

**ISegmentsElement**
implements the previous function of the IBlock interface and also has 2 additional methods: InvalidateElement and InvalidateSegments. These methods are analogous to the InvalidateInlines(bool updateText) method. 
But since RichTextBlock contains a collection of Bloks, not Inlines, the name of this method has been generalized for both types. 
Since the RichTextBlock type, in addition to text, also has an [InlineUIContainer](https://learn.microsoft.com/en-us/dotnet/api/system.windows.documents.inlineuicontainer?view=windowsdesktop-7.0) child element, the updateText parameter was removed, and the method was split into two methods.

ISegmentsElement type is needed in order not to write "RichTextBlock or TextBlock" condition every time.

**SegmentsCollection**
is only needed to generalize InlineCollection and BlockCollection.
The only change I added to it is to check if the element is a Paragraph for the [GetPreorderTree](https://github.com/unoplatform/uno/blob/5c4648510710c4500d449efcad6116679ca536fc/src/Uno.UI/UI/Xaml/Documents/InlineCollection.cs#L44) method.

<h3>Still under development</h3>

- Paragraph space between lines;
- Paragraph TextIndent;
- Paragraph Hyperlink support;
- Runtime tests;

<h3>Problems</h3>

Unfortunately, I did not understand how to add the default RichTextBlock style to the project.
I see it as follows:

```XML
<Style TargetType="RichTextBlock">
</Style>
```
But in the project I found only styles with x:Key specified.

<h3>Latest preview</h3>

https://github.com/unoplatform/uno/assets/69314237/eadfd0b2-2231-4b6d-9a4e-d6a5db333704

<h3>Platforms</h3>

- [ ] Windows
- [ ] Mobile
- [x] Skia
- [ ] Wasm


limeniye:
Latest preview

https://github.com/unoplatform/uno/assets/69314237/eadfd0b2-2231-4b6d-9a4e-d6a5db333704

PR #14950, opened 2024-01-02
fix: Fix ContentDialog not properly resizing

### LINKED ISSUE #14647: [Skia] Dialogue doesn't resize on Window Maximize
### Current behavior

The Dialogue doesn't resize or update it's position when the Maximize button is pressed on the Window.

image
image

### Expected behavior

The Dialogue should resize and update its position accordingly.

1. Open SampesApp
2. Open some Fluent Dialogue Window (the same issue in on non-Fluent, but may not be as noticeable)
3. Maximize the Window 

### Workaround

_No response_

### Works on UWP/WinUI

Yes

### Environment

_No response_

### NuGet package version(s)

_No response_

### Affected platforms

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

Youssef1313:
GitHub Issue (If applicable): closes #14647

## PR Type

## PR Checklist

## Other information


src/Uno.UI/UI/Xaml/Window.skia.cs
```diff
@@ -50,11 +50,10 @@ internal void OnNativeSizeChanged(Size size)
 			}
 
 			Bounds = newBounds;
+			ApplicationView.GetForCurrentView().SetVisibleBounds(newBounds);
```
Youssef1313, line 53: @MartinZikmund FYI as this will create a merge conflict with multiwindow PR.

We need to ensure that visible bounds is set properly before we raise size changed because subscribers can read this value.

MartinZikmund (approved):

src/Uno.UI/UI/Xaml/Window.skia.cs
MartinZikmund: Any chance this change can be test covered? Maybe via some test hook?

Youssef1313: @MartinZikmund I wasn't able to write a test, but added in assert in SamplesApp. Not sure how the assert will behave on Android though (maybe the sizes I'm comparing could have differences due to status bar)

ajpinedam (approved):

src/Uno.UI/UI/Xaml/Window.skia.cs
Youssef1313: In Wasm implementation, there is this comment:

https://github.com/unoplatform/uno/blob/72c0786f059068144b7f5b0c2f4dac65bdd68b86/src/Uno.UI/UI/Xaml/Window.wasm.cs#L117-L118

So it seems the change is actually wrong, which means we shouldn't rely on the right visible bounds when we are raising size changed.

Youssef1313: In WinUI, `ApplicationView` is unusable. It always throws an exception.
In UWP, SizeChanged is raised before VisibleBoundsChanged, however, the new VisibleBounds value can be correctly read from SizeChanged event.

So, the sequence is like:

- Window.Bounds and ApplicationView.VisibleBounds are both updated.
- Window.SizeChanged is raised
- ApplicationView.VisibleBoundsChanged is raised

Currently, we raise VisibleBoundsChanged as soon as we update the visible bounds. This is not correct. Since the fix will involve more code that has been changed in multiwindow, I'll wait for this PR.

src/SamplesApp/SamplesApp.Shared/App.Assertions.cs
```diff
@@ -203,4 +210,12 @@ private void AssertIssue10313ResumingAfterActivate()
 			Assert.Fail("Resuming should never be triggered unless the app is suspended.");
 		}
 	}
+
+	private static void AssertWindowSize(WindowSizeChangedEventArgs e)
+	{
+		// The idea is just to make sure the bounds are set correctly before WindowSize event is fired.
+		var applicationViewBounds = ApplicationView.GetForCurrentView().VisibleBounds;
```
MartinZikmund, line 217: Does this work on WinUI? I think the API just throws 🤔

Youssef1313: @MartinZikmund Yes, it throws on WinUI. See https://github.com/unoplatform/uno/pull/14950#discussion_r1440244473

Youssef1313: We have *internal* usages of `ApplicationView.GetForCurrentView()` which is why we care about VisibleBounds being correctly set when in SizeChanged event.

jeromelaban: Though this would fail in the winui sample, right?

Youssef1313: Yup. I wrapped the problematic call under `HAS_UNO` in the last commit.

jeromelaban:
@Youssef1313 could you fix the conflicts?

PR #14959, opened 2024-01-04
perf: Avoid LOH allocations in RenderTargetBitmap

### LINKED ISSUE #14955: [Skia] `RenderTargetBitmap` improvements
## Current situation

- `RenderTargetBitmap` is IDisposable in Uno while it's not in WinUI. We should maintain parity with WinUI.
- `RenderTargetBitmap` is using ArrayPool for renting the array. Generally, RenderTargetBitmap will tend to ask for a very large array that the .NET ArrayPool implementation will just always allocate a new array. So, effectively we are always allocating a large array and not using anything from the pool.
- The allocated array ends up in the Large Object Heap (which might be a performance issue)

Two solutions come to mind:

### Solution 1: Avoid `_buffer` field

Currently, in `RenderAsBgra8_Premul`, we get an SKBitmap and copy its pixels to `_buffer` field.

Then, we dispose the SKBitmap, and we use the `_buffer` array for two purposes:

1. Creating an SKBitmap back again, if needed (e.g, if RenderTargetBitmap is used as `Image.Source`)
2. In `GetPixelsAsync`

Both purposes can still be achieved with SKBitmap instead of a `_buffer` array.

The remaining part will be disposing the SKBitmap, which we will have to do in the destructor (unfortunately)

Note that this approach is Skia-specific. We might find similar ideas for other platforms, but then this will introduce more platform-specific code to RenderTargetBitmap.

**EDIT:** I tried this approach, but it's problematic as we need the array anyways in `GetPixelsAsync` (because `Buffer` needs either an array of bytes or `Memory<byte>`, but we have a `Span<byte>` that we can't convert to `Memory<byte>` as far as I know.

### Solution 2: Using `SegmentedArray`s

We can port `SegmentedArray` implementation from Roslyn and use it instead of a regular array. This will help with LOH allocations.

**EDIT:** This too couldn't work, as SkiaSharp needs a pointer to contiguous memory.

---

Some profiling will be needed to know the performance impact of the changes. This will be most notable in snapshot tests.

Youssef1313:
GitHub Issue (If applicable): closes #14955

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Media/Imaging/RenderTargetBitmap.cs
+++ b/src/Uno.UI/UI/Xaml/Media/Imaging/RenderTargetBitmap.cs
@@ -25,7 +25,7 @@ namespace Microsoft.UI.Xaml.Media.Imaging
 	{
 #if !__ANDROID__
 		// This is to avoid LOH array allocations
-		private unsafe struct UnmanagedArrayOfBytes : IDisposable
+		private unsafe class UnmanagedArrayOfBytes
 		{
 			public nint Pointer;
 			public int Length { get; }
@@ -34,6 +34,7 @@ namespace Microsoft.UI.Xaml.Media.Imaging
 			{
 				Length = length;
 				Pointer = Marshal.AllocHGlobal(length);
+				GC.AddMemoryPressure(length);
 			}
 
 			public byte this[int index]
@@ -48,9 +49,10 @@ namespace Microsoft.UI.Xaml.Media.Imaging
 				}
 			}
 
-			public void Dispose()
+			~UnmanagedArrayOfBytes()
 			{
 				Marshal.FreeHGlobal(Pointer);
+				GC.RemoveMemoryPressure(Length);
 			}
 		}
 
@@ -225,7 +227,7 @@ namespace Microsoft.UI.Xaml.Media.Imaging
 #if !__ANDROID__
 				unsafe
 				{
-					var mem = new UnmanagedMemoryManager<byte>((byte*)_buffer.Value.Pointer.ToPointer(), _bufferSize);
+					var mem = new UnmanagedMemoryManager<byte>((byte*)_buffer.Pointer.ToPointer(), _bufferSize);
 					return Task.FromResult<IBuffer>(new Buffer(mem.Memory.Slice(0, _bufferSize)));
 				}
 #else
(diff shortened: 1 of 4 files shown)


src/Uno.UI/UI/Xaml/Media/Imaging/RenderTargetBitmap.cs
```diff
@@ -142,36 +221,57 @@ public IAsyncOperation<IBuffer> GetPixelsAsync()
...
-		void IDisposable.Dispose()
+#if !__ANDROID__
+		~RenderTargetBitmap()
 		{
-			if (_buffer is not null)
+			if (_buffer.HasValue)
 			{
-				ArrayPool<byte>.Shared.Return(_buffer);
+				_buffer.Value.Dispose();
 			}
```
jeromelaban, line 247: Can't the buffer handle that by itself?

src/Uno.UI/UI/Xaml/Media/Imaging/RenderTargetBitmap.cs
```diff
@@ -14,14 +14,89 @@
...
+		// This is to avoid LOH array allocations
+		private unsafe struct UnmanagedArrayOfBytes : IDisposable
+		{
+			public nint Pointer;
+			public int Length { get; }
+
+			public UnmanagedArrayOfBytes(int length)
+			{
+				Length = length;
+				Pointer = Marshal.AllocHGlobal(length);
```
jeromelaban, line 36: We'll likely need to [increase memory pressure](https://learn.microsoft.com/en-us/dotnet/api/system.gc.addmemorypressure?view=net-8.0) as a result of this allocation. (same foes for the dispose).

src/Uno.UI/UI/Xaml/Media/Imaging/RenderTargetBitmap.cs
```diff
@@ -14,14 +14,89 @@
...
+			public override Span<T> GetSpan() => new Span<T>(_pointer, _length);
+
+			/// <summary>
+			/// Provides access to a pointer that represents the data (note: no actual pin occurs)
+			/// </summary>
+			public override MemoryHandle Pin(int elementIndex = 0)
+			{
+				if (elementIndex < 0 || elementIndex >= _length)
+					throw new ArgumentOutOfRangeException(nameof(elementIndex));
+				return new MemoryHandle(_pointer + elementIndex);
```
jeromelaban, line 85: Is the returned MemoryHandle disposed properly?

src/Uno.UI/UI/Xaml/Media/Imaging/RenderTargetBitmap.cs
Youssef1313: I don't see a way for that :'(

Youssef1313: Maybe switch `UnmanagedArrayOfBytes` to class instead of struct and put that code in UnmanagedArrayOfBytes's finalizer?

src/Uno.UI/UI/Xaml/Media/Imaging/RenderTargetBitmap.cs
Youssef1313: @jeromelaban It's probably caller's responsibility, e.g:

https://github.com/unoplatform/uno/blob/c500ee9d04f9e2d4fc9e959674adb246afd11fab/src/Uno.UI.Composition/Composition/SkiaCompositionSurface.skia.cs#L69

I'll try to do refactoring that will let us do the work without `MemoryManager<T>` at all

jeromelaban (approved):

PR #14968, opened 2024-01-05
chore: Enable nullable reference types in Uno.UWP

Youssef1313:

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/SourceGenerators/Uno.UI.Tasks/Assets/RetargetAssets.cs
+++ b/src/SourceGenerators/Uno.UI.Tasks/Assets/RetargetAssets.cs
@@ -32,7 +32,7 @@ namespace Uno.UI.Tasks.Assets
 
 		public string? AndroidAssetsPrefix { get; set; }
 
-		public string IosAppManifest { get; set; } = null!;
+		public string? IosAppManifest { get; set; }
 
 		[Required]
 		public string DefaultLanguage { get; set; } = null!;
@@ -172,7 +172,7 @@ namespace Uno.UI.Tasks.Assets
 			}
 
 			RetargetedAssets = retargetdAssets.ToArray();
-			PartialAppManifests = GenerateFontPartialManifest(fontAssets, IosAppManifest);
+			PartialAppManifests = GenerateFontPartialManifest(fontAssets, IosAppManifest!);
 		}
 
 
--- a/src/SourceGenerators/Uno.UI.Tasks/Assets/RetargetAssets.cs
+++ b/src/SourceGenerators/Uno.UI.Tasks/Assets/RetargetAssets.cs
@@ -25,35 +25,35 @@ namespace Uno.UI.Tasks.Assets
 		public bool UseHighDPIResources { get; set; }
 
 		[Required]
-		public string TargetPlatform { get; set; }
+		public string TargetPlatform { get; set; } = null!;
 
 		[Required]
-		public string IntermediateOutputPath { get; set; }
+		public string IntermediateOutputPath { get; set; } = null!;
 
-		public string AndroidAssetsPrefix { get; set; }
+		public string? AndroidAssetsPrefix { get; set; }
 
-		public string IosAppManifest { get; set; }
-
-		[Required]
-		public string DefaultLanguage { get; set; }
+		public string IosAppManifest { get; set; } = null!;
 
 		[Required]
-		public ITaskItem[] ContentItems { get; set; }
+		public string DefaultLanguage { get; set; } = null!;
+
+		[Required]
+		public ITaskItem[] ContentItems { get; set; } = null!;
 
 		[Output]
-		public ITaskItem[] Assets { get; set; }
+		public ITaskItem[]? Assets { get; set; }
 
 		[Output]
-		public ITaskItem[] RetargetedAssets { get; set; }
+		public ITaskItem[]? RetargetedAssets { get; set; }
 
 		[Output]
-		public ITaskItem[] PartialAppManifests { get; set; }
+		public ITaskItem[]? PartialAppManifests { get; set; }
 
 		public override bool Execute()
 		{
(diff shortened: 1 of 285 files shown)


src/Uno.UWP/Devices/Enumeration/Internal/Providers/Midi/MidiDeviceClassProviderBase.iOSmacOS.cs
```diff
@@ -79,19 +79,19 @@ public void WatchStop()
 			_client.ObjectAdded -= MidiObjectAdded;
 			_client.ObjectRemoved -= MidiObjectRemoved;
 			_client.PropertyChanged -= MidiObjectChanged;
-			_client = null;
 			_client.Dispose();
+			_client = null;
 			WatchStopped?.Invoke(this, null);
```
Youssef1313, line 84: The change here should be fixing a NullReferenceException bug.

src/Uno.UWP/ApplicationModel/Calls/PhoneCallManager.Android.cs
```diff
@@ -20,16 +20,16 @@ static PhoneCallManager()
 					"PhoneCallManager was used too early in the application lifetime. " +
 					"Android app context needs to be available.");
 			}
-			_telephonyManager = (TelephonyManager)ContextHelper.Current
-				.GetSystemService(Context.TelephonyService);
+			_telephonyManager = (TelephonyManager)ContextHelper.Current!
```
jeromelaban, line 23: This may actually be an error. ContextHelper.Current must not be null here, and it would otherwise cause a NRE if used too early.

src/Uno.UWP/Data/Xml/Dom/XmlCDataSection.cs
```diff
@@ -25,38 +25,38 @@ public string Data
 		public object Prefix
 		{
 			get => _backingDataSection.Prefix;
-			set => _backingDataSection.Prefix = value?.ToString();
+			set => _backingDataSection.Prefix = value.ToString()!;
```
jeromelaban, line 28: What happens when setting null to Prefix on WinRT?

src/Uno.UWP/Data/Xml/Dom/XmlComment.cs
```diff
@@ -25,38 +25,38 @@ public string Data
 		public object Prefix
 		{
 			get => _backingComment.Prefix;
-			set => _backingComment.Prefix = value?.ToString();
+			set => _backingComment.Prefix = value.ToString()!;
```
jeromelaban, line 28: Same here, what happens on null with WinRT?

src/Uno.UWP/Data/Xml/Dom/XmlEntityReference.cs
```diff
@@ -17,38 +17,38 @@ internal XmlEntityReference(XmlDocument owner, SystemXmlEntityReference backingE
 		public object Prefix
 		{
 			get => _backingEntityReference.Prefix;
-			set => _backingEntityReference.Prefix = value?.ToString();
+			set => _backingEntityReference.Prefix = value.ToString()!;
```
jeromelaban, line 20: Same here with null on winrt

jeromelaban (commented):
(more coming)

src/Uno.UWP/ApplicationModel/Calls/PhoneCallManager.Android.cs
Youssef1313: @jeromelaban There is no way for the compiler to know if this is used early or not.

I actually noticed that I made `ContextHelper.Current` non-nullable and added the suppression inside the property getter so the suppression is no longer needed here.

src/SourceGenerators/Uno.UI.Tasks/Assets/RetargetAssets.cs
```diff
@@ -24,35 +24,35 @@ public partial class RetargetAssets_v0 : Task
...
 
 		[Required]
-		public string IntermediateOutputPath { get; set; }
+		public string IntermediateOutputPath { get; set; } = null!;
 
-		public string AndroidAssetsPrefix { get; set; }
+		public string? AndroidAssetsPrefix { get; set; }
 
-		public string IosAppManifest { get; set; }
+		public string IosAppManifest { get; set; } = null!;
```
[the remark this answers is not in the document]
Youssef1313: There will have to be a suppression somewhere anyways. The value has to be non-null if building for iOS. I'll move the suppression elsewhere.

src/Uno.UWP/Data/Xml/Dom/XmlCDataSection.cs
Youssef1313: It seems setting to null is accepted on WinUI. Tested with the following snippet (not really a CData but I'd assume all node types will behave the same, `Prefix` is an interface member anyways and the annotation of all implementers should be the same of the interface):

```csharp
        var doc = new XmlDocument();
        doc.LoadXml(@"
<p1:Hello xmlns:p1=""https://platform.uno/prefix1"" xmlns:p2=""https://platform.uno/prefix2"">
    <Child>ChildContent</Child>
</p1:Hello>
");

        doc.FirstChild.Prefix = null;
```

I was following the same nullability annotation from `System.Xml`. But looks like System.Xml actually allows for null even though it's not annotated nullable.

I'll adjust but will have to have suppressions to account for the wrong System.Xml annotations.

Youssef1313: btw, our current behavior (relying on System.Xml) doesn't match WinUI.

On WinUI, if you set the prefix to null and try to get the value back, it returns null (same as you assigned).

In System.Xml (and hence Uno), if you set the prefix to null and try to get the value back, it returns empty string.

PR #14972, opened 2024-01-06
Miscellaneous scrolling fixes

### LINKED ISSUE #14832: [WASM] Match ScrollViewer native scrolling with managed scrolling
now that we can test the sample app more easily, I tried looking at this sample:

image

It looks like the navigation works with the keys, which is good, but it seems that the scrollbars don't follow the actual position when done through keys. 

If it's an issue that was there previously, let's open an issue to track, if not, let's try to fix as well.

_Originally posted by @jeromelaban in https://github.com/unoplatform/uno/issues/13850#issuecomment-1852856566_

See also https://github.com/unoplatform/uno/pull/13850#issuecomment-1855575551 for more details.

### LINKED ISSUE #14965: [WASM] Suggestion is not selected when pressing down arrow key in suggestion list of AutoSuggestBox
### Current behavior

This is the AutoSuggestBox issue on WASM.
When typing keyword to text box of AutoSuggestBox, SuggestionList popup will come.
Here, when I press down arrow key once, the first suggestion isn't selected. 
But then when I press up arrow key, last suggestion is selected. 
As a result, the down arrow key doesn't work in suggestion list of AutoSuggestBox

### Expected behavior

_No response_

[1.zip(updated)](https://github.com/unoplatform/uno/files/13861864/1.zip)

### Workaround

_No response_

### Works on UWP/WinUI

None

### Environment

_No response_

### NuGet package version(s)

5.1.0-dev.1219

### Affected platforms

_No response_

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

ramezgerges:
GitHub Issue (If applicable): closes #14832, closes #14965

Fixes multiple problems:
1. ScrollBar not moving as the ScrollViewer gets scrolled with keyboard on WASM (#14832)
2. ListView scrolling instead of changing selection when arrow keys are pressed (regression from #13850).
3. Keyboard scrolling on WASM scrolling incorrectly when arrow keys are held longer than needed due to discrepancy between the native ScrollContentPresenter scrolling and the managed keyboard scrolling. I've decided to turn managed keyboard scrolling off for WASM as matching the numbers precisely isn't worth the hassle.
4. On WASM, the scrollbars would freeze in place if a long scroll occurs (either violent pointer wheel scrolling or arrow key held) until the end of the scrolling duration. This is due to event processing and layoutting happening on one thread, while rendering is done on another.

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.cs
@@ -1322,6 +1322,16 @@ namespace Microsoft.UI.Xaml.Controls
 					}
 				}
 			}
+
+#if __WASM__
+			// On WASM, a large wheel scroll can be a large number of OnScroll events in sequence.
+			// In that case, the queue will be drowning with scroll events before any chance of layout
+			// updates. The ScrollContentPresenter will scroll smoothly since the native scrolling/rendering
+			// is on a separate thread, but the ScrollBars will be frozen until the end of the (long) scrolling
+			// duration.
+			_horizontalScrollbar?.Arrange(_horizontalScrollbar.LayoutSlot);
+			_verticalScrollbar?.Arrange(_verticalScrollbar.LayoutSlot);
+#endif
 		}
 
 		// Presenter to Control, i.e. OnPresenterZoomed
--- a/src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.cs
@@ -1599,7 +1599,10 @@ namespace Microsoft.UI.Xaml.Controls
 			}
 		}
 
-#if !__ANDROID__ && !__IOS__ // ScrollContentPresenter.[Horizontal|Vertical]Offset not implemented on Android and iOS
+// On WASM, we could choose to scroll in the managed layer and suppress the native scrolling
+// but it can lead to some chaotic scenarios where it's really difficult to reconcile the
+// numbers between ScrollViewer and ScrollContentPresenter, so we choose to keep the scrolling native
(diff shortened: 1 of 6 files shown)


src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.wasm.cs
```diff
@@ -94,9 +94,13 @@ private protected override void OnUnloaded()
 		{
 			base.OnUnloaded();
 			RemoveHandler(KeyDownEvent, new KeyEventHandler(OnKeyDown));
-			RemoveHandler(PointerWheelChangedEvent, new KeyEventHandler(OnKeyDown));
+			RemoveHandler(PointerWheelChangedEvent, new PointerEventHandler(OnPointerWheelChanged));
```
Youssef1313, line 97: 😨

src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.cs
```diff
@@ -1599,12 +1609,18 @@ protected override void OnPointerReleased(PointerRoutedEventArgs args)
 			}
 		}
 
-#if !__ANDROID__ && !__IOS__ // ScrollContentPresenter.[Horizontal|Vertical]Offset not implemented on Android and iOS
+		// On WASM, we could choose to scroll in the managed layer and suppress the native scrolling
+		// but it can lead to some chaotic scenarios where it's really difficult to reconcile the
+		// numbers between ScrollViewer and ScrollContentPresenter, so we choose to keep the scrolling native
+#if UNO_HAS_MANAGED_SCROLL_PRESENTER || UNO_REFERENCE_API
```
Youssef1313, line 1615: It doesn't feel right to have the override for reference but not for Wasm. Hopefully, the CI should detect this and fail.

jeromelaban (approved):

PR #15009, opened 2024-01-09
fix(FrameworkTemplatePool): Detach native view from parent when recycling

ebariche:
## PR Type

- Bugfix

## PR Checklist


src/Uno.UI/UI/Xaml/FrameworkTemplatePool.cs
```diff
@@ -324,6 +325,13 @@ private void Recycle()
 			{
 				for (var x = 0; x < count; x++)
 				{
+#if __ANDROID__
+					if (((View)array[x]).Parent is BindableView bindableView)
```
jeromelaban, line 329: Would it be best to use `array[x] is View` instead of a cast? We're assuming that the list will always be `View` instances, but may should not cost more to avoid a type exception.

jeromelaban (approved):

PR #15050, opened 2024-01-13
fix: Port GetMinMax from WinUI

Youssef1313:

## PR Type

## PR Checklist

## Other information


src/Uno.UI/UI/LayoutHelper.cs
```diff
@@ -38,36 +38,50 @@ internal static void Deconstruct(this Size size, out double width, out double he
 		[Pure]
 		internal static (Size min, Size max) GetMinMax(this IFrameworkElement e)
```
Youssef1313, line 39: NOTE: Consider the following situation:

- e.Height: NaN
- e.MinHeight: 0
- e.MaxHeight: 100

Previously, the minSize.Height would be zero.
Now, it will be `Max(Min(maxHeight, height), minHeight)` = `Max(Min(100, inf), 0)` = `Max(100, 0)` = 100

So this is a behavior change that matches WinUI. The visible effect is not yet known. Waiting for screenshot comparison results from CI.

Youssef1313: Nothing is interesting in screenshot comparison results. But I think it's a good idea to merge since this matches WinUI.

src/Uno.UI/UI/Xaml/DependencyPropertyDetails.cs
```diff
@@ -130,6 +130,8 @@ internal DependencyPropertyDetails(DependencyProperty property, Type dependencyO
 		/// <param name="precedence">The precedence level to set the value at</param>
 		internal void SetValue(object? value, DependencyPropertyValuePrecedences precedence)
 		{
+			Property.ValidateValue(value);
```
jeromelaban, line 133: That seems like a very hot path. Is it done the same way in winui?

Youssef1313: @jeromelaban It's quite similar. Every call to SetValue is going to be validated. I'd guess this call would be negligible compared to the other work done by SetValue

src/Uno.UI/UI/Xaml/DependencyProperty.cs
```diff
@@ -72,6 +72,25 @@ private DependencyProperty(string name, Type propertyType, Type ownerType, Prope
 			CachedHashCode = _name.GetHashCode() ^ ownerType.GetHashCode();
 		}
 
+		// This is our equivalent of WinUI's ValidateXXX methods in PropertySystem.cpp, e.g, CDependencyObject::ValidateFloatValue
+		internal void ValidateValue(object value)
+		{
+			if (this == FrameworkElement.MaxWidthProperty ||
+				this == FrameworkElement.MinWidthProperty ||
+				this == FrameworkElement.MaxHeightProperty ||
+				this == FrameworkElement.MinHeightProperty)
```
jeromelaban, line 81: Thinking about it again, we have `_flags` that would avoid making this particular set of checks. Let's use that instead.

jeromelaban: Let's do that one for now, until we add additional validations.

jeromelaban (approved):

ajpinedam (approved):

PR #15051, opened 2024-01-14
perf: Use a singleton ancestors dictionary

### LINKED ISSUE #13110: `AncestorsDictionary` should be pooled
`AncestorsDictionary` is instantiated and then disposed at the end of the method. Pooling can help improve performance.

On SamplesApp startup, we create about 1,544 AncestorsDictionary instances, totalling 37,056 bytes.

Youssef1313:
GitHub Issue (If applicable): closes #13110

## PR Type

## PR Checklist

## Other information


src/Uno.UI/UI/Xaml/DependencyObjectStore.cs
```diff
@@ -1587,7 +1588,7 @@ private void PropagateInheritedNonLocalProperties(DependencyObjectStore? childSt
 			// properties.
 
 			// Ancestors is a local cache to avoid walking up the tree multiple times.
-			var ancestors = new AncestorsDictionary();
+			var ancestors = _ancestorsDictionary;
```
jeromelaban, line 1591: I'm not sure this is enough. The problem here is that the hashtable may be cleared too early, causing ancestors to be cleaned in when children are processed. We may not have the appropriate amount of validation for that scenario either...

Youssef1313: The containing method turned out to be re-entrant.

So, using a singleton is a potential behavioral change that we don't have coverage for by the current tests.

However, the existing behavior might also be wrong if a parent could change after isAncestor is cached.

jeromelaban: It may be the case, but there are many manipulations that can happen when templates are applied/removed.

One thing that we could do is pool those dictionaries, as there may only be as many as the maximum depths of the current tree.

Youssef1313: Yup. Pooling makes most sense for now to at least preserve the original behavior. In future we may want to re-validate whether this code has a bug in certain scenarios.

Issue #14735, opened 2023-12-06
[Gtk] MediaPlayerElement does not load properly if the Transport controls are not enabled


jeromelaban:
### Current behavior

`When_MediaPlayerElement_Added_In_Opening` and related tests are failing if the MTC are not enabled (could be caused by the control being 0,0?)

Other tests failing:
- `When_MediaPlayerElement_Check_TransportControlVisibility`
- `When_MediaPlayerElement_SetIsFullWindow_Check_Fullscreen`
- `When_MediaPlayerElement_SetSource_Check_Play`

### Expected behavior

Tests pass.

### Works on UWP/WinUI

None

ramezgerges:
Further testing is needed to see if we can enable more tests.

@jeromelaban the video link for tests seems extremely fragile. Not sure if it's a local problem on my side or not, but maybe a more stable CDN solution would work better?

PR #15196, opened 2024-01-23
fix(mpe): show player when transport controls are disabled

ramezgerges:
GitHub Issue (If applicable): #14735

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/AddIns/Uno.UI.MediaPlayer.Skia.Gtk/GtkMediaPlayer.cs
+++ b/src/AddIns/Uno.UI.MediaPlayer.Skia.Gtk/GtkMediaPlayer.cs
@@ -307,10 +307,7 @@ public partial class GtkMediaPlayer : FrameworkElement
 				videoWidth = videoTrack.Value.Width;
 				videoHeight = videoTrack.Value.Height;
 
-				if (videoTrack.Value.SarDen != 0)
-				{
-					return true;
-				}
+				return true;
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


ramezgerges:
`When_MediaPlayerElement_SetIsFullWindow_Check_Fullscreen` seems to be failing in CI only, but the remaining tests are finally passing.

jeromelaban (approved):

ajpinedam (approved):

ramezgerges:
It seems that the first test among the newly enabled tests always fails. Initially, CI was failing because of `When_MediaPlayerElement_SetIsFullWindow_Check_Fullscreen`. I disabled it and now `When_MediaPlayerElement_SetSource_Check_Play` is not passing (it was passing in CI when `When_MediaPlayerElement_SetIsFullWindow_Check_Fullscreen` was enabled). I suspect it may have to do with caching the sample video file after the first test, so the later tests are done before  the timeout hits (although that would be weird as well since I've increased the timeouts to be 20secs instead of 8, it's just a 5mb file). I will test with a different host of the same video.

src/AddIns/Uno.UI.MediaPlayer.Skia.Gtk/GtkMediaPlayer.cs
```diff
@@ -307,10 +307,7 @@ private void GtkMediaPlayer_LayoutUpdated(object? sender, object e)
 				videoWidth = videoTrack.Value.Width;
 				videoHeight = videoTrack.Value.Height;
 
-				if (videoTrack.Value.SarDen != 0)
-				{
-					return true;
-				}
+				return true;
 			}
```
ramezgerges, line 311: @jeromelaban What do you think about this? I'm assuming you added this initially so that `SarDen` being non-zero marks the preparation of a video. This would make sense, but we don't need to wait for SarDen to start layoutting. The details (width and height) can be present even though `SarDen` is not filled out yet. I've failed to pinpoint exactly when or why `SarDen` is filled, but proceeding without `SarDen` temporarily seems to fix the tests (and also works when I've tested manually).

ramezgerges: Actually, I think the original might be right. This change makes it so that the MPE takes space for the video even though it hasn't loaded yet, filling with black until the playback starts. I will need to think about this further.

jeromelaban: I recall that checking for SarDen was about waiting for the natural size of the video to be available.

PR #15245, opened 2024-01-29
Core support for keyboard accelerators

### LINKED ISSUE #17129: Core support for Keyboard Accelerators

MartinZikmund:
GitHub Issue (If applicable): closes #17129

## PR Type

## PR Checklist

## Other information

- [x] `Get/Set` KeyboardAccelerator ToolTip (ToolTipService)


src/Uno.UI/UI/Xaml/Controls/ToolTip/ToolTipService.Properties.cs
carldebilly, line 1: I don't understand why there's ToolTip changes in a PR about keyboard accelerators.

MartinZikmund: @carldebilly they are closely tied - specifically keyboard accelerators automatically generate their own tooltip on the target controls which shows up when the control does not have one set locally

src/Uno.UI/UI/Xaml/DependencyPropertyDetailsCollection.cs
```diff
@@ -186,11 +186,11 @@ public DependencyPropertyDetails GetPropertyDetails(DependencyProperty property)
...
-				ref var propertyEntry = ref entries[offset + bucketRemainder];
-
-				if (propertyEntry == null)
+				var propertyEntry = entries[offset + bucketRemainder];
+				if (propertyEntry is null)
 				{
 					propertyEntry = new DependencyPropertyDetails(property, _ownerType, property == _dataContextProperty || property == _templatedParentProperty);
+					_entries[offset + bucketRemainder] = propertyEntry;
 
 					if (TryResolveDefaultValueFromProviders(property, out var value))
```
MartinZikmund, line 195: Had to make this change because in `TryResolveDefaultValueFromProviders` the `KeyboardAcceleratorCollection` instance got created, which then leads to it setting its `Parent` to the `UIElement` instance which owns it - this then can cause `entries` to get resized, which means the original `ref` is no longer valid and all kinds of hell breaks loose. Because the `entries` collection size only increases here, it seemed safe to do it this way, but please let me know if you see some red herrings @jeromelaban @ebariche

src/Uno.UWP/UI/Core/Internal/KeyboardStateTracker.cs
```diff
@@ -95,4 +96,14 @@ private static void SetStateOnNonSideKeys(VirtualKey key)
...
 	}
+
+	internal static void Reset()
+	{
+		// Clearing state when debugger is attached would
+		// make it hard to debug key events.
+		if (!System.Diagnostics.Debugger.IsAttached)
+		{
+			_keyStates.Clear();
+		}
```
MartinZikmund, line 107: Not sure whether to go with `Debugger.IsAttached` or just `#if DEBUG` here. My thought is that if someone is debugging an Uno app project, this will make it easier to debug key-related scenarios as opposed to `#if` approach

Youssef1313: I think `Debugger.IsAttached` is more suitable, as the debugger is really what will cause the window to be deactivated.

src/Uno.UI/UI/Xaml/DependencyPropertyDetailsCollection.cs
jeromelaban: We'll likely need to make this change in a separate PR with specialized tests.

carldebilly (approved):
LGTM

PR #15308, opened 2024-02-01
fix(textbox): caret color should follow theme

### LINKED ISSUE #15307: [Skia] Caret of `TextBox` is black in dark mode
### Current behavior

Nearly impossible to see the black caret in dark mode of Skia TextBox

### Expected behavior

Should be of light color 

_No response_

### Workaround

_No response_

### Works on UWP/WinUI

Yes

### Environment

_No response_

### NuGet package version(s)

5.1.31

### Affected platforms

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

Tested on WPF but I think it affected on all skia platforms

ramezgerges:
GitHub Issue (If applicable): closes #15307

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.skia.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.skia.cs
@@ -3230,11 +3230,12 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 			SUT.Focus(FocusState.Programmatic);
 			await WindowHelper.WaitForIdle();
 
-			var screenshot = await UITestHelper.ScreenShot(SUT);
+			var random = new Random();
 			var i = 0;
 			for (; i < 20; i++)
 			{
-				await Task.Delay(100);
+				await Task.Delay(random.Next(75, 126));
+				var screenshot = await UITestHelper.ScreenShot(SUT);
 				if (HasColorInRectangle(screenshot, new Rectangle(0, 0, screenshot.Width / 2, screenshot.Height), Colors.Black) &&
 					!HasColorInRectangle(screenshot, new Rectangle(screenshot.Width / 2, 0, screenshot.Width / 2, screenshot.Height), Colors.Black))
 				{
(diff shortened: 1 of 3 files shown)


jeromelaban:
@ramezgerges Nice! Can you add runtime test validations? Also, I would have thought that this would be following some brush from the template.

src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.skia.cs
```diff
@@ -228,6 +229,7 @@ private void UpdateTextBoxView()
 						_caretRect.Height = Math.Ceiling(rect.Height);
 						_caretRect.SetValue(Canvas.LeftProperty, rect.Left);
 						_caretRect.SetValue(Canvas.TopProperty, rect.Top);
+						_caretRect.Fill = DefaultBrushes.TextForegroundBrush;
```
MartinZikmund, line 232: Is this called in case the theme is changed too?

ramezgerges: Yes. I've tested it and it immediately worked, so I didn't think too much about it. Theoretically, changing the theme would cause a redraw (and maybe an entire layout cycle?) that would lead to CaretFound firing. I should probably invalidate the inlines just to make sure. Nice catch.

jeromelaban:
> @jeromelaban Nope. TextBox text stuff is all done internally in WinUI. There's just a `ContentElement` ScrollViewer in the template and it's basically a black box for WinUI to fill it in (we do the same on our side, except we add a `DisplayBlock` TextBlock instead of whatever WinUI is doing).

Indeed, there's no public API to change the color.

> As for tests, I thought about it but I'm not sure how one would test that. It would probably be an ImageAssert that tests a specific pixel, but these kinds of tests tend to be very buggy and/or flaky. Any better ideas?

Indeed, the caret blinks, but we can likely still make a small loop with some random delay that will still validate that the caret is present.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.skia.cs
```diff
@@ -3211,6 +3213,71 @@ public async Task When_FeatureConfiguration_Changes()
...
+			await UITestHelper.Load(SUT);
+
+			SUT.Focus(FocusState.Programmatic);
+			await WindowHelper.WaitForIdle();
+
+			var screenshot = await UITestHelper.ScreenShot(SUT);
+			var i = 0;
+			for (; i < 20; i++)
+			{
+				await Task.Delay(100);
```
jeromelaban, line 3237: Let's wait using a random delay so we don't align with the blinking unwillingly.

jeromelaban (approved):

PR #15323, opened 2024-02-01
refactor(sdk): Avoid possible double-import and fix dead condition

Youssef1313:

## PR Type

- `$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'skia'` is always false
- Import for the same targets file is duplicated with different conditions. If both conditions are true, this will be a double import.

## What is the new behavior?

- Used UnoRuntimeIdentifier instead as the condition, and did the same for Wasm
- Combined the conditions into a single import.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.Sdk/targets/Uno.Common.targets
+++ b/src/Uno.Sdk/targets/Uno.Common.targets
@@ -21,7 +21,7 @@
 	<Import Project="Uno.Common.MacOS.targets"
 		Condition=" $([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'macos' " />
 	<Import Project="Uno.Common.Skia.targets"
-		Condition="'$(UnoRuntimeIdentifier)' == 'Skia' OR $(MSBuildProjectName.Contains('.Skia.')) " />
+		Condition="'$(UnoRuntimeIdentifier)' == 'Skia' OR $([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'skia' OR $(MSBuildProjectName.Contains('.Skia.')) " />
 	<Import Project="Uno.Common.Wasm.targets"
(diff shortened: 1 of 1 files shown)


src/Uno.Sdk/targets/Uno.Common.targets
```diff
@@ -21,13 +21,10 @@
...
 		Condition=" $([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'macos' " />
 	<Import Project="Uno.Common.Skia.targets"
-		Condition=" $([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'skia' " />
-	<Import Project="Uno.Common.Skia.targets"
-		Condition=" $(MSBuildProjectName.Contains('.Skia.')) " />
-	<Import Project="Uno.Common.Wasm.targets"
-		Condition=" $([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'browser' " />
+		Condition="'$(UnoRuntimeIdentifier)' == 'Skia' OR $(MSBuildProjectName.Contains('.Skia.')) " />
 	<Import Project="Uno.Common.Wasm.targets"
-		Condition=" $(MSBuildProjectName.EndsWith('.Wasm')) or $(MSBuildProjectName.EndsWith('.WebAssembly')) " />
```
jeromelaban, line 30: Those are expected conditions that will be valid in the future. Let's keep them as-is for now.

dansiegel:
@Youssef1313 closing this as this isn't a change that is needed. Work is ongoing and the conditions are by design and intention. I realize that on the surface it appears that you could doubly import the targets however they are designed to support 2 different project systems (which is why they are intentionally separated) so it is extremely unlikely that a double import would occur.

PR #15327, opened 2024-02-02
remove redundant exception on throw

SimonCropp:
## PR Type

- Code style update (formatting)

### DIFF
--- a/src/SamplesApp/SamplesApp.Shared/App.xaml.cs
+++ b/src/SamplesApp/SamplesApp.Shared/App.xaml.cs
@@ -224,7 +224,7 @@ namespace SamplesApp
 								_ =>
 								{
 									Console.WriteLine($"WatchDog detecting a stall in the dispatcher after {timeout}, terminating the app");
-									throw new Exception($"Watchdog failed");
+									throw new($"Watchdog failed");
 								});
 						}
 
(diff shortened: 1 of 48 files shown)


src/SamplesApp/SamplesApp.Shared/App.xaml.cs
Youssef1313, line 1: Thanks for the contribution @SimonCropp!

While I personally prefer not using target-typed new for these cases, I'd like to know what @jeromelaban @MartinZikmund think.

SimonCropp: yeah i understand target typed new is a little controversial. i find the throw case to be less so. since we already have ‘throw’ so close to the context that it removes any ambiguity

MartinZikmund: I'm a bit conflicted here as I feel that maybe instead we should adjust these generic exceptions to throw a specific exception type that is more descriptive 🤔 . In many of these cases a `InvalidOperationException` would be better than plain `Exception` for example

SimonCropp: > generic exceptions to throw a specific exception type that is more descriptive 

not a fan of that. you should only throw a specific exception if there is a use case for someone to catch it. otherwise `Exception` with a clear message is a better option. when u have a clear message `InvalidOperationException ` adds no value when compared to `Exception `

jeromelaban: I don't think we should be using implicit typing here. While the compiler knows what the target is, it really feels odd.

As for `InvalidOperationException`, it is typed and makes it somehow easier to filter out of other types of underlying .NET exceptions.

PR #15328, opened 2024-02-02
use char over string

SimonCropp:
since the code paths have better perf

## PR Type


jeromelaban (approved):
Thanks for your contribution, @SimonCropp! Note that not all of Uno projects are running on .NET 7 or later (particularly the Roslyn and Visual Studio bits), so some of your changes cannot be included.

SimonCropp:
@jeromelaban thanks for the heads up. i will keep that in mind moving forward

PR #15343, opened 2024-02-03
chore: remove packages from PackageCache

SimonCropp:
as per the readme in that file

> For obvious reasons, files in this folder should not be included in source control.


jeromelaban:
Thanks for your contribution. At this time, these packages are still needed, as we're using a custom version of benchmarkdotnet. Until we've moved back to the main version of BDN, we need to keep those packages.

PR #15355, opened 2024-02-05
feat: Add support for InteractionTracker

dr1rrb:
Not intended to be merged, only to provide early feedback.

### DIFF
--- a/src/Uno.UI.Composition/Composition/Visual.skia.cs
+++ b/src/Uno.UI.Composition/Composition/Visual.skia.cs
@@ -17,6 +17,10 @@ public partial class Visual : global::Microsoft.UI.Composition.CompositionObject
 	private Vector2 _anchorPoint = Vector2.Zero; // Backing for scroll offsets
 	private int _zIndex;
 
+	// TODO: On other platforms, we could let Visual delegate to NativeOwner (UIElement) for these two properties.
+	internal bool IsTouchPadRedirected { get; set; }
+	internal bool IsPointerWheelRedirected { get; set; }
+
 	public CompositionClip? Clip
 	{
 		get => _clip;
--- a/src/Uno.UI.Composition/Composition/Visual.skia.cs
+++ b/src/Uno.UI.Composition/Composition/Visual.skia.cs
@@ -10,6 +10,7 @@ using SkiaSharp;
 using Uno.Extensions;
 using Uno.UI.Composition;
 using Uno.UI.Composition.Composition;
+using Windows.Foundation;
 
 namespace Microsoft.UI.Composition;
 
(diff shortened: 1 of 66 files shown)


src/Uno.UI.Runtime.Skia.Gtk/UI/Controls/UnoEventBox.cs
```diff
@@ -7,6 +7,23 @@ namespace Uno.UI.Runtime.Skia.Gtk;
 
 internal class UnoEventBox : EventBox
 {
+	public UnoEventBox()
+	{
+		var swipe = new GestureSwipe(this);
+		swipe.End += Swipe_End;
+		swipe.Begin += Swipe_Begin;
```
dr1rrb, line 14: Those kind of handling has proven to be difficult to maintain / make uniform on all targets, we should avoid them. Instead we should listen to raw pointer events and convert then interpret the gestures like `Swipe` (cf. `GestureRecognizer`)

src/Uno.UI.Composition/Composition/Visual.skia.cs
```diff
@@ -143,4 +154,51 @@ private protected DrawingSession BeginDrawing(SKSurface surface, in DrawingFilte
 
 		return session;
 	}
+
+	internal void RedirectTouchForPointerId(uint pointerId)
```
dr1rrb, line 158: Sounds not really related to a `Visual`, I would try to not add such methods here to avoid to end with a too massive class.

src/Uno.UI/UI/Xaml/UIElement.Pointers.cs
```diff
@@ -311,23 +311,44 @@ internal HitTestability GetHitTestVisibility()
 			var that = (UIElement)sender.Owner;
 			var src = PointerRoutedEventArgs.LastPointerEvent?.OriginalSource as UIElement ?? that;
 
+			// Note: ManipulationStarting is raised in WinUI, even if RedirectWithManipulation was called.
 			that.SafeRaiseEvent(ManipulationStartingEvent, new ManipulationStartingRoutedEventArgs(src, that, args));
+
+			if (that.Visual.RedirectedPointerId == args.Pointer.Id)
```
dr1rrb, line 317: Why is this handling is done here? I would suspect that we should do it directly in the `InputManager.PointerManager` (and maybe the `IUnoCorePointerInputSource`). I would suspect that once a pointer is being capture by the `InteractionTracker` events are no longer bubbling in the visual tree.

src/Uno.UI.Runtime.Skia.Gtk/UI/Controls/UnoEventBox.cs
Youssef1313: Not currently relying on them. Will cleanup

dr1rrb:
Also, regarding my investigations on Android, I would suspect that the `InteractionTracker` should be maintained on the "compositor thread". So even if we don't have one yet, we should make sure to not have a deep dependency with the visual tree (i.e. `UIELement` and some `Visual` properties).

Youssef1313:
Closing temporarily while I'm working locally and pushing changes to avoid unnecessary CI runs.

PR #15408, opened 2024-02-08
perf: Move Loaded/Unloaded to use OnLoaded/OnUnloaded to avoid delegate combine

jeromelaban:
Reduces allocations for ScrollViewer constructor:
image

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.cs
@@ -44,8 +44,6 @@ using View = Microsoft.UI.Xaml.UIElement;
 using _ScrollContentPresenter = Microsoft.UI.Xaml.Controls.ScrollContentPresenter;
 #else
 using _ScrollContentPresenter = Microsoft.UI.Xaml.Controls.IScrollContentPresenter;
-using System.Security;
-
 #endif
 
 #if HAS_UNO_WINUI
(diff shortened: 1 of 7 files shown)


src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.iOS.cs
```diff
@@ -38,7 +38,7 @@ public partial class ScrollViewer : ContentControl, ICustomClippingElement
 			SetScrollableContainer();
 		}
 
-		private protected override void OnLoaded()
+		private partial void OnLoadedPartial()
 		{
 			SetScrollableContainer();
 			base.OnLoaded();
```
Youssef1313, line 44: This will be a duplicate call to `base.OnLoaded`

src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.cs
```diff
@@ -44,6 +44,8 @@
 using _ScrollContentPresenter = Microsoft.UI.Xaml.Controls.ScrollContentPresenter;
 #else
 using _ScrollContentPresenter = Microsoft.UI.Xaml.Controls.IScrollContentPresenter;
+using System.Security;
+
```
Youssef1313, line 48: Minor: this seems to be added unintentionally

Youssef1313 (approved):
Very nice!

PR #15632, opened 2024-02-26
feat(skia): Support `ElementCompositionPreview.SetIsTranslationEnabled`

Youssef1313:

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI.Composition/Composition/CompositionObject.cs
+++ b/src/Uno.UI.Composition/Composition/CompositionObject.cs
@@ -22,18 +22,32 @@ namespace Microsoft.UI.Composition
 		{
 			ApiInformation.TryRaiseNotImplemented(GetType().FullName!, "The compositor constructor is not available, as the type is not implemented");
 			Compositor = new Compositor();
+			Properties = GetProperties();
 		}
 
 		internal CompositionObject(Compositor compositor)
 		{
 			Compositor = compositor;
+			Properties = GetProperties();
 		}
+
+		public CompositionPropertySet Properties { get; }
 
 		public Compositor Compositor { get; }
 
 		public CoreDispatcher Dispatcher => CoreDispatcher.Main;
 
 		public string? Comment { get; set; }
+
+		private CompositionPropertySet GetProperties()
+		{
+			if (this is CompositionPropertySet @this)
+			{
+				return @this;
+			}
+
+			return new CompositionPropertySet(Compositor);
+		}
 
 		private protected T ValidateValue<T>(object? value)
 		{
--- a/src/Uno.UI.Composition/Composition/CompositionObject.cs
+++ b/src/Uno.UI.Composition/Composition/CompositionObject.cs
@@ -16,22 +16,21 @@ namespace Microsoft.UI.Composition
 	public partial class CompositionObject : IDisposable
 	{
 		private readonly ContextStore _contextStore = new ContextStore();
+		private CompositionPropertySet? _properties;
 		private Dictionary<string, CompositionAnimation>? _animations;
 
 		internal CompositionObject()
 		{
 			ApiInformation.TryRaiseNotImplemented(GetType().FullName!, "The compositor constructor is not available, as the type is not implemented");
 			Compositor = new Compositor();
-			Properties = GetProperties();
 		}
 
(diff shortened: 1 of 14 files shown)


src/Uno.UI.Composition/Composition/CompositionObject.cs
```diff
@@ -22,19 +22,33 @@ internal CompositionObject()
 		{
 			ApiInformation.TryRaiseNotImplemented(GetType().FullName!, "The compositor constructor is not available, as the type is not implemented");
 			Compositor = new Compositor();
+			Properties = GetProperties();
 		}
 
 		internal CompositionObject(Compositor compositor)
 		{
 			Compositor = compositor;
+			Properties = GetProperties();
```
dr1rrb, line 31: Could we lazy init this ? We are creating a lot of `CompositionObject`, it will probably be a win to create it only when needed.

src/Uno.UI.Composition/Composition/CompositionObject.cs
```diff
@@ -22,19 +22,33 @@ internal CompositionObject()
...
+
 		public Compositor Compositor { get; }
 
 		public CoreDispatcher Dispatcher => CoreDispatcher.Main;
 
 		public string? Comment { get; set; }
 
+		private CompositionPropertySet GetProperties()
+		{
+			if (this is CompositionPropertySet @this)
```
dr1rrb, line 44: Is this window's behavior?

src/Uno.UI.Composition/Composition/Visual.cs
```diff
@@ -32,6 +32,18 @@ internal Visual(Compositor compositor) : base(compositor)
 
 		internal VisualInteractionSource? VisualInteractionSource { get; set; }
 
+		internal bool IsTranslationEnabled { get; set; }
+
+		internal Vector3 GetTotalOffset()
+		{
+			if (IsTranslationEnabled && Properties.TryGetVector3("Translation", out var translation) == CompositionGetValueStatus.Succeeded)
```
dr1rrb, line 39: I would suspect that the `"Translation"` should be reflected in the `TransformMatrix`? If so, you should arrange the `TransformableObjectExtensions.GetTransform` extension method instead (BTW I would probably move that method as an extensions method to keep the `Visual` class as simple as possible)

src/Uno.UI.Composition/Composition/CompositionObject.cs
Youssef1313: Seems like it, yeah.

src/Uno.UI.Composition/Composition/CompositionObject.cs
Youssef1313: Sure.

src/Uno.UI.Composition/Composition/Visual.cs
Youssef1313: @dr1rrb Nope, it looks like it's not reflected in TransformMatrix. It remains identity on WinUI after setting the translation.

src/Uno.UI.Composition/Composition/Uno/VisualExtensions.cs
```diff
@@ -0,0 +1,16 @@
+using System.Numerics;
+
+namespace Microsoft.UI.Composition;
+
+internal static class VisualExtensions
+{
+	internal static Vector3 GetTotalOffset(this Visual visual)
+	{
+		if (visual.IsTranslationEnabled && visual.Properties.TryGetVector3("Translation", out var translation) == CompositionGetValueStatus.Succeeded)
```
dr1rrb, line 9: Well in order to keep the `Properties` lazily init, we should avoid to use the `Properties` property here. A solution could be to add an internal `TryGetProperty` on the `Visual` (which almost breaks the benefit to have this as an extensions method ... your choice ^^)

dr1rrb (approved):

src/Uno.UI.Composition/Composition/Uno/VisualExtensions.cs
Youssef1313: @dr1rrb Properties is only accessed if IsTranslationEnabled is true. The default is false.

PR #15672, opened 2024-02-27
fix: DeferLoadStrategy should work even if no visibility is set

### LINKED ISSUE #15625: DeferLoadStrategy not working in Uno.UI 5.*
### Current behavior

I am heavily relying on DeferLoadStrategy="Lazy" in my XAML code to improve startup time (and to exclude platform-specific UI).
After upgrading from version 4.10.26 to 5.1.69, I realized that portions of the UI have been visible that should not have been.

So I downgraded to 5.0.156 and other versions, but with the same effect. Only the last 4.* version (being 4.10.34) works as expected. All 5.* versions load, process and show the XAML that is encapsulated within DeferLoadStrategy="Lazy" immediately.

### Expected behavior

The code within a DeferLoadStrategy="Lazy" block should not be loaded and stay invisible until it was found via "FindName".

```
<Grid DeferLoadStrategy="Lazy" x:Name="lazyGrid">
   <TextBlock>Hallo</TextBlock>
</Grid>
```

Hallo is shown, but should not (yet)

### Workaround

_No response_

### Works on UWP/WinUI

None

### Environment

### NuGet package version(s)

5.1.69 and any other verison of Uno.UI >= 5.0

### Affected platforms

Android

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

Tested under Android only.

ramezgerges:
GitHub Issue (If applicable): closes #15625

## PR Type

## PR Checklist

## Other information


src/Uno.UI/UI/Xaml/ElementStub.cs
```diff
@@ -198,17 +182,6 @@ private void Materialize(bool isVisibilityChanged)
 					_isMaterializing = true;
 
 					_content = SwapViews(oldView: (FrameworkElement)this, newViewProvider: ContentBuilder);
-					var targetDependencyObject = _content as DependencyObject;
```
jeromelaban, line 201: Changing this behavior may cause unintended consequences, as changing the visibility through a binding is supposed to be materializing the element.

ramezgerges: From my testing, it seems that WinUI doesn't care about the visibility at all. Setting the visibility in xaml to anything doesn't really affect whether the element is materialized or not. Binding the visibility also doesn't affect the materialization. The WinUI docs also don't mention anything about visibility. Our `SimpleDeferLoadStrategy` tests this already and on WinUi, the element that's supposed to be loaded when the visibility changes, in fact never loads.

jeromelaban: Got it. So it was a uno-specific behavior, which is likely not being used anymore, let's note it as a breaking change anyways, even if it's not a binary breaking change.

jeromelaban (approved):

src/Uno.UI/UI/Xaml/ElementStub.cs
queequac: Should be noted that this seems to be a breaking change only with respect to versions 5.x of Uno, because older Uno versions did also not materialize any element when changing visibility.

MartinZikmund (approved):

MartinZikmund (approved):

jeromelaban:
@ramezgerges could you fix the conflicts?

jeromelaban:
@ajpinedam could you take a look at why the tests is failing? @ramezgerges is not able to test on mac at this time.

ajpinedam:
@ramezgerges 

I tested this and validated the tests are failing because of the changes added to this file: `src/Uno.UI/UI/Xaml/Media/VisualTreeHelper.cs`

I reverted these changes locally, and the tests passed.

For the test `When_Binding_xLoad_Nested_With_ElementStub_LoadCount()`

when calling: 

```csharp
var tb01Stub = SUT.FindFirstChild<ElementStub>(e => e.Name == "tb01")!;
var tb02Stub = SUT.FindFirstChild<ElementStub>(e => e.Name == "tb02")!;
var panel01Stub = SUT.FindFirstChild<ElementStub>(e => e.Name == "panel01")!;
var panel03Stub = SUT.FindFirstChild<ElementStub>(e => e.Name == "panel03")!;
```

All these calls are returning null because 

```csharp
FindFirstChild()
  -> GetAllChildren() 
    -> VisualTreeEnumerable.Enumerate()
       -> BranchEnumerator() 
         -> VisualTreeHelper.GetChildrenCount()
```
For both `VisualTreeHelper.GetChildrenCount()` and `VisualTreeHelper.GetChildren()`, we are telling it to skip `ElementStub` objects.

```csharp
return (reference as UIElement)?
    GetChildren()
    .Where(c => c is not ElementStub) //This line here
    .ElementAtOrDefault(childIndex);
```

Btw: you will need to refresh the branch with master

cc: @jeromelaban

jeromelaban:
@ramezgerges there are still failing tests, could you take a look, thanks!

PR #15779, opened 2024-03-06
fix: copy libUnoNativeMac.dylib into Sample.Skia.Generic app

jeromelaban:
Adjusts macOS binaries location

### DIFF
--- a/src/SamplesApp/SamplesApp.Skia.Generic/SamplesApp.Skia.Generic.csproj
+++ b/src/SamplesApp/SamplesApp.Skia.Generic/SamplesApp.Skia.Generic.csproj
@@ -74,6 +74,12 @@
 	<Import Project="..\..\SourceGenerators\Uno.UI.Tasks\Content\Uno.UI.Tasks.targets" Condition="'$(SkipUnoResourceGeneration)' == '' " />
 	<Import Project="..\..\..\build\nuget\uno.winui.runtime-replace.targets" />
 
+	<Target Name="BuildUnoNativeMac" BeforeTargets="Compile" Condition="$([MSBuild]::IsOsPlatform('OSX'))">
+		<Copy
+			SourceFiles="$(MSBuildThisFileDirectory)../../Uno.UI.Runtime.Skia.MacOS/UnoNativeMac/build/$(Configuration)/libUnoNativeMac.dylib"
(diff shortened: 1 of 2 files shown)


src/SamplesApp/SamplesApp.Skia.Generic/SamplesApp.Skia.Generic.csproj
```diff
@@ -74,6 +74,12 @@
 	<Import Project="..\..\SourceGenerators\Uno.UI.Tasks\Content\Uno.UI.Tasks.targets" Condition="'$(SkipUnoResourceGeneration)' == '' " />
 	<Import Project="..\..\..\build\nuget\uno.winui.runtime-replace.targets" />
 
+	<Target Name="BuildUnoNativeMac" BeforeTargets="Compile" Condition="$([MSBuild]::IsOsPlatform('OSX'))">
```
Youssef1313, line 77: This is likely to be problematic as it may run after `_GenerateCompileDependencyCache`, so it may cause MSBuild to always think this project isn't up to date and re-build it.

I think it should be `BeforeTargets="BeforeCompile"` to avoid that.

spouliot: You're right it builds the native project each time - sadly it does so even if `BeforeCompile`.

The _good_ news is that it's quite fast so it's barely noticeable and this only apply to this sample.

Youssef1313: I didn't mean that the `Target` will execute every time, but anyway, `_GenerateCompileDependencyCache` actually won't consider the copied file at all so this is safe as-is.

But to elaborate on what I meant (which is not correct in this context), it's something like the following (in the context of a single build):

1. Project `X` is being built
    1. `_GenerateCompileDependencyCache` is run
    1. A new file is added that affects compilation
1. Another project is being built:
    1. The other project references `X`, and will also attempt to build it.
    1. Expectation is that `X` won't be built again as it was already.
    1. However, it will build again because when `_GenerateCompileDependencyCache` executes again, it will be different from the first run.

This can even be vulnerable to races when `1` and `2` are being built in parallel.

But again, the above is also to explain my original thought, which is wrong because the copied file isn't included in `_GenerateCompileDependencyCache` at all.

Youssef1313 (approved):

PR #15780, opened 2024-03-06
fix(textbox): set Handled so it does not beep on macOS

jeromelaban:

## PR Type

- Bugfix

Unhandled keyboard events (most keys) are making macOS beep.

## What is the new behavior?

* Cut (Cmd+X) and Copy (Cmd+C) beep if selection is empty (!Handled) otherwise silence
* Paste (Cmd+V) never beeps - even if the clipboard is empty
* Back (Delete) never beeps - even if there's nothing to delete
* Other keys (default) if we modify the text then it's handled

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.skia.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.skia.cs
@@ -449,14 +449,17 @@ public partial class TextBox
 				}
 				return;
 			case VirtualKey.X when ctrl:
+				args.Handled = selectionLength != 0;
 				CutSelectionToClipboard();
 				selectionLength = 0;
 				text = Text;
 				break;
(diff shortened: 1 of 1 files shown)


ramezgerges:
@jeromelaban @spouliot CI is complaining because there are runtime tests that check that Handled in fact should _not_ be set. I just tested on winui again and it indeed is never set on copy/paste/text input. Perhaps we will need to break this on MacOS specifically to prevent the beeping, or find another workaround.

MartinZikmund:
I think we should just do this top-level so that the WinUI bubbling is preserved as it is supposed to be. We could just provide a `Func<bool> FeatureConfiguration.ShouldHandleKeyDown` where people could override the default behavior not to handle certain presses.

spouliot:
I'll do a different PR so the Mac never beeps. We can revisit this later if it _offends_ Mac users (like me) too much 😄

PR #15748, opened 2024-03-05
fix: explicit Measure calls should invalidate parents if DesiredSize changes

ramezgerges:
GitHub Issue (If applicable): closes #15644

## PR Type

What kind of change does this PR introduce?

## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)
- [ ] [Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [ ] Validated PR `Screenshots Compare Test Run` results.
- [ ] Contains **NO** breaking changes
- [ ] Associated with an issue (GitHub or internal) and uses the [automatic close keywords](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue).
- [ ] Commits must be following the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) specification.

## Other information

Internal Issue (If applicable):

PR #15815, opened 2024-03-08
revert: "fix: explicit Measure calls should invalidate parents if DesiredSize changes"

ramezgerges:
#15748 broke screenshot comparisons (kudos to@Youssef1313 for the heads up). Let's revert until I can take a look at it.


Youssef1313 (approved):

ramezgerges:
Screenshot tests are actually also broken on WinUI, so it wasn't a regression due to the PR, but a bug that was uncovered with the PR. Will fix asap.

PR #15875, opened 2024-03-13
Skia hittesting with visuals

### LINKED ISSUE #15854: TransformToVisual should rely on Visual's total matrix

Currently the `UIElement.TransformToVisual`(https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.uielement.transformtovisual?view=winrt-22621) is implemented by walking up (and down) the tree of `UIElement` and computing the transform matrix based on captured layouting values, `RenderTransform`, scrolling offsets etc.

But on skia, where we properly draw an position each item by our own, we already have that exact matrix available on [`SKCanvas.TotalMatrix`](https://learn.microsoft.com/en-us/dotnet/api/skiasharp.skcanvas.totalmatrix?view=skiasharp-2.88) (which basically represents the "visual to root transform").

We should capture it on each rendering loop and use it in the `TransformToVisual` implementation.

### Why is this needed

This would make the `TransformToVisual` incredibly faster and, as those is used a lot in input dispatching, also make the inputs much faster.

### For which platform

_No response_

_No response_

ramezgerges:
GitHub Issue (If applicable): closes #15854

## PR Type

## PR Checklist

## Other information


src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_ScrollViewer.cs
```diff
@@ -1326,5 +1327,37 @@ public async Task When_ReversedMouseWheel_Then_ScrollInReversedDirection()
 			sut.VerticalOffset.Should().Be(0);
 #endif
 		}
+
+		[TestMethod]
+#if !__SKIA__
+		[Ignore("Only skia uses Visuals for TransformToVisual. The visual-less implementation adjusts the offset on the SCP itself instead of the child.")]
+#endif
```
Youssef1313, line 1334: Instead of ignoring, I'd assert against the current behavior on all platforms

src/Uno.UI/UI/Xaml/UIElement.cs
```diff
@@ -528,42 +528,67 @@ internal static Matrix3x2 GetTransform(UIElement from, UIElement to)
 				return Matrix3x2.Identity;
 			}
 
-#if UNO_REFERENCE_API // Depth is defined properly only on WASM and Skia
-			// If possible we try to navigate the tree upward so we have a greater chance
-			// to find an element in the parent hierarchy of the other element.
-			if (to is { } && from.Depth < to.Depth)
+#if __SKIA__
+			if (from.Visual?.TotalMatrix is { } fromMatrix && (to is null || to.Visual?.TotalMatrix is { }))
```
Youssef1313, line 532: I'm not really sure about this. TotalMatrix could be non-null but having an outdated value (when we're in between arrange and render)

Youssef1313: Maybe we should set `TotalMatrix` when measure or arrange is invalidated? I'm not sure though if there are any edge cases where `TotalMatrix` could change without arrange being invalidated?

src/Uno.UI.Composition/Composition/ShapeVisual.skia.cs
```diff
@@ -47,11 +49,14 @@ internal override void Render(in DrawingSession parentSession)
...
 	internal override void Draw(in DrawingSession session)
 	{
+		var canvas = session.Canvas;
 		if (ViewBox is { } viewBox)
 		{
-			session.Canvas.ClipRect(viewBox.GetRect(), antialias: true);
+			canvas.ClipRect(viewBox.GetRect(), antialias: true);
 		}
 
+		TotalMatrix = canvas.TotalMatrix;
```
Youssef1313, line 58: Is it possible that we don't read `TotalMatrix` from the canvas, and update it right away when Visual properties change?

That way, we will not even need to calculate transforms on every render frame. So, while rendering we will be just calling `canvas.SetMatrix(TotalMatrix)` where `TotalMatrix` is updated as the Visual properties are being updated. This has to be carefully done though as changes will need to happen recursively to all descendants, and it needs proper handling when a new child is being added.

src/Uno.UI/UI/Xaml/UIElement.cs
Youssef1313: > I'm waiting for your lifecycle stuff to merge to see what we can do about that.

Yeah it also *might* be possible to do something with the enhanced lifecycle by controlling when we do render ourselves. But I'm not 100% sure if that will end up being possible/viable.

src/Uno.UI.Composition/Composition/ShapeVisual.skia.cs
ramezgerges: I don't think that's very worthwhile. You'll need to change the matrix on every element while drawing anyway, so you wouldn't gain much by precalculating the matrix for each element. You'd basically be setting the matrix to a new value instead of concatenating the preexisting matrix with the current element's transformation matrix. It shouldn't matter very much I think.

Youssef1313: It matters from a code clarity/correctness point of view. The idea is to ensure that `TotalMatrix` always has the up-to-date value even if render didn't happen.

Performance-wise, it's unclear to me if it will be slower, faster, or *almost* the same. **Maybe** it won't matter that much and the difference will be too small to be noticable.

ramezgerges: I've tried this, and it's surprisingly working very well. We should not consider the `TotalMatrix` as the single source of truth of where the element visually actually is (along with the clipping).

ramezgerges:
@dr1rrb I've found a problem after the latest changes with EVP and it turns out that ScrollContentPresenter (and probably all ScrollPorts) should NOT be affected by their own viewport clipping. Meaning that when you scroll a SCP, you should see the EVP effects appeering on the children, but not on the SCP itself. In a sense, the SCP is never "clipping by itself".

src/Uno.UI.Composition/Composition/Visual.skia.cs
```diff
@@ -16,6 +14,45 @@ public partial class Visual : global::Microsoft.UI.Composition.CompositionObject
...
+				}
+
+				_totalMatrix = matrix;
+
+			}
+
+			return _totalMatrix;
+		}
+	}
+	internal string? TotalMatrixString => string.Join(", ", TotalMatrix.Values);
```
Youssef1313, line 55:
```suggestion

#if DEBUG
	internal string? TotalMatrixString => string.Join(", ", TotalMatrix.Values);
#endif
```

src/Uno.UI.Composition/Composition/ContainerVisual.skia.cs
```diff
@@ -45,15 +45,30 @@ internal void ResetRenderOrder()
 		IsChildrenRenderOrderDirty = false;
 	}
 
-	internal override void Draw(in DrawingSession session)
+	internal override void Draw(in DrawingSession session, SKMatrix initialTransform)
```
Youssef1313, line 48: I feel like we should make the parameter `in`, as it's the case with `DrawingSession`?

src/Uno.UI/UI/Xaml/Media/VisualTreeHelper.cs
```diff
@@ -444,11 +444,176 @@ internal static IReadOnlyList<_View> ClearChildren(UIElement view)
 			return default;
 		}
 
+#if __SKIA__ // optimized implementation for skia that uses visuals directly
+		/// <param name="position">The absolute position relative to the window origin</param>
+		private static (UIElement? element, Branch? stale) SearchDownForTopMostElementAt(
```
dr1rrb, line 449: Please do not clone it ... if I'm right the difference is only on how we resolve the `clippingBounds` and the `renderingBounds`, those could be extracted in a platform specific method. Creating a complete separate implementation will make it more complex to maintain and diverge.

src/Uno.UI.Composition/Composition/Visual.skia.cs
```diff
@@ -16,6 +15,49 @@ public partial class Visual : global::Microsoft.UI.Composition.CompositionObject
...
+		{
+			if (_matrixDirty)
+			{
+				_matrixDirty = false;
+
+				// Start out with the final matrix of the parent
+				var matrix = Parent?.TotalMatrix ?? SKMatrix.Identity;
+
+				// Set the position of the visual on the canvas (i.e. change coordinates system to the "XAML element" one)
+				var totalOffset = this.GetTotalOffset();
```
dr1rrb, line 40: Please use `Matrix3x2` or `Matrix4x4` as much as possible on convert it only once at the end to an `SKMatrix` (those are optimized to use CPU specific instruction set as much as possible)

src/Uno.UI.Composition/Composition/Visual.skia.cs
```diff
@@ -16,6 +15,49 @@ public partial class Visual : global::Microsoft.UI.Composition.CompositionObject
...
+	private SKMatrix _totalMatrix = SKMatrix.Identity;
+
+	/// <returns>true if wasn't dirty</returns>
+	protected internal virtual bool SetMatrixDirty() => _matrixDirty = true;
+
+	/// <summary>
+	/// This is the final transformation matrix from the origin for this Visual.
+	/// </summary>
+	[DebuggerDisplay("{TotalMatrixString}")]
+	internal SKMatrix TotalMatrix
```
dr1rrb, line 28: Why have a `SKMatrix` here ? I would use a `Matrix4x4` instead.

src/Uno.UI.Composition/Composition/Visual.skia.cs
```diff
@@ -29,7 +71,6 @@ public Vector2 AnchorPoint
 		set
 		{
 			SetProperty(ref _anchorPoint, value);
-			Compositor.InvalidateRender(this);
```
dr1rrb, line 32: Is it expected to remove this ?

src/Uno.UI/UI/Xaml/FrameworkElement.EffectiveViewport.cs
```diff
@@ -205,11 +206,6 @@ IDisposable IFrameworkElement_EffectiveViewport.RequestViewportUpdates(bool isIn
 				return;
 			}
 
-			if (!isInitial && viewport == _parentViewport)
-			{
-				return;
-			}
-
```
dr1rrb, line 212: Removing this might cause invalid viewport event to be raised.

src/Uno.UI.Composition/Composition/Visual.skia.cs
ramezgerges: It's an unnecessary call. SetProperty already calls `InvalidateRender` internally.

src/Uno.UI/UI/Xaml/Controls/TextBlock/TextVisual.skia.cs
```diff
@@ -27,7 +27,7 @@ public TextVisual(Compositor compositor, TextBlock owner) : base(compositor)
 			_owner = new WeakReference<TextBlock>(owner);
 		}
 
-		internal override void Draw(in DrawingSession session)
+		internal override void Draw(in DrawingSession session, in SKMatrix initialTransform)
```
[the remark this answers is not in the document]
ramezgerges: The `TotalMatrix` does not take into account the initial state of the canvas (e.g. scaling or offsets), only the transformations that are done by the UIElements/Visuals. So, we need to apply the "initial state of the canvas" before we apply the `TotalMatrix`.

Previously, we modified (i.e. left multiplied) the canvas matrix when we `BeginDrawing` each visual, so the effect on the root would propagate everywhere. Now, we need to reapply it whenever we set the matrix using `TotalMatrix`.

This is only relevant to `RenderTargetBitmap` and high DPI renderers. the `initialTransform` will be the identity matrix otherwise.

src/Uno.UI/UI/Xaml/FrameworkElement.EffectiveViewport.cs
ramezgerges: There is another check inside `PropagateEffectiveViewportChange`, so this one was redundant (and incorrect after the changes since SCP needs to propagate the change without itself being affected by the change)

src/Uno.UI/UI/Xaml/Controls/TextBlock/TextVisual.skia.cs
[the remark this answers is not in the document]
dr1rrb: > Now, we need to reapply it whenever we set the matrix using TotalMatrix.

Not sure to understand why? The `TotalMatrix` is "the transform of the visual relative to the root of the visual tree", if the root is already transformed (e.g. DPI) it should be applied when we start the drawing but should not impact the `TotalMatrix`?

> This is only relevant to RenderTargetBitmap and high DPI renderers. the initialTransform will be the identity matrix otherwise.

That's what I was suspecting, in that case why not pre-concat parent's `TotalMatrix.Inverse()` before drawing the content?

src/Uno.UI/UI/Xaml/FrameworkElement.EffectiveViewport.cs
dr1rrb: Ok 👍(Anyway there are some tests to validate those, just so you are aware if they fail :) )

src/Uno.UI/UI/Xaml/FrameworkElement.EffectiveViewport.cs
```diff
@@ -375,7 +375,7 @@ internal ViewportInfo GetParentViewport()
 				_effectiveViewportChanged?.Invoke(this, new EffectiveViewportChangedEventArgs(parentViewport.Effective));
 			}
 
-			if (_childrenInterestedInViewportUpdates is { Count: > 0 } && (isInitial || viewportUpdated))
+			if (_childrenInterestedInViewportUpdates is { Count: > 0 } && (isInitial || viewportUpdated || _lastScrollOffsets != ScrollOffsets))
```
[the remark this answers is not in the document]
dr1rrb: Ok makes sense 👍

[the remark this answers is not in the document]
dr1rrb: Indeed you changed the `GetEffectiveViewport` accordingly ... but only on skia. You will probably either have to change also on other platform, either have to make this new condition platform specific.

src/Uno.UI/UI/Xaml/Controls/TextBlock/TextVisual.skia.cs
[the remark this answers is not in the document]
ramezgerges: > if the root is already transformed (e.g. DPI) it should be applied when we start the drawing but should not impact the TotalMatrix?

Exactly.

When we draw a visual, we directly `SKCanvas.SetMatrix` the precalculated `TotalMatrix`, so any initial matrix set on the SKCanvas will be lost. To counteract this, we pass this initial matrix along so that when we copy (not multiply) the TotalMatrix to the canvas matrix, we take into account this initial matrix.

Alternatively, we could rely on the fact that the parent was drawn before the current visual was drawn, so instead of SetMatrix, we could just preconcatenate the transformation from `current.TotalMatrix` to `parent.TotalMatrix` (i.e. `current.TotalMatrix * parent.TotalMatrix.Inverse()`). This indeed was the initial implementation, but this fails with `RenderTargetBitmap`, where the parent might not be part of the drawing.

src/Uno.UI/UI/Xaml/FrameworkElement.EffectiveViewport.cs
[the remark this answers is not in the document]
ramezgerges: I guess this new condition will always match viewportUpdated on the other platforms? I would prefer keeping it (redundantly) instead of adding `#if __SKIA__`.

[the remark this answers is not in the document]
dr1rrb: Indeed ... a small comment would be great then :)

src/Uno.UI.Composition/Composition/Visual.skia.cs
ramezgerges: Done. I just remembered why: Matrix4x4 uses the weird column-major style that contradicts the convention everyone else uses for no good reason 🤦

src/Uno.UI/UI/Xaml/Media/VisualTreeHelper.cs
ramezgerges: It was too messy when everything was in one big function, but oddly enough, it looks quite pleasant after merging them again.

src/Uno.UI/UI/Xaml/Controls/TextBlock/TextVisual.skia.cs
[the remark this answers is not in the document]
dr1rrb: Ok get it, it's because I was still under the original idea to capture the `TotalMatrix` provided by the canvas itself during drawing, but finally it's the kind-off the opposite, we are computing it and forcefully setting it on the canvas by our own.

> we could rely on the fact that the parent was drawn before the current visual was drawn

That's how we work currently ... and if we are no longer relying on this, we can remove all the `Canvas.Save` and `Canvas.Restore`. Not sure yet what is the best approach ...

> but this fails with RenderTargetBitmap, where the parent might not be part of the drawing.

This case can be handled in the `RenderRootVisual` (which is already the case with the `ignoreLocation` flag ... where we should actually use the `Parent.TotalTransform.Inverse()` instead of only translating to the opposite offsets).



In all cases, the `initialTransform` should be moved into the `DrawingSession` and IMO should be named something like `OriginTransform` or `RootTransform` as it's only about getting the origin of the drawing plan.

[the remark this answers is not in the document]
ramezgerges: The problem with "caching" the canvas' TotalMatrix during drawing is that it will give outdated values if you modify an element and request TransformToVisual before we've had a chance to redraw.

In fact, this is exactly what the original version did. It would fall back to the old `GetTransform` implementation until we redraw, but it's cleaner to do it in reverse: calculate the TotalMatrix on the spot whenever you need it, and don't recalculate it on redrawing.

About removing the `Save`/`Restore` calls, I guess we might still need them for the clipping/filters. I'm kind of too scared of possible edge cases in some weird visuals :)

Youssef1313 (approved):
LGTM. Let's wait for @dr1rrb's approval though.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Repeater/Given_ItemsRepeater.cs
```diff
@@ -484,7 +484,7 @@ public async Task When_ItemSignificantlyHigher_Then_VirtualizeProperly()
 
 			item3UpdatedVerticalOffset.Should().Be(item3OriginalVerticalOffset); // Confirm that item #3 has not been moved down
 			var result = await UITestHelper.ScreenShot(sut.Repeater);
-			ImageAssert.HasColorAt(result, 10, 10, Colors.FromARGB("#008000")); // For safety also check it's effectively the item 3 that is visible
+			ImageAssert.HasColorAt(result, 10, 10 + (int)Math.Ceiling(sut.Scroller.VerticalOffset), Colors.FromARGB("#008000")); // For safety also check it's effectively the item 3 that is visible
```
dr1rrb, line 487: No this does not really test what is expected, if the `result` as changed due to clipping updates, you should change it to :

```suggestion
			var result = await UITestHelper.ScreenShot(sut.Root);
			ImageAssert.HasColorAt(result, 15, 15, Colors.FromARGB("#008000")); // For safety also check it's effectively the item 3 that is visible
```

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Composition/Given_CompositionClip.cs
```diff
@@ -1,366 +0,0 @@
-﻿using System;
```
dr1rrb, line 1: Why do you remove all those tests? Do they have to be udpated instead?

src/Uno.UI/UI/Xaml/UIElement.cs
```diff
@@ -528,6 +530,17 @@ internal static Matrix3x2 GetTransform(UIElement from, UIElement to)
 				return Matrix3x2.Identity;
 			}
 
+#if __SKIA__
+			Matrix4x4.Invert(to?.Visual?.TotalMatrix ?? Matrix4x4.Identity, out var invertedTotalMatrix);
```
dr1rrb, line 534: Question: how does winUI react if on of the `UIElement` is not in the visual tree?

src/Uno.UI/UI/Xaml/UIElement.cs
```diff
@@ -528,6 +530,17 @@ internal static Matrix3x2 GetTransform(UIElement from, UIElement to)
...
+			var finalTransform = (from.Visual.TotalMatrix * invertedTotalMatrix).ToMatrix3x2();
```
dr1rrb, line 535: Can the `Visual` be `null` on skia? You are allowing it to be `null` for the `to` but not the `from`.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Composition/Given_CompositionClip.cs
ramezgerges: They have been flaky for a while now. I rewrote them as UI samples to be checked by screenshot comparisons

src/Uno.UI/UI/Xaml/UIElement.cs
ramezgerges: No it can't, just a typo on my end.

src/Uno.UI.Composition/Composition/Visual.skia.cs
```diff
@@ -16,6 +15,51 @@ public partial class Visual : global::Microsoft.UI.Composition.CompositionObject
...
 	private Vector2 _anchorPoint = Vector2.Zero; // Backing for scroll offsets
 	private int _zIndex;
+	private bool _matrixDirty = true;
+	private Matrix4x4 _totalMatrix = Matrix4x4.Identity;
+
+	/// <returns>true if wasn't dirty</returns>
+	internal virtual bool SetMatrixDirty() => _matrixDirty = true;
+
+	/// <summary>
+	/// This is the final transformation matrix from the origin for this Visual.
```
agneszitte, line 25:
```suggestion
	/// This is the final transformation matrix from the origin of this Visual.
```

src/Uno.UI.Composition/Composition/Visual.skia.cs
```diff
@@ -52,7 +95,7 @@ internal int ZIndex
 	internal ShadowState? ShadowState { get; set; }
 
 	/// <summary>
-	/// Render a root visual.
+	/// Render a visual as if it's the root visual.
 	/// </summary>
 	/// <param name="surface">The surface on which this visual should be rendered.</param>
 	/// <param name="ignoreLocation">A boolean which indicates if the location of the root visual should be ignored (so it will be rendered at 0,0).</param>
```
agneszitte, line 101:
```suggestion
	/// <param name="ignoreLocation">A boolean that indicates if the location of the root visual should be ignored (so it will be rendered at 0,0).</param>
```

src/Uno.UI/UI/Xaml/FrameworkElement.EffectiveViewport.cs
```diff
@@ -289,9 +285,13 @@ private ViewportInfo GetEffectiveViewport(ViewportInfo parentViewport)
 
 				// The visible window of the SCP
 				// TODO: We should constrains the clip only on axis on which we can scroll
```
agneszitte, line 287:
```suggestion
				// TODO: We should constrain the clip to only an axis on which we can scroll
```

src/Uno.UI.Composition/Composition/ContainerVisual.skia.cs
```diff
@@ -56,4 +56,19 @@ internal override void Draw(in DrawingSession session)
 			children[i].Render(in session);
 		}
 	}
+
+	internal override bool SetMatrixDirty()
```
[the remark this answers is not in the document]
dr1rrb: We can defer this to another PR

src/Uno.UI/UI/Xaml/Media/VisualTreeHelper.cs
```diff
@@ -444,11 +444,15 @@ internal static IReadOnlyList<_View> ClearChildren(UIElement view)
 			return default;
 		}
 
+		/// <param name="position">
+		/// On skia: The absolute position relative to the window origin.
+		/// Everywhere else: The position relative to the parent (i.e. the position in parent coordinates).
+		/// </param>
 		private static (UIElement? element, Branch? stale) SearchDownForTopMostElementAt(
-			Point posRelToParent,
+			Point position,
```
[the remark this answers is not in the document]
ramezgerges: I fixed the tracing strings and forgot to fix the comments. The comment for `position` is already added as an xml doc.

src/Uno.UI/UI/Xaml/UIElement.cs
ramezgerges: It returns a `MatrixTransform` with `Matrix == Identity`.

src/Uno.UI.RuntimeTests/Tests/Microsoft_UI_Xaml_Controls/Given_RatingControl.cs
```diff
@@ -41,7 +41,8 @@ public async Task When_Loaded_Then_Unloaded_Tap()
 
 		// Tap RatingControl
 		ratingControl.Value = 1;
-		var tapTarget = ratingControl.TransformToVisual(null).TransformPoint(new Point(ratingControl.ActualWidth * 0.9, ratingControl.ActualHeight / 2));
+		// We don't use ActualWidth because of https://github.com/unoplatform/uno/issues/15982
+		var tapTarget = ratingControl.TransformToVisual(null).TransformPoint(new Point(112 * 0.9, ratingControl.ActualHeight / 2));
```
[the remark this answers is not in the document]
ramezgerges: It does. AFAIK, DPI generally shouldn't change the "logical" numbers, it just scales the drawing canvas by the scaling ratio

PR #16008, opened 2024-03-22
fix: Adjust global usings suppression for HR

jeromelaban:


src/Uno.Sdk/targets/Uno.Common.Wasm.targets
```diff
@@ -41,6 +41,39 @@
...
+		-->
+
+		<PropertyGroup>
+			<_FakeGlobalUsingsFile>$(IntermediateOutputPath)/FakeGlobalUsings.g.cs</_FakeGlobalUsingsFile>
+			<_FakeGlobalUsingsAssemblyName>$(AssemblyName.Replace('.','_'))</_FakeGlobalUsingsAssemblyName>
+		</PropertyGroup>
+
+		<ItemGroup>
+			<_UsingsDedup Include="@(Using)" Name="%(Identity)" KeepDuplicates="false" />
+			<_GlobalUsingMocksLines Include="@(_UsingsDedup->'namespace %(Name) { class __PlaceHolderDoNotUse_$(_FakeGlobalUsingsAssemblyName) { } }')" />
```
dansiegel, line 62: Since this is a work around for Microsoft.AspNetCore.* can we filter on this for the usings?

jeromelaban: It's a bit larger than that, as it also includes Microsoft.Extensions

dansiegel (approved):

PR #16104, opened 2024-04-03
[SDK] Maui Embedding fixes

dansiegel:

- closes unoplatform/uno-private#417
- closes unoplatform/uno-private#418

## PR Type

- Bugfix

- Automatic package lookup incorrectly selects version when you go from 1.1.9 to 1.1.10
- AndroidX Collections package not found with 1.4.0.2
- UnoFeatures not correctly set with props file

## What is the new behavior?

- NuGet Versions are now correctly sorted so that 1.1.10 would be selected as the more recent version over 1.1.9
- AndroidX Collections group now uses 1.4.0.1 which all packages in the group have a version for
- UnoFeatures has moved to a targets file to be properly set. This was not picking up the Features in the csproj


dansiegel:

jeromelaban (approved):

PR #16153, opened 2024-04-05
Enable warnings as errors

dansiegel:

- n/a

## PR Type

- Project automation

Build Warnings are treated normally

## What is the new behavior?

Build Warnings in the 5.1/5.2 templates are now set to treat Warnings as Errors to help prevent regressions that result in build warnings.

### DIFF
--- a/src/SolutionTemplate/5.2/uno52blank/Directory.Build.props
+++ b/src/SolutionTemplate/5.2/uno52blank/Directory.Build.props
@@ -5,7 +5,7 @@
     <Nullable>enable</Nullable>
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
     <UnoSdkDebugging>true</UnoSdkDebugging>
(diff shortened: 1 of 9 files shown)


ajpinedam (approved):

src/SolutionTemplate/5.2/uno52blank/Directory.Build.props
```diff
@@ -5,6 +5,7 @@
     <Nullable>enable</Nullable>
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
     <UnoSdkDebugging>true</UnoSdkDebugging>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
Youssef1313, line 8: Should we add this [here](https://github.com/unoplatform/uno/blob/8d386cad726054a15f53c8803941c5c8b77caed8/src/SolutionTemplate/Directory.Build.props) instead?

PR #16156, opened 2024-04-06
Clip improvements

Youssef1313:

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI.Composition/Composition/CompositionClip.skia.cs
+++ b/src/Uno.UI.Composition/Composition/CompositionClip.skia.cs
@@ -17,7 +17,7 @@ partial class CompositionClip
 	{
 		if (GetBoundsCore(visual) is { } bounds)
 		{
-			return bounds.Transform(this.TransformMatrix);
+			return TransformMatrix.Transform(bounds);
 		}
 
 		return null;
--- a/src/Uno.UI.Composition/Composition/CompositionClip.skia.cs
+++ b/src/Uno.UI.Composition/Composition/CompositionClip.skia.cs
@@ -8,58 +8,13 @@ namespace Microsoft.UI.Composition;
 
 partial class CompositionClip
 {
-	internal Rect? GetRect(Visual visual)
-	{
-		if (this is InsetClip insetClip)
-		{
-			return new Rect(
-				x: insetClip.LeftInset,
-				y: insetClip.TopInset,
-				width: visual.Size.X - insetClip.LeftInset - insetClip.RightInset,
-				height: visual.Size.Y - insetClip.TopInset - insetClip.BottomInset);
-		}
-		else if (this is RectangleClip rectangleClip)
-		{
-			return new Rect(
-				x: rectangleClip.Left,
-				y: rectangleClip.Top,
-				width: rectangleClip.Right - rectangleClip.Left,
-				height: rectangleClip.Bottom - rectangleClip.Top);
-		}
+	/// <summary>
+	/// Returns the bounds of the clip. The clip itself could be non-rectangular, e.g, rounded rectangle or path.
+	/// </summary>
+	internal virtual Rect? GetBounds(Visual visual)
+		=> null;
 
-		return null;
-	}
-
-	internal void Apply(SKCanvas canvas, Visual visual)
+	internal virtual void Apply(SKCanvas canvas, Visual visual)
 	{
-		if (this is InsetClip insetClip)
-		{
-			var rect = new SKRect(
-				insetClip.LeftInset,
-				insetClip.TopInset,
-				visual.Size.X - insetClip.RightInset,
-				visual.Size.Y - insetClip.BottomInset);
-			canvas.ClipRect(rect, SKClipOperation.Intersect, true);
-		}
-		else if (this is RectangleClip rectangleClip)
-		{
-			canvas.ClipRoundRect(rectangleClip.SKRoundRect, SKClipOperation.Intersect, true);
-		}
-		else if (this is CompositionGeometricClip geometricClip)
-		{
-			switch (geometricClip.Geometry)
-			{
-				case CompositionPathGeometry { Path.GeometrySource: SkiaGeometrySource2D geometrySource }:
-					canvas.ClipPath(geometrySource.Geometry, antialias: true);
-					break;
-				case CompositionPathGeometry cpg:
-					throw new InvalidOperationException($"Clipping with source {cpg.Path?.GeometrySource} is not supported");
-				case null:
-					// null is nop
-					break;
-				default:
-					throw new InvalidOperationException($"Clipping with {geometricClip.Geometry} is not supported");
-			}
-		}
 	}
 }
--- a/src/Uno.UI.Composition/Composition/CompositionClip.skia.cs
+++ b/src/Uno.UI.Composition/Composition/CompositionClip.skia.cs
@@ -2,6 +2,7 @@
 using System;
 using System.Linq;
 using SkiaSharp;
+using Uno.Extensions;
 using Windows.Foundation;
 
 namespace Microsoft.UI.Composition;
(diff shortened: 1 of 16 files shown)


src/Uno.UI/UI/Xaml/Media/VisualTreeHelper.cs
```diff
@@ -487,19 +487,23 @@ internal static IReadOnlyList<_View> ClearChildren(UIElement view)
 				TRACE($"- renderTransform: {tr.ToMatrix(element.RenderTransformOrigin, element.ActualSize.ToSize())}");
 
 #if __SKIA__
-			var transformToElement = element.TransformToVisual(null);
+			var transformToElement = UIElement.GetTransform(element, null);
```
Youssef1313, line 490: `TransformToVisual` will allocate `MatrixTransform` unnecessarily, which is simply just a wrapper around `Matrix` (being a DP).

Using `GetTransform` to return `Matrix` right away is more performant in two ways:

1. Avoids the `MatrixTransform` allocation
2. Avoids calling into the DP system when setting `MatrixTransform.Matrix` dependency property, which then calls into a property changed callback

src/Uno.UI/UI/Xaml/UIElement.cs
```diff
@@ -844,7 +844,30 @@ private static void InnerUpdateLayout(UIElement root)
...
+			{
+				ApplyNativeClip(Rect.Empty);
+				OnViewportUpdated();
+			}
+			else
+			{
+				var rect = clip.Rect;
+				if (clip.Transform is { } clipTransform)
+				{
+					rect = clipTransform.TransformBounds(rect);
```
dr1rrb, line 864: On skia, we should instead map this to the `CompositionClip.TransformMatrix`.

src/Uno.UI/UI/Xaml/Media/VisualTreeHelper.cs
```diff
@@ -487,19 +487,23 @@ internal static IReadOnlyList<_View> ClearChildren(UIElement view)
...
 			// The maximum region where the current element and its children might draw themselves
 			// This is expressed in the window (absolute) coordinate space.
 			var clippingBounds = element.Visual.ViewBox?.GetRect() is { } rect
-				? transformToElement.TransformBounds(new Rect(rect.Left, rect.Top, rect.Width, rect.Height))
+				? transformToElement.Transform(rect)
 				: Rect.Infinite;
+			if (element.Visual.Clip?.GetRect(element.Visual) is { } clip)
+			{
+				clippingBounds = clippingBounds.IntersectWith(transformToElement.Transform(clip)) ?? default;
+			}
```
dr1rrb, line 500: Note: This still does not supports non rectangular clip (like rounded rect). It's sufficient for now, but we will need at some point to reverse the logic to do something like `Visual.HitTest(Point)` (which could be recursive by its own).

src/Uno.UI.Composition/Composition/CompositionViewBox.skia.cs
```diff
@@ -2,11 +2,23 @@
...
+			top: Offset.Y,
+			right: Offset.X + Size.X,
+			bottom: Offset.Y + Size.Y);
+
+	internal Rect GetRect()
+		=> new(
+			x: Offset.X,
+			y: Offset.Y,
+			width: Size.X,
+			height: Size.Y);
```
dr1rrb, line 23: This should be in `CompositionViewBox.cs` file

src/Uno.UI/UI/Xaml/UIElement.cs
```diff
@@ -844,7 +844,30 @@ private static void InnerUpdateLayout(UIElement root)
 
 		internal void ApplyClip()
 		{
-#if __CROSSRUNTIME__
+#if __SKIA__
+			// On Skia specifically, we separate the two types of clipping.
+			// First, from Clip DP (handled in this code path)
+			// That clipping propagates to Visual.Clip through ApplyNativeClip.
+			// Second is clipping calculated from FrameworkElement.GetClipRect during arrange.
+			// That clipping propagates to ViewBox during arrange.
```
dr1rrb, line 852: This comment adds points to the idea to move corner radius directly in the `CompositionViewBox`

src/Uno.UI.Composition/Composition/Visual.skia.cs
```diff
@@ -205,6 +212,9 @@ private DrawingSession BeginDrawing(SKSurface surface, SKCanvas canvas, in Drawi
 		// Note: The Clip is applied after the transformation matrix, so it's also transformed.
 		Clip?.Apply(canvas, this);
 
+		// CornerRadiusClip applies to the children only.
+		CornerRadiusClip?.Apply(canvas, this);
```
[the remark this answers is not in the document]
dr1rrb: Let keep that for later : https://github.com/unoplatform/uno/issues/16294

src/Uno.UI.Composition/Composition/Visual.skia.cs
```diff
@@ -77,6 +78,12 @@ internal Matrix4x4 TotalMatrix
 		set => SetProperty(ref _clip, value);
 	}
 
+	internal RectangleClip? CornerRadiusClip
```
dr1rrb, line 81:
```suggestion
	/// <summary>
	/// DO NOT USE: This a a temporary property to properly support the corner radius. 
	/// It should be removed by https://github.com/unoplatform/uno/issues/16294.
	/// This clipping should be applied only on Children elements (i.e. in the context of a ContainerVisual)
	/// </summary>
	internal RectangleClip? CornerRadiusClip
```

src/Uno.UI/UI/Xaml/UIElement.cs
Youssef1313: This is the `UIElement.Clip` property. It's not a composition clip. Or am I misunderstanding something?

src/Uno.UI/UI/Xaml/Media/VisualTreeHelper.cs
Youssef1313: Better support for hit testing in such scenarios will be brought by https://github.com/unoplatform/uno/pull/16201. It's still not a `Visual.HitTest` though.

src/Uno.UI/UI/Xaml/UIElement.cs
dr1rrb: On ski the `ApplyNativeClip` will actually only set the `Visual.Clip` property which is of type `CompositionClip` and itself which has a `TransformMatrix` property. SO the idea is that instead of transforming the `rect` here, I would just push the transform to the `CompositionClip.TransformMatrix`. This will later allow us to animate that transform "independently".

dr1rrb (approved):

MartinZikmund (approved):

src/Uno.UI/UI/Xaml/UIElement.skia.cs
```diff
@@ -354,12 +354,19 @@ internal virtual void OnArrangeVisual(Rect rect, Rect? clip)
...
 
-				Visual.Clip = Visual.Compositor.CreateRectangleClip(
+				var compositionClip = Visual.Compositor.CreateRectangleClip(
 					top: (float)roundedRectClip.Top,
 					left: (float)roundedRectClip.Left,
 					bottom: (float)roundedRectClip.Bottom,
 					right: (float)roundedRectClip.Right
 				);
+
+				if (Clip.Transform is { } clipTransform)
```
dr1rrb, line 364: Should we add the transformation as a parameter of the method on skia instead? It sounds weird to me that the rect is read by caller but transform is read independently here.

src/Uno.UI.Composition/Composition/CompositionClip.skia.cs
```diff
@@ -10,8 +11,23 @@ partial class CompositionClip
...
 	/// <summary>
 	/// Returns the bounds of the clip. The clip itself could be non-rectangular, e.g, rounded rectangle or path.
+	/// Note that this already handles TransformMatrix
 	/// </summary>
-	internal virtual Rect? GetBounds(Visual visual)
+	internal Rect? GetBounds(Visual visual)
+	{
+		if (GetBoundsCore(visual) is { } bounds)
+		{
+			return bounds.Transform(this.TransformMatrix);
```
dr1rrb, line 20: I would do the opposite actually, transformation are computed using matrix, the `rect.Transform` is an extensions method kept for backward compatibility:

``` suggestion
			return TransformMatrix.Transform(rect);
```

dr1rrb (approved):

PR #16201, opened 2024-04-10
Skia shapes hittesting

### LINKED ISSUE #15855: `HitTest` should rely on `Visual`

The hit-testing algorithm should rely on `Visual` instead of `UIElement`'s position. Basically on Windows, the hist tests is all about "does the element as draw something non transparent on the given pixel".

### Why is this needed

* It would improve performance by allowing us to properly hit-test on the compositor thread (when we will finally have it - https://github.com/unoplatform/uno/issues/15854  might be needed first) in sync with the frame currently visible, dispatch to `InteractionTracker` if relevant or pushing it to the `UIElement` on UI thread.
* Fix issue with `Shape` hit-testing : a `Shape` should be hit-testable only on drawn pixels, not on the whole "drawing area/rectangle" like it is currently

### For which platform

_No response_

_No response_

ramezgerges:
GitHub Issue (If applicable): closes #15855, #17675

## PR Type

## PR Checklist

## Other information


ramezgerges:
this builds on top of #15875

src/Uno.UI.Composition/Composition/CompositionSpriteShape.skia.cs
```diff
@@ -13,8 +14,15 @@ public partial class CompositionSpriteShape : CompositionShape
 		private SKPaint? _strokePaint;
 		private SKPaint? _fillPaint;
 
+		internal SKPath? LastDrawnStrokePath { get; private set; }
+		internal SKPath? LastDrawnFillPath { get; private set; }
```
Youssef1313, line 18: Not sure if there is an unmanaged memory leak here.

ramezgerges: Yeah, I noticed that too. According to the documentation, the unmanaged memory will be freed in the managed object's destructor.

ramezgerges: Disposing sometimes throws weird native errors. It should be safe to not dispose. The remark on `Dispose` says this:
>Always dispose the object before you release your last reference to the <see cref="T:SkiaSharp.SKNativeObject" />. Otherwise, the resources it is using will not be freed until the garbage collector calls the finalizer.

image

Youssef1313: Not disposing and relying on the GC may have performance impacts, I think. Finalizers aren't cheap, and an explicit Dispose call should be calling `GC.SuppressFinalize` (inside SkiaSharp)

What errors did you get when disposing?

src/Uno.UI.Composition/Composition/CompositionSpriteShape.skia.cs
```diff
@@ -76,10 +85,11 @@ internal override void Draw(in DrawingSession session)
 					// So, to get a correct stroke geometry, we must apply the transformations first.
 
 					// Get the stroke geometry, after scaling has been applied.
-					using var strokeGeometry = strokePaint.GetFillPath(geometryWithTransformations);
+					var strokeGeometry = strokePaint.GetFillPath(geometryWithTransformations);
 
 					stroke.UpdatePaint(fillPaint, strokeGeometry.Bounds);
 
+					LastDrawnStrokePath = strokeGeometry;
```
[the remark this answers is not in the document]
ramezgerges: No longer needed with the latest changes.

src/Uno.UI.Composition/Composition/BorderVisual.skia.cs
```diff
@@ -81,8 +84,11 @@ private protected override void OnPropertyChangedCore(string? propertyName, bool
 				_borderPathValid = false; // to update _borderPath if previously skipped
 				if (BorderBrush is not null && _borderShape is null)
 				{
-					// we need this to track get notified on brush updates.
-					SetProperty(ref _borderShape, Compositor.CreateSpriteShape());
+					// we need this to get notified on brush updates
+					// (BorderBrush internals change -> BorderShape is notified through FillBrush -> render invalidation)
+					BorderShape = Compositor.CreateSpriteShape();
+
+					_borderShape!.Geometry = Compositor.CreatePathGeometry();
```
dr1rrb, line 91: This will probably led to 2 invalidation, it's usually better to fully configure the object before assigning it rather than assign it and then update its properties.

src/Uno.UI.Composition/Composition/BorderVisual.skia.cs
```diff
@@ -254,8 +252,8 @@ private void UpdatePathsAndCornerClip()
 
 	private unsafe void UpdateBackgroundPath(bool useInnerBorderBoundsAsAreaForBackground, SKSize innerArea, SKSize outerArea, SKPoint* outerRadii, SKPoint* innerRadii)
 	{
-		_backgroundPath ??= new SKPath();
-		_backgroundPath.Reset();
+		var backgroundPath = new SKPath();
+		backgroundPath.Reset();
```
dr1rrb, line 256: Sounds useless after a `new`

src/Uno.UI.Composition/Composition/CompositionSpriteShape.skia.cs
```diff
@@ -141,9 +131,63 @@ private static SKPaint TryCreateAndClearPaint(in Visual.PaintingSession session,
 				}
 			}
 
-			paint.ColorFilter = session.Filters.OpacityColorFilter;
```
dr1rrb, line 144: I get why you want to remove it ... but it has the advantage to make sure that we don t forget to do it. If there is no other reason than code clarity, I think I prefer the previous version.

src/Uno.UI.Composition/Composition/CompositionSpriteShape.skia.cs
```diff
@@ -141,9 +131,63 @@ private static SKPaint TryCreateAndClearPaint(in Visual.PaintingSession session,
...
+						var transform = CombinedTransformMatrix;
+						SKPath geometryWithTransformations;
+						if (transform.IsIdentity)
+						{
+							geometryWithTransformations = geometry;
+						}
+						else
+						{
+							geometryWithTransformations = new SKPath();
+							geometry.Transform(transform.ToSKMatrix(), geometryWithTransformations);
```
dr1rrb, line 155: The signature of this method is weird ... it is something new? If so, either make the `geometryWithTransformations` a `ref` parameter, either make the method to return a new instance.

src/Uno.UI.Composition/Composition/CompositionSpriteShape.skia.cs
```diff
@@ -141,9 +131,63 @@ private static SKPaint TryCreateAndClearPaint(in Visual.PaintingSession session,
...
+					{
+						_geometryWithTransformations = null;
+					}
+					break;
+			}
+		}
+
+		internal override bool HitTest(Point point)
+		{
+			point = CombinedTransformMatrix.Inverse().Transform(point);
```
dr1rrb, line 170: This should be in the condition below

src/Uno.UI.Composition/Composition/ShapeVisual.skia.cs
```diff
@@ -54,4 +55,27 @@ internal override void Paint(in PaintingSession session)
 
 		return shape;
 	}
+
+	/// <remarks>This does NOT take the clipping into account.</remarks>
```
dr1rrb, line 59: Is it applied somewhere else? If not this would introduce regression on skia I think

src/Uno.UI.Composition/Composition/SkiaGeometrySource2D.skia.cs
```diff
@@ -8,16 +8,15 @@ namespace Microsoft.UI.Composition
...
 		{
 			Geometry = source ?? throw new ArgumentNullException(nameof(source));
 		}
 
-		public SKPath Geometry { get; set; }
+		/// <remarks>
+		/// DO NOT MODIFY THIS SKPath. CREATE A NEW SkiaGeometrySource2D INSTEAD.
+		/// This can lead to nasty invalidation bugs where the SKPath changes without notifying anyone.
+		/// </remarks>
+		public SKPath Geometry { get; }
```
dr1rrb, line 20: Should we refactor this class to keep this path private ?

src/Uno.UI.Composition/Composition/Visual.skia.cs
```diff
@@ -123,6 +123,33 @@ internal Matrix4x4 TotalMatrix
 			}
 
 			return _totalMatrix;
+
+			Matrix4x4 GetTransform()
```
dr1rrb, line 127: Should we have a `CombinedMatrix` also here ?

src/Uno.UI/UI/Xaml/Controls/ScrollContentPresenter/ScrollContentPresenter.cs
```diff
@@ -235,6 +235,30 @@ internal override bool IsViewHit()
 			=> new Rect(default, RenderSize);
 #endif
 
+#if UNO_HAS_MANAGED_SCROLL_PRESENTER
+		internal void HookScrollEvents(ScrollViewer sv)
+		{
+			// Mouse wheel support
+			sv.PointerWheelChanged += PointerWheelScroll;
+
+			// Touch scroll support
```
dr1rrb, line 244:
```suggestion
			// Touch scroll support
			// Note: Events are hooked on the SCP itself, not the ScrollViewer
```

src/Uno.UI/UI/Xaml/Controls/ScrollContentPresenter/ScrollContentPresenter.cs
```diff
@@ -235,6 +235,30 @@ internal override bool IsViewHit()
 			=> new Rect(default, RenderSize);
 #endif
 
+#if UNO_HAS_MANAGED_SCROLL_PRESENTER
```
dr1rrb, line 238: Why are you moving this in the shared file, while we already have a the `SCP.Managed.cs` file for this? If needed, we can rename the file `SCP.managed_pt.cs`

src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.cs
```diff
@@ -1012,10 +1012,23 @@ protected override void OnApplyTemplate()
...
 			{
 				presenter.ScrollOwner = this;
 				Presenter = presenter;
+
+				// Note: the way WinUI does scrolling is very different, and doesn't use
+				// PointerWheelChanged changes, etc.
+				// We can either subscribe on the ScrollViewer or the SCP directly, but due to
+				// the way hit-testing works (see #16201), the SCP will not receive any pointer
+				// events. On WinUI, this is also the case: pointer presses are received on the SV,
+				// not on the SCP.
```
dr1rrb, line 1028: Even if the SV and SCP are weird controls, this is suspicious ... we are not receiving any pointer events on the SCP on WinUI? It is a real feature or a bug?

src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.Managed.cs
```diff
@@ -24,7 +24,11 @@ private bool ChangeViewNative(double? horizontalOffset, double? verticalOffset,
 			=> (_presenter as ScrollContentPresenter)?.Set(horizontalOffset, verticalOffset, disableAnimation: disableAnimation) ?? true;
 
 		private partial void OnLoadedPartial() { }
-		private partial void OnUnloadedPartial() { }
+
+		private partial void OnUnloadedPartial()
+		{
+			Presenter?.UnhookScrollEvents(this);
```
dr1rrb, line 30: We `UnHook` on un-load, but we are not re-subscribing on next load. Also as we already have a ref to the `Owner`, I think it would be better to just keep everything in the SCP and just subscribe to the load/unload events from it

src/Uno.UI/UI/Xaml/Media/VisualTreeHelper.cs
```diff
@@ -659,9 +659,14 @@ internal static void ClearChildren(UIElement view)
 				}
 			}
 
+#if __SKIA__
+			// We still need to check the renderingBounds in case the element is clipped.
```
dr1rrb, line 663: Is this a TODO before we merge the PR ?

src/Uno.UI.Composition/Composition/CompositionSpriteShape.skia.cs
ramezgerges: I moved it out of the function because I need to use the function (TryCreateAndClearPaint) in the hit-testing logic (that runs without a `session`). I guess I can create an additional function that redirects to this one + adds this line back so we can keep both versions.

src/Uno.UI.Composition/Composition/CompositionSpriteShape.skia.cs
ramezgerges: Not new. It's from SkiaSharp, not something we added. I agree, it looks really weird, but there is no overload that returns a new instance. There is an overload that takes 1 paramter (the matrix) that makes the modification in place, so I can't even add an extension method that returns a new instance.

src/Uno.UI.Composition/Composition/CompositionSpriteShape.skia.cs
ramezgerges: I adjust the signature to require a (possibly null) `colorFilter` parameter.

src/Uno.UI.Composition/Composition/ShapeVisual.skia.cs
ramezgerges: Yes. the way hit-testing works in this PR is that a visual hit-tests positively checks the clip area as before AND also HitTest returns true. This is why we don't need to take clipping into account here.

src/Uno.UI.Composition/Composition/Visual.skia.cs
ramezgerges: Right now, probably not. I added a new `CombinedTransformMatrix` property in CompositionShape because I needed to listen to changes to it. If we later need something similar here (e.g. so that we know exactly when to recalculate the TotalMatrix, instead of the current naive MatrixDirty logic), we can go ahead and introduce a similar CombinedMatrix property. Maybe in another PR.

src/Uno.UI/UI/Xaml/Controls/ScrollContentPresenter/ScrollContentPresenter.cs
ramezgerges: No idea :). Moved to the `Managed` partial.

src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.cs
ramezgerges: It is very suspicious indeed. Yes, we do NOT receive any pointer events on the SCP on WinUI: we don't get WheelChanged at all, and we get Pressed and Released on the SV only. But I guess it makes some kind of sense? In theory, the SCP doesn't really draw anything, it's just a boring ContentPresenter that takes care of scrolling, so it it has nothing that hit-test can check against; ergo, it's not hit-testable and doesn't receive pointer events.

src/Uno.UI/UI/Xaml/Media/VisualTreeHelper.cs
ramezgerges: No, it's just a (probably unnecessary) note for the next line. I adjusted the comments.

MartinZikmund:
@dr1rrb can you please re-review this one? Thanks a lot!

src/Uno.UI.Composition/Composition/ShapeVisual.skia.cs
dr1rrb: Great, this match what we did here https://github.com/unoplatform/uno/pull/17633/commits/f5c4c7c43d86413f3596c1322c42353b8464e837# 👍 (Once both PR have been merged, we will have to converge APIs)

src/Uno.UI/UI/Xaml/Controls/ScrollContentPresenter/ScrollContentPresenter.Managed.cs
```diff
@@ -65,17 +66,57 @@ public bool CanVerticallyScroll
 
 			_strategy.Initialize(this);
 
+			ManipulationMode = ManipulationModes.TranslateX | ManipulationModes.TranslateY; // Updated in PrepareTouchScroll!
+		}
+
+		internal void HookScrollEvents(ScrollViewer sv)
```
dr1rrb, line 72:
```suggestion
		private void HookScrollEvents(ScrollViewer sv)
```

src/Uno.UI/UI/Xaml/Controls/ScrollContentPresenter/ScrollContentPresenter.Managed.cs
```diff
@@ -65,17 +66,57 @@ public bool CanVerticallyScroll
...
 			// Touch scroll support
-			ManipulationMode = ManipulationModes.TranslateX | ManipulationModes.TranslateY; // Updated in PrepareTouchScroll!
+			// Note: Events are hooked on the SCP itself, not the ScrollViewer
 			ManipulationStarting += PrepareTouchScroll;
 			ManipulationStarted += TouchScrollStarted;
 			ManipulationDelta += UpdateTouchScroll;
 			ManipulationCompleted += CompleteTouchScroll;
 		}
 
+		internal void UnhookScrollEvents(ScrollViewer sv)
```
dr1rrb, line 92:
```suggestion
		private void UnhookScrollEvents(ScrollViewer sv)
```

src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.Managed.cs
dr1rrb: Is this still relevant ? (If so, dismiss my comment about making the method `private` ;) )

src/Uno.UI/UI/Xaml/Media/VisualTreeHelper.cs
```diff
@@ -659,9 +659,14 @@ internal static void ClearChildren(UIElement view)
 				}
 			}
 
-			// We didn't find any child at the given position, validate that element can be touched (i.e. not HitTestability.Invisible),
-			// and the position is in actual bounds (which might be different than the clipping bounds)
+			// We didn't find any child at the given position, validate that element can be touched,
+			// and the position is in actual bounds(which might be different than the clipping bounds)
+#if __SKIA__
+			if (element.HitTest(transformToElement.Inverse().Transform(testPosition)) && renderingBounds.Contains(testPosition))
```
dr1rrb, line 665: FYI @Youssef1313  (we will have to unify API once we merge both PR)

src/Uno.UI/UI/Xaml/Shapes/Shape.cs
```diff
@@ -168,8 +168,13 @@ public DoubleCollection StrokeDashArray
 		);
 		#endregion
 
+		// Do not invoke base.IsViewHit(): We don't have to have de FrameworkElement.Background to be hit testable!
 		internal override bool IsViewHit()
-			=> Fill != null; // Do not invoke base.IsViewHit(): We don't have to have de FrameworkElement.Background to be hit testable!
+			=> Fill != null
+#if __SKIA__ // we only add this condition for skia because it's the only platform with proper hit-testing support for shapes. If we add it for other platforms, we get a different but still inaccurate behaviour, so we prefer to keep the behaviour as is.
+				|| Stroke != null
```
dr1rrb, line 175: FYI @Youssef1313 :)

src/Uno.UI.Composition/Composition/SkiaGeometrySource2D.skia.cs
dr1rrb: @ramezgerges ? If possible I would expose needed APIs on this `SkiaGeometrySource2D` and keep that `Geometry` private rather than expose something that is risky.

dr1rrb (approved):

src/Uno.UI/UI/Xaml/Shapes/Shape.cs
Youssef1313: Great. That's the same change we'll need for Wasm as well so it will be `#if __SKIA__ || __WASM__` once **both** PRs are in.

src/Uno.UI.Composition/Composition/SkiaGeometrySource2D.skia.cs
ramezgerges: I'm open to the idea (but maybe in a separate PR), but I'm not really sure how you want this to go? What would we expose instead of the SKPath? Also, this doesn't fix the problem of someone with a reference to the SKPath (before constructing the SkiaGeometrySource2D) modifying it. I thought about cloning the SKPath in the constructor, but I'm not sure about the performance implications.

src/Uno.UI.Composition/Composition/BorderVisual.skia.cs
```diff
@@ -264,30 +265,40 @@ private unsafe void UpdateBackgroundPath(bool useInnerBorderBoundsAsAreaForBackg
 			roundRect.Handle,
 			&rect,
 			useInnerBorderBoundsAsAreaForBackground ? innerRadii : outerRadii);
-		_backgroundPath.AddRoundRect(roundRect);
-		_backgroundPath.Close();
+		backgroundPath.AddRoundRect(roundRect);
+		backgroundPath.Close();
+
+		// Unfortunately, this will cause an unnecessary render invalidation
+		((CompositionPathGeometry)_backgroundShape!.Geometry!).Path = new CompositionPath(new SkiaGeometrySource2D(backgroundPath));
```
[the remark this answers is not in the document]
ramezgerges: There shouldn't be any "users" here. The `BackgroundPath` is private to the BorderVisual (but was marked internal in another PR because it was needed for BackgroundTransition) and it not exposed to the user. If somehow the `Geometry` is not CompositionPathGeometry, we should indeed crash.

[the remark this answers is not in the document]
ramezgerges: I made it private now.

src/Uno.UI.Composition/Composition/SkiaGeometrySource2D.skia.cs
MartinZikmund: Let's resolve for now, can be handled in separate PR. The class itself is internal anyway, so it should not be a big problem

PR #16300, opened 2024-04-15
Fix Skia / SVG package references

dansiegel:

- unoplatform/uno-private#455

## PR Type

- Bugfix
- Feature

Uno.WinUI.Svg is never included
SkiaSharp packages are always included for Non-Windows/WASM heads

## What is the new behavior?

SVG and SkiaSharp packages are only included when building non-Windows/WASM projects when the `Skia` UnoFeature has been specified. The SVG package is only included on a Head project and is not included for libraries.


dansiegel:

PR #16343, opened 2024-04-17
fix(vsix): preserve the selected profile when chosen first

jeromelaban:
closes https://github.com/unoplatform/private/issues/426

### DIFF
--- a/src/Uno.UI.RemoteControl.VS/EntryPoint.ActiveProfileSync.cs
+++ b/src/Uno.UI.RemoteControl.VS/EntryPoint.ActiveProfileSync.cs
@@ -93,18 +93,17 @@ public partial class EntryPoint : IDisposable
 				// If the current profile already matches the targetframework we're going to
 				// prefer using it. It can happen if the change is initiated through the profile
 				// selector.
-				var currentCompatible = profiles
-					.Where(profileFilter)
-					.FirstOrDefault(p => p.Name == _debuggerObserver.CurrentActiveDebugProfile);
+				var selectedProfile = profiles
+					.FirstOrDefault(p => profileFilter(p) && p.Name == _debuggerObserver.CurrentActiveDebugProfile);
 
 				// Otherwise, select the first compatible profile.
-				var firstCompatible = profiles.Find(p => profileFilter(p));
+				selectedProfile ??= profiles.Find(p => profileFilter(p));
 
-				if ((currentCompatible ?? firstCompatible) is { } profile)
+				if (selectedProfile is not null)
 				{
-					_debugAction?.Invoke($"Setting profile {profile}");
+					_debugAction?.Invoke($"Setting profile {selectedProfile}");
 
-					await _debuggerObserver.SetActiveLaunchProfileAsync(profile.Name);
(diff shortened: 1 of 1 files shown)


src/Uno.UI.RemoteControl.VS/EntryPoint.ActiveProfileSync.cs
```diff
@@ -71,36 +71,40 @@ private async Task OnDebugFrameworkChangedAsync(string? previousFramework, strin
...
 
-					await _debuggerObserver.SetActiveLaunchProfileAsync(desktopProfile.Name);
-				}
-			}
-			else if (targetFrameworkIdentifier == Windows10TargetFrameworkIdentifier)
+				_ => (Func<ILaunchProfile, bool>?)null
+			};
+
+			if (profileFilter is not null)
 			{
```
agneszitte, line 92:
```suggestion
if (profileFilter != null)
{
    // If the current profile already matches the targetframework we're going to
    // prefer using it. It can happen if the change is initiated through the profile selector.
    var currentCompatible = profiles.FirstOrDefault(p => p.Name == _debuggerObserver.CurrentActiveDebugProfile && profileFilter(p));
    // Otherwise, select the first compatible profile.
    var firstCompatible = currentCompatible ?? profiles.FirstOrDefault(profileFilter);

    if (firstCompatible != null)
    {
        _debugAction?.Invoke($"Setting profile {firstCompatible}");
        await _debuggerObserver.SetActiveLaunchProfileAsync(firstCompatible.Name);
    }
}
```
Maybe we can streamline the evaluation of `currentCompatible` and `firstCompatible` with something like that?

src/Uno.UI.RemoteControl.VS/EntryPoint.ActiveProfileSync.cs
```diff
@@ -71,36 +71,40 @@ private async Task OnDebugFrameworkChangedAsync(string? previousFramework, strin
...
+					=> (ILaunchProfile p) => p.CommandName.Equals("MsixPackage", StringComparison.OrdinalIgnoreCase),
 
-					await _debuggerObserver.SetActiveLaunchProfileAsync(desktopProfile.Name);
-				}
-			}
-			else if (targetFrameworkIdentifier == Windows10TargetFrameworkIdentifier)
+				_ => (Func<ILaunchProfile, bool>?)null
+			};
+
+			if (profileFilter is not null)
```
agneszitte, line 91:
```suggestion
if (profileFilter != null)
{
    // If the current profile already matches the targetframework we're going to
    // prefer using it. It can happen if the change is initiated through the profile selector.
    var currentCompatible = profiles.FirstOrDefault(p => p.Name == _debuggerObserver.CurrentActiveDebugProfile && profileFilter(p));
    // Otherwise, select the first compatible profile.
    var firstCompatible = currentCompatible ?? profiles.FirstOrDefault(profileFilter);

    if (firstCompatible != null)
    {
        _debugAction?.Invoke($"Setting profile {firstCompatible}");
        await _debuggerObserver.SetActiveLaunchProfileAsync(firstCompatible.Name);
    }
}
```
Maybe we can streamline the evaluation of `currentCompatible` and `firstCompatible` with something like that?

jeromelaban:

PR #16361, opened 2024-04-18
Handle duplicate XAML files in codegen

jeromelaban:
GitHub Issue (If applicable): closes https://github.com/unoplatform/private/issues/433

### DIFF
--- a/src/SourceGenerators/SourceGeneratorHelpers/Helpers/GeneratorExecutionContextExtensions.cs
+++ b/src/SourceGenerators/SourceGeneratorHelpers/Helpers/GeneratorExecutionContextExtensions.cs
@@ -42,6 +42,7 @@ namespace Uno.Roslyn
 	public class MSBuildItem
 	{
 		private readonly GeneratorExecutionContext Context;
+		private static MSBuildItemIdentityComparer? _identityComparer;
 
 		internal MSBuildItem(AdditionalText file, GeneratorExecutionContext context)
 		{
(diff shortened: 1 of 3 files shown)


jeromelaban:

src/SourceGenerators/SourceGeneratorHelpers/Helpers/GeneratorExecutionContextExtensions.cs
```diff
@@ -67,5 +68,17 @@ public string GetMetadataValue(string name)
...
 			return string.IsNullOrEmpty(metadataValue) ? "" : metadataValue!;
 		}
+
+		public static IEqualityComparer<MSBuildItem> IdentityComparer
+			=> _identityComparer ??= new MSBuildItemIdentityComparer();
+
+		private class MSBuildItemIdentityComparer : IEqualityComparer<MSBuildItem>
+		{
+			public bool Equals(MSBuildItem x, MSBuildItem y)
+				=> string.Equals(x.Identity, y.Identity, System.StringComparison.OrdinalIgnoreCase);
```
Youssef1313, line 78: Not sure if we should use StringComparison based on file system's case sensitivity?

Youssef1313 (approved):

src/SourceGenerators/SourceGeneratorHelpers/Helpers/GeneratorExecutionContextExtensions.cs
jeromelaban: I wondered, the problem is that it's not easy to determine if the file system should be case sensitive, as it depends on the file location. In our case, having case sensitivity differences is likely to cause other bugs.

agneszitte (approved):

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlCodeGeneration.cs
```diff
@@ -172,8 +172,9 @@ public XamlCodeGeneration(GeneratorExecutionContext context)
 			var applicationDefinitionItems = GetWinUIItems("ApplicationDefinition").Concat(GetWinUIItems("UnoApplicationDefinition"));
 
 			var xamlItems = pageItems
+				.Except(applicationDefinitionItems, MSBuildItem.IdentityComparer)
 				.Concat(applicationDefinitionItems)
```
Youssef1313, line 176: Except(applicationDefinitionItems).Concat(applicationDefinitionItems) is quite confusing 😕

jeromelaban: It's the order, I want to keep the original itemgroup, name favoring the `ApplicationDefinition` over the `Page.

PR #16373, opened 2024-04-19
Hot Restart fixes

jeromelaban:
GitHub Issue (If applicable): closes https://github.com/unoplatform/uno-private/issues/463

## PR Type

- Bugfix

## What is the new behavior?

- Ensures that PublishFolderType is set to avoid `The file XXX does not specify a 'PublishFolderType' metadata, and a default value could not be calculated. The file will not be copied to the app bundle.`
- Ensures that assets generation properly sets and uses the `TargetPath` property to avoid HotRestart to write incorrect file paths


jeromelaban:

PR #15543, opened 2024-02-19
feat: Support for touch/pen input in WPF

MartinZikmund:
GitHub Issue (If applicable): closes https://github.com/unoplatform/uno/issues/15542

## PR Type

What kind of change does this PR introduce?

- Feature

## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)
- [ ] [Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [ ] Validated PR `Screenshots Compare Test Run` results.
- [x] Contains **NO** breaking changes
- [x] Associated with an issue (GitHub or internal) and uses the [automatic close keywords](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue).
- [x] Commits must be following the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) specification.

Issue #14393, opened 2023-11-11
[Skia.WPF] Multi-Touch support


lindexi:
### What would you like to be added

Adding the multi-touch support to UNO.Skia.WPF.

### Why is this needed

The original WPF has very good touch support. But that's what UNO.Skia.WPF is missing right now. See https://github.com/unoplatform/uno/issues/4309

### For which platform

Skia.WPF

### Anything else we need to know?

WPF distinguishes WM_Pointer from Mouse and Stylus and Touch events.

lindexi:
I'm delighted to be involved in the development of this feature, and I'm also quite familiar with the touch and ink modules of WPF.

Youssef1313:
@lindexi I don't think this was implemented. So, re-opening it.

lindexi:
Thank you @Youssef1313

lindexi:
After reviewing the source code of UNO, I believe we can modify the original method of only obtaining events from the Mouse, to simultaneously obtaining events from both the Stylus and Mouse. Moreover, when the Mouse determines that the current event is from the Stylus or Touch, it will automatically ignore the subsequent processing logic. According to the touch behavior of WPF, the Touch event is triggered first, then promoted to the Stylus event, and finally promoted to the Mouse event. At the same time, Pen messages will trigger the Stylus event first, and then be promoted to the Mouse event. This means that for simple processing, using the Stylus event can handle both Touch and Pen situations. When using the Stylus event, multi-touch support will be automatically enabled. Also, by using the GetPropertyValue method and passing in StylusPointProperties.Width and StylusPointProperties.Height, we can obtain the possible size of the touch, which can be used to fill in the Microsoft.UI.Input.PointerPointProperties.ContactRect attribute. This will be beneficial for WPF applications on touch screens using the UNO framework.

MartinZikmund:
I think the ability to distinguish input types is working since https://github.com/unoplatform/uno/pull/15543 , but maybe the additional properties are missing?

lindexi:
@MartinZikmund Sorry for delay. As you can see, the  `WpfCorePointerInputSource` will only add the Mouse events and it do not add any Stylus and Touch event. That means it can not handle the multi-touch events.

lindexi:
@MartinZikmund I create https://github.com/unoplatform/uno/pull/16378 and could you review my code?

MartinZikmund:
The changes are merged now!

PR #16378, opened 2024-04-20
feat(wpf): Add Multi-Touch support on WPF platform

lindexi:
GitHub Issue (If applicable): #14393

## PR Type

- Feature

In the current WPF platform, if multi-touch is used, only the touch message from one finger touching the screen will be received.

## What is the new behavior?

We can receive the multi-touch events message.

## PR Checklist

- [ ] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features) Not need.
- [ ] [Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable) I test pass by UnoIslandsSamplesApp.Skia.Wpf

## Other information

Reference https://github.com/unoplatform/uno/pull/15543

Cc: @MartinZikmund

### DIFF
--- a/src/Uno.UI.Runtime.Skia.Wpf/Input/WpfCorePointerInputSource.cs
+++ b/src/Uno.UI.Runtime.Skia.Wpf/Input/WpfCorePointerInputSource.cs
@@ -17,6 +17,8 @@ using Windows.Foundation;
 using Windows.System;
 using Windows.UI.Core;
 using Windows.UI.Input;
+using Point = System.Windows.Point;
+using Rect = Windows.Foundation.Rect;
 using WpfControl = System.Windows.Controls.Control;
 using WpfMouseEventArgs = System.Windows.Input.MouseEventArgs;
 
@@ -52,9 +54,15 @@ internal sealed class WpfCorePointerInputSource : IUnoCorePointerInputSource
 
 		_hostControl.MouseEnter += HostOnMouseEnter;
 		_hostControl.MouseLeave += HostOnMouseLeave;
+
+		_hostControl.StylusMove += HostControlOnStylusMove;
+		_hostControl.StylusDown += HostControlOnStylusDown;
+		_hostControl.StylusUp += HostControlOnStylusUp;
+
 		_hostControl.MouseMove += HostOnMouseMove;
 		_hostControl.MouseDown += HostOnMouseDown;
 		_hostControl.MouseUp += HostOnMouseUp;
+
 		_hostControl.LostMouseCapture += HostOnMouseCaptureLost;
 
 		// Hook for native events
@@ -103,7 +111,7 @@ internal sealed class WpfCorePointerInputSource : IUnoCorePointerInputSource
 		=> _hostControl.ReleaseMouseCapture();
 
 	#region Native events
-	private void HostOnMouseEvent(WpfMouseEventArgs args, TypedEventHandler<object, PointerEventArgs>? @event, [CallerArgumentExpression(nameof(@event))] string eventName = "")
+	private void HostOnMouseEvent(InputEventArgs args, TypedEventHandler<object, PointerEventArgs>? @event, [CallerArgumentExpression(nameof(@event))] string eventName = "")
 	{
 		var current = SynchronizationContext.Current;
 		try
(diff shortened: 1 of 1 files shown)


lindexi:
I am currently attempting to build a test application for use on my touch device. I am encountering some issues with the UNO build, but I suspect it might be due to my local configuration. I am in the process of trying to fix it.

lindexi:
I fixed the build issues and I test pass by UnoIslandsSamplesApp.Skia.Wpf

MartinZikmund:
I think the problem with this approach was that the input events always eventually go to Mouse events, with StylusDevice set, so subscribing to both touch and mouse will mean all events will be triggered twice. But not completely sure, just trying to remember why I didn't go this route. Could you differentiate between touches via the StylusDevice property instead?

lindexi:
@MartinZikmund Thank you. You're absolutely right. Whether it's touch or mouse events, they ultimately get converted into mouse events. In the WPF framework, touch events are converted into mouse events using the Promote technique. However, this conversion process results in loss of information. In the WPF framework, only the touch information from the first finger in a multi-touch scenario is converted into mouse information, while the information from other fingers is lost. Moreover, the conversion process loses touch dimensions such as width, height, and pressure sensitivity. This is why I need to separately listen to multiple events. Although listening to both Stylus and Mouse events means that I will receive two event notifications during the touch process, fortunately, I can determine in the mouse event whether the current mouse event was triggered by touch, allowing me to filter out the touch events that have already been processed.

MartinZikmund:
@lindexi oh I see it now, filtering out events with `StylusDevice` set! That's very nice!

MartinZikmund (approved):

lindexi:
Thank you @MartinZikmund

src/Uno.UI.Runtime.Skia.Wpf/Input/WpfCorePointerInputSource.cs
```diff
@@ -230,21 +259,86 @@ private IntPtr OnWmMessage(IntPtr hwnd, int msg, IntPtr wparamOriginal, IntPtr l
...
+			};
+		}
+		else if (args is StylusEventArgs stylusEventArgs)
+		{
+			pointerId = (uint)stylusEventArgs.StylusDevice.Id;
+			position = stylusEventArgs.GetPosition(_hostControl);
+
+			properties = new()
+			{
+				IsLeftButtonPressed = true,
```
dr1rrb, line 295: For touch, this has to be `false` for release and cancel.
For pen, it has to be `true` only when `!stylusEventArgs.InAir`.

Also for the pen do we have information about the barel button and the eraser (for the `IsRightButtonPressed` and `IsEraser`)?

src/Uno.UI.Runtime.Skia.Wpf/Input/WpfCorePointerInputSource.cs
```diff
@@ -230,21 +259,86 @@ private IntPtr OnWmMessage(IntPtr hwnd, int msg, IntPtr wparamOriginal, IntPtr l
...
+				IsPrimary = true,
+				IsInRange = !stylusEventArgs.InAir,
```
dr1rrb, line 297: This is actually always `true` except for some `Cancel` (and maybe also `Release` for touch).

src/Uno.UI.Runtime.Skia.Wpf/Input/WpfCorePointerInputSource.cs
lindexi: @dr1rrb Yeah. Get the barel button by this code:

```
        private void MainWindow_StylusMove(object sender, StylusEventArgs e)
        {
            if (e.StylusDevice.StylusButtons.GetStylusButtonByGuid(StylusPointProperties.BarrelButton.Id).StylusButtonState == StylusButtonState.Down)
            {
                
            }
        }
```

lindexi: Thank you @dr1rrb  and I create the PR in https://github.com/unoplatform/uno/pull/16517 . Could you review my code?

src/Uno.UI.Runtime.Skia.Wpf/Input/WpfCorePointerInputSource.cs
dr1rrb: From what I remember of my investigations, the `IsInRange` is about "do we know where the pointer currently is?", no matter if it's pressed or not.

* For mouse: does the mouse is still connected on the computer (or the USB cable as been unplugged)
* For pen: does the pen is in the detectable range (over) or touching the screen (pressed)
* For touch: does a finger is touching the screen

So when we raise a pointer event it's almost always `true` (except some `Cancel` and `Release` as mentioned). BUT you can actually keep a reference of a pointer events args (or the [`Pointer`](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.input.pointerroutedeventargs.pointer?view=windows-app-sdk-1.5)) for longer than the pointer event handler scope! Properties on that args/`Pointer` will then be updated by WinUI (this is currently not supported by uno BTW), and the `IsInRange` can become `false` in those cases.

(Note: not sure about the `IsInRange` state in windowed environment, when the pointer is above another window)

Issue #14349, opened 2023-11-08
[Skia] Compositor thread


MartinZikmund:
### What would you like to be added

Run composition operations on separate thread

### For which platform

Skia (WPF), Skia (GTK on Linux/macOS/Windows), Skia (Linux Framebuffer)

MartinZikmund:
Duplicate of https://github.com/unoplatform/uno-private/issues/1310

PR #16403, opened 2024-04-22
Skia Composition refactoring

ramezgerges:
GitHub Issue (If applicable): stepping stone for https://github.com/unoplatform/uno/issues/14349

## PR Type

## PR Checklist

## Other information


src/Uno.UI.Composition/Composition/Visual.skia.cs
```diff
@@ -178,7 +195,17 @@ internal virtual void Render(in DrawingSession parentSession)
...
 		using var session = BeginDrawing(in parentSession);
-		Draw(in session);
+		if (_requiresRepaint)
+		{
+			_requiresRepaint = false;
+			_recorder ??= new SKPictureRecorder();
+			var recorderSession = session with { Canvas = _recorder.BeginRecording(new SKRect(float.NegativeInfinity, float.NegativeInfinity, float.PositiveInfinity, float.PositiveInfinity)) };
+			// To debug what exactly gets repainted, replace the following line with `Draw(in session);`
+			Draw(in recorderSession);
+			_picture = _recorder.EndRecording();
```
dr1rrb, line 205: Should we consider to have a try/finally for this?

src/Uno.UI.Composition/Composition/Visual.skia.cs
```diff
@@ -178,7 +195,17 @@ internal virtual void Render(in DrawingSession parentSession)
 		}
 
 		using var session = BeginDrawing(in parentSession);
-		Draw(in session);
+		if (_requiresRepaint)
+		{
+			_requiresRepaint = false;
+			_recorder ??= new SKPictureRecorder();
```
dr1rrb, line 201: Since we are actually interested only by the `_picture`, can we pool the `SKPictureRecorder` instead of having one per `Visual`?

src/Uno.UI.Composition/Composition/Visual.skia.cs
```diff
@@ -178,7 +195,17 @@ internal virtual void Render(in DrawingSession parentSession)
...
+			var recorderSession = session with { Canvas = _recorder.BeginRecording(new SKRect(float.NegativeInfinity, float.NegativeInfinity, float.PositiveInfinity, float.PositiveInfinity)) };
+			// To debug what exactly gets repainted, replace the following line with `Draw(in session);`
+			Draw(in recorderSession);
+			_picture = _recorder.EndRecording();
+		}
+
+		session.Canvas.DrawPicture(_picture);
```
dr1rrb, line 208: Does the `_picture` is pixels based or only display list (a.k.a. "drawing directives")? On all platforms?

src/Uno.UI.Composition/Composition/ShapeVisual.skia.cs
```diff
@@ -1,96 +1,32 @@
...
-				shapes[i].Render(in session);
-			}
-		}
-
-		// Second we render the children
-		base.Render(in parentSession);
-	}
-
 	/// <inheritdoc />
 	internal override void Draw(in DrawingSession session)
```
dr1rrb, line 10: Should we rename this to "Paint" ? (and `PaintSession`)

src/Uno.UI.Composition/Composition/Visual.skia.cs
dr1rrb: Confirmed it's  display list that are store:

> The picture recorder is used to record drawing operations made to a [SKCanvas](https://learn.microsoft.com/en-us/dotnet/api/skiasharp.skcanvas?view=skiasharp-2.88) and stored in a [SKPicture](https://learn.microsoft.com/en-us/dotnet/api/skiasharp.skpicture?view=skiasharp-2.88).

https://learn.microsoft.com/en-us/dotnet/api/skiasharp.skpicturerecorder?view=skiasharp-2.88

ramezgerges: Yes. If I remember correctly, I read in some chromium design documents that SkPicture's resulting bitmap is cached as a texture after the first time it's drawn on an SKCanvas, but I'm not sure.

src/Uno.UI.Composition/Composition/Visual.skia.cs
ramezgerges: I don't think so. External code cannot access the internal painting logic, so an exception here means that something internal went catastrophically wrong, so I suppose we should actually crash.

src/Uno.UI.Composition/Composition/Visual.cs
```diff
@@ -133,6 +133,7 @@ public Vector3 RotationAxis
 		private protected override void OnPropertyChangedCore(string? propertyName, bool isSubPropertyChange)
 		{
 			Compositor.InvalidateRender(this);
+			InvalidatePaint(); // TODO: only repaint when "dependent" properties are changed
```
dr1rrb, line 136: Thinking about this (and the default private `_requiresRepaint = true;`), I realized that actually most of our `Visual` in uno apps will never have anything to draw, and we will create an incredible amount of empty `_picture` that are blank (not even considering the cost of the recording canvas etc.).

src/Uno.UI.Composition/Composition/Visual.skia.cs
```diff
@@ -178,21 +204,72 @@ internal virtual void Render(in DrawingSession parentSession)
...
 	}
 
 	/// <summary>
 	/// Draws the content of this visual.
 	/// </summary>
 	/// <param name="session">The drawing session to use.</param>
-	internal virtual void Draw(in DrawingSession session)
+	internal virtual void Paint(in PaintingSession session) { }
+
+	private Vector3 GetTotalOffset()
```
Youssef1313, line 238: I feel like we may want to keep this `internal` **even if** it's only used in `Visual` class. In case someone wanted to access `Offset` externally, it may be useful that they see IntelliSense suggesting `GetTotalOffset()` so they have to think which one they need.

src/Uno.UI.Composition/Composition/Uno/DrawingSession.skia.cs
```diff
@@ -10,9 +10,9 @@ namespace Uno.UI.Composition;
 // Accessing Surface.Canvas is slow due to SkiaSharp interop.
 // Avoid using .Surface.Canvas and use .Canvas right away.
 /// <param name="RootTransform">The transform matrix to the root visual of this drawing session (which isn't necessarily the identity matrix due to scaling (DPI) and/or RenderTargetBitmap.</param>
-internal record struct DrawingSession(SKSurface Surface, SKCanvas Canvas, in DrawingFilters Filters, in Matrix4x4 RootTransform) : IDisposable
+internal record struct PaintingSession(SKSurface Surface, SKCanvas Canvas, in DrawingFilters Filters, in Matrix4x4 RootTransform) : IDisposable
```
Youssef1313, line 13: While we're here, let's make it a `readonly struct`

src/Uno.UI.Composition/Composition/Visual.skia.cs
```diff
@@ -12,15 +14,39 @@ namespace Microsoft.UI.Composition;
 
 public partial class Visual : global::Microsoft.UI.Composition.CompositionObject
 {
+	// Since painting (and recording) is done on the UI thread, we need a single SKPictureRecorder per UI thread.
+	// If we move to a UI-thread-per-window model, then we need multiple recorders.
+	[ThreadStatic] private static SKPictureRecorder? _recorder;
```
Youssef1313, line 19: The more common formatting is attribute on separate line.

```suggestion
	[ThreadStatic]
	private static SKPictureRecorder? _recorder;
```

src/Uno.UI.Composition/Composition/LayerVisual.cs
```diff
@@ -1,12 +0,0 @@
-﻿#nullable enable
-
-namespace Microsoft.UI.Composition
-{
-	public partial class LayerVisual : global::Microsoft.UI.Composition.ContainerVisual
```
Youssef1313, line 5: This change adds a public constructor to `LayerVisual` that shouldn't be there.

src/Uno.UI.Composition/Composition/CompositionObject.Context.cs
```diff
@@ -7,118 +7,110 @@ namespace Microsoft.UI.Composition
 {
 	public partial class CompositionObject
 	{
-		private class ContextStore
```
Youssef1313, line 10: A nested class was separating the store from the composition object quite well. I personally prefer that separation over making `CompositionObject` class more large.

If there are performance concerns, we might make it a struct.

@dr1rrb What do you think?

ramezgerges: IMHO, it doesn't make sense to have a nested class that has the same lifetime as it's outer class and most of its methods have corresponding "passthrough methods" in the outer class. It just creates unnecessary indirections.

src/Uno.UI.Composition/Composition/Visual.skia.cs
ramezgerges: I don't have a strong opinion on this, but I would argue that internal code that isn't inside Visual should _not_ need to read the "total offset" (or any offset really) but only set it. Keeping it private makes it clear that you should think twice if you really need to depend on it.

src/Uno.UI.Composition/Composition/Visual.skia.cs
```diff
@@ -177,23 +184,60 @@ internal virtual void Render(in DrawingSession parentSession)
 			return;
 		}
 
-		using var session = BeginDrawing(in parentSession);
-		Draw(in session);
+		using var session = parentSession.WithOpacity(Opacity);
+		SetupLocalCoordinatesAndClip(in session);
```
dr1rrb, line 188: Why did you remove the `BeginDrawing`? This correspond to the logic of that method and the `PaintingSession` was designed to abstract the notion/count of `Canvas.Save` and `Canvas.Restore`.

Here you create a new session in the `WIthOpacity` and then do the `Save` in `SetupLocalCoordinatesAndClip`. Invoking one of the 2 method without invoking the other one will cause issue. Please merge them back. 

(And it was the purpose of the `PushOpacity`: setuping current session without creating a new one ... but its probably no longer relevant since I think we does that only here).

src/Uno.UI.Composition/Composition/Visual.skia.cs
```diff
@@ -177,23 +184,60 @@ internal virtual void Render(in DrawingSession parentSession)
...
+		Paint(in session);
+
+		// The CornerRadiusClip doesn't affect the visual itself, only its children
+		CornerRadiusClip?.Apply(session.Canvas, this);
+
+		foreach (var child in GetChildrenInRenderOrder())
+		{
+			child.Render(in session);
+		}
```
dr1rrb, line 197: Not sure I'm a fan of this ... sounds like if we are again trying to support all cases in the `Visual` instead of deferring the responsibility in each sub-class.

src/Uno.UI.Composition/Composition/Uno/PaintingSession.skia.cs
```diff
@@ -10,16 +11,19 @@ namespace Uno.UI.Composition;
...
 		// We try to keep the filter ref as long as possible in order to share the same filter.OpacityColorFilter
 		if (opacity is not 1.0f)
 		{
-			var filters = session.Filters;
-			session = session with { Filters = filters with { Opacity = filters.Opacity * opacity } };
+			var filters = Filters;
+			return this with { Filters = filters with { Opacity = filters.Opacity * opacity } };
 		}
+
+		return this;
```
dr1rrb, line 26: Cf. comment in the `Visual.skia.cs#Render` ... but I think this is dangerous: the number of `PaintingSession` (and their disposal) is linked to the number of `Save` we do on the `Canvas`. "Creating" a session should not be implciit like here. 

(Note: I used a `record` only for the light syntax, but we should consider to actually convert it to a standard `struct` to avoid the risk to implicitly create new sessions like here with the `with` keyword)

src/Uno.UI.Composition/Composition/CompositionObject.Context.cs
```diff
@@ -7,118 +7,110 @@ namespace Microsoft.UI.Composition
...
-		{
-			private readonly object _lock = new object();
-			private List<ContextEntry>? _contextEntries;
+		private readonly object _lock = new object();
```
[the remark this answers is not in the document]
ramezgerges: `_contextEntries` is created only as needed, so I went with the renaming. I think we scan still go without a separate lock object, but I try to not be too clever when it comes to concurrency problems.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Media/Given_AcrylicBrush.cs
```diff
@@ -23,6 +23,7 @@ public class Given_AcrylicBrush
 #if __SKIA__ && HAS_UNO
 	[TestMethod]
 	[RunsOnUIThread]
+	[Ignore("Crashes")]
```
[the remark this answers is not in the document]
ramezgerges: My bad. I ignored it to see what else fails in CI, but then forgot to unignore it.

src/Uno.UI.Composition/Composition/Visual.skia.cs
ramezgerges: I definitely agree with this. In fact, I just went a step further (I most likely went way too far with this actually): a creating and disposing a `PaintingSession` should not concern anything outside the `Visual` that creates it to start a new "local session". I moved `PaintingSession` to be nested inside `Visual` and tried to make it work in an RAII-ish way. Let me know if you think it's too over the top.

src/Uno.UI.Composition/Composition/Visual.skia.cs
```diff
@@ -177,52 +187,75 @@ internal virtual void Render(in DrawingSession parentSession)
...
+		var canvas = parentSession.Canvas;
+		var rootTransform = parentSession.RootTransform;
+		// We try to keep the filter ref as long as possible in order to share the same filter.OpacityColorFilter
+		var filters = Opacity is 1.0f
+			? parentSession.Filters
+			: parentSession.Filters with { Opacity = parentSession.Filters.Opacity * Opacity };
 
-		// CornerRadiusClip applies to the children only.
-		CornerRadiusClip?.Apply(canvas, this);
+		var session = ((IPrivateSessionFactory)new PaintingSession.SessionFactory()).CreateInstance(this, surface, canvas, filters, rootTransform);
```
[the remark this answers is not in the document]
ramezgerges: I've removed the `PaintaingSessionWrapper`, but I've kept the `IPrivateSessionFactory` as there's apparently no other way to make a nested class only constructable within the outer class. It does look a bit ugly, but it's not a lot of code and it locks the lifetime of the PaintingSession objects within the Visual, so I would think it's a net win.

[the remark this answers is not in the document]
Youssef1313: An alternative that comes to mind could be an analyzer that errors when a paining session is created outside Visual.

PR #16479, opened 2024-04-26
refactor: Remove unnecessary `OnChildChangedPartial`

MartinZikmund:

## PR Type

We have `AddChild` and `RemoveChild` on each target, no need to make partials for them

## What is the new behavior?

Removed unnecessary partials.

## PR Checklist

## Other information


src/Uno.UI/UI/Xaml/Controls/Border/Border.iOSmacOS.cs
```diff
@@ -20,20 +20,8 @@
...
-	partial void OnChildChangedPartial(UIElement previousValue, UIElement newValue)
-	{
-		previousValue?.RemoveFromSuperview();
-
-		if (newValue != null)
-		{
-			AddSubview(newValue);
-		}
-
-		UpdateBorder();
```
Youssef1313, line 34: Was this call not necessary on iOS and macOS?

src/Uno.UI/UI/Xaml/Controls/Border/Border.cs
```diff
@@ -31,255 +31,260 @@
...
-			);
-
-		private void OnChildChanged(UIElement oldValue, UIElement newValue)
+		if (newValue is not null)
 		{
-			ReAttachChildTransitions(oldValue, newValue);
-
-			OnChildChangedPartial(oldValue, newValue);
+			AddChild(newValue);
 		}
```
Youssef1313, line 105: I think `AddChild` internally does a null check (at least on Skia)?
If that's the case on all platforms, I think we can just call `AddChild` unconditionally here.

src/Uno.UI/UI/Xaml/Controls/Border/Border.iOSmacOS.cs
MartinZikmund: `BorderLayerRenderer` attaches itself to `LayoutUpdated` event, so I assume this is superfluous as it will happen anyway.

src/Uno.UI/UI/Xaml/Controls/Border/Border.cs
Youssef1313: Just confirmed, this is the case only on Skia and Wasm, so let's keep the null check as it's likely needed for all mobile platforms.

MartinZikmund: The check is not uniform unfortunately. I could move the check to `VisualTreeHelper.AddChild` though (same for `RemoveChild`)

src/Uno.UI/UI/Xaml/Controls/Border/Border.iOSmacOS.cs
Youssef1313: Just to make sure, I think `AfterUpdateBorderPartial` will not be called when `LayoutUpdated` is fired. Only `BorderLayerRenderer.Update` is called. Want to make sure this is intended.

Youssef1313 (approved):

ajpinedam (approved):

src/Uno.UI/UI/Xaml/Controls/Border/Border.iOSmacOS.cs
jeromelaban: I'd tend to think that `AfterUpdateBorderPartial` is still required as well.

MartinZikmund: Yes, adjusted the code so that the logic happens inside `BorderLayerRenderer` itself, which simplifies things

PR #16481, opened 2024-04-26
test: Avoid the use of IdleSynchronizer.Wait in runtime tests

Youssef1313:

## PR Type

On Wasm, `IdleSynchronizer.Wait` is currently no-op as we don't have threading.

https://github.com/unoplatform/uno/blob/df5929ab80d13eaaa03a1fc62756440d65d3d253/src/Uno.UI.RuntimeTests/MUX/Utilities/IdleSynchronizer.cs#L11-L25

This is functionally incorrect. It may be the root cause of some test failures that are specific to Wasm, and is a bigger problem for enhanced lifecycle because in some cases we really need to wait for idle whereas previously it would have been okay.

## What is the new behavior?

Use `await TestServices.WindowHelper.WaitForIdle` instead.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI.RuntimeTests/MUX/Utilities/IdleSynchronizer.cs
+++ b/src/Uno.UI.RuntimeTests/MUX/Utilities/IdleSynchronizer.cs
@@ -8,6 +8,7 @@ namespace MUXControlsTestApp.Utilities
 {
 	internal class IdleSynchronizer
 	{
+		[Obsolete("Don't use IdleSynchronizer.Wait. Use 'await TestServices.WindowHelper.WaitForIdle()' instead.")]
 		public static void Wait()
 		{
 #if __WASM__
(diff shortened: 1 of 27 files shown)


MartinZikmund (approved):

src/Uno.UI.RuntimeTests/MUX/Utilities/IdleSynchronizer.cs
```diff
@@ -8,6 +8,7 @@ namespace MUXControlsTestApp.Utilities
 {
 	internal class IdleSynchronizer
 	{
+		[Obsolete("Don't use IdleSynchronizer.Wait. Use 'await TestServices.WindowHelper.WaitForIdle()' instead.")]
```
MartinZikmund, line 11: Maybe worth even banning the API?

Youssef1313: As it's internal, obsoletion should be okay.

Usually, the BannedApiAnalyzer is useful for two scenarios:

1. An API that comes from a package/library that we don't control.
2. An API that we control but is either public and we don't want to obsolete it publicly, or we want the obsoletion for specific project.

But if it's an API that we want to "ban everywhere", I think Obsolete should be good.

PR #16509, opened 2024-04-29
fix: Adjust desktop hot load under linux

jeromelaban:
This change ensures that delta updates do not fail on missing assemblies, which can happen when performing HR reload on linux and HR picks up WPF assemblies that are not loaded in the current process.


jeromelaban:

agneszitte (approved):

PR #16714, opened 2024-05-14
fix(visualstates): fix rolling back properties when the next state doesn't modify a property, but modifies one of its subproperties

ramezgerges:

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/DataBinding/BindingPath.BindingItem.cs
+++ b/src/Uno.UI/DataBinding/BindingPath.BindingItem.cs
@@ -12,10 +12,8 @@ namespace Uno.UI.DataBinding
 {
 	internal partial class BindingPath
 	{
-		private sealed class BindingItem : IBindingItem, IDisposable
+		private sealed class BindingItem : IBindingItem, IDisposable, IEquatable<BindingItem>
 		{
-			private delegate void PropertyChangedHandler(object? previousValue, object? newValue, bool shouldRaiseValueChanged);
-
 			private ManagedWeakReference? _dataContextWeakStorage;
 			private Flags _flags;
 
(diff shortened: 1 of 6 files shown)


src/Uno.UI/DataBinding/BindingPath.BindingItem.cs
```diff
@@ -540,6 +538,17 @@ public void NewValue()
...
+
+			/// <remarks>
+			/// This is defined so that 2 BindingItems are equal if they refer to the same property on the same object
+			/// even if they're a part of 2 different "chains".
+			/// </remarks>
+			public bool Equals(BindingItem? that)
+			{
+				return that != null
+					&& !DependencyObjectStore.AreDifferent(this.DataContext, that.DataContext)
+					&& this.PropertyName.Replace(")", "").Replace("(", "").Split(':', '.')[^1] == that.PropertyName.Replace(")", "").Replace("(", "").Split(':', '.')[^1];
```
ramezgerges, line 550: What's blocking this PR is this line. We need to find a way to compare two `BindingItem.PropertyNam`s and figure out if they point to the same thing. E.g. `(Microsoft.UI.Xaml.Controls:Border.Background)` and `Background` are the same thing. There are also expressions with indexers and some other weird stuff.

ramezgerges:
cc. @kazo0, this should fix the Toolkit visual states regression.

kazo0:
fyi @ramezgerges, this original fix seemed to have fixed https://github.com/unoplatform/uno/issues/14262

So rolling it back might unfix it, maybe?

@eriklimakc maybe check from this branch if the issue happens again

ramezgerges:
@kazo0 It shouldn't. The title is a bit misleading :). This PR doesn't "roll back" the previous fix, but it fixes a certain scenario where some properties set by animations should not be rolled back (they get rolled back after the other PR when they shouldn't). So it does indeed move some cases to the old behaviour, but it's very unlikely that that was what fixed #14262.

eriklimakc:
It works fine with the changes from this PR :)

MartinZikmund (approved):

PR #16562, opened 2024-05-02
feat: introduce a new BorderVisual to handle rendering the border and the background

ramezgerges:
This fixes a bunch of rendering problems in borders and simplifies the logic significantly.

GitHub Issue (If applicable): closes #16518, closes #16294

## PR Type

What kind of change does this PR introduce?

## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)
- [ ] [Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [ ] Validated PR `Screenshots Compare Test Run` results.
- [ ] Contains **NO** breaking changes
- [ ] Associated with an issue (GitHub or internal) and uses the [automatic close keywords](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue).
- [ ] Commits must be following the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) specification.

## Other information

Internal Issue (If applicable):

PR #16741, opened 2024-05-15
refactor: Remove reflection usage in Brush

dr1rrb:
## Refactoring (no functional changes, no api changes)
Remove reflection usage in Brush

We use reflection to invoke event handler

## What is the new behavior?
No reflection!

## PR Checklist

### DIFF
--- a/src/Uno.UI.RuntimeTests/Tests/Uno_UI_Helpers/Given_WeakEventManager.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Uno_UI_Helpers/Given_WeakEventManager.cs
@@ -1,0 +1,103 @@
+﻿#nullable enable
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Uno.UI.Helpers;
+
+namespace Uno.UI.RuntimeTests.Tests.Uno_UI_Helpers;
+
+[TestClass]
+public class Given_WeakEventManager
+{
+	private sealed class EventPublisher
+	{
+		private readonly WeakEventManager _manager = new();
+
+		internal event Action Event
+		{
+			add => _manager.AddEventHandler(value);
+			remove => _manager.RemoveEventHandler(value);
+		}
+	}
+
+	private sealed class EventSubscriber
+	{
+		public void M() { }
+	}
+
+	[TestMethod]
+	public void When_ManySubscriptions_Then_DoesNotLeak()
+	{
+		var publisher = new EventPublisher();
+		var weakRefs = new List<WeakReference<EventSubscriber>>();
+		Subscribe(publisher, weakRefs);
+
+		for (var i = 0; i < 10; i++)
+		{
+			GC.Collect();
+			GC.WaitForPendingFinalizers();
+		}
+
+		foreach (var x in weakRefs)
+		{
+			Assert.IsFalse(x.TryGetTarget(out _));
+		}
+	}
+
+	[TestMethod]
+	public void When_ReEnter_Then_AllHandlersInvokedProperly()
+	{
+		var sut = new WeakEventManager();
+		int handler1Count = 0, handler2Count = 0;
+
+		sut.AddEventHandler(Handler1, "Event1");
+		sut.AddEventHandler(Handler2, "Event1");
+		sut.AddEventHandler(Handler1_2, "Event2");
+		sut.AddEventHandler(Handler2_2, "Event2");
+
+		sut.HandleEvent("Event1");
+
+		Assert.AreEqual(3, handler1Count);
+		Assert.AreEqual(3, handler2Count);
+
+		sut.HandleEvent("Event2");
+
+		Assert.AreEqual(4, handler1Count);
+		Assert.AreEqual(4, handler2Count);
+
+		void Handler1()
+		{
+			handler1Count++;
+			sut.HandleEvent("Event2");
+		}
+
+		void Handler1_2()
+		{
+			handler1Count++;
+		}
+
+		void Handler2()
+		{
+			handler2Count++;
+			sut.HandleEvent("Event2");
+		}
+
+		void Handler2_2()
+		{
+			handler2Count++;
+		}
+	}
+
+	private void Subscribe(EventPublisher publisher, List<WeakReference<EventSubscriber>> weakRefs)
+	{
+		for (var i = 0; i < 1000; i++)
+		{
+			var subscriber = new EventSubscriber();
+			publisher.Event += subscriber.M;
+			weakRefs.Add(new WeakReference<EventSubscriber>(subscriber));
+		}
+	}
+}
--- a/src/Uno.UI.RuntimeTests/Tests/Uno_UI_Helpers/Given_WeakEventManager.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Uno_UI_Helpers/Given_WeakEventManager.cs
@@ -21,6 +21,8 @@ public class Given_WeakEventManager
 			add => _manager.AddEventHandler(value);
 			remove => _manager.RemoveEventHandler(value);
 		}
+
+		public void RaiseEvent() => _manager.HandleEvent("Event");
 	}
 
 	private sealed class EventSubscriber
@@ -86,6 +88,60 @@ public class Given_WeakEventManager
 		}
 
 		void Handler2_2()
+		{
+			handler2Count++;
+		}
+	}
+
+	[TestMethod]
+	public void When_UnSubscribeInHandler()
+	{
+		var sut = new WeakEventManager();
+		int handler1Count = 0, handler2Count = 0;
+
+		sut.AddEventHandler(Handler1, "Event1");
+		sut.AddEventHandler(Handler2, "Event1");
+
+		sut.HandleEvent("Event1");
+		sut.HandleEvent("Event1");
+
+		Assert.AreEqual(1, handler1Count);
+		Assert.AreEqual(2, handler2Count);
+
+		void Handler1()
+		{
+			handler1Count++;
+			sut.RemoveEventHandler(Handler1, "Event1");
+		}
+
+		void Handler2()
+		{
+			handler2Count++;
+		}
+	}
+
+	[TestMethod]
+	public void When_UnSubscribeInHandler2()
+	{
+		var pub = new EventPublisher();
+		int handler1Count = 0, handler2Count = 0;
+
+		pub.Event += Handler1;
+		pub.Event += Handler2;
+
+		pub.RaiseEvent();
+		pub.RaiseEvent();
+
+		Assert.AreEqual(1, handler1Count);
+		Assert.AreEqual(2, handler2Count);
+
+		void Handler1()
+		{
+			handler1Count++;
+			pub.Event -= Handler1;
+		}
+
+		void Handler2()
 		{
 			handler2Count++;
 		}
--- a/src/Uno.UI.RuntimeTests/Tests/Uno_UI_Helpers/Given_WeakEventManager.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Uno_UI_Helpers/Given_WeakEventManager.cs
@@ -1,4 +1,5 @@
 ﻿#nullable enable
+#if HAS_UNO
 
 using System;
 using System.Collections.Generic;
@@ -157,3 +158,4 @@ public class Given_WeakEventManager
 		}
 	}
 }
+#endif
--- a/src/Uno.UI/Helpers/WeakEvents/WeakEventManager.cs
+++ b/src/Uno.UI/Helpers/WeakEvents/WeakEventManager.cs
@@ -102,7 +102,11 @@ internal sealed class WeakEventManager
 		{
 			if (_enumeratingHandlersOfEvent == eventName)
 			{
+#if NET8_0_OR_GREATER
 				handlers = handlers![..];
+#else
+				handlers = handlers!.ToArray();
+#endif
 			}
 
 			handlers!.Add(handler);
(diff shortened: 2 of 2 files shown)


carldebilly (approved):

src/Uno.UI/Helpers/WeakEvents/WeakEventManager.cs
```diff
@@ -6,130 +6,110 @@
...
-		{
-			if (string.IsNullOrEmpty(eventName))
-				throw new ArgumentNullException(nameof(eventName));
+	private readonly ConditionalWeakTable<object, Dictionary<string, List<Action>>> _handlersPerTarget = new();
+	private Dictionary<string, List<Action>>? _staticHandlers;
+	private string? _enumeratingHandlersOfEvent;
 
-			if (handler == null)
-				throw new ArgumentNullException(nameof(handler));
+	public void AddEventHandler(Action handler, [CallerMemberName] string eventName = "")
```
Youssef1313, line 17: Let's keep it nullable. The problem is that events are usually declared nullable, so the compiler thinks `value` in the context of add/remove is also nullable. Making this nullable helps not having to suppress nullability warning in callsites.

src/Uno.UI/Helpers/WeakEvents/WeakEventManager.cs
```diff
@@ -6,130 +6,110 @@
...
-			if (!_eventHandlers.TryGetValue(eventName, out List<Subscription>? subscriptions))
-				return;
+			RemoveFrom(handlersPerEvent, eventName, handler);
+		}
+	}
 
-			for (int n = subscriptions.Count - 1; n >= 0; n--)
+	public void HandleEvent(string eventName)
+	{
+		_enumeratingHandlersOfEvent = eventName; // Note: We do not support re-entrancy!
```
Youssef1313, line 60: I think re-entrancy should actually be allowed. It may not have been allowed in the existing implementation, but that means the existing implementation is not correct :)

src/Uno.UI/Helpers/WeakEvents/WeakEventManager.cs
```diff
@@ -6,130 +6,110 @@
...
 				}
 
-				if (current.Subscriber?.Target == handlerTarget && current.Handler == methodInfo)
+				var count = handlers.Count;
+				for (var i = 0; i < count; i++)
 				{
-					// Found the match, we can break
-					subscriptions.RemoveAt(n);
-					break;
+					handlers[i]();
```
Youssef1313, line 74: The handler could end up unsubscribing another handler. This is an existing issue though

src/Uno.UI/Helpers/WeakEvents/WeakEventManager.cs
```diff
@@ -6,130 +6,110 @@
...
+	private void AddTo(Dictionary<string, List<Action>> handlersPerEvent, string eventName, Action handler)
+	{
+		if (handlersPerEvent.TryGetValue(eventName, out var handlers))
 		{
-			public Subscription(WeakReference? subscriber, MethodInfo handler)
+			if (_enumeratingHandlersOfEvent == eventName)
 			{
-				Subscriber = subscriber;
-				Handler = handler ?? throw new ArgumentNullException(nameof(handler));
+				handlers = [..handlers];
```
Youssef1313, line 90: Is this copying to prevent the issue when a handler is mutating the list?

Youssef1313 (approved):
Quite scary class to touch, but I don't see leak concerns with the new impl. Just few comments.

I'd also love us to have this well-tested.

e.g (either runtime test or unit test should be fine, though my preference is always runtime test if no platform-specifics are involved):

```csharp
using System;
using Uno.UI.Helpers;

namespace Uno.UI.RuntimeTests.Tests.Uno_UI;

[TestClass]
public class Given_WeakEventManager
{
	private sealed class EventPublisher
	{
		private readonly WeakEventManager _manager = new();

		internal event Action Event
		{
			add => _manager.AddEventHandler(value);
			remove => _manager.RemoveEventHandler(value);
		}
	}

	private sealed class EventSubscriber
	{
		public void M() { }
	}

	[TestMethod]
	public void When_Many_Subscriptions_Should_Not_Leak()
	{
		var publisher = new EventPublisher();
		var weakRefs = new List<WeakReference<EventSubscriber>>();
		Subscribe(publisher, weakRefs);

		for (int i = 0; i < 10; i++)
		{
			GC.Collect();
			GC.WaitForPendingFinalizers();
		}

		foreach (var x in weakRefs)
		{
			Assert.IsFalse(x.TryGetTarget(out _));
		}
	}

	private void Subscribe(EventPublisher publisher, List<WeakReference<EventSubscriber>> weakRefs)
	{
		for (int i = 0; i < 1000; i++)
		{
			var subscriber = new EventSubscriber();
			publisher.Event += subscriber.M;
			weakRefs.Add(new WeakReference<EventSubscriber>(subscriber));
		}
	}
}

```

Maybe also have additional tests around a handler that tries to subscribe/unsubscribe, re-entrancy, etc

src/Uno.UI/Helpers/WeakEvents/WeakEventManager.cs
dr1rrb: By re-entrancy here I mean that we are not allowed to raise an event while we are already raising an event ... but actually re-thinking about it, it could be easily fixed without perf impact, will fix it ^^

src/Uno.UI/Helpers/WeakEvents/WeakEventManager.cs
dr1rrb: Yes is has been fixed in https://github.com/unoplatform/uno/pull/16562, that's why pointed me on that class. It's fixed in the add/remove handler: I clone the list at this time (so we will not pay the cost to clone handler's list on each raise unless it's effectively needed!)

src/Uno.UI/Helpers/WeakEvents/WeakEventManager.cs
dr1rrb: Yes :)

src/Uno.UI/Helpers/WeakEvents/WeakEventManager.cs
Youssef1313: Getting `null` shouldn't happen, so throwing is fine I think.

It's mostly an artifact that the `value` parameter is having the same type as the event. The nullability in the event type is mostly around when there are no subscribers so that `MyEvent.Invoke` doesn't crash. But whenever we do `+= Something` or `-= Something`, the `Something` is the value and should never be null, even if the type is nullable.

src/Uno.UI.RuntimeTests/Tests/Uno_UI_Helpers/Given_WeakEventManager.cs
Youssef1313, line 1: FYI there are some tests from MAUI [here](https://github.com/dotnet/maui/blob/main/src/Core/tests/UnitTests/WeakEventManagerTests.cs). I haven't looked at them, but just in case there is any interesting test scenario there.

src/Uno.UI/Helpers/WeakEvents/WeakEventManager.cs
```diff
@@ -3,133 +3,131 @@
...
+		if (string.IsNullOrEmpty(eventName))
 		{
-			if (string.IsNullOrEmpty(eventName))
-				throw new ArgumentNullException(nameof(eventName));
-
-			if (handler == null)
-				throw new ArgumentNullException(nameof(handler));
-
-			AddEventHandler(eventName, handler.Target, handler.GetMethodInfo());
+			throw new ArgumentNullException(nameof(eventName));
```
ebariche, line 21: We could use : https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception.throwifnull?view=net-8.0

src/Uno.UI/Helpers/WeakEvents/WeakEventManager.cs
```diff
@@ -3,133 +3,131 @@
...
-						target.Remove(subscription);
-					}
-					else
-					{
-						subscription.Handler.Invoke(subscriber, null);
-					}
-				}
-			}
+			// Allow handler to be nullable to not having to deal with nullability checks in callsites, but this indicates a bug.
+			Debug.Fail($"Got a null event handler for event {eventName}.");
```
ebariche, line 27: Debug.Assert() ?

src/Uno.UI/Helpers/WeakEvents/WeakEventManager.cs
```diff
@@ -3,133 +3,131 @@
...
+			AddTo(_handlersPerTarget.GetValue(target, static _ => new()), eventName, handler);
+		}
+	}
 
-			RemoveEventHandler(eventName, handler.Target, handler.GetMethodInfo());
+	public void RemoveEventHandler(Action? handler, [CallerMemberName] string eventName = "")
+	{
+		if (string.IsNullOrEmpty(eventName))
+		{
+			throw new ArgumentNullException(nameof(eventName));
```
ebariche, line 46: Same

src/Uno.UI/Helpers/WeakEvents/WeakEventManager.cs
```diff
@@ -3,133 +3,131 @@
...
-			{
-				targets = new List<Subscription>();
-				_eventHandlers.Add(eventName, targets);
-			}
-			else
-			{
-				targets.RemoveAll(subscription => subscription.Subscriber is { IsAlive: false });
-			}
+			// Allow handler to be nullable to not having to deal with nullability checks in callsites, but this indicates a bug.
+			Debug.Fail($"Got a null event handler for event {eventName}.");
```
ebariche, line 52: Same

src/Uno.UI/Helpers/WeakEvents/WeakEventManager.cs
```diff
@@ -3,133 +3,131 @@
...
+	{
+		if (handlersPerEvent.TryGetValue(eventName, out var handlers))
 		{
-			public Subscription(WeakReference? subscriber, MethodInfo handler)
+			if (_enumeratingHandlersOfEvent == eventName)
 			{
-				Subscriber = subscriber;
-				Handler = handler ?? throw new ArgumentNullException(nameof(handler));
+				handlers = [..handlers];
+				handlersPerEvent[eventName] = handlers;
```
ebariche, line 109: We could use : https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.collectionsmarshal.getvaluereforadddefault?view=net-8.0#system-runtime-interopservices-collectionsmarshal-getvaluereforadddefault-2(system-collections-generic-dictionary((-0-1))-0-system-boolean@)

src/Uno.UI/Helpers/WeakEvents/WeakEventManager.cs
```diff
@@ -3,133 +3,131 @@
...
 
-			public override bool Equals(object? obj) => obj is Subscription other && Equals(other);
+	private void RemoveFrom(Dictionary<string, List<Action>> handlersPerEvent, string eventName, Action handler)
+	{
+		if (handlersPerEvent.TryGetValue(eventName, out var handlers))
+		{
+			if (_enumeratingHandlersOfEvent == eventName)
+			{
+				handlers = [..handlers];
+				handlersPerEvent[eventName] = handlers;
```
ebariche, line 127: Same

src/Uno.UI/Helpers/WeakEvents/WeakEventManager.cs
dr1rrb: The assert is the condition above, if we reach this point we already failed!

src/Uno.UI/Helpers/WeakEvents/WeakEventManager.cs
dr1rrb: Cool didn't know about this one 👍

src/Uno.UI/Helpers/WeakEvents/WeakEventManager.cs
```diff
@@ -3,133 +3,126 @@
...
+		if (handler is null)
 		{
-			if (string.IsNullOrEmpty(eventName))
-				throw new ArgumentNullException(nameof(eventName));
-
-			if (handler == null)
-				throw new ArgumentNullException(nameof(handler));
-
-			AddEventHandler(eventName, handler.Target, handler.GetMethodInfo());
+			// Allow handler to be nullable to not having to deal with nullability checks in callsites, but this indicates a bug.
```
agneszitte, line 24:
```suggestion
			// Allow handler to be nullable to not have to deal with nullability checks in callsites, but this indicates a bug.
```

src/Uno.UI/Helpers/WeakEvents/WeakEventManager.cs
```diff
@@ -3,133 +3,126 @@
...
-			if (handler == null)
-				throw new ArgumentNullException(nameof(handler));
+	public void RemoveEventHandler(Action? handler, [CallerMemberName] string eventName = "")
+	{
+		ArgumentException.ThrowIfNullOrEmpty(eventName);
 
-			RemoveEventHandler(eventName, handler.Target, handler.GetMethodInfo());
+		if (handler is null)
+		{
+			// Allow handler to be nullable to not having to deal with nullability checks in callsites, but this indicates a bug.
```
agneszitte, line 46:
```suggestion
			// Allow handler to be nullable to not have to deal with nullability checks in callsites, but this indicates a bug.
```

MartinZikmund:
The failures here don't seem related to the changes, maybe just CI being funky

src/Uno.UI.RuntimeTests/Tests/Uno_UI_Helpers/Given_WeakEventManager.cs
```diff
@@ -0,0 +1,161 @@
...
+		public void RaiseEvent() => _manager.HandleEvent("Event");
+	}
+
+	private sealed class EventSubscriber
+	{
+		public void M() { }
+	}
+
+	[TestMethod]
+	public void When_ManySubscriptions_Then_DoesNotLeak()
```
Youssef1313, line 35: This is failing on iOS @dr1rrb

jeromelaban:
Superseeded by https://github.com/unoplatform/uno/pull/18899

PR #16754, opened 2024-05-16
fix: Support preserving DP values when type replacement happens in HR

### LINKED ISSUE #16723: [HR] Reloading a control does apply outer attached properties
### Current behavior

When changing a property in StatusBar.xaml in a debugging session, the StatusBar will expand at max height 

https://github.com/unoplatform/uno/assets/1608424/4f1b9105-2214-43f5-aadb-1d57c532f363

### Expected behavior

_No response_

1- Open the project [Test.zip](https://github.com/unoplatform/uno/files/15311010/Test.zip)
2- Start the debugging session
3- Open StatusBar.xaml then change line 12 to Background="Red" then click Save.

The whole StatusBar will goes max height.

### Workaround

_No response_

### Works on UWP/WinUI

Yes

### Environment

### NuGet package version(s)

_No response_

### Affected platforms

Skia (WPF)

### IDE

_No response_

### IDE version

Latest VS Code

### Relevant plugins

_No response_

_No response_

Youssef1313:
GitHub Issue (If applicable): closes #16723

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI.RuntimeTests/Tests/HotReload/Frame/HRApp/Tests/Given_TypeReplacement.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/HotReload/Frame/HRApp/Tests/Given_TypeReplacement.cs
@@ -1,0 +1,47 @@
+﻿using System.Reflection.Metadata;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Uno.Extensions;
+using Uno.UI;
+using Uno.UI.Extensions;
+using Uno.UI.Helpers;
+using Uno.UI.RuntimeTests.Tests.HotReload.Frame.HRApp.Tests;
+using Uno.UI.RuntimeTests.Tests.HotReload.Frame.Pages;
+
+namespace Uno.UI.RuntimeTests.Tests.HotReload.Frame.HRApp.Tests;
+
+[TestClass]
+[RunsOnUIThread]
+public class Given_TypeReplacement : BaseTestClass
+{
+	public const string OriginalColor = "Green";
+	public const string UpdatedColor = "Orange";
+
+	[TestMethod]
+	public async Task When_DP_And_AttachedDP_Set_On_Old_Instance()
+	{
+		var ct = new CancellationTokenSource(TimeSpan.FromSeconds(30)).Token;
+
+		var page = new HR_DPUpdates_PageUsingComponent();
+		UnitTestsUIContentHelper.Content = page;
+
+		UserControl oldComponent = page.myComponent;
+
+		await HotReloadHelper.UpdateServerFileAndRevert<HR_DPUpdates_Component>(
+			OriginalColor,
+			UpdatedColor,
+			() =>
+			{
+				var grid = (Grid)page.Content;
+				var component = (UserControl)grid.Children.Single();
+				Assert.AreNotEqual(oldComponent, component);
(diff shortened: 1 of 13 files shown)


src/Uno.UI.RuntimeTests/Tests/HotReload/Frame/HRApp/Tests/Given_TypeReplacement.cs
```diff
@@ -0,0 +1,77 @@
...
+			() =>
+			{
+				var grid = (Grid)page.Content;
+				var component = (UserControl)grid.Children.Single();
+				Assert.AreNotEqual(oldComponent, component);
+				Assert.IsTrue(component.GetType().Name.Contains("#"));
+				Assert.AreEqual(1, Grid.GetRow(component));
+				Assert.AreEqual("Hello tag", component.Tag as string);
+				// Visually, the color is updated, but this assert still fails with "Actual" being "Green" :/
+				//Assert.AreEqual(Microsoft.UI.Colors.Orange, ((SolidColorBrush)((Grid)component.Content).Background).Color);
```
jeromelaban, line 41: I think this we still need to fix this. @dr1rrb did you notice something similar during your tests?

Youssef1313: @jeromelaban This is probably an issue not relevant to the changes in this PR (i.e, an existing issues). So I think it doesn't block this PR from merge and could be looked at separately?

jeromelaban: The thing is that it does not validate the behavior properly, so it becomes relevant.

dr1rrb: Don't remember to have seen such issue in HR testing ... and I'm unable to find any good reason for this assert to fail:/

I supposed you already tried to yield the UI thread before the assertion? Otherwise in order to effectively test that changes occurred without fixing this (weird) issue (please open a new one), you can try to add a `TextBlock Text="Green"` in your component and validate that text changed ...

Youssef1313: > The thing is that it does not validate the behavior properly, so it becomes relevant.

I think the Tag assert and the other binding tests would fail so it still validates the fix to a good extent

Youssef1313: This is actually a regression, and currently there is no good way I can think of to fix this. We'll need to discuss more

dr1rrb: As discussed, I think we copy only the value with the `Local` precedence and only if the new instance has no value for this precedence. All other precedences are either "contextual" to where the control is being loaded or the current state of the control that should be rested by the control itself on load. The idea is only to preserve the changes made by the user since the original instance has been loaded but still applied the changes made in the xaml file.

jeromelaban (approved):

Issue #7002, opened 2021-09-08
LayoutUpdated event is not fired properly


dr1rrb:
### Current behavior

The `LayoutUpdated` event is fired only on elements that has been [re-]layouted

### Expected behavior

* It's like a static event that has to be fired on any `UIElement` associated to the current thread
* The sender of the event has to be `null`

More info: https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.frameworkelement.layoutupdated?view=winrt-20348#remarks

### How to reproduce it (as minimally and precisely as possible)

In a blank app, in the `MainPage` constructor, add
```cs
new Border().LayoutUpdated += (snd, e) => "".ToString(); 
```

The event handler is never invoked while it's on UWP!

### Workaround

* Don't rely on the sender like on UWP
* Subscribe to `Window.Current.Content.LayoutUpdated` to get notify also for element that are not in the visual tree.

### Works on UWP/WinUI

Yes

### Environment

Uno.UI / Uno.UI.WebAssembly / Uno.UI.Skia

### NuGet package version(s)

3.10.0-dev.632

### Affected platforms

iOS, Android, WebAssembly, WebAssembly renderers for Xamarin.Forms, macOS, Skia (WPF), Skia (GTK on Linux/macOS/Windows), Skia (Tizen)

### Anything else we need to know?

Didn't really validated on all platforms ... just an assumption based of current implementation.

jeromelaban:
Related to https://github.com/unoplatform/uno/issues/3519

Youssef1313:
Wrote some runtime tests for this to prepare after #14936.

```csharp
		[TestMethod]
		[RunsOnUIThread]
		public async Task When_LayoutUpdated()
		{
			string s = "";
			var button = new Button();
			button.LayoutUpdated += (sender, args) =>
			{
				Assert.IsNull(sender);
				Assert.IsNull(args);
				s += "button1 ";
			};
			button.LayoutUpdated += (sender, args) =>
			{
				Assert.IsNull(sender);
				Assert.IsNull(args);
				s += "button2 ";
			};

			var border = new Border
			{
				Width = 100,
				Height = 100,
				Background = new SolidColorBrush(Colors.Red),
			};

			border.LayoutUpdated += (sender, args) =>
			{
				Assert.IsNull(sender);
				Assert.IsNull(args);
				s += "border1 ";
			};

			border.LayoutUpdated += (sender, args) =>
			{
				Assert.IsNull(sender);
[...]

PR #16761, opened 2024-05-17
fix: Fix LayoutUpdated/SizeChanged on Skia and Wasm

Youssef1313:
GitHub Issue (If applicable): Skia and Wasm side of #7002

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_FrameworkElement_And_Leak.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_FrameworkElement_And_Leak.cs
@@ -213,7 +213,11 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml
 			};
 
 			var _holders = new ConditionalWeakTable<DependencyObject, Holder>();
-			void TrackDependencyObject(DependencyObject target) => _holders.Add(target, new Holder());
+			void TrackDependencyObject(DependencyObject target) => _holders.Add(target, new Holder(HolderUpdate));
+
+			var maxCounter = 0;
+			var activeControls = 0;
+			var maxActiveControls = 0;
 
 			// Ensure Holder counter is reset between individual control tests.
 			Holder.Reset();
@@ -226,24 +230,47 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml
 
 			for (int i = 0; i < count; i++)
 			{
-				await MaterializeControl(controlType, _holders, rootContainer);
+				await MaterializeControl(controlType, _holders, maxCounter, rootContainer);
 			}
 
 			TestServices.WindowHelper.WindowContent = null;
 
-			for (int i = 0; i < 10; i++)
+			void HolderUpdate(int value)
+			{
+				_ = rootContainer!.Dispatcher.RunAsync(CoreDispatcherPriority.High,
+					() =>
+					{
+						maxCounter = Math.Max(value, maxCounter);
+						activeControls = value;
+						maxActiveControls = maxCounter;
+					}
+				);
+			}
+
+			var sw = Stopwatch.StartNew();
+
+			var endTime = TimeSpan.FromSeconds(5);
+			var maxTime = TimeSpan.FromSeconds(30);
+
+			var lastActiveControls = activeControls;
+
+			while (sw.Elapsed < endTime && sw.Elapsed < maxTime && activeControls != 0)
 			{
 				GC.Collect();
 				GC.WaitForPendingFinalizers();
 
-				if (Holder.Counter == 0)
-				{
-					break;
-				}
-
 				// Waiting for idle is required for collection of
 				// DispatcherConditionalDisposable to be executed
 				await TestServices.WindowHelper.WaitForIdle();
+
+				if (lastActiveControls != activeControls)
+				{
+					// Expand the timeout if the count has changed, as the
+					// GC may still be processing levels of the hierarcy on iOS
+					endTime += TimeSpan.FromMilliseconds(500);
+				}
+
+				lastActiveControls = activeControls;
 			}
 
 #if TRACK_REFS
@@ -254,7 +281,7 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml
 			var retainedMessage = "";
 
 #if __IOS__ || __ANDROID__
-			if (Holder.Counter != 0)
+			if (activeControls != 0)
 			{
 				var retainedTypes = _holders.AsEnumerable().Select(ExtractTargetName).JoinBy(";");
 				Console.WriteLine($"Retained types: {retainedTypes}");
@@ -268,9 +295,9 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml
 			// to always go to zero during runtime tests. If the count of active objects
 			// is arbitrarily below the half of the number of top-level objects.
 			// created, we can assume that enough objects were collected entirely.
-			Assert.IsTrue(Holder.Counter < count, retainedMessage);
+			Assert.IsTrue(activeControls < count, retainedMessage);
 #else
-			Assert.AreEqual(0, Holder.Counter, retainedMessage);
+			Assert.AreEqual(0, activeControls, retainedMessage);
 #endif
 
 #if __IOS__ || __ANDROID__
@@ -287,7 +314,7 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml
 			}
 #endif
 
-			async Task MaterializeControl(Type controlType, ConditionalWeakTable<DependencyObject, Holder> _holders, ContentControl rootContainer)
+			async Task MaterializeControl(Type controlType, ConditionalWeakTable<DependencyObject, Holder> _holders, int maxCounter, ContentControl rootContainer)
 			{
 				var item = (FrameworkElement)Activator.CreateInstance(controlType)!;
 				TrackDependencyObject(item);
@@ -383,18 +410,19 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml
 
 		private class Holder
 		{
+			private readonly Action<int> _update;
 			private static int _counter;
 
-			internal static int Counter => _counter;
-
-			public Holder()
+			public Holder(Action<int> update)
 			{
-				Interlocked.Increment(ref _counter);
+				_update = update;
+				_update(++_counter);
 			}
 
 			~Holder()
 			{
-				Interlocked.Decrement(ref _counter);
+				var counter = Interlocked.Decrement(ref _counter);
+				_update(counter);
 			}
 
 			public static void Reset() => _counter = 0;
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_FrameworkElement_And_Leak.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_FrameworkElement_And_Leak.cs
@@ -254,7 +254,7 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml
 			var retainedMessage = "";
 
 #if __IOS__ || __ANDROID__
-			if (activeControls != 0)
+			if (Holder.Counter != 0)
 			{
 				var retainedTypes = _holders.AsEnumerable().Select(ExtractTargetName).JoinBy(";");
 				Console.WriteLine($"Retained types: {retainedTypes}");
@@ -268,7 +268,7 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml
 			// to always go to zero during runtime tests. If the count of active objects
 			// is arbitrarily below the half of the number of top-level objects.
 			// created, we can assume that enough objects were collected entirely.
-			Assert.IsTrue(activeControls < count, retainedMessage);
+			Assert.IsTrue(Holder.Counter < count, retainedMessage);
 #else
 			Assert.AreEqual(0, Holder.Counter, retainedMessage);
 #endif
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_FrameworkElement_And_Leak.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_FrameworkElement_And_Leak.cs
@@ -177,7 +177,7 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml
 			LeakTestStyles.All
 #endif
 			)]
-		[DataRow(typeof(MediaPlayerElement), 15)] // Commented out temporarily.
+		[DataRow(typeof(MediaPlayerElement), 15)]
 		public async Task When_Add_Remove(object controlTypeRaw, int count, LeakTestStyles leakTestStyles = LeakTestStyles.All)
 		{
 			if (leakTestStyles.HasFlag(LeakTestStyles.Default))
@@ -213,11 +213,7 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml
 			};
 
 			var _holders = new ConditionalWeakTable<DependencyObject, Holder>();
-			void TrackDependencyObject(DependencyObject target) => _holders.Add(target, new Holder(HolderUpdate));
-
-			var maxCounter = 0;
-			var activeControls = 0;
-			var maxActiveControls = 0;
+			void TrackDependencyObject(DependencyObject target) => _holders.Add(target, new Holder());
 
 			// Ensure Holder counter is reset between individual control tests.
 			Holder.Reset();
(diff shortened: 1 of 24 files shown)


src/Uno.UI/UI/Xaml/FrameworkElement.cs
```diff
@@ -180,10 +181,17 @@ internal void RaiseSizeChanged(SizeChangedEventArgs args)
...
 			SizeChanged?.Invoke(this, args);
+#if !UNO_HAS_ENHANCED_LIFECYCLE
 			_renderTransform?.UpdateSize(args.NewSize);
+#endif
 #endif
 		}
 
+#if UNO_HAS_ENHANCED_LIFECYCLE
+		internal void UpdateRenderTransformSize(Size newSize)
+			=> _renderTransform?.UpdateSize(newSize);
```
dr1rrb, line 192: FYI I think this is actually useless on skia (and it should be) ... we can keep it if we want to keep the contract of `NativeRenderTransformAdapter` unified between platforms, but I think it would make sense to update it also.

src/Uno.UI/UI/Xaml/FrameworkElement.cs
```diff
@@ -752,13 +760,77 @@ internal override bool IsViewHit()
...
+				while (true)
+				{
+					EventHandler<object> eventHandler2 = eventHandler;
+					EventHandler<object> value2 = (EventHandler<object>)Delegate.Combine(eventHandler2, value);
+					eventHandler = Interlocked.CompareExchange(ref _layoutUpdated, value2, eventHandler2);
+					if ((object)eventHandler == eventHandler2)
+					{
+						break;
+					}
+				}
```
dr1rrb, line 786: You are never re-evaluating the `_layoutUpdated` in case of conflict, so you will never exit the loop.

BTW Is this from WinUI (which would explain the bug)? Otherwise the common names for such scenario are `original` and `updated`, suffixing with number is really not helpful.

```suggestion
				while (true)
				{
					var original = _layoutUpdated;
					var updated = (EventHandler<object>)Delegate.Combine(original, value);
					if (Interlocked.CompareExchange(ref _layoutUpdated, updated, original) == original)
					{
						break;
					}
				}
```

src/Uno.UI/UI/Xaml/FrameworkElement.cs
```diff
@@ -752,13 +760,77 @@ internal override bool IsViewHit()
...
+				while (true)
+				{
+					EventHandler<object> eventHandler2 = eventHandler;
+					EventHandler<object> value2 = (EventHandler<object>)Delegate.Remove(eventHandler2, value);
+					eventHandler = Interlocked.CompareExchange(ref _layoutUpdated, value2, eventHandler2);
+					if ((object)eventHandler == eventHandler2)
+					{
+						break;
+					}
+				}
```
dr1rrb, line 811: Same here:

```suggestion
				while (true)
				{
					var original = _layoutUpdated;
					var updated = (EventHandler<object>)Delegate.Remove(original, value);
					if (original == updated // `value` was not in the invocation list
						|| Interlocked.CompareExchange(ref _layoutUpdated, updated, original) == original)
					{
						break;
					}
				}
```

src/Uno.UI/UI/Xaml/Internal/EventManager.uno.cs
```diff
@@ -0,0 +1,140 @@
...
+			=> ReferenceEquals(Element, other.Element) && OldSize == other.OldSize;
+
+		public override int GetHashCode()
+			=> HashCode.Combine(Element, OldSize);
+
+		public static bool operator ==(SizeChangedQueueItem left, SizeChangedQueueItem right)
+			=> left.Equals(right);
+
+		public static bool operator !=(SizeChangedQueueItem left, SizeChangedQueueItem right)
+			=> !(left == right);
```
dr1rrb, line 50: Why not using a `record` here ?

src/Uno.UI/UI/Xaml/Internal/EventManager.uno.cs
```diff
@@ -0,0 +1,140 @@
...
+		_layoutUpdatedSubscribers.Remove(element);
+	}
+
+	public void RaiseLayoutUpdated()
+	{
+		var clone = _layoutUpdatedSubscribers.ToArray();
+		for (int i = 0; i < clone.Length; i++)
+		{
+			clone[i].Key.OnLayoutUpdated();
+		}
```
dr1rrb, line 69: The `CollectionWeakTable` is [thread safe](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.conditionalweaktable-2?view=net-8.0#thread-safety), including [the enumeration](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.conditionalweaktable-2.system-collections-generic-ienumerable-system-collections-generic-keyvaluepair-tkey-tvalue---getenumerator?view=net-8.0#remarks), you don't have to clone the collection (which should be avoided as this is a hot-path).

```suggestion
		foreach(var subscriber in _layoutUpdatedSubscribers)
		{
			subscriber.Key.OnLayoutUpdated();
		}
```

src/Uno.UI/UI/Xaml/Controls/Border/BorderLayerRenderer.cs
```diff
@@ -29,7 +29,11 @@ public BorderLayerRenderer(FrameworkElement owner)
 
 		_owner.Loaded += (s, e) => Update();
 		_owner.Unloaded += (s, e) => Clear();
-		_owner.LayoutUpdated += (s, e) => Update();
+#if UNO_HAS_ENHANCED_LIFECYCLE
+		_owner.SizeChanged += (_, _) => Update();
+#else
+		_owner.LayoutUpdated += (_, _) => Update();
+#endif
```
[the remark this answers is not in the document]
Youssef1313: It should, but didn't work on Android

src/Uno.UI/UI/Xaml/FrameworkElement.Layout.crossruntime.cs
```diff
@@ -350,9 +351,22 @@ private void InnerMeasureCore(Size availableSize)
...
+
 			// DesiredSize must include margins
 			LayoutInformation.SetDesiredSize(this, desiredSize);
 
+			var newWidth = this.GetActualWidth();
+			var newHeight = this.GetActualHeight();
+
+			if (oldWidth != newWidth || oldHeight != newHeight)
+			{
+				// This is HACKY, and is specific to TextBlock which overrides GetActualWidth and GetActualHeight to return DesiredSize.
```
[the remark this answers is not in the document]
Youssef1313: @dr1rrb We'are comparing `Actual[Width|Height]` after a SetDesiredSize call.

For all elements, DesiredSize isn't going to change `Actual[Width|Height]`. However, for `TextBlock`, it's the case because of:

https://github.com/unoplatform/uno/blob/30014dd5103817fc401eff20a913cb4c3cba4b22/src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.cs#L1216-L1217

src/Uno.UI/UI/Xaml/FrameworkElement.cs
Youssef1313: @dr1rrb This is just the decompiled code when you have a regular field-like event in C#. https://sharplab.io/#v2:EYLgtghglgdgNAExAagD4AEBMBGAsAKHQGYACLEgYRIG8CT6zSBTANyZgBcSBRNzgCQgwEAGyYAnADwB7YACsmAYw4A+EgBkIAT2kBXDgFUADgggcmCANwEAvkA=

src/Uno.UI/Microsoft/UI/Xaml/Controls/ProgressBar/ProgressBar.cs
```diff
@@ -69,7 +69,13 @@ public ProgressBar()
 
 			SizeChanged += (snd, evt) => OnSizeChange();
 
+#if !UNO_HAS_ENHANCED_LIFECYCLE
+			// Uno-specific: TODO: Investigate why we need this. It's a quite very old workaround from 2020
+			// https://github.com/unoplatform/uno/commit/641bbc9483f33c64d5eddc474f069c07b79039ba
+			// So, maybe it's no longer needed.
+			// For now, we're sure it's no longer needed with enhanced lifecycle (actually, it's problematic if it exists there)
```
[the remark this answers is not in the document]
Youssef1313: Here: https://github.com/unoplatform/uno/blob/2dbe845645483db4416d0ab9f5eedb5998ab456d/src/SamplesApp/UITests.Shared/Microsoft_UI_Xaml_Controls/ProgressBar/WinUIProgressBarPage.xaml.cs#L48

This sample listens to visual state change and updates the text

So, the ProgressBar goes into Indeterminate, causing TextBlock.Text to change which will cause LayoutUpdated, then `OnSizeChange` will cause a visual state to change from Indeterminate to Updating, etc. So, you go into a loop keeping dancing between Indeterminate and Updating states

src/Uno.UI/UI/Xaml/FrameworkElement.cs
Youssef1313: Why is it useless on Skia?

src/Uno.UI/UI/Xaml/Internal/EventManager.uno.cs
Youssef1313: This is not about thread-safety as it will happen on the UI thread anyways. It's more about re-entrancy.

src/Uno.UI/UI/Xaml/Internal/EventManager.uno.cs
Youssef1313: Records will use `IEqualityComparer<T>.Default`. In case of `Size` we don't implement `IEquatable<T>` so I think the default comparer will be `ObjectEqualityComparer` which will probably end up going through .NET's default value type equality which uses reflection (and some boxing will probably happen as well)

src/Uno.UI/Microsoft/UI/Xaml/Controls/ProgressBar/ProgressBar.cs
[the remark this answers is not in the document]
Youssef1313: And it's probably worse than just the "cycle". It will really mean that any change anywhere in the visual tree will call `OnSizeChange` which will cause visual state transition to Updating that will likely mess up with ProgressBar and cause it to flicker.

src/Uno.UI/UI/Xaml/Internal/EventManager.uno.cs
Youssef1313: For reference: https://github.com/microsoft/microsoft-ui-xaml/blob/2a60e27c591846556fa9ec4d8f305afdf0f96dc1/dxaml/xcp/dxaml/lib/DXamlCore.cpp#L2179-L2215

src/Uno.UI/UI/Xaml/FrameworkElement.cs
Youssef1313: The whole idea behind the way the compiler generates this is thread-safety. I'm not sure if subscribing/unsubscribing to LayoutUpdated requires the UI thread in WinUI, in which case we can simplify this code. But for now, I preferred to keep it matching with what the compiler is lowering to.

src/Uno.UI/UI/Xaml/FrameworkElement.Layout.crossruntime.cs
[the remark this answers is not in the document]
dr1rrb: Ok ... but why we need to raise the event twice? Is it really expected?

src/Uno.UI/Microsoft/UI/Xaml/Controls/ProgressBar/ProgressBar.cs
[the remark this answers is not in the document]
dr1rrb: Would be great to have few details about this in the comment :)

src/Uno.UI/UI/Xaml/FrameworkElement.cs
dr1rrb: Oh indeed on line 781 the original `eventHandler` is being updated ... that why names matter! If it's from de-compiled *-**generated** code (so not from winUI) we need to give real names to variables.

And if it's only about replicates what's done by compiler, you can simply convert your private field `_layoutUpdated` to an `event` and do the `+=` and `-=` on it.

src/Uno.UI/UI/Xaml/Internal/EventManager.uno.cs
dr1rrb: As the `Size` is a `struct` I think it will instead compare its fields directly without any boxing nor reflection ... but I'm not sure neither so it makes sense to implement it by your own. Would be nice to add some comments to explain why you decided to implement the `IEquatable` here then ^^

src/Uno.UI/UI/Xaml/Internal/EventManager.uno.cs
jeromelaban: Could we use the array pool for that? This is likely going to be called a lot.

dr1rrb: Indeed I can see problem if you remove and then add back an handler for instance ... again some comments would help :)

(I still consider that we can avoid to clone the collection on something that can be considered as a hot-path, but if it replicates the code from winUI it's fine and we will change it only if we detect a performance hit)

src/Uno.UI/UI/Xaml/FrameworkElement.cs
dr1rrb: Well I was assuming that all `Transform` info were pushed to the `Visual` and they will be applied during the painting / rendering using the size of the `Visual` itself. But actually looking at the implementation of the `NativeRenderTransformAdapter`(https://github.com/unoplatform/uno/blob/master/src/Uno.UI/Media/NativeRenderTransformAdapter.skia.cs#L28C60-L28C68) we are computing the matrix using the `CurrentSize`... which is probably wrong but not the subject of this PR!

src/Uno.UI/UI/Xaml/FrameworkElement.cs
Youssef1313: I'm plugging extra logic so can't use a regular event. Will rename variables

src/Uno.UI/UI/Xaml/Internal/EventManager.uno.cs
Youssef1313: Yeah no reflection will happen as `Size` override `Equals`. But I think boxing will still happen because the `ObjectEqualityComparer` will end up calling `object.Equals(object)` method. I'll comment on this to clarify

src/Uno.UI/UI/Xaml/Internal/EventManager.uno.cs
Youssef1313: @jeromelaban I don't see a way to use array pool as ConditionalWeakTable doesn't provide a `Count` or `Length` property :/

We could go with counting it ourselves though, I think.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_FrameworkElement_And_Leak.cs
```diff
@@ -230,47 +226,24 @@ Type GetType(string s)
...
+			for (int i = 0; i < 10; i++)
 			{
 				GC.Collect();
 				GC.WaitForPendingFinalizers();
 
-				// Waiting for idle is required for collection of
-				// DispatcherConditionalDisposable to be executed
-				await TestServices.WindowHelper.WaitForIdle();
-
-				if (lastActiveControls != activeControls)
```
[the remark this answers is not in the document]
Youssef1313: I think it was passing on iOS CI just fine. So it's probably something that may have changed in .NET at some point?

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_ContentPresenter.cs
```diff
@@ -317,24 +319,22 @@ WeakReference SetContent()
 		}
 	}
 
-	private async Task AssertCollectedReference(WeakReference reference)
+	private void AssertCollectedReference(WeakReference reference)
```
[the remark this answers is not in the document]
Youssef1313: It was passing on CI

src/Uno.UI/UI/Xaml/Internal/EventManager.uno.cs
Youssef1313: Actually, adding elements while iterating isn't a problem. `ConditionalWeakTable` really does what we need and re-entrancy won't be an issue.

src/Uno.UI/Microsoft/UI/Xaml/Controls/ProgressBar/ProgressBar.cs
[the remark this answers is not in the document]
Youssef1313: I extended the comment so it's clearer what the problem is.

src/Uno.UI/UI/Xaml/Internal/EventManager.uno.cs
Youssef1313: Added a comment as to why not using a record.

src/Uno.UI/UI/Xaml/Internal/EventManager.uno.cs
Youssef1313: Fixed now

src/Uno.UI/UI/Xaml/FrameworkElement.cs
dr1rrb: But you can still use a private event:, eg: 

``` csharp
		private event EventHandler<object> _layoutUpdated;

		public event EventHandler<object> LayoutUpdated
		{
			add
			{
#if UNO_HAS_ENHANCED_LIFECYCLE
				var isFirstSubscriber = _layoutUpdated is null;
#endif
				_layoutUpdated += value;
#if UNO_HAS_ENHANCED_LIFECYCLE
				if (isFirstSubscriber)
				{
					Uno.UI.Extensions.DependencyObjectExtensions.GetContext(this).EventManager.AddLayoutUpdatedEventHandler(this);
				}
#endif
			}
```

Youssef1313: Ah, yes. Indeed. Will do it that way to make the code cleaner.

src/Uno.UI/UI/Xaml/FrameworkElement.Layout.crossruntime.cs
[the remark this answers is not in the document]
Youssef1313: It shouldn't be raising it twice, but thinking more, it will be raising it once in a case where it shouldn't have been raised.

It looks like currently we don't have any good event to listen for here:

https://github.com/unoplatform/uno/blob/519c422535527d70deda85e391b3a70b360c9481/src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.skia.cs#L202

Will investigate more next week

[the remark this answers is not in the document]
Youssef1313: Fixed

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_FrameworkElement_And_Leak.cs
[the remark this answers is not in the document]
Youssef1313: This one started to fail consistently in CI, so reverted. Let's see if it will be passing on all platforms.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_ContentPresenter.cs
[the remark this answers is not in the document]
Youssef1313: Reverted, though this one didn't seem to cause any issues for iOS CI.

On Skia though, an extra WaitForIdle was needed. I didn't put time to understand why it's, as the test fails on WinUI as well with the way it's previously written.

MartinZikmund (approved):

PR #16772, opened 2024-05-17
feat: Native Element Hosting for X11 + airpspace for X11, WPF and GTK

ramezgerges:
GitHub Issue (If applicable): closes https://github.com/unoplatform/uno-private/issues/295

## PR Type

## PR Checklist

## Other information


src/Uno.UI/UI/Xaml/UIElement.skia.cs
```diff
@@ -60,9 +60,31 @@ public bool UseLayoutRounding
 				typeof(UIElement),
 				new FrameworkPropertyMetadata(true, propertyChangedCallback: (o, _) => (o as IBorderInfoProvider)?.UpdateBorderThickness()));
 
+		/// <summary>
+		/// Represents the final calculated opacity of the element.
+		/// </summary>
+		/// <remarks>
+		/// This property should never be directly set, and its value should always be calculated through coercion (see <see cref="CoerceHitTestVisibility(DependencyObject, object, bool)"/>.
+		/// </remarks>
+		[GeneratedDependencyProperty(DefaultValue = 1.0, Options = FrameworkPropertyMetadataOptions.AffectsRender | FrameworkPropertyMetadataOptions.Inherits, CoerceCallback = true, ChangedCallback = false)]
```
Youssef1313, line 69: Not a fan of increasing coercion usages.

src/Uno.UI/UI/Xaml/UIElement.skia.cs
```diff
@@ -60,9 +60,31 @@ public bool UseLayoutRounding
 				typeof(UIElement),
 				new FrameworkPropertyMetadata(true, propertyChangedCallback: (o, _) => (o as IBorderInfoProvider)?.UpdateBorderThickness()));
 
+		/// <summary>
+		/// Represents the final calculated opacity of the element.
+		/// </summary>
+		/// <remarks>
+		/// This property should never be directly set, and its value should always be calculated through coercion (see <see cref="CoerceHitTestVisibility(DependencyObject, object, bool)"/>.
```
Youssef1313, line 67: Looks like a copy/paste error

src/Uno.UI/UI/Xaml/UIElement.skia.cs
```diff
@@ -60,9 +60,31 @@ public bool UseLayoutRounding
...
+
+		internal double CalculatedOpacity
+		{
+			get => GetCalculatedOpacityValue();
+			set => SetCalculatedOpacityValue(value);
+		}
+
+		private object CoerceCalculatedOpacity(object baseValue)
+		{
+			// The HitTestVisibilityProperty is never set directly. This means that baseValue is always the result of the parent's CoerceCalculatedOpacity.
```
Youssef1313, line 80: Looks like a copy/paste error

src/Uno.UI.Runtime.Skia/X11NativeWindow.cs
```diff
@@ -0,0 +1,11 @@
+namespace Uno.UI.Runtime.Skia;
+
+public class X11NativeWindow
```
Youssef1313, line 3:
```suggestion
public sealed class X11NativeWindow
```

src/Uno.UI.Runtime.Skia/X11NativeWindow.cs
```diff
@@ -0,0 +1,11 @@
...
+{
+	public IntPtr WindowId { get; private set; }
```
Youssef1313, line 5:
```suggestion
	public IntPtr WindowId { get; }
```

src/Uno.UI.Runtime.Skia.X11/X11_Bindings/X11Helper.cs
```diff
@@ -334,6 +371,59 @@ private unsafe static void SetMotifWMHints(X11Window x11Window, bool on, IntPtr?
...
+		rectangle.X = x;
+		rectangle.Y = y;
+		rectangle.W = w;
+		rectangle.H = h;
+		var _ = XUnionRectWithRegion(&rectangle, region, region);
+
+		return region;
+	}
+
+	[DllImport(libX11Randr)]
```
Youssef1313, line 400:
```suggestion
	[LibraryImport(libX11Randr)]
```

src/Uno.UI.Runtime.Skia.X11/X11_Bindings/X11Helper.cs
```diff
@@ -313,6 +328,28 @@ private unsafe static void SetMotifWMHints(X11Window x11Window, bool on, IntPtr?
 	public static partial IntPtr XCreateImage(IntPtr display, IntPtr visual, uint depth, int format, int offset,
 		IntPtr data, uint width, uint height, int bitmap_pad, int bytes_per_line);
 
+	[LibraryImport(libX11)]
+	public static partial int XClearWindow(IntPtr display, IntPtr window);
+
+	[DllImport(libX11)]
```
Youssef1313, line 334:
```suggestion
	[LibraryImport(libX11)]
```

src/Uno.UI.Composition/Composition/Visual.skia.cs
```diff
@@ -162,7 +193,7 @@ internal void RenderRootVisual(SKSurface surface, Vector2? offsetOverride = null
 	/// Position a sub visual on the canvas and draw its content.
 	/// </summary>
 	/// <param name="parentSession">The drawing session of the <see cref="Parent"/> visual.</param>
-	private void Render(in PaintingSession parentSession)
+	private void Render(in PaintingSession parentSession, bool isFlyoutSurface)
```
Youssef1313, line 196: Should this flag be part of the PaintingSession?

src/Uno.UI/UI/Xaml/UIElement.skia.cs
Youssef1313: As it's not a very common scenario, I'd keep a static `List<WeakReference>` of `ContentPresenter` that are used for native embedding, then, on *any* opacity update in the visual tree, calculate the opacity for things that are in the list. This is beneficial IMO because:

1. Avoids coercion, which ideally should be removed completely.
2. Moving the logic to where it really belongs, which is ContentPresenter. The CalculatedOpacity is really not relevant to UIElement, which also avoids increasing the size of `UIElement`s unnecessarily (even though the increase isn't a big deal)
3. Avoids the calculation of opacity in many cases where it's actually irrelevant/unused
4. Opacity isn't something that will change very often, esp with native embedding, so the extra cost of re-calculating the `List<WeakReference>` isn't a big deal.

src/Uno.UI/UI/Xaml/UIElement.skia.cs
```diff
@@ -60,9 +60,31 @@ public bool UseLayoutRounding
...
+			return baseValue is double d1 ? d1 * Opacity : Opacity;
+		}
+
 		partial void OnOpacityChanged(DependencyPropertyChangedEventArgs args)
 		{
 			UpdateOpacity();
+			CalculatedOpacity = (double)args.NewValue;
```
Youssef1313, line 87: There is a comment suggesting that this property is never set directly, but it's actually set directly here.

src/Uno.UI.Composition/Composition/Visual.skia.cs
ramezgerges: For now, probably not. Right now, all the "flyout surface" logic is only visible in Visual. Until we have a very clear reason for why a subclass would need to know about whether the current drawing session is for flyouts or not, let's keep the "layering" details in Visual. This follows the recent changes that pack all the rendering (not painting) logic in Visual directly.

src/Uno.UI.Composition/Composition/Visual.skia.cs
```diff
@@ -22,6 +22,37 @@ public partial class Visual : global::Microsoft.UI.Composition.CompositionObject
 	private bool _matrixDirty = true;
 	private Matrix4x4 _totalMatrix = Matrix4x4.Identity;
 
+	// a visual is a flyout visual if it's directly set by SetAsFlyoutVisual or is a child of a flyout visual
+	// This doesn't yet handle the case of a child visual changing parents. i.e., once a flyout visual, always a flyout visual.
```
ramezgerges, line 26: note to self: this comment should be removed since it's already fixed with `_isFlyoutVisualInherited`

src/Uno.UI.Composition/Composition/Visual.skia.cs
Youssef1313: I'm not sure though if this can introduce perf regression given how deep the recursion can go. I *think* the whole idea of having `PaintingSession` passed as `in` is to avoid copying individual fields. But introducing a parameter this way will have an extra copy, which kind-of defeats the purpose.

src/Uno.UI.Runtime.Skia.X11/X11_Bindings/X11Helper.cs
ramezgerges: DllImport handles string and bool marshalling automatically, so I'm leaving a couple of imports as DllImports.

src/Uno.UI.Composition/Composition/Visual.skia.cs
ramezgerges: I'm not particularly convinced that this would cause any perf regressions. Let's keep it this way until we have a clear example where it does.

src/Uno.UI.Composition/Composition/Visual.cs
```diff
@@ -24,6 +24,7 @@ public partial class Visual : CompositionObject, I3DTransformableObject
 		private float _opacity = 1.0f;
 		private CompositionCompositeMode _compositeMode;
 		private ICompositionTarget? _compositionTarget;
+		private ContainerVisual? _parent;
```
jeromelaban, line 27: This may cause memory leaks, particularly if there are events involved. We may need to validate visuals collections.

src/Uno.UI.Composition/Composition/Visual.skia.cs
jeromelaban: I agree with moving this to the PaintingSession. There's no point in adding additional context here, particularly when it's created at the same time as the PaintingSession instance.

Also, shouldn't this be using popup-related wording? A flyout is a popup.

src/Uno.UI.Runtime.Skia.Gtk/Extensions/GtkNativeElementHostingExtension.cs
```diff
@@ -1,100 +1,200 @@
...
+
+		var button1 = new Button("Button 1");
+		var button2 = new Button("Button 2");
+		hbox.Add(button1);
+		hbox.Add(button2);
+
+		vbox.PackStart(hbox, false, false, 0);
+
+		return vbox;
+	}
```
jeromelaban, line 87: Test code that should be removed? If not, let's move to a separate class.

src/Uno.UI.Runtime.Skia.MacOS/MacSkiaHost.cs
```diff
@@ -25,7 +25,6 @@ static MacSkiaHost()
 		MacOSFileSavePickerExtension.Register();
 		MacOSFolderPickerExtension.Register();
 		MacOSLauncherExtension.Register();
-		MacOSNativeElementHostingExtension.Register();  // seems used only for native elements, which Skia avoids ?!?
 		MacOSNativeWindowFactoryExtension.Register();
```
jeromelaban, line 28: We'll need to have this PR testsed on mac

src/Uno.UI.Runtime.Skia.Wpf/Extensions/WpfNativeElementHostingExtension.cs
```diff
@@ -50,39 +55,85 @@ public void DetachNativeElement(object owner, object content)
...
+		if (content is System.Windows.UIElement contentAsUIElement)
+		{
+			contentAsUIElement.Measure(new System.Windows.Size(availableSize.Width, availableSize.Height));
+			return new Windows.Foundation.Size(contentAsUIElement.DesiredSize.Width, contentAsUIElement.DesiredSize.Height);
+		}
+
+		return Windows.Foundation.Size.Empty;
+	}
+
+	public object CreateSampleComponent(string text)
```
jeromelaban, line 111: Let's move to a separate class, if possible.

src/Uno.UI.Runtime.Skia.Wpf/Extensions/WpfNativeElementHostingExtension.cs
```diff
@@ -50,39 +55,85 @@ public void DetachNativeElement(object owner, object content)
...
-			&& GetOverlayLayer(xamlRoot) is { } layer
-			&& contentAsFE.Parent == layer;
+	public void ChangeNativeElementVisibility(object content, bool visible)
+	{
+		if (content is System.Windows.UIElement contentAsUIElement)
+		{
+			contentAsUIElement.Visibility = visible ? System.Windows.Visibility.Visible : System.Windows.Visibility.Collapsed;
+		}
+	}
+	public void ChangeNativeElementOpacity(object content, double opacity)
```
jeromelaban, line 70:
```suggestion
	}
	
	public void ChangeNativeElementOpacity(object content, double opacity)
```

src/Uno.UI.Runtime.Skia.X11/X11NativeElementHostingExtension.cs
```diff
@@ -0,0 +1,285 @@
...
+					{
+						return true;
+					}
+				}
+			}
+		}
+
+		return FindWindowById(_display, nativeWindow.WindowId, root) != IntPtr.Zero;
+	}
+	public void AttachNativeElement(object content)
```
jeromelaban, line 100:
```suggestion
	}
	
	public void AttachNativeElement(object content)
```

src/Uno.UI.Runtime.Skia.X11/X11NativeElementHostingExtension.cs
```diff
@@ -0,0 +1,285 @@
...
+			XWindowAttributes attributes = default;
+			var _2 = XLib.XGetWindowAttributes(_display, nativeWindow.WindowId, ref attributes);
+			attributes.override_direct = /* True */ 1;
+
+			unsafe
+			{
+				IntPtr attr = Marshal.AllocHGlobal(Marshal.SizeOf(attributes));
+				Marshal.StructureToPtr(attributes, attr, false);
+				var _3 = X11Helper.XChangeWindowAttributes(_display, nativeWindow.WindowId, (IntPtr)XCreateWindowFlags.CWOverrideRedirect, (XSetWindowAttributes*)attr.ToPointer());
+				Marshal.FreeHGlobal(attr);
```
jeromelaban, line 118: This should be in a finally

src/Uno.UI.Runtime.Skia.X11/X11NativeElementHostingExtension.cs
```diff
@@ -0,0 +1,285 @@
...
+	}
+
+	public void DetachNativeElement(object content)
+	{
+		if (content is X11NativeWindow nativeWindow
+			&& XamlRoot is { } xamlRoot
+			&& X11Manager.XamlRootMap.GetHostForRoot(xamlRoot) is X11XamlRootHost host)
+		{
+			using var _1 = X11Helper.XLock(_display);
+			var _2 = XLib.XQueryTree(_display, nativeWindow.WindowId, out IntPtr root, out _, out var children, out _);
```
jeromelaban, line 148: You should be able to use `_` directly, afaik. The only scenario scenario that needs `_1` or similar is with using. Still, it feels like the usings "fake discards" should be named to avoid having to renumber...

src/Uno.UI.Runtime.Skia.X11/X11NativeElementHostingExtension.helper.cs
```diff
@@ -0,0 +1,189 @@
...
+// Thanks to Jörg Seebohn for providing an example on how to use X SHAPE
+// https://gist.github.com/je-so/903479/834dfd78705b16ec5f7bbd10925980ace4049e17
+internal partial class X11NativeElementHostingExtension : ContentPresenter.INativeElementHostingExtension
+{
+	private const string SampleVideoLink = "https://uno-assets.platform.uno/tests/uno/big_buck_bunny_720p_5mb.mp4";
+
+	/// replace the executable and the args with whatever you have locally. This is only used
+	/// for internal debugging. However, make sure that you can set a unique title to the window,
+	/// so that you can then look it up.
+	public object? CreateSampleComponent(string text)
```
jeromelaban, line 21: Same here, let's move to a separate class

src/Uno.UI.Runtime.Skia.X11/X11_Bindings/X11Helper.cs
jeromelaban: Let's indicate that as a comment

src/Uno.UI/UI/Xaml/Controls/ContentPresenter/ContentPresenter.skia.cs
```diff
@@ -46,96 +47,127 @@ partial class ContentPresenter
...
+			Content,
+			arrangeRect,
+			clippingBounds);
 	}
 
-	partial void ArrangeNativeElement(Rect arrangeRect)
+	partial void AttachNativeElement()
 	{
-		if (IsNativeHost)
+#if DEBUG
```
jeromelaban, line 99: Debug.Assert is already done only in debug.

src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/ContentPresenter/ContentPresenter_NativeEmbedding.xaml.cs
```diff
@@ -0,0 +1,54 @@
...
+	[Sample("ContentPresenter", Name = "ContentPresenter_NativeEmbedding")]
+	public sealed partial class ContentPresenter_NativeEmbedding : UserControl
+	{
+		public ContentPresenter_NativeEmbedding()
+		{
+			this.InitializeComponent();
+		}
+	}
+
+	public sealed partial class NativeControlHostWithText : ContentControl
```
jeromelaban, line 17: We'll need to document how to create nuget packages that can use the airspace integration.

src/Uno.UI.Composition/Composition/Visual.cs
ramezgerges: I'm not sure what you're referring to. `Parent` was an auto property already. I just added a backing field so that  I can add additional logic in  the setter.

src/Uno.UI.Runtime.Skia.X11/X11NativeElementHostingExtension.cs
ramezgerges: It shouldn't matter since `XChangeWindowAttributes` doesn't throw exceptions. If the call is incorrect for some reason, we will just crash.

src/Uno.UI/UI/Xaml/Controls/ContentPresenter/ContentPresenter.skia.cs
ramezgerges: The `#if DEBUG` is for the `_nativeElementAttached` assignment mainly.

src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/ContentPresenter/ContentPresenter_NativeEmbedding.xaml.cs
ramezgerges: Let's do that in a separate PR after testing it thoroughly. We may need to change the API a little bit and/or notice some specific points that need attention in the docs.

src/Uno.UI.Runtime.Skia.Gtk/Extensions/GtkNativeElementHostingExtension.cs
ramezgerges: It's "test code" but we need to keep it for debugging. I would much rather keep everything tight in one class instead of creating a new interface and registering it with `ApiExtensibility` (especially since for X11, this method needs the added context of the specific ContentPresenter). I moved these methods to a separate partial for now.

src/Uno.UI.Composition/Composition/Visual.cs
jeromelaban: If it was there, no problem then.

src/Uno.UI.Runtime.Skia.Gtk/Extensions/GtkNativeElementHostingExtension.cs
jeromelaban: Separate partial is fine. Thanks.

src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/ContentPresenter/ContentPresenter_NativeEmbedding.xaml
```diff
@@ -0,0 +1,47 @@
...
+			 xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+			 xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+			 xmlns:not_skia="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+			 xmlns:skia="http://uno.ui/skia"
+			 mc:Ignorable="d skia"
+			 xmlns:uc="using:Uno.UI.Samples.Converters"
+			 d:DesignHeight="300"
+			 d:DesignWidth="400">
+
+	<UserControl.Resources>
```
jeromelaban, line 14: For this sample, could you add more information about what should be visible? Which actions should be taken, and what to validate? Thanks!

src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/ContentPresenter/ContentPresenter_NativeEmbedding.xaml
```diff
@@ -0,0 +1,47 @@
...
+		<uc:FromNullableBoolToVisibilityConverter x:Key="FromNullableBoolToVisibilityConverter"
```
jeromelaban, line 15: This is what I have on linux:

image

ramezgerges: Oh, I left `mpv` as the process to run in the sample code, but I should probably change it so that it tries a bunch of different options.

MartinZikmund (approved):
Impressive ⭐

jeromelaban (approved):

PR #16827, opened 2024-05-21
Make UnoFeatures ignore casing

### LINKED ISSUE #16470: [SDK] Make UnoFeature case insensitive

Support for case insensitive features so that, for instance, MVVM and Mvvm are supported. Making the build warn if a feature isn't known at the quiet msbuild level would also be useful.

### Why is this needed

Make it easier to author.

### For which platform

_No response_

_No response_

dansiegel:

- closes #16470

## PR Type

- Feature

UnoFeature selection is case sensitive

## What is the new behavior?

UnoFeatures now ignore the case and cast everything ToLower

### DIFF
--- a/src/Uno.Sdk/targets/Uno.Features.targets
+++ b/src/Uno.Sdk/targets/Uno.Features.targets
@@ -1,7 +1,7 @@
 <Project>
 
 	<PropertyGroup>
-		<UnoFeatures>$([System.Text.RegularExpressions.Regex]::Replace($(UnoFeatures), '\s', '').ToLower())</UnoFeatures>
+		<UnoFeatures>$([System.Text.RegularExpressions.Regex]::Replace($(UnoFeatures), '\s', '').ToLowerInvariant())</UnoFeatures>
 		<UnoFeatures Condition="$(UnoFeatures) == ''">;</UnoFeatures>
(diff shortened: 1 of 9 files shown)


src/Uno.Sdk/targets/Uno.Features.targets
```diff
@@ -1,13 +1,13 @@
 <Project>
 
 	<PropertyGroup>
-		<UnoFeatures>$([System.Text.RegularExpressions.Regex]::Replace($(UnoFeatures), '\s', ''))</UnoFeatures>
+		<UnoFeatures>$([System.Text.RegularExpressions.Regex]::Replace($(UnoFeatures), '\s', '').ToLower())</UnoFeatures>
```
jeromelaban, line 4: I think we need ToLowerInvariant().

jeromelaban (approved):

agneszitte (commented):
@dansiegel, @jeromelaban 
No need to update the related documentation to precise that is not case sensitive?
https://platform.uno/docs/articles/features/using-the-uno-sdk.html?tabs=uno-packages#uno-platform-features

ajpinedam (approved):

spouliot (approved):

Issue #16587, opened 2024-05-06
[Android] FileOpenPicker returns invalid file path for videos selected


arianeleonard:
### Current behavior

When using FileOpenPicker's PickSingleFIleAsync or PickMultipleFIlesAsync and setting the SuggestedStartLocation property to PickerLocationId.VideosLibrary to select one or multiple videos, the path returned seems invalid on Android. When using FileInfo(string filename) to get the file information like the extension and size of the image, an exception is triggered for Android.

### Expected behavior

FileOpenPicker's PickSingleFIleAsync and PickMultipleFIlesAsync should return a valid file path for videos selected.

### How to reproduce it (as minimally and precisely as possible)

1. Launch the app on Android [VideoSelection.zip](https://github.com/unoplatform/uno/files/15223102/VideoSelection.zip)
2. Select "Select videos" button
3. Select some videos from the gallery and select "Add"
4. Exception should be displayed

### Works on UWP/WinUI

None

### Environment

Uno.WinUI / Uno.WinUI.WebAssembly / Uno.WinUI.Skia

### NuGet package version(s)

Uno WInUI : 5.2.108

### Affected platforms

Android

### IDE

Visual Studio 2022

### IDE version

17.9.2

jeromelaban:
@arianeleonard paths returned by the file picker are supposed to be opaque (the path is not necessarily supposed to be useable through `System.IO` APIs). You should make a copy of the StorageFile you get through Windows.Storage APIs, such as `FileIO` and similar first, then make `System.IO` operations on that copy after.

Out of curiosity, why do you need to interact the file directly?

MartinZikmund:
@arianeleonard Which equivalent `Windows.Storage` APIs would solve your usecase? We should tackle this by implementing them so that it is transparent to the target platform.

jeromelaban:
@arianeleonard we're adjusted the documentation in https://github.com/unoplatform/uno/pull/17193 to reflect the fact that the path provided by the `AndroidFilePicker` cannot return an actual file path (it would otherwise break the write-back feature).

It's best to use the `StorageFile` APIs to get the information you need, and if you need an actual file on the local storage using `System.IO`, make sure to create a copy of the file so you can manipulate it directly.

PR #16863, opened 2024-05-24
fix(droid): FileOpenPicker public path

ajpinedam:
GitHub Issue (If applicable): closes https://github.com/unoplatform/uno/issues/16587

## PR Type

- Bugfix

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UWP/Storage/Pickers/FileOpenPicker.Android.cs
+++ b/src/Uno.UWP/Storage/Pickers/FileOpenPicker.Android.cs
@@ -1,4 +1,4 @@
-﻿#nullable enable
+#nullable enable
 using System;
 using System.Collections.Generic;
 using System.Globalization;
@@ -9,6 +9,16 @@ using System.Threading.Tasks;
 using Android.App;
 using Android.Content;
 using Uno.UI;
+using Uno.UI.Dispatching;
+using Windows.UI.Core;
+using Uno.Extensions;
+using AndroidUri = Android.Net.Uri;
+using Android.Provider;
+using System.IO;
+using Java.Security;
+using Android.Webkit;
+using Android.Content.PM;
+using Uno.Foundation.Logging;
 
 namespace Windows.Storage.Pickers
 {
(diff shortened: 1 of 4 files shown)


ajpinedam:
Sample app running with a new Manual Test"

| Picking Videos | Picking Photos |
| :--: |  :--: |
uno_android_fileopenpicker_videos_fileinfo |  uno_android_fileopenpicker_photos_fileinfo |

src/Uno.UWP/Storage/Pickers/FileOpenPicker.Android.cs
```diff
@@ -240,5 +250,440 @@ private string GetMimeType()
 
 		internal void RegisterOnBeforeStartActivity(Action<Intent> intentAction)
 			=> _intentAction = intentAction;
+
+		static class FileSystemUtils
+		{
+			const string FolderHash = "2203693cc04e0be7f4f024d5f9499e13";
```
jeromelaban, line 256: Where is this hash coming from?

MartinZikmund:
I'm not sure if this is reliable, as the user can pick from any arbitrary location and the path will in many cases be not accessible via an actual path, only via that Android SAF framework. There is however a way to cache access to previously opened files (maybe it is just about storing the content URI, I don't remember exactly), so we could use that to implement FutureAccessList API

src/Uno.UWP/Storage/Pickers/FileOpenPicker.Android.cs
ajpinedam: This is just a directory name. We can name it differently and have a different value, like "UnoPlatform"

MartinZikmund:
Ah, I understand now, so this does copies the picked files into the temporary storage of the app? I'm not sure if that is the ideal, as this means that if a potentially large file is picked, it will take time to copy it as well as it will require additional space on the device. Plus, I think that this change means one cannot do actual read/write changes to picked files, as modifications will actually be done to this private copy instead, so this could even be a breaking change. I guess if that is the only way, it should be responsibility of the user to copy the file to temporary storage instead, as the path that is shown there is not the "actual" path to the source file anyway. What do you think @jeromelaban?

MartinZikmund:
Sorry about that close, accidental :-D

ajpinedam:
@jeromelaban  @MartinZikmund  here's the PR with the documentation changes related to this issue: https://github.com/unoplatform/uno/pull/17143

ajpinedam:
Closing as we are not going to go this route as these changes break some other functionality. Refer to the documentation for more information.

PR #15951, opened 2024-03-20
feat(skia): Gif support

Youssef1313:
GitHub Issue (If applicable): Skia part of #9598.

## PR Type

What kind of change does this PR introduce?

## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)
- [ ] [Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [ ] Validated PR `Screenshots Compare Test Run` results.
- [ ] Contains **NO** breaking changes
- [ ] Associated with an issue (GitHub or internal) and uses the [automatic close keywords](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue).
- [ ] Commits must be following the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) specification.

## Other information

Internal Issue (If applicable):

PR #16896, opened 2024-05-28
perf(skia): only repaint a Visual when necessary

ramezgerges:

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI.Composition/Composition/CompositionSpriteShape.skia.cs
+++ b/src/Uno.UI.Composition/Composition/CompositionSpriteShape.skia.cs
@@ -39,10 +39,11 @@ namespace Microsoft.UI.Composition
 					{
 						// workaround until SkiaSharp adds support for SaveLayerRec
 						fillPaint.FilterQuality = SKFilterQuality.High;
-						session.Canvas.SaveLayer(fillPaint);
-						session.Canvas.Scale(1.0f / session.Canvas.TotalMatrix.ScaleX);
-						session.Canvas.DrawSurface(session.Surface, new(-session.Canvas.TotalMatrix.TransX, -session.Canvas.DeviceClipBounds.Top + session.Canvas.LocalClipBounds.Top));
-						session.Canvas.Restore();
+						var sessionCanvas = session.Surface.Canvas;
+						sessionCanvas.SaveLayer(fillPaint);
+						sessionCanvas.Scale(1.0f / sessionCanvas.TotalMatrix.ScaleX);
+						sessionCanvas.DrawSurface(session.Surface, new(-sessionCanvas.TotalMatrix.TransX, -sessionCanvas.DeviceClipBounds.Top + sessionCanvas.LocalClipBounds.Top));
+						sessionCanvas.Restore();
 					}
 					else
 					{
--- a/src/Uno.UI.Composition/Composition/CompositionSpriteShape.skia.cs
+++ b/src/Uno.UI.Composition/Composition/CompositionSpriteShape.skia.cs
@@ -39,7 +39,7 @@ namespace Microsoft.UI.Composition
 					{
 						// workaround until SkiaSharp adds support for SaveLayerRec
 						fillPaint.FilterQuality = SKFilterQuality.High;
-						// For effect brushes, we need to draw directly on the surface's canvas, otherwise
+						// Here, we need to draw directly on the surface's canvas, otherwise
 						// you can get an AccessViolationException. session.Canvas is possibly a RecorderCanvas
 						// from an SKPictureRecorder, so we bypass it and use Surface.Canvas.
 						var sessionCanvas = session.Surface.Canvas;
--- a/src/Uno.UI.Composition/Composition/CompositionSpriteShape.skia.cs
+++ b/src/Uno.UI.Composition/Composition/CompositionSpriteShape.skia.cs
@@ -39,6 +39,9 @@ namespace Microsoft.UI.Composition
 					{
 						// workaround until SkiaSharp adds support for SaveLayerRec
 						fillPaint.FilterQuality = SKFilterQuality.High;
+						// For effect brushes, we need to draw directly on the surface's canvas, otherwise
+						// you can get an AccessViolationException. session.Canvas is possibly a RecorderCanvas
+						// from an SKPictureRecorder, so we bypass it and use Surface.Canvas.
 						var sessionCanvas = session.Surface.Canvas;
 						sessionCanvas.SaveLayer(fillPaint);
 						sessionCanvas.Scale(1.0f / sessionCanvas.TotalMatrix.ScaleX);
--- a/src/Uno.UI.Composition/Composition/CompositionSpriteShape.skia.cs
+++ b/src/Uno.UI.Composition/Composition/CompositionSpriteShape.skia.cs
@@ -40,8 +40,10 @@ namespace Microsoft.UI.Composition
 						// workaround until SkiaSharp adds support for SaveLayerRec
 						fillPaint.FilterQuality = SKFilterQuality.High;
 						// Here, we need to draw directly on the surface's canvas, otherwise
-						// you can get an AccessViolationException. session.Canvas is possibly a RecorderCanvas
-						// from an SKPictureRecorder, so we bypass it and use Surface.Canvas.
+						// you can get an AccessViolationException (most likely because DrawSurface needs the
+						// receiver SKCanvas object to be the same as the first argument's SKSurface.Canvas).
+						// session.Canvas is possibly a RecorderCanvas from an SKPictureRecorder, so we bypass
+						// it and use Surface.Canvas.
 						var sessionCanvas = session.Surface.Canvas;
 						sessionCanvas.SaveLayer(fillPaint);
 						sessionCanvas.Scale(1.0f / sessionCanvas.TotalMatrix.ScaleX);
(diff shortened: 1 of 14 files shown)


src/Uno.UI/UI/Xaml/Media/Imaging/WriteableBitmap.skia.cs
```diff
@@ -13,9 +13,7 @@ partial class WriteableBitmap
 
 		private protected override bool TryOpenSourceSync(int? targetWidth, int? targetHeight, out ImageData image)
 		{
-			_surface ??= new SkiaCompositionSurface();
-
-			_surface.CopyPixels(PixelWidth, PixelHeight, _buffer.AsReadOnlyMemory());
+			_surface = SkiaCompositionSurface.LoadFromPixels(PixelWidth, PixelHeight, _buffer.AsReadOnlyMemory());
```
Youssef1313, line 16: Previously, we were potentially re-using the same surface.

I think this change is not necessary for this PR and we can be better on the safe side and keep `CopyPixels` as instance method.

ramezgerges: No, it's necessary for exactly that reason. Using the same "surface" was wrong, since the SkiaCompositionSurface is just a wrapper for a simple SKImage and changing the image should cause render invalidation, which didn't previously happen because we were changing the internal SKImage without changing the surface instance.

src/Uno.UI.Composition/Composition/Visual.skia.cs
```diff
@@ -30,6 +39,17 @@ internal virtual bool SetMatrixDirty()
 		return !matrixDirty;
 	}
 
+	public void InvalidatePaint()
```
Youssef1313, line 42: It shouldn't be public

src/Uno.UI.Composition/Composition/CompositionSurfaceBrush.skia.cs
```diff
@@ -77,16 +77,16 @@ private static bool TryGetSkiaCompositionSurface(ICompositionSurface? surface, [
 
 		internal override void UpdatePaint(SKPaint fillPaint, SKRect bounds)
 		{
-			if (TryGetSkiaCompositionSurface(Surface, out var scs))
+			if (TryGetSkiaCompositionSurface(Surface, out var scs) && scs is { Image: SKImage image })
```
dr1rrb, line 80: If I'm right, the `Image` can be `null` anymore, you should just make the `Image` non nullable and remove the check here.

src/Uno.UI.Composition/Composition/SpriteVisual.skia.cs
```diff
@@ -67,5 +68,10 @@ internal override void Paint(in PaintingSession session)
 				);
 			}
 		}
+
+		internal override bool RequiresRepaintOnEveryFrame =>
+			Brush is CompositionEffectBrush { HasBackdropBrushInput: true } or IOnlineBrush { IsOnline: true };
```
dr1rrb, line 73: :)

src/Uno.UI.Composition/Composition/SpriteVisual.skia.cs
```diff
@@ -54,10 +54,11 @@ internal override void Paint(in PaintingSession session)
 			if (Brush is CompositionEffectBrush { HasBackdropBrushInput: true })
 			{
 				// workaround until SkiaSharp adds support for SaveLayerRec, see https://github.com/mono/SkiaSharp/issues/2773
-				session.Canvas.SaveLayer(_paint);
-				session.Canvas.Scale(1.0f / session.Canvas.TotalMatrix.ScaleX);
-				session.Canvas.DrawSurface(session.Surface, new(-session.Canvas.TotalMatrix.TransX, -session.Canvas.DeviceClipBounds.Top + session.Canvas.LocalClipBounds.Top));
-				session.Canvas.Restore();
+				var sessionCanvas = session.Surface.Canvas;
```
dr1rrb, line 57:
```suggestion
				var sessionCanvas = session.Canvas;
```

src/Uno.UI/UI/Xaml/Media/Imaging/ImageSourceHelpers.cs
```diff
@@ -34,16 +35,13 @@ public static async Task<ImageData> ReadFromStreamAsBytesAsync(Stream stream, Ca
...
-
-		if (result.success)
+		if (SkiaCompositionSurface.LoadFromStream(imageStream) is { } surface)
 		{
 			return Task.FromResult(ImageData.FromCompositionSurface(surface));
 		}
 		else
 		{
-			var exception = new InvalidOperationException($"Image load failed ({result.nativeResult})");
+			var exception = new InvalidOperationException($"Image load failed");
```
dr1rrb, line 44: There are some case where the `SkiaCompositionSurface.LoadFromStream` raise an exception that is logged, dropped and we return `null`. We should not hide useful information to the users on the reason why the image load failed. You should instead changed the `LoadFromStream` (or create an overload):

```
bool SkiaCompositionSurface.TryLoadFromStream(
    [NotNullWhen(true)] SkiaCompositionSurface? surface, 
    [NutNullWhen(false) Exception? error)
```

and actually thinking about it, as the `SkiaCompositionSurface` only contains an `SKImage` we could even drop it  and just return an `SKImage`:


```
bool SkiaCompositionSurface.TryLoadFromStream(
    [NotNullWhen(true)] SKImage? image, 
    [NutNullWhen(false) Exception? error)
```

Not sure if we could image scenarios where we would return something else than an `SKImage` ?

src/Uno.UI/UI/Xaml/Media/Imaging/WriteableBitmap.skia.cs
Youssef1313: I don't see anything wrong with re-using the same surface. It just currently doesn't work the way it should.

@jeromelaban This is the root cause for the ProgressRing issue where we are reusing the same surface and not causing invalidate render. PR https://github.com/unoplatform/uno/pull/17059 has "what I think" the proper fix.

ramezgerges: The entire idea of a "surface" is misleading. It's not a surface, it's just a dumb wrapper for an SKImage because we can't implement `ICompositionSurface` directly on the SKImage. It doesn't make a lot of sense to "reuse the same surface" because there is no "surface". We would be "reusing the same wrapper but change the the thing is wraps", which is a lot of work to just "replace the thing". By making the wrapper a readonly struct, we make reasoning about the relationship between the wrapper and what it wraps a lot easier.

src/Uno.UI.Composition/Composition/CompositionSpriteShape.skia.cs
```diff
@@ -39,10 +39,11 @@ internal override void Paint(in Visual.PaintingSession session)
 					{
 						// workaround until SkiaSharp adds support for SaveLayerRec
 						fillPaint.FilterQuality = SKFilterQuality.High;
-						session.Canvas.SaveLayer(fillPaint);
-						session.Canvas.Scale(1.0f / session.Canvas.TotalMatrix.ScaleX);
-						session.Canvas.DrawSurface(session.Surface, new(-session.Canvas.TotalMatrix.TransX, -session.Canvas.DeviceClipBounds.Top + session.Canvas.LocalClipBounds.Top));
-						session.Canvas.Restore();
+						var sessionCanvas = session.Surface.Canvas;
```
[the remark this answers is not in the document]
ramezgerges: With effect brushes specifically, we have to bypass the `RecorderCanvas` and draw directly on the surface canvas. This is because effect brushes usually after an effect to an existing drawing, and this effect won't appear if you use the recording canvas. I will add a comment.

src/Uno.UI.Composition/Composition/ShapeVisual.skia.cs
```diff
@@ -31,6 +34,19 @@ internal override void Paint(in PaintingSession session)
 		base.Paint(in session);
 	}
 
+	private protected override void OnPropertyChangedCore(string? propertyName, bool isSubPropertyChange)
+	{
+		base.OnPropertyChangedCore(propertyName, isSubPropertyChange);
+		if (propertyName == nameof(Shapes))
+		{
+			_needsContinuousUpdates = _shapes?.OfType<CompositionSpriteShape>().Any(s => s.FillBrush is CompositionEffectBrush { HasBackdropBrushInput: true } or IOnlineBrush { IsOnline: true }) ?? false;
```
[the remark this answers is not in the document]
ramezgerges: We get notifications to changes on the collection for free because of the `CompositionObject._contextEntries`. I manually tested that this indeed propagates since I had the same concerns.

src/Uno.UI/UI/Xaml/Media/Imaging/ImageSourceHelpers.cs
ramezgerges: Love this pattern. As for returning an SKImage directly: SkiaCompositionSurface will soon be more complex than a simple SKImage (https://github.com/unoplatform/uno/pull/15951/).

src/Uno.UI/UI/Xaml/Media/Imaging/WriteableBitmap.skia.cs
ramezgerges: As we discussed, https://github.com/unoplatform/uno/pull/15951/ will break by some of the changes in SkiaCompositionSurface. Let's wait for it to merge fix and then see what to do.

PR #17033, opened 2024-06-07
Revert "fix(imagebrush): downscaling images was very blurry"

Youssef1313:
Since the merge of this PR, we noticed failures in the new added test and it was skipped. Now, noticed the same failure for other tests that existed before this PR, so it's a regression.

```
info: Uno.UI.Samples.Tests.UnitTestsControl[0]
      Running test When_Stretch(Uniform, False)
GLIB UNHANDLED EXCEPTIONSystem.ArgumentNullException: Value cannot be null. (Parameter 'bitmap')
   at SkiaSharp.SKImage.FromBitmap(SKBitmap bitmap)
   at Microsoft.UI.Composition.CompositionSurfaceBrush.UpdatePaint(SKPaint fillPaint, SKRect bounds) in C:\a\1\s\src\Uno.UI.Composition\Composition\CompositionSurfaceBrush.skia.cs:line 91
   at Microsoft.UI.Composition.CompositionSpriteShape.Paint(PaintingSession& session) in C:\a\1\s\src\Uno.UI.Composition\Composition\CompositionSpriteShape.skia.cs:line 19
   at Microsoft.UI.Composition.BorderVisual.Paint(PaintingSession& session) in C:\a\1\s\src\Uno.UI.Composition\Composition\BorderVisual.skia.cs:line 127
   at Microsoft.UI.Composition.Visual.Render(PaintingSession& parentSession) in C:\a\1\s\src\Uno.UI.Composition\Composition\Visual.skia.cs:line 222
   at Microsoft.UI.Composition.Visual.Render(PaintingSession& parentSession) in C:\a\1\s\src\Uno.UI.Composition\Composition\Visual.skia.cs:line 232
   at Microsoft.UI.Composition.Visual.Render(PaintingSession& parentSession) in C:\a\1\s\src\Uno.UI.Composition\Composition\Visual.skia.cs:line 232
   at Microsoft.UI.Composition.Visual.Render(PaintingSession& parentSession) in C:\a\1\s\src\Uno.UI.Composition\Composition\Visual.skia.cs:line 232
   at Microsoft.UI.Composition.Visual.Render(PaintingSession& parentSession) in C:\a\1\s\src\Uno.UI.Composition\Composition\Visual.skia.cs:line 232
   at Microsoft.UI.Composition.Visual.Render(PaintingSession& parentSession) in C:\a\1\s\src\Uno.UI.Composition\Composition\Visual.skia.cs:line 232
[...]

PR #16816, opened 2024-05-21
fix(imagebrush): downscaling images was very blurry

ramezgerges:
GitHub Issue (If applicable): closes #

## PR Type

What kind of change does this PR introduce?

## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)
- [ ] [Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [ ] Validated PR `Screenshots Compare Test Run` results.
- [ ] Contains **NO** breaking changes
- [ ] Associated with an issue (GitHub or internal) and uses the [automatic close keywords](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue).
- [ ] Commits must be following the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) specification.

## Other information

Internal Issue (If applicable):

PR #17035, opened 2024-06-07
fix(surfacebrush): check that the final size isn't empty before resizing images.

ramezgerges:
GitHub Issue (If applicable): #17033
Let's see if this is the problem. My guess is that we might try to render the image right before a needed layout cycle. Looking at SkiaSharp's source, it returns null if you try to resize to a zero rectangle.

### DIFF
--- a/src/Uno.UI.Composition/Composition/CompositionSurfaceBrush.skia.cs
+++ b/src/Uno.UI.Composition/Composition/CompositionSurfaceBrush.skia.cs
@@ -83,36 +83,38 @@ namespace Microsoft.UI.Composition
 
 				if (backgroundArea.Width <= 0 || backgroundArea.Height <= 0)
 				{
-					return;
-				}
-
-				// Adding image downscaling in the shader matrix directly is very blurry
-				// since the default downsampler in Skia is really low quality (but really fast).
-				// We force Lanczos instead.
-				// https://github.com/mono/SkiaSharp/issues/520#issuecomment-444973518
-#pragma warning disable CS0612 // Type or member is obsolete
-				var resizedBitmap = scs.Image.ToSKBitmap().Resize(scs.Image.Info.WithSize((int)backgroundArea.Size.Width, (int)backgroundArea.Size.Height), SKBitmapResizeMethod.Lanczos3.ToFilterQuality());
-#pragma warning restore CS0612 // Type or member is obsolete
-				var resizedImage = SKImage.FromBitmap(resizedBitmap);
-
-				var matrix = Matrix3x2.CreateTranslation((float)backgroundArea.Left, (float)backgroundArea.Top);
-				matrix *= TransformMatrix;
-				var imageShader = SKShader.CreateImage(resizedImage, SKShaderTileMode.Decal, SKShaderTileMode.Decal, matrix.ToSKMatrix());
-
-				if (UsePaintColorToColorSurface)
-				{
-					// use the set color instead of the image pixel values. This is what happens on WinUI.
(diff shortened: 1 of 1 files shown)


src/Uno.UI.Composition/Composition/CompositionSurfaceBrush.skia.cs
```diff
@@ -81,6 +81,11 @@ internal override void UpdatePaint(SKPaint fillPaint, SKRect bounds)
 			{
 				var backgroundArea = GetArrangedImageRect(new Size(scs.Image!.Width, scs.Image.Height), bounds);
 
+				if (backgroundArea.Width <= 0 || backgroundArea.Height <= 0)
+				{
+					return;
+				}
```
Youssef1313, line 87: Instead of this, I'd do it this way:

```csharp
var resizedBitmap = scs.Image.ToSKBimap();
if (backgroundArea.Width > 0 && backgroundArea.Height > 0)
{
    resizedBitmap = resizedBitmap.Resize(...);
}

var resizedImage = SKImage.FromBitmap(resizedBitmap);
// ...
```

The reason I prefer this way is to ensure as we can that there are no surprises regarding any behavioral changes. For example, I'm not sure if it's possible this return can keep the `fillPaint` unchanged when it should have changed.

ramezgerges: Makes sense. It's important that the `fillPaint` is cleared before sending it to `UpdatePaint` anyway, but sure.

jeromelaban: Shouldn't we check for null at the location of the original issue? Was the null only returned because of the size?

Youssef1313: The null is returned by the `Resize` call. If we'll just ignore the resizing if `Resize` is null, that could mask an issue if it's something other than the dimensions.

ramezgerges: Actually now thinking about it, proceeding without resizing if `backgroundArea` is empty looks really weird to me. If the area is empty, we shouldn't be doing anything. At the very least, we shouldn't be proceeding with a full-size image. I will prefer to keep the PR as is.

Youssef1313: It's a behavior change that's unrelated to the `ArgumentNullException`. If the current behavior wrong, we need to have a scenario showing that (and preferably a runtime test)

ramezgerges: It's not really a behavior change. The invariant was and should continue to be that after whatever processing we do, the image we render is now sized at `backgroundArea`. Before #16816, we always resized but with terrible quality. After #16816, we crash if resizing failed. With this PR, we don't draw at all if we can't resize. Proceeding with painting with an image shader that doesn't match the `backgroundArea` would be the only change that breaks this invariant.

Youssef1313: Then maybe we should cleear any possibly existing Shader on fillPaint?

ramezgerges: Yes, good point.

Youssef1313 (approved):

jeromelaban:
@mergifyio backport release/stable/5.3

PR #17048, opened 2024-06-07
feat: always include Lottie on Desktop

### LINKED ISSUE #17024: Include Lottie by default for the desktop target

Lottie does not add any weight in the desktop target at this time, we can include it by default.

### Why is this needed

Have ProgressRing work by default without adding dependencies.

### For which platform

_No response_

_No response_

dansiegel:

- closes #17024

## PR Type

- Bugfix

Lottie is not always included for the Desktop target

## What is the new behavior?

Lottie is now included always for the Desktop target

### DIFF
--- a/src/Uno.Sdk/targets/Uno.Implicit.Packages.ProjectSystem.Uno.targets
+++ b/src/Uno.Sdk/targets/Uno.Implicit.Packages.ProjectSystem.Uno.targets
@@ -17,7 +17,7 @@
 	</ItemGroup>
 
 	<ItemGroup Condition="$(IsBrowserWasm) != 'true' AND '$(IsPackable)' != 'true'">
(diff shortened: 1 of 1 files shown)


jeromelaban (approved):

agneszitte (approved):

src/Uno.Sdk/targets/Uno.Implicit.Packages.ProjectSystem.Uno.targets
```diff
@@ -4,7 +4,7 @@
 		in order for VS and C# Dev Kit to show nuget references in their respective solution explorers.
 		The version is not required, and VS/Code waits for some design-time targets to be executed to evaluate it.
 	-->
-	<ItemGroup Condition="$(UnoFeatures.Contains(';lottie;')) OR $(UnoFeatures.Contains(';material;')) OR $(UnoFeatures.Contains(';cupertino;'))">
+	<ItemGroup Condition="$(UnoFeatures.Contains(';lottie;')) OR $(UnoFeatures.Contains(';material;')) OR $(UnoFeatures.Contains(';cupertino;')) OR $(IsDesktop) == 'true'">
```
jeromelaban, line 7: We need to do the same with SkiaSharp.Views.Uno.WinUI also (so says the build).

carldebilly (approved):

PR #17062, opened 2024-06-08
ci: Always use .NET 9 SDK

Youssef1313:

## PR Type

## PR Checklist

## Other information


build/ci/.azure-devops-android-tests.yml
```diff
@@ -48,6 +48,11 @@ jobs:
 
   - template: templates/gitversion.yml
 
+  - script: |
+      cd $(build.sourcesdirectory)/src/SamplesApp/SamplesApp.netcoremobile
+      dotnet workload restore
+    displayName: Workload restore
```
jeromelaban, line 54: This should not be needed if we have run uno.check. Doing this will likely result in build failures down the road.

Why did you need to add it? We may need to change uno.check instead.

build/ci/.azure-devops-docs.yml
```diff
@@ -27,6 +27,11 @@ jobs:
 
   - template: templates/dotnet-mobile-install-windows.yml
 
+  - script: |
+      cd src/Uno.UI
+      dotnet workload restore Uno.UI.netcoremobile.csproj
+    displayName: Workload restore
```
jeromelaban, line 33: Same here.

src/Uno.UI.Runtime.Skia.Wpf/Builder/WpfHostBuilder.cs
```diff
@@ -26,6 +26,36 @@ public WpfHostBuilder()
 		}
 	}
 
+	private static void AdjustPath(string dotnetSharedDirectory, string runtimeVersion, ref string windowsDesktopFrameworkPath)
+	{
+		// In at least preview SDKs, the .NET runtime version is different from the .NET Desktop runtime version for the same .NET SDK.
```
jeromelaban, line 31: Interesting change. We may need to adjust this for the net9 release as well.

build/ci/.azure-devops-android-tests.yml
Youssef1313: @jeromelaban It's all about maintenance hassle of manifests, especially that now we are just using .NET 9 SDK ad building for both net8.0/net9.0 TFMs. `dotnet workload restore` will always do the right thing and just install the right workloads. Otherwise, it's a nightmare to maintain different manifests and having to create ones that work with .NET 9 SDK but for net8.0 builds, etc. And I think if we will later build both net8.0 and net9.0 in the same job, we will need two workload installation. But `dotnet workload restore` should get us working almost with zero maintenance cost. It hopefully shouldn't result in build failures, but we'll see.

src/Uno.UI.Runtime.Skia.Wpf/Builder/WpfHostBuilder.cs
Youssef1313: So far it's looking like preview-specific. But we'll see how it goes with stable SDK

build/ci/templates/dotnet-mobile-install-mac.yml
```diff
@@ -11,20 +8,9 @@ steps:
   - template: jdk-setup.yml
 
   - powershell: |
-      & dotnet tool update --global uno.check --version ${{ parameters.UnoCheck_Version_Previous }} --add-source https://api.nuget.org/v3/index.json
-      & uno-check -v --ci --non-interactive --fix --skip androidsdk --skip androidemulator --skip xcode --skip gtk3 --skip vswin --skip vsmac --manifest ${{ parameters.UnoCheck_Manifest_Previous }}
+      & dotnet tool update --global uno.check --version ${{ parameters.UnoCheck_Version }} --add-source https://api.nuget.org/v3/index.json
```
jeromelaban, line 11: use `--pre-major`

jeromelaban (approved):

PR #17124, opened 2024-06-13
fix: builds fail with latest MSAL updates

dansiegel:

- unoplatform/uno.templates#743 - Build Failure

## PR Type

- Bugfix

We do not add Microsoft.Identity.Client.Extensions.MSAL
We add MSAL packages across all TFMs when the `AuthenticationMSAL` UnoFeature is present

## What is the new behavior?

We now also provide the top level dependency for Microsoft.Identity.Client.Extensions.MSAL
We exclude the MSAL packages but not the Extensions package from unsupported targets.

### DIFF
--- a/src/Uno.Sdk/packages.json
+++ b/src/Uno.Sdk/packages.json
@@ -125,9 +125,10 @@
 	},
 	{
 		"group": "MsalClient",
-		"version": "4.61.1",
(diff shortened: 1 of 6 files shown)


jeromelaban (approved):

jeromelaban:

Issue #15863, opened 2024-03-12
Changing UnoImage to Content throw and error in Visual Studio


mcNets:
### Current behavior

Changing Build property from `UnoImage` to `Content` of a .svg image added to a new solution (Windows & Skia) throws and error:

> Property value is not valid.
> An error has occurred while saving the edited properties listed below:
>	Build Action
>     One or more values are invalid. value is object

### Expected behavior

Build property should change from UnoImage to Content.

### How to reproduce it (as minimally and precisely as possible)

- Generate a new solution for Windows and Skia. I used this command:
```
dotnet new unoapp -o unosvgcli -id "com.companyname.unosvgcli" -pub "O=unosvgcli" -platforms "windows" -platforms "gtk" -pwa False -theme "fluent" -presentation "none" -config false -di false -log "none" -nav "blank" -http false -loc false -server false -tests "none" -toolkit false -dsp false
```

- Add an Images folder to the Assets folder in the main project.

- Add a .svg file to the Assets/Images folder using 'Add existing item'. This lines are added to the main project:

```
  <ItemGroup>
    <Content Remove="Assets\Images\uno.svg" />
  </ItemGroup>
  <ItemGroup>
    <None Remove="Assets\Images\uno.svg" />
  </ItemGroup>
  ```

- In the properties tool window try to change from UnoImage 

- Try to change Build property from `UnoImage` to `Content` in the properties tool window.

Visual studio throws this error message:

> Property value is not valid.
> An error has occurred while saving the edited properties listed below:
>	Build Action
>     One or more values are invalid. value is object

### Workaround

You can manually change:

```
  <ItemGroup>
    <Content Remove="Assets\Images\uno.svg" />
  </ItemGroup>
  <ItemGroup>
    <None Remove="Assets\Images\uno.svg" />
  </ItemGroup>
  ```

to

```
  <ItemGroup>
    <Content Include="Assets\Images\uno.svg" />
  </ItemGroup>
  <ItemGroup>
    <None Remove="Assets\Images\uno.svg" />
  </ItemGroup>
  ```

### Works on UWP/WinUI

None

### Anything else we need to know?

Uno templates: 5.1.25

Visual Studio 17.10 Prev. 1

agneszitte:
@mcNets Thank you for opening the issue with the details. Really appreciated!

For anyone looking at this issue, there are two main things here:
[...]

PR #17139, opened 2024-06-13
Allow explicitly added SVG as Content to be excluded from UnoImage

dansiegel:

- #15863

## PR Type

- Bugfix

- SVG's not inside of an SVG directory will always be included by default as an UnoImage.
- Various items including Content, Page, & UnoImage are also duplicated as None items

## What is the new behavior?

- Items that have been explicitly added as Content will be excluded from the default UnoImage items
- Items added as Content, Page, UnoImage, etc are now explicitly removed from the None Item group.


ajpinedam (approved):

agneszitte (approved):

jeromelaban:
@dansiegel can you track this new build, the last build failed because of a missing asset.

Issue #1285, opened 2019-07-22
Given_ApplicationDataContainer_When_GetAllKeys fails on iOS


MartinZikmund:
## Current behavior

Given_ApplicationDataContainer_When_GetAllKeys fails on iOS when run on my device. The problem seems to be the fact, that even after clearing, the `NSUserDefaults.StandardUserDefaults` contains a ton of weird default settings like:

> AppleLanguages, AppleLocale, NSInterfaceStyle, AppleITunesStoreItemKinds, AppleKeyboardsExpanded, PKKeychainVersionKey, ApplePasscodeKeyboards , AppleKeyboards, NSLanguages, NSInterfaceStyle

One of those being a date, which contains `:`, which Uno, unfortunately, uses as deserialization separator. This makes the `DataTypeSerializer` crash with exception.

## Expected behavior

Runtime tests should run from SamplesApp without issues.

## How to reproduce it (as minimally and precisely as possible)

Run Runtime tests via the SamplesApp on iOS.

## Environment

Nuget Package:

Package Version(s):

Affected platform(s):
- [x] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] Windows
- [ ] Build tasks

Visual Studio
- [ ] 2017 (version: )
- [x] 2019 (version: 16.2 Preview 3)
- [ ] for Mac (version: )

Relevant plugins
- [ ] Resharper (version: )

xperiandri:
Critical bug. Having it you kan only save a fixed set of values keys for which you hardcode in code

MartinZikmund:
The cause is that these come from the global domain:

https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/UserDefaults/AboutPreferenceDomains/AboutPreferenceDomains.html

We should be able to filter them out, will do that as part of App data changes PR

MartinZikmund:
Still relevant to both Skia and Native renderer

MartinZikmund:
# Repro generator

Runtime test(s) added in branch [`dev/mazi/repro/1285`](https://github.com/unoplatform/uno/compare/dev/mazi/repro/1285):
- `src/Uno.UI.RuntimeTests/Tests/Windows_Storage/Given_ApplicationDataContainer.cs` → `When_GetEnumerator_Returns_Correct_Values_Not_Keys`
[...]

PR #17183, opened 2024-06-17
Handle non-Uno settings in `DataTypeSerializer` gracefully, add tests

MartinZikmund:
GitHub Issue (If applicable): part of #1285

## PR Type

- Bugfix

## PR Checklist

## Other information


carldebilly (approved):

MartinZikmund:
@jeromelaban This is to handle cases when a component the app is using (could be a 3rd party control for example), writes a setting using native APIs, which is a string which contains a colon. This previously threw an exception, now it just gracefully returns `null`. Alternative would be to just return the full string as the fallback. Not sure which approach is better. Technically for Uno-saved values we should never hit this. But not sure if `ApplicationData` should return `null` for non-Uno strings, or return the string value itself

PR #17208, opened 2024-06-18
fix(selector): restore SelectorItem visual state after recycling

### LINKED ISSUE #16542: Samples list loses selection state on scrolling
### Current behavior

https://github.com/unoplatform/uno/assets/31348972/68a87f5e-a539-4021-99a6-ebcf6b70feb3

### Expected behavior

_No response_

_No response_

### Workaround

_No response_

### Works on UWP/WinUI

None

### Environment

_No response_

### NuGet package version(s)

_No response_

### Affected platforms

_No response_

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

ramezgerges:
GitHub Issue (If applicable): closes #16542

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/Primitives/Selector.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Primitives/Selector.cs
@@ -63,6 +63,8 @@ namespace Microsoft.UI.Xaml.Controls.Primitives
 		/// </summary>
 		private readonly HashSet<DataTemplate> _itemTemplatesThatArentContainers = new HashSet<DataTemplate>();
 
+		private string _selectedContainerStateBeforeRecycling;
+
 		public Selector()
 		{
 
@@ -127,6 +129,8 @@ namespace Microsoft.UI.Xaml.Controls.Primitives
 
 		internal virtual void OnSelectedItemChanged(object oldSelectedItem, object selectedItem, bool updateItemSelectedState)
 		{
+			_selectedContainerStateBeforeRecycling = null;
+
 			var wasSelectionUnset = oldSelectedItem == null && (!GetItems()?.Contains(null) ?? false);
 			var isSelectionUnset = false;
 			var items = GetItems();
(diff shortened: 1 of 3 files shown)


src/Uno.UI/UI/Xaml/Controls/Primitives/Selector.cs
```diff
@@ -510,13 +514,26 @@ private void OnCollectionViewCurrentChanged(object sender, object e)
...
 		}
 
+		internal void OnContainerRecycled(DependencyObject element)
+		{
+			if (element is SelectorItem { IsSelected: true } selectorItem)
+			{
+				_selectedContainerStateBeforeRecycling = VisualStateManager.GetCurrentState(selectorItem, "CommonStates")?.Name;
+			}
+		}
+
```
ramezgerges, line 524: I'm not particularly fond of this fix, but it's the only way I found to get it working. We need to restore the visual state of the recycle containers somehow. I'm not exactly sure if WinUI restores the visual state for all the containers, or only the one that IsSelected. Non-selected containers normally don't have interesting visual states, so you can't really notice the difference. The selected container specifically needs to be restored because we need to restore the `Pressed` state of the container (which is the related issue adds a grey background).

jeromelaban: In theory, visual states are supposed to be a reflection of the IsSelected and other properties, though that may not be true for pointer related states.

ramezgerges: Yeah, exactly. The problem is that we can't easily recreate the (pointer-related) visual state of the recycled-then-restored containers. This is due to some weird (but accurate) behaviour in the way SelectorItem does visual states, where the current visual state might not reflect the current "state". E.g., clicking an item might keep it in the `Pressed` visual state even thought it's no longer pressed (which is what happens in the repro of the related issue). In that case, trying to use the current state (i.e. IsSelected, IsEnabled, IsPressed, etc.) to go to the correct visual state will fail, so we need to remember it somewhere.
This problem seems to only happen with selected items. Other items can recreate their visual states naturally from their current states as far as I can tell.

I just noticed now that this fix assumes a single selected item, but ListViews can have multiple selection, so I need to look into that as well.

PR #17224, opened 2024-06-19
Revert "fix: Don't use target platform version in nuspec"

Youssef1313:
Reverts unoplatform/uno#16997


jeromelaban:
Let's keep this one for 5.4.

PR #16762, opened 2024-05-17
Make Samples app use fluent styles by default

MartinZikmund:
GitHub Issue (If applicable): closes #

## PR Type

What kind of change does this PR introduce?

## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)
- [ ] [Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [ ] Validated PR `Screenshots Compare Test Run` results.
- [ ] Contains **NO** breaking changes
- [ ] Associated with an issue (GitHub or internal) and uses the [automatic close keywords](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue).
- [ ] Commits must be following the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) specification.

## Other information

Internal Issue (If applicable):

PR #17357, opened 2024-07-01
fix(treeview): fix dragging in some cases causing an NRE

### LINKED ISSUE #17342: [Skia] `TreeView` on drag enter NRE
### Current behavior

When randomly dragging items in `TreeView` multiple times, the OnDragEnter causes NRE `draggedOverNode was null`:
image

### Expected behavior

_No response_

1. Create a new project with TreeView control:
```xml
      <TreeView x:Name="TreeViewControl">
        <TreeView.ItemTemplate>
          <DataTemplate>
            <TreeViewItem ItemsSource="{Binding Children}"
                          Content="{Binding Name}" />
          </DataTemplate>
        </TreeView.ItemTemplate>
      </TreeView>
```

```csharp
        TreeViewControl.ItemsSource = new ObservableCollection<TreeViewItemModel>
        {
            new TreeViewItemModel
            {
                Name = "Root 1",
                Children = new ObservableCollection<TreeViewItemModel>
                {
                    new TreeViewItemModel { Name = "Child 1.1" },
                    new TreeViewItemModel { Name = "Child 1.2" }
                }
            },
            new TreeViewItemModel
            {
                Name = "Root 2",
                Children = new ObservableCollection<TreeViewItemModel>
                {
                    new TreeViewItemModel { Name = "Child 2.1" }
                }
            }
        };

    public class TreeViewItemModel
    {
        public string Name { get; set; }
        public ObservableCollection<TreeViewItemModel> Children { get; set; } = new ObservableCollection<TreeViewItemModel>();
    }
```
2. Try to drag the Item multiple times between nodes, it randomly results in NRE

### Workaround

_No response_

### Works on UWP/WinUI

None

### Environment

_No response_

### NuGet package version(s)

Uno.WinUI `5.4.0-dev.121`

### Affected platforms

Skia (WPF), Skia (Linux X11), Skia (macOS), Skia (Linux Framebuffer), Skia (GTK)

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

ramezgerges:
GitHub Issue (If applicable): closes #17342

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI.RuntimeTests/Tests/Microsoft_UI_Xaml_Controls/Given_TreeView.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Microsoft_UI_Xaml_Controls/Given_TreeView.cs
@@ -318,6 +318,86 @@ namespace Uno.UI.RuntimeTests.Tests.Microsoft_UI_Xaml_Controls
 
 			Assert.IsFalse(TestableTreeViewItem.DraggingThrewException);
 		}
+
+		[TestMethod]
+#if !HAS_INPUT_INJECTOR
+		[Ignore("InputInjector is only supported on skia")]
+#elif HAS_UNO && !HAS_UNO_WINUI
+		[Ignore("Fails on UWP branch as mixing WUX and MUX types causes errors.")]
+#endif
+		public async Task When_TreeViewItem_Dragged_NRE2()
+		{
+			using var _ = new DisposableAction(() => TestableTreeViewItem.DraggingThrewException = false);
+			var treeView = new TreeView
+			{
+				ItemTemplate = new DataTemplate(() =>
+				{
+					var tvi = new TestableTreeViewItem();
+					tvi.SetBinding(TreeViewItem.ItemsSourceProperty, new Binding("Items"));
+					tvi.SetBinding(ContentControl.ContentProperty, new Binding("Label"));
+					return tvi;
+				}),
+				ItemsSource = new ObservableCollection<TestTreeNodeModel>
+				{
+					new TestTreeNodeModel("Root 1")
+					{
+						Items =
+						{
+							new TestTreeNodeModel("Child 1.1"),
+							new TestTreeNodeModel("Child 1.2"),
+						}
+					},
+					new TestTreeNodeModel("Root 2")
+					{
+						Items =
+						{
+							new TestTreeNodeModel("Child 2.1"),
+						}
+					},
+				}
+			};
+
+			await UITestHelper.Load(treeView);
+
+			var injector = InputInjector.TryCreate() ?? throw new InvalidOperationException("Failed to init the InputInjector");
+			using var mouse = injector.GetMouse();
+
+			treeView.RootNodes[0].IsExpanded = true;
+			await WindowHelper.WaitForIdle();
+
+			var ttv = treeView.TransformToVisual(null);
+
+			var test = new[]
+			{
+				("move", new Point(103, 29)),
+				("press", new Point()),
+				("move", new Point(103, 36)),
+				("move", new Point(103, 40)),
+				("move", new Point(103, 47)),
+				("release", new Point()),
+			};
+
+
+			foreach (var (action, position) in test)
+			{
+				if (action == "press")
+				{
+					mouse.Press();
+				}
+				else if (action == "release")
+				{
+					mouse.Release();
+				}
+				else
+				{
+					Assert.AreEqual("move", action);
+					mouse.MoveTo(ttv.TransformPoint(position), 1);
+				}
+				await WindowHelper.WaitForIdle();
+			}
+
+			Assert.IsFalse(TestableTreeViewItem.DraggingThrewException);
+		}
 #endif
 
 		[TestMethod]
--- a/src/Uno.UI/UI/Xaml/UIElement.Pointers.cs
+++ b/src/Uno.UI/UI/Xaml/UIElement.Pointers.cs
@@ -733,7 +733,11 @@ namespace Microsoft.UI.Xaml
 			// but actually removing them from the visual tree needs a layout cycle. Without waiting here,
 			// we can also get a DragEnter event on one of the to-be-collapsed containers in the same
 			// pointer event. This crashes the dragging logic.
-			await NativeDispatcher.Main.EnqueueAsync(() => { }, NativeDispatcherPriority.Idle);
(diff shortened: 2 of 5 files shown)


src/Uno.UI.RuntimeTests/Tests/Microsoft_UI_Xaml_Controls/Given_TreeView.cs
```diff
@@ -318,6 +318,86 @@ public async Task When_TreeViewItem_Dragged_NRE()
 
 			Assert.IsFalse(TestableTreeViewItem.DraggingThrewException);
 		}
+
+		[TestMethod]
+#if !HAS_INPUT_INJECTOR
+		[Ignore("InputInjector is only supported on skia")]
+#elif HAS_UNO && !HAS_UNO_WINUI
+		[Ignore("Fails on UWP branch as mixing WUX and MUX types causes errors.")]
```
Youssef1313, line 326: I think it should be testable on UWP tree. What kind of error do you get?

ramezgerges: Not sure, it was a long time ago (this is a copy from the test above it). It's most likely a mistake in the test itself, but it seems like it would be a waste of time to find where the problem is exactly.

src/Uno.UI/UI/Xaml/UIElement.Pointers.cs
```diff
@@ -727,6 +728,17 @@ private async Task<DataPackageOperation> StartDragAsyncCore(PointerPoint pointer
...
+			// We need to give a chance for  for layout updates, etc. This is particularly problematic with TreeView
+			// dragging where the DragStarting event on the TreeView will "internally" collapse some nodes,
+			// but actually removing them from the visual tree needs a layout cycle. Without waiting here,
+			// we can also get a DragEnter event on one of the to-be-collapsed containers in the same
+			// pointer event. This crashes the dragging logic.
+			// The way WinUI asynchronously handles StartDragAsync (relevant code is in AutomaticDragHelper::AutomaticDragHelper::HandlePointerMovedEventArgs)
+			// is not very clear and the way it interacts with layout timing is not similar at all to
+			// what we do in Uno, so the closest thing is to wait for Low/Idle here.
+			// Note that waiting for Idle causes some unrelated problems in ListView dragging, so we choose Low.
+			await NativeDispatcher.Main.EnqueueAsync(() => { }, NativeDispatcherPriority.Low);
```
dr1rrb, line 740: > the DragStarting event on the TreeView will "internally" collapse some nodes
FYI, we are handling the visual tree modification [using the `PointerEventDispatchResult`](https://github.com/unoplatform/uno/blob/9c24c59ce294f335a7c6884e45d8415638fc9f36/src/Uno.UI/UI/Xaml/UIElement.Pointers.cs#L243) I think this is similar to what you are facing here (but it's for managed pointers only).

> we can also get a DragEnter event on one of the to-be-collapsed containers
Awaiting a somehow random delay here does not make sure that it won't be the case anymore, especially if we have "recursed" updates (like altering again the visual tree in `OnLoaded` or `OnUnloaded` of one of your containers). At some point I would simply invoke `UIElement.UpdateLayout`. Is it something you already tried?

ramezgerges: Yes, `UpdateLayout` was already a fix I added in https://github.com/unoplatform/uno/pull/16762 (and forgot about it :/) but it looked very hacky to me. The delay here is not random. It's a delay with low enough priority to allow an invalidated visual tree to get layouted.

On WinUI, if you subscribe to `DragOver` on the dragging target and then do something like this on the dragging source:
```csharp
private void UIElement_OnDragStarting(UIElement sender, DragStartingEventArgs args)
    {
        Console.WriteLine("dragstarting");
        e.InvalidateMeasure(); // e is a UIElement in the visual tree with an overriden MeasureOverride that prints "measured"
        DispatcherQueue.TryEnqueue(DispatcherQueuePriority.High, () => Console.WriteLine("high"));
        DispatcherQueue.TryEnqueue(DispatcherQueuePriority.Normal, () => Console.WriteLine("normal"));
        DispatcherQueue.TryEnqueue(DispatcherQueuePriority.Low, () =>
        {
            Console.WriteLine("low");
            DispatcherQueue.TryEnqueue(DispatcherQueuePriority.Low, () => Console.WriteLine("lowlow"));
        });
    }
```

When you drag an element, you get this:
```
dragstarting
high
normal
low
measured
lowlow
dragover // this comes from the dragging target
dragover // this comes from the dragging target
dragover // this comes from the dragging target
dragover // this comes from the dragging target
...
```

So it's not only that it's updating the visual tree. It's also allowing any operations enqueued on the dispatcher to go through first.
> FYI, we are handling the visual tree modification [using the PointerEventDispatchResult](https://github.com/unoplatform/uno/blob/9c24c59ce294f335a7c6884e45d8415638fc9f36/src/Uno.UI/UI/Xaml/UIElement.Pointers.cs#L243) I think this is similar to what you are facing here (but it's for managed pointers only).

Totally forgot about that one, thanks! But hittesting is not the problem here. The problem is NOT that we incorrectly hit-test an element that's no longer in the visual tree. The problem is that the collapsed element is still in the tree even though the TreeView removed the node from its internal representation. This is fine and expected because the visual tree will be updated to match this internal representation on the next `MeasureOverride`. This is all behaving as intended. It only becomes problematic when you can a SINGULAR PointerMove event that both:
1. fires a DragStarting event on the dragging source (i.e. this PointerMove event is just enough to move out of the "drag box". This causes the TreeView to collapse and remove the dragged TreeViewItem and all its children from its internal representation. 
2. also fires a DragEnter+DragOver on one of the TreeViewItems that was just "collapsed internally". The callback of DragOver in TreeViewItem still assumes that it (and some other nodes) are still in the internal representation of the parent TreeView.

So the fix is to somehow leave some breathing room after 1. so that 2. no longer happens. The closest fix I could find is the one in this PR.

dr1rrb: > but it looked very hacky to me

This is not a hack, it's instead a good way to stabilize the tree after an event has been raised and user can have updated the tree. And BTW this is how the `ItemsRepeater` works in some circumstances.

> But hittesting is not the problem here

Yes, it was only an option to have a way to defer only if needed.


And anyway looking at your "events log" on WinUI, you change is probably the most appropriate indeed.

ajpinedam (approved):

PR #17558, opened 2024-07-16
Workaround: (iOS) animation crash when navigating

ajpinedam:

## PR Type

- Bugfix

A few crashes on iOS apps have been reported when navigating. We saw in the logs this crash is related to animations. We could not reproduce the error, but after enabling nullable, we found a few places where we could prevent a crash.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Media/Animation/Animators/GPUFloatValueAnimator.iOSmacOS.cs
+++ b/src/Uno.UI/UI/Xaml/Media/Animation/Animators/GPUFloatValueAnimator.iOSmacOS.cs
@@ -306,7 +306,7 @@ namespace Microsoft.UI.Xaml.Media.Animation
 		{
 			return CreateCoreAnimation(view, "opacity", value => new NSNumber(value));
 		}
-		private UnoCoreAnimation InitializeTranslateCoreAnimation(TranslateTransform transform, IBindingItem animatedItem)
+		private UnoCoreAnimation? InitializeTranslateCoreAnimation(TranslateTransform transform, IBindingItem animatedItem)
 		{
 			if (animatedItem.PropertyName.Equals("X")
 				|| animatedItem.PropertyName.Equals(TranslateTransformX)
@@ -326,7 +326,7 @@ namespace Microsoft.UI.Xaml.Media.Animation
 			}
 		}
 
-		private UnoCoreAnimation InitializeRotateCoreAnimation(RotateTransform transform, IBindingItem animatedItem)
+		private UnoCoreAnimation? InitializeRotateCoreAnimation(RotateTransform transform, IBindingItem animatedItem)
 		{
 			if (animatedItem.PropertyName.Equals("Angle")
 				|| animatedItem.PropertyName.Equals(RotateTransformAngle)
@@ -340,7 +340,7 @@ namespace Microsoft.UI.Xaml.Media.Animation
 			}
 		}
 
-		private UnoCoreAnimation InitializeScaleCoreAnimation(ScaleTransform transform, IBindingItem animatedItem)
+		private UnoCoreAnimation? InitializeScaleCoreAnimation(ScaleTransform transform, IBindingItem animatedItem)
 		{
 			if (animatedItem.PropertyName.Equals("ScaleX")
 				|| animatedItem.PropertyName.Equals(ScaleTransformX)
@@ -360,7 +360,7 @@ namespace Microsoft.UI.Xaml.Media.Animation
 			}
 		}
 
-		private UnoCoreAnimation InitializeSkewCoreAnimation(SkewTransform transform, IBindingItem animatedItem)
+		private UnoCoreAnimation? InitializeSkewCoreAnimation(SkewTransform transform, IBindingItem animatedItem)
 		{
 			// We need to review this.  This won't play along if other transforms are happening at the same time since we are animating the whole transform
 			if (transform.View is not _View view)
@@ -370,7 +370,7 @@ namespace Microsoft.UI.Xaml.Media.Animation
 					this.Log().Warn("The View property of the SkewTransform is null.");
 				}
 
-				throw new InvalidOperationException("The View property of the SkewTransform is null.");
+				return null;
 			}
 
 			if (animatedItem.PropertyName.Equals("AngleX")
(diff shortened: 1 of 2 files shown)


src/Uno.UI/UI/Xaml/Media/Animation/Animators/GPUFloatValueAnimator.iOSmacOS.cs
```diff
@@ -37,10 +38,10 @@ internal class GPUFloatValueAnimator : IValueAnimator
 		private float _to;
 		private float _from;
 		private long _duration;
-		private IEnumerable<IBindingItem> _bindingPath;
-		private FloatValueAnimator _valueAnimator;
-		private UnoCoreAnimation _coreAnimation;
-		private IEasingFunction _easingFunction;
+		private IEnumerable<IBindingItem>? _bindingPath;
```
Youssef1313, line 41: Can this really be null? It looks like it's set in the constructor?

src/Uno.UI/UI/Xaml/Media/Animation/Animators/GPUFloatValueAnimator.iOSmacOS.cs
```diff
@@ -37,10 +38,10 @@ internal class GPUFloatValueAnimator : IValueAnimator
...
+		private FloatValueAnimator? _valueAnimator;
```
Youssef1313, line 42: This one is only null after dispose. It's probably an issue if the object is dealt with after it was disposed.

Usually, these cases are annotated as non-null, then in `Dispose`, it's done with null suppression, `_valueAnimator = null!`.

src/Uno.UI/UI/Xaml/Media/Animation/Animators/GPUFloatValueAnimator.iOSmacOS.cs
```diff
@@ -421,12 +441,23 @@ private UnoCoreAnimation InitializeCompositeCoreAnimation(CompositeTransform tra
...
+
+			if (view is null)
+			{
+				if (this.Log().IsEnabled(Uno.Foundation.Logging.LogLevel.Warning))
+				{
+					this.Log().Warn("The View property of the Transform is null.");
+				}
+
+				throw new InvalidOperationException("The View property of the Transform is null.");
+			}
```
Youssef1313, line 459: If we are disallowing null here, I think the parameter shouldn't be annotated as nullable, and callers should make sure to not pass null

src/Uno.UI/UI/Xaml/Media/Animation/Animators/GPUFloatValueAnimator.iOSmacOS.cs
ajpinedam: We are making it null on the dispose method https://github.com/unoplatform/uno/blob/master/src/Uno.UI/UI/Xaml/Media/Animation/Animators/GPUFloatValueAnimator.iOSmacOS.cs#L557. Should we change it there?

src/Uno.UI/UI/Xaml/Media/Animation/Animators/GPUFloatValueAnimator.iOSmacOS.cs
ajpinedam: Alright, I will follow this. 

"This one is only null after dispose. It's probably an issue if the object is dealt with after it was disposed."
I believe that's what's happening, that for some reason, during the navigation, we are dealing with a disposed Animation object and reaching a point where some fields are already `null`

src/Uno.UI/UI/Xaml/Media/Animation/Animators/GPUFloatValueAnimator.iOSmacOS.cs
Youssef1313: Ah, missed that line. In that case, I think Dispose should do `_bindingPath = null!` and keep the field non-nullable. Attempts to operate on a disposed object look fundamentally wrong and may be the actual root cause.

**However,** depending on how complex is it to fix the use-after-dispose issue, it may still be acceptable to tolerate nulls. It may be worth adding a comment though

src/Uno.UI/UI/Xaml/Media/Animation/Animators/GPUFloatValueAnimator.iOSmacOS.cs
Youssef1313: I think that's the real bug then, the object is used in a disposed state. But as mentioned in https://github.com/unoplatform/uno/pull/17558#discussion_r1679814958, we could tolerate for use-after-dispose if it ends up complicated to fix it the right way

src/Uno.UI/UI/Xaml/Media/Animation/Animators/GPUFloatValueAnimator.iOSmacOS.cs
ajpinedam: Updated. Making the field non-nullable and validating for null on the constructor should work

src/Uno.UI/UI/Xaml/Media/Animation/Animators/GPUFloatValueAnimator.iOSmacOS.cs
ajpinedam: Updated

src/Uno.UI/UI/Xaml/Media/Animation/Animators/GPUFloatValueAnimator.iOSmacOS.cs
ajpinedam: @Youssef1313, thanks for your input. Yes, this is a workaround since we have not been able to reproduce the actual error but found that by doing these changes we could prevent crashes or make it crash in a more evident place

jeromelaban: An additional information was provided for the nullref, it's failing at : 

https://github.com/unoplatform/uno/blob/161c3c1959ecaa00dd55c658d099b4c49f6feeed/src/Uno.UI/UI/Xaml/Media/Animation/Animators/GPUFloatValueAnimator.iOSmacOS.cs#L436-L438

jeromelaban:
@ajpinedam the build is failing: https://dev.azure.com/uno-platform/Uno%20Platform/_build/results?buildId=134453&view=logs&j=1e7f85e3-6361-5c87-810d-f02571af2764&t=7e6f2d69-ff70-5fe7-19ca-d07ee2ed79de&l=339

jeromelaban:
@mergifyio backport release/stable/5.3

jeromelaban:

PR #17827, opened 2024-08-04
perf: DependencyObjectStore

Youssef1313:

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_FrameworkElement_And_Leak.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_FrameworkElement_And_Leak.cs
@@ -420,8 +420,10 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml
 				var SUT = new Grid();
 				root.Children.Add(SUT);
 
+#if UNO_HAS_UIELEMENT_IMPLICIT_PINNING
 				Assert.IsFalse((SUT as IDependencyObjectStoreProvider).Store.AreHardReferencesEnabled);
 				Assert.IsNotNull(SUT.GetParent());
+#endif
 
 				TestServices.WindowHelper.WindowContent = root;
 				await TestServices.WindowHelper.WaitForIdle();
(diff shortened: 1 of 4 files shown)


Youssef1313:
@MartinZikmund This PR was probably ready to merge as it was green. But the build is stale as it's quite old and there are now conflicts. The idea here is that using a WeakReference should only be required on iOS where circular references are problematic. Other platforms shouldn't use a WeakReference which has a slight performance impact.

MartinZikmund:
@ebariche is this perf PR still going to be included?

jeromelaban:
This change seems to be removing hard reference support, and I'm not sure it's worth it at this time.

MartinZikmund:
Closing this as superseded — thanks @Youssef1313, the idea landed, just by a different route.

Since this PR was opened, `DependencyObjectStore` was dissolved into `DependencyObject` itself (on `feature/breakingchanges`). The store and its owner are now the same object, so `ActualInstance` is literally `=> this` — which gets what `_hardOriginalObjectRef` was after, without needing a second field at all.

The other half of this PR — `DependencyPropertyDetailsCollection` holding its owner strongly instead of through a pooled `ManagedWeakReference` — landed on `dev/mazi/dostore-strongself`:

- `refactor: Type the details-collection owner as DependencyObject`
- `refactor: Remove the inert store hard-reference toggles` (the `TryEnableHardReferences`/`DisableHardReferences` pair on that collection)
- `test: Pin DependencyObject collectibility with a strong store cycle` — a `When_Store_Populated_Then_Owner_Collected` regression test that fails if the collection, or a closure capturing it, ever escapes its owner and pins the whole graph

The dead code this PR cleaned up as a consequence — the `if (actualInstanceAlias != null)` wrapper in `InnerSetValue` and its `else { Parent = null; }` fallback, plus the `!` suppressions — was picked up there too, in `refactor: Make ActualInstance non-nullable`.

@jeromelaban, on the hard-reference concern from your review: it is preserved. Only the **self** hard reference went away, and only because it became inert once the store and the DO were the same object. The **parent** hard reference is untouched — `_hardParentRef`, `FeatureConfiguration.DependencyObject.IsStoreHardReferenceEnabled` and `AreHardReferencesEnabled` all still work, and a loaded element still pins its parent.

One note on the original motivation ("a `WeakReference` should only be required on iOS where circular references are problematic"): that constraint no longer applies now that the native iOS UI layer is gone, so the strong reference is unconditional rather than platform-gated.

PR #17948, opened 2024-08-15
fix(skia): make UIElement.HitTest strictly stronger than UIElement.IsViewHit

ramezgerges:
GitHub Issue (If applicable): followup to #16201

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/Border/Border.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Border/Border.cs
@@ -322,17 +322,7 @@ public partial class Border : FrameworkElement
 
 	internal override bool CanHaveChildren() => true;
 
-	internal override bool IsViewHit() => IsViewHitImpl(this);
-
-	internal static bool IsViewHitImpl(FrameworkElement element)
-	{
-		_Debug.Assert(element is Panel
-			|| element is Border
-			|| element is ContentPresenter
-		);
-
-		return element.Background != null;
-	}
+	internal override bool IsViewHit() => ((IBorderInfoProvider)this).IsViewHit();
 
 #if !UNO_HAS_BORDER_VISUAL
 	private void UpdateBorder() => BorderRenderer.Update();
--- a/src/Uno.UI/UI/Xaml/Controls/Popup/PopupPanel.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Popup/PopupPanel.cs
@@ -296,4 +296,7 @@ internal partial class PopupPanel : Panel
 
 		return Popup?.IsLightDismissEnabled ?? false;
 	}
+
+	// This shouldn't be needed, see the above comment in IsViewHit
+	internal override bool HitTest(Point point) => IsViewHit();
 }
--- a/src/Uno.UI/UI/Xaml/Controls/Popup/PopupPanel.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Popup/PopupPanel.cs
@@ -297,6 +297,9 @@ internal partial class PopupPanel : Panel
 		return Popup?.IsLightDismissEnabled ?? false;
 	}
 
-	// This shouldn't be needed, see the above comment in IsViewHit
+	// This shouldn't be needed, but due to the way we implement Popups, we might
+	// need to make Popups not hit-testable even if the PopupPanel has a Background.
+#if __SKIA__
 	internal override bool HitTest(Point point) => IsViewHit();
+#endif
 }
--- a/src/Uno.UI/UI/Xaml/Media/VisualTreeHelper.cs
+++ b/src/Uno.UI/UI/Xaml/Media/VisualTreeHelper.cs
@@ -501,12 +501,6 @@ namespace Microsoft.UI.Xaml.Media
 				clippingBounds = clippingBounds.IntersectWith(transformToElement.Transform(clip)) ?? default;
 			}
 			TRACE($"- clipping (absolute): {clippingBounds.ToDebugString()}");
-
-			// The region where the current element draws itself.
-			// Be aware that children might be out of this rendering bounds if no clipping defined.
-			// This is expressed in the window (absolute) coordinate space.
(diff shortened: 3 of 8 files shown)


src/Uno.UI/UI/Xaml/Controls/Popup/PopupPanel.cs
```diff
@@ -296,4 +296,10 @@ internal override bool IsViewHit()
 
 		return Popup?.IsLightDismissEnabled ?? false;
 	}
+
+	// This shouldn't be needed, but due to the way we implement Popups, we might
+	// need to make Popups not hit-testable even if the PopupPanel has a Background.
+#if __SKIA__
+	internal override bool HitTest(Point point) => IsViewHit();
+#endif
 }
```
Youssef1313, line 305: Not really feeling comfortable with this. I'd personally prefer if we still keep the HitTestability.Visible check in `SearchDownForTopMostElementAt`. Then at some point we can try to remove IsViewHit completely, at least on Skia. That's kinda my personal preference.

src/Uno.UI.Composition/Composition/Visual.skia.cs
```diff
@@ -344,6 +345,9 @@ private PaintingSession CreateLocalSession(in PaintingSession parentSession)
 		return session;
 	}
 
+	/// <remarks>This does NOT take the clipping into account.</remarks>
+	internal virtual bool HitTest(Point point) => false;
```
Youssef1313, line 349: Somewhat weird to be in Visual but not all visuals have proper override. Maybe not reachable for those visuals? In this case maybe we can throw here or at least add a failing assert

src/Uno.UI/UI/Xaml/Controls/TextBlock/TextVisual.skia.cs
```diff
@@ -34,5 +34,7 @@ internal override void Paint(in PaintingSession session)
 				owner.Inlines.Draw(in session);
 			}
 		}
+
+		internal override bool HitTest(Point point) => new Rect(Offset.X, Offset.Y, Size.X, Size.Y).Contains(point);
```
Youssef1313, line 38: I don't see how this override is reachable?

src/Uno.UI.Composition/Composition/Visual.skia.cs
ramezgerges: Yeah, you're right. Definitely not reachable. I thought I would move the logic to be inside TextVisual instead of the TextBlock itself, but because the TextVisual is not `TextBlock.Visual`, but a child of it, it doesn't work out.

Youssef1313: In that case let's keep it specifically to ShapeVisual

src/Uno.UI/UI/Xaml/Media/VisualTreeHelper.cs
```diff
@@ -662,12 +656,19 @@ internal static (UIElement? element, Branch? stale) SearchDownForTopMostElementA
...
 			// and the position is in actual bounds(which might be different than the clipping bounds)
 #if __SKIA__
-			if (element.HitTest(transformToElement.Inverse().Transform(testPosition)) && renderingBounds.Contains(testPosition))
+			if (element.HitTest(transformToElement.Inverse().Transform(testPosition)))
 #else
-
 			if (elementHitTestVisibility == HitTestability.Visible && renderingBounds.Contains(testPosition))
 #endif
 			{
+				// UIElement.HitTest should be a stronger condition than elementHitTestVisibility == HitTestability.Visible
```
Youssef1313, line 664: UIElement.HitTest just does Visual hit testing, which isn't aware of whether the UIElement's background is NULL or not. Aren't we creating transparent background in both cases? Also, what if the user gets the Visual and modifies the background, in this case they are not the same. I tend to think we should really restore `elementHitTestVisibility == HitTestability.Visible` check

dr1rrb: Not sure if it's the case so far, but the `Visual` **could** be aware if ~the background is `null`~ a pixel has not been painted, or not if it has been painted using with transparent color. If so, the `elementHitTestVisibility == HitTestability.Visible` check would indeed only be about what hit-testability we push to the native element and therefor should be less restrictive than the `IsViewHit()`.

src/Uno.UI/UI/Xaml/Controls/Border/Border.cs
```diff
@@ -322,17 +322,7 @@ protected override void OnBackgroundChanged(DependencyPropertyChangedEventArgs e
...
-	internal static bool IsViewHitImpl(FrameworkElement element)
-	{
-		_Debug.Assert(element is Panel
-			|| element is Border
-			|| element is ContentPresenter
-		);
-
-		return element.Background != null;
-	}
+	internal override bool IsViewHit() => ((IBorderInfoProvider)this).IsViewHit();
```
Youssef1313, line 325: This may change behavior on non-Skia. It was already not correct behavior, but now will be incorrect in a different way, I think.

ramezgerges: We could guard this with `#if __SKIA__`, but the change makes sense in general I think, so I would say it's better to change the behaviour on non-skia targets even if it's still not accurate.

src/Uno.UI/UI/Xaml/Controls/Popup/PopupPanel.cs
Youssef1313: Relying on Visual here is probably buggy due to bugs in popup implementation itself that will need to be their own separate work. In this case, `IsViewHit` can return false while the visual can actually hit test. This basically makes me more convinced into keeping the check in the VisualTreeHelper as done with #17973. My opinion here is that #17973 is enough for now. It's not worth the hassle around trying to force HitTest to be a stronger condition.

For now, the only place we know it's not stronger is PopupPanel, and this PR forces it to be strongly by basically just calling IsViewHit.

(btw, if we proceeded with this PR, it will be a better idea to `return IsViewHit() && base.HitTest(point)` I think)

PR #17961, opened 2024-08-16
refactor: Cleanup hit test coercion

Youssef1313:

## PR Type

## PR Checklist

## Other information


MartinZikmund (approved):

MartinZikmund (approved):

src/Uno.UI/UI/Xaml/UIElement.Pointers.cs
```diff
@@ -268,7 +273,34 @@ internal struct PointerEventDispatchResult
...
-			return HitTestVisibility;
+			var element = this;
+			while (element is not null)
+			{
+				if (!element.IsEnabledOverride() || !element.IsHitTestVisible)
+				{
+					return HitTestability.Collapsed;
+				}
+
+				element = GetParentUIElement(element);
```
dr1rrb, line 284: I don't see any interest to walk the tree up instead of relying on a pre-computed property (that is pushed to native elements on wasm for instance). This will be perform for each pointer so I would assume this change would have a negative perf impact.

src/Uno.UI/UI/Xaml/DragDrop/DropUITarget.cs
```diff
@@ -116,7 +102,7 @@ public IAsyncOperation<DataPackageOperation> DropAsync(CoreDragInfo dragInfo)
 			var target = VisualTreeHelper.HitTest(
 				dragInfo.Position,
 				xamlRoot,
-				getTestability: GetDropHitTestability,
+				isForDrop: true,
```
dr1rrb, line 105: This reduce the flexibilty of the `HitTest` method, not sure about the gain?

src/Uno.UI/UI/Xaml/UIElement.Pointers.cs
dr1rrb: It should not: the down propagation is computed only when an element is being added into the tree or something that changes testability is updated... 

Even if this approach is closer to winUI, we need to pre-compute the testability for native elements, so I don't see any interest to try to change the behavior.

src/Uno.UI/UI/Xaml/DragDrop/DropUITarget.cs
Youssef1313: @dr1rrb I can try to refactor it back to gain a bit of flexibility back

PR #18093, opened 2024-08-31
perf: Experiment with different approach for DP storage

Youssef1313:

## PR Type

Too much allocations happening inside DependencyPropertyDetailsCollection for offsets and entries arrays.

## What is the new behavior?

This PR experiments an approach that's more memory friendly.

Here is a comparison during SamplesApp startup.

image

Most significant part of the diff:

image

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/DependencyPropertyDetailsCollection.cs
+++ b/src/Uno.UI/UI/Xaml/DependencyPropertyDetailsCollection.cs
@@ -2,6 +2,7 @@
 
 using System;
 using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
 using Microsoft.UI.Xaml.Data;
 using Uno.Buffers;
 using Uno.UI.DataBinding;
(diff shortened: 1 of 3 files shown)


ajpinedam (approved):
The change looks good, as do the numbers you shared.

src/Uno.UI/UI/Xaml/DependencyPropertyDetailsCollection.cs
```diff
@@ -22,15 +23,10 @@ partial class DependencyPropertyDetailsCollection : IDisposable
 		private DependencyPropertyDetails? _dataContextPropertyDetails;
 		private DependencyPropertyDetails? _templatedParentPropertyDetails;
 
-		private readonly static ArrayPool<short> _offsetsPool = ArrayPool<short>.Shared;
-		private readonly static LinearArrayPool<DependencyPropertyDetails?> _pool = LinearArrayPool<DependencyPropertyDetails?>.CreateAutomaticallyManaged(BucketSize, 16);
+		private DependencyPropertyDetails[] _entries = Array.Empty<DependencyPropertyDetails>();
```
jeromelaban, line 26: This change basically removes the pooling to allocate every time. This is curious that this leads to improvements.

Youssef1313: Pooling doesn't seem very much beneficial in this case, I think. Those arrays are more commonly long living, while pooling is more useful for short living objects.

Youssef1313:
NOTES:

- From my profiling, I don't see much of a regression with regards to CPU time, but the memory usage is a lot better with this PR
- If we think this approach has an impact on CPU time, we should consider one (or a combination) of these alternatives:
   - Adding a new DP option to specify the storage mode. The default (IMO) would be the new approach which is optimized more for memory, and an option "OptimizeForCPU" is added to use the old approach for REALLY very commonly accessed DPs.
   - Depending on device available memory, select the mode
   - Add a FeatureConfiguration for users to specify the (default) storage mode, based on their preferences.

PR #18231, opened 2024-09-17
fix: Use Rgba8888 for Skia/macOS/metal

spouliot:

https://github.com/ScottPlot/ScottPlot/issues/4257

## PR Type

- Bugfix

Metal-based rendering was using `Bgra888` and copying `Rgba888` memory surface into it is swapping the `R` and `B` channels.

## What is the new behavior?

Metal-based rendering was using `Rgba888`

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI.Runtime.Skia.MacOS/MacOSWindowHost.cs
+++ b/src/Uno.UI.Runtime.Skia.MacOS/MacOSWindowHost.cs
@@ -108,7 +108,7 @@ internal class MacOSWindowHost : IXamlRootHost, IUnoKeyboardInputSource, IUnoCor
 
 		// we can't cache anything since the texture will be different on next calls
 		using var target = MacOSMetalRenderer.CreateTarget(_context!, nativeWidth, nativeHeight, texture);
-		using var surface = SKSurface.Create(_context, target, GRSurfaceOrigin.TopLeft, SKColorType.Bgra8888);
+		using var surface = SKSurface.Create(_context, target, GRSurfaceOrigin.TopLeft, SKColorType.Rgba8888);
 
 		surface.Canvas.Scale(scale, scale);
 
--- a/src/Uno.UI.Runtime.Skia.MacOS/UnoNativeMac/UnoNativeMac/UNOMetalViewDelegate.m
+++ b/src/Uno.UI.Runtime.Skia.MacOS/UnoNativeMac/UnoNativeMac/UNOMetalViewDelegate.m
@@ -19,7 +19,7 @@
         _device = mtkView.device;
         self.queue = [_device newCommandQueue];
         
-        mtkView.colorPixelFormat = MTLPixelFormatBGRA8Unorm;
+        mtkView.colorPixelFormat = MTLPixelFormatRGBA8Unorm;
         mtkView.depthStencilPixelFormat = MTLPixelFormatDepth32Float_Stencil8;
         mtkView.sampleCount = 1;
         // this property has no effect on x86_64, only on arm64, and is required for sampling (which acrylicbrush does)


MartinZikmund (approved):

agneszitte:
@jeromelaban should we backport this for 5.4?

agneszitte:
We would need to test these changes on a Mac intel to decide if the changes are ok to backport to 5.4

ArchieCoder:
@spouliot @agneszitte Thanks for fixing this issue, I was not aware. Regarding the backport, does it mean the fix will be available in a future big update? 5.4.5 was released yesterday. I don't know the merge structure.

jeromelaban:
@ArchieCoder We have yet to validate the change, but you can validate that it does what you need using 5.5-dev builds.

ArchieCoder:
@spouliot I tested with 5.5.0-dev.66 and I have the same effect as before (screenshot shows the app and the global.json)

Screenshot 2024-09-27 at 10 29 08 AM

spouliot:
Unexpected. I'll check if `5.5.0-dev.66` has the fix...

Just in case, did you do a clean after updating ? are you using a arm64 or Intel Mac ? thanks!

spouliot:
So looking into `5.5.0-dev.66` @ https://nuget.info/packages/Uno.Sdk/5.5.0-dev.66
points to `UnoVersion* | 5.5.0-dev.512`, meaning https://nuget.info/packages/Uno.WinUI/5.5.0-dev.512
which points to commit 0f422bf564aa567cd5363bd07cf3616c1cb5fca7 
which is newer than my fix https://github.com/unoplatform/uno/commit/0dfe51edd0d3a9b8e4b6dce3d9f86806c61ffc9a
so it should be included in your build.

Make sure a `restore` was made, just in case. I'll double check it on my side asap...

ArchieCoder:
@spouliot How do I clean or do a restore? I don't even see this in VS Code. I deleted the folders obj/bin. It had no effect. I have arm64.

spouliot:
> How do I clean or do a restore? 

@ArchieCoder usually `dotnet restore` [1] .

> I don't even see this in VS Code.

That's done automatically by our VS Code extension - so that should not be an issue.

> I have arm64.

Thanks, I'll be checking stuff on my side...

[1] but `nuget restore` or `msbuild /r` are older ways to restore (but don't use them anymore).

spouliot:
@ArchieCoder I can duplicate with the same packages... 😞 weird

ArchieCoder:
@spouliot sad for the bug but happy that you see the same thing

agneszitte:
> @ArchieCoder I can duplicate with the same packages... 😞 weird

@spouliot that means that more fixes will be needed for https://github.com/ScottPlot/ScottPlot/issues/4257 right?

spouliot:
> @spouliot that means that more fixes will be needed for https://github.com/ScottPlot/ScottPlot/issues/4257 ?

@agneszitte yes

spouliot:
Ok, so previous, _similar_ issues have tinted my glasses. Unlike them this happens on **both** software and hardware(metal) rendering - which otherwise are working normally.

Still the error does not seems to come from ScottPlot. Took m a while to convince myself but  I can now duplicate the issue with other samples that, like ScottPlot, also use `SKXamlCanvas`.

Note: I can't be 100% sure but I _probably_ executed the Catalyst (instead of desktop) version of the app when testing my previous fix :-(

spouliot:
Could be related to https://github.com/mono/SkiaSharp/pull/2918
c.c. @Youssef1313

Youssef1313:
@spouliot Just to be sure, could you actually be hitting the SkiaSharp bug, then your fix will break when run against fixed SkiaSharp version?

IIRC, the version we run on is 2.88.7 which doesn't have the fix, I think?

Youssef1313:
We need to align our usages everywhere, really. Otherwise it's getting too problematic

spouliot:
> IIRC, the version we run on is 2.88.7 which doesn't have the fix, I think?

I'm running with 2.88.8 but I'll double check if the fix is present (or not).

> We need to align our usages everywhere, really. Otherwise it's getting too problematic

Yes. There's some platforms that only support some format(s), but most do support whatever is requested.

Youssef1313:
Per release notes, it seems the fix is included in 2.88.9 p1.1

https://github.com/mono/SkiaSharp/releases/tag/v2.88.9-preview.1.1

spouliot:
So `2.88.9 p1.1` is the fix... (compiled from git)
I'll review if other of my local changes are needed since there's a mix of surface types.

PR #18327, opened 2024-09-27
fix: Fix UpdateFileAsync requesting HR before VS detects file change

dr1rrb:
## Bugfix
Fix UpdateFileAsync requesting HR before VS detects file change

HR is requested before VS detects file update

## What is the new behavior?
We wait hopefully long enough after file change to trigger HR on VS, *AND* if a no-changes is reported, we re-trigger a second HR on VS.

## PR Checklist

### DIFF
--- a/src/Uno.UI.RemoteControl.Server.Processors/HotReload/ServerHotReloadProcessor.cs
+++ b/src/Uno.UI.RemoteControl.Server.Processors/HotReload/ServerHotReloadProcessor.cs
@@ -28,7 +28,7 @@ namespace Uno.UI.RemoteControl.Host.HotReload
 	{
 		private FileSystemWatcher[]? _watchers;
 		private CompositeDisposable? _watcherEventsDisposable;
-		private IRemoteControlServer _remoteControlServer;
+		private readonly IRemoteControlServer _remoteControlServer;
 
 		public ServerHotReloadProcessor(IRemoteControlServer remoteControlServer)
 		{
@@ -219,6 +219,7 @@ namespace Uno.UI.RemoteControl.Host.HotReload
 			private ImmutableHashSet<string> _filePaths;
 			private int /* HotReloadResult */ _result = -1;
 			private CancellationTokenSource? _deferredCompletion;
+			private int _noChangesRetry;
 
 			public long Id { get; } = Interlocked.Increment(ref _count);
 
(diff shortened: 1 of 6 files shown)


src/Uno.UI.RemoteControl.Server.Processors/HotReload/ServerHotReloadProcessor.cs
```diff
@@ -530,9 +552,41 @@ private async Task ProcessUpdateFile(UpdateFile message)
...
+				}
+
+				if (dir is null)
+				{
+					return;
+				}
+
+				var tcs = new TaskCompletionSource();
+				using var watcher = new FileSystemWatcher(dir.FullName);
+				watcher.Changed += async (snd, e) =>
```
ajpinedam, line 578: should we unsubscribe from this event at some point?

ajpinedam (approved):

src/Uno.UI.RemoteControl.Server.Processors/HotReload/ServerHotReloadProcessor.cs
nickrandolph: @ajpinedam this shouldn't be required since the FileSystemWatcher is disposable. The context of watcher only exists for the duration of the WaitForFileUpdated method, and since this method blocks waiting for the TaskCompletionSource, I think we can rely on Dispose being invoked to clean up the event handler.

src/Uno.UI.RemoteControl.Server.Processors/HotReload/ServerHotReloadProcessor.cs
```diff
@@ -530,9 +552,41 @@ private async Task ProcessUpdateFile(UpdateFile message)
...
+				{
+					if (e.FullPath.Equals(file.FullName, StringComparison.OrdinalIgnoreCase))
+					{
+						await Task.Delay(500);
+						tcs.TrySetResult();
+					}
+				};
+				watcher.EnableRaisingEvents = true;
+
+				await Task.WhenAny(tcs.Task, Task.Delay(TimeSpan.FromSeconds(2)));
```
nickrandolph, line 588: Since we're putting a timeout on how long to await this task, we should wrap this in a try-catch so we can log when the timeout is exceeded - this would aid debugging if we come across scenarios where the timeout is hit and we can't trace why HR isn't behaving

nickrandolph: I've raised this as a follow up issue: https://github.com/unoplatform/uno/issues/18335

nickrandolph (approved):

src/Uno.UI.RemoteControl.Server.Processors/HotReload/ServerHotReloadProcessor.cs
dr1rrb: For logging purpose yes it would make sense indeed. But FYI from a HR perspective, we have a second step that will re-request a HR to the IDE if the delay is too short. And only then the client might get an invalid "no changes".

About exception handling and possible crash due to an handled exception : this not really a "timeout", it's only a task that will complete at some point, it won't raise any exception (especially since there is no `CancellationToken` here). The only thing that can throw is the creation of the task itself (e.g. configuration of the `watcher`) which is already catch by the server.

jeromelaban (approved):

carldebilly (approved):
LGTM

PR #18328, opened 2024-09-28
fix: Restore net9 threading

jeromelaban:


src/Uno.UI.Dispatching/Native/NativeDispatcher.wasm.cs
```diff
@@ -2,13 +2,17 @@
...
-
+using System.Threading;
 using Uno.Foundation.Logging;
 
 namespace Uno.UI.Dispatching
 {
 	internal sealed partial class NativeDispatcher
 	{
+#if NET9_0_OR_GREATER
+		private static SynchronizationContext _jsContext = SynchronizationContext.Current;
```
Youssef1313, line 13: It feels confusing as to which SynchronizationContext is used.

We set our own here: https://github.com/unoplatform/uno/blob/52754f744fe56d7d14499e571110e36fc0592a9f/src/Uno.UI/UI/Xaml/Application.wasm.cs#L89-L90

And, on .NET 9, I think .NET runtime will attempt to use their internal `JSSynchronizationContext`

Which one is going to win?
I think .NET will only set JSSynchronizationContext if no synchronization context is already set.

So:

1. If .NET sets it first, we'll overwrite it in Application.Start
2. If we set it first, .NET will find non-null synchronization context and will not set their own.

jeromelaban: I agree, this is a try-out in order figure that specific scenario. The idea is that we do not care which one is set, as long as we explicitly use the JS one to trigger the dispatcher on order to come back on the main thead. The important part will be to capture it as early as possible, and it may be before app.wasm.cs does it.

jeromelaban:
More work is needed, let's close for now.

PR #18346, opened 2024-10-01
fix: Fix Binding that should be treated as regular property

### LINKED ISSUE #18345: Handle cases where `Binding` markup extension is treated as "regular" property value
This is related to https://github.com/unoplatform/kahua-private/issues/205

### Current behavior

```xaml
<Page x:Class="UnoApp10.MainPage"
      xmlns:local="using:UnoApp10"
      Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">
  <StackPanel>
    <local:MyCustomPanel x:Name="SUT" MyBinding="{Binding Hello2}"></local:MyCustomPanel>
  </StackPanel>
</Page>
```

```csharp
using Microsoft.UI.Xaml.Data;

namespace UnoApp10;

{
    public MainPage()
    {
        this.InitializeComponent();
    }
}

public partial class MyCustomPanel : Panel
{
    private BindingBase _myBinding;

    public BindingBase MyBinding { get => _myBinding; set => _myBinding = value; /* add breakpoint here */ }
}
```

On WinUI, this should set the binding as "regular" property, which is not the case in Uno.

**IMPORTANT NOTES:**

1. If the `MyBinding` property was a DP, then the binding markup extension will be treated regularly as a binding, not property value.
2. `MyBinding` being a DP is not simply that there is a static `MyBindingProperty`, the registration name **must** match the C# property name. So, something like `DependencyProperty.Register("_MyBinding", ...)` will not consider `MyBinding` as a DP, hence it will be regular property value.

### Expected behavior

Match WinUI

_No response_

### Workaround

_No response_

### Works on UWP/WinUI

None

### Environment

_No response_

### NuGet package version(s)

_No response_

### Affected platforms

_No response_

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

Youssef1313:
GitHub Issue (If applicable): closes #18345

## PR Type

## PR Checklist

## Other information


Youssef1313:
This PR needs IsPropMethodCall to be completed. The reason this is needed is that now `Control.TabNavigation` bindings are broken, because we are declaring it this way:

image

This means, the DP registry doesn't really have a property with name `TabNavigation`. On WinUI, the property is actually registered, and is marked as IsPropMethodCall so that it can delegate to `TabFocusNavigation`.

PR #18216, opened 2024-09-16
feat(android): GLCanvasElement

ramezgerges:
GitHub Issue (If applicable): https://github.com/unoplatform/uno/issues/9405, follows from #16621

## PR Type

What kind of change does this PR introduce?

## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)
- [ ] [Unit Tests and/or UI Tests](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [ ] Validated PR `Screenshots Compare Test Run` results.
- [ ] Contains **NO** breaking changes
- [ ] Associated with an issue (GitHub or internal) and uses the [automatic close keywords](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue).
- [ ] Commits must be following the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) specification.

## Other information

Internal Issue (If applicable):

PR #18408, opened 2024-10-08
fix(WriteableBitmap): Invalidation should cause a redrawing

ramezgerges:
GitHub Issue (If applicable): required for https://github.com/unoplatform/uno/pull/18216

## PR Type

## PR Checklist

## Other information


src/Uno.UI/UI/Xaml/Controls/Border/BorderLayerRenderer.Android.cs
```diff
@@ -43,11 +43,6 @@ partial void UpdatePlatform()
 			var newState = new BorderLayerState(drawArea.Size, _borderInfoProvider);
 			var previousLayoutState = _currentState;
 
-			if (newState.Equals(previousLayoutState))
-			{
-				return;
-			}
-
```
Youssef1313, line 50: This is concerning.

ramezgerges: Not really. The original code is more surprising actually. Just because the "layout slot" didn't change, doesn't mean that what is being drawn is the same. In fact, there is no sound way to make this sort of optimization in the BorderLayerRenderer. If an ImageSource fires `Invalidated`, then it known better. The BorderLayerRenderer (which has less information) cannot decide that the invalidation is unnecessary.

Youssef1313: In that case, BorderLayerRenderer should subscribe to `ImageSource.Invalidated` event, and set a flag `forceUpdate`. Then we do `if (forceUpdate || newState.Equals(previousLayoutState))`. That way, we fix that bug and still maintain the perf optimization (even if it's quite ugly to do so)

ramezgerges: That's almost already the case (`Update` is only called when it really needs updating) except for this:
```csharp
#if UNO_HAS_ENHANCED_LIFECYCLE
		_owner.SizeChanged += (_, _) => Update();
#else
		// Using SizeChanged on other platforms SHOULD work. But it didn't work on Android
		// for unknown reason. For now, we are using SizeChanged only on enhanced lifecycle
		// platforms where we are sure it works correctly.
		_owner.LayoutUpdated += (_, _) => Update();
#endif
```

So I'm going to add a `forceUpdate` flag, but it would be a lot cleaner to figure out why SizeChanged doesn't work and to just call `Update` properly.

ramezgerges: Actually, upon further reflection, we would be inserting `forceUpdate` in almost all calls and it would be very error-prone to accurately track which calls should `forceUpdate`. I guess the better way of doing that would be making the ` newState.Equals(previousLayoutState)` check only in the LayoutUpdated subscription.

Youssef1313: I'm not following why subscribing to ImageSource.Invalidated won't work

MartinZikmund: I think when padding changes and size remains the same, that should mean the invalidation should not be needed - as only the size of the content adjusted, but things like background/border still remain the same with the new padding. While it would be the best to `Update` only when actually needed, there seems to be just too many options that can trigger it, so the layer comparison + explicit invalidation based on brush changes should be sufficient (as long as all the bases are covered). `Brushes` are the only ones "problematic" enough, as they can change while the instance remains the same, but otherwise the layer state properties can be compared quite reliably

ramezgerges: I have made the requested changed. For the record, continuing to do these unsound ad-hoc optimizations without even profiling will continue to bite us.

MartinZikmund (approved):

Issue #18401, opened 2024-10-07
Frame content does not render when using FeatureConfiguration.Frame.UseWinUIBehavior on android


DavidM29:
### Current behavior

Frame content does not render when setting `FeatureConfiguration.Frame.UseWinUIBehavior = true;`.

### Expected behavior

Content should get rendered.

### How to reproduce it (as minimally and precisely as possible)

- Run the attached app on android with `FeatureConfiguration.Frame.UseWinUIBehavior = true;`, Test Button is not visible.
- Run without `FeatureConfiguration.Frame.UseWinUIBehavior = true;` and the Test Button is visible.

[UnoApp1.zip](https://github.com/user-attachments/files/17277186/UnoApp1.zip)

### Workaround

Have not found any workarounds yet.

### Works on UWP/WinUI

Yes

### Affected platforms

Android

### IDE

Visual Studio 2022

### IDE version

17.11.4

MartinZikmund:
After investigation I have found out the problem is that Uno.Toolkit automatically applies the `NativeFramePresenter` as the `Frame` template. This does not fare well with the WinUI `Frame` behavior, as that requires the `Frame` to manage `Page` instances and caching as compared to the frame presenter itself.

If you set `UseWinUIBehavior` to true on Android and iOS, you also need to override the default style for the control. You can do this by explicitly setting the Style to `XamlDefaultFrame:`

```xml
<Frame Style="{StaticResource XamlDefaultFrame}" />
```

Or by creating an implicit style based on XamlDefaultFrame:

```xml
<Style TargetType="Frame" BasedOn="XamlDefaultFrame" />
```

PR #18460, opened 2024-10-14
Don't allow mixing WinUI Frame behavior with native frame presenter

MartinZikmund:
GitHub Issue (If applicable): closes https://github.com/unoplatform/uno/issues/18401

## PR Type

- Feature
- Documentation

When WinUI Frame behavior is used, content does not load correctly.

## What is the new behavior?

Error log informs the user the mode is not compatible with `NativeFramePresenter` - using `GetType().Name` to handle both the internal `NativeFramePresenter` as well as the one coming from `Uno.Toolkit`.

## PR Checklist

## Other information


src/Uno.UI/UI/Xaml/Controls/Frame/Frame.partial.mux.cs
```diff
@@ -48,6 +49,17 @@ protected override void OnApplyTemplate()
 
 		base.OnApplyTemplate();
 
+#if HAS_UNO
+		if (this.TemplatedRoot?.GetType().Name?.Contains("NativeFramePresenter", StringComparison.OrdinalIgnoreCase) == true)
```
Youssef1313, line 53: Ordinal should just work, I think.

```suggestion
		if (this.TemplatedRoot?.GetType().Name?.Contains("NativeFramePresenter", StringComparison.Ordinal) == true)
```

src/Uno.UI/UI/Xaml/Controls/Frame/Frame.partial.mux.cs
```diff
@@ -48,6 +49,17 @@ protected override void OnApplyTemplate()
 
 		base.OnApplyTemplate();
 
+#if HAS_UNO
+		if (this.TemplatedRoot?.GetType().Name?.Contains("NativeFramePresenter", StringComparison.Ordinal) == true)
+		{
+			// It is not possible to use the WinUI behavior with a NativeFramePresenter.
+			if (this.Log().IsEnabled(LogLevel.Error))
+			{
+				this.Log().LogError("WinUI Frame behavior is not compatible with NativeFramePresenter. Set the Frame.Style to '{StaticResource XamlDefaultFrame}' instead.");
```
kazo0, line 58: Should we perhaps "undo" the feature flags for NatveFramePresenter if `UseWinUIBehavior` for `Frame` is set?

https://github.com/unoplatform/uno/blob/4babcb40873138e6108b387d4732506c02c26824/src/Uno.UI/FeatureConfiguration.cs#L470C23-L470C53

MartinZikmund: The trouble here is that in case of the Toolkit `NativeFramePresenter` I don't think they are set - the style coming from toolkit is set always irrespective of the flags, unless I missed something. If that is the case, then I think this solution will apply to both cases.

Secondly, you could mix WinUI and non-WinUI behavior frames in a single apps, in which case this hint may be better.

doc/articles/controls/Frame.md
```diff
@@ -12,3 +12,15 @@ uid: Uno.Controls.Frame
 ## Using Frame with Uno
 
 To improve performance during navigation, `Frame` on Android, iOS, and WebAssembly targets operates in different way than in WinUI. Whereas WinUI follows `NavigationCacheMode` property on individual `Page` instances, on iOS and Android we keep the individual page instances in the back stack in memory by default. This way the can be quickly surfaced back to the user during back navigation. This behavior can be controlled using the `FeatureConfiguration.Frame.UseWinUIBehavior` property. This defaults to `true` on Skia targets and to `false` on Android, iOS and WebAssembly.
```
ramezgerges, line 14:
```suggestion
To improve performance during navigation, `Frame` on Android, iOS, and WebAssembly targets operates in different way than in WinUI. Whereas WinUI follows `NavigationCacheMode` property on individual `Page` instances, on iOS and Android we keep the individual page instances in the back stack in memory by default. This way they can be quickly surfaced back to the user during back navigation. This behavior can be controlled using the `FeatureConfiguration.Frame.UseWinUIBehavior` property. This defaults to `true` on Skia targets and to `false` on Android, iOS and WebAssembly.
```

src/Uno.UI/UI/Xaml/Controls/Frame/Frame.partial.mux.cs
```diff
@@ -48,6 +49,17 @@ protected override void OnApplyTemplate()
 
 		base.OnApplyTemplate();
 
+#if HAS_UNO
+		if (this.TemplatedRoot?.GetType().Name?.Contains("NativeFramePresenter", StringComparison.Ordinal) == true)
```
ramezgerges, line 53: Not sure if I'm following, but maybe
```suggestion
		if (useWinUIBehavior && this.TemplatedRoot?.GetType().Name?.Contains("NativeFramePresenter", StringComparison.Ordinal) == true)
```
?

ramezgerges (approved):

src/Uno.UI/UI/Xaml/Controls/Frame/Frame.partial.mux.cs
```diff
@@ -48,6 +49,17 @@ protected override void OnApplyTemplate()
 
 		base.OnApplyTemplate();
 
+#if HAS_UNO
+		if (_useWinUIBehavior && this.TemplatedRoot?.GetType().Name?.Contains("NativeFramePresenter", StringComparison.Ordinal) == true)
```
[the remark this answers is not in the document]
MartinZikmund: Good point will try to rebase and verify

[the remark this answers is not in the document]
MartinZikmund: Works as expected!

PR #18468, opened 2024-10-15
fix: X11's render timer interval was set to 1 frame per 16 ticks, not 16 ms

ramezgerges:

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI.Runtime.Skia.X11/X11XamlRootHost.cs
+++ b/src/Uno.UI.Runtime.Skia.X11/X11XamlRootHost.cs
@@ -100,7 +100,7 @@ internal partial class X11XamlRootHost : IXamlRootHost
 		_configureCallback = configureCallback;
 
 		_renderTimer = new DispatcherTimer();
-		_renderTimer.Interval = new TimeSpan(1000 / 16);
+		_renderTimer.Interval = TimeSpan.FromMilliseconds(16.66); // 60 hz
 		_renderTimer.Tick += (sender, o) =>
 		{
 			if (Interlocked.Exchange(ref _needsConfigureCallback, 0) == 1)
@@ -110,8 +110,8 @@ internal partial class X11XamlRootHost : IXamlRootHost
 
 			if (_renderDirty)
 			{
-				_renderer?.Render();
 				_renderDirty = false;
+				_renderer?.Render();
 			}
 		};
 		_renderTimer.Start();


src/Uno.UI.Runtime.Skia.X11/X11XamlRootHost.cs
```diff
@@ -110,8 +110,8 @@ public X11XamlRootHost(X11WindowWrapper wrapper, Window winUIWindow, XamlRoot xa
 
 			if (_renderDirty)
 			{
-				_renderer?.Render();
 				_renderDirty = false;
+				_renderer?.Render();
```
Youssef1313, line 114: Does the order really matter here?

Youssef1313 (approved):

src/Uno.UI.Runtime.Skia.X11/X11XamlRootHost.cs
ramezgerges: It probably won't be noticeable, but there's a theoretical race condition where we during `_renderer?.Render()`, another thread might set `_renderDirty` to true and it that case it will be reset back to false after `Render()` ends, so we will lag a frame in that case.

src/Uno.UI.Runtime.Skia.X11/X11XamlRootHost.cs
```diff
@@ -100,7 +100,7 @@ public X11XamlRootHost(X11WindowWrapper wrapper, Window winUIWindow, XamlRoot xa
 		_configureCallback = configureCallback;
 
 		_renderTimer = new DispatcherTimer();
-		_renderTimer.Interval = new TimeSpan(1000 / 16);
+		_renderTimer.Interval = TimeSpan.FromMilliseconds(16.66); // 60 hz
```
jeromelaban, line 103: We probably should be setting this to be configurable.

ramezgerges: Like a `FeatureConfiguration` option? I would rather not expand `FeatureConfiguration` even more, it's already quite bloated. I would guess that 60 fps is smooth enough. Is there a specific scenario where beefier machines could benefit from the extra frames?

jeromelaban: It's mostly for less performant machines than anything else at this point. I'd expose it though the platform builder, though.

MartinZikmund (approved):

jeromelaban:

PR #18472, opened 2024-10-15
chore: remove negative HR indicator margins

### LINKED ISSUE #18471: HR indicator is using negative margins
### Current behavior

The HR indicator is using -3 margins because the paths of the SVGs its using have a "margin" included in them already. This leads to the HR indicator being offcenter when no extra margin is applied. If we fix the SVG paths, the -3 margins won't be needed anymore.

This also might help with https://github.com/unoplatform/uno/issues/17837 where we need to use -10 margins.

rajamatt:
GitHub Issue (If applicable): closes #18471

## PR Type

### Refactoring (no functional changes, no api changes)

- SVG paths were updated to new paths which don't have an already "included" margin (the size of the new paths were 5 pixels bigger than this one, hence the margins of 5 that were added in some places)
- Moves the size constraints Height and Width from the root grid to the HR Icon itself

Light:
image

Dark:
image


ajpinedam (approved):
Looks good to me.

jeromelaban:

dr1rrb (approved):

PR #18476, opened 2024-10-15
perf: reduce SKPaint and SKPath allocations

### LINKED ISSUE #18473: [Linux] `Path` is not clipping inside a ScrollViewer.
### Current behavior

In the repro below, scrolling the scrollviewer will make the `Path` inside remain visible even though it should be partially clipped. The shocking part is that this only happens on Linux. The WPF target on Windows clips accurately.

### Expected behavior

The `Path` should only be visible within the viewport of the ScrollViewer.

```xaml
	<StackPanel>
		<Border Height="100" HorizontalAlignment="Stretch" Background="Blue" />
		<ScrollViewer Height="30">
			<StackPanel>
				<Path Data="m 0,0 c -0.639,0 -1.276,0.243 -1.765,0.729 -0.977,0.974 -0.98,2.557 -0.006,3.535 L 16.925,23.032 -1.768,41.725 c -0.976,0.976 -0.976,2.559 0,3.535 0.977,0.976 2.559,0.976 3.536,0 L 22.225,24.803 c 0.975,-0.975 0.976,-2.555 0.004,-3.532 L 1.771,0.736 C 1.283,0.245 0.642,0 0,0" Fill="Red" />
			</StackPanel>
		</ScrollViewer>
	</StackPanel>
```

### Workaround

_No response_

### Works on UWP/WinUI

None

### Environment

_No response_

### NuGet package version(s)

_No response_

### Affected platforms

_No response_

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

ramezgerges:
GitHub Issue (If applicable):

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI.Composition/Composition/CompositionSpriteShape.skia.cs
+++ b/src/Uno.UI.Composition/Composition/CompositionSpriteShape.skia.cs
@@ -76,7 +76,7 @@ namespace Microsoft.UI.Composition
 					// On Windows, the stroke is simply 1px, it doesn't scale with the height.
 					// So, to get a correct stroke geometry, we must apply the transformations first.
 
-					var strokeFillPath = SkiaExtensions.GetTempSKPath();
+					var strokeFillPath = SkiaHelper.GetTempSKPath();
 					// Get the stroke geometry, after scaling has been applied.
 					strokePaint.GetFillPath(geometryWithTransformations, strokeFillPath);
 
@@ -95,7 +95,7 @@ namespace Microsoft.UI.Composition
 
 		private static SKPaint GetTempPaint(bool isStroke, SKColorFilter? colorFilter, bool isHighQuality = false)
 		{
-			var paint = SkiaExtensions.GetTempSKPaint();
+			var paint = isStroke ? SkiaHelper.GetTempSKPaint() : SkiaHelper.GetAnotherTempSKPaint();
 			paint.IsAntialias = true;
 			paint.ColorFilter = colorFilter;
 
@@ -162,7 +162,7 @@ namespace Microsoft.UI.Composition
 					var strokePaint = GetTempStrokePaint(null);
 					strokePaint.StrokeWidth = StrokeThickness;
 
-					var hitTestStrokeFillPath = SkiaExtensions.GetTempSKPath();
+					var hitTestStrokeFillPath = SkiaHelper.GetTempSKPath();
 					strokePaint.GetFillPath(geometryWithTransformations, hitTestStrokeFillPath);
 					if (hitTestStrokeFillPath.Contains((float)point.X, (float)point.Y))
 					{
--- a/src/Uno.UI.Composition/Composition/CompositionSpriteShape.skia.cs
+++ b/src/Uno.UI.Composition/Composition/CompositionSpriteShape.skia.cs
@@ -3,6 +3,7 @@
 using Windows.Foundation;
 using SkiaSharp;
 using Uno;
+using Uno.Disposables;
 using Uno.Extensions;
 
 namespace Microsoft.UI.Composition
@@ -17,7 +18,7 @@ namespace Microsoft.UI.Composition
 			{
 				if (FillBrush is { } fill)
 				{
-					var fillPaint = GetTempFillPaint(session.Filters.OpacityColorFilter);
+					using var fillPaintDisposable = GetTempFillPaint(session.Filters.OpacityColorFilter, out var fillPaint);
 
 					if (Compositor.TryGetEffectiveBackgroundColor(this, out var colorFromTransition))
 					{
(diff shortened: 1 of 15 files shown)


src/Uno.UI.Composition/Composition/CompositionSpriteShape.skia.cs
```diff
@@ -126,9 +136,38 @@ private static SKPaint TryCreateAndClearPaint(ref SKPaint? paint, bool isStroke,
...
+					paint.Typeface = null;
+				}
 			}
 
+			paint.IsStroke = isStroke;
+
+			// uno-specific defaults
+			paint.Color = SKColors.White;   // Transparent color wouldn't draw anything
+			paint.IsAutohinted = true;
+			// paint.IsAntialias = true; // IMPORTANT: don't set this to true by default. It breaks canvas clipping on Linux for some reason.
```
Youssef1313, line 164: This is super super curious. Antialiasing it generally an important thing for us, I think.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Shapes/Given_Path.cs
```diff
@@ -41,5 +45,29 @@ public void Should_Not_Include_Control_Points_Bounds()
 			Assert.IsTrue(Math.Abs(50 - SUT.DesiredSize.Height) <= 1, $"Actual size: {SUT.DesiredSize}");
 #endif
 		}
+
+		// This tests a workaround for a Linux-specific skia bug.
+		// This doesn't actually repro for some reason, so it's here for documentation purposes.
```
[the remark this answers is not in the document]
ramezgerges: The test passes on Linux even without this PR. RenderTargetBitmap renders on a new surface, which seems to prevent the problem from appearing. The problem only appears with the renderer's SKSurface.

src/Uno.UI.Composition/Composition/CompositionSpriteShape.skia.cs
ramezgerges: It is indeed :/. The point of this change is to set antialiasing as needed instead of enabling it by default, which still sounds silly but it seems to be the only somewhat reasonable fix. Removing this method entirely and creating a new SKPath when needed would also fix this. I might try to remove it and profile the difference.

jeromelaban: Does it still happen with SkiaSharp 3 ?

ramezgerges: I didn't think to try that at all. It works fine with SkiaSharp 3!

ramezgerges: The `TryCreateAndClear***` methods that looked hacky to me above. Unfortunately, profiling should a significant impact when I removed them.

While investigating the problem, I found that we can actually get away with not allocating almost any `SKPaint`s across the entire app. Almost all our `SKPaint` usages are very temporary, usually just to immediately draw something. So, we can "borrow" `SKPaint` instances when needed and it turns out that there's an `SKPaint.Reset` method that cleans the `SKPaint` instances between usages, so this also takes care of these `TryCreateAndClear***` methods that can now soundly depend on `Reset` instead of the error-prone manual cleanup that we did before.

At this point, this is mostly a perf PR, that just so happens to also fix the problem with clipping on Linux.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Shapes/Given_Path.cs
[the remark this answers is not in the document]
MartinZikmund: In this case I would suggest creating a sample page which illustrates the problem and is marked as manual test for QA to verify against regressions

src/Uno.UI.Composition/Composition/SkiaHelper.skia.cs
```diff
@@ -0,0 +1,43 @@
...
+
+		public static SKPaint GetTempSKPaint()
+		{
+			// https://api.skia.org/classSkPaint.html#a6c7118c97a0e8819d75aa757afbc4c49
+			// "This is equivalent to replacing SkPaint with the result of SkPaint()."
+			_tempPaint.Reset();
+			return _tempPaint;
+		}
+
+		public static SKPaint GetAnotherTempSKPaint()
```
MartinZikmund, line 35: Instead of this I would probably prefer `GetTempSKPaint(int id)` where you would have a `List<SKPaint>` that would create a new one when `id >= Count`. For now two are enough, but some future API could need 3, this would make it easier.

ramezgerges: It was bugging me too. I rewrote SkiaHelper to use ObjectPool, which does what you suggest without the resizing aspect (which shouldn't be needed anyway).

src/Uno.UI.Composition/Composition/CompositionSpriteShape.skia.cs
ramezgerges: Update: even after all the changes, the problem still reproduces on Linux, unless we disable anti-aliasing, which is a non-starter as Youssef said, or unless we close the SKPath everytime we use it for painting (so basically the same as remove the `TryCreate***` methods, which turned out to be very impactful).

For now, the only real fix is to move to SkiaSharp 3

ramezgerges:
@MartinZikmund 
The improvement is most noticeable in a continuously-rendering app (e.g. with animation) and some shapes.

Profiling the `HitTest_Shapes`  samples for around a minute while holding down tab (to shift the focus visual and invalidating the render) gives this:

SKPath: 4.57 MB / 185 MB before vs 179.1 KB / 170 MB after 
SKPaint: 25 KB / 185 MB before vs 432 B (not KB) / 170 MB after

Seems like optimizing SKPaint didn't matter a whole lot, but it was very high in the profiler graph in a different profiling session earlier today. I don't totally remember the scenario I was testing with.

MartinZikmund (approved):

PR #18505, opened 2024-10-17
chore: remove SkiaSharp.HarfBuzz and SkiaSharp.Views dependencies when unnecessary

ramezgerges:
@jeromelaban I noticed SkiaSharp.Views being a dependenct in the wpf runtime project. It doesn't seem to be necessary at all. Can we remove it?


jeromelaban:
> @jeromelaban I noticed SkiaSharp.Views being a dependenct in the wpf runtime project. It doesn't seem to be necessary at all. Can we remove it?

Yes, it was probably originally there because of a early implementation. Good catch!

jeromelaban (approved):

ramezgerges:
@jeromelaban Not sure if this should have been in a separate PR, but all Skia.Runtime.<some_target> have a useless SkiaSharp.HarfBuzz dependency. I removed those too.

Issue #13543, opened 2023-09-06
[iOS] `PersonPicture` does not load the picture when the source is delayed


Marc-Antoine-Soucy:
### Current behavior

![image](https://github.com/unoplatform/uno/assets/90481654/f4e9086d-32db-4af6-addf-a59d9a636ade)
The image becomes gray after the source is set. 

### Expected behavior

Just like on windows, the image should load properly
![image](https://github.com/unoplatform/uno/assets/90481654/4ac0d6b1-8e47-4dfc-8eba-531c942a9f06)

### How to reproduce it (as minimally and precisely as possible)

In the xaml:

```
<PersonPicture ProfilePicture="{Binding ImagePath}"
			 Height="100"
			 Width="100"
			 HorizontalAlignment="Center"
			 VerticalAlignment="Center" />
```

In the codeBehind:
```
public MainPage()
{
	this.InitializeComponent();
	_ = SetImagePathWithDelay();
}

public string ImagePath { get; } = "ms-appx:///BasicApp/Assets/james_bondi.png";

private async Task SetImagePathWithDelay()
{
	await Task.Delay(1000);
	DataContext = this;
}
```

[BasicApp.zip](https://github.com/unoplatform/uno/files/12542519/BasicApp.zip)

### Workaround

You can use an Images with a border or grid around it to have a similar final result as the PersonPicture. 

### Works on UWP/WinUI

None

### NuGet package version(s)

```
<PackageReference Include="Uno.Resizetizer" Version="1.2.0-dev.32" />
<PackageReference Include="Uno.WinUI" Version="5.0.0-dev.2437" />
<PackageReference Include="CommunityToolkit.Mvvm" Version="8.1.0" />
<PackageReference Include="Microsoft.Extensions.Logging.Console" Version="7.0.0" />
<PackageReference Include="Uno.Extensions.Logging.OSLog" Version="1.6.0" />
<PackageReference Include="Uno.UI.Adapter.Microsoft.Extensions.Logging" Version="5.0.0-dev.2437" />
<PackageReference Include="Uno.WinUI.RemoteControl" Version="5.0.0-dev.2437" Condition="'$(Configuration)'=='Debug'" />
```

### Affected platforms

iOS

### IDE

Visual Studio 2022

### IDE version

17.6.5

### Anything else we need to know?

It looks the same as if you put the source as null.

agneszitte:
[...]

PR #18587, opened 2024-10-25
fix(ios): PersonPicture late binding

ajpinedam:
GitHub Issue (If applicable): closes https://github.com/unoplatform/uno/issues/13543

## PR Type

- Bugfix

on iOS, when doing a late binding (setting the context to the page after it has been loaded), the PersonPicture image stays gray.

## PR Checklist

## Other information

### DIFF
--- a/src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/PersonPictureTests/PersonPictureLateBindingPage.xaml.cs
+++ b/src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/PersonPictureTests/PersonPictureLateBindingPage.xaml.cs
@@ -13,6 +13,9 @@ namespace MUXControlsTestApp
 	[SampleControlInfo(
 		"MUX",
 		"PersonPictureLateBinding",
+		description: "Shows a PersonPicture control and a Textblock with a name. \n" +
+		"The PersonPicture control image is set after 1 second delay. \n" +
+		"The image should be displayed after the delay.",
 		isManualTest: true)]
 #pragma warning disable UXAML0002 // does not explicitly define the Microsoft.UI.Xaml.Controls.UserControl base type in code behind.
 	public sealed partial class PersonPictureLateBindingPage
--- a/src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/PersonPictureTests/PersonPictureLateBindingPage.xaml.cs
+++ b/src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/PersonPictureTests/PersonPictureLateBindingPage.xaml.cs
@@ -1,0 +1,37 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Collections.ObjectModel;
+using MUXControlsTestApp.Utilities;
+using Uno.UI.Samples.Controls;
+using Windows.ApplicationModel.Contacts;
+using System.Threading.Tasks;
+
+namespace MUXControlsTestApp
+{
+	[SampleControlInfo(
+		"MUX",
+		"PersonPictureLateBinding",
+		isManualTest: true)]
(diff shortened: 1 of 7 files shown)


src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/PersonPictureTests/PersonPictureLateBindingPage.xaml.cs
```diff
@@ -0,0 +1,40 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
```
Youssef1313, line 2: Is this ported from WinUI? I can't find it in the sources (https://github.com/search?q=repo%3Amicrosoft%2Fmicrosoft-ui-xaml%20PersonPictureLateBindingPage&type=code)

src/Uno.UI/Microsoft/UI/Xaml/Controls/PersonPicture/PersonPicture.cs
```diff
@@ -203,23 +203,33 @@ void UpdateIfReady()
 
 		var templateSettings = TemplateSettings;
 		templateSettings.ActualInitials = initials;
-		if (imageSrc != null)
+
+		if (imageSrc is not null)
```
Youssef1313, line 207: Are the changes in this method refactoring only? It may be better to keep this closer to the original code from WinUI to ease future revisions of the port.

src/Uno.UI/Microsoft/UI/Xaml/Controls/PersonPicture/PersonPicture.cs
```diff
@@ -203,23 +203,33 @@ void UpdateIfReady()
...
 		{
 			templateSettings.ActualImageBrush = null;
 		}
 
+#if __IOS__
+		if (templateSettings.ActualImageBrush is ImageBrush brush)
+		{
+			brush.ImageOpened += RefreshPhoto;
+		}
+#endif
```
Youssef1313, line 232: As WinUI doesn't have this code, it likely means that there is some other behavior that's different between WinUI and Uno. Do we know?

src/SamplesApp/UITests.Shared/Windows_UI_Xaml_Controls/PersonPictureTests/PersonPictureLateBindingPage.xaml.cs
ajpinedam: Oh this is the test! Good catch, No, it is not ported from WinUI. I copied another PersonPicture test and it seems I forgot to delete this part. Deleting it now.

src/Uno.UI/Microsoft/UI/Xaml/Controls/PersonPicture/PersonPicture.cs
ajpinedam: Good point. Will revert back

src/Uno.UI/Microsoft/UI/Xaml/Controls/PersonPicture/PersonPicture.cs
ajpinedam: I could not find anything. The different behaviour, if any, could be either in the `Image` or the `ImageSource`. The problem here is that for iOS the Image loading is taking longer and the code above is not blocking the UI to wait for image to complete loading before changing the VisualState which is then causing an unrecoverable error.

Tried enforcing the UI thread in different areas with no success.

cc: @Youssef1313

jeromelaban: Does it work in the desktop target? If so, what's making it work in that context?

jeromelaban: Thanks for the update. Let's go with the workaround for now.

jeromelaban (approved):

PR #18659, opened 2024-11-01
Make skia runtime tests green locally on Debug

### LINKED ISSUE #18715: Keyboard state reset for debugging scenario
Yes, this one was failing (the name of the failing test is in the commit message). Like I said above, I would definitely be in favour of these debug-friendly options being explicitly activated (like FeatureConfiguration, but for internal use). The only part I don't like is that they're on by default.

_Originally posted by @ramezgerges in https://github.com/unoplatform/uno/pull/18659#discussion_r1829545494_

ramezgerges:
GitHub Issue (If applicable): closes #18715

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/SamplesApp/SamplesApp.Shared/App.xaml.cs
+++ b/src/SamplesApp/SamplesApp.Shared/App.xaml.cs
@@ -159,11 +159,8 @@ namespace SamplesApp
 			}
 
 			var sw = Stopwatch.StartNew();
-#if DEBUG
-			if (System.Diagnostics.Debugger.IsAttached)
-			{
-				// this.DebugSettings.EnableFrameRateCounter = true;
-			}
+#if WINAPPSDK && DEBUG
+			// this.DebugSettings.EnableFrameRateCounter = true;
 #endif
 			AssertInitialWindowSize();
 
--- a/src/SamplesApp/SamplesApp.Shared/App.xaml.cs
+++ b/src/SamplesApp/SamplesApp.Shared/App.xaml.cs
@@ -546,6 +546,7 @@ namespace SamplesApp
 #if HAS_UNO
 			Uno.UI.FeatureConfiguration.TextBox.UseOverlayOnSkia = false;
 			Uno.UI.FeatureConfiguration.ToolTip.UseToolTips = true;
+			Uno.UI.FeatureConfiguration.DependencyProperty.ValidatePropertyOwnerOnReadWrite = true;
 
 			Uno.UI.FeatureConfiguration.Font.DefaultTextFontFamily = "ms-appx:///Uno.Fonts.OpenSans/Fonts/OpenSans.ttf";
 #endif
--- a/src/Uno.UI.RuntimeTests/MUX/Utilities/TestHelpers.cs
+++ b/src/Uno.UI.RuntimeTests/MUX/Utilities/TestHelpers.cs
@@ -139,7 +139,7 @@ namespace MUXControlsTestApp.Utilities
 {
 	public static class TestUtilities
 	{
-		public static int DefaultWaitMs = Debugger.IsAttached ? 120000 : 5000;
+		public static int DefaultWaitMs = 5000;
 
 		public static async Task SetAsVisualTreeRoot(FrameworkElement element)
 		{
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_Image.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_Image.cs
@@ -663,8 +663,8 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 			await WindowHelper.WaitFor(() => imageOpened);
 			var screenshot = await TakeScreenshot(image);
 			ImageAssert.HasColorAt(screenshot, 5, screenshot.Height / 2, Color.FromArgb(0xFF, 0xED, 0x1B, 0x24), tolerance: 5);
-			ImageAssert.HasColorAt(screenshot, screenshot.Width / 2 - 10, screenshot.Height / 2, Color.FromArgb(0xFF, 0xED, 0x1B, 0x24), tolerance: 5);
-			ImageAssert.HasColorAt(screenshot, screenshot.Width / 2 + 10, screenshot.Height / 2, Color.FromArgb(0xFF, 0x23, 0xB1, 0x4D), tolerance: 5);
+			ImageAssert.HasColorAt(screenshot, screenshot.Width / 2.2f, screenshot.Height / 2, Color.FromArgb(0xFF, 0xED, 0x1B, 0x24), tolerance: 5);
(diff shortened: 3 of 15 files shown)


src/SamplesApp/SamplesApp.Shared/App.xaml.cs
```diff
@@ -159,11 +159,8 @@ override void OnLaunched(LaunchActivatedEventArgs e)
 			}
 
 			var sw = Stopwatch.StartNew();
-#if DEBUG
-			if (System.Diagnostics.Debugger.IsAttached)
```
jeromelaban, line 163: Is this failing in debug somehow? I don't mind the change, though.

src/Uno.UI.RuntimeTests/MUX/Utilities/TestHelpers.cs
```diff
@@ -139,7 +139,7 @@ namespace MUXControlsTestApp.Utilities
 {
 	public static class TestUtilities
 	{
-		public static int DefaultWaitMs = Debugger.IsAttached ? 120000 : 5000;
```
jeromelaban, line 142: The point of this particular line is to make it possible to break inside a test for a little while and still make it pass. Is it failing otherwise?

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_Image.cs
```diff
@@ -663,8 +663,8 @@ public async Task When_Exif_Rotated_MsAppx_Unequal_Dimensions()
 			await WindowHelper.WaitFor(() => imageOpened);
 			var screenshot = await TakeScreenshot(image);
 			ImageAssert.HasColorAt(screenshot, 5, screenshot.Height / 2, Color.FromArgb(0xFF, 0xED, 0x1B, 0x24), tolerance: 5);
-			ImageAssert.HasColorAt(screenshot, screenshot.Width / 2 - 10, screenshot.Height / 2, Color.FromArgb(0xFF, 0xED, 0x1B, 0x24), tolerance: 5);
-			ImageAssert.HasColorAt(screenshot, screenshot.Width / 2 + 10, screenshot.Height / 2, Color.FromArgb(0xFF, 0x23, 0xB1, 0x4D), tolerance: 5);
+			ImageAssert.HasColorAt(screenshot, screenshot.Width / 2.2f, screenshot.Height / 2, Color.FromArgb(0xFF, 0xED, 0x1B, 0x24), tolerance: 5);
```
jeromelaban, line 666: Why is it different in debug?

src/Uno.UI/UI/Xaml/DependencyObjectStore.cs
```diff
@@ -792,25 +790,24 @@ private void WritePropertyEventTrace(int eventId, DependencyProperty property, D
 		[MethodImpl(MethodImplOptions.AggressiveInlining)]
 		private void ValidatePropertyOwner(DependencyProperty property)
 		{
-			if (_validatePropertyOwner)
-			{
-				var isFrameworkElement = _originalObjectType.Is(typeof(FrameworkElement));
-				var isMixinFrameworkElement = _originalObjectRef.Target is IFrameworkElement && !isFrameworkElement;
+#if DEBUG
```
jeromelaban, line 793: The original idea here is that if the debugger is attached, even in release builds that come from nuget, this check is still done for app developers. Changing it for a compile time conditional will remove that check.

src/Uno.UWP/UI/Core/Internal/KeyboardStateTracker.cs
```diff
@@ -97,13 +97,5 @@ private static void SetStateOnNonSideKeys(VirtualKey key)
...
-	internal static void Reset()
-	{
-		// Clearing state when debugger is attached would
-		// make it hard to debug key events.
-		if (!System.Diagnostics.Debugger.IsAttached)
-		{
-			_keyStates.Clear();
-		}
-	}
+	internal static void Reset() => _keyStates.Clear();
```
jeromelaban, line 100: Is it failing otherwise? If so, @MartinZikmund may have some input for debugging.

src/SamplesApp/SamplesApp.Shared/App.xaml.cs
ramezgerges: Not this instance, but something was failing because of a `Debugger.IsAttached` condition somewhere, so I removed all these checks everywhere. I've been bitten by those before too and it always wastes a lot of time because you'd never guess that attaching the debugger would change the behaviour.

src/Uno.UI.RuntimeTests/MUX/Utilities/TestHelpers.cs
ramezgerges: I understand, but I'm still very against any `Debugger.IsAttached` checks. We can put some commented-out options that can be turned on on demand, but leaving these checks by default has burnt me way too many times (not this specific instance, but it's the same point).

src/Uno.UWP/UI/Core/Internal/KeyboardStateTracker.cs
ramezgerges: Yes, this one was failing (the name of the failing test is in the commit message). Like I said above, I would definitely be in favour of these debug-friendly options being explicitly activated (like FeatureConfiguration, but for internal use). The only part I don't like is that they're on by default.

While we're at it, whenever I'm debugging something popup-related,  I tend to manually delete the code that dismisses all popups when the Uno window is alt-tabbed. We could also add this to the list of things that are adjusted when in the "debug-friendly mode".

src/Uno.UI/UI/Xaml/DependencyObjectStore.cs
ramezgerges: Ah, I see. Good point. Can we make this a FeatureConfigutarion option instead? We can explicitly keep the checks internally for SamplesApp (always) and make the public builds run the checks if Debugger.IsAttached. Actually, I would prefer to make the checks run if `Debugger.IsAttached` OR `DEBUG`, as these checks are really useful when developping even when a debugger isn't attached.

jeromelaban: Sure, a feature configuration would be fine indeed.

jeromelaban (approved):

src/Uno.UI/UI/Xaml/Controls/AppBar/AppBar.Partial.cs
```diff
@@ -769,7 +769,8 @@ private void OnIsOpenChanged(bool isOpen)
 		{
 			// If the AppBar is not live, then wait until it's loaded before
 			// responding to changes to opened state and firing our Opening/Opened events.
-			if (!IsInLiveTree)
+			// Uno Specific: using IsLoaded instead of IsInLiveTree, which makes more sense because OnOpening (called below) -> SetupOverlayState expects OnApplyTemplate to have already been called
+			if (!IsLoaded)
 			{
```
Youssef1313, line 774: This was not a correct change.

PR #18679, opened 2024-11-04
Update get-started-counter-csharp-mvux.md

JohannesMP:
- The [MainModel](https://platform.uno/docs/articles/getting-started/counterapp/get-started-counter-csharp-mvux.html?tabs=vs#mainmodel) walks you through the creation of `MainModel.cs` which defines the partial record `MainModel`.
- However the [DataBinding](https://platform.uno/docs/articles/getting-started/counterapp/get-started-counter-csharp-mvux.html?tabs=vs#data-binding) only refers to `MainViewModel` which is never defined.

This PR renames `MainViewModel` references in the **DataBinding** section to be `MainModel`, as the examples would not compile otherwise.

----

Alternatively if the intent was to use `MainViewModel` then the `MainModel` section should be updated instead.

### DIFF
--- a/doc/articles/getting-started/counterapp/get-started-counter-csharp-mvux.md
+++ b/doc/articles/getting-started/counterapp/get-started-counter-csharp-mvux.md
@@ -89,12 +89,12 @@ Also, for more information on all the template options, see [Using the Uno Platf
 
 ## Data Binding
 
-Now that we have the **`MainViewModel`** class, we can update the **`MainPage`** to use data binding to connect the UI to the application logic.
+Now that we have the **`MainModel`** class, we can update the **`MainPage`** to use data binding to connect the UI to the application logic.
 
-- Let's add the **`DataContext`** to our page. To do so, add `.DataContext(new MainViewModel(), (page, vm) => page` before `.Background(...)`. Remember to close the **`DataContext`** expression with a `)` at the end of the code. It should look similar to the code below:
+- Let's add the **`DataContext`** to our page. To do so, add `.DataContext(new MainModel(), (page, vm) => page` before `.Background(...)`. Remember to close the **`DataContext`** expression with a `)` at the end of the code. It should look similar to the code below:
 
     ```csharp
-    this.DataContext(new MainViewModel(), (page, vm) => page
+    this.DataContext(new MainModel(), (page, vm) => page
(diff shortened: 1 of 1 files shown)


Kunal22shah:
@JohannesMP these renames aren't necessary. The the `MainViewModel` that is used for dataContext is auto generated using the new MVUX v3 generator see here : [Upgrading Extensions Reactive Version](https://platform.uno/docs/articles/external/uno.extensions/doc/Learn/Mvux/reactive-migration.html#mvux-generated-code)
and : https://platform.uno/docs/articles/external/workshops/simple-calc/modules/MVUX-CSharp/04-App%20Architecture/README.html#binding-to-properties-in-the-ui 

Which is different from the `MainModel`, this is the default now, so all new Uno Apps should have this
`[assembly: Uno.Extensions.Reactive.Config.BindableGenerationTool(3)]` in GlobalUsings
and the app would compile.


if it's not compiling on your side then maybe you aren't on the later sdk versions where this was added. Let me know if I can help you with more info on the same : )

JohannesMP:
> @JohannesMP these renames aren't necessary. The the `MainViewModel` that is used for dataContext is auto generated using the new MVUX v3 generator see here : [Upgrading Extensions Reactive Version](https://platform.uno/docs/articles/external/uno.extensions/doc/Learn/Mvux/reactive-migration.html#mvux-generated-code) and : https://platform.uno/docs/articles/external/workshops/simple-calc/modules/MVUX-CSharp/04-App%20Architecture/README.html#binding-to-properties-in-the-ui
> 
> Which is different from the `MainModel`, this is the default now, so all new Uno Apps should have this `[assembly: Uno.Extensions.Reactive.Config.BindableGenerationTool(3)]` in GlobalUsings and the app would compile.
> 
> if it's not compiling on your side then maybe you aren't on the later sdk versions where this was added. Let me know if I can help you with more info on the same : )

Thank you for clarifying! 

In that case I would recommend that the sample/tutorial app docs should mention this.

Right now it just says "Now that we have the `MainViewModel` class ..." after creating `MainModel`. 
image

jeromelaban:
Indeed, the name of the class in DataBinding is incorrect, it was fixed here: https://github.com/unoplatform/uno/pull/18708

PR #18758, opened 2024-11-09
feat: Platform runtime helper

### LINKED ISSUE #11545: Add an API to get the current Uno Platform target at runtime

Currently it is possible to conditionally compile code to run on specific platforms, but there is no easy way to check the current Uno target at runtime. This would be especially useful to differentiate between Skia targets.

This could be a simple static method that would return an enum with Windows, iOS, SkiaWpf, SkiaGtk, etc.

## Why is this needed:

Runtime checks

## For which Platform:

All

mcNets:
GitHub Issue (If applicable): closes https://github.com/unoplatform/uno/issues/11545

## TODO

- [x] The helper was moved to Uno.WinRT and hence does not correctly check for Skia vs non-Skia target - needs to be based on the actual current platform value.
- [ ] Validate tests pass across all targets including native
- [x] Validate the changes did not affect runtime tests platform recognition
- [x] Add the enum extensions as linked to the uno toolkit project
- [x] Validate no additional public members were included
- [x] Is it safe to use static ctors (e.g. Skia desktop)?
- [x] Add startup assert to Samples app to validate that the platform is not unknown

## PR Type

- Feature

You have to use conditionals to knot the current platform runtime.

## What is the new behavior?

You can use RuntimeHelper.CurrentPlatform enumeration.

## PR Checklist

## Other information

Follow this conversation:
https://github.com/unoplatform/uno/discussions/18432#discussioncomment-10946686

I've had some issues with the filtered solutions, I've had been able to test it in Skia only.

Please, someone else should check the other platforms.

https://github.com/unoplatform/uno/issues/11545


src/Uno.UI/Helpers/PlatformRuntimeHelper.cs
```diff
@@ -0,0 +1,34 @@
...
+#elif __MACCATALYST__
+		UnoRuntimePlatform.MacCatalyst;
+#elif __MACOS__
+		UnoRuntimePlatform.MacOSX;
+#elif __WASM__
+		UnoRuntimePlatform.WebAssembly;
+#elif __SKIA__
+		SkiaPlatform;
+#elif !HAS_UNO
+		UnoRuntimePlatform.Windows;
```
Youssef1313, line 29: This feels like a dead code?

src/Uno.UI/Helpers/PlatformRuntimeHelper.cs
```diff
@@ -0,0 +1,34 @@
...
+#else
+		UnoRuntimePlatform.Unkonwn;
```
Youssef1313, line 31: #error ?

src/Uno.UI/Helpers/UnoRuntimePlatform.cs
```diff
@@ -0,0 +1,24 @@
...
+	MacCatalyst,
+	MacOSX,
+	WebAssembly,
+	Windows,
+	Skia,
+	SkiaGtk,
+	SkiaWpf,
+	SkiaX11,
+	SkiaFrameBuffer,
+	SkiaMacOS,
```
Youssef1313, line 22: SkiaIslands?

src/Uno.UI/Helpers/PlatformRuntimeHelper.cs
mcNets: What other conditional could I use?

Youssef1313: For Windows, you don't need it here. It should be handled by code in Uno.UI.Toolkit

src/Uno.UI.Toolkit/UnoRuntimePlatform.cs
```diff
@@ -0,0 +1,18 @@
...
+	MacOSX,
+	WebAssembly,
+	Windows,
+	Skia,
+	SkiaGtk,
+	SkiaWpf,
+	SkiaX11,
+	SkiaFrameBuffer,
+	SkiaMacOS,
+	Unkonwn
```
jeromelaban, line 17:
```suggestion
	Unknown
```

src/Uno.UI/Helpers/UnoRuntimePlatform.cs
spouliot, line 22: When\where\how would `Skia*` (e.g. `SkiaMacOS`) be returned ?

There's no code to allow to set it (from the host) since it's returned from a method. An internal `Set...` method (only for `__SKIA__` could be used by the different hosts).

Also if `Skia*` values are returned when would the lone `Skia` value be returned ?

You might want to add an extension method, on the enum, like `IsSkia()` to make it easier to detect all Skia-based platforms (and remove `Skia`).

MartinZikmund: In such case I would go with `SkiaWpfIslands` to make it even more specific

src/Uno.UI/Helpers/UnoRuntimePlatform.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Threading.Tasks;
+
+namespace Uno.UI.Helpers;
+
+public enum UnoRuntimePlatform
+{
+	Android,
+	iOS,
+	MacCatalyst,
+	MacOSX,
```
MartinZikmund, line 14: I think this entry is not necessary, as it is actually only temporary anyway, as we will be dropping the legacy macOS implementation

src/Uno.UI/Helpers/UnoRuntimePlatform.cs
MartinZikmund: This sounds good, in some cases you want to make some code skia specific (maybe because of native rendering limitations), in other cases it is the "native platform" that matters only; so a combination of `IsSkia()` and `GetPlatform()` makes sense; and dropping `Skia` enum value completely (and use unknown as the "default")

src/Uno.UI/Helpers/PlatformRuntimeHelper.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace Uno.UI.Helpers;
+
+public static class PlatformRuntimeHelper
+{
+	internal static UnoRuntimePlatform SkiaPlatform { get; set; } = UnoRuntimePlatform.Skia;
+
+	public static UnoRuntimePlatform Current => GetPlatform();
+
+	private static UnoRuntimePlatform GetPlatform() =>
```
MartinZikmund, line 15: Please add a runtime test that gets the value from `GetPlatform()` at runtime and checks that the result is not `Unknown` - this will future proof us against potential new platforms being added and us forgetting to update this code path

src/Uno.UI/Helpers/UnoRuntimePlatform.cs
```diff
@@ -0,0 +1,24 @@
...
+	WebAssembly,
+	Windows,
+	Skia,
+	SkiaGtk,
+	SkiaWpf,
+	SkiaX11,
+	SkiaFrameBuffer,
+	SkiaMacOS,
+	Unknown
```
MartinZikmund, line 23: I think `Unknown` should be the default value - (`= 0`)

MartinZikmund:
@mcNets Nice PR! I know we are thorough with the review, but as this can be a very useful and frequently used API and it is public, we want to make sure to get it right 🙏

mcNets:
@MartinZikmund I've set default to Unknown, I removed IsAndroid, IsIOS, etc, and changed SkiaWpf by SkiaWpfIslands. I'll merged main branch again but still not green,.

MartinZikmund:
@mcNets the test belongs in Uno.UI.RuntimeTests as it needs to run on each platform at runtime. There is this folder - https://github.com/unoplatform/uno/tree/master/src/Uno.UI.RuntimeTests/Tests/Uno_UI_Toolkit where you can add your new test class. Theoretically you can also add platform-specific tests in that test class as well - e.g.:

- `GetPlatform_Not_Unknown()`
- `#if __ANDROID__ -> GetPlatform_IsAndroid()`
- `#if __iOS__ -> GetPlatform_IsiOS()`
- etc.

mcNets:
@MartinZikmund Do you know why CodeQL test is always failling?

jeromelaban:
> @MartinZikmund Do you know why CodeQL test is always failling?

This was fixed a few merges ago, the next build will succeed.

mcNets:
@MartinZikmund I've added the tests now, I've checked with Skia and it pass.

@Youssef1313 , @spouliot Would you mind taking a look at the unresolved conversations and let me know if can be closed?

mcNets:
@jeromelaban there are still some test that they fail, and it seems that at least one of them is required. Any advice on what should I look for?

MartinZikmund:
@mcNets the fails occurs specifically on WPF Islands:

image

mcNets:
@MartinZikmund these are the ones I've added to ckeck the Runtime Platform helpers.

image

image

It seems it is not using/finding the new namspace

```
src\Uno.UI.Toolkit\RuntimeHelper.cs(6,14): Error CS0234: The type or namespace name 'Helpers' does not exist in the namespace 'Uno.UI' (are you missing an assembly reference?)
```

mcNets:
Just to keep it alive.

mcNets:
I'll work on it

MartinZikmund:
Some of the checks are failing, getting `Unknown` platform, can you check the runtime tests?

mcNets:
> Some of the checks are failing, getting `Unknown` platform, can you check the runtime tests?

Sorry @MartinZikmund, I just rebased my branch and pushed it again. I still need to work on it.

src/Uno.UI.Toolkit/RuntimeHelper.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Uno.UI.Helpers;
+
+namespace Uno.UI.Toolkit;
+
+public static class PlatformRuntimeHelper
```
jeromelaban: It makes sense to have this helper to include it for winappsdk, but in that case, the usefulness of the `Uno.UI.Helpers` helper is very limited to uno users and could cause namespace and `#if` fiddling.

I see two ways here, either we :
1. Don't include that API for winappsdk, and let app developers `#if WINDOWS` or similar around it
2. Hide the `Uno.UI.Helpers` version from the public API, so it can still be used internally, and only keep the `Uno.UI.Toolkit` one.

The first one has the advantage of avoiding to have two versions seemingly of the same classes, but forces `#if` uses.

The second one has the advantage of being available everywhere, but is located in an unrelated namespace.

I think I prefer the second, just for the sake of developer convenience.

mcNets:
Hi @MartinZikmund 

I'm not able to see what is failing. They didn't seem related to this PR.

Would you mind to take a look?

MartinZikmund:
@mcNets will take a look asap!

src/Uno.UI.Runtime.Skia.Win32/Hosting/Win32Host.cs
```diff
@@ -107,6 +110,8 @@ public Win32Host(Func<Application> appBuilder, bool preloadVlc)
 			mediaExtensionType.GetMethod("PreloadVlc", BindingFlags.Static | BindingFlags.Public)?.Invoke(null, null);
 		}
 
+		PlatformRuntimeHelper.SkiaPlatform = UnoRuntimePlatform.SkiaWin32;
```
jeromelaban: This is likely not needed because it's already set in the static ctor?

src/Uno.UI.Runtime.Skia.Wpf/Hosting/WpfHost.cs
```diff
@@ -42,6 +45,8 @@ internal WpfHost(Func<WinUIApplication> appBuilder, Func<WpfApplication>? wpfApp
 		_current = this;
 		_dispatcher = _wpfApp.Dispatcher;
 		_appBuilder = appBuilder;
+
+		PlatformRuntimeHelper.SkiaPlatform = UnoRuntimePlatform.SkiaWpf;
```
jeromelaban: Same here, set in the static ctor.

src/Uno.UI.Runtime.Skia.X11/Hosting/X11ApplicationHost.cs
```diff
@@ -109,6 +112,8 @@ public X11ApplicationHost(Func<Application> appBuilder, int renderFrameRate = 60
 
 		_appBuilder = appBuilder;
 
+		PlatformRuntimeHelper.SkiaPlatform = UnoRuntimePlatform.SkiaX11;
```
jeromelaban: Same here, static ctor

src/Uno.UI/Helpers/Xaml/UnoRuntimePlatformExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+#if __SKIA__
+		true;
+#else
+	false;
+#endif
+
+
+	public static bool IsIOS(this UnoRuntimePlatform platform) => platform == UnoRuntimePlatform.NativeIOS || platform == UnoRuntimePlatform.SkiaIOS;
+
+	public static bool IsWindows(this UnoRuntimePlatform platform) => platform == UnoRuntimePlatform.NativeWinUI;
```
jeromelaban: IsWindows could be confusing because windows can also mean uno desktop.

```suggestion
	public static bool IsWindowsAppSdk(this UnoRuntimePlatform platform) => platform == UnoRuntimePlatform.NativeWinUI;
```

src/Uno.UI/Helpers/UnoRuntimePlatform.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace Uno.UI.Helpers;
+
+public enum UnoRuntimePlatform
```
MartinZikmund: Please make the helpers in Uno.UI/Helpers `internal` only, so that from the public API perspective only the `Uno.UI.Toolkit` version is visible

src/Uno.UI/Helpers/PlatformRuntimeHelper.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace Uno.UI.Helpers;
+
+public static class PlatformRuntimeHelper
```
MartinZikmund: Please make the helpers in Uno.UI/Helpers `internal` only, so that from the public API perspective only the `Uno.UI.Toolkit` version is visible

src/Uno.UI/Helpers/Xaml/UnoRuntimePlatformExtensions.cs
```diff
@@ -0,0 +1,23 @@
+namespace Uno.UI.Helpers;
+
+public static class UnoRuntimePlatformExtensions
```
MartinZikmund: Please make the helpers in Uno.UI/Helpers `internal` only, so that from the public API perspective only the `Uno.UI.Toolkit` version is visible

MartinZikmund:
@mcNets I'm making some additional changes to reuse the helper across Uno, Runtime tests and Toolkit without having duplicate types, but getting close!

mcNets:
@MartinZikmund Do I have to make any changes to the code?

MartinZikmund:
@mcNets its mostly infrastructure changes, will take care of it

mcNets:
Hi @MartinZikmund, I have a lot of conflicts with my fork and my local repo.

src/Uno.UI/Helpers/UnoRuntimePlatform.cs
MartinZikmund: The enum keeps the generic `SkiaIslands` value. The WPF-specific Islands host (`Uno.UI.XamlHost.Skia.Wpf`) was removed in 6.7, so a WPF-specific value is no longer needed.

src/Uno.UI/Helpers/UnoRuntimePlatform.cs
MartinZikmund: Addressed — the legacy native macOS entry was dropped. The enum now carries only `NativeMacCatalyst` and `SkiaMacOS` (the current AppKit-based Skia macOS host).

src/Uno.UWP/Helpers/RuntimePlatformHelper.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace Uno.UI.Helpers;
+
+public static class PlatformRuntimeHelper
+{
+	internal static UnoRuntimePlatform SkiaPlatform { get; set; } = UnoRuntimePlatform.Skia;
+
+	public static UnoRuntimePlatform Current => GetPlatform();
+
+	private static UnoRuntimePlatform GetPlatform() =>
```
MartinZikmund: Done — added `When_Platform_Is_Known` asserting `RuntimePlatformHelper.Current` is not `Unknown`, plus a SamplesApp startup assert (`AssertRuntimePlatformIsKnown`) to future-proof against unwired platforms.

src/Uno.UI/Helpers/UnoRuntimePlatform.cs
MartinZikmund: Done — `Unknown = 0` is the default for the public and internal enums (`None = 0` in the runtime-tests `[Flags]` variant).

src/Uno.UI/Helpers/UnoRuntimePlatform.cs
MartinZikmund: Done — the enum, helper and extensions are `internal` (namespace `Uno.Helpers`) in the Uno.WinRT projects; only the `Uno.UI.Toolkit` versions are public.

src/Uno.UI/Helpers/PlatformRuntimeHelper.cs
MartinZikmund: Done — `RuntimePlatformHelper` is `internal` in `Uno.Helpers`; the public surface is `Uno.UI.Toolkit.RuntimePlatformHelper`.

src/Uno.UI/Helpers/Xaml/UnoRuntimePlatformExtensions.cs
MartinZikmund: Done — the extension methods are `internal` in `Uno.Helpers` and `public` only when the shared source is linked into `Uno.UI.Toolkit`.

MartinZikmund:
### Finalized + rebased onto latest `master` (6.7)

The branch was rebased onto current `origin/master` (the previous base was ~4800 commits behind) and the incremental commits were squashed into a single feature commit. The implementation was completed to a consistent, compiling state. Summary of what changed:

**Shared enum** (`src/Uno.UWP/Helpers/UnoRuntimePlatform.cs`) — one source, three forms via `#if`:
- `internal RuntimePlatform` (`Uno.Helpers`) in the Uno.WinRT projects
- `public RuntimeTestPlatforms` (`[Flags]`, `Microsoft.VisualStudio.TestTools.UnitTesting`) in `Uno.UI.RuntimeTests` — kept byte-identical in behavior to the existing `master` enum so the test-conditioning (`PlatformConditionAttribute`/`RuntimeTestsPlatformHelper`) is unaffected
- `public UnoRuntimePlatform` (`Uno.UI.Toolkit`) — the public API. `Unknown = 0` is the default.

**Detection** (`Uno.Helpers.RuntimePlatformHelper`, internal): `GetPlatform()` is now runtime-first — a Skia host registers the actual platform via `SkiaPlatform`, which takes precedence over the compile-time native ladder. This fixes the "doesn't correctly check Skia vs non-Skia" item: since platform symbols are mutually exclusive per build, the host-set value is authoritative for all Skia targets (incl. Skia-on-Android/Apple/WASM).

**Public surface** (`Uno.UI.Toolkit`): `RuntimePlatformHelper.Current` + the `IsSkia()/IsIOS()/IsAndroid()/IsWasm()/IsMacCatalyst()/IsTvOS()/IsWindowsAppSdk()` extensions are now public (shared source linked into every Toolkit variant). The Uno-internal mirror stays `internal`.

**Hosts**: each Skia host sets its platform in its (static where available) constructor using `Uno.Helpers.RuntimePlatformHelper`/`RuntimePlatform`. The redundant X11 instance-ctor assignment was removed.

**Notable decisions (please confirm):**
- `SkiaAppleUIKit` was dropped; the AppleUIKit host now resolves the specific per-OS value (`SkiaIOS`/`SkiaTvOS`/`SkiaMacCatalyst`) at runtime, consistent with `Skia.Android → SkiaAndroid` and so `IsIOS()`/`IsTvOS()`/`IsMacCatalyst()` stay correct.
- The WPF Skia host (`Uno.UI.Runtime.Skia.Wpf`) and `Uno.UI.XamlHost.Skia.Wpf` no longer exist on `master` (6.7), so their changes were dropped and `SkiaWpf` is not part of the enum.

**Tests / assert**: `Given_RuntimePlatformDetected` validates the public API (`Current != Unknown` everywhere, `IsSkia()` on Skia, `NativeWinUI` on WinAppSDK) using `[PlatformCondition]`. A SamplesApp startup assert (`AssertRuntimePlatformIsKnown`) guards against unwired platforms.

**Validation**: the three enum/extension/helper configurations compile cleanly in isolation (Toolkit / Uno.WinRT / RuntimeTests forms, 0 warnings). A full multi-target build + the runtime tests still need to run on CI.

The earlier inline review threads (including the Copilot findings, which targeted the pre-refactor structure) have been addressed and resolved.

PR #18899, opened 2024-11-22
perf(brush): Don't use reflection to invoke brush updates

jeromelaban:


src/Uno.UI/UI/Xaml/Media/Brush.cs
```diff
@@ -9,20 +10,23 @@
...
+using Windows.UI.Core;
 #endif
 
 namespace Microsoft.UI.Xaml.Media
 {
 	[TypeConverter(typeof(BrushConverter))]
 	public partial class Brush : DependencyObject
 	{
-		private readonly WeakEventManager _weakEventManager = new();
+		private List<WeakEventHelper.GenericEventHandler>? _invalidateRenderHandlers;
```
Youssef1313, line 21: From my previous testing, I recall this was leaking.

I never seen perf issues from the current approach when profiling, and even if there is a perf issue, I feel @dr1rrb's approach is safer.

jeromelaban: Either previous approaches have issues, the original one is using reflection and is very slow on wasm, and the used by David uses CWT which is costly as well. I agree that it might on some instances, particularly if instances are not collected, but we can improve that with GCHandles and gen2 GC detection now that we're in net8+.

Youssef1313: If it's problematic on Wasm specifically I think you might consider going with conditional code to use different approach on different platforms?

The main issue with WeakEventHelper is that the `WeakReference` instances are leaked (which is not detected by `When_Add_Remove`, still it's a memory leak).

On desktop, I never seen a perf issue for the existing approach, and it doesn't leak. I can't speak of other platforms though as I haven't profiled on them.

The real issue is that there is no good solution for weak events in C#. Everyone is inventing their own solution, and all the solutions have problems.

jeromelaban: I understand you concerns, but yes, it's sufficient, and using reflection for such a critical path is not a good way to handle this for any platform. I agree that the origin of the problem has not been addressed, but every version of the runtime provides its features that provide a new way for handling it.

I'll try with the GCHandle approach, we'll see.

src/SamplesApp/SamplesApp.Wasm/SamplesApp.Wasm.csproj
```diff
@@ -17,6 +17,9 @@
 		<WasmShellIncludeWindowsCompatibility>false</WasmShellIncludeWindowsCompatibility>
 		<AllowUnsafeBlocks>true</AllowUnsafeBlocks>
 		
+		<!-- Required for net9 workloads -->
+		<AppendRuntimeIdentifierToOutputPath>false</AppendRuntimeIdentifierToOutputPath>
```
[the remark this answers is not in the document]
jeromelaban: It is, it's needed since net9 stable to run the samples app.

jeromelaban:

jeromelaban:
@Mergifyio backport release/stable/5.4

PR #18910, opened 2024-11-25
perf: Don't use multicast delegates for DependencyObjectCollection.VectorChanged

jeromelaban:
This change avoids the high cost of multicast delegates, particularly on mono-AOT environments. ~We also do not need the thread safety guarantees~.

Example:

image

### DIFF
--- a/src/Uno.UI/UI/Xaml/DependencyObjectCollection.cs
+++ b/src/Uno.UI/UI/Xaml/DependencyObjectCollection.cs
@@ -28,7 +28,8 @@ namespace Microsoft.UI.Xaml
 	{
 		private object _vectorChangedHandlersLock = new();
 
-		// Explicit handlers list to avoid the cost of multicast delegates handling
+		// Explicit handlers list to avoid the cost of generic multicast
+		// delegates handling on mono's AOT.
 		private List<VectorChangedEventHandler<T>> _vectorChangedHandlers;
 
 		public event VectorChangedEventHandler<T> VectorChanged
--- a/src/Uno.UI/UI/Xaml/DependencyObjectCollection.cs
+++ b/src/Uno.UI/UI/Xaml/DependencyObjectCollection.cs
@@ -45,7 +45,14 @@ namespace Microsoft.UI.Xaml
 			{
 				lock (_vectorChangedHandlersLock)
 				{
-					(_vectorChangedHandlers ??= new()).Remove(value);
+					var list = _vectorChangedHandlers ??= new();
+
+					var lastIndex = list.LastIndexOf(value);
+
+					if (lastIndex != -1)
+					{
+						list.RemoveAt(lastIndex);
+					}
 				}
 			}
 		}
@@ -229,32 +236,59 @@ namespace Microsoft.UI.Xaml
 
 		private void RaiseVectorChanged(CollectionChange change, int index)
 		{
-			lock (_vectorChangedHandlersLock)
+			// Gets an executable list that does not need to be locked
+			int GetInvocationList(out VectorChangedEventHandler<T> single, out VectorChangedEventHandler<T>[] array)
 			{
-				if (_vectorChangedHandlers is { Count: > 0 })
+				lock (_vectorChangedHandlersLock)
 				{
-					var args = new VectorChangedEventArgs(change, (uint)index);
-
-					if (_vectorChangedHandlers.Count == 1)
+					if (_vectorChangedHandlers is { Count: > 0 })
 					{
-						_vectorChangedHandlers[0].Invoke(this, args);
-					}
-					else
-					{
-						// Clone the array to account for reentrancy.
-						var handlersClone = ArrayPool<VectorChangedEventHandler<T>>.Shared.Rent(_vectorChangedHandlers.Count);
-						_vectorChangedHandlers.CopyTo(handlersClone, 0);
-
-						// Use the original count, the rented array may be larger.
-						var count = _vectorChangedHandlers.Count;
-
-						for (int i = 0; i < count; i++)
+						if (_vectorChangedHandlers.Count == 1)
 						{
-							handlersClone[i].Invoke(this, args);
+							single = _vectorChangedHandlers[0];
+							array = null;
+							return 1;
 						}
+						else
+						{
+							single = null;
 
-						ArrayPool<VectorChangedEventHandler<T>>.Shared.Return(handlersClone);
+							array = ArrayPool<VectorChangedEventHandler<T>>.Shared.Rent(_vectorChangedHandlers.Count);
+							_vectorChangedHandlers.CopyTo(array, 0);
+
+							return _vectorChangedHandlers.Count;
+						}
 					}
+				}
+
+				single = null;
+				array = null;
+				return 0;
+			}
+
+			var count = GetInvocationList(out var single, out var array);
+
+			if (count > 0)
+			{
+				var args = new VectorChangedEventArgs(change, (uint)index);
+
+				if (count == 1)
+				{
+					single.Invoke(this, args);
+				}
+				else
+				{
+					for (int i = 0; i < count; i++)
+					{
+						ref var handler = ref array[i];
+						handler.Invoke(this, args);
+
+						// Clear the handle immediately, so we don't
+						// call ArrayPool.Return with clear.
+						handler = null;
+					}
+
+					ArrayPool<VectorChangedEventHandler<T>>.Shared.Return(array);
 				}
 			}
 		}
--- a/src/Uno.UI/UI/Xaml/DependencyObjectCollection.cs
+++ b/src/Uno.UI/UI/Xaml/DependencyObjectCollection.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Buffers;
 using System.Collections;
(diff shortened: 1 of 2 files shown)


jeromelaban:

src/Uno.UI/UI/Xaml/DependencyObjectCollection.cs
```diff
@@ -205,7 +212,17 @@ internal List<T>.Enumerator GetEnumeratorFast()
 			=> _list.GetEnumerator();
 
 		private void RaiseVectorChanged(CollectionChange change, int index)
-			=> VectorChanged?.Invoke(this, new VectorChangedEventArgs(change, (uint)index));
+		{
+			if (_vectorChangedHandlers is not null)
```
Youssef1313, line 216: This avoids the event args allocation in case an event subscription happened then it was unsubscribed. In this case, the list will be empty and not null.

```suggestion
			if (_vectorChangedHandlers is { Count: > 0 })
```

src/Uno.UI/UI/Xaml/DependencyObjectCollection.cs
```diff
@@ -205,7 +212,17 @@ internal List<T>.Enumerator GetEnumeratorFast()
...
+			{
+				var args = new VectorChangedEventArgs(change, (uint)index);
+
+				foreach (var handler in _vectorChangedHandlers)
+				{
+					handler.Invoke(this, args);
+				}
```
Youssef1313, line 223: This feels too dangerous tbh. I don't see any guarantee that the handler won't attempt to subscribe/unsubscribe from the event which would mutate the list.

src/Uno.UI/UI/Xaml/DependencyObjectCollection.cs
```diff
@@ -25,7 +25,14 @@ public partial class DependencyObjectCollectionBase : DependencyObject
...
 	{
-		public event VectorChangedEventHandler<T> VectorChanged;
+		// Explicit handlers list to avoid the cost of multicast delegates handling
+		private List<VectorChangedEventHandler<T>> _vectorChangedHandlers;
+
+		public event VectorChangedEventHandler<T> VectorChanged
+		{
+			add => (_vectorChangedHandlers ??= new()).Add(value);
+			remove => (_vectorChangedHandlers ??= new()).Remove(value);
+		}
```
Youssef1313, line 36: I'm pretty confident that the subscription/unsubscription is implemented in a thread-safe way in WinUI. If you do this:

```csharp
        private async void myButton_Click(object sender, RoutedEventArgs e)
        {
            DependencyObjectCollection x = new();
            for (int i = 0; i < 100; i++)
            {
                int iCopy = i;
                _ = Task.Run(() =>
                {
                    x.VectorChanged += (sender, ev) =>
                    {
                        Debug.WriteLine($"VectorChanged: {iCopy}");
                    };
                });
            }

            await Task.Delay(500); // give a chance to all tasks to be completed.
            x.Add(new Button());
        }
```

This should print the numbers from 0 to 99 (not ordered for sure).

So definitely the change in the PR is a divergence from WinUI. And people sometimes does crazy things with Reactive and execute code in background threads. So it's not very unlikely that this will break something working already.

src/Uno.UI/UI/Xaml/DependencyObjectCollection.cs
jeromelaban: Agreed, there's a need for copying when executing if the list changes, though copying may cause a different set of trouble, will looking into it further.

src/Uno.UI/UI/Xaml/DependencyObjectCollection.cs
jeromelaban: It may be, though in those objects are not supposed to be manipulated from a background thread. Fair comment anyways, it can't be moved away from this kind of behavior.

jeromelaban: I've adjusted the locking and enumeration, that should do for those scenarios.

src/Uno.UI/UI/Xaml/DependencyObjectCollection.cs
```diff
@@ -205,7 +228,36 @@ internal List<T>.Enumerator GetEnumeratorFast()
 			=> _list.GetEnumerator();
 
 		private void RaiseVectorChanged(CollectionChange change, int index)
-			=> VectorChanged?.Invoke(this, new VectorChangedEventArgs(change, (uint)index));
+		{
```
Youssef1313, line 231: I'm wondering what will happen in this case:

```csharp
DependencyObjectCollection x = new();
x.VectorChanged += (_, _) =>
{
    var handle = new AutoResetEvent(false);

    _ = Task.Run(() =>
    {
        // Is this going to dead lock?
        x.VectorChanged += (_, _) => { };
        handle.Set();
    });

    handle.WaitOne();
};
```

Sorry for such crazy scenario. While it may be unlikely to see it in user apps, I think you may still be looking forward to having a correct implementation.

src/Uno.UI/UI/Xaml/DependencyObjectCollection.cs
```diff
@@ -25,7 +26,29 @@ public partial class DependencyObjectCollectionBase : DependencyObject
...
+				{
+					(_vectorChangedHandlers ??= new()).Add(value);
+				}
+			}
+
+			remove
+			{
+				lock (_vectorChangedHandlersLock)
+				{
+					(_vectorChangedHandlers ??= new()).Remove(value);
```
Youssef1313, line 48: This is a change in behavior I believe. `List<T>.Remove` will remove the first occurrence, while regular events removes the last occurrence.

So, if you do:

```
collection.VectorChanged += firstHandler;
collection.VectorChanged += secondHandler;
collection.VectorChanged += firstHandler; // EXACTLY the same instance.
collection.VectorChanged -= firstHandler;

// something that triggers the event.

// which handler is invoked first?
```

Youssef1313: For this one, it might be good to also verify what is the behavior of WinUI, as I'm not very sure what WinUI behavior here is. I haven't verified.

src/Uno.UI/UI/Xaml/DependencyObjectCollection.cs
jeromelaban: That's fair, It's very unlikely, but possible nonetheless. I've adjusted it.

src/Uno.UI/UI/Xaml/DependencyObjectCollection.cs
jeromelaban: It's removing the last. I've adjusted it.

src/Uno.UI/UI/Xaml/DependencyObjectCollection.cs
Youssef1313: > It's very unlikely

Yup. Hence mentioning "crazy scenario" 😄

src/Uno.UI/UI/Xaml/DependencyObjectCollection.cs
```diff
@@ -25,7 +26,37 @@ public partial class DependencyObjectCollectionBase : DependencyObject
...
+		private List<VectorChangedEventHandler<T>> _vectorChangedHandlers;
+
+		public event VectorChangedEventHandler<T> VectorChanged
+		{
+			add
+			{
+				lock (_vectorChangedHandlersLock)
+				{
+					(_vectorChangedHandlers ??= new()).Add(value);
+				}
```
carldebilly, line 42: Adding twice the same delegate will register it twice?

carldebilly (approved):

src/Uno.UI/UI/Xaml/DependencyObjectCollection.cs
```diff
@@ -25,7 +26,37 @@ public partial class DependencyObjectCollectionBase : DependencyObject
...
+
+		// Explicit handlers list to avoid the cost of generic multicast
+		// delegates handling on mono's AOT.
+		private List<VectorChangedEventHandler<T>> _vectorChangedHandlers;
+
+		public event VectorChangedEventHandler<T> VectorChanged
+		{
+			add
+			{
+				lock (_vectorChangedHandlersLock)
```
dr1rrb, line 39: I would assume we don't need a `lock` here since it's about `DependencyObject` the event should be touched only from the UI thread?

dr1rrb (approved):

src/Uno.UI/UI/Xaml/DependencyObjectCollection.cs
jeromelaban: Like it would for a normal event.

src/Uno.UI/UI/Xaml/DependencyObjectCollection.cs
jeromelaban: They may be registered from a background thread, even if unlikely. It's there to keep the original behavior.

jeromelaban:
@Mergifyio backport release/stable/5.5

jeromelaban:
@Mergifyio backport release/stable/5.4

PR #18952, opened 2024-11-28
Revert "chore: Revert "Make skia runtime tests green locally on Debug""

ramezgerges:
It doesn't seem like #18909 was the culprit after all, so let's try again.

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/AppBar/AppBar.Partial.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/AppBar/AppBar.Partial.cs
@@ -769,7 +769,8 @@ namespace Microsoft.UI.Xaml.Controls
 		{
 			// If the AppBar is not live, then wait until it's loaded before
 			// responding to changes to opened state and firing our Opening/Opened events.
-			if (!IsInLiveTree)
+			// Uno Specific: using IsLoaded instead of IsInLiveTree, which makes more sense because OnOpening (called below) -> SetupOverlayState expects OnApplyTemplate to have already been called
+			if (!IsLoaded)
 			{
 				return;
 			}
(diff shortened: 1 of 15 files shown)


jeromelaban (approved):

ajpinedam (approved):

src/Uno.UI/UI/Xaml/Controls/AppBar/AppBar.Partial.cs
```diff
@@ -769,7 +769,8 @@ private void OnIsOpenChanged(bool isOpen)
 		{
 			// If the AppBar is not live, then wait until it's loaded before
 			// responding to changes to opened state and firing our Opening/Opened events.
-			if (!IsInLiveTree)
+			// Uno Specific: using IsLoaded instead of IsInLiveTree, which makes more sense because OnOpening (called below) -> SetupOverlayState expects OnApplyTemplate to have already been called
+			if (!IsLoaded)
```
Youssef1313, line 773: This change looks suspicious to me. Why is ported code needing change now?

ramezgerges: A bug appears when you try to set `IsOpen` before loading. `OnIsOpenChanged` calls `SetupOverlayState`, which expects `m_tpLayoutRoot` to be set already. `OnApplyTemplate` only fires in `OnPostLoading` (I'm not sure if you aligned this part or not), so we need the logic to check for `IsLoaded` instead of `IsInLiveTree`.

Youssef1313: Template application became more aligned.

If there is still a behavior mismatch, it would be good to at least explain it precisely in the comment.

The other thing is, why would such a bug happen now with the other changes? Or is this change unrelated to the rest? It's just so confusing to see such a change in this PR.

ramezgerges: The bug (and the change) were discovered because `CanEnterAndLeaveLiveTree` was failing, so it went next to the other "make tests green" changes. The name of the test is in the commit message (although in hindsight, I should've probably added it to the comment and made it clearer overall). It could be improved, but I don't think it's worth another CI run.

PR #19228, opened 2025-01-14
perf(anim): Don't refresh triggers and transitions during parsing

jeromelaban:
## What is the new behavior?

`VisualState` instances changes in `VisualStateGroup` won't trigger updates during XAML parsing, avoiding intermediate discarded animations.


jeromelaban:

PR #19411, opened 2025-01-31
fix: Fix possible null ref

dr1rrb:
## Bugfix
- Fix possible null ref
- Don't skip elements without handlers


jeromelaban:

PR #19438, opened 2025-02-04
fix(hr): Don't import Microsoft.IO.RecyclableMemoryStream using nuget

jeromelaban:
This avoids importing this package into the running app and cause dependency issues if the package is updated by the user.


ajpinedam (approved):
Good

jeromelaban:

Issue #19433, opened 2025-02-04
ENC1002 error when deleting item with x:Name set


nickrandolph:
### Current behavior

Start with simple page that contains a Button with x:Name set

```xml
<Page x:Class="UnoApp36.MainPage"
      xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
      xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
      xmlns:local="using:UnoApp36"
      Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">
    <ScrollViewer>
        <Grid>
            <StackPanel HorizontalAlignment="Center"
                        VerticalAlignment="Center">
                <Button x:Name="MyButton" Content="Button1"  />
            </StackPanel>
        </Grid>
    </ScrollViewer>
</Page>
```
Run the application
Delete the Button (in XAML in VS)
Trigger HR

![Image](https://github.com/user-attachments/assets/937db33b-7c93-4388-a3f6-ab6d349a0e3e)

### Works on UWP/WinUI

None

nickrandolph:
Note that this happens even using the fix from https://github.com/unoplatform/uno/pull/19418
ie "Uno.Sdk.Private": "5.6.44"

jeromelaban:
I cannot repro in VS 17.13 Preview 3

dr1rrb:
I do repro
VS 17.12.1
Sdk.Private 5.6.44

looking at it

dr1rrb:
Note: If I do any changes before removing the button (like changing the content of the button), then I'm unable to repro

agneszitte:
I am not able to reproduce this issue with latest **Uno.Sdk 5.6.26**.

--- 
Tested with:
- Templates/SDK: **5.6.26**
- VS version: VS 2022 (64-bit) - Version **17.12.4**
- VS VSIX: [5.6.23](https://dev.azure.com/uno-platform/uno-private/_build/results?buildId=154495)
- [BlankAppVsWizardNet9.zip](https://github.com/user-attachments/files/18713878/BlankAppVsWizardNet9.zip)

agneszitte:
I am **not** able to reproduce this issue with latest **Uno.Sdk 5.6.28**.

--- 
Tested with:
- Templates/SDK: **5.6.28**
- VS version: VS 2022 (64-bit) - Version **17.12.4**
[...]

PR #19440, opened 2025-02-04
fix(vs): Use single TFM build feature from VS

jeromelaban:
## PR Type

- Bugfix

## What is the new behavior?

Removes workaround used to build a single target framework by rewriting the TFMs, instead use the ootb feature from VS. This also fixes multi-TFMs issues with `ENC1002: Cannot apply changes -- unexpected error: 'Object reference not set to an instance of an object.'`

Fixes https://github.com/unoplatform/uno/issues/19433, fixes https://github.com/unoplatform/uno.hotdesign/issues/3290


jeromelaban:

dr1rrb (approved):

PR #19638, opened 2025-03-05
fix: android PhoneCallManager

### LINKED ISSUE #19633: [Android] PhoneCallManager not working on API 31+
### Current behavior

Using `PhoneCallManager.ShowPhoneCallUI("123456789", "UnoApp142");` throws a `SecurityException` when calling on an Android device with API 31+

See: https://developer.android.com/reference/android/telephony/TelephonyManager#listen(android.telephony.PhoneStateListener,%20int)

We should be handling the new suggested method in the android docs from API 31 and above. This new method may require a permissions check for READ_PHONE_STATE and possibly others, as defined in the Android docs

[UnoApp142.zip](https://github.com/user-attachments/files/19075582/UnoApp142.zip)

### Expected behavior

It should behave like older Android versions:

Image

rajamatt:
GitHub Issue (If applicable): closes #19633 

## PR Type

- Bugfix

Telephony.Listen is deprecated for API >= 31

## What is the new behavior?

Now we use Telephony.Listen if API < 31 and Telephony.RegisterTelephonyCallback if API >= 31

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UWP/ApplicationModel/Calls/PhoneCallManager.Android.cs
+++ b/src/Uno.UWP/ApplicationModel/Calls/PhoneCallManager.Android.cs
@@ -33,19 +33,23 @@ namespace Windows.ApplicationModel.Calls
 			}
 			else
 			{
-				TryRegisterTelephonyCallbackAsync();
+				RegisterTelephonyCallbackAsync();
 			}
 #pragma warning restore CA1422 // Validate platform compatibility
 #pragma warning restore CS0618 // TelephonyManager is obsolete in API 31
 		}
 
-		private static async void TryRegisterTelephonyCallbackAsync()
+		private static async void RegisterTelephonyCallbackAsync()
 		{
+#pragma warning disable CS0618 // TelephonyManager is obsolete in API 31
+#pragma warning disable CA1422 // Validate platform compatibility
 			if (await Extensions.PermissionsHelper.CheckPermission(CancellationToken.None, Android.Manifest.Permission.ReadPhoneState) ||
 				await Extensions.PermissionsHelper.TryGetPermission(CancellationToken.None, Android.Manifest.Permission.ReadPhoneState))
 			{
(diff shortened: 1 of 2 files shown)


src/Uno.UWP/ApplicationModel/Calls/PhoneCallManager.Android.cs
```diff
@@ -20,11 +21,33 @@ static PhoneCallManager()
...
+#pragma warning restore CA1422 // Validate platform compatibility
+#pragma warning restore CS0618 // TelephonyManager is obsolete in API 31
+		}
+
+		private static async void RegisterTelephonyCallbackAsync()
+		{
 #pragma warning disable CS0618 // TelephonyManager is obsolete in API 31
 #pragma warning disable CA1422 // Validate platform compatibility
-			_telephonyManager.Listen(new CallStateListener(), PhoneStateListenerFlags.CallState);
+			if (await Extensions.PermissionsHelper.CheckPermission(CancellationToken.None, Android.Manifest.Permission.ReadPhoneState) ||
```
kazo0, line 46: we should only need the `TryGetPermission` call here. It will return `true` if the permission has already been granted

rajamatt: should we add an exception if the permission isn't allowed? like this https://github.com/unoplatform/uno/blob/f8382a25892a8848cb17a6bbc9a7dcfe254971c5/src/Uno.UWP/Media/SpeechRecognition/SpeechRecognizer.Android.cs#L24-L27

rajamatt: I don't think we should be throwing since READ_PHONE_STATE isn't needed apart from ShowPhoneCallUI(). And would also likely lead to a crash if we throw.

Normally this specific permission is granted at runtime, along with other Phone permissions that aren't needed in this case. If it's included in the manifest then that's fine too.

kazo0 (approved):

kazo0:
@Mergifyio backport release/stable/5.6

PR #19644, opened 2025-03-06
feat: Update `AppBar` to winui3/release/1.6.5

MartinZikmund:

## PR Type

## PR Checklist

## Other information


MartinZikmund:
Blocked by https://github.com/unoplatform/uno/issues/19649

MartinZikmund:
Closing in favour of #20196, which now subsumes this PR.

Both PRs set out to port `AppBar` from the WinUI C++ sources. #20196 is the one that got there: it ports `AppBar_Partial.{h,cpp}` **and** `ApplicationBarService_Partial.cpp` (~2,850 lines) and wires them into `Page`, `XamlRoot` and `BackButtonIntegration`, so `Page.TopAppBar`/`BottomAppBar` actually do something. This branch never left draft, has had no commits since 2025-03-06, and does not compile at its head — `ShouldUseParentedLTE` uses an undeclared `rootDO` and contains `rootDO is PopupRoot popupRoot is { }`, which is not valid C#; and `OnLoaded`/`OnUnloaded`/`OnBackButtonPressedImpl` call an `ApplicationBarService`, a `XamlRoot.GetApplicationBarService()` and a `Page.AppBarClosedSizeChanged()` that this branch's 11 files never add.

I also checked the one thing that might have argued for keeping this branch — its `winui3/release/1.6.5` target versus #20196's `1.7.1` — and it makes no difference. Seven of the nine `AppBar*` files are byte-identical between the two tags, and the only two hunks that differ in `AppBar_Partial.{cpp,h}` resolve to identical behaviour once the shipped Containment change-IDs are enabled.

**Grafted across into #20196 before closing:**

- `48b45f9a8e0` — the generic `LightDismissOverlayHelper`. This branch's version is the faithful port of WinUI's `LightDismissOverlayHelper.h`, which is a free-standing helper with a `template <typename T> ResolveIsOverlayVisibleForControl` plus a separate `IsOverlayVisibleForMode`. #20196 had collapsed it into a single `AppBar`-typed method inside the `AppBar` folder and dropped `IsOverlayVisibleForMode`. The two-member generic shape, the `IHasLightDismissOverlay` interface standing in for the C++ duck typing, and the file's location outside `AppBar` are all now in #20196, so `ContentDialog`/`Flyout`/`CommandBar` can reuse it when they're ported. Same commit routes `Auto` mode through `XboxUtility.IsOnXbox()`, which is the WinUI call site.
- `dde393f46eb` — the XML doc comments on the public `AppBarClosedDisplayMode` enum, and WinUI's verbatim `AppBarMode.Inline` note from `JoltClasses.h:32`.

**Deliberately not carried over:**

- `AppBar.Native.cs` (`#if HAS_NATIVE_COMMANDBAR`) — that symbol is defined nowhere in the repo, so the file is unreachable in every build configuration, and `ICustomClippingElement` has no counterpart in `AppBar_Partial.h`. It's also the same class of dead conditional that #20196's reviewers spent three rounds getting removed for the 7.0 Skia-only line.
- `XboxUtility.IsOnXbox => false` — the name was worth keeping, the body was not. Hardcoding `false` would have replaced Uno's real cached device-family probe (`SharedHelpers.IsOnXbox()`) with a constant and silently broken `LightDismissOverlayMode="Auto"` on Xbox. #20196 now has the WinUI call-site name forwarding to the working probe.
- The `AppBar.Partial.cs` → `AppBar.partial.mux.cs` split, the `AppBar.partial.h.mux.cs` header, the MUX reference-comment convention and the private `TemplateSettings` setter — all already present in #20196 independently, so nothing is lost.
- The ported `AppBar.partial.mux.cs` body itself. Beyond the compile errors, its `TryQueryDisplayModesStatesGroup` fix combined with the retained `if (IsOpen == isOpen)` fallback would have made `Opened`/`Closed` fire twice per toggle — worse than `master`. #20196 gets both halves right.

Thanks for the `LightDismissOverlayHelper` shape — that was the genuinely better call of the two, and it's the reason #20196 is now closer to the C++ than it was.

PR #19789, opened 2025-03-31
test: Adjust test retrie filter syntax

jeromelaban:


jeromelaban:
Issue is not about the filter syntax https://github.com/unoplatform/uno/issues/19794

PR #19883, opened 2025-04-09
Fix Image_ImageSource_PixelSize

jeromelaban:


src/Uno.UI/UI/Xaml/Media/ImageSource.crossruntime.cs
```diff
@@ -105,10 +105,22 @@ private void Open()
...
 				else if (TryOpenSourceAsync(ct, null, null, out var asyncImg))
 				{
+#if __WASM__
+					// On Wasm single-threaded ContinueWith with
+					// TaskContinuationOptions/TaskScheduler is not functional
+					asyncImg.ContinueWith(t =>
+					{
+						if (t.IsCompletedSuccessfully)
+						{
+							OnOpened(t.Result);
```
dlemstra, line 115: Saw this PR and was curious. Should you not also do `OnOpened(t.IsFaulted ? ImageData.FromError(t.Exception) : t.Result);` here? And just put `}, ct, TaskContinuationOptions.None, TaskScheduler.FromCurrentSynchronizationContext());` inside an ifdef?

PR #19905, opened 2025-04-10
fix: Remove unused x64 dependency

jeromelaban:


kazo0 (approved):

src/Uno.UI.Runtime.Skia.Win32.Support/NativeMethods.txt
```diff
@@ -148,7 +148,6 @@ GetPointerPenInfo
 GetPointerTouchInfo
 GetPointerType
 GetRegionData
-GetSaveFileName
```
kazo0, line 151: Just tagging to make sure this wasn't an accidental change, not sure if related

PR #20139, opened 2025-04-30
fix: Adjust for linker dependency ordering

jeromelaban:
Ensures that the xaml trimming pass provides the proper parameters to the linker.


jeromelaban:
@mergifyio backport release/stable/6.0

spouliot (approved):

PR #20166, opened 2025-05-02
feat(import external docs): allow contributor url for import

DevTKSS:

## PR Type

- Bugfix
- Feature

- Script allows only importing from uno github url, which makes it difficult (for me not possible) for non-team contributors to validate the PR if it includes doc changes that would be best to build locally
- while it takes one simple paramter, its not following current PS v7.5 structure with begin-process-end and the function the current code contains lies mid in the executed code which leaves unclear when its called (at this line or maybe only if called by name?) seperated this below the main function to clear this up.
- using alias instead of Get-... Set-... is not recommended from PSAnalyzer so updated this.

## What is the new behavior?

- Added a parameter that extends the script capabilities to use the custom github url and not only (until now) the sha to import.
- Changed the PS to advanced script, not only for brevity than also to future allow Verbose, Debug cmdlet and Testing this script
- Changed alias to current recommended

## PR Checklist

- [x] Contains **NO** breaking changes => to be evaluated, but I dont think its a breaking change

## Other information

Problem this need a last bit of help with:
after running the main script, when it returns to the known `_Test.ps1` the Script executing path does not match coming from the Location change in the main script for importing the externals.
thats most likly just a quick fix but I do not know.

causing problems for most of my docs PR, e.g.:
- https://github.com/unoplatform/uno.extensions/pull/2734
- https://github.com/unoplatform/uno.extensions/pull/2710
- etc.

### DIFF
--- a/doc/import_external_docs.ps1
+++ b/doc/import_external_docs.ps1
@@ -68,7 +68,7 @@ function Invoke-ImportExternalDocs {
             $repoBranch = $external_docs[$repoPath]
 
             # if the repository is already cloned, just update it, not clone it again, because git will otherwise fail
-            if (-Not (Test-Path $repoPath -and (Get-Location -Stack).Path -contains (Resolve-Path $repoPath).Path)) {
+            if (-Not (Test-Path $repoPath)) {
                 Write-Output "Cloning $repoPath ($repoUrl@$repoBranch)..."
                 git clone $repoUrl $repoPath
                 Assert-ExitCodeIsZero -ErrorMessage "Failed to clone repository $repoUrl."
--- a/doc/import_external_docs.ps1
+++ b/doc/import_external_docs.ps1
@@ -98,4 +98,25 @@ git config advice.detachedHead $detachedHeadConfig
 
 Pop-Location
 
-Set-PSDebug -Off
+        if ($hasErrors) {
+            Write-Output "Some external repositories failed to import or update. Check the logs for details."
+        }
+        else {
+            Write-Output "All external repositories are imported and up to date."
+        }
+        Write-Output "All external repositories are imported and up to date."
+    }
+}
+
+function Assert-ExitCodeIsZero {
+    param (
+        [string]$ErrorMessage = "A command failed with a non-zero exit code."
+    )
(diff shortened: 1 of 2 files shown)


DevTKSS:
@kazo0 @agneszitte tryed to fix the import script, but this need a last bit of help to be ready

kazo0:
@jeromelaban I'm not sure if this is something we'd want to change? However, what would be the easiest way for @DevTKSS to be able to test their docs changes locally given that they are on a fork?

I was imagining that they would be able to change this line: https://github.com/unoplatform/uno/blob/c753d158026c852b1303ce1027471e91063471b8/doc/import_external_docs.ps1#L24 to point to their fork url instead and they could comment out the unneeded repos from here: https://github.com/unoplatform/uno/blob/c753d158026c852b1303ce1027471e91063471b8/doc/import_external_docs_test.ps1#L9-L20

DevTKSS:
@kazo0 so like do it in a two step thing? but that would potentially fail because after the import, the docfx and dotnet-serve command is called right after, so this will defintly fail. And currently, when I maybe did already run this once (and its telling me the branch with the sha of my fork is not on the uno-origin (of corse not ;) ) ) then on second run where I maybe would have added another commit to fix links, then it is failing because the repo (no matter which) is already existing at external, directory, so maybe this part of the problem would require a clean up before checking out again? meaning the checkouts of the uno-origin, non fork external imports if unclear worded.

DevTKSS:
currently that partial import would fail also because the docfx would attempt to resolve all links no matter if it might be the first of two import steps, which is of corse not possible until all externals are there.

doc/import_external_docs.ps1
```diff
@@ -1,101 +1,124 @@
...
+        Write-Output "Current setup:"
+        $external_docs
+
+        $ErrorActionPreference = 'Stop'
+       
+        if (-Not (Test-Path articles\external)) {
+            mkdir articles\external -ErrorAction Continue
+        }
+
+        Push-Location articles\external
```
jeromelaban, line 52: It's best to avoid cosmetic changes with functional changes. Could you make a PR only with the custom git change?

PR #20298, opened 2025-05-09
feat(skia): collapse visual subtrees into one SKPicture when the subtree doesn't change for a bit

### LINKED ISSUE #20282: [Skia] Improve partial rendering

We're already using partial rendering with SKPicture for some areas, but we should be using recursive rendering for portions of the tree without having to recurse into visuals if they have not changed.

This is particularly important for WebAssembly where with the interpreter, a lot of managed code is executed when drawing frames even if visuals have not changed, and moving that rendering to Skia would bring it down to pure wasm.

### Why is this needed

Wasm intepreter performance, and to reduce the overall CPU consumption when drawing frames, on all targets..

### Renderer

- [ ] Skia
- [ ] Native

### For which platforms

_No response_

Revive this PR: https://github.com/unoplatform/uno/pull/19829

ramezgerges:
GitHub Issue (If applicable): closes https://github.com/unoplatform/uno/issues/20282

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI.Composition/Composition/ContainerVisual.skia.cs
+++ b/src/Uno.UI.Composition/Composition/ContainerVisual.skia.cs
@@ -172,7 +172,13 @@ public partial class ContainerVisual : Visual
 
 	internal override int GetSubTreeVisualCount()
 	{
-		_subtreeVisualCount ??= Children.Count + Children.InnerList.Aggregate(0, (acc, visual) => acc + visual.GetSubTreeVisualCount());
+		var acc = 0;
+		foreach (var visual in Children.InnerList)
+		{
+			acc += visual.GetSubTreeVisualCount();
+		}
+		_subtreeVisualCount ??= Children.Count + acc;
+
 		return _subtreeVisualCount.Value;
 	}
 }
--- a/src/Uno.UI.Composition/Composition/ContainerVisual.skia.cs
+++ b/src/Uno.UI.Composition/Composition/ContainerVisual.skia.cs
@@ -172,6 +172,10 @@ public partial class ContainerVisual : Visual
 
 	internal override int GetSubTreeVisualCount()
 	{
+		if (_subtreeVisualCount is { } count)
+		{
+			return count;
+		}
 		var acc = 0;
 		foreach (var visual in Children.InnerList)
 		{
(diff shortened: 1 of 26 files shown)


ajpinedam (approved):

src/Uno.UI.Composition/Composition/ContainerVisual.skia.cs
```diff
@@ -25,6 +26,13 @@ partial void InitializePartial()
 		{
 			IsChildrenRenderOrderDirty = true;
 
+			var parent = this;
+			while (parent is not null && parent._subtreeVisualCount is not null)
+			{
+				parent._subtreeVisualCount = null;
```
Youssef1313, line 32: Wouldn't it be more performant if you adjust this based on the given NotifyCollectionChangedAction?

- Remove: subtract one
- Add: add one
- Replace/Move: do nothing
- Reset: set back to null to re-calculate

ramezgerges: That's fair but the average number of children of a visual is likely around 2. Looping through 3 children and reading a cached value is hardly more expensive and it leaves the code a lot cleaner and easier to debug when we make invalidation bugs. We can revisit this if it shows up while profiling.

src/Uno.UI.Composition/Composition/ContainerVisual.skia.cs
```diff
@@ -161,4 +169,16 @@ internal override bool SetMatrixDirty()
...
+	internal override int GetSubTreeVisualCount()
+	{
+		var acc = 0;
+		foreach (var visual in Children.InnerList)
+		{
+			acc += visual.GetSubTreeVisualCount();
+		}
+		_subtreeVisualCount ??= Children.Count + acc;
+
+		return _subtreeVisualCount.Value;
```
Youssef1313, line 182: ⚠️ Here you will loop to calculate `acc` even if the value is already cached?

ramezgerges:
The differences are massive on skia/wasm where going to JS and back is expensive. In an artificial scenario with N buttons in a WrapPanel, the numbers before and after differ by a full order of magnitude in some cases:

After (fps and frame time):
N = 10 -> 50-52fps ~7.5ms
N = 14 (right before threshold for opt) -> 57fps ~5.5ms
N = 15 -> roughly the same as above
20 -> roughly the same as above
100 -> 30-35fps ~6.5ms
200 -> ~31fps 8.5ms (was 14fps 50ms for the first second or so before the opt hit)
500 -> ~19fps ~14ms (was ~14fps ~110ms for the first few seconds before the opt hit)
1000 -> ~19fps ~24ms (was ~4fps ~215ms for the first few seconds before the opt hit)

Before:
10 -> ~7ms
14 -> ~8.5ms
500 -> ~110ms
1000 -> ~200ms

jeromelaban:
@mergifyio backport release/stable/6.0

Issue #20476, opened 2025-05-21
[WASM] TextTrimming does not work in combination with Width


AuderixDev:
### Current behavior

WASM:

![Image](https://github.com/user-attachments/assets/ff20336b-db53-4d43-95d2-22d232143ce5)

CharacterEllipsis only works if WordWrapping is enabled. 
TextBlock-Size (Width) is calculated incorrect with TextTrimming/Wrapping enabled. 

This leads to a Left-Alignment, or it looks like it is left aligned....

### Expected behavior

1. I expected that, with `TextTrimming="CharacterEllipsis"` the 3 dots appear (`...`) like in WinAppSDK
2. I expected that, Trimming works with the `Width`/`MaxWidth` Property like in WinAppSDk
3. I expected that, if I set a `HorizontalAlignment="Center"` the trimmed Text will pe centered in the TextBlock like in the WinAppSDK

This is the same code on WinAppSDK:

![Image](https://github.com/user-attachments/assets/cd8f9477-4f65-44e5-a1f3-e3d71300734e)

### How to reproduce it (as minimally and precisely as possible)

http://playground.platform.uno/#bf60c50f

The code: 
```XAML
    <StackPanel Orientation="Vertical" HorizontalAlignment="Center">

        <TextBlock Text="This should end with dots: " />
        <Border Background="Green">
            <TextBlock Text="Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum."
                   TextTrimming="CharacterEllipsis"
                   HorizontalAlignment="Center"
                   MaxWidth="250"
                   MaxLines="1" />
        </Border>

        <TextBlock Text="This should be centered: " Margin="0,10,0,0" />
        <Border Background="Green">
[...]

PR #20478, opened 2025-05-21
fix: Avoid setting webkit properties for MaxLines = 1 when not wrapping

MartinZikmund:
GitHub Issue (If applicable): part of https://github.com/unoplatform/uno/issues/20476

## PR Type

- Bugfix

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI/UI/Xaml/UIElement.TextHelper.wasm.cs
+++ b/src/Uno.UI/UI/Xaml/UIElement.TextHelper.wasm.cs
@@ -134,15 +134,20 @@ namespace Microsoft.UI.Xaml
 			}
 		}
 
-		internal void SetMaxLines(object localValue)
+		internal void SetMaxLines(object maxLines, object textWrapping)
 		{
-			if (localValue == DependencyProperty.UnsetValue)
+			var isWrapping = textWrapping is TextWrapping.Wrap or TextWrapping.WrapWholeWords;
+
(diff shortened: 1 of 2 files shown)


jeromelaban:
@MartinZikmund let's validate the screenshots for this one, it's very easy to break with the native renderer.

src/Uno.UI/UI/Xaml/UIElement.TextHelper.wasm.cs
```diff
@@ -134,15 +134,20 @@ internal void SetFontSize(object localValue)
...
+			// webkit properties, as we will have a single line anyway, and these properties would
+			// otherwise interfere with text-overflow ellipsis functionality.
+			if (maxLines == DependencyProperty.UnsetValue || (!isWrapping && maxLines is 1))
 			{
 				this.ResetStyle("display", "-webkit-line-clamp", "webkit-box-orient");
 			}
 			else
 			{
-				var value = (int)localValue;
+				var value = (int)maxLines;
```
jeromelaban, line 150: Can we add a runtime test that validates the final `p` size?

MartinZikmund:
Unfortunately, this approach does not work too reliably, as in case the `Text` itself contains multiple lines, `MaxLines` should restrict it. This change would not allow for such a behavior, unless we explicitly checked for new line characters in the text as well, which however, makes the logic a bit fragile...

PR #20648, opened 2025-05-30
fix: AppBarButton not responding to clicks on mobile

Kunal22shah:
GitHub Issue (If applicable): closes https://github.com/unoplatform/ziidms-private/issues/72

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_CommandBar.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_CommandBar.cs
@@ -178,6 +178,60 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 			SUT.IsOpen = true;
 			await WindowHelper.WaitForIdle();
 		}
+
+		[TestMethod]
+#if !HAS_INPUT_INJECTOR
+		[Ignore("InputInjector is not supported on this platform.")]
+#endif
+		public async Task When_LoadUnload_CommandBar_IsOpen_Resets_OnReload()
+		{
+			var SUT = new CommandBar
+			{
+				SecondaryCommands =
+				{
+					new AppBarButton { Label = "SecondaryCommand", Name = "SecondaryButton" }
+				}
+			};
+
+			await UITestHelper.Load(SUT);
+			await WindowHelper.WaitForIdle();
+			await Task.Delay(1000);
+
+			var moreBtn = (Button)SUT.FindName("MoreButton");
+			Assert.IsNotNull(moreBtn);
+			var injector = InputInjector.TryCreate() ?? throw new InvalidOperationException("Failed to init InputInjector");
+			using var finger = injector.GetFinger();
+			Point GetCenter(Rect rect) => new Point(rect.Left + rect.Width / 2, rect.Top + rect.Height / 2);
+			finger.Press(GetCenter(moreBtn.GetAbsoluteBounds()));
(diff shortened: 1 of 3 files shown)


MartinZikmund:
We will also need a test to cover this change

src/Uno.UI/UI/Xaml/Controls/AppBar/AppBar.Partial.cs
```diff
@@ -154,6 +154,29 @@ private protected override void OnLoaded()
...
+				expandBtn.Click += OnExpandButtonClick;
+				m_expandButtonClickEventHandler.Disposable =
+					Disposable.Create(() => expandBtn.Click -= OnExpandButtonClick);
+			}
+			if (GetTemplateChild("MoreButton") is ButtonBase moreButton)
+			{
+				moreButton.Click += OnExpandButtonClick;
+				m_expandButtonClickEventHandler.Disposable =
+					Disposable.Create(() => moreButton.Click -= OnExpandButtonClick);
+			}
```
jeromelaban: Let's add a runtime test for this

kazo0: This is quite similar to the logic inside of `OnApplyTemplate`. We could extract this to re-use the code. Unless we don't want to modify the `OnApplyTemplate` as it is ported code? @MartinZikmund?

MartinZikmund: Although it would touch ported code, in this case it would probably make sense. As long as the Uno specific code is marked clearly, it should be ok

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_CommandBar.cs
```diff
@@ -178,6 +178,41 @@ public async Task When_IsOpen_True_LayoutCycle()
...
+
+			WindowHelper.WindowContent = SUT;
+			await WindowHelper.WaitForIdle();
+
+			var injector = InputInjector.TryCreate() ?? throw new InvalidOperationException("Failed to init InputInjector");
+			using var finger = injector.GetFinger();
+			var secBtn = (AppBarButton)SUT.FindName("SecondaryButton");
+			var secBounds = secBtn.GetAbsoluteBounds();
+			finger.Press(secBounds.Left + secBounds.Width / 2, secBounds.Bottom + 20);
+			finger.Release();
```
kazo0: This seems to be testing a click on a button that is in the overflow menu? Shouldn't be we pressing on the expand/more button and validating that the IsOpen properly gets set to True?

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_CommandBar.cs
```diff
@@ -178,6 +178,41 @@ public async Task When_IsOpen_True_LayoutCycle()
...
+				SecondaryCommands =
+				{
+					new AppBarButton { Label = "SecondaryCommand", Name = "SecondaryButton" }
+				}
+			};
+
+			await UITestHelper.Load(SUT);
+			await WindowHelper.WaitForIdle();
+
+			SUT.IsOpen = true;
```
kazo0: We should be opening the AppBar by performing a click here instead of forcing to IsOpen=true. So we can validate that the click works before and after reloading

kazo0 (approved):

PR #20698, opened 2025-06-04
build: fix xaml merging race on android/skia

ramezgerges:
**GitHub Issue:** closes #

When building for android/skia, the app crashes on startup with a resource resolution error. It turns out that mergedstyles are sometimes completely blank, which explains the missing resources. This is because both Uno.UI.Reference and Uno.UI.Skia are running the XamlMerge target so on my machine it seems like the one run from the reference project always wins.
## PR Checklist ✅

### DIFF
--- a/src/Uno.UI/Uno.UI.Reference.csproj
+++ b/src/Uno.UI/Uno.UI.Reference.csproj
@@ -140,6 +140,5 @@
 	</ItemGroup>
 
 	<Import Project="MixinGeneration.targets" />
-	<Import Project="XamlMerge.targets" />
 
 </Project>


MartinZikmund (approved):

MartinZikmund:
This is the right fix, as the `xaml` files are not automatically included in reference build (`EnableAutomaticXamlPageInclusion` is `false`)

src/Uno.UI/Uno.UI.Reference.csproj
```diff
@@ -140,6 +140,5 @@
 	</ItemGroup>
 
 	<Import Project="MixinGeneration.targets" />
-	<Import Project="XamlMerge.targets" />
```
jeromelaban, line 143: > This is the right fix, as the `xaml` files are not automatically included in reference build (`EnableAutomaticXamlPageInclusion` is `false`)

Let's make sure it won't cause issues on platforms that rely on the presence of the `GlobalStaticResources` on the reference assemblies (it may only appear when running from the uno.sdk, not inside the samples app).

MartinZikmund: Ah, that's a tricky one  🤔 . I guess in that case we could provide just a blank "GlobalStaticResources" class a stub for `Reference` specifically

ramezgerges: I've tested Gallery with the CI artifacts and it seems fine without a GlobalStaticResources for the reference project.

ajpinedam (approved):

PR #20713, opened 2025-06-05
fix(droid): webview insets

### LINKED ISSUE #19858: Android WebView does not respect AdjustResize with Uno.Sdk 5.6.51
### Current behavior

When using "Uno.Sdk": "5.6.51" on Android, setting `SoftInputMode = AdjustResize` has no effect when displaying a WebView.
This causes the on-screen keyboard to cover focused input fields, especially when the HTML content contains forms or interactive inputs.
In previous versions like 5.6.22, the keyboard triggered proper layout resizing and scrolling as expected.

https://github.com/user-attachments/assets/9a039f87-908e-469c-81e0-9ac935c72e69

### Expected behavior

`SoftInputMode = AdjustResize` should shrink the WebView’s visible area and allow input elements (like <input>, <textarea>) to stay visible when the keyboard is shown — just like it did in version 5.6.22.

https://github.com/yeahg-dev/UnoAndroidWebViewExample

### Workaround

_No response_

### Works on UWP/WinUI

None

### Environment

_No response_

### NuGet package version(s)

_No response_

### Affected platforms

Android

### IDE

Visual Studio Code

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

ajpinedam:
**GitHub Issue:** closes #19858 

## PR Type:

- 🐞 Bugfix

Android when using WebViews the Insets are not being applied correctly causing the inputs to be covered by the keyboard

## PR Checklist ✅

### DIFF
--- a/doc/articles/common-issues-android.md
+++ b/doc/articles/common-issues-android.md
@@ -4,6 +4,30 @@ uid: Uno.UI.CommonIssues.Android
 
 # Issues related to Android projects
 
+## Insets not working when using WebView
+
+There is a known issue on Android with the [Edge-to-Edge](https://developer.android.com/develop/ui/views/layout/edge-to-edge) feature, where the bottom Insets do not function correctly when using a WebView. This flaw prevents the inputs from being shifted upward, which results in the keyboard covering them.
+
+> [!IMPORTANT]
+> Edge-to-Edge is enforced on devices running Android 15 with a Target SDK of 35 or higher.
+
+A workaround for this issue is to navigate to the `MainActivity.Android.cs` file in your Uno Platform application, override the `OnCreate` method, and call `WindowCompat.SetDecorFitsSystemWindows()`, setting the second parameter, decorFitsSystemWindows, to true.
+
+Your code should be structured as follows:
+
+```csharp
+public class MainActivity : Microsoft.UI.Xaml.ApplicationActivity
+{
+  protected override void OnCreate(Bundle bundle)
+  {
+      base.OnCreate(bundle);
+      WindowCompat.SetDecorFitsSystemWindows(Window, true);
+  }
+}
+```
+
+Learn more about this issue [on Google's issues tracker](https://issuetracker.google.com/issues/311256305?pli=1) website.
+
 ## ADB0020 - The package does not support the CPU architecture of this device
 
 This error may occur when deploying an application to a physical device with ARM architecture. To resolve this issue, you will need to add the following to your csproj anywhere inside the `<PropertyGroup>` tag:
--- a/doc/articles/common-issues-android.md
+++ b/doc/articles/common-issues-android.md
@@ -28,6 +28,9 @@ public class MainActivity : Microsoft.UI.Xaml.ApplicationActivity
 
 Learn more about this issue [on Google's issues tracker](https://issuetracker.google.com/issues/311256305?pli=1) website.
 
+> [!IMPORTANT]
+> Using the workaround mentioned above will deactivate some benefits of the Edge-To-Edge feature. For instance, your top navigation bar will no longer be "invisible," which removes the immersive appearance associated with edge-to-edge design.
+
 ## ADB0020 - The package does not support the CPU architecture of this device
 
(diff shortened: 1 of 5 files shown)


src/Uno.UI/UI/Xaml/ApplicationActivity.Android.cs
```diff
@@ -267,6 +268,8 @@ protected override void OnCreate(Bundle bundle)
 				EdgeToEdge.Enable(this);
 			}
 
+			WindowCompat.SetDecorFitsSystemWindows(Window, FeatureConfiguration.AndroidSettings.IsEdgeToEdgeEnabled);
```
MartinZikmund: This should not be needed  🤔 - https://developer.android.com/develop/ui/views/layout/edge-to-edge-manually#lay-out-in-full-screen Edge to Edge mode sets this to `false`, so this would kind of "undo" a part of the edge to edge functionality.

MartinZikmund (changes requested):

src/Uno.UI/UI/Xaml/ApplicationActivity.Android.cs
MartinZikmund: Only thing that comes to mind end users could do this manually, but I think it would break other scenarios to do it directly within Uno

ajpinedam: Not entirely, the Edge-to-Edge is still there, even with this removed (will upload a video in a moment). 

The reason is that there is a on-going problem/bug on Android that is causing this error, specifically when using WebView. I tried removing this only when the WebView is the content of the page, but it's hard and I could not find a way to do it.

Link to the android issue: https://issuetracker.google.com/issues/311256305?pli=1

ajpinedam: Looking closely, the edge-to-edge effect is not fully achieved when using the code above.

|  without code (e2e fully enabled) | with code above |
| :--: | :---: |
|uno_android_webview_inset_edge| uno_android_webview_inset|

Should we implement this fix behind feature flags and document that it's an issue coming from Android?

ajpinedam: Should we implement this fix behind feature flags and document that it's an issue coming from Android?

MartinZikmund: I think it could be just documented as a workaround in `WebView` docs - as people can manually call this API in their `MainActivity` if they choose to - it is just a one liner, so I think a feature flag is not needed for this one. I'm still surprised it has an impact here though 🤔 ...

ajpinedam: @MartinZikmund  removed the change and updated the PR with the docs and the workaround

doc/articles/common-issues-android.md
```diff
@@ -4,6 +4,30 @@ uid: Uno.UI.CommonIssues.Android
...
+
+Your code should be structured as follows:
+
+```csharp
+public class MainActivity : Microsoft.UI.Xaml.ApplicationActivity
+{
+  protected override void OnCreate(Bundle bundle)
+  {
+      base.OnCreate(bundle);
+      WindowCompat.SetDecorFitsSystemWindows(Window, true);
```
jeromelaban, line 24: Let's also explain the drawbacks of setting this specific property, if any (otherwise, we'd have set it by default inside uno?)

ajpinedam: @jeromelaban Re "otherwise, we'd have set it by default inside uno"  That was the original fix but that would revert part of the Edge-to-Edge feature.  here's a thread with a discussion with Martin about it: https://github.com/unoplatform/uno/pull/20713#discussion_r2128421177

jeromelaban: Agreed, but let's explain it in the docs anyways :)

ajpinedam: Updated the docs indicating the mentioned above

jeromelaban (approved):

MartinZikmund (approved):

ajpinedam:
https://github.com/Mergifyio backport release/stable/6.0

Issue #17300, opened 2024-06-25
Expand `IsPropMethodCall` usage: bit-pack UIElement booleans, calculated BitmapImage/Control properties, and precedence parity


Youssef1313:
The core `IsPropMethodCall` infrastructure landed in #22747 (merged 2026-04-17), along with three initial migrations (`UIElement.IsHitTestVisible`, `TextBox.CanUndo`, `TextBox.CanRedo`) and runtime tests in `Given_DependencyProperty_PropMethodCall`. This issue tracks the remaining work that the original scope and the discussion in #17285 called for.

#### 1. Bit-pack `UIElement` boolean DPs

`UIElement.Properties.cs` currently carries an explicit TODO:

```csharp
// TODO: Add more PropMethodCall and combine them in BitVector32
private bool _isHitTestVisible = true;
```

Only `IsHitTestVisible` has been migrated, and it still occupies a full `bool` field. The goal is a single `BitVector32` (or equivalent bitfield) on `UIElement` shared by all boolean DPs, so each costs 1 bit instead of a full slot.

Candidates to migrate alongside `IsHitTestVisible`:

- `IsTapEnabled`
- `IsDoubleTapEnabled`
- `IsRightTapEnabled`
- `IsHoldingEnabled`
- `AllowDrop`
- `CanDrag`
- `IsAccessKeyScope`
- `ExitDisplayModeOnAccessKeyInvoked`

#### 2. Calculated properties (the #17285 case)

`BitmapImage` / `BitmapSource` / `RenderTargetBitmap` `PixelWidth` and `PixelHeight` are still plain `SetValue`-driven DPs. They should be migrated to `PropMethodCall` so the value can be derived (image metadata when available, otherwise the retained natural size) the way WinUI does:

```cpp
if (IsMetadataAvailable())
{
    return m_imageMetadataView->GetImageMetadata()->width;
}
else
{
    auto width = GetWidth();
    if (width != 0) { return width; }
    else { return m_retainedNaturalWidth; }
}
```

#### 3. `Control.IsEnabled` through `Enter`

Called out in the #17285 discussion: `IsEnabled` needs `PropMethodCall` semantics so that visual-tree inheritance of the disabled state via `Enter` works correctly. Not addressed by #22747.

#### 4. Precedence parity: \"local first, then Style → Style skipped\"

[...]

PR #20724, opened 2025-06-05
perf(skia): Cache UseLayoutRounding

ebariche:
## PR Type:

- ⚡Perf

## PR Checklist ✅

### DIFF
--- a/src/Uno.UI/UI/Xaml/UIElement.skia.cs
+++ b/src/Uno.UI/UI/Xaml/UIElement.skia.cs
@@ -52,16 +52,14 @@ namespace Microsoft.UI.Xaml
 
 		public bool UseLayoutRounding
 		{
-			get => (bool)this.GetValue(UseLayoutRoundingProperty);
-			set => this.SetValue(UseLayoutRoundingProperty, value);
+			get => GetUseLayoutRoundingValue();
+			set => SetUseLayoutRoundingValue(value);
 		}
 
-		public static DependencyProperty UseLayoutRoundingProperty { get; } =
-			DependencyProperty.Register(
-				nameof(UseLayoutRounding),
-				typeof(bool),
-				typeof(UIElement),
-				new FrameworkPropertyMetadata(true, propertyChangedCallback: (o, _) => (o as IBorderInfoProvider)?.UpdateBorderThickness()));
+		[GeneratedDependencyProperty(DefaultValue = true, ChangedCallbackName = nameof(OnUseLayoutRoundingChanged))]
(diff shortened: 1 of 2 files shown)


MartinZikmund (approved):

src/Uno.UI/UI/Xaml/UIElement.skia.cs
```diff
@@ -52,16 +52,14 @@ public UIElement()
 
 		public bool UseLayoutRounding
 		{
-			get => (bool)this.GetValue(UseLayoutRoundingProperty);
```
Youssef1313, line 55: I'm surprised that reading the property is expensive and curious to know what part of the DP system is showing up in traces.

Youssef1313: NOTE: Given that the property is `MetaDataPropertyInfoFlags::IsPropMethodCall` in WinUI, this change is fine and reasonable until Uno implements `IsPropMethodCall` (tracked by https://github.com/unoplatform/uno/issues/17300). After `IsPropMethodCall` is implemented, you can try using it and see if the perf is still a concern.

It would still be good to understand exactly where the perf hit is, and what scenario is being profiled. So, it's easier for future devs who may want to switch to `IsPropMethodCall` to have a reasonable idea on how to ensure the switch doesn't regress perf.

jeromelaban: Any kind of call to the DP system is more expensive than reading a local cached value being the property getter, so it always makes sense to add this cache for high-traffic properties. Even with `IsPropMethodCall`, it will still go through the weeds of DPs and will be expensive on recurrent reads.

Youssef1313: > Any kind of call to the DP system is more expensive than reading a local cached value being the property getter

Sure, that's definitely true. But what's the scenario being profiled? How large the improvement is? Which path of DP is showing in the traces? There is a path when the property value was never set before, which doesn't even create DependencyPropertyDetails, and there is another path when there is a DependencyPropertyDetails. Knowing which path is showing in traces, and some numbers from before/after traces would be nice IMO.

jeromelaban: > some numbers from before/after traces would be nice IMO

That we agree on :)

ebariche: This DP is read on Measure/Arrange/Clip for Borders/Grids/TextBlocks/ScrollViewer/...

Numbers are measured on a synthetic bench (Dopes Change) for 11s ~, real traffic would be less but it still piles up.

Before:

image

After:

image

PR #20730, opened 2025-06-06
docs: remove initializelogging calls from suggested migration code

kazo0:
We should not include the `App.InitializeLogging()` call in the migration code for Skia rendered app. It is misleading as this method may not always exist in the App.xaml.cs depending on the template used

### DIFF
--- a/doc/articles/migrating-to-uno-6.md
+++ b/doc/articles/migrating-to-uno-6.md
@@ -94,6 +94,8 @@ To upgrade:
     ```csharp
     public static int Main(string[] args)
     {
+        // code omitted for brevity
+
         Microsoft.UI.Xaml.Application.Start(_ => _app = new App());
 
         return 0;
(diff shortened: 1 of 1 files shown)


doc/articles/migrating-to-uno-6.md
```diff
@@ -94,8 +94,6 @@ To upgrade:
     ```csharp
     public static int Main(string[] args)
     {
-        App.InitializeLogging();
```
jeromelaban, line 97: This is required for the blank template. We should make it so that adding it is conditional to another call being present somewhere else or at least explain how to check if it's needed here.

kazo0: It's not relevant to the changes we are specifying here that are required when migrating an existing app to Uno 6 Skia

We could just replace with `// Code omitted for brevity`

ajpinedam (approved):
Good!

kazo0:
@Mergifyio backport release/stable/6.0

PR #20808, opened 2025-06-17
fix(skia): make sure the UI tree is clean before rendering

ramezgerges:
**GitHub Issue:** closes https://github.com/unoplatform/uno.hotdesign/issues/4327

## PR Checklist ✅

### DIFF
--- a/src/Uno.UI.Runtime.Skia.Wpf/Rendering/SoftwareWpfRenderer.cs
+++ b/src/Uno.UI.Runtime.Skia.Wpf/Rendering/SoftwareWpfRenderer.cs
@@ -50,9 +50,10 @@ internal class SoftwareWpfRenderer : IWpfRenderer
 
 		using var _ = _fpsHelper.BeginFrame();
 
-		while (_host.RootElement is { } rootElement && (rootElement.IsArrangeDirtyOrArrangeDirtyPath || rootElement.IsMeasureDirtyOrMeasureDirtyPath))
+		if (_host.RootElement is { } rootElement && (rootElement.IsArrangeDirtyOrArrangeDirtyPath || rootElement.IsMeasureDirtyOrMeasureDirtyPath))
 		{
-			rootElement.UpdateLayout();
+			_host.InvalidateRender();
+			return;
 		}
 
 		int width, height;
--- a/src/Uno.UI.Runtime.Skia.Wpf/Rendering/SoftwareWpfRenderer.cs
+++ b/src/Uno.UI.Runtime.Skia.Wpf/Rendering/SoftwareWpfRenderer.cs
@@ -50,6 +50,11 @@ internal class SoftwareWpfRenderer : IWpfRenderer
 
 		using var _ = _fpsHelper.BeginFrame();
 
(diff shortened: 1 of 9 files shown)


Xiaoy312 (approved):

MartinZikmund (approved):

src/Uno.UI.Runtime.Skia.Wpf/Rendering/SoftwareWpfRenderer.cs
```diff
@@ -50,6 +50,11 @@ public void Render(DrawingContext drawingContext)
 
 		using var _ = _fpsHelper.BeginFrame();
 
+		while (_host.RootElement is { } rootElement && (rootElement.IsArrangeDirtyOrArrangeDirtyPath || rootElement.IsMeasureDirtyOrMeasureDirtyPath))
+		{
+			rootElement.UpdateLayout();
+		}
```
jeromelaban: This really feels like it could be placed in a central location. Could `RenderRootVisualAndReturnPath` be a candidate?

src/Uno.UI.Runtime.Skia.Wpf/Rendering/SoftwareWpfRenderer.cs
```diff
@@ -50,6 +50,11 @@ public void Render(DrawingContext drawingContext)
 
 		using var _ = _fpsHelper.BeginFrame();
 
+		while (_host.RootElement is { } rootElement && (rootElement.IsArrangeDirtyOrArrangeDirtyPath || rootElement.IsMeasureDirtyOrMeasureDirtyPath))
+		{
+			rootElement.UpdateLayout();
```
jeromelaban: This also feels like what leads to the render being called before any layout is done could be another bug. Aren't we supposed to invalidate the layout on invalidatearrange/invalidatemeasure?

src/Uno.UI.Runtime.Skia.Wpf/Rendering/SoftwareWpfRenderer.cs
ramezgerges: It really does, but all the composition stuff have no visibility into Uno.UI so we're forced to do it this way.

src/Uno.UI.Runtime.Skia.Wpf/Rendering/SoftwareWpfRenderer.cs
ramezgerges: I think you meant "invalidate the render on invalidatearrange/invalidatemeasure"? No, that shouldn't be the case. Layouting should change properties on the Visual associated with the UIElement and _that_ should invalidate the visual. In the case of TextBox/TextBlock, measuring (which called InlinesCollection.Measure) does not invalidate the render. Now, the obvious fix would be to add an invalidation line somewhere there but there is still a big problem. What's happening right now means that we're rendering partial (i.e. not fully laid out) frames. I think we should actually have both fixes.

src/Uno.UI.Runtime.Skia.Wpf/Rendering/SoftwareWpfRenderer.cs
jeromelaban: Having a separate method that does it would have been best.

PR #20833, opened 2025-06-19
fix: Fail more explicitly when `BrowserHtmlElement` APIs are used with an invalid id

jeromelaban:
**GitHub Issue:** closes https://github.com/unoplatform/kahua-private/issues/300

## PR Type:

- 🐞 Bugfix

Fail with a verbose message when the ID is incorrect.

## PR Checklist ✅

### DIFF
--- a/src/Uno.UI.Runtime.Skia.WebAssembly.Browser/ts/Runtime/BrowserNativeElementHostingExtension.ts
+++ b/src/Uno.UI.Runtime.Skia.WebAssembly.Browser/ts/Runtime/BrowserNativeElementHostingExtension.ts
@@ -82,19 +82,19 @@ namespace Uno.UI.NativeElementHosting {
 		}
 
 		public static attachNativeElement(content: string) {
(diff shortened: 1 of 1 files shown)


ramezgerges (approved):

ramezgerges:
@jeromelaban I'm wondering if there's a way to keep a reference to the html element even if the id changes? At the time of creating the feature, I didn't know we can send JSObject handles across FFI boundaries. Maybe we can use JSObject instead of element IDs?

jeromelaban:
> @jeromelaban I'm wondering if there's a way to keep a reference to the html element even if the id changes? At the time of creating the feature, I didn't know we can send JSObject handles across FFI boundaries. Maybe we can use JSObject instead of element IDs?

We probably could change it to avoid this lookup, indeed. It won't improve performance though, since the same kind of lookup is done, but only allow for the ID to be set dynamically. We're allowing changing the ID through the `CreateElement` API, so we'll revisit if it's becoming a problem.

PR #20841, opened 2025-06-20
fix: Handled virtual key was not correct

MartinZikmund:
**GitHub Issue:** closes https://github.com/unoplatform/add-private/issues/43

- `Down` key is always handled by `ScrollViewer`, even if scroll did not actually happen
- We always scroll and prevent focus if scrolling is possible in given direction

- `Down` key handling corrected
- We only scroll when the next focused item would be out of the viewport

## PR Checklist ✅

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.focus.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.focus.cs
@@ -1,0 +1,168 @@
+﻿#nullable enable
+
+using Microsoft.UI.Xaml.Input;
+using Windows.Foundation;
+using static Microsoft.UI.Xaml.Controls._Tracing;
+
+namespace Microsoft.UI.Xaml.Controls;
+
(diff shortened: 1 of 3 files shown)


src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.focus.cs
```diff
@@ -0,0 +1,168 @@
+﻿#nullable enable
```
ramezgerges, line 1: I'm assuming this is from WinUI?

ramezgerges (approved):

src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.focus.cs
MartinZikmund: Yes, this is ported code, but from the `ScrollView` control. The `ScrollViewer` code is very different from ours and it would not be possible to align, at least for now...

PR #21041, opened 2025-07-14
fix: ios statusbar when background is set before all windows are loaded

### LINKED ISSUE #1416: Adjust StatusBar Extensions to utilize the newly implemented StatusBar APIs directly in Uno
Once https://github.com/unoplatform/uno/pull/20687 merges, we can remove the platform specific code inside of the Status Bar Extensions:

https://github.com/unoplatform/uno.toolkit.ui/blob/main/src/Uno.Toolkit.UI/Behaviors/StatusBar.Android.cs
https://github.com/unoplatform/uno.toolkit.ui/blob/main/src/Uno.Toolkit.UI/Behaviors/StatusBar.iOS.cs

And we should be able to directly set the StatusBar.ForegroundColor and BackgroundColor without needing the Android/iOS specifics (allowing this to then work on Skia)

Outstanding question on this. Should we be bumping the dependency of Uno.WinUI in Toolkit for this? Or simply remove the platform specifics for the StatusBar extensions and as long as the client app is using a latest version of Uno with the StatusBar APIs implemented, this should would without bumping Uno.WinUI inside of Toolkit??

rajamatt:
**GitHub Issue:** closes https://github.com/unoplatform/uno.toolkit.ui/issues/1416

## PR Type:

When launching an app with StatusBar Extensions from Toolkit, the initial window would have the correct `utu:StatusBar.Background` color. However, the main window would not have the color since it is a new [`WindowsAndStatusBarFrame`](https://github.com/unoplatform/uno/blob/df246eb5d53680d043ebeb2757e5131b915d60fc/src/Uno.UWP/UI/ViewManagement/StatusBar/StatusBar.iOS.cs#L80).

We listen for when the app is active and can receive focus from the user (meaning the windows have loaded) and set the StatusBar background when that happens. 

## PR Checklist ✅

### DIFF
--- a/src/Uno.UWP/UI/ViewManagement/StatusBar/StatusBar.iOS.cs
+++ b/src/Uno.UWP/UI/ViewManagement/StatusBar/StatusBar.iOS.cs
@@ -28,7 +28,7 @@ namespace Windows.UI.ViewManagement
 				UIApplication.DidBecomeActiveNotification,
 				_ => SetStatusBarBackgroundColor(_backgroundColor)
 			);
-			
+
 			NSNotificationCenter.DefaultCenter.AddObserver(
 				UIApplication.DidChangeStatusBarOrientationNotification,
 				_ => SetStatusBarBackgroundColor(_backgroundColor)
--- a/src/Uno.UWP/UI/ViewManagement/StatusBar/StatusBar.iOS.cs
+++ b/src/Uno.UWP/UI/ViewManagement/StatusBar/StatusBar.iOS.cs
@@ -25,6 +25,11 @@ namespace Windows.UI.ViewManagement
 		partial void InitializePartial()
 		{
 			NSNotificationCenter.DefaultCenter.AddObserver(
+				UIApplication.DidBecomeActiveNotification,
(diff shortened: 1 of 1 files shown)


src/Uno.UWP/UI/ViewManagement/StatusBar/StatusBar.iOS.cs
```diff
@@ -24,6 +24,11 @@ public sealed partial class StatusBar
 	{
 		partial void InitializePartial()
 		{
+			NSNotificationCenter.DefaultCenter.AddObserver(
```
ajpinedam, line 27: The PR title says "ios skia," but if I'm not mistaken, this change (due to the file) only affects the Native iOS implementation. 

@MartinZikmund  could you please confirm?

jeromelaban: That's true, it's indeed iOS Skia, but this layer (uno.winrt) is platform specific, iOS files really are running on iOS.

jeromelaban (approved):

PR #21111, opened 2025-07-22
docs: Add server project basic docs

### LINKED ISSUE #21090: [WASM][Server] WebAssembly should support https by default instead of build fail
### Current behavior 🐛

changing the applicationUrl value to use https in launchsettings (same ports) and using the provided server project doing the same there, the build will instantly fail with the message that it failed to find the server:

Image

### Expected behavior 🎯

as dotnet runtime and most api we might want to send requests to, will require the usage of antiforgery and cors, it should be the default or even possible, to use https instead of http

create a new uno app including wasm target, server project and for example custom auth.
start it and see the notification at the top bar.

go to the uno app launchsettings and set the http to https, same for the server project

will not longer build and fail with the shown message.

### Workaround 🛠️

None! does not build when setting the url to https even while our server defines `app.UseHttpsRedirection()`

### Renderer 🎨

- [x] Skia
- [ ] Native

### Affected platforms 📱💻🖥️

_No response_

6.1.23

### IDE version 🧑‍💻

17.14.9

If our apps are only able to use http, this will make the user get notifyed that the page he is surfing on, does not support https and he shall not type in any passwords etc. now, if our app would then be configured by template already to use authentication with e.g. customauth, that very first view he will get presented, IS a Login UI. See the Problem?
By that our produced app is marked as NOT trustable.

in best case those features should get build in, so for example, this is what I would suggest to be usable gathered from multiple resources you can find linked there:

```csharp
public static class SecurityServiceExtensions
{
    public static IServiceCollection AddSecurityServices(this IServiceCollection services, IConfiguration configuration)
    {

        // Etsy Authentication requires Cross-site request forgery (CSRF) protection,
        // <see href="https://developers.etsy.com/documentation/essentials/authentication#requesting-an-oauth-token"/>
        // so we need to add the Antiforgery service.
        services.AddAntiforgery(setupAction =>
        {
            setupAction.Cookie.SecurePolicy = CookieSecurePolicy.Always; // Always use HTTPS
        });

        // Add CORS service policy as needed for wasm Multi-threaded applications
        // <see href="https://github.com/dotnet/runtime/blob/main/src/mono/wasm/features.md#multi-threading"/>
        // Configuration reference:
        // <see href="https://substack.com/home/post/p-164745710"/>
        services.AddCors(options =>
        {
            var AllowedHosts = configuration["AllowedHosts"]?
                                              .ToLowerInvariant().Split(';', StringSplitOptions.RemoveEmptyEntries)
                                              ?? throw new ArgumentNullException("AllowedHosts"); ;

            string[] Methods = [
                HttpMethods.Get,
                HttpMethods.Post,
                HttpMethods.Put,
                HttpMethods.Delete,
                HttpMethods.Patch
            ];
            string[] Headers = [
                HeaderNames.ContentType,
                HeaderNames.Cookie, // Adding Cookie header to allow cookies in CORS requests, as I want to use Cookie Authentication
                HeaderNames.Authorization,
                // Adding headers referring to MDN Docs
                // <see href="https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CORS" />
                // to avoid problems with browsers like Safari or Nightly
                HeaderNames.ContentLanguage,
                HeaderNames.AcceptLanguage,
                HeaderNames.AcceptEncoding,
                // Adding Preflight request headers from samples
                // <see href="https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#preflighted_requests"/>
                HeaderNames.Origin,
                HeaderNames.UserAgent,
                HeaderNames.AccessControlRequestHeaders,
                HeaderNames.AccessControlRequestMethod,
                HeaderNames.Host
            ];


            options.AddPolicy("MyCorsPolicy", policy =>
            {
                policy.WithOrigins()
                .WithMethods(Methods)
                .WithHeaders(Headers)
                .DisallowCredentials() // Credentials are not required for cors requests as we use cookies
                .SetPreflightMaxAge(TimeSpan.FromMinutes(10));
            });

        });
        return services;
    }
    public static IApplicationBuilder UseSecurityServices(this IApplicationBuilder app)
    {
        app.UseCookiePolicy();
        app.UseAntiforgery();
        app.UseCors("MyCorsPolicy");
        return app;
    }
}
```

jeromelaban:
**GitHub Issue:** closes https://github.com/unoplatform/uno/issues/21090


DevTKSS:
if including the WebAssembly Target Platform is important, maybe we should add this information?
and would you have something against some visual content? :)
image
maybe showing where to find it in the templates and... dont know, is there something related to the Server to know about PWA? otherwhise, maybe a link to the PWA docs (I think I seens some in the past 🤔 ) and that this is a consideration maybe? dont know if it is. Or do we have known issues so we should consider not to combine those? I had refit http files in my new project while choosen kiota in the wizard, thats why I am asking. Not that this intermixing would then confuse the user 👍 here, some wizard image for this option maybe
image

jeromelaban:
> maybe showing where to find it in the templates and... dont know, is there something related to the Server to know about PWA? otherwhise, maybe a link to the PWA docs (I think I seens some in the past 🤔 ) and that this is a consideration maybe? dont know if it is. Or do we have known issues so we should consider not to combine those? I had refit http files in my new project while choosen kiota in the wizard, thats why I am asking. Not that this intermixing would then confuse the user 👍 here, some wizard image for this option maybe

Thanks. There's no need to explain how to create a project, we're detailing this in the getting started documentation.

DevTKSS:
@jeromelaban 👍 seperate PR then I think :)
hm but for the intro of this page this is should maybe tell about if this is assuming we do include specific targets or if this is independant of them and as the Wizard does file the Server Option below "Features" the intro sentence maybe should be more like... 🤔 
"
As part of the uno platform templates, you can opt in to include a ASP.NET Core server project to your solution to allow the hosting of both an API server as well as a WebAssembly app.
"
it reads a bit... in german you would say "holprig" 😅

DevTKSS:
will try if this works 👍

DevTKSS:
@jeromelaban sorry, no does not work out with the https 👎

DevTKSS:
attached repro etc to the linked issue

DevTKSS:
@jeromelaban missing this asp net core https trust certificate setting here to be linked or introduced to execute:
https://learn.microsoft.com/en-us/aspnet/core/security/enforcing-ssl?view=aspnetcore-9.0&tabs=visual-studio%2Clinux-sles#trust-the-aspnet-core-https-development-certificate
not sure where you would add it (server or app?) but I think this might be important to do?

jeromelaban:
> @jeromelaban missing this asp net core https trust certificate setting here to be linked or introduced to execute:
https://learn.microsoft.com/en-us/aspnet/core/security/enforcing-ssl?view=aspnetcore-9.0&tabs=visual-studio%2Clinux-sles#trust-the-aspnet-core-https-development-certificate
not sure where you would add it (server or app?) but I think this might be important to do?

It's generally done by VS automatically, but for other platforms it may be needed, thanks.

DevTKSS:
@jeromelaban not sure if you did notice in the video, but if I did read your guide correctly, you do tell that the https additional url has to be put into our wasm client app? Maybe you did meant the server? because somehow, as soon as I did that, it had that socked error again 🤔

jeromelaban:
> @jeromelaban not sure if you did notice in the video, but if I did read your guide correctly, you do tell that the https additional url has to be put into our wasm client app? Maybe you did meant the server? because somehow, as soon as I did that, it had that socked error again 🤔

The section named `Enabling HTTPS for the app project` is about only the app project, not the server project. If you're running the server as startup project, https is already selected by default by the VS debugger. If the server project is running not for you in https, then swap the urls in `applicationUrls` in `launchSettings.json` to place https first. Then if you get the mixed content error, the rest of the doc of this PR is applicable again.

DevTKSS:
@jeromelaban I am trying to see, where I am making the mistake, because I would think  I did what it told and were curios about that addition on the app != server 🤷 and our templated app url has just 1 member and our server 2 🤔

MartinZikmund (approved):

PR #21145, opened 2025-07-24
fix: Basic split rendering on WASM

### LINKED ISSUE #21130: WASM app is blanking out when resizing
### Current behavior 🐛

When some Uno Skia WASM apps are resized, they are blanking out temporarily. Interestingly, it is not happening in all applications, e.g. it happens reliably in Samples app and Chefs, but not in Uno Gallery

https://github.com/user-attachments/assets/b44d5bac-102e-4926-860f-c8fe041cab52

### Expected behavior 🎯

Should never happen

_No response_

### Workaround 🛠️

_No response_

### Renderer 🎨

- [x] Skia
- [ ] Native

### Affected platforms 📱💻🖥️

WebAssembly

[6.2.0-dev.28](https://www.nuget.org/packages/Uno.Sdk/6.2.0-dev.28)

### IDE version 🧑‍💻

VS 2022

_No response_

MartinZikmund:
**GitHub Issue:** closes https://github.com/unoplatform/uno/issues/21130

## PR Checklist ✅

### DIFF
--- a/src/Uno.UI.Runtime.Skia.WebAssembly.Browser/Rendering/BrowserRenderer.cs
+++ b/src/Uno.UI.Runtime.Skia.WebAssembly.Browser/Rendering/BrowserRenderer.cs
@@ -8,6 +8,8 @@ using System.Diagnostics;
 using Uno.UI.Helpers;
 using Microsoft.UI.Xaml.Media;
 using Uno.UI.Dispatching;
+using Uno.UI.Xaml.Core;
+using System.Threading;
 
 namespace Uno.UI.Runtime.Skia;
 
@@ -33,6 +35,7 @@ internal partial class BrowserRenderer
 	private GRBackendRenderTarget? _renderTarget;
 	private SKSurface? _surface;
 	private SKCanvas? _canvas;
+	private SKPicture? _picture;
 
 	private SKSizeI? _lastSize;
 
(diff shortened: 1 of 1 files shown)


MartinZikmund:
@ramezgerges this seems to help, but neither seems _correct_ still... I think it does make sense we are rendering while there is another arrange already queued, that should not be a big problem - only issue would be if render would happen during an already running arrange, that could mess up things. We will need to look into this further, as it is really odd, but in the current state any "persistent" manipulation like dragging completely freezes up the UI until the pointer stops moving

MartinZikmund:
@mergifyio backport release/stable/6.1

src/Uno.UI.Runtime.Skia.WebAssembly.Browser/Rendering/BrowserRenderer.cs
```diff
@@ -60,18 +63,48 @@ private void Initialize()
...
 	}
 
-	internal void InvalidateRender() => NativeMethods.Invalidate(_nativeSwapChainPanel);
-
-	private void RenderFrame()
+	internal void InvalidateRender()
 	{
-		using var _ = _fpsHelper.BeginFrame();
-
 		if (_host.RootElement is { } rootElement && (rootElement.IsArrangeDirtyOrArrangeDirtyPath || rootElement.IsMeasureDirtyOrMeasureDirtyPath))
```
ramezgerges, line 68: Having the dirty check inside InvalidateRender and returning early without forcing a later invalidation can lead to the UI not updating.
Here's a scenario
1. 2 UIElements UI1 and UI2 are dirty
2. UI1 updated its layout and in the process called InvalidateRender, which did nothing because UI2 is still dirty.
3. UI2 goes through its layout cycle but nothing is updated and therefore no new InvalidateRender call happens.
4. Result: UI1 changed dimensions but the UI was never updated.

MartinZikmund:
Closing in favor of new split render pr

PR #21199, opened 2025-08-04
chore!: Remove some use of [Export]

jonpryor:
Context: https://github.com/unoplatform/uno-private/issues/1422
Context: https://github.com/dotnet/maui/commit/70fbb835d74b7dd6268bfa51ba8f450d0bea22dd

While exploring the use of NativeAOT on Android for .NET 10, a sticking point is that `[Java.Interop.ExportAttribute]` does not "play nice" with NativeAOT; if a class contains an `[Export]` attributed-method, and that class is encountered during runtime, then an exception is thrown:

	F DOTNET  : FATAL UNHANDLED EXCEPTION: System.InvalidOperationException: Methods such as __export__ are not implemented!
	F DOTNET  :    at Microsoft.Android.Runtime.ManagedTypeManager.RegisterNativeMembers(JniType, Type, ReadOnlySpan`1) + 0x386
	F DOTNET  :    at Java.Interop.ManagedPeer.RegisterNativeMembers(IntPtr jnienv, IntPtr klass, IntPtr n_nativeClass, IntPtr n_methods) + 0x195

This can be worked around for *instance* methods, but not for static methods.

Update `ApplicationActivity*.cs` and `NativeApplication.cs` to remove `[Export]`, and update any such attributed methods to throw a `NotSupportedException`.

Add a new `UnoWebViewHandlerJavascriptInterface.java` Java class, and update `UnoWebViewHandler.Android.cs` to implement `IUnoWebViewHandlerJavascriptInterface`, a'la dotnet/maui@70fbb835. This removes the use of `[Export]` from `UnoWebViewHandler.Android.cs`.

TODO: Review and update other uses of `[Export]`, such as in `UIElement.Android.cs`.

BREAKING CHANGE: the following methods now throw `NotSupportedException`:

  * `Microsoft.UI.Xaml.ApplicationActivity.GetTypeAssemblyFullName(string)` (Android renderer)
  * `Microsoft.UI.Xaml.ApplicationActivity.GetTypeAssemblyFullName(string)` (Android+Skia renderer)
  * `Microsoft.UI.Xaml.NativeApplication.GetTypeAssemblyFullName(string)`

**GitHub Issue:** closes #

## PR Checklist ✅

### DIFF
--- a/src/Uno.UI/UI/Xaml/NativeApplication.cs
+++ b/src/Uno.UI/UI/Xaml/NativeApplication.cs
@@ -152,9 +152,9 @@ namespace Microsoft.UI.Xaml
 		/// </summary>
 		/// <param name="type">A type full name</param>
 		/// <returns>The assembly that contains the specified type</returns>
-		[Export(nameof(GetTypeAssemblyFullName))]
 		[global::System.ComponentModel.EditorBrowsable(global::System.ComponentModel.EditorBrowsableState.Never)]
-		public static string GetTypeAssemblyFullName(string type) => Type.GetType(type)?.Assembly.FullName;
+		public static string GetTypeAssemblyFullName(string type) =>
+			throw new NotSupportedException("`static` methods with [Export] are not supported on NativeAOT.");
 
 		private class ActivityCallbacks : Java.Lang.Object, IActivityLifecycleCallbacks
 		{
--- a/src/Uno.UI/UI/Xaml/NativeApplication.cs
+++ b/src/Uno.UI/UI/Xaml/NativeApplication.cs
@@ -152,9 +152,15 @@ namespace Microsoft.UI.Xaml
 		/// </summary>
 		/// <param name="type">A type full name</param>
 		/// <returns>The assembly that contains the specified type</returns>
+#if NET10_0_OR_GREATER
 		[global::System.ComponentModel.EditorBrowsable(global::System.ComponentModel.EditorBrowsableState.Never)]
 		public static string GetTypeAssemblyFullName(string type) =>
 			throw new NotSupportedException("`static` methods with [Export] are not supported on NativeAOT.");
+#else   // !NET10_0_OR_GREATER
+		[Export(nameof(GetTypeAssemblyFullName))]
+		[global::System.ComponentModel.EditorBrowsable(global::System.ComponentModel.EditorBrowsableState.Never)]
+		public static string GetTypeAssemblyFullName(string type) => Type.GetType(type)?.Assembly.FullName;
+#endif  // !NET10_0_OR_GREATER
 
 		private class ActivityCallbacks : Java.Lang.Object, IActivityLifecycleCallbacks
 		{
--- a/src/Uno.UI.BindingHelper.Android/Uno/UI/UnoWebViewHandlerJavascriptInterface.java
+++ b/src/Uno.UI.BindingHelper.Android/Uno/UI/UnoWebViewHandlerJavascriptInterface.java
@@ -1,5 +1,5 @@
 package Uno.UI;
 
-public interface UnoWebViewHandlerJavascriptInterface {
-	void postMessage(String message);
+public abstract class UnoWebViewHandlerJavascriptInterface {
+	public abstract void postMessage(String message);
 }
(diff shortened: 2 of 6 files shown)


jonpryor:
Multiple implicit and explicit questions here:

1. *Will it build*? Does anything on CI call the `GetTypeAssemblyFullName()` methods which now throw `NotSupportedException`?
2. Is this change *acceptable*?
3. And/or should this change be *conditional* on .NET 10? i.e. .NET 9 builds work as before, while .NET 10 builds throw or just remove the method outright.

jonpryor:
Rebased atop acd096ea1c5d26efa6462ea2a24cc83ff289e7bb, which makes for a known comparison point: https://dev.azure.com/uno-platform/Uno%20Platform/_build/results?buildId=172962&view=results

acd096ea only shows test failures on **Tests - iOS Native** and **Publish - iOS Testflight**.  Nothing Android.

Compare to the current build, which has various Android-related failures: https://dev.azure.com/uno-platform/Uno%20Platform/_build/results?buildId=173133&view=ms.vss-test-web.build-test-results-tab

src/Uno.UI/UI/Xaml/NativeApplication.cs
```diff
@@ -152,9 +152,9 @@ internal bool TryHandleIntent(Intent intent)
 		/// </summary>
 		/// <param name="type">A type full name</param>
 		/// <returns>The assembly that contains the specified type</returns>
-		[Export(nameof(GetTypeAssemblyFullName))]
 		[global::System.ComponentModel.EditorBrowsable(global::System.ComponentModel.EditorBrowsableState.Never)]
-		public static string GetTypeAssemblyFullName(string type) => Type.GetType(type)?.Assembly.FullName;
+		public static string GetTypeAssemblyFullName(string type) =>
+			throw new NotSupportedException("`static` methods with [Export] are not supported on NativeAOT.");
```
jeromelaban, line 158: This change will break this particular scenario:

https://github.com/unoplatform/Uno.UITest/blob/94d027295b779e28064aebf99aeaee2b393ad558/src/Uno.UITest.Helpers/Helpers/AppQueryExtensions.cs#L21

Uno itself is not using this `.XamlParent` helper for Uno.UITest, which is why it's likely working properly. Others maybe using it, but it's probably not possible to tell who's using it in the wild.

We could keep it for native renderers, and accept the fact that native renderers won't support native AOT. What's the build error when this attribute exists? Can it be ignored at app compiled time? If so, it may be good enough to keep it for native.

jonpryor: I think that scenario is *already* broken, due to line 31: https://github.com/unoplatform/Uno.UITest/blob/94d027295b779e28064aebf99aeaee2b393ad558/src/Uno.UITest.Helpers/Helpers/AppQueryExtensions.cs#L31

```csharp
var controlType = "md5" + GetMd5Hash(nameSpace + ":" + assembly) + "." + typeName;
```

Use of MD5 for package names was removed in dotnet/java-interop@baf43b09, circa *2019*.  MD5 has not been supported since *Xamarin.Android 11.3.0*, before the .NET migration!

This code is *broken*, and has been for quite some time.

jonpryor: The question remains around what to do about now.  The "safest" approach may be to restrict this to .NET 10 builds, a'la:

```csharp
#if !NET10_0_OR_GREATER
    [Export(nameof(GetTypeAssemblyFullName))]
#endif // !NET10_0_OR_GREATER
    [global::System.ComponentModel.EditorBrowsable(global::System.ComponentModel.EditorBrowsableState.Never)]
    public static string GetTypeAssemblyFullName(string type) =>
#if NET10_0_OR_GREATER
        throw new NotSupportedException("`static` methods with [Export] are not supported on NativeAOT.");
#else  // NET10_0_OR_GREATER
        Type.GetType(type)?.Assembly.FullName;
#endif // NET10_0_OR_GREATER
```

The *other* question is where this change belongs: separately (i.e. "this PR"), or as part of a .NET 10 PR?  The conditional change described above could live in PR.

jeromelaban: Right, let's break for net10, that will be good enough, particularly it's been broken for a known codepath for a while :)

jeromelaban (approved):

PR #21308, opened 2025-08-20
fix: Managed pointer input pass-through native layer on iOS

MartinZikmund:
**GitHub Issue:** closes https://github.com/unoplatform/kahua-private/issues/328

Pointer input is not passing through the native layer due to clipping

- Input is captured in a new top level layer which contains the clipped native layer
- Native layer is pass-through unless native elements are inside
- In such case we hit test against the clipping layer and let all "clipped" input pass through to the input layer

## PR Checklist ✅

### DIFF
--- a/src/Uno.UI.Runtime.Skia.AppleUIKit/Devices/Input/TopViewLayer.cs
+++ b/src/Uno.UI.Runtime.Skia.AppleUIKit/Devices/Input/TopViewLayer.cs
@@ -1,8 +1,5 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
+﻿using CoreAnimation;
+using CoreGraphics;
 using Foundation;
 using UIKit;
 using Uno.UI.Runtime.Skia.AppleUIKit;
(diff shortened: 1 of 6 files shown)


src/Uno.UI.Runtime.Skia.AppleUIKit/Devices/Input/TopViewLayer.cs
```diff
@@ -0,0 +1,49 @@
...
+
+/// <summary>
+/// Top layer that receives all touch events and forwards them to the
+/// managed pointer handling. It also contains  <see cref="NativeOverlayLayer" /> which
+/// will only capture input of the native element within.
+/// Input handling must be in a layer (compared to override RootViewController.TouchesXXX methods)
+/// in order to be able to properly get the multitouch events.
+/// </summary>
+[Register("TopViewLayer")]
+internal class TopViewLayer : UIView
```
ramezgerges, line 17: Why do we need a TopViewLayer when NativeOverlayLayer has a HitTest method?

ramezgerges (approved):

jeromelaban (approved):

dr1rrb (approved):
LGTM, but we need to test multi-touch with this change on a real device before next release

src/Uno.UI.Runtime.Skia.AppleUIKit/Devices/Input/TopViewLayer.cs
MartinZikmund: Because this is what will capture the actual managed input. Previously `NativeOverlayLayer` had input handling, but holes caused it not to capture the input that happened within those "holes" - causing the original issue

PR #21317, opened 2025-08-21
fix(nativedispatchersynchronizationcontext): always enqueue on Normal

ramezgerges:
Our NativeDispatcherSynchronizationContext should always enqueue on Normal. Currently, it enqueues on the same priority as the current dispatcher job. This is publicly visible implications, e.g. if a user `await`s any task on the UI thread, it should always enqueue on Normal. For example, pointer events on X11 are scheduled on `High`, but user code that subscribes to pointer events and awaits a task should still be put on Normal.

Investigated during https://github.com/unoplatform/private/issues/894

### DIFF
--- a/src/Uno.UI.Dispatching/Native/NativeDispatcher.cs
+++ b/src/Uno.UI.Dispatching/Native/NativeDispatcher.cs
@@ -31,8 +31,6 @@ namespace Uno.UI.Dispatching
 			new Queue<Delegate>(), // Idle
 		};
 
-		private readonly NativeDispatcherSynchronizationContext[] _synchronizationContexts;
-
 		private readonly object _gate = new();
 
 		private readonly long _startTime;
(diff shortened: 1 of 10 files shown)


src/Uno.UI.Dispatching/Native/NativeDispatcher.cs
```diff
@@ -57,30 +55,14 @@ private NativeDispatcher()
...
 			_currentPriority = NativeDispatcherPriority.Normal;
 
-			_synchronizationContexts = new NativeDispatcherSynchronizationContext[]
-			{
-				new(this, NativeDispatcherPriority.High),
-				new(this, NativeDispatcherPriority.Normal),
-				new(this, NativeDispatcherPriority.Low),
-				new(this, NativeDispatcherPriority.Idle),
-			};
+			SynchronizationContext = new NativeDispatcherSynchronizationContext(this);
```
jeromelaban, line 58: For a given task or async method, the queuing of continuations should stay in the same priority. Changing the priority should be done through another explicit requeue on the dispatcher.

jeromelaban: Update: WinUI restores the normal priority on continuations.

jeromelaban (approved):

MartinZikmund (approved):

PR #21256, opened 2025-08-13
chore: Update to .NET 10 Preview 7

jonpryor:
Context: https://github.com/unoplatform/uno/pull/19934
Context: https://github.com/unoplatform/uno/pull/21140
Context: https://github.com/dotnet/sdk/issues/48949

unoplatform/uno#19934 was the first attempt to build unoplatform/uno with .NET 10.  It *built* (mostly), but it had *lots* of unit test failures (over 3700) that were not understood at the time.

unoplatform/uno#21140 was an attempt to cleanup & simplify #19934. It partially succeeded, but required additional reconsideration for a cleaner and more understandable merge.

(Later, PR #21183 helped explain many of the unit test failures that PR #19934 observed: Xamarin.UITest doesn't work reliably under .NET 9! See microsoft/appcenter#2646 and microsoft/appcenter#327!)

Some of the contents of #19934 and #21140 have been split out into separately merged PRs such as #21230 (CS0246 fixes) and #21199 (remove `[Export]` in some scenarios).

Squash existing "known good" and related changes:

  * doc updates

  * Bumping target framework versions,

  * Use uno.check 1.32.0-dev.45, which knows about .NET 10 Preview 7.

  * `$(TreatWarningsAsErrors)`=true + NuGet warnings.  For example:

        error NU1510: Warning As Error:
        PackageReference System.Private.Uri will not be pruned.
        Consider removing this package from your dependencies, as it is likely unnecessary.

  * `$(TreatWarningsAsErrors)`=true + CS8604 warnings due to nullability changes, particularly in iOS.

  * `$(TreatWarningsAsErrors)`=true + IL2\* warnings, some of which @jonpryor was able to fix in a reasonable timeframe, and some of which was thrown into `$(NoWarn)` to deal with later.

  * Remove hardcoded `$(TargetFramework)`=net8.0 from `Uno.UI.RemoteControl.TestProcessor.csproj`, so that it builds, and remove other uses of `net8.0` from everywhere other than `src/SolutionTemplate`.

  * Provide `%(PackageReference.Version)` for packages that were producing NU1015 errors, specifically `Uno.Fonts.OpenSans`.

  * Remove `@(PackageReference)` values which produced NU1510 warnings such as:

        warning NU1510: PackageReference Microsoft.Win32.Registry will not be pruned. Consider removing this package from your dependencies, as it is likely unnecessary.

[...]

PR #21359, opened 2025-08-27
chore: build & run SamplesApp.Skia.netcoremobile with NativeAOT on Android

jonpryor:
Context: https://github.com/unoplatform/uno/pull/21256 / 52a6f3e4c792f258a5f9aee91958cd0aa0c6d3a5
Context: https://github.com/dotnet/android/pull/10461
Context: https://github.com/dotnet/android/issues/10457
Context: https://github.com/dotnet/android/issues/10463
Context: https://github.com/AwesomeAssertions/AwesomeAssertions/issues/290

unoplatform/uno#21256 added support for building with .NET 10, and
one of the new features in .NET 10 is that Android has (very!)
preliminary preview support for [NativeAOT][0].

As building `SamplesApp.Skia.netcoremobile.csproj` with NativeAOT
takes a significant amount of time and disk space, we've decided to
introduce a new `Tests - Android+NativeAOT Skia` stage to build and
run these unit tests within an Android+NativeAOT environment.
To help reduce disk usage, after building the `.apk` we delete the
`obj` directory.

Update `android-run-skia-runtime-tests.sh` to always create
the `$(build.sourcesdirectory)/build/uitests-failure-results` path
before existing with an error, as failure to do so means that the
`PublishBuildArtifacts@1` YAML task fails:

	##[error]Publishing build artifacts failed with an error: Not found PathtoPublish: /agent/_work/1/s/build/uitests-failure-results

which in turn means subsequent `DownloadBuildArtifacts@0` /
**Download previous test runs failed tests** steps *also* always fail:

	##[error]Artifact uitests-android-nativeaot-failure-results not found for build 174635. Please ensure you have published artifacts in any previous phases of the current build.

Update `ApplicationData.GetRoamingFolder()` to explicitly create
`Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData)`.
(Apparently this isn't created by default under NativeAOT?)
This avoids a startup assertion:

	I NativeAotFromAndroid: App failed to initialize: System.IO.DirectoryNotFoundException: IO_PathNotFound_Path, /data/user/0/uno.platform.samplesapp.skia/files/.config/b63c4306-f361-42d0-bc1d-5be385a95c78.txt
	I NativeAotFromAndroid:    at System.IO.FileSystem.DeleteFile(String) + 0xe7
	I NativeAotFromAndroid:    at SamplesApp.App.<AssertApplicationData>g__AssertCanCreateFile|32_3(StorageFolder) + 0x10c
	I NativeAotFromAndroid:    at SamplesApp.App.AssertApplicationData() + 0xa5
	I NativeAotFromAndroid:    at SamplesApp.App..ctor() + 0x2a3
	I NativeAotFromAndroid:    at SamplesApp.Droid.Application.<>c.<.ctor>b__0_0() + 0x18
	I NativeAotFromAndroid:    at Microsoft.UI.Xaml.NativeApplication.<OnActivityStarted>b__7_0() + 0x12
	I NativeAotFromAndroid:    at Uno.UI.Runtime.Skia.Android.AndroidHost.<Run>g__CreateApp|2_10(ApplicationInitializationCallbackParams _) + 0xf
	I NativeAotFromAndroid:    at Microsoft.UI.Xaml.Application.StartPartial(ApplicationInitializationCallback callback) + 0xb5
	I NativeAotFromAndroid:    at Uno.UI.Runtime.Skia.Android.AndroidHost.Run() + 0x306

Enable NativeAOT builds by updating
`SamplesApp.Skia.netcoremobile.csproj` to set the [`$(PublishAot)`][1]
property to the `$(SkiaPublishAot)` MSBuild property.  This allows
us to build a single project for NativeAOT --
`SamplesApp.Skia.netcoremobile.csproj` -- *without* trying to build
every referenced project for NativeAOT, which is what happens if you
instead try `dotnet publish -p:PublishAot=true …`, which fails:

	% dotnet build -c Release -p:PublishAot=true src/SamplesApp/SamplesApp.Skia.netcoremobile/SamplesApp.Skia.netcoremobile.csproj -bl
	…
	…/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(121,5): error NETSDK1207: Ahead-of-time compilation is not supported for the target framework.

	% dotnet publish src/SamplesApp/SamplesApp.Skia.netcoremobile/SamplesApp.Skia.netcoremobile.csproj \
	  -c Release -r android-x64 -f net10.0-android -p:UnoTargetFrameworkOverride=net10.0-android -p:SkiaPublishAot=true -bl
	# succeeds… after 15 minutes…

This also requires updating `$(NoWarn)` to ignore the hundreds of
IL trimmer warnings.  We're just trying to see where things stand for now.

Additionally, we need to publish with `-r android-x64` in order to
avoid the build error:

	…/Microsoft.NETCore.Native.Publish.targets(92,5): error MSB3030: Could not copy the file "bin/Release/net10.0-android/native/SamplesApp.so" because it was not found.

Because it's `bin/Release/net10.0-android/android-{arm64,x64}/native/SamplesApp.so`!

Oddly, using `-r android-x64` still results in *both* ABIs being
included, which appears to be a unoplatform/uno "bug":

	% unzip -l src/SamplesApp/SamplesApp.Skia.netcoremobile/bin/Release/net10.0-android/android-x64/publish/uno.platform.samplesapp.skia-Signed.apk | grep SamplesApp.so
	763563120  08-28-2025 19:47   lib/x86_64/libSamplesApp.so
	757982224  08-28-2025 19:48   lib/arm64-v8a/libSamplesApp.so

The need for `-r android-x64` may be related to dotnet/android#10457.

.NET Crypto support isn't propertly initialized in Android+NativeAOT
apps in .NET 10 RC1; see dotnet/android#10463.  This may be fixed in
dotnet/android#10461, but in the meantime we can manually call
`AndroidCryptoNative_InitLibraryOnLoad()` so that methods such as
`SHA1.Create()` won't throw.

Exclude tests which don't pass under NativeAOT.  These fall into
three major categories:

 1. Failures that don't make sense, likely due to unknown bugs within
    the NativeAOT toolchain itself, failing due to
    `TargetParameterCountException` or `InvalidOperationException`
    described below.  Some of these are worked around (see below),
    while others are ignored via e.g.:

        [Ignore("DataRowAttribute.GetData() wraps data in an extra array under NativeAOT; not yet understood why.")]

 2. Failures which are due to `AwesomeAssertions` using
    `MethodInfo.MakeGenericMethod()`; see
    AwesomeAssertions/AwesomeAssertions#290:

        Unhandled exception. System.NotSupportedException: 'AwesomeAssertions.Equivalency.Steps.GenericEnumerableEquivalencyStep.HandleImpl[System.Int32](AwesomeAssertions.Equivalency.Steps.EnumerableEquivalencyValidator,System.Object[],System.Collections.Generic.IEnumerable`1[System.Int32])' is missing native code. MethodInfo.MakeGenericMethod() is not compatible with AOT compilation. Inspect and fix AOT related warnings that were generated when the app was published. For more information see https://aka.ms/nativeaot-compatibility
           at System.Reflection.Runtime.MethodInfos.RuntimeNamedMethodInfo`1.GetUncachedMethodInvoker(RuntimeTypeInfo[], MemberInfo) + 0x2a
           at System.Reflection.Runtime.MethodInfos.RuntimeMethodInfo.Invoke(Object, BindingFlags, Binder, Object[], CultureInfo) + 0x3f
           at AwesomeAssertions.Equivalency.Steps.GenericEnumerableEquivalencyStep.Handle(Comparands, IEquivalencyValidationContext, IValidateChildNodeEquivalency) + 0x21a
           at AwesomeAssertions.Equivalency.EquivalencyValidator.TryToProveNodesAreEquivalent(Comparands, IEquivalencyValidationContext) + 0x129
           at AwesomeAssertions.Equivalency.EquivalencyValidator.AssertEquality(Comparands, EquivalencyValidationContext) + 0x43
           at AwesomeAssertions.Collections.GenericCollectionAssertions`3.BeEquivalentTo[TExpectation](IEnumerable`1, Func`2, String, Object[]) + 0x1f9

    Basically, the `.BeBeEquivalentTo()` extension method cannot
    currently work in NativeAOT environments.

    These are ignored via:

        [Ignore(".BeEquivalentTo() unsupported under NativeAOT; see: https://github.com/AwesomeAssertions/AwesomeAssertions/issues/290")]

 3. Other limitations within the NativeAOT environment.  For example,
    Android+NativeAOT does not yet have a GC bridge, meaning every
    `Java.Lang.Object` subclass instance is *never* collected by the
    GC unless explicitly `.Dispose()`d

Update `UnitTestsControl.cs` to try to work around some NativeAOT
"weirdness" and provide better error messages when tests fail in
certain scenarios.  Previously, many tests would fail with one of:

	System.Reflection.TargetParameterCountException: Arg_ParmCnt
	    at System.Reflection.DynamicInvokeInfo.ThrowForArgCountMismatch() + 0x83
	    at System.Reflection.DynamicInvokeInfo.Invoke(Object, IntPtr, Object[], BinderBundle, Boolean) + 0x136
	    at Internal.Reflection.Execution.MethodInvokers.InstanceMethodInvoker.Invoke(Object, Object[], BinderBundle, Boolean) + 0x4f
	    at Internal.Reflection.Core.Execution.MethodBaseInvoker.Invoke(Object, Object[], Binder, BindingFlags, CultureInfo) + 0x48
	    at System.Reflection.Runtime.MethodInfos.RuntimeMethodInfo.Invoke(Object, BindingFlags, Binder, Object[], CultureInfo) + 0x70
	    at Uno.UI.Samples.Tests.UnitTestsControl.<>c__DisplayClass67_2.<<ExecuteTestsForInstance>g__InvokeTestMethod|5>d.MoveNext() + 0x4b9

or:

	System.InvalidOperationException: Missing parameter does not have default value
	    at Uno.UI.Samples.Tests.UnitTestsControl.ExpandArgumentsWithDefaultValues(Object[] methodArguments, ParameterInfo[] methodParameters) + 0x138
	    at Uno.UI.Samples.Tests.UnitTestsControl.<>c__DisplayClass67_4.<<ExecuteTestsForInstance>b__12>d.MoveNext() + 0x92
	--- End of stack trace from previous location ---
	    at Uno.UI.Samples.Tests.UnitTestsControl.<>c__DisplayClass67_2.<<ExecuteTestsForInstance>g__InvokeTestMethod|5>d.MoveNext() + 0x549

A problem is that the reported test that was failing would be
reported as e.g. `When_Add_Remove(System.Object[])`, which is a test
method which *does not exist*; it's actually
`When_Add_Remove(object, int, LeakTestStyles, RuntimeTestPlatforms)`.
Additionally, which parameter doesn't have a default value?
Or how does the parameter count not match? Or…

An intermediate form of this commit would wrap the
`TargetParameterCountException` in an `InvalidOperationException`,
resulting in messages such as:

	System.InvalidOperationException: Exception thrown while invoking Uno.UI.Tests.Windows_Globalization.Given_NumeralSystemTranslator.When_NumeralSystemIsMtei(System.String, System.String) with arguments { System.Object[]{ 1 as System.String, ꯱ as System.String } }.
	---> System.Reflection.TargetParameterCountException: Parameter count mismatch.

Note that `When_NumeralSystemIsMtei(System.String, System.String)`
takes two arguments, but it's given *one* argument with value
`{{ System.Object[]{ 1 as System.String, ꯱ as System.String } }}`,
i.e. *the values are there*, just "wrapped" in an extra array.

Update `UnitTestsControl.InvokeTestMethod()` to always call
`ExpandArgumentsWithDefaultValues()` as part of test method
invocation, so that we have a centralized place to look for such
"extra array wrapping".  Update `ExpandArgumentsWithDefaultValues()`
so that if it encounters an array for a parameter type which isn't
an array, the array is expanded as additional arguments.

Log various "weird" scenarios via `Console.WriteLine()` to make it
easier to get a "complete" listing of such failing methods by using
`adb logcat` output.

TODO? the migration to AwesomeAssertions in 80c07056 results in some
"bizarre" failure messages, e.g.

	Microsoft.VisualStudio.TestTools.UnitTesting.AssertFailedException: Expected value to be less than or equal to 81 because
	TextBox/SingleTextBox;Microsoft.UI.Xaml.VisualStateManager;Microsoft.UI.Xaml.Controls.Grid;…;ElementStub/HorizontalScrollBar;…

which is *truncated* at *4000* characters.  *Something* is wonky there.

[0]: https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/
[1]: https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/?tabs=windows%2Cnet8#publish-native-aot-using-the-cli
[2]: https://developer.android.com/guide/topics/manifest/application-element#debug
[3]: https://github.com/dotnet/java-interop/commit/90ac202e9a82bd13b9e160686bd14b5a5ddca0e9

### DIFF
--- a/build/ci/tests/.azure-devops-tests-android-nativeaot-skia.yml
+++ b/build/ci/tests/.azure-devops-tests-android-nativeaot-skia.yml
@@ -1,0 +1,131 @@
+parameters:
+  vmLinuxPool: ''
+  UNO_UWP_BUILD: ''
+  XAML_FLAVOR_BUILD: ''
+
+jobs:
+##
+## Android+NativeAOT Skia
+##
+
+- job: Android_NativeAOT_Tests
+  displayName: ' ' ## Name is concatenated with the matrix group name
+  timeoutInMinutes: 45
+  dependsOn: Skia_Android_NativeAOT_Tests_Build
+  pool: ${{ parameters.vmLinuxPool }}
+
+  variables:
+    CI_Build: true
+    SourceLinkEnabled: false
+    NUGET_PACKAGES: $(Agent.WorkFolder)/.nuget
+
+  strategy:
+    matrix:
+
+      'Runtime Tests 0':
+        ANDROID_SIMULATOR_APILEVEL: 34
+        UITEST_TEST_MODE_NAME: RuntimeTests
+        UNO_UITEST_BUCKET_ID: RuntimeTests
+        UITEST_RUNTIME_TEST_GROUP: 0
+        UITEST_RUNTIME_TEST_GROUP_COUNT: 5
+        SAMPLEAPP_ARTIFACT_NAME: uitests-android-nativeaot-skia-build
+        TARGETPLATFORM_NAME: net10
+        UITEST_TEST_TIMEOUT: '2600s'
+
+      'Runtime Tests 1':
+        ANDROID_SIMULATOR_APILEVEL: 34
+        UITEST_TEST_MODE_NAME: RuntimeTests
+        UNO_UITEST_BUCKET_ID: RuntimeTests
+        UITEST_RUNTIME_TEST_GROUP: 1
+        UITEST_RUNTIME_TEST_GROUP_COUNT: 5
+        SAMPLEAPP_ARTIFACT_NAME: uitests-android-nativeaot-skia-build
+        TARGETPLATFORM_NAME: net10
+        UITEST_TEST_TIMEOUT: '2600s'
+
+      'Runtime Tests 2':
+        ANDROID_SIMULATOR_APILEVEL: 34
+        UITEST_TEST_MODE_NAME: RuntimeTests
+        UNO_UITEST_BUCKET_ID: RuntimeTests
+        UITEST_RUNTIME_TEST_GROUP: 2
+        UITEST_RUNTIME_TEST_GROUP_COUNT: 5
+        SAMPLEAPP_ARTIFACT_NAME: uitests-android-nativeaot-skia-build
+        TARGETPLATFORM_NAME: net10
+        UITEST_TEST_TIMEOUT: '2600s'
+
+      'Runtime Tests 3':
+        ANDROID_SIMULATOR_APILEVEL: 34
+        UITEST_TEST_MODE_NAME: RuntimeTests
+        UNO_UITEST_BUCKET_ID: RuntimeTests
+        UITEST_RUNTIME_TEST_GROUP: 3
+        UITEST_RUNTIME_TEST_GROUP_COUNT: 5
+        SAMPLEAPP_ARTIFACT_NAME: uitests-android-nativeaot-skia-build
+        TARGETPLATFORM_NAME: net10
+        UITEST_TEST_TIMEOUT: '2600s'
+
+      'Runtime Tests 4':
+        ANDROID_SIMULATOR_APILEVEL: 34
+        UITEST_TEST_MODE_NAME: RuntimeTests
+        UNO_UITEST_BUCKET_ID: RuntimeTests
+        UITEST_RUNTIME_TEST_GROUP: 4
+        UITEST_RUNTIME_TEST_GROUP_COUNT: 5
+        SAMPLEAPP_ARTIFACT_NAME: uitests-android-nativeaot-skia-build
+        TARGETPLATFORM_NAME: net10
+        UITEST_TEST_TIMEOUT: '2600s'
+
+  steps:
+  - checkout: self
+    clean: true
+
+  - bash: |
+        echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | sudo tee /etc/udev/rules.d/99-kvm4all.rules
+        sudo udevadm control --reload-rules
+        sudo udevadm trigger --name-match=kvm
+    displayName: 'Enable KVM'
+
+  - task: DownloadPipelineArtifact@2
+    displayName: Downloading $(SAMPLEAPP_ARTIFACT_NAME)
+    inputs:
+      artifact: $(SAMPLEAPP_ARTIFACT_NAME)
+      path: '$(build.sourcesdirectory)/build/$(SAMPLEAPP_ARTIFACT_NAME)/android'
+
+  - task: DownloadBuildArtifacts@0
+    condition: gt(variables['System.JobAttempt'], 1)
+    continueOnError: true
+    displayName: Download previous test runs failed tests
+    inputs:
+        artifactName: uitests-android-nativeaot-failure-results
+        downloadPath: '$(build.sourcesdirectory)/build'
+
+  - template: ../templates/dotnet-install.yml
+
+  - bash: |
+      chmod +x $(build.sourcesdirectory)/build/test-scripts/android-run-skia-runtime-tests.sh
+      UNO_UITEST_APP_ID=uno.platform.samplesapp.skia.nativeaot \
+        $(build.sourcesdirectory)/build/test-scripts/android-run-skia-runtime-tests.sh
+
+    displayName: Run Android+NativeAOT Skia Tests
+
+  - task: PublishTestResults@2
+    condition: always()
+    inputs:
+      testRunTitle: 'Android+NativeAOT Skia Runtime Tests $(UITEST_RUNTIME_TEST_GROUP)'
+      testResultsFormat: 'NUnit'
+      testResultsFiles: '$(build.sourcesdirectory)/build/*TestResult*.xml'
+      failTaskOnFailedTests: true
+
+  - task: PublishBuildArtifacts@1
+    condition: always()
+    retryCountOnTaskFailure: 3
+    inputs:
+      PathtoPublish: $(build.artifactstagingdirectory)
+      ArtifactName: android-nativeaot-skia-results
+      ArtifactType: Container
+
+  - task: PublishBuildArtifacts@1
+    condition: always()
+    displayName: Publish Failed Tests Results
+    retryCountOnTaskFailure: 3
+    inputs:
+      PathtoPublish: $(build.sourcesdirectory)/build/uitests-failure-results
+      ArtifactName: uitests-android-nativeaot-failure-results
+      ArtifactType: Container
--- a/src/SamplesApp/SamplesApp.Skia.netcoremobile/SamplesApp.Skia.netcoremobile.csproj
+++ b/src/SamplesApp/SamplesApp.Skia.netcoremobile/SamplesApp.Skia.netcoremobile.csproj
@@ -32,6 +32,9 @@
 		-->
 		<NoWarn>$(NoWarn);NU1703;SYSLIB1045</NoWarn>
 
+		<!-- Ignore linker warnings for now -->
+		<NoWarn>$(NoWarn);IL2026;IL2057;IL2062;IL2065;IL2067;IL2068;IL2070;IL2072;IL2075;IL2077;IL2080;IL2104;IL2111;IL2122;IL3000;IL3002;IL3050;IL3053</NoWarn>
+
 		<!--
 		aab is the default packaging format in net6 API 31.
 		We need an APK for deployment on simulators.
@@ -53,6 +56,8 @@
 
 		<!-- Required when not building inside VS to avoid dependency issues -->
 		<PreBuildUnoUITasks Condition="'$(BuildingInsideVisualStudio)'==''">true</PreBuildUnoUITasks>
+
+		<PublishAot Condition=" '$(SkiaPublishAot)' != '' ">$(SkiaPublishAot)</PublishAot>
 	</PropertyGroup>
 
 	<ItemGroup>
@@ -291,6 +296,7 @@
 
 	<ItemGroup>
 		<TrimmerRootDescriptor Include="LinkerConfig.xml" />
+		<TrimmerRootAssembly Include="SamplesApp.Skia" RootMode="All" />
 	</ItemGroup>
 
 	<ItemGroup>
--- a/build/ci/tests/.azure-devops-tests-android-nativeaot-skia-build.yml
+++ b/build/ci/tests/.azure-devops-tests-android-nativeaot-skia-build.yml
@@ -1,0 +1,61 @@
+parameters:
+  vmLinuxPool: ''
+  UNO_UWP_BUILD: ''
+  XAML_FLAVOR_BUILD: ''
+
+jobs:
+- job: Skia_Android_NativeAOT_Tests_Build
+  displayName: 'Build Skia Android+NativeAOT Samples App'
+  timeoutInMinutes: 60
+  cancelTimeoutInMinutes: 1
+
+  pool: ${{ parameters.vmLinuxPool }}
+
+  variables:
+    CombinedConfiguration: Release|Any CPU
+    CI_Build: true
+
+    UNO_UWP_BUILD: ${{ parameters.UNO_UWP_BUILD }}
+    XAML_FLAVOR_BUILD: ${{ parameters.XAML_FLAVOR_BUILD }}
+
+  steps:
+  - checkout: self
+    clean: true
+
+  - template: ../templates/gitversion.yml
+  - template: ../templates/dotnet-mobile-install-linux.yml
+    parameters:
+      UnoCheckParameters: '--tfm net10.0-android'
+
+  - powershell: dotnet publish src/SamplesApp/SamplesApp.Skia.netcoremobile/SamplesApp.Skia.netcoremobile.csproj -c Release -r android-x64 -f net10.0-android -p:UnoTargetFrameworkOverride=net10.0-android -p:ApplicationId=uno.platform.samplesapp.skia.nativeaot -p:SkiaPublishAot=true /bl:$(build.artifactstagingdirectory)/logs/build-skia-android-nativeaot.binlog
+    displayName: Build Android+NativeAOT Skia Head
+
+  - script: >
+      rm -Rf src/SamplesApp/SamplesApp.Skia.netcoremobile/obj
+    workingDirectory: $(Build.SourcesDirectory)
+    displayName: Delete obj folder to free up disk space
+
+  - task: CopyFiles@2
+    displayName: 'Copy Generated Android APK'
+    inputs:
+      SourceFolder: $(build.sourcesdirectory)/src/SamplesApp/SamplesApp.Skia.netcoremobile/bin/Release/net10.0-android/android-x64/publish/
+      Contents: 'uno.platform.samplesapp.skia.nativeaot-Signed.apk'
+      TargetFolder: $(build.artifactstagingdirectory)/android-nativeaot
+      CleanTargetFolder: false
+      OverWrite: false
+      flattenFolders: false
+
+  - task: PublishPipelineArtifact@1
+    displayName: 'Publish Android+NativeAOT Binaries'
+    retryCountOnTaskFailure: 3
+    inputs:
+      targetPath: $(build.artifactstagingdirectory)/android-nativeaot
+      ArtifactName: uitests-android-nativeaot-skia-build
+
+  - task: PublishBuildArtifacts@1
+    retryCountOnTaskFailure: 3
+    condition: always()
+    inputs:
+      PathtoPublish: $(build.artifactstagingdirectory)/logs
+      ArtifactName: skia-samples-app-binlog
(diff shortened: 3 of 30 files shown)


jonpryor:
There's going to be another SNAFU with this idea: https://github.com/dotnet/java-interop/commit/90ac202e9a82bd13b9e160686bd14b5a5ddca0e9

.NET 10 Previews 2 through 7 have a bug in .NET for Android wherein the app crashes when attempting to use boolean fields.  As this PR builds `SamplesApp.Skia.netcoremobile` against net10.0-android, it thus encounters this behavior and crashes; from [android-device-log-RuntimeTests-0-RuntimeTests.interim.txt](https://dev.azure.com/uno-platform/_apis/resources/Containers/36138191/android-nativeaot-skia-results?itemPath=android-nativeaot-skia-results%2Fandroid-skia-34-net10%2Fandroid-device-log-RuntimeTests-0-RuntimeTests.interim.txt):

```
F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
F DEBUG   : Build fingerprint: 'google/sdk_gphone64_x86_64/emu64xa:14/UE1A.230829.036.A4/12096271:user/release-keys'
F DEBUG   : Revision: '0'
F DEBUG   : ABI: 'x86_64'
F DEBUG   : Timestamp: 2025-08-27 20:06:55.197697199+0000
F DEBUG   : Process uptime: 351s
F DEBUG   : Cmdline: uno.platform.samplesapp.skia
F DEBUG   : pid: 4518, tid: 4518, name: samplesapp.skia  >>> uno.platform.samplesapp.skia <<<
F DEBUG   : uid: 10190
F DEBUG   : signal 6 (SIGABRT), code -1 (SI_QUEUE), fault addr --------
F DEBUG   : Abort message: 'JNI DETECTED ERROR IN APPLICATION: attempt to access field boolean android.widget.RelativeLayout$LayoutParams.alignWithParent of type boolean with the wrong type byte: 0x77
F DEBUG   :     in call to SetByteField
F DEBUG   :     from void mono.java.lang.NativeDispatcherImplementor.n_run()'
F DEBUG   :     rax 0000000000000000  rbx 00007ffe9831c418  rcx 000072d75ab19610  rdx 0000000000000006
F DEBUG   :     r8  000072d6c8633e91  r9  000072d6c8633e91  r10 00007ffe9831c420  r11 0000000000000203
F DEBUG   :     r12 000072d688675cb0  r13 00007ffe9831c608  r14 00000000000011a6  r15 00000000000011a6
F DEBUG   :     rdi 00000000000011a6  rsi 00000000000011a6
F DEBUG   :     rbp 0000000000000000  rsp 00007ffe9831c410  rip 000072d75ab19610
F DEBUG   : 11 total frames
F DEBUG   : backtrace:
F DEBUG   :       #00 pc 0000000000061610  /apex/com.android.runtime/lib64/bionic/libc.so (abort+192) (BuildId: fa337969c798946280caa45e2d71a2e7)
F DEBUG   :       #01 pc 00000000007fabab  /apex/com.android.art/lib64/libart.so (art::Runtime::Abort(char const*)+1627) (BuildId: b6dc79e02101ea00827a35a55ab6597a)
F DEBUG   :       #02 pc 0000000000037abd  /apex/com.google.mainline.primary.libs@340724000/lib64/libbase.so/a5cce001552a532d111a59af51bbb6cc5e93475522f033f9807fa5f8ad346840deb8c6475491b3425ed7ef34094fb412761afd70b2205820f67d8a4c71c39832/libbase.so (android::base::SetAborter(std::__1::function<void (char const*)>&&)::$_0::__invoke(char const*)+61) (BuildId: d848a07685150bc1df3af6b95e4bc316)
F DEBUG   :       #03 pc 0000000000036fe2  /apex/com.google.mainline.primary.libs@340724000/lib64/libbase.so/a5cce001552a532d111a59af51bbb6cc5e93475522f033f9807fa5f8ad346840deb8c6475491b3425ed7ef34094fb412761afd70b2205820f67d8a4c71c39832/libbase.so (android::base::LogMessage::~LogMessage()+354) (BuildId: d848a07685150bc1df3af6b95e4bc316)
F DEBUG   :       #04 pc 00000000005ed357  /apex/com.android.art/lib64/libart.so (art::JavaVMExt::JniAbort(char const*, char const*)+2695) (BuildId: b6dc79e02101ea00827a35a55ab6597a)
F DEBUG   :       #05 pc 00000000005ed3c5  /apex/com.android.art/lib64/libart.so (art::JavaVMExt::JniAbortV(char const*, char const*, __va_list_tag*)+85) (BuildId: b6dc79e02101ea00827a35a55ab6597a)
F DEBUG   :       #06 pc 00000000005de0af  /apex/com.android.art/lib64/libart.so (art::(anonymous namespace)::ScopedCheck::AbortF(char const*, ...) (.__uniq.99033978352804627313491551960229047428)+191) (BuildId: b6dc79e02101ea00827a35a55ab6597a)
F DEBUG   :       #07 pc 00000000005e40bd  /apex/com.android.art/lib64/libart.so (art::(anonymous namespace)::ScopedCheck::CheckFieldAccess(art::ScopedObjectAccess&, _jobject*, _jfieldID*, bool, art::Primitive::Type) (.__uniq.99033978352804627313491551960229047428)+2301) (BuildId: b6dc79e02101ea00827a35a55ab6597a)
F DEBUG   :       #08 pc 00000000005e4caf  /apex/com.android.art/lib64/libart.so (art::(anonymous namespace)::CheckJNI::SetField(char const*, _JNIEnv*, _jobject*, _jfieldID*, bool, art::Primitive::Type, art::(anonymous namespace)::JniValueType) (.__uniq.99033978352804627313491551960229047428)+863) (BuildId: b6dc79e02101ea00827a35a55ab6597a)
F DEBUG   :       #09 pc 00000000005cf197  /apex/com.android.art/lib64/libart.so (art::(anonymous namespace)::CheckJNI::SetByteField(_JNIEnv*, _jobject*, _jfieldID*, signed char) (.__uniq.99033978352804627313491551960229047428.llvm.1924743731234850023)+39) (BuildId: b6dc79e02101ea00827a35a55ab6597a)
F DEBUG   :       #10 pc 0000000000e71155  <anonymous:64797000>
```

(I had forgotten encountering this crash, but the crash is why I fixed the java-interop bug two weeks ago…)

Immediate-term, I'll build `SamplesApp.Skia.netcoremobile` as a net9 app, just to verify that artifact name/etc. behavior works as expected.

Longer-term, we either wait for the next September .NET 10 Preview release -- which should have the fix for this abort -- *or* we alter `SamplesApp.Skia.netcoremobile` to avoid hitting `RelativeLayout.LayoutParams.alignWithParent`.

I'm partial to the latter, as it will speed things up.

jonpryor:
"Interestingly", the [crash noted above](https://github.com/unoplatform/uno/pull/21359#issuecomment-3229629874)? It's caused by making the app Debuggable!  This patch:

```diff
diff --git a/src/SamplesApp/SamplesApp.Skia.netcoremobile/Android/Main.Android.cs b/src/SamplesApp/SamplesApp.Skia.netcoremobile/Android/Main.Android.cs
index 79e647c76c..c3601f89f6 100644
--- a/src/SamplesApp/SamplesApp.Skia.netcoremobile/Android/Main.Android.cs
+++ b/src/SamplesApp/SamplesApp.Skia.netcoremobile/Android/Main.Android.cs
@@ -40,6 +40,7 @@ using Uno.UI;
 namespace SamplesApp.Droid
 {
 	[global::Android.App.ApplicationAttribute(
+		Debuggable = true,
 		Label = "@string/SamplesAppName",
 		Icon = "@mipmap/icon",
 		Banner = "@drawable/banner",
```

breaks *both* **Tests - Android Skia** and the (currently nearly identical) **Tests - Android+NativeAOT Skia**.

Offhand, I'm not sure why that would happen…but this *does* explain my confusion yesterday around why *both* **Tests - Android Skia** and **Tests - Android+NativeAOT Skia** would be crashing in the same way when running different binaries!

TIL, and I'm still confuzzled.

build/ci/tests/.azure-devops-tests-android-nativeaot-skia.yml
```diff
@@ -0,0 +1,131 @@
...
+
+      'Runtime Tests 1':
+        ANDROID_SIMULATOR_APILEVEL: 34
+        UITEST_TEST_MODE_NAME: RuntimeTests
+        UNO_UITEST_BUCKET_ID: RuntimeTests
+        UITEST_RUNTIME_TEST_GROUP: 1
+        UITEST_RUNTIME_TEST_GROUP_COUNT: 5
+        SAMPLEAPP_ARTIFACT_NAME: uitests-android-nativeaot-skia-build
+        TARGETPLATFORM_NAME: net10
+        UITEST_TEST_TIMEOUT: '2600s'
```
ramezgerges, line 43: The first one has UITEST_RUNTIME_TEST_GROUP_COUNT=4 but the others (correctly) have UITEST_RUNTIME_TEST_GROUP_COUNT=5

src/SamplesApp/SamplesApp.Skia.netcoremobile/SamplesApp.Skia.netcoremobile.csproj
```diff
@@ -53,6 +57,8 @@
 
 		<!-- Required when not building inside VS to avoid dependency issues -->
 		<PreBuildUnoUITasks Condition="'$(BuildingInsideVisualStudio)'==''">true</PreBuildUnoUITasks>
+
+		<PublishAot Condition=" '$(SkiaPublishAot)' != '' ">$(SkiaPublishAot)</PublishAot>
```
ramezgerges, line 60: I'm wondering why we need SkiaPublishAot on top of PublishAot

src/SamplesApp/SamplesApp.Skia.netcoremobile/SamplesApp.Skia.netcoremobile.csproj
```diff
@@ -12,6 +12,7 @@
 		<AssemblyName>SamplesApp</AssemblyName>
 		<DefineConstants>$(DefineConstants);XAMARIN;HAS_UNO</DefineConstants>
 		<DefineConstants Condition=" '$(UseMonoRuntime)' == 'false' ">$(DefineConstants);RUNTIME_CORECLR</DefineConstants>
+		<DefineConstants Condition=" '$(SkiaPublishAot)' == 'true' ">$(DefineConstants);RUNTIME_NATIVE_AOT</DefineConstants>
```
ramezgerges: I've seen this added in multiple csprojs. Maybe this can be moved to Directory.Build.targets?

ramezgerges (approved):

src/SamplesApp/SamplesApp.Skia.netcoremobile/SamplesApp.Skia.netcoremobile.csproj
jonpryor: @ramezgerges asked:

> I'm wondering why we need SkiaPublishAot on top of PublishAot

That's in the PR message:

> Enable NativeAOT builds by updating `SamplesApp.Skia.netcoremobile.csproj` to set the `$(PublishAot)` property to the `$(SkiaPublishAot)` MSBuild property. This allows us to build a single project for NativeAOT -- `SamplesApp.Skia.netcoremobile.csproj` -- without trying to build every referenced project for NativeAOT, which is what happens if you instead try `dotnet publish -p:PublishAot=true …`:
> 
> ```dotnetcli
> % dotnet build -c Release -p:PublishAot=true src/SamplesApp/SamplesApp.Skia.netcoremobile/SamplesApp.Skia.netcoremobile.csproj -bl
> …
> …/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(121,5): error NETSDK1207: Ahead-of-time compilation is not supported for the target framework.
> 
> % dotnet publish src/SamplesApp/SamplesApp.Skia.netcoremobile/SamplesApp.Skia.netcoremobile.csproj \
>   -c Release -r android-x64 -f net10.0-android -p:UnoTargetFrameworkOverride=net10.0-android -p:SkiaPublishAot=true -bl
> # succeeds… after 15 minutes…
> ```

src/SamplesApp/SamplesApp.Skia.netcoremobile/SamplesApp.Skia.netcoremobile.csproj
jonpryor: I have moved this to `Directory.Build.props`.  Let's see if it builds!

jonpryor: It built. Resolving…

PR #21360, opened 2025-08-27
chore: work around floating point precision issues in CreateLocalSession's assertion

### LINKED ISSUE #21356: Skia/Transforms: - Debug assertion on `SKMatrix` uses exact equality — flakes when `rootTransform` is defined
### Current behavior 🐛

`Visual.skia.cs` file: When running in Debug builds we have a Debug.Assert that compares canvas.TotalMatrix to an expected SKMatrix. The code uses exact equality (==) on SKMatrix instances and causes intermittent assertion failures when rootTransform is present. The values differ by tiny floating-point rounding errors (e.g. due to matrix multiplications), so the assertion fails even though the matrices are effectively equal for rendering purposes.

Code location: https://github.com/unoplatform/uno/blob/master/src/Uno.UI.Composition/Composition/Visual.skia.cs#L612

### Expected behavior 🎯

The Debug.Assert should compare matrices with a small epsilon tolerance (an element-wise "almost equal") rather than strict equality, to avoid flaky assertion failures caused by floating-point rounding.

The assertion triggers when rootTransform is defined and TotalMatrix has been multiplied with it (TotalMatrix * rootTransform); small rounding differences across platforms or calculation order make the exact-equality check fail.

### Workaround 🛠️

The workaround right now is to comment the `Debug.Assert()` line to prevent application from crashing.

## Proposed fix
Use an element-wise approximate comparison (with a configurable epsilon) instead of `==`. Do this in the Debug.Assert (or compute the matrices beforehand and assert using a helper method). Also avoid expensive matrix conversions inside asserts by computing expected/actual once and reusing them.

The epsilon value can be tuned; around 1e-4 is typical for SKMatrix single-precision comparisons. You should use the `CompositionMathHelpers.IsCloseReal()` to check the values.

### Renderer 🎨

- [x] Skia
- [ ] Native

### Affected platforms 📱💻🖥️

Desktop (Windows)

_No response_

### IDE version 🧑‍💻

_No response_

_No response_

ramezgerges:
**GitHub Issue:** closes https://github.com/unoplatform/uno/issues/21356

## PR Checklist ✅

### DIFF
--- a/src/Uno.UI.Composition/Composition/Visual.skia.cs
+++ b/src/Uno.UI.Composition/Composition/Visual.skia.cs
@@ -13,6 +13,7 @@ using SkiaSharp;
 using Uno.Disposables;
 using Uno.Extensions;
 using Uno.Helpers;
+using Uno.UI.Composition;
 using Uno.UI.Composition.Composition;
 
 namespace Microsoft.UI.Composition;
@@ -613,7 +614,7 @@ public partial class Visual : global::Microsoft.UI.Composition.CompositionObject
 			Debug.Assert(Unsafe.IsNullRef(ref rootTransform)
 				? canvas.TotalMatrix == TotalMatrix.ToSKMatrix()
 				// Due to the limit precision of doubles, instead of comparing the two matrices directly we compare the Frobenius norm of their difference to zero
-				: (canvas.TotalMatrix.ToMatrix4x4() - TotalMatrix * rootTransform).ToSKMatrix().Values.Sum(i => i * i) < 1e-5);
+				: CompositionMathHelpers.IsCloseRealZero((canvas.TotalMatrix.ToMatrix4x4() - TotalMatrix * rootTransform).ToSKMatrix().Values.Sum(i => i * i), 1e-5f));
 		}
 #endif
 	}
--- a/src/Uno.UI.Composition/Composition/Visual.skia.cs
+++ b/src/Uno.UI.Composition/Composition/Visual.skia.cs
@@ -5,6 +5,7 @@ using System;
 using System.Buffers;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
 using System.Numerics;
 using System.Runtime.CompilerServices;
 using Microsoft.CodeAnalysis.PooledObjects;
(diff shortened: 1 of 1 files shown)


carldebilly (commented):
Summing components to compare floats is not a sound metric (can cancel opposing errors and hides per‑component drift). We should compare each component (or a defined norm) with a justified epsilon.
* Logic is inlined “in place”. This should be centralized in a small helper (e.g. FloatingPointAssert / MathTolerance.AreClose) so tolerance policy lives in one spot.
* No tests validating the chosen epsilon (need at least: passes just under, fails just over).
* Epsilon appears arbitrary; please document why the chosen value is safe.
* Using LINQ here (in a performance‑critical path) adds unnecessary allocations/iterator overhead; use a straightforward loop or direct field access instead.

src/Uno.UI.Composition/Composition/Visual.skia.cs
```diff
@@ -609,7 +610,10 @@ private unsafe void CreateLocalSession(in PaintingSession parentSession, out Pai
 #if DEBUG
 		else
 		{
-			Debug.Assert(Unsafe.IsNullRef(ref rootTransform) ? canvas.TotalMatrix == TotalMatrix.ToSKMatrix() : canvas.TotalMatrix == (TotalMatrix * rootTransform).ToSKMatrix());
+			Debug.Assert(Unsafe.IsNullRef(ref rootTransform)
+				? canvas.TotalMatrix == TotalMatrix.ToSKMatrix()
+				// Due to the limit precision of doubles, instead of comparing the two matrices directly we compare the Frobenius norm of their difference to zero
+				: (canvas.TotalMatrix.ToMatrix4x4() - TotalMatrix * rootTransform).ToSKMatrix().Values.Sum(i => i * i) < 1e-5);
```
carldebilly: I'm pretty sure the Frobenius approach here is not the right one.

If you're concerned by the performance here, you should consider using SIMD or AVX, btw. In a tested helper, obviously.

ramezgerges:
I'm already doing a Frobenius norm (i.e. the sum of element squares) which should be good enough.
* I switched to the CompositionMathHelpers method you suggested. The policy needs to be hardcoded here because for drawing we can tolerate a much higher tolerance. Being accurate to just 3-4 decimal places is good enough actually.
* I'm waiting for you to test it manually in your repro. Having an automated test is not that important since it's just an assertion and the repro is not trivial.
* It's a debug assertion. The assemblies pushed to Nuget won't have this compiled in, so I prefer LINQ for the readability.

src/Uno.UI.Composition/Composition/Visual.skia.cs
ramezgerges: Why not? I don't really know which way would work best, I used whatever googling gave me. Performance is irrelevant here since it's just a Debug assertion,

carldebilly (approved):

src/Uno.UI.Composition/Composition/Visual.skia.cs
carldebilly: Fair enough, that's true that's just a debug assertion. But it is still in critical path, no?

ramezgerges: Yeah, it is, but published packages only include Release assemblies, so this is all not included. For our testing, it shouldn't regress the performance too much. We will reassess if it turns out to be a problem.

ajpinedam (approved):

PR #21398, opened 2025-09-02
fix: Enhance error message for StorageFolder use in ApplicationData

jeromelaban:
## PR Type:

- 🐞 Bugfix

When `ApplicationData` is used too early, it will now fail with an actionable error.


MartinZikmund (approved):

build/cSpell.json
```diff
@@ -123,105 +123,106 @@
 	"waitlist",
 	"subviews"
   ],
-  "ignoreWords": [
```
MartinZikmund: Is this formatting change intentional?

jeromelaban: It's not... VS changed it by itself, will revert.

PR #21428, opened 2025-09-05
fix: Visual state refresh on load

MartinZikmund:
**GitHub Issue:** closes https://github.com/unoplatform/kahua-private/issues/339

On control load we are refreshing all visual state groups based on their triggers. This is ok, as long as the states in the group have any state triggers. For non-trigger groups, we should avoid refreshing the state.

We do not refresh state for non-trigger groups

## PR Checklist ✅


src/Uno.UI/UI/Xaml/VisualStateGroup.cs
```diff
@@ -516,6 +516,11 @@ static bool Match(VisualTransition transition, string from, string to)
 
 		internal void RefreshStateTriggers(bool force = false)
 		{
+			if (!HasStateTriggers())
+			{
+				return;
+			}
+
```
ramezgerges, line 525: I'm wondering why an unnecessary refresh causes GoToState to function incorrectly?

ramezgerges (approved):

src/Uno.UI/UI/Xaml/VisualStateGroup.cs
MartinZikmund: For visual states which do not have triggers but are set before the control loads (e.g. the user control in the tests for this PR), we were reverting their state from the one explicitly set to "null" - undoing the changes that were applied

MartinZikmund:
TODO: Check what happens if some states are not trigger based, whereas some are -  if a state is set explicitly via `GoToState` and no tirgger-based state applies during the refresh, should the explicitly set state be undone?

jeromelaban (approved):

PR #21439, opened 2025-09-08
fix: Volume range should be 0.0 to 1.0, not percent based

### LINKED ISSUE #21128: MediaPlayer with LibVLCSharp: Volume control broken
### Current behavior 🐛

When changing the volume using MediaPlayer.Volume property, no sound can be heard. Volume control is not possible. Volume range on the MediaPlayer.Volume property is defined as 0.0 to 1.0.

### Expected behavior 🎯

Volume control should work as expected.

Use MediaPlayer with LibVLC extension, e.g. on Desktop target.

### Workaround 🛠️

Multiply volume with 100, so it is 0.0 to 100.0.

### Renderer 🎨

- [x] Skia
- [ ] Native

### Affected platforms 📱💻🖥️

Desktop (Windows)

6.1.23

### IDE version 🧑‍💻

Visual Studio 2022 17.14.2

Volume in LibVLCSharp is defined like this:

        /// <summary>
        /// Get/Set the volume in percents (0 = mute, 100 = 0dB)
        /// </summary>
        public int Volume
        {
            get => Native.LibVLCAudioGetVolume(NativeReference);
            set => Native.LibVLCAudioSetVolume(NativeReference, value);
        }

Volume in MediaPlayer is defined like this:

    //
    // Summary:
    //     Get or sets the audio volume for media playback.
    //
    // Returns:
    //     The audio volume for media playback. The allowed range of values is 0 to 1. Values
    //     outside of this range will be clamped.
    public double Volume
    { ...

So the value from MediaPlayer.Volume value must be scaled with factor 100 to apply it to LibVLC and vice versa (in src/AddIns/Uno.UI.MediaPlayer.Skia.X11/SharedMediaPlayerExtension.cs). But currently it is just rounded, leading to the observed problems.

lukasf:
**GitHub Issue:** closes #21128

## PR Type:

- 🐞 Bugfix

Although the Volume of MediaPlayer is defined as range 0.0-1.0, the Uno implementation incorrectly assumes it to be 0.0-100.0. When using standard conform volume on MediaPlayer, sound is inaudible.

Additionally, on the Windows platform, LibVLC uses a per-process volume control, displayed in the Windows volume mixer. So all LibVLC instances in one process share the same volume and volume changes done on one instance apply to all instances. Also, volume changes in the Windows volume mixer apply to all instances. But the "VolumeChanged" event from LibVLC was not handled, so instances reported wrong (old) volume, if the volume was changed in the mixer or in a different instance.

Fixed up the Uno implementation to use the correct 0.0-1.0 volume range on all platforms.

Additionally, the VolumeChanged event from LibVLC is handled to allow propagation of extenal volume changes (though windows volume mixer or other LibVLC instance), so the Volume property always reports the actual volume.

## PR Checklist ✅


lukasf:
Closing in favor of #21440

PR #21440, opened 2025-09-08
fix: Change MediaPlayer volume range to 0.0-1.0

### LINKED ISSUE #21128: MediaPlayer with LibVLCSharp: Volume control broken

lukasf:
**GitHub Issue:** closes #21128

## PR Type:

- 🐞 Bugfix

Although the Volume of MediaPlayer is defined as range 0.0-1.0, the Uno implementation incorrectly assumes it to be 0.0-100.0. When using standard conform volume on MediaPlayer, sound is inaudible.

Additionally, on the Windows platform, LibVLC uses a per-process volume control, displayed in the Windows volume mixer. So all LibVLC instances in one process share the same volume and volume changes done on one instance apply to all instances. Also, volume changes in the Windows volume mixer apply to all instances. But the "VolumeChanged" event from LibVLC was not handled, so instances reported wrong (old) volume, if the volume was changed in the mixer or in a different instance.

Fixed up the Uno implementation to use the correct 0.0-1.0 volume range on all platforms.

Additionally, the VolumeChanged event from LibVLC is handled to allow propagation of extenal volume changes (though windows volume mixer or other LibVLC instance), so the Volume property always reports the actual volume.

## PR Checklist ✅

### DIFF
--- a/src/AddIns/Uno.UI.MediaPlayer.Skia.X11/SharedMediaPlayerExtension.cs
+++ b/src/AddIns/Uno.UI.MediaPlayer.Skia.X11/SharedMediaPlayerExtension.cs
@@ -20,6 +20,7 @@ using Uno.UI.Dispatching;
 using MediaPlayer = Windows.Media.Playback.MediaPlayer;
 using Windows.Web.Http.Headers;
 using Uno.Logging;
+using System.Timers;
 
 #if IS_MPE_WIN32
 namespace Uno.UI.MediaPlayer.Skia.Win32;
@@ -195,21 +196,16 @@ public class SharedMediaPlayerExtension : IMediaPlayerExtension
 		VlcPlayer.Buffering += (o, a) => weakRef.GetTarget()?.OnBuffering(o, a);
 		VlcPlayer.Paused += (o, a) => weakRef.GetTarget()?.OnPaused(o, a);
 		VlcPlayer.TimeChanged += (o, a) => weakRef.GetTarget()?.OnTimeChanged(o, a); // PositionChanged fires way too frequently (probably every frame). We use TimeChanged instead.
-		VlcPlayer.VolumeChanged += (o, a) => weakRef.GetTarget()?.OnVlcVolumeChanged(o, a);
 
-		if (PlatformHelper.IsWindows)
-		{
-			// Update the player volume from VLC volume, since all VLC instances have
-			// shared volume control on Windows.
-			OnVlcVolumeChanged(null, null!);
-		}
-
+		//VlcPlayer.VolumeChanged += (o, a) => weakRef.GetTarget()?.OnVlcVolumeChanged(o, a);
 		// We need to start a timer to update the playback state, since libVLC doesn't
 		// provide a way to get the end of buffering without polling.
-		var timer = new DispatcherTimer { Interval = TimeSpan.FromMilliseconds(16) };
-		EventHandler<object> timerOnTick = (_, _) => weakRef.GetTarget()?.OnTick();
-		timer.Tick += timerOnTick;
-		_timerDisposable = Disposable.Create(() => timer.Tick -= timerOnTick);
+		// Also, attaching to VolumeChanged with debugger attached causes crashes on some systems,
+		// so we poll the volume on the timer as well.
+		var timer = new System.Timers.Timer(16);
+		ElapsedEventHandler timerOnTick = (_, _) => weakRef.GetTarget()?.OnTick();
+		timer.Elapsed += timerOnTick;
+		_timerDisposable = timer;
 		timer.Start();
 	}
 
@@ -480,19 +476,20 @@ public class SharedMediaPlayerExtension : IMediaPlayerExtension
 	{
 		// This is primarily to update the Buffering status, since libVLC doesn't
 		// expose a BufferingEnded event.
+		MediaPlaybackState? state = null;
 		switch (VlcPlayer.State)
 		{
 			case VLCState.Opening:
-				Player.PlaybackSession.PlaybackState = MediaPlaybackState.Opening;
+				state = MediaPlaybackState.Opening;
 				break;
 			case VLCState.Buffering:
-				Player.PlaybackSession.PlaybackState = MediaPlaybackState.Buffering;
+				state = MediaPlaybackState.Buffering;
 				break;
 			case VLCState.Playing:
-				Player.PlaybackSession.PlaybackState = MediaPlaybackState.Playing;
+				state = MediaPlaybackState.Playing;
 				break;
 			case VLCState.Paused:
-				Player.PlaybackSession.PlaybackState = MediaPlaybackState.Paused;
+				state = MediaPlaybackState.Paused;
 				break;
 			case VLCState.Stopped:
 			case VLCState.Ended:
@@ -500,19 +497,25 @@ public class SharedMediaPlayerExtension : IMediaPlayerExtension
 			case VLCState.NothingSpecial:
 				break;
 		}
-	}
 
-	private void OnVlcVolumeChanged(object? _, MediaPlayerVolumeChangedEventArgs _1)
-	{
-		NativeDispatcher.Main.Enqueue(() =>
+		if (state != null && state != Player.PlaybackSession.PlaybackState)
 		{
-			var volume = VlcPlayer.Volume;
-			if (_vlcPlayerVolume != volume && volume != -1)
+			NativeDispatcher.Main.Enqueue(() =>
 			{
-				_vlcPlayerVolume = volume;
+				Player.PlaybackSession.PlaybackState = state.Value;
+			});
+		}
+
+		// We also update the volume, in case it has has been changed externally.
+		var volume = VlcPlayer.Volume;
+		if (_vlcPlayerVolume != volume && volume != -1)
+		{
+			_vlcPlayerVolume = volume;
+			NativeDispatcher.Main.Enqueue(() =>
+			{
 				double newVolume = volume / 100.0; // VlcPlayer.Volume is in [0, 100]
-				Events?.RaiseExternalVolumeChanged(newVolume);
-			}
-		});
+				Events?.RaiseVolumeChanged(newVolume);
+			});
+		}
 	}
 }
--- a/src/AddIns/Uno.UI.MediaPlayer.Skia.X11/SharedMediaPlayerExtension.cs
+++ b/src/AddIns/Uno.UI.MediaPlayer.Skia.X11/SharedMediaPlayerExtension.cs
@@ -28,7 +28,7 @@ namespace Uno.UI.MediaPlayer.Skia.Win32;
 namespace Uno.UI.MediaPlayer.Skia.X11;
 #endif
 
-public class SharedMediaPlayerExtension : IMediaPlayerExtension
+internal class SharedMediaPlayerExtension : IMediaPlayerExtension
 {
 	private static int _vlcInitialized;
 	private static LibVLC _vlc = null!;
--- a/src/AddIns/Uno.UI.MediaPlayer.Skia.X11/SharedMediaPlayerExtension.cs
+++ b/src/AddIns/Uno.UI.MediaPlayer.Skia.X11/SharedMediaPlayerExtension.cs
@@ -39,7 +39,7 @@ public class SharedMediaPlayerExtension : IMediaPlayerExtension
 	private int _playlistIndex = -1; // -1 if no playlist or empty playlist, otherwise the 0-based index of the current track in the playlist
 	private MediaPlaybackList? _playlist; // only set and used if the current _player.Source is a playlist
 
-	private double _vlcPlayerVolume;
+	private int _vlcPlayerVolume = 100;
 
 	// the current effective url (e.g. current video in playlist) that is set natively
 	// DO NOT READ OR WRITE THIS. It's only used to RaiseSourceChanged.
@@ -195,20 +195,24 @@ public class SharedMediaPlayerExtension : IMediaPlayerExtension
 		VlcPlayer.Buffering += (o, a) => weakRef.GetTarget()?.OnBuffering(o, a);
 		VlcPlayer.Paused += (o, a) => weakRef.GetTarget()?.OnPaused(o, a);
 		VlcPlayer.TimeChanged += (o, a) => weakRef.GetTarget()?.OnTimeChanged(o, a); // PositionChanged fires way too frequently (probably every frame). We use TimeChanged instead.
+		VlcPlayer.VolumeChanged += (o, a) => weakRef.GetTarget()?.OnVlcVolumeChanged(o, a);
 
-		_vlcPlayerVolume = VlcPlayer.Volume;
+		if (PlatformHelper.IsWindows)
+		{
+			// Update the player volume from VLC volume, since all VLC instances have
+			// shared volume control on Windows.
+			OnVlcVolumeChanged(null, null!);
+		}
 
-		// For some reason, subscribing to VolumeChanged, even with an empty lambda, causes
-		// a native crash. Here's the crazy part: this only happens when a debugger is attached.
-		// This does not happen when a debugger is not attached even in debug builds. To work around
-		// this, we poll for the volume in OnTick instead.
-		// VlcPlayer.VolumeChanged += OnVolumeChanged;
+		// We need to start a timer to update the playback state, since libVLC doesn't
+		// provide a way to get the end of buffering without polling.
 		var timer = new DispatcherTimer { Interval = TimeSpan.FromMilliseconds(16) };
 		EventHandler<object> timerOnTick = (_, _) => weakRef.GetTarget()?.OnTick();
 		timer.Tick += timerOnTick;
 		_timerDisposable = Disposable.Create(() => timer.Tick -= timerOnTick);
 		timer.Start();
 	}
+
 
 	~SharedMediaPlayerExtension()
 	{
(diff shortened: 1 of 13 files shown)


src/AddIns/Uno.UI.MediaPlayer.Skia.X11/SharedMediaPlayerExtension.cs
```diff
@@ -195,21 +195,25 @@ public SharedMediaPlayerExtension(Windows.Media.Playback.MediaPlayer player)
...
+			// Update the player volume from VLC volume, since all VLC instances have
+			// shared volume control on Windows.
+			OnVlcVolumeChanged(null, null!);
+		}
 
-		// For some reason, subscribing to VolumeChanged, even with an empty lambda, causes
-		// a native crash. Here's the crazy part: this only happens when a debugger is attached.
-		// This does not happen when a debugger is not attached even in debug builds. To work around
-		// this, we poll for the volume in OnTick instead.
-		// VlcPlayer.VolumeChanged += OnVolumeChanged;
```
ramezgerges, line 205: You have ignored this comment. When I tested on my side, your subscription crashes the app when a debugger is attached. We can stick with the original logic as it was, unless there's a problem with it.

src/Uno.UWP/Media/Playback/IMediaPlayerEventsExtension.cs
```diff
@@ -25,6 +25,11 @@ public interface IMediaPlayerEventsExtension
 		/// </summary>
 		void RaiseVolumeChanged();
 
+		/// <summary>
+		///	Raises the <see cref="MediaPlayer.VolumeChanged"/> event
+		/// </summary>
+		void RaiseExternalVolumeChanged(double newVolume);
```
ramezgerges: This method is unnecessary. RaiseVolumeChanged is enough.

src/Uno.UWP/Media/Playback/MediaPlayer.cs
```diff
@@ -58,7 +58,7 @@ public double Volume
 			get => _volume;
 			set
 			{
-				_volume = value;
+				_volume = Math.Clamp(value, 0.0, 1.0);
```
ramezgerges: From WinUI's docs: "The audio volume for media playback. The allowed range of values is 0 to 1. Values outside of this range will be ignored.". So, I would instead ignore instead of clamping.

src/Uno.UI.Runtime.Skia.MacOS/UI/Xaml/Controls/MediaPlayerElement/MacOSMediaPlayerExtension.cs
```diff
@@ -256,7 +256,7 @@ public void NextTrack()
 
 	public void OnVolumeChanged()
 	{
-		NativeUno.uno_mediaplayer_set_volume(_nativePlayer, (float)(_player.Volume / 100));
+		NativeUno.uno_mediaplayer_set_volume(_nativePlayer, (float)_player.Volume);
```
ramezgerges, line 259: This needs to be tested on MacOS.

src/Uno.UI/UI/Xaml/Controls/MediaPlayerElement/MediaTransportControls.MediaPlayer.cs
```diff
@@ -459,13 +459,25 @@ private void OnVolumeChanged(object sender, RangeBaseValueChangedEventArgs e)
...
+				_mediaPlayer.Volume = e.NewValue / 100.0;
 			}
 
 			UpdateVolumeMuteStates();
 			ResetControlsVisibilityTimer();
 		}
 
+		private void OnPlayerVolumeChanged(MediaPlayer sender, object e)
+		{
+			_ = Dispatcher.RunAsync(CoreDispatcherPriority.Normal, () =>
```
ramezgerges: You don't need to enqueue on the dispatcher here, you're already on the UI thread.

src/Uno.UI/UI/Xaml/Controls/MediaPlayerElement/MediaTransportControls.MediaPlayer.cs
```diff
@@ -459,13 +459,25 @@ private void OnVolumeChanged(object sender, RangeBaseValueChangedEventArgs e)
 		{
 			if (_mediaPlayer is not null)
 			{
-				_mediaPlayer.Volume = e.NewValue;
+				_mediaPlayer.Volume = e.NewValue / 100.0;
```
ramezgerges: This will affect iOS and Android too. I see you made adjustments only to MacOS, Win32/X11 (the shared file), and WASM.

src/AddIns/Uno.UI.MediaPlayer.Skia.X11/SharedMediaPlayerExtension.cs
lukasf: It did not happen to me, no matter if I was running debug or release builds of the sample app. So I thought it might have been an older bug which has been fixed inbetween. I will test again.

src/Uno.UWP/Media/Playback/IMediaPlayerEventsExtension.cs
lukasf: RaiseVolumeChanged is called when someone sets the Volume property, and it will cause the volume to be applied to the LibVLC player instance. RaiseExternalVolumeChanged is instead called when the volume of LibVLC has been changed from outside, e.g. by using the Windows volume mixer. In that case, we need to set the backing field of the property (otherwise our MediaPlayer reports an incorrect, old value) and raise VolumeChanged. But we must not set the volume on LibVLC, as this might cause an infinite loop. That's why two separate events are needed.

src/Uno.UWP/Media/Playback/MediaPlayer.cs
lukasf: Right, I will change this.

src/Uno.UI.Runtime.Skia.MacOS/UI/Xaml/Controls/MediaPlayerElement/MacOSMediaPlayerExtension.cs
lukasf: The change seems logical, but I agree it should be tested. I don't have any Mac devices to test though.

src/Uno.UI/UI/Xaml/Controls/MediaPlayerElement/MediaTransportControls.MediaPlayer.cs
lukasf: MediaPlayer is not a DispatcherObject. Events can (and often will) be called from background threads. It's properties can also be accessed on any thread. So when changing the Volume property from a background thread, the VolumeChanged will also come on that background thread. But also when the user changes the volume through Windows volume mixer, the change is detected by LibVLC and VolumeChanged is raised (through RaiseExternalVolumeChanged), but here again it is called on some background thread of LibVLC, not on the UI thread.

src/Uno.UI/UI/Xaml/Controls/MediaPlayerElement/MediaTransportControls.MediaPlayer.cs
lukasf: I only found Wasm, MacOS and the shared LibVLC based implementation. I assumed the shared extension is used by Android and iOS as well. Searching again, I did not find any other implementations.

Can you point me to where they are?

lukasf: Nevermind, I just found them. They do not use the extension mechanism, which is why I missed them. I will apply the change there as well.

src/Uno.UWP/Media/Playback/MediaPlayer.cs
lukasf: Done

src/Uno.UI/UI/Xaml/Controls/MediaPlayerElement/MediaTransportControls.MediaPlayer.cs
lukasf: Fixed up Android and iOS implementations. I also verified the API specs for all plarforms: All platforms use volume range 0.0 to 1.0, except for LibVLC, so this is the only platform where we need to re-scale the new volume value.

src/Uno.UWP/Media/Playback/IMediaPlayerEventsExtension.cs
ramezgerges: I'm fine with this for now but my point was that the infinite loop scenario shouldn't happen because the "native value" and the "uno value" should be the same, so no further VolumeChanged events should keep going back and forth.

src/AddIns/Uno.UI.MediaPlayer.Skia.X11/SharedMediaPlayerExtension.cs
ramezgerges: We will need to set this back. Even if it works on your end, it doesn't work on mine on two different machines. I've also seen independent non Uno-related reports of VLC hanging on VolumeChanged, so it's definitely not something specific to my setup.

src/Uno.UI/UI/Xaml/Controls/MediaPlayerElement/MediaTransportControls.MediaPlayer.cs
ramezgerges: That's not how Uno works. all UI interactions and all public events are fired on the UI thread (through the dispatcher). VolumeChanged is called from RaiseVolumeChanged and when changing the Volume property, which both happen on the UI thread. For RaiseExternalVolumeChanged, it's the job of the platform-specific extention to call RaiseExternalVolumeChanged on the UI thread, which is already done for the shared libvlc implementation (which is the only one to call RaiseExternalVolumeChanged).

src/Uno.UWP/Media/Playback/IMediaPlayerEventsExtension.cs
ramezgerges: Actually, after further consideration, this doesn't make sense to me. There are 2 directions for volume changes to flow.
1. Managed to native. This happens when MediaPlayer.Volume changes, which calls OnVolumeChanged which then forwards the managed change in volume to the native media player implementation. This is platform specific.
2. Native to managed. This is done through RaiseVolumeChanged which is called from the platform-specific implementation/extension. This is already for "external VolumeChanged".

src/AddIns/Uno.UI.MediaPlayer.Skia.X11/SharedMediaPlayerExtension.cs
lukasf: We could theoretically detect if a Debugger is attached and have different logic for that case, but it increases complexity and does not add much benefit. So then it's probably best rolling back to the old behavior.

src/Uno.UWP/Media/Playback/IMediaPlayerEventsExtension.cs
lukasf: I just added this as an extra safeguard for infinite loops. Rounding issues can be tricky and VLC uses int values instead of float/double, so rounding is bound to happen in both directions. It might work without RaiseExternalVolumeChanged, but I'd rather not have surprises. It does not have any downsides to have a safe implementation for the back channel, does it?

I did not get your last comment. Yes these are the two scenarios. The second scenario is not covered in the current implementation. The new volume value from VLC is stored in the shared extension, but it is not propagated back to the MediaPlayer.Volume property. My code implements the second scenario through the (safe) RaiseExternalVolumeChanged method.

src/AddIns/Uno.UI.MediaPlayer.Skia.X11/SharedMediaPlayerExtension.cs
lukasf: I have just re-tested this and could not reproduce any hanging, neither in Debug, nor in Release build, both with Debugger attached. Still, planning to remove this. But I noticed that the current implementation uses a DispatcherTimer with a very high frequency (16ms). This puts a pretty high tax on the UI thread. Probably in 99.99% of the timer invocations, nothing ever changes, so this is wasting lots of precious UI thread resources for nothing. It would be a lot more efficient to use a normal background timer and only switch to the dispatcher thread if a change is actually detected. If you agree, I would change the implementation to use a background timer, to free up resources on the UI thread. Only if playback status or volume actually changes, we need to go to the UI thread.

ramezgerges: Sure, that could be useful. Thanks.

src/Uno.UWP/Media/Playback/IMediaPlayerEventsExtension.cs
ramezgerges: What I'm saying is that RaiseVolumeChanged and RaiseExternalVolumeChanged are the same thing. The second scenario is already covered with RaiseVolumeChanged, so unless I'm missing something, RaiseExternalVolumeChanged is redundant, no? E.g. in the libvlc implementation, RaiseVolumeChanged was called in OnTick (the code you removed and replaced with OnVlcVolumeChanged).

src/Uno.UI/UI/Xaml/Controls/MediaPlayerElement/MediaTransportControls.MediaPlayer.cs
lukasf: I noticed that all event-raising methods in the MediaPlayerExtension are explicitly dispatched to the UI thread already inside the extension. So indeed, this event will always be raised on the UI thread. I can remove the dispatcher call here.

Still, FYI: On WinUI/UWP MediaPlayer, events like MediaOpened, MediaFailed, etc are *never* called on the UI thread. So the Uno behavior is different, compared to WinUI...

ramezgerges: Oh, that's very interesting actually, thanks for pointing that out. Let's keep the behavior as it is (everything on the UI thread) to avoid regressing exiting apps that depend on that behavior, but very good catch.

src/Uno.UWP/Media/Playback/IMediaPlayerEventsExtension.cs
lukasf: I just checked the implementation again. The old code indeed called RaiseVolumeChanged. But this method will only cause the VolumeChanged event to be raised on the MediaPlayer. But it does not apply the new volume to the MediaPlayer.Volume property (see src/Uno.UWP/Media/Playback/MediaPlayer.Events.others.cs). So the MediaPlayer would have its VolumeChanged event raised, but it would still report the old, not updated volume on its Volume property. The new RaiseExternalVolumeChanged method has the "newVolume" parameter, which is applied to the MediaPlayer.Volume property backing field. Only then, the value of the MediaPlayer.Volume property is really updated.

We could also unify both methods by adding an *optional* nullable newVolume parameter to the RaiseVolumeChanged method. If the parameter is != null, it is applied to the backing field, otherwise only the event is called. That way, we don't have two methods but only one, which might be a bit better.

ramezgerges: Indeed, RaiseVolumeChanged should be taking a volume parameter read from the platform-specific media player provider. For now, let's make this parameter optional like you said and only propagate it for the libvlc implementation.

src/AddIns/Uno.UI.MediaPlayer.Skia.X11/SharedMediaPlayerExtension.cs
lukasf: Done

src/Uno.UWP/Media/Playback/IMediaPlayerEventsExtension.cs
lukasf: Done

src/Uno.UI/UI/Xaml/Controls/MediaPlayerElement/MediaTransportControls.MediaPlayer.cs
lukasf: Done

src/Uno.UI.Runtime.Skia.MacOS/UI/Xaml/Controls/MediaPlayerElement/MacOSMediaPlayerExtension.cs
lukasf: I cannot test iOS and MacOS. Other platforms seem to work fine.

ramezgerges (approved):

MartinZikmund:
@lukasf seems good! Can you please rebase on latest master and change the diffignore to point to 6.3 as base?

MartinZikmund (approved):

lukasf:
@MartinZikmund Rebased to master. I updated diffignore to 6.4, because that is latest now. I hope that is correct.

ramezgerges:
@lukasf Thanks for your contribution!

Issue #21244, opened 2025-08-11
Add Uno.NotImplemented attribute to FrameworkElement RenderPhase Property


rafael-rosa-knowcode:
### What would you like to be added? 💡

As the RenderPhase is NotImplemented, 

Not for [skia](https://github.com/unoplatform/uno/blob/a0386c80b1654064fd2d9121598485f930b484d3/src/Uno.UI/UI/Xaml/FrameworkElement.Interface.skia.cs#L64)
<img width="769" height="368" alt="Image" src="https://github.com/user-attachments/assets/5693e8db-90de-4ef8-9c86-57bbcd33b6fa" />

Not for [Reference](https://github.com/unoplatform/uno/blob/a0386c80b1654064fd2d9121598485f930b484d3/src/Uno.UI/UI/Xaml/FrameworkElement.Interface.reference.cs#L89)

<img width="962" height="865" alt="Image" src="https://github.com/user-attachments/assets/c251503e-4ddf-4620-9a51-fdd86a606f47" />

**I would like to request that the property has the Uno.NotImplemented attribute**

### Why is this needed? 🤔

I am using a filter with the tag Uno.Notimpleted and this property not implemented is being returned.

### Renderer 🎨

- [x] Skia
- [x] Native

### For which platforms 📱💻🖥️

All platforms 🌍

lukasf:
Instead of throwing NotImplementedException, can't you just no-nop in the get and set method? If the platform does not support render phase, it should irgnore the attribute. Right now, I need to define multiple data templates twice, once with render phase and once without. I'd really like to use a single data template. Platforms should make use of render phase, if they can, otherwise just ignore it. This would really help reduce unneccessary clutter.

jeromelaban:
@lukasf yes indeed, this is a better implementation at this time, throwing is not particularly useful.

booskander:
QA: review done

Implementation follows the contributer's suggestion.

PR #21441, opened 2025-09-08
fix: Do not throw when setting RenderPhase property on Skia

lukasf:
**GitHub Issue:** Related to #21244

## PR Type:

- 🐞 Bugfix

When the x:RenderPhase attribute is set, a NotImplementedException is raised at startup. It is not possible to share DataTemplates between platforms which do support RenderPhase and platform which don't (e.g. Skia).

On WinUI, the x:RenderPhase can be applied to any UI element. But it is only ever used when applied to ListView or GridView ItemTemplate. Applying it to unsupported elements in WinUI does not throw exceptions.

Setting x:RenderPhase on unsupported elements or platforms does not throw exceptions anymore. It just does not have any effect. So now DataTemplates can be easily shared between all platforms.

## PR Checklist ✅

### DIFF
--- a/src/Uno.UI/UI/Xaml/FrameworkElement.Interface.skia.cs
+++ b/src/Uno.UI/UI/Xaml/FrameworkElement.Interface.skia.cs
@@ -7,6 +7,7 @@ using System.Collections.Generic;
 using Windows.Foundation;
 using System.Runtime.CompilerServices;
 using System.Text;
+using Uno;
 using Uno.UI.Xaml;
 
 namespace Microsoft.UI.Xaml
@@ -61,13 +62,21 @@ namespace Microsoft.UI.Xaml
 			return finalSize;
 		}
 
+		[NotImplemented]
 		public int? RenderPhase
 		{
-			get => throw new NotImplementedException();
-			set => throw new NotImplementedException();
+			get
+			{
+				global::Windows.Foundation.Metadata.ApiInformation.TryRaiseNotImplemented("Microsoft.UI.Xaml.FrameworkElement", "RenderPhase");
+				return null;
+			}
+			set
+			{
+				global::Windows.Foundation.Metadata.ApiInformation.TryRaiseNotImplemented("Microsoft.UI.Xaml.FrameworkElement", "RenderPhase");
+			}
 		}
 
-		public void ApplyBindingPhase(int phase) => throw new NotImplementedException();
+		public void ApplyBindingPhase(int phase) { }
 
 		#region Background DependencyProperty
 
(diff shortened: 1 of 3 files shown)


src/Uno.UI/UI/Xaml/FrameworkElement.Interface.reference.cs
```diff
@@ -88,10 +88,10 @@ protected virtual void OnBackgroundChanged(DependencyPropertyChangedEventArgs e)
...
 		public int? RenderPhase
 		{
-			get => throw new NotImplementedException();
-			set => throw new NotImplementedException();
+			get;
+			set;
 		}
 
-		public void ApplyBindingPhase(int phase) => throw new NotImplementedException();
+		public void ApplyBindingPhase(int phase) { }
```
jeromelaban: Thanks for your contribution! 

We should still keep this one, the reference implementation is not supposed to be executed in any way.

src/Uno.UI/UI/Xaml/FrameworkElement.Interface.skia.cs
```diff
@@ -63,11 +63,11 @@ public Size AdjustArrange(Size finalSize)
...
 		public int? RenderPhase
 		{
-			get => throw new NotImplementedException();
-			set => throw new NotImplementedException();
+			get;
+			set;
 		}
 
-		public void ApplyBindingPhase(int phase) => throw new NotImplementedException();
+		public void ApplyBindingPhase(int phase) { }
```
jeromelaban: Let's do the "not implemented" logging as found here:

https://github.com/unoplatform/uno/blob/c03e234c1e4f30af6909786b39d19007009fb551/src/Uno.UWP/UI/ViewManagement/UISettings.cs#L78

it will log once per app instance.

Same for the others.

lukasf: Okay, will do that. Shall I also add the NotImplemented attribute, while I am at it?

src/Uno.UI/UI/Xaml/FrameworkElement.Interface.reference.cs
lukasf: Okay

src/Uno.UI/UI/Xaml/FrameworkElement.Interface.skia.cs
jeromelaban: Yes, let's do so indeed. Thanks!

src/Uno.UI/UI/Xaml/FrameworkElement.Interface.reference.cs
lukasf: Done

src/Uno.UI/UI/Xaml/FrameworkElement.Interface.skia.cs
lukasf: Done

jeromelaban (approved):

lukasf:
Thank you @jeromelaban, looking forward to seeing this in one of the upcoming builds!

PR #21470, opened 2025-09-12
fix(RenderTargetBitmap): Don't clip the rendered element using its current parent layout slot

jeromelaban:
## PR Type:

- 🐞 Bugfix

This allows the rendering of UIElement instances using any arbitrary size, regardless of their actual parent size.

Skia Renderer only.

### DIFF
--- a/src/Uno.UI/UI/Xaml/Media/Imaging/RenderTargetBitmap.skia.cs
+++ b/src/Uno.UI/UI/Xaml/Media/Imaging/RenderTargetBitmap.skia.cs
@@ -43,46 +43,58 @@ namespace Microsoft.UI.Xaml.Media.Imaging
 			var renderSize = element.RenderSize;
 			var visual = element.Visual;
 
-			if (renderSize is { IsEmpty: true } or { Width: 0, Height: 0 })
+			var previousClip = visual.LayoutClip;
+
+			try
 			{
-				return (0, 0, 0);
+				// Remove any existing layout clip, we want to render the full element, not
+				// the clipped part based on the existing parent's layout slot.
+				visual.LayoutClip = null;
+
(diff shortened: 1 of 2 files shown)


Xiaoy312 (approved):

ajpinedam (approved):

src/Uno.UI/UI/Xaml/Media/Imaging/RenderTargetBitmap.skia.cs
```diff
@@ -43,46 +43,58 @@ private static (int ByteCount, int Width, int Height) RenderAsBgra8_Premul(UIEle
...
-			if (renderSize is { IsEmpty: true } or { Width: 0, Height: 0 })
+			var previousClip = visual.LayoutClip;
+
+			try
 			{
-				return (0, 0, 0);
-			}
+				// Remove any existing layout clip, we want to render the full element, not
+				// the clipped part based on the existing parent's layout slot.
+				visual.LayoutClip = null;
```
ramezgerges, line 52: I Initially didn't like hot-fixing the clip here and I thought it would be a lot nicer to send this as a parameter to RenderRootVisual itself. But actually, this is exactly what we have to do because only the "layout clip" should be removed but not the clip from UIElement.Clip (I've tested on WinUI), so if we send a parameter to RenderRootVisual to ignore the clipping altogether for the "root element", then that's actually incorrect.

ramezgerges (approved):

src/Uno.UI/UI/Xaml/Media/Imaging/RenderTargetBitmap.skia.cs
jeromelaban: Will make a separate PR for the comment update.

PR #21492, opened 2025-09-18
fix(xamlreader): Don't fail on ItemsPanelTemplate inside a frameworktemplate

jeromelaban:
## PR Type:
- 🐞 Bugfix

Using an ItemsPaneltemplate inside a DataTemplate fails with:

```
The type 'Microsoft.UI.Xaml.Controls.ItemsPanelTemplate' does not support direct content.
```

This scenario is now properly supported.


ajpinedam (approved):
Good!

MartinZikmund (approved):

PR #21497, opened 2025-09-19
fix: `Maximizing`/`Minimizing` windows before activation

### LINKED ISSUE #21435: Cannot initialize window as maximized
### Discussed in https://github.com/unoplatform/uno/discussions/21339

<div type='discussions-op-text'>

<sup>Originally posted by **Tad29Eng** August 22, 2025</sup>
I create desktop app, how to startup main app window in maximize size (same as pressing maximize button in window) 
</div>

I try use in App.xaml.cs

```csharp
 var presenter = (OverlappedPresenter)MainWindow.AppWindow.Presenter;
 presenter.Maximize();
```

but this not worked, when app runing exception thrown: System.NullReferenceException Microsoft.UI.Windowing.OverlappedPresenter.Maximize()

MartinZikmund:
**GitHub Issue:** closes https://github.com/unoplatform/uno/issues/21435

Trying to initialize windows on Windows crashes with NRE.

Does not happen.

## PR Checklist ✅

### DIFF
--- a/src/Uno.UWP/Microsoft/UI/Windowing/OverlappedPresenter.cs
+++ b/src/Uno.UWP/Microsoft/UI/Windowing/OverlappedPresenter.cs
@@ -135,7 +135,7 @@ public partial class OverlappedPresenter : AppWindowPresenter
 		}
 		else
 		{
-			_pendingState = OverlappedPresenterState.Maximized;
+			_pendingState = OverlappedPresenterState.Restored;
 		}
 	}
 
--- a/src/Uno.UWP/Microsoft/UI/Windowing/OverlappedPresenter.cs
+++ b/src/Uno.UWP/Microsoft/UI/Windowing/OverlappedPresenter.cs
@@ -1,5 +1,3 @@
-using System.ComponentModel;
-using System.Runtime.CompilerServices;
 using Microsoft.UI.Windowing.Native;
 
 namespace Microsoft.UI.Windowing;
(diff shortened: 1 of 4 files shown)


src/Uno.UWP/Microsoft/UI/Windowing/OverlappedPresenter.cs
```diff
@@ -94,7 +103,17 @@ public bool IsAlwaysOnTop
...
-	public void Minimize(bool activateWindow) => Native.Minimize(activateWindow);
+	public void Minimize(bool activateWindow)
+	{
+		if (Native is not null)
+		{
+			Native.Minimize(activateWindow);
+		}
+		else
+		{
+			_pendingState = OverlappedPresenterState.Maximized;
```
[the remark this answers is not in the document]
ajpinedam: This comment makes sense

src/Uno.UWP/Microsoft/UI/Windowing/OverlappedPresenter.cs
```diff
@@ -77,13 +76,23 @@ public bool IsAlwaysOnTop
 
 	public bool HasTitleBar => _hasTitleBar;
 
-	public OverlappedPresenterState State => Native?.State ?? OverlappedPresenterState.Restored;
+	public OverlappedPresenterState State => Native?.State ?? _pendingState ?? OverlappedPresenterState.Restored;
```
ramezgerges, line 79: Doesn't having a pendingState here remove the need to do the same pendingState change in win32-specific code?

ramezgerges (approved):

src/Uno.UWP/Microsoft/UI/Windowing/OverlappedPresenter.cs
MartinZikmund: Unfortunately, it does not - for Win32 specifically the Maximize/Minimize methods would cause the window to be displayed before its actual activation, so when the pending state here is applied (and causes `Maximize()`), the Win32 does not actually maximize but rather just remembers that pending state for the actual initial window activation. It is a bit clunky unfortunately, but this seems to be the only way to actually initialize the window in the already maximized state instead of showing it and immediately subsequently maximizing it.

PR #21571, opened 2025-10-07
fix: Ensure pointer events are handled by the native element host

clairernovotny:
Fixes various mouse issues related to mouse events not being processed by code running in the native element host on Skia WASM

**GitHub Issue:** closes # https://github.com/unoplatform/uno-private/issues/505

Mouse events aren't seen by the native element host. Clicking does not do anything when it should.

Clicking works, scrolling with the wheel works.

## PR Checklist ✅

### DIFF
--- a/src/Uno.UI.Runtime.Skia.WebAssembly.Browser/ts/Runtime/BrowserPointerInputSource.ts
+++ b/src/Uno.UI.Runtime.Skia.WebAssembly.Browser/ts/Runtime/BrowserPointerInputSource.ts
@@ -59,6 +59,8 @@
 
 		private _source: any;
 		private _bootTime: Number;
+		// Cached reference to #uno-native-element-host. Refreshed if detached/replaced.
+		private _nativeElementHost: HTMLElement | null = null;
 
 		private constructor(manageSource: any) {
 			this._bootTime = Date.now() - performance.now();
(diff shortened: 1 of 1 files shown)


dr1rrb:
Looks great, but instead of validating source for each pointer event, are we able to subscribe to pointer events for the "Native host" [when we create it](https://github.com/unoplatform/uno/blob/46652420669fc9baec50e5a233013dff8d10cc11/src/Uno.UI.Runtime.Skia.WebAssembly.Browser/ts/Runtime/BrowserNativeElementHostingExtension.ts#L49) and request a `stopPropagation` instead ? This should be enough to prevent the  `BrowserInputSource` to receive events and would be a bit more performant.

dr1rrb:
> Looks great, but instead of validating source for each pointer event, are we able to subscribe to pointer events for the "Native host" [when we create it](https://github.com/unoplatform/uno/blob/46652420669fc9baec50e5a233013dff8d10cc11/src/Uno.UI.Runtime.Skia.WebAssembly.Browser/ts/Runtime/BrowserNativeElementHostingExtension.ts#L49) and request a `stopPropagation` instead ? This should be enough to prevent the `BrowserInputSource` to receive events and would be a bit more performant.

As @clairernovotny mentioned indeed this is not a valid solution as we are subscribing on phase 1 `capture: true` https://github.com/unoplatform/uno/pull/21571/files#diff-08e99cbcd93e57d1662cd7560a26060ff6ad6291ab31aab55711a06adea5008cR76 so the `body` will get the event first.

src/Uno.UI.Runtime.Skia.WebAssembly.Browser/ts/Runtime/BrowserPointerInputSource.ts
```diff
@@ -81,6 +83,48 @@
...
+
+			while (currentNode !== null) {
+				// Identity compare for cheapest match.
+				if (currentNode === host) {
+					return true;
+				}
+
+				// Shadow DOM hop if applicable; otherwise climb light DOM.
+				const rootNode: any = typeof (currentNode as any).getRootNode === "function"
+					? (currentNode as any).getRootNode()
```
dr1rrb: I suspect that will work only with component that uses a "shadow root" (which is probably the case with the monaco editor), otherwise [it will return directly the `HTMLDocument`](https://developer.mozilla.org/en-US/docs/Web/API/Node/getRootNode#return_value) which is a parent of the `_nativeElementHost`.

src/Uno.UI.Runtime.Skia.WebAssembly.Browser/ts/Runtime/BrowserPointerInputSource.ts
```diff
@@ -81,6 +83,48 @@
...
+					: null;
+
+				if (rootNode !== null && rootNode.host instanceof Node) {
```
dr1rrb: This is probably why it's working actually, the `host` property doesn't seem to exist on a [`Node`](https://developer.mozilla.org/en-US/docs/Web/API/Node) so the `rootNode.host instanceof Node` is always `false` so we always use the `currentNode.parentNode`.

Unless I misunderstood something, I think could just remove all the `getRootNode()` and just iterate the `currentNode.parentNode`

clairernovotny: Thanks, agree on both points.

* `getRootNode()` returns `Document` when there’s no shadow root, so we should fall back to `parentNode`.
* My check was wrong. `host` is on `ShadowRoot`, not on `Node`, so `rootNode.host instanceof Node` won't work... I'll fix it to use `instanceof ShadowRoot`.

If Monaco or anything inside the host uses Shadow DOM, a plain `parentNode` climb won’t cross the boundary and we’ll misclassify events.

src/Uno.UI.Runtime.Skia.WebAssembly.Browser/ts/Runtime/BrowserPointerInputSource.ts
```diff
@@ -81,6 +83,59 @@
...
+				const parent = (currentNode as any).parentNode as Node | null;
+				if (parent) {
+					currentNode = parent;
+					continue;
+				}
+
+				// Only if parentNode is null, check for a shadow boundary.
+				const rootNode: any = typeof (currentNode as any).getRootNode === "function"
+					? (currentNode as any).getRootNode()
+					: null;
```
dr1rrb: According to the support table, [the `getRootNode` is supported by all modern browser](https://caniuse.com/mdn-api_node_getrootnode), IMO we can safely use it without performing check:


```suggestion
				const rootNode = currentNode.getRootNode();
```

dr1rrb (approved):

PR #21638, opened 2025-10-19
feat(nativeelementhosting): Z-order support

### LINKED ISSUE #21376: [WebView2] [Desktop/Wasm] `WebView2.Visibility` is not working as expected
### Current behavior 🐛

1) Cannot hide a WebView2 element with `Visibility.Collapsed` in the code behind.
2) With two WebView2 elements, the z-order seems to be in conflict with `WebView2.Visibility`. The last WebView2 will always be on top regardless of their Visibility.

This makes it so we can't create a scenario where we have two WebViews and want to hide/show them based on a currently selected tab.

https://github.com/user-attachments/assets/27a14491-2cf4-47e8-b73f-166a815646f0

### Expected behavior 🎯

On Android (it is working, but even then, I feel like the WebViews might always be on-screen (just hidden by the other WebViews) which might not be the behavior we are aiming for):

https://github.com/user-attachments/assets/9621e656-0635-4265-8620-0d334d69793d

1) Add a `WebView2` and set its source to something like https://www.microsoft.com (and Visibility="Collapsed")
2) Add a button and define the click event in the code behind to hide/show the WebView
3) Launch the app on Desktop/Wasm and notice that the button can show the WebView but not hide it.

[WebViewSpike.zip](https://github.com/user-attachments/files/22032909/WebViewSpike.zip)

### Workaround 🛠️

Using `MyGrid.Children.Remove(MyWebView)` and `MyGrid.Children.Add(WebView)` works.

### Renderer 🎨

- [x] Skia
- [ ] Native

### Affected platforms 📱💻🖥️

Desktop (Windows), WebAssembly

Uno.Sdk 6.2.29

### IDE version 🧑‍💻

Visual Studio 2022 17.14.12

In the repro app, using `WebView2.Visibility` is working fine. Probably because the z-order of elements is decided differently.

ramezgerges:
**GitHub Issue:** closes https://github.com/unoplatform/uno/issues/21376

## PR Checklist ✅


MartinZikmund:
@ramezgerges Does this mean if there are two native elements on top of each other are technically still there, even if the bottom one is `Collapsed`? I'm thinking whether that could still pose problems, for example if the one on top would not be fully opaque - then the one underneath would still be visible to the user, even though it should be `Collapsed`.

ramezgerges:
@MartinZikmund Indeed. This complicates the lifecycle a little bit (at least until https://github.com/unoplatform/uno-private/issues/742),  but I've adjusted the changes to completely remove out-of-view native elements.

Xiaoy312 (approved):

src/Uno.UI.Runtime.Skia.MacOS/UnoNativeMac/UnoNativeMac/UNONative.h
```diff
@@ -35,6 +33,4 @@ void uno_native_measure(NSView* element, double childWidth, double childHeight,
 
 void uno_native_set_opacity(NSView* element, double opacity);
 
-void uno_native_set_visibility(NSView* element, bool visible);
```
spouliot, line 38: The p/invoke for this function (located in `src/Uno.UI.Runtime.Skia.MacOS/Native/NativeUno.cs`) also needs to be removed.

spouliot (approved):

ajpinedam (approved):

PR #21640, opened 2025-10-20
perf: Memory allocation improvements

ebariche:
## PR Type:

- ⚡Performance

## PR Checklist ✅

### DIFF
--- a/src/Uno.UI.Composition/Composition/CompositionGeometricClip.skia.cs
+++ b/src/Uno.UI.Composition/Composition/CompositionGeometricClip.skia.cs
@@ -28,6 +28,8 @@ partial class CompositionGeometricClip
 		return null;
 	}
 
+	private static readonly SKPath _spareTransformedPath = new();
+
 	internal override SKPath? GetClipPath(Visual visual)
 	{
 		if (Geometry is not null)
@@ -39,7 +41,8 @@ partial class CompositionGeometricClip
 				var path = geometrySource.Geometry;
 				if (!TransformMatrix.IsIdentity)
 				{
-					var transformedPath = new SKPath();
+					var transformedPath = _spareTransformedPath;
+					transformedPath.Rewind();
 					path.Transform(TransformMatrix.ToSKMatrix(), transformedPath);
 					path = transformedPath;
 				}
--- a/src/Uno.UI.Composition/Composition/RectangleClip.skia.cs
+++ b/src/Uno.UI.Composition/Composition/RectangleClip.skia.cs
@@ -1,5 +1,6 @@
 ﻿#nullable enable
 
+using System;
 using System.Numerics;
 using SkiaSharp;
 using Uno.Extensions;
@@ -9,9 +10,8 @@ namespace Microsoft.UI.Composition;
 
 partial class RectangleClip
 {
-	private static readonly SKPoint[] _radiiStore = new SKPoint[4];
-
 	private SKRoundRect? _skRoundRect;
+	private static readonly SKPath _spareClipPath = new();
 
 	private protected override Rect? GetBoundsCore(Visual visual)
 	{
@@ -22,10 +22,13 @@ partial class RectangleClip
 			height: Bottom - Top);
 	}
 
+	// The path returned here is reused, do not cache
 	internal override SKPath GetClipPath(Visual visual)
 	{
-		var path = new SKPath();
+		var path = _spareClipPath;
+		path.Rewind();
 		path.AddRoundRect(GetClipRoundedRect(visual));
+
 		return path;
 	}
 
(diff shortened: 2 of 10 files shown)


src/Uno.UI.Composition/Composition/Visual.skia.cs
```diff
@@ -37,15 +37,15 @@ public partial class Visual : global::Microsoft.UI.Composition.CompositionObject
 
 	// Since painting (and recording) is done on the UI thread, we need a single SKPictureRecorder per UI thread.
 	// If we move to a UI-thread-per-window model, then we need multiple recorders.
-	[ThreadStatic]
-	private static SKPictureRecorder? _recorder;
+	// [ThreadStatic]
```
jeromelaban: Remove?

src/Uno.UI.Composition/Composition/CompositionGeometricClip.skia.cs
```diff
@@ -39,7 +41,8 @@ partial class CompositionGeometricClip
 				var path = geometrySource.Geometry;
 				if (!TransformMatrix.IsIdentity)
 				{
-					var transformedPath = new SKPath();
+					var transformedPath = _spareTransformedPath;
+					transformedPath.Rewind();
```
ramezgerges, line 45: The path created or reused here is returned by GetClipPath, and there is no contract that says that the caller of GetClipPath cannot modify the returned path. If the allocations in GetClipPath showed up heavily in a trace, we can think about tracing all the different execution paths and making sure that the path returned from GetClipPath is either not modified, or cloned before being modified. But currently, this change is not guaranteed to be safe and I can already see at least one call that modified the returned path (GetNativeViewPath calls GetPostPaintingClip which calls GetClipPath and then there's a postClip.Transform call).

src/Uno.UI.Composition/Composition/RectangleClip.skia.cs
```diff
@@ -24,7 +24,8 @@ partial class RectangleClip
 
 	internal override SKPath GetClipPath(Visual visual)
 	{
-		var path = new SKPath();
+		var path = _spareClipPath;
+		path.Rewind();
```
ramezgerges, line 29: Same comment here.

src/Uno.UI.Composition/Composition/Visual.skia.cs
```diff
@@ -207,8 +207,11 @@ Matrix4x4 GetTransform()
 	/// </remarks>
 	internal void InvalidatePaint()
 	{
-		_picture?.Dispose();
-		_picture = null;
+		if (_picture != IntPtr.Zero)
+		{
+			UnoSkiaApi.sk_refcnt_safe_unref(_picture);
+			_picture = IntPtr.Zero;
+		}
```
ramezgerges, line 211: Definitely optional, but whenever we swap out SkiaSharp calls for direct P/Invoke'd skia calls, it would be great if we could keep the equivalent SkiaSharp calls that got removed in a comment so that we can restore them back easily to check if there's something wrong with the changes.

src/Uno.UWP/Storage/Streams/IBufferByteAccess.cs
```diff
@@ -0,0 +1,9 @@
+using System;
+
+namespace Windows.Storage.Streams
+{
+	public interface IBufferByteAccess
```
ramezgerges: What's the point of this interface? There is only one implementor anyway (Buffer).

ebariche:
Improves https://github.com/unoplatform/kahua-private/issues/349

src/Uno.UI.Composition/Composition/CompositionGeometricClip.skia.cs
ebariche: InsetClip.GetClipPath() was already cached ! So that contract already exists ! Anyways a 100,000 objects for 85s of runtime is not acceptable !

src/Uno.UI.Composition/Composition/RectangleClip.skia.cs
ebariche: See other.

src/Uno.UI/Helpers/SkiaRenderHelper.skia.cs
```diff
@@ -15,36 +15,55 @@ namespace Uno.UI.Helpers;
...
+	private static readonly SKPictureRecorder _recorder = new();
+
+	// This is used all the time, on all platforms but X11, when no native elements are present - DO NOT MODIFY
+	private static readonly SKPath _emptyClipPath = new();
+
+	// This is used on X11, when no native elements are present - DO NOT MODIFY
+	private static int _x11Width;
+	private static int _x11Height;
+	private static float _x11Scale;
+	private static SKPath? _x11ClipPath;
```
ramezgerges: Nothing about this is X11-specific, so the naming is misleading. This is used when `invertPath`, which happens to be the case only on X11 for now, but not necessarily.

src/Uno.UI/Helpers/SkiaRenderHelper.skia.cs
```diff
@@ -54,38 +73,81 @@ internal static void RenderPicture(SKCanvas canvas, SKPicture? picture, SKColor
...
+
+			invertedPath.Op(clipPath, SKPathOp.Difference, invertedPath);
+
+			clipPath.Dispose();
+
 			return invertedPath;
 		}
+	}
+
+	private static SKPath GetOrUpdateX11ClippingPath(int width, int height, float scale)
```
ramezgerges: Same comment about the unnecessary mention of X11, and it should be made clear in the name that this is the empty native clip case.

ramezgerges (approved):

PR #21646, opened 2025-10-20
fix: Adjust devserver startup

jeromelaban:
## PR Type:

- 🐞 Bugfix

Adjust devserver startup on non-windows systems

## PR Checklist ✅

### DIFF
--- a/src/Uno.UI.DevServer.Cli/McpProxy.cs
+++ b/src/Uno.UI.DevServer.Cli/McpProxy.cs
@@ -1,13 +1,6 @@
-using System.Collections.Concurrent;
-using System.CommandLine;
-using System.Diagnostics;
-using System.Net;
-using System.Net.Sockets;
-using System.Text;
+using System;
+using System.Collections.Generic;
 using System.Text.Json;
-using Microsoft.AspNetCore.Http;
-using Microsoft.AspNetCore.Mvc;
-using Microsoft.Extensions.AI;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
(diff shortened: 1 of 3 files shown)


src/Uno.UI.DevServer.Cli/McpProxy.cs
```diff
@@ -99,7 +109,8 @@ public async Task<int> RunAsync(string hostPath, int port, List<string> forwarde
 		if (controller.ExitCode != 0)
 		{
 			_logger.LogError("DevServer controller exited with code {ExitCode}", controller.ExitCode);
-			var so = await stdoutCapture; var se = await stderrCapture;
+			var so = await stdoutCapture;
```
Youssef1313: NOTE: Waiting for stdout then waiting for stderr can lead to deadlocks.

The stream buffers are somewhere around 4kb.

If the child process tries to write more than 4kb to stderr, this will deadlock.

- Child process will successfully write the first 4kb.
- Parent process here is waiting to consume the whole stdout.
- Child process cannot write more to stderr, because it's waiting for the parent process to consume from the stream so that there is space in the buffer to write.
- Parent process isn't consuming stderr because it's waiting for stdout.
- Deadlock.

Current approach is so risky.

Youssef1313: Current approach can also easily lead to threadpool starvation, on Windows specifically.

jeromelaban: Indeed, there were also two competing implementation for the same purpose, that should do now. Thanks!

PR #21679, opened 2025-10-25
fix(GlobalStaticResources): add an exclude xml doc tag to the generated type definition

### LINKED ISSUE #18616: Annotate generated GlobalStaticResources class

Please annotate the generated `GlobalStaticResources` class with either `[GeneratedCode]` and/or `[EditorBrowsableAttribute(Never)]` so it doesn't pollute library documentation generated by docfx :)

### Why is this needed

image

### For which platform

_No response_

_No response_

DevTKSS:
**GitHub Issue:** closes #18616

## PR Type:

- 🐞 Bugfix

the public accessor level of the GlobalStaticResources class is resulting in publication of this generated type

Refering to the docfx documentation:

> The `<exclude />` documentation comment excludes the type or member on a per API basis using C# documentation comment

the added tag should generally already work for excluding this type from the api docs

Also adding the EditorBrowsable Attribute, for the case that someone might not use docfx then another docs creation tool that might not consider the exclude tag, the attribute should serve as redundant backup

## PR Checklist ✅


DevTKSS:
@MartinZikmund 🤔 if I read the log right, its looking for the files beeing produced, but all of them are listed in the end just not maybe in the order it seems to expect them?
I just implemented the points from the issue report at the position Jerome linked.

 Is there something missing that needs to be done more?

I included the type for now in my filterConfig.yml for my docfx build but would be cool if this wouldn't be required in the future 😁

MartinZikmund (approved):

PR #22025, opened 2025-12-03
fix(CalendarView): wrong string value when Spanish culture

ajpinedam:
**GitHub Issue:** closes https://github.com/unoplatform/add-private/discussions/51

## PR Type:

- 🐞 Bugfix

Wrong string format when CalendarView uses es-Es (Spanish Language)

image

Screenshot 2025-12-02 231819

## PR Checklist ✅

### DIFF
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_CalendarView.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_CalendarView.cs
@@ -201,6 +201,7 @@ public class Given_CalendarView
 			MinDate = DateTimeOffset.Now.AddDays(-10),
 			MaxDate = DateTimeOffset.Now.AddDays(10)
 		};
+
 		Assert.AreEqual(2, calendar.SelectedDates.Count);
 		Assert.AreEqual(day1, calendar.SelectedDates[0]);
 		Assert.AreEqual(day2, calendar.SelectedDates[1]);
@@ -302,6 +303,32 @@ public class Given_CalendarView
 			offsets.Zip(offsets.Skip(1)).All(x => x.Second >= x.First),
 			$"should never rewind back: (v-offsets: {string.Join(", ", offsets)})"
 		);
+	}
+
+	[TestMethod]
+	public async Task When_Spanish_Language()
+	{
+		var calendarView = new CalendarView();
+		TestServices.WindowHelper.WindowContent = calendarView;
+		await TestServices.WindowHelper.WaitForLoaded(calendarView);
+		calendarView.Language = "es-ES";
+		await TestServices.WindowHelper.WaitForIdle();
+		calendarView.SetDisplayDate(new DateTimeOffset(new DateTime(2024, 1, 1)));
+		await TestServices.WindowHelper.WaitForIdle();
+		Assert.AreEqual("enero de 2024", calendarView.TemplateSettings.HeaderText);
+	}
+
+	[TestMethod]
+	public async Task When_English_Language()
+	{
+		var calendarView = new CalendarView();
+		TestServices.WindowHelper.WindowContent = calendarView;
+		await TestServices.WindowHelper.WaitForLoaded(calendarView);
+		calendarView.Language = "en"; //default culture usually is en
+		await TestServices.WindowHelper.WaitForIdle();
+		calendarView.SetDisplayDate(new DateTimeOffset(new DateTime(2024, 1, 1)));
+		await TestServices.WindowHelper.WaitForIdle();
+		Assert.AreEqual("January 2024", calendarView.TemplateSettings.HeaderText);
 	}
 }
 #endif
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_CalendarView.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_CalendarView.cs
@@ -308,26 +308,26 @@ public class Given_CalendarView
 	[TestMethod]
 	public async Task When_Spanish_Language()
 	{
-		var calendarView = new CalendarView();
+		var calendarView = new CalendarView()
+		{
+			Language = "es-ES"
+		};
+		calendarView.SetDisplayDate(new DateTimeOffset(new DateTime(2024, 1, 1)));
 		TestServices.WindowHelper.WindowContent = calendarView;
 		await TestServices.WindowHelper.WaitForLoaded(calendarView);
-		calendarView.Language = "es-ES";
-		await TestServices.WindowHelper.WaitForIdle();
-		calendarView.SetDisplayDate(new DateTimeOffset(new DateTime(2024, 1, 1)));
-		await TestServices.WindowHelper.WaitForIdle();
 		Assert.AreEqual("enero de 2024", calendarView.TemplateSettings.HeaderText);
 	}
 
 	[TestMethod]
 	public async Task When_English_Language()
 	{
-		var calendarView = new CalendarView();
+		var calendarView = new CalendarView()
+		{
+			Language = "en-US"
+		};
+		calendarView.SetDisplayDate(new DateTimeOffset(new DateTime(2024, 1, 1)));
 		TestServices.WindowHelper.WindowContent = calendarView;
 		await TestServices.WindowHelper.WaitForLoaded(calendarView);
-		calendarView.Language = "en"; //default culture usually is en
-		await TestServices.WindowHelper.WaitForIdle();
-		calendarView.SetDisplayDate(new DateTimeOffset(new DateTime(2024, 1, 1)));
-		await TestServices.WindowHelper.WaitForIdle();
 		Assert.AreEqual("January 2024", calendarView.TemplateSettings.HeaderText);
 	}
 }
--- a/src/Uno.UWP/Globalization/DateTimeFormatting/DateTimeFormatter.cs
+++ b/src/Uno.UWP/Globalization/DateTimeFormatting/DateTimeFormatter.cs
@@ -802,7 +802,7 @@ public sealed partial class DateTimeFormatter
 				return string.Empty;
 			}
 
-			char singleQuoteChar = '\'';
+			const char singleQuoteChar = '\'';
 			var builder = new StringBuilder();
 			int i = 0;
 			int len = str.Length;
@@ -844,17 +844,16 @@ public sealed partial class DateTimeFormatter
 				}
 				else
 				{
-					char lastChar = currentChar;
 					int count = 1;
 					i++;
 
-					while (i < len && str[i] == lastChar && str[i] != '\'')
+					while (i < len && str[i] == currentChar)
 					{
 						count++;
 						i++;
 					}
 
-					AddToBuilder(builder, lastChar, count);
+					AddToBuilder(builder, currentChar, count);
 				}
 			}
 
--- a/src/Uno.UWP/Globalization/DateTimeFormatting/DateTimeFormatter.cs
+++ b/src/Uno.UWP/Globalization/DateTimeFormatting/DateTimeFormatter.cs
@@ -806,12 +806,20 @@ public sealed partial class DateTimeFormatter
 
 			foreach (var part in IsWithinSingleQuotesRegex().Split(str))
 			{
-				// add quoted literal as is, without the quotes
-				if (part.StartsWith('\''))
+				// skip empty block
+				if (part.Length == 0)
 				{
-					builder.Append(part[1..^1]);
+					continue;
 				}
-				// for non-quoted parts, further segment them by continous character
+				// add quoted literal as is, without the quotes
+				else if (part.StartsWith('\''))
+				{
+					if (part.Length > 2)
+					{
+						builder.Append(part[1..^1]);
+					}
+				}
+				// for non-quoted parts, further segment them by grouping repeated character(s)
 				else
 				{
 					foreach (var segment in DateTimeFormatPartsRegex().EnumerateMatches(part))
@@ -820,6 +828,7 @@ public sealed partial class DateTimeFormatter
 					}
 				}
 			}
+
 			return builder.ToString();
 		}
 	}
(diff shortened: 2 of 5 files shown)


src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_CalendarView.cs
```diff
@@ -303,5 +304,31 @@ public async Task When_NextMonth_InQuickSequence()
...
+	{
+		var calendarView = new CalendarView();
+		TestServices.WindowHelper.WindowContent = calendarView;
+		await TestServices.WindowHelper.WaitForLoaded(calendarView);
+		calendarView.Language = "en"; //default culture usually is en
+		await TestServices.WindowHelper.WaitForIdle();
+		calendarView.SetDisplayDate(new DateTimeOffset(new DateTime(2024, 1, 1)));
+		await TestServices.WindowHelper.WaitForIdle();
+		Assert.AreEqual("January 2024", calendarView.TemplateSettings.HeaderText);
+	}
```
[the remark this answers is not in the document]
ajpinedam: These cases are not in the C#/Dotnet DateFormat in any of the cultures

src/Uno.UWP/Globalization/DateTimeFormatting/DateTimeFormatter.cs
```diff
@@ -797,27 +797,68 @@ void AddToBuilder(StringBuilder builder, char lastChar, int count)
...
+								break;
+							}
+						}
+						else
+						{
+							literal.Append(str[i]);
+							i++;
+						}
+					}
+
```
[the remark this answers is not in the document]
ajpinedam: This will never happen. That is not part of the C#/dotnet culture info datatimeformat

src/Uno.UWP/Globalization/DateTimeFormatting/DateTimeFormatter.cs
```diff
@@ -797,27 +797,68 @@ void AddToBuilder(StringBuilder builder, char lastChar, int count)
...
+						i++;
+					}
+
+					AddToBuilder(builder, lastChar, count);
 				}
 			}
 
-			AddToBuilder(builder, lastChar, count);
 			return builder.ToString();
 		}
```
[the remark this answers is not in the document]
ajpinedam: added already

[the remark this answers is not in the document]
ajpinedam: will add later. Thanks

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_CalendarView.cs
Xiaoy312, line 332: wouldnt it better to setup everything, and then wait:
```suggestion
	public async Task When_Spanish_Language()
	{
		var calendarView = new CalendarView();
		calendarView.Language = "es-ES";
		calendarView.SetDisplayDate(new DateTimeOffset(new DateTime(2024, 1, 1)));
		
		TestServices.WindowHelper.WindowContent = calendarView;
		await TestServices.WindowHelper.WaitForLoaded(calendarView);

		Assert.AreEqual("enero de 2024", calendarView.TemplateSettings.HeaderText);
	}

	[TestMethod]
	public async Task When_English_Language()
	{
		var calendarView = new CalendarView();
		calendarView.Language = "en"; // default culture usually is en
		calendarView.SetDisplayDate(new DateTimeOffset(new DateTime(2024, 1, 1)));

		TestServices.WindowHelper.WindowContent = calendarView;
		await TestServices.WindowHelper.WaitForLoaded(calendarView);
		
		Assert.AreEqual("January 2024", calendarView.TemplateSettings.HeaderText);
	}
```

src/Uno.UWP/Globalization/DateTimeFormatting/DateTimeFormatter.cs
```diff
@@ -797,27 +797,68 @@ void AddToBuilder(StringBuilder builder, char lastChar, int count)
 
 		string ConstructPattern(string str)
 		{
+			if (string.IsNullOrEmpty(str))
+			{
+				return string.Empty;
+			}
+
+			char singleQuoteChar = '\'';
```
Xiaoy312:
```suggestion
			const char SingleQuoteChar = '\'';
```

src/Uno.UWP/Globalization/DateTimeFormatting/DateTimeFormatter.cs
```diff
@@ -797,27 +797,68 @@ void AddToBuilder(StringBuilder builder, char lastChar, int count)
...
+					builder.Append(literal);
 				}
 				else
 				{
-					count++;
+					char lastChar = currentChar;
+					int count = 1;
+					i++;
+
+					while (i < len && str[i] == lastChar && str[i] != '\'')
```
Xiaoy312:
```suggestion
					while (i < len && str[i] == lastChar && str[i] != SingleQuoteChar)
```

src/Uno.UWP/Globalization/DateTimeFormatting/DateTimeFormatter.cs
```diff
@@ -797,27 +797,68 @@ void AddToBuilder(StringBuilder builder, char lastChar, int count)
...
 
-					lastChar = str[i];
-					count = 1;
+					while (i < len)
+					{
+						if (str[i] == singleQuoteChar)
+						{
+							if (i + 1 < len && str[i + 1] == singleQuoteChar)
+							{
+								literal.Append(singleQuoteChar);
```
Xiaoy312: here we have the case for triple `'''` in a row, which you add as a `'`, this doesnt sound right
according to https://learn.microsoft.com/en-us/dotnet/standard/base-types/custom-date-and-time-format-strings, it would be open-close-empty-literal and open-literal...
the only way to insert quote (both single or double) would be to use `\`-escape or weave in another different quote-sign literal (`'"'` or `"'"` d-quote between s-quote or vise-versa)

just test with `DateTime.Now.ToString("MMMM '''''''''de' yyyy")` it should produces `December de 2025`
but this formatter produces the intermediate value of: `{month.full} ''''de {year.full}`

src/Uno.UWP/Globalization/DateTimeFormatting/DateTimeFormatter.cs
```diff
@@ -797,27 +797,68 @@ void AddToBuilder(StringBuilder builder, char lastChar, int count)
...
+							i++;
+						}
+					}
+
+					builder.Append(literal);
 				}
 				else
 				{
-					count++;
+					char lastChar = currentChar;
```
Xiaoy312: kinda useless when not mutating `currentChar`, or re-assigning `lastChar`
```suggestion
```

src/Uno.UWP/Globalization/DateTimeFormatting/DateTimeFormatter.cs
Xiaoy312: we are within the else block of
```cs
if (currentChar == singleQuoteChar) // L822
```
there is no need to check `str[i] is not SingleQuoteChar`, since you just confirmed `str[i] == lastChar/*aka currentChar*/` which cant be single quote

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_CalendarView.cs
ajpinedam: Other tests do it this way, did not confirmed if this is required to be in the UI thread. Confirming and will let you know

src/Uno.UWP/Globalization/DateTimeFormatting/DateTimeFormatter.cs
ajpinedam: Understand, but these values (the format) are not randomly generated by the user, but they will come from the standard C# CultureInfo Format:

https://github.com/unoplatform/uno/blob/master/src/Uno.UWP/Globalization/DateTimeFormatting/DateTimeFormatter.cs#L479

So the chance of having a ''' is close to zero.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_CalendarView.cs
ajpinedam: It worked. Updated the tests with your suggestions

src/Uno.UWP/Globalization/DateTimeFormatting/DateTimeFormatter.cs
Xiaoy312: let me propose this func:
```
	foreach (var part in Regex.Split(str, @"('[^']+')"))
	{
		// add quoted literal as is, without the quotes
		if (part.StartsWith('\''))
		{
			builder.Append(part[1..^1]);
		}
		// for non-quoted parts, further segment them by repeating character(s)
		else
		{
			foreach (var segment in Regex.EnumerateMatches(part, @"(.)(\1+)?"))
			{
				AddToBuilder(builder, part[segment.Index], segment.Length);
			}
		}
	}
	
	return builder.ToString();
```

jeromelaban: The regexes can work, but only if we make it generated ones.

ajpinedam: pushed changes with generated Regexs

src/Uno.UWP/Globalization/DateTimeFormatting/DateTimeFormatter.cs
```diff
@@ -797,27 +797,36 @@ void AddToBuilder(StringBuilder builder, char lastChar, int count)
...
 				}
 			}
-
-			AddToBuilder(builder, lastChar, count);
 			return builder.ToString();
 		}
 	}
 
+	[GeneratedRegex(@"('[^']+')")]
+	private static partial Regex IsWithinSingleQuotesRegex();
```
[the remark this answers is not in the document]
Xiaoy312: good point, but the both regex (not as suggested) and the outter loop impl needs to be updated

src/Uno.UWP/Globalization/DateTimeFormatting/DateTimeFormatter.cs
Xiaoy312: here is the updated impl that handles empty quote
```cs
	// naive implementation that only escapes single-quoted literal
	foreach (var part in IsWithinSingleQuotesRegex().Split(str))
	{
		// skip empty block
		if (part.Length == 0)
		{
		}
		// add quoted literal as is, without the quotes
		else if (part.StartsWith('\''))
		{
			if (part.Length > 2)
			{
				builder.Append(part[1..^1]);
			}
		}
		// for non-quoted parts, further segment them by grouping repeated character(s)
		else
		{
			foreach (var segment in DateTimeFormatPartsRegex().EnumerateMatches(part))
			{
				AddToBuilder(builder, part[segment.Index], segment.Length);
			}
		}
	}
...
private static Regex IsWithinSingleQuotesRegex() => new Regex(@"('[^']*')"); // note: edit this to use the generator syntax
```

Youssef1313 (commented):
It's best IMO to have an additional direct unit test for DateTimeFormatter specifically to show what the original bug in DateTimeFormatter is, and to ease investigation if this broke in the future.

src/Uno.UWP/Globalization/DateTimeFormatting/DateTimeFormatter.cs
```diff
@@ -797,27 +797,45 @@ void AddToBuilder(StringBuilder builder, char lastChar, int count)
...
+			{
+				return string.Empty;
+			}
+
 			var builder = new StringBuilder();
-			char lastChar = str[0];
-			int count = 1;
-			for (int i = 1; i < str.Length; i++)
+
+			foreach (var part in IsWithinSingleQuotesRegex().Split(str))
```
Youssef1313: If I get it correctly, this can be done much more easily without regex.
i.e, keep a track of boolean variable e.g, `bool isInLiteral = false`
Then when looping, if `isInLiteral` is true, just append directly to the builder without calling `AddToBuilder`.

And when you encounter a `'`, flip the `isInLiteral`.

Youssef1313 (commented):
For future reference on what's going on here, `new CultureInfo("es-ES").DateTimeFormat.YearMonthPattern` produces `MMMM 'de' yyyy`.

The old logic used to replace the `d` with the day, and keeps the quotes.

ajpinedam:
> For future reference on what's going on here, `new CultureInfo("es-ES").DateTimeFormat.YearMonthPattern` produces `MMMM 'de' yyyy`.
> 
> The old logic used to replace the `d` with the day, and keeps the quotes.

Right. That's exactly the issue. Documented this on the original issue/discussion.

src/Uno.UWP/Globalization/DateTimeFormatting/DateTimeFormatter.cs
Xiaoy312: true, that was the initial plan
but since we did have to handle both the quoted state and extracting repeating char in the same loop
i gave up after wasting 30min on this, and went back to the "good ol' reliable" regex

src/Uno.UWP/Globalization/DateTimeFormatting/DateTimeFormatter.cs
```diff
@@ -797,27 +797,45 @@ void AddToBuilder(StringBuilder builder, char lastChar, int count)
...
-					AddToBuilder(builder, lastChar, count);
-
-					lastChar = str[i];
-					count = 1;
+					continue;
+				}
+				// add quoted literal as is, without the quotes
+				else if (part.StartsWith('\''))
+				{
+					if (part.Length > 2)
```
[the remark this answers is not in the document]
ajpinedam: We don't need to capture the empty string

src/Uno.UWP/Globalization/DateTimeFormatting/DateTimeFormatter.cs
```diff
@@ -797,27 +797,45 @@ void AddToBuilder(StringBuilder builder, char lastChar, int count)
...
+					}
 				}
 			}
 
-			AddToBuilder(builder, lastChar, count);
 			return builder.ToString();
 		}
 	}
 
+	[GeneratedRegex(@"('[^']+')")]
```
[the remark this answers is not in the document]
ajpinedam: Thanks, but this is exactly what we want

Xiaoy312 (approved):

PR #22071, opened 2025-12-09
chore: handle release branches for auto docs publish

kazo0:
This pull request enhances the external documentation updater workflow to better support release branches and clarifies its behavior in both code and documentation. The main improvement is that when the updater is run against a `release/stable/*` branch, it will now attempt to update each repository to the latest matching `release/stable/*` branch (by version) if available, falling back to the default branch otherwise. The workflow documentation and commit message have also been updated to clearly explain this behavior.

**Branch selection logic improvements:**

* The updater script (`.github/actions/docs-update/action.yml`) now detects if the workflow is running against a `release/stable/*` branch and, for each repository, attempts to use the latest `release/stable/*` branch by version as the source. If no such branch exists, it falls back to the repository's default branch. [[1]](diffhunk://#diff-101b50ad773a292d834f0ce2351f91bc07e8c5dc6ad8cb418a71d0ee1c00c107R47-R56) [[2]](diffhunk://#diff-101b50ad773a292d834f0ce2351f91bc07e8c5dc6ad8cb418a71d0ee1c00c107L65-R127)

**Documentation and workflow updates:**

* The workflow commit message and body (`.github/workflows/docs-updater.yml`) have been updated to clearly describe the new branch selection behavior for both default and release branch runs.


.github/actions/docs-update/action.yml
```diff
@@ -61,23 +71,60 @@ runs:
...
+                $targetBranch = $defaultBranch
+                $branchDescription = $defaultBranch
+
+                if ($useReleaseBranches) {
+                    # Try to locate the latest release/stable/* branch for this repo
+                    $branchesUrl = "https://api.github.com/repos/unoplatform/$repo/branches?per_page=100"
+                    $branches = Invoke-RestMethod -Uri $branchesUrl -Headers @{
+                        "Accept" = "application/vnd.github.v3+json"
+                        "User-Agent" = "Uno-Docs-Updater"
+                    }
```
jeromelaban: We have a lot of branches, it might not be enough.

Kunal22shah (approved):

PR #22161, opened 2025-12-17
Dev/xygu/20251216/hotreload-fwktemplate-leak

Xiaoy312:
**GitHub Issue:** closes unoplatform/uno-private#1670

## PR Type:
- 🐞 Bugfix
- 🔄 Refactoring (no functional changes, no api changes)

1. FrameworkTemplate can leak if its view-factory method isn't static
2. FrameworkTemplatePool still creates an entry for every data-template even if it is disabled

1. FrameworkTemplate nows holds its view-factory with the new weak-delegate
2. FrameworkTemplatePool wont create any entry when disabled

## PR Checklist ✅

## Other information ℹ️
view-factory method are usually generated with `static` modifier, except when hot-reload is enable (which implies in DEBUG only)

### DIFF
--- a/src/Uno.UI/Helpers/WeakDelegate.cs
+++ b/src/Uno.UI/Helpers/WeakDelegate.cs
@@ -1,0 +1,67 @@
+﻿#nullable enable
+
+using System;
+using System.Reflection;
+
+namespace Uno.UI.Helpers;
+
+internal class WeakDelegate
+{
+	public static WeakDelegate<TDelegate>? Create<TDelegate>(TDelegate? d)
+		where TDelegate : Delegate
+	{
+		return d is { } ? new(d) : null;
+	}
+}
+
+/// <summary>
+/// Represents a delegate that references its <see cref="Delegate.Target"/> weakly, allowing the target to be eventually garbage collected.
+/// </summary>
+/// <remarks>Use <see cref="WeakDelegate{TDelegate}"/> to hold a reference to a delegate without preventing its
+/// target object from being collected by the garbage collector. This is useful for scenarios where you want to
+/// avoid memory leaks caused by strong references. Only delegates with a single target are supported;
+/// multicast delegates are not allowed.</remarks>
+/// <typeparam name="TDelegate">The type of delegate to be referenced. Must derive from <see cref="Delegate"/>.</typeparam>
+internal class WeakDelegate<TDelegate> where TDelegate : Delegate
+{
+	public WeakReference? Instance { get; init; }
+	public MethodInfo Method { get; init; }
+
+	// static method delegate doesn't capture any target instance, so we can just reuse it
+	private TDelegate? _staticDelegate;
+
+	public WeakDelegate(TDelegate d)
+	{
+		if (!d.HasSingleTarget)
+		{
+			throw new NotImplementedException("Multi-cast delegate not supported");
+		}
+
+		if (d.Target is { } t)
+		{
+			Instance = new WeakReference(t);
+			Method = d.Method;
+		}
+		else
+		{
+			_staticDelegate = d;
+			Method = d.Method; // still used outside of this class for logging
+		}
+	}
+
+	public object? Target => Instance?.Target;
+
+	/// <summary>
+	/// Gets the delegate instance that targets the referenced object and method,
+	/// may be null if the target is no longer available.
+	/// </summary>
+	/// <remarks>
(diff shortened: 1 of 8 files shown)


Xiaoy312:
todo: tests

src/Uno.UI/Helpers/WeakDelegate.cs
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	public object? Target => Instance?.Target;
+
+	/// <summary>
+	/// Gets the delegate instance that targets the referenced object and method,
+	/// may be null if the target is no longer available.
+	/// </summary>
+	/// <remarks>
+	/// DO NOT cache the returned delegate, as it keep a strong reference to the target like the original delegate.
```
Xiaoy312:
```suggestion
	/// DO NOT cache the returned delegate, as it keeps a strong reference to the target just like the original delegate.
```

src/Uno.UI/UI/Xaml/FrameworkTemplatePool.cs
```diff
@@ -273,6 +272,11 @@ internal int GetPooledTemplatesCount()
 				info.PooledInstances.RemoveAt(position);
 
 				if (_trace.IsEnabled)
+				{
+					_trace.WriteEventActivity(TraceProvider.ReuseTemplate, EventOpcode.Send, new[] { instance.GetType().ToString() });
+				}
+
+				if (this.Log().IsEnabled(Uno.Foundation.Logging.LogLevel.Debug))
```
Xiaoy312, line 279: just restoring old code that was here, which i messed in #17645

src/Uno.UI/UI/Xaml/FrameworkTemplatePool.cs
```diff
@@ -395,7 +405,8 @@ private void OnParentChanged(object instance, object? key, DependencyObjectParen
 
 		private void TryReuseTemplateRoot(object instance, object? key, object? newParent, bool shouldCleanUpTemplateRoot)
 		{
-			if (!_isPoolingEnabled)
+			if (!_isPoolingEnabled ||
+				TryGetTemplatePool(key as FrameworkTemplate ?? throw new InvalidOperationException($"Received {key} but expecting {typeof(FrameworkTemplate)}")) is not { } info)
```
Xiaoy312: just move it back, with a if-null-return

src/Uno.UI/Helpers/WeakDelegate.cs
```diff
@@ -0,0 +1,58 @@
...
+	/// </remarks>
+	public TDelegate? Delegate => Instance switch
+	{
+		// for instanced method delegate, only try to create if the instance reference is still alive
+		{ Target: { } t } => System.Delegate.CreateDelegate(typeof(TDelegate), t, Method),
```
dr1rrb: We are re-creating delegate on each access ... we need to validate the performance of this! (Not sure if the `CreateDelegate` involves reflection).

src/Uno.UI/Helpers/WeakDelegate.cs
```diff
@@ -0,0 +1,58 @@
...
+		{ } => null,
+
+		// for static method delegate, just create it directly
+		null => System.Delegate.CreateDelegate(typeof(TDelegate), null, Method)
```
dr1rrb: If it's a static method (which will be the case in our case in release), we should just keep the instance of the original delegate instead or re-creating it (on each access!), so it will have 0 perf impact.

src/Uno.UI/Helpers/WeakDelegate.cs
```diff
@@ -0,0 +1,58 @@
...
+	} as TDelegate;
```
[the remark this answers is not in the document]
Xiaoy312: signature is guaranteed by TDelegate

src/Uno.UI/Helpers/WeakDelegate.cs
Xiaoy312: invoking a delegate is quite fast, almost comparable to direct call
that said, since we are rebuilding it on every invocation, there is a concern here
it would argue that, for its current sole usage, template builder, the bulk of work would be in the builder itself.
and for now, we can rest assured that this wont impact RELEASE with static delegate being cached

ajpinedam (approved):

MartinZikmund:
@mergifyio backport release/stable/6.5

PR #22289, opened 2026-01-07
`OnBackPressedCallback` support for Android 16+

### LINKED ISSUE #11748: Add built-in support for `OnBackPressedCallback` on Android

Currently we use the `OnBackPressed` handler on Android, which is now deprecated. Instead `OnBackPressedCallback` should be used, which however has a different handling approach - when it is registered, you are required to handle the back press (so there is no way to decide whether the back press will be handled when it is already being executed). We will probably need some new API for this to let user register/unregister the callback on demand - and something which will also work as a replacement of `SystemNavigationManager` on other targets (e.g. WinAppSDK, where `SystemNavigationManager` does not exist).

It seems the Application Activity's `DispatchKeyEvent` for the back button is triggered *before* the callback, which could be an opportunity to trigger the `SystemNavigationManager` and handle it there, but it is not going to be able to handle the back gesture probably.

Create a new API that will fit this new approach of explicitly registering a callback when the app can go back in-app.

## For which Platform:

- [ ] WebAssembly
- [x] Android
- [ ] iOS
- [ ] macOS (AppKit)
- [ ] Mac Catalyst
- [ ] Skia
  - [ ] WPF
  - [ ] GTK (Linux)
  - [ ] Linux Framebuffer
  - [ ] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

MartinZikmund:
**GitHub Issue:** closes https://github.com/unoplatform/uno/issues/11748

On Android 16+ `OnBackPressed` is no longer called. Instead `OnBackPressedCallback` needs to be used, but it has different semantics.

- `OnBackPressedCallback` supported
- Documentation added

## PR Checklist ✅

### DIFF
--- a/doc/articles/features/hardware-back-button.md
+++ b/doc/articles/features/hardware-back-button.md
@@ -29,7 +29,7 @@ private void OnBackRequested(object sender, BackRequestedEventArgs e)
 ```
 
 > [!NOTE]
-> On Android 16+ (API level 36+), the behavior differs. See the [Android 16+ behavior](#android-16-behavior) section below for details.
+> On Android 16+ (API level 36+), the behavior differs. See the [Android 16+ behavior](#android-16+-behavior) section below for details.
 
 When `Handled` is set to `true`, the OS will not continue processing the request. If not set or set to `false`, the OS will navigate away from the application.
 
--- a/doc/articles/features/hardware-back-button.md
+++ b/doc/articles/features/hardware-back-button.md
@@ -32,16 +32,16 @@ When `Handled` is set to `true`, the OS will not continue processing the request
 
 Make sure to unsubscribe from the event when no longer needed.
 
-## Android 15+ behavior
+## Android 16+ behavior
 
-Starting with Android 15 (API level 35), the back navigation behavior has changed due to the [predictive back gesture](https://developer.android.com/guide/navigation/custom-back/predictive-back-gesture) feature. On these versions, the `Handled` property is ignored, and the subscription state of the `BackRequested` event determines whether the app handles back navigation:
+Starting with Android 16 (API level 36), the back navigation behavior has changed due to the [predictive back gesture](https://developer.android.com/guide/navigation/custom-back/predictive-back-gesture) feature. On these versions, the `Handled` property is ignored, and the subscription state of the `BackRequested` event determines whether the app handles back navigation:
 
 - **When subscribed**: The app is assumed to handle back navigation. Back button presses are consumed by the app.
 - **When unsubscribed**: The system handles back navigation (e.g., exits the app or navigates back in the task stack).
 
-### Best practices for Android 15+
+### Best practices for Android 16+
 
-To ensure proper behavior on Android 15+, subscribe to `BackRequested` only when your app can handle back navigation, and unsubscribe when it cannot. A common pattern is to subscribe on `Loaded` and unsubscribe on `Unloaded`:
+To ensure proper behavior on Android 16+, subscribe to `BackRequested` only when your app can handle back navigation, and unsubscribe when it cannot. A common pattern is to subscribe on `Loaded` and unsubscribe on `Unloaded`:
 
 ```csharp
 public MyPage()
@@ -67,7 +67,7 @@ private void OnBackRequested(object sender, BackRequestedEventArgs e)
     if (this.Frame.CanGoBack)
     {
         this.Frame.GoBack();
-        // On Android 15+, the Handled property is ignored.
+        // On Android 16+, the Handled property is ignored.
         // The subscription itself indicates the app handles back navigation.
         e.Handled = true;
     }
@@ -75,4 +75,4 @@ private void OnBackRequested(object sender, BackRequestedEventArgs e)
 ```
 
 > [!NOTE]
-> On Android versions prior to 15, the `Handled` property continues to work as expected. Setting it to `true` prevents the OS from processing the back request, while `false` allows the OS to handle it.
+> On Android versions prior to 16, the `Handled` property continues to work as expected. Setting it to `true` prevents the OS from processing the back request, while `false` allows the OS to handle it.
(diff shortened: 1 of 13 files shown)


src/Uno.UI/UI/Xaml/ApplicationActivity.BackPressed.Android.cs
```diff
@@ -0,0 +1,59 @@
...
+	private void InitializeBackPressedCallback()
+	{
+		// On Android 15+ (API 35+), use OnBackPressedCallback for predictive back gesture support.
+		// The callback is enabled/disabled based on BackRequested subscription state.
+		if ((int)Build.VERSION.SdkInt >= 35)
+		{
+			_backPressedCallback = new UnoOnBackPressedCallback(
+				enabled: SystemNavigationManager.GetForCurrentView().HasBackRequestedSubscribers);
+			OnBackPressedDispatcher.AddCallback(this, _backPressedCallback);
+			SystemNavigationManager.BackRequestedSubscribersChanged += OnBackRequestedSubscribersChanged;
```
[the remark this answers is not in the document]
morning4coffe-dev: I think this is a valid point

morning4coffe-dev (approved):
The sub to BackRequested now implicitly marks the event as handled on A16+, regardless of whether the user sets args.Handled = true, do we/should we count with the possibility of user setting it to false? I think it would make the behavior closer to... well UWP 😅.

doc/articles/features/hardware-back-button.md
```diff
@@ -31,3 +31,48 @@ private void OnBackRequested(object sender, BackRequestedEventArgs e)
 When `Handled` is set to `true`, the OS will not continue processing the request. If not set or set to `false`, the OS will navigate away from the application.
```
MartinZikmund, line 34: Make sure to make it clear that this is no longer the case on Android 16+

MartinZikmund:
@copilot Address open comments

MartinZikmund:
> The sub to BackRequested now implicitly marks the event as handled on A16+, regardless of whether the user sets args.Handled = true, do we/should we count with the possibility of user setting it to false? I think it would make the behavior closer to... well UWP 😅.

@morning4coffe-dev that's the thing, this is impossible to do on A16+ - on A16 you either handle and then you **must** handle, or you don't...

src/SamplesApp/SamplesApp.Skia.netcoremobile/Android/AndroidManifest.xml
```diff
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android" android:installLocation="auto">
-	<uses-sdk android:minSdkVersion="21" android:targetSdkVersion="35" />
+	<uses-sdk android:minSdkVersion="21" android:targetSdkVersion="36" />
```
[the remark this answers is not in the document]
MartinZikmund: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/unoplatform/uno/pull/22289#discussion_r2668163888)

morning4coffe-dev:
@MartinZikmund The idea I got would be to then programmatically call the termination of the app, if the dev sets handled to false. But testing this with A16, Google Pixel, it works mostly well, the navigation is halted. I found an edge case, when I do it multiple times after eachother, the app sometimes crashes and closes itself anyway. I can try to logcat the error if you won't be able to repro: 

https://github.com/user-attachments/assets/9934797d-122b-4e0b-b336-ee958e14fb51

Update: this is related to pointers, fix will be in a different PR, not related to those changes.

jeromelaban (approved):

ajpinedam (approved):

doc/articles/features/hardware-back-button.md
```diff
@@ -28,6 +28,53 @@ private void OnBackRequested(object sender, BackRequestedEventArgs e)
...
+
+private void OnBackRequested(object sender, BackRequestedEventArgs e)
+{
+    if (this.Frame.CanGoBack)
+    {
+        this.Frame.GoBack();
+        // The subscription itself indicates the app handles back navigation.
+        // So setting e.Handled has no effect and is ignored.
+    }
+}
```
Xiaoy312, line 76: This shouldnt be handled at the page level... We should handle it on a global/app level:
```suggestion
public partial class App : Application
{
	protected Window? MainWindow { get; private set; }
	
	public App()
	{
		this.InitializeComponent();

		SystemNavigationManager.GetForCurrentView().BackRequested += OnBackRequested;
	}

	private void OnBackRequested(object sender, BackRequestedEventArgs e)
	{
		if (MainWindow?.Content is Frame { CanGoBack: true } frame)
		{
			frame.GoBack();
		}
	}
}
```

doc/articles/features/hardware-back-button.md
```diff
@@ -28,6 +28,53 @@ private void OnBackRequested(object sender, BackRequestedEventArgs e)
...
+    SystemNavigationManager.GetForCurrentView().BackRequested -= OnBackRequested;
+}
+
+private void OnBackRequested(object sender, BackRequestedEventArgs e)
+{
+    if (this.Frame.CanGoBack)
+    {
+        this.Frame.GoBack();
+        // The subscription itself indicates the app handles back navigation.
+        // So setting e.Handled has no effect and is ignored.
```
Xiaoy312, line 74: if setting e.Handled has no effect... then whats the point? if it meant is for communication between multiple subscribers of this event, we should clarify that.

also, we should be able to give control back to the OS-back-navigation, like when the frame navigation stack is empty

doc/articles/features/hardware-back-button.md
MartinZikmund: That's the problem - in case of Android 16+, it has to be handled on page level/or in some navigation service, because it has to be attached only in case the frame can go back and always immediately unsusbscribed after...

doc/articles/features/hardware-back-button.md
MartinZikmund: Yes, in case of Android 16+ there is no effect, because when the app is subscribed to the event, it has to **always** handle it. It is a quite annoying change, but that's how they did it... https://developer.android.com/guide/navigation/custom-back/predictive-back-gesture

PR #22372, opened 2026-01-15
fix: UIElement Event Registration under NativeAOT

jonpryor:
Context: 559b1f7866028a1d23f74ba43c3e8614a8e9f284
Context: https://github.com/unoplatform/uno-private/issues/1697#issuecomment-3752449412

(Repetition is fun!)

While running the uno.chefs app on macOS under NativeAOT:

	dotnet publish -c Release -r osx-x64 -f net10.0-desktop -p:TargetFrameworkOverride=net10.0-desktop \
	  -bl Chefs/Chefs.csproj -p:SelfContained=true -p:UseSkiaRendering=true\
	  -p:PublishAot=true -p:IsAotCompatible=true \
	  -p:IlcGenerateMapFile=true -p:IlcGenerateMstatFile=true -p:IlcGenerateDgmlFile=true -p:IlcGenerateMetadataLog=true \
	  -p:EmitCompilerGeneratedFiles=true -p:CompilerGeneratedFilesOutputPath=`pwd`/_gen
	Chefs/bin/Release/net10.0-desktop/osx-x64/publish/Chefs

while using Uno 6.5.0-dev.104 (and later), the Login page doesn't work properly.  Specifically, after launching the app, click **Skip** to get to the Login page, and:

  * The `Username` field can be selected, but text cannot be entered.
  * The `Password` field cannot be selected via the mouse, but you can tab to it from the `Password` field.
  * You cannot enter text into the `Password` field.
  * You cannot check the `[ ] Remember me` checkbox.

You *can* click the **Sign in with Apple** and **Sign in with Google** buttons.

Why don't the `<TextBox/>` or `<PasswordBox/>` controls work?

Given that this is only under NativeAOT, the obvious answer is "Reflection!", but from where?

The less obvious answer is *event registration*, specifically the interaction between `Control.SubscribeToPostKeyDown()` and `UIElement.GetIsEventOverrideImplemented(Type, string, Type[])`:

	partial class Control {
	  [UnconditionalSuppressMessage("Trimming", "IL2072", Justification = "Types manipulated here have been marked earlier")]
	  private void SubscribeToPostKeyDown() {
	    if (GetIsEventOverrideImplemented(GetType(), nameof(OnPostKeyDown), _keyArgsType))
	      PostKeyDown += OnPostKeyDownHandler;
	  }
	}
	partial class UIElement {
	  private protected static bool GetIsEventOverrideImplemented(
	    [DynamicallyAccessedMembers(…)] Type type, string name, Type[] args)
	  {
	    var method = type.GetMethod(name, BindingFlags.NonPublic | BindingFlags.Instance, null, args, null);
	    return method != null && method.IsVirtual
	      && method.DeclaringType != typeof(UIElement)
	      && method.DeclaringType != typeof(Control);
	  }
	}

In order for text input to work, the type must subscribe to the `PostKeyDown` event.  In order for e.g. `<TextBox/>` to subscribe to the `PostKeyDown` event, it must override the `OnPostKeyDown()` method, *and* `Control` must *determine* that `OnPostKeyDown()` was overridden.  This override check involves Reflection by way of `Type.GetMethod()`, which in turn requires that `OnPostKeyDown` be included in "reflection metadata."

It was not:

	% grep OnPostKeyDown Chefs/obj/Release/net10.0-desktop/osx-x64/native/Chefs.metadata.csv
	3432e2bf, ConstantStringValue, "OnPostKeyDown", ""

We need *more* than `ConstantStringValue` for it be useful. For example, if we add `[DynamicDependency]` to `OnPostKeyDown()`:

	partial class TextBox {
	  [DynamicDependency(nameof(OnPostKeyDown))]
	  private protected override void OnPostKeyDown(KeyRoutedEventArgs args) => …;
	}

that *fixes* the "keyboard input doesn't work" problem, and `Chefs.metadata.csv` contains:

	% grep OnPostKeyDown Chefs/obj/Release/net10.0-desktop/osx-x64/native/Chefs.metadata.csv
	5009119c, Method, "[HasThis]  System.Void OnPostKeyDown(Microsoft.UI.Xaml.Input.KeyRoutedEventArgs)", "34117ead 56117ebb 6202258a 42117ec7"
	34117ead, ConstantStringValue, "OnPostKeyDown", ""
	42117ec7, CustomAttribute, "System.Diagnostics.CodeAnalysis.DynamicDependencyAttribute.[HasThis]  System.Void .ctor(System.String)(\"OnPostKeyDown\")(ctor: Internal.Metadata.NativeFormat.Handle", "6c0aa81f 34117ead"

However, this is not a scalable answer.  *Lots* of codepaths touch `GetIsEventOverrideImplemented()`, notably
`Control.EvaluateImplementedControlRoutedEvents(Type)`.

Do we really want (or need?) to add every `Control` method to reflection metadata?

It also doesn't fix the inabilty to interract with the checkbox.

Additionally, some parts of System.Reflection *do not require* "reflection metadata".  In particular, it is possible to obtain `MethodInfo` instances *without* using `Type.GetMethod()`, by using *delegates*:

	Action a = () => {};
	MethodInfo method = a.Method;

This does not require "reflection metadata", and works under NativeAOT.

Rework `Control.EvaluateImplementedControlRoutedEvents()` et al to use *delegates* instead of `Type.GetMethod()`:

	partial class UIElement {
	  private protected static bool GetIsEventOverrideImplemented<T>(Action<T> a) =>
	    GetIsEventOverrideImplemented((Delegate)a);
	  private protected static bool GetIsEventOverrideImplemented(Delegate d) =>
	    d.Method.IsVirtual && d.Method.DeclaringType != typeof(UIElement) && d.Method.DeclaringType != typeof(Control);
	}
	partial class Control {
	  private void SubscribeToPostKeyDown() {
	    if (GetIsEventOverrideImplemented(OnPostKeyDown))
	      PostKeyDown += OnPostKeyDownHandler;
	  }
	}

This works because delegates know about method overrides.

This not only works under NativeAOT, but also allows us to remove many `[UnconditionalSuppressMessage]` annotations.

However, in order for this to work, many methods which were previously `static` must now be instance methods, such as
`Control.EvaluateImplementedControlRoutedEvents()`.

For backward compatibility/migration purposes, preserve the previous `static` methods and `[Obsolete]` them, directing code to use the instance method alternatives.

**GitHub Issue:** closes #

## PR Checklist ✅

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
@@ -330,23 +330,21 @@ namespace Microsoft.UI.Xaml.Controls
 #endif
 
 #if !__NETSTD_REFERENCE__
-		[System.Diagnostics.CodeAnalysis.UnconditionalSuppressMessage("Trimming", "IL2072", Justification = "Types manipulated here have been marked earlier")]
 		private void SubscribeToPostKeyDown()
 		{
-			if (GetIsEventOverrideImplemented(GetType(), nameof(OnPostKeyDown), _keyArgsType))
+			if (GetIsEventOverrideImplemented(OnPostKeyDown))
 			{
 				PostKeyDown += OnPostKeyDownHandler;
 			}
 		}
 
-		[UnconditionalSuppressMessage("Trimming", "IL2072", Justification = "Types manipulated here have been marked earlier")]
 		private void SubscribeToOverridenRoutedEvents()
 		{
 			// Overridden Events are registered from constructor to ensure they are
 			// registered first in event handlers.
 			// https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.control.onpointerpressed#remarks
 
-			var implementedEvents = GetImplementedRoutedEventsForType(GetType());
+			var implementedEvents = GetImplementedRoutedEvents();
 
 			if (HasFlag(implementedEvents, RoutedEventFlag.PointerPressed))
 			{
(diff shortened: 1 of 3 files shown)


src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
```diff
@@ -1196,6 +1337,7 @@ protected virtual void OnLostFocus(RoutedEventArgs e) { }
 		private static readonly Type[] _manipInertiaArgsType = new[] { typeof(ManipulationInertiaStartingRoutedEventArgs) };
 		private static readonly Type[] _manipCompletedArgsType = new[] { typeof(ManipulationCompletedRoutedEventArgs) };
 
+		[Obsolete("Use instance EvaluateImplementedControlRoutedEvents()")]
```
ramezgerges: This is not part of any public API, so I would rather just remove the obsolete method instead of keeping it around.

src/Uno.UI/UI/Xaml/UIElement.cs
```diff
@@ -179,6 +211,19 @@ internal static RoutedEventFlag GetImplementedRoutedEventsForType(
...
+
+			if (GetIsEventOverrideImplemented(OnBringIntoViewRequested))
+			{
+				result |= RoutedEventFlag.BringIntoViewRequested;
+			}
+
+			return result;
+		}
+
+		[Obsolete("Use EvaluateImplementedUIElementRoutedEvents() instance method")]
```
ramezgerges: same here

ramezgerges (approved):

src/Uno.UI/UI/Xaml/UIElement.cs
jonpryor: I was concerned -- without checking -- that some other assembly with `[InternalsVisibleTo]` would be using this member.

It doesn't look like any other assembly does, so it feels safe to remove: https://github.com/search?q=org%3Aunoplatform+EvaluateImplementedUIElementRoutedEvents&type=code

src/Uno.UI/UI/Xaml/Controls/Control/Control.cs
jonpryor: I was concerned -- without checking -- that some other assembly with `[InternalsVisibleTo]` would be using this member.

It doesn't look like any other assembly does, so it feels safe to remove: https://github.com/search?q=org%3Aunoplatform%20EvaluateImplementedControlRoutedEvents&type=code

jeromelaban (approved):

PR #22504, opened 2026-01-29
ci: bump Xcode to 26.2, update iOS 18.5, and fix macOS template tests

agneszitte:
## Changes
- Update iOS simulator runtime to 18.5 for macOS-15 runners
- Bump Xcode from 26.0.1 to 26.2 for macOS builds
- Add macOS template-test safeguards: skip Xcode validation and download iOS 26.2 runtime if missing

## Why
- macOS-15 runners removed older runtimes (iOS 17.x/18.4 and iOS 26.0), which broke template tests even after the version bumps.
- The .NET iOS workload still performs strict Xcode checks, so CI needs to bypass validation and ensure a supported simulator runtime is present to keep the templates tests green.

### Refs : 
- https://github.com/actions/runner-images/issues/13392
- https://github.com/actions/runner-images/issues/13570

PR #22521, opened 2026-01-29
fix(ci): Stabilize iOS builds on macOS-15

agneszitte:
## Changes

More adjustments after  https://github.com/unoplatform/uno/pull/22504

- Disable strict Xcode version validation for iOS SamplesApp builds.
## Why

macOS‑15 runners ship Xcode 26.2, which fails strict .NET iOS validation and has removed older runtimes, causing the SamplesApp build to fail.
 
 ### Ref
 
 - https://github.com/actions/runner-images/issues/13570


kazo0 (approved):

agneszitte:
PR reverted with https://github.com/unoplatform/uno/pull/22589

PR #22533, opened 2026-02-02
fix: ios skia textbox arrow text navigation

ajpinedam:
**GitHub Issue:** closes https://github.com/unoplatform/chorus-private/issues/38

## PR Type:

- 🐞 Bugfix

## PR Checklist ✅


src/Uno.UI.Runtime.Skia.AppleUIKit/UI/Xaml/Controls/TextBox/MultilineInvisibleTextBoxDelegate.cs
```diff
@@ -57,6 +70,22 @@ public override bool ShouldChangeText(UITextView textView, NSRange range, string
...
+
+	private void SyncSelectionToTextView(UITextView textView)
+	{
+		if (_textBoxViewExtension.GetTarget()?.Owner.TextBox is { } textBox &&
+		textView.SelectedTextRange is { } selectedRange)
+		{
+			var selectionStart = (int)textView.GetOffsetFromPosition(textView.BeginningOfDocument, selectedRange.Start);
+			var selectionEnd = (int)textView.GetOffsetFromPosition(textView.BeginningOfDocument, selectedRange.End);
+			textBox.Select(selectionStart, selectionEnd - selectionStart);
+		}
```
ramezgerges, line 86: I'm wondering if we can get the "directionality" of the selection as well? i.e. whether end of the selection is at the left or right of the selection. If so, we can use TextBox.SelectInternal instead of TextBox.Select, which can handle "negative directions".

ramezgerges (approved):

src/Uno.UI.Runtime.Skia.AppleUIKit/UI/Xaml/Controls/TextBox/MultilineInvisibleTextBoxDelegate.cs
ajpinedam: I see that the SelectInternal has the same parameters as Select (different implementations though) 

```csharp
internal void SelectInternal(int selectionStart, int selectionLength) { }
```

but that means we could just pass the same values

```csharp
textBox.SelectInternal(selectionStart, selectionEnd - selectionStart);
```

ajpinedam: Never mind, I see what you mean now looking at the implementation of the `SelectInternal()` method. Let me continue investigating.

ajpinedam: @ramezgerges  validated. With the `UITextRange` we cannot tell whether the scroll is to the left or to the right.

>In UIKit's UITextRange:
>
>Start is always the earlier position in the document
>End is always the later position in the document
>These are normalized positions, so you can't determine from them alone which end the user dragged to create the selection.

PR #22715, opened 2026-02-23
docs: add root-cause debugging protocol for agents

agneszitte:
This updates AGENTS guidance to steer agents toward root-cause-first debugging for crash and indexing issues, instead of symptom-first guard patching. It adds mandatory diagnosis and validation evidence protocols so confidence and verification claims are explicit and reproducible.

> [!NOTE]
> We are keeping this guidance in AGENTS.md for now. We can refactor it into an Uno-core-specific skill later if AGENTS.md grows too large after syncing with @jeromelaban 

Note: No related issue (Internal maintenance / Approved by Team member: @agneszitte).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -138,6 +138,85 @@ Auto-generated stubs marked with `[Uno.NotImplemented]` allow compilation but wa
 
 ## Development Workflow
 
+### Root-Cause First Debugging Protocol (MANDATORY)
+
+When fixing crashes, rendering issues, or selection/indexing bugs,
+agents must follow this order:
+
+1. **Reproduce first**
+   - Capture exact repro steps and expected vs actual behavior.
+   - Keep one known-good repro path and rerun it after each meaningful change.
+
+2. **Identify the broken invariant**
+   - Prefer state/lifecycle invariants over symptom-level checks.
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -138,6 +138,85 @@ Auto-generated stubs marked with `[Uno.NotImplemented]` allow compilation but wa
...
+
+1. **Reproduce first**
+   - Capture exact repro steps and expected vs actual behavior.
+   - Keep one known-good repro path and rerun it after each meaningful change.
+
+2. **Identify the broken invariant**
+   - Prefer state/lifecycle invariants over symptom-level checks.
+   - For text/layout pipelines, verify that derived structures
+       (maps, indices, cluster lists, line metadata) are rebuilt from
+       final post-mutation state.
```
ramezgerges: Feels too specific and will probably through agents off.

agneszitte: @ramezgerges what do you suggest here, please?

Would this wording work better?

“Prefer state/lifecycle invariants over symptom-level checks.
For pipelines that derive secondary state, verify those derived structures (for example: maps, indices, caches, or metadata) are rebuilt from final post-mutation state.”

ajpinedam (approved):

PR #22716, opened 2026-02-23
fix(win32): avoid flickering when attaching windows

### LINKED ISSUE #22694: [Skia.Win32] WebView2 flickering, position flash, and airspace due to child HWND hosting
## Current behavior

The Win32 (Skia) backend hosts WebView2 via a child HWND using `CreateCoreWebView2ControllerAsync`. This causes three distinct visual defects:

### 1. White flash on show/hide/recreate

Toggling WebView2 visibility or recreating the control (e.g. navigating between pages, re-templating) produces a visible white flash. This is especially noticeable against dark-themed content. The flash occurs because the child HWND surface is composited by DWM independently from the parent window — when the HWND is shown, its surface contains the default white background for one or more frames before WebView2's multi-process renderer paints real content.

Setting `DefaultBackgroundColor` to a dark color before navigation reduces the contrast but does not eliminate the flash — there is still a frame where the HWND surface contains the default background, not the rendered page content.

### 2. Position flash at screen origin

On WebView2 creation, the HWND briefly appears near the top-left corner of the primary monitor before jumping to its correct position inside the app window. This happens because:

1. [`Win32NativeWebView.cs` creates the HWND](https://github.com/nicknow/AiEditorReferences/blob/main/Win32NativeWebView.cs) with `HWND.Null` as parent and `CW_USEDEFAULT` for position — making it a top-level window at the OS-chosen default position
2. `Win32NativeElementHostingExtension.AttachNativeElement` reparents it via `SetParent(childHwnd, parentHwnd)` on a subsequent frame
3. Between creation and reparenting, DWM composites the HWND at its initial top-level position for one or more vsync cycles

### 3. Airspace (z-order violation)

The WebView2 child HWND always renders on top of Skia content, regardless of the XAML visual tree z-order. Skia renders to the window DC via WGL `SwapBuffers` (DWM Layer 1), while child HWNDs are composited on DWM Layer 3 — which is unconditionally above Layer 1. This means:

- XAML elements that should appear above the WebView2 (popups, tooltips, dropdowns, overlays) are hidden behind it
- The current mitigation (`SetWindowRgn` in `Win32NativeElementHostingExtension`) provides coarse rectangular clipping but cannot handle transparency, rounded corners, or smooth edges

## Expected behavior

- No white flash when showing, hiding, or recreating WebView2
- No position flash — WebView2 content should appear in-place at the correct position
- XAML elements should be able to render on top of WebView2 according to their z-order in the visual tree

## How to reproduce

Minimal Uno app that demonstrates all three issues:

```xml
<!-- MainPage.xaml -->
<Page x:Class="WebView2FlickerRepro.MainPage"
      Background="#1E1E1E">

    <Grid>
        <Grid.RowDefinitions>
            <RowDefinition Height="Auto"/>
            <RowDefinition Height="*"/>
        </Grid.RowDefinitions>

        <StackPanel Orientation="Horizontal" Spacing="8" Padding="8">
            <Button Content="Toggle Visibility" Click="ToggleVisibility_Click"/>
            <Button Content="Recreate" Click="Recreate_Click"/>
            <TextBlock x:Name="StatusText" Foreground="White" VerticalAlignment="Center"/>
        </StackPanel>

        <Grid Grid.Row="1">
            <!-- WebView2 -->
            <WebView2 x:Name="MyWebView"
                       Source="https://example.com"
                       DefaultBackgroundColor="#1E1E1E"/>

            <!-- This border SHOULD render on top of the WebView2 but won't (airspace) -->
            <Border Background="#80FF0000"
                    Width="200" Height="200"
                    HorizontalAlignment="Center"
                    VerticalAlignment="Center"
                    CornerRadius="16">
                <TextBlock Text="I should be on top of WebView2"
                           Foreground="White"
                           HorizontalAlignment="Center"
                           VerticalAlignment="Center"/>
            </Border>
        </Grid>
    </Grid>
</Page>
```

```csharp
// MainPage.xaml.cs
{
    public MainPage()
    {
        InitializeComponent();
    }

    private void ToggleVisibility_Click(object sender, RoutedEventArgs e)
    {
        // Issue 1: Each toggle produces a white flash
        MyWebView.Visibility = MyWebView.Visibility == Visibility.Visible
            ? Visibility.Collapsed
            : Visibility.Visible;
        StatusText.Text = $"Visibility: {MyWebView.Visibility}";
    }

    private void Recreate_Click(object sender, RoutedEventArgs e)
    {
        // Issue 1 + 2: Recreating produces white flash AND position flash at screen origin
        var parent = (Grid)MyWebView.Parent;
        var source = MyWebView.Source;
        parent.Children.Remove(MyWebView);

        var newWebView = new WebView2
        {
            Source = source,
            DefaultBackgroundColor = Windows.UI.Color.FromArgb(255, 30, 30, 30)
        };
        Grid.SetRow(newWebView, 0);
        parent.Children.Insert(0, newWebView);
        MyWebView = newWebView;

        StatusText.Text = "Recreated";
    }
}
```

**Steps to reproduce:**

1. Create a new Uno app targeting `net10.0-desktop` (Win32/Skia backend)
2. Add the XAML and code-behind above
3. Run on Windows 11
4. **Issue 1**: Click "Toggle Visibility" rapidly — observe white flash on each toggle
5. **Issue 2**: Click "Recreate" — observe the WebView2 HWND briefly flash near the top-left of the monitor before appearing in the correct position
6. **Issue 3**: Observe the red semi-transparent overlay — it is hidden behind the WebView2 instead of rendering on top

## Root cause analysis

All three issues stem from hosting WebView2 via `CreateCoreWebView2ControllerAsync` (windowed/HWND mode). This is a fundamental limitation of the Win32 HWND composition model.

### DWM layer model

Windows Desktop Window Manager composites content in four layers:

```
Layer 4: DirectComposition topmost visuals
Layer 3: Child HWNDs          ← WebView2 lives here (always on top of parent)
Layer 2: DirectComposition non-topmost visuals
Layer 1: GDI / OpenGL window DC  ← Skia renders here (WGL SwapBuffers)
```

**Flickering**: When a child HWND is shown or created, DWM composites its surface immediately — but the surface may not yet contain rendered content (WebView2 renders in a separate process). The gap between "surface visible" and "content painted" is the flash.

**Position flash**: The HWND is created as a top-level window (`parent: HWND.Null`) then reparented. DWM composites it at the initial position for at least one frame before `SetParent` moves it.

**Airspace**: Layer 3 is unconditionally above Layer 1. No amount of z-ordering in the XAML tree can make Skia content (Layer 1) render above a child HWND (Layer 3).

### Relevant source files

| File | Role |
|------|------|
| [`Win32NativeWebView.cs`](https://github.com/nicknow/AiEditorReferences/blob/main/Win32NativeWebView.cs) | Creates child HWND with `HWND.Null` parent, calls `CreateCoreWebView2ControllerAsync` (windowed mode) |
| [`Win32NativeElementHostingExtension.cs`](https://github.com/nicknow/AiEditorReferences/blob/main/Win32NativeElementHostingExtension.cs) | Reparents via `SetParent`, clips via `SetWindowRgn` |
| [`Win32WindowWrapper.Rendering.OpenGl.cs`](https://github.com/nicknow/AiEditorReferences/blob/main/Win32WindowWrapper.Rendering.OpenGl.cs) | Skia renders to window DC via WGL `SwapBuffers` (Layer 1) |
| [`RenderSurfaceType.cs`](https://github.com/nicknow/AiEditorReferences/blob/main/RenderSurfaceType.cs) | Renderer enum: only `Software` and `OpenGL` — no DirectComposition option |

### Why this cannot be fixed within the HWND model

The parent process and the browser process (WebView2) paint independently, and DWM composites them independently. There is no Win32 synchronization primitive that atomically makes an HWND visible with content already painted. Workarounds (delayed visibility, `DefaultBackgroundColor`, `WS_EX_COMPOSITED`) reduce severity but cannot eliminate the flash because:

- `DefaultBackgroundColor` only controls the pre-navigation background — the HWND surface is still composited before the browser process paints
- `WS_EX_COMPOSITED` enables GDI double-buffering, which doesn't apply to WGL or WebView2's multi-process renderer
- `BeginDeferWindowPos`/`EndDeferWindowPos` batch position changes but don't control paint timing
- `WM_SETREDRAW` suppresses repaints temporarily but re-enabling it triggers the same flash

## Proposed solution: DirectComposition rendering mode

WinUI does not have these issues because it uses [`CreateCoreWebView2CompositionControllerAsync`](https://learn.microsoft.com/en-us/microsoft-edge/webview2/concepts/windowed-vs-visual-hosting) instead of `CreateCoreWebView2ControllerAsync`. The WebView2 renders into a DirectComposition visual — not a child HWND.

The full fix would add a `DirectComposition` renderer to the Win32 backend where both Skia and WebView2 render into the same DComp visual tree:

```
Win32 Window (WS_EX_NOREDIRECTIONBITMAP)
└── DComp Target
    ├── Skia Visual: SkiaSharp → ANGLE (OpenGL ES → D3D11) → IDCompositionSurface
    └── WebView2 Visual: CoreWebView2CompositionController → RootVisualTarget
```

- **No child HWND** — eliminates position flash and HWND lifecycle flash
- **Atomic composition** — DComp commits are atomic; content appears only when ready
- **True z-ordering** — Skia and WebView2 are sibling DComp visuals, z-ordered by tree position
- **No airspace problem** — XAML elements can render on top of WebView2

### Scope of changes (Win32 backend only)

| Change | File | Description |
|--------|------|-------------|
| Modify | `RenderSurfaceType.cs` | Add `DirectComposition` enum value |
| Modify | `Win32WindowWrapper.cs` | Use `WS_EX_NOREDIRECTIONBITMAP` when DComp renderer is active |
| Modify | `Win32NativeWebView.cs` | Use `CreateCoreWebView2CompositionControllerAsync` + input forwarding via `SendMouseInput` |
| Modify | `Win32NativeElementHostingExtension.cs` | DComp path: create `IDCompositionVisual` instead of child HWND |
| New | `Win32WindowWrapper.Rendering.DirectComposition.cs` | DComp renderer: D3D11 → ANGLE → Skia into `IDCompositionVirtualSurface` |
| New | `AngleEglBridge.cs` | ANGLE EGL display/context management for D3D11↔OpenGL ES bridge |
| New | `DCompInterop.cs` | COM interfaces for `IDCompositionDevice`, `IDCompositionTarget`, `IDCompositionVisual` |
| Dependency | ANGLE native binaries | `libEGL.dll` + `libGLESv2.dll` (~10MB) |

No other platforms (WASM, macOS, Linux, iOS, Android) are affected. The existing `Software` and `OpenGL` renderers remain unchanged.

### Prior art

- **WinUI**: Uses composition mode natively for WebView2 ([`WebView2.cpp` → `CreateCoreObjects()`](https://github.com/nicknow/AiEditorReferences/blob/main/WebView2.cpp))
- **Avalonia**: Renders Skia into DComp surfaces via ANGLE ([`DirectCompositedWindowSurface.cs`](https://github.com/AvaloniaUI/Avalonia/blob/master/src/Windows/Avalonia.Win32/DComposition/DirectCompositedWindowSurface.cs), [`AngleWin32EglDisplay.cs`](https://github.com/AvaloniaUI/Avalonia/blob/master/src/Windows/Avalonia.Win32/OpenGl/Angle/AngleWin32EglDisplay.cs))
- **Wice framework**: .NET implementation of DComp + WebView2 composition ([`WebView.cs`](https://github.com/aelyo-softworks/Wice))
- **WebView2 API Sample**: DComp visual tree setup ([`ViewComponent.cpp`](https://github.com/MicrosoftEdge/WebView2Samples))

### Immediate partial fix (lower effort)

A smaller change that fixes flickering but NOT airspace: use `CreateCoreWebView2CompositionControllerAsync` with a DComp overlay, while Skia continues rendering via WGL. This avoids the ANGLE dependency entirely (~250-400 new lines, zero new binary dependencies) but WebView2 would still always render on top of Skia content (Layer 2 above Layer 1).

## Environment

- Uno.WinUI: 5.x / 6.x (Win32/Skia backend)
- .NET: 9 / 10
- OS: Windows 10/11
- WebView2 Runtime: Evergreen

## Related issues

- #21376 — WebView2 Visibility not working as expected (z-order / airspace symptom)
- #22641 — WebView2 control steals focus from window

ramezgerges:
**GitHub Issue:** closes https://github.com/unoplatform/uno/issues/22694

## PR Checklist ✅

### DIFF
--- a/src/Uno.UI.Runtime.Skia.Win32/Native/Win32NativeElementHostingExtension.cs
+++ b/src/Uno.UI.Runtime.Skia.Win32/Native/Win32NativeElementHostingExtension.cs
@@ -31,6 +31,7 @@ internal class Win32NativeElementHostingExtension : ContentPresenter.INativeElem
 	private Rect _lastArrangeRect;
 	private string? _lastFinalSvgClipPath;
 	private HRGN _lastClipHrgn;
+	private bool _showWindowOnNextArrange;
 
 	public Win32NativeElementHostingExtension(ContentPresenter presenter)
 	{
@@ -85,7 +86,8 @@ internal class Win32NativeElementHostingExtension : ContentPresenter.INativeElem
 		PInvoke.SetWindowLong((HWND)window.Hwnd, WINDOW_LONG_PTR_INDEX.GWL_EXSTYLE, oldExStyleVal | (int)WINDOW_EX_STYLE.WS_EX_LAYERED);
 		PInvoke.SetWindowLong((HWND)window.Hwnd, WINDOW_LONG_PTR_INDEX.GWL_STYLE, (oldStyleVal | (int)WINDOW_STYLE.WS_CLIPSIBLINGS) & ~(int)WINDOW_STYLE.WS_CAPTION); // removes the title bar and borders
 
-		_ = PInvoke.ShowWindow((HWND)window.Hwnd, SHOW_WINDOW_CMD.SW_SHOWNORMAL);
+		_ = PInvoke.ShowWindow((HWND)window.Hwnd, SHOW_WINDOW_CMD.SW_HIDE);
+		_showWindowOnNextArrange = true; // only show the window after the first arrange to avoid the split-second flicker between showing the window and positioning/clipping it correctly.
 
 		var oldParent = PInvoke.SetParent((HWND)window.Hwnd, Hwnd);
 		if (oldParent == HWND.Null && Marshal.GetLastWin32Error() != 0)
@@ -99,12 +101,6 @@ internal class Win32NativeElementHostingExtension : ContentPresenter.INativeElem
 
 	private unsafe void OnRenderingNegativePathReevaluated(object? sender, SKPath path)
 	{
-		if (path != _lastClipPath)
-		{
-			_lastClipPath.Rewind();
-			_lastClipPath.AddPath(path);
-		}
-
 		_tempPath.Rewind();
 		_tempPath.AddRect(_lastArrangeRect.ToSKRect());
 		path.Op(_tempPath, SKPathOp.Intersect, _tempPath);
@@ -321,6 +317,12 @@ internal class Win32NativeElementHostingExtension : ContentPresenter.INativeElem
 
 		_lastFinalSvgClipPath = null; // force reapply clip path after arranging
 		OnRenderingNegativePathReevaluated(this, _lastClipPath);
+
+		if (_showWindowOnNextArrange)
+		{
+			_showWindowOnNextArrange = false;
+			_ = PInvoke.ShowWindow((HWND)window.Hwnd, SHOW_WINDOW_CMD.SW_SHOWNORMAL);
+		}
 	}
 
 	public Size MeasureNativeElement(object content, Size childMeasuredSize, Size availableSize)
--- a/src/Uno.UI/UI/Xaml/Controls/ContentPresenter/ContentPresenter.skia.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/ContentPresenter/ContentPresenter.skia.cs
@@ -69,6 +69,7 @@ partial class ContentPresenter
 			{
 				//If in visual tree, attach immediately. If not, don't attach since Enter will attach later.
 				AttachNativeElement();
+				ArrangeNativeElement();
 			}
 		}
 		else
@@ -87,8 +88,16 @@ partial class ContentPresenter
 		_nativeElementHostingExtension.Value!.AttachNativeElement(Content);
 		_nativeHosts.Add(this);
 		var ct = ((CompositionTarget)Visual.CompositionTarget)!;
-		ct.FrameRendered += ArrangeNativeElement;
-		_frameRenderedDisposable = Disposable.Create(() => ct.FrameRendered -= ArrangeNativeElement);
+		ct.FrameRendered += OnFrameRendered;
+		_frameRenderedDisposable = Disposable.Create(() => ct.FrameRendered -= OnFrameRendered);
+	}
+
+	private void OnFrameRendered()
+	{
+		if (_nativeElementAttached)
+		{
+			ArrangeNativeElement();
+		}
 	}
 
 	partial void DetachNativeElement(object content)
@@ -171,6 +180,7 @@ partial class ContentPresenter
 						// We're detaching the native element as it's no longer in view, but conceptually, it's still in the tree, so IsNativeHost is still true
 						Debug.Assert(host.IsNativeHost);
 						host._nativeElementAttached = false;
+						host._lastArrangeRect = null;
 						host._nativeElementHostingExtension.Value!.DetachNativeElement(host.Content);
 					}
 					else


ajpinedam (approved):

clairernovotny (approved):
The ghost window no longer appears, though there is some focus stealing still. May not be related.

PR #22719, opened 2026-02-24
Bitmap decode size

MartinZikmund:
**GitHub Issue:** closes https://github.com/unoplatform/uno-private/issues/1342

## PR Checklist ✅

### DIFF
--- a/src/Uno.UI.Composition/Composition/FrameProviderFactory.skia.cs
+++ b/src/Uno.UI.Composition/Composition/FrameProviderFactory.skia.cs
@@ -13,6 +13,9 @@ internal static class FrameProviderFactory
 		=> new SingleFrameProvider(image);
 
 	public static bool TryCreate(SKManagedStream stream, Action onFrameChanged, [NotNullWhen(true)] out IFrameProvider? provider)
+		=> TryCreate(stream, onFrameChanged, null, null, out provider);
+
+	public static bool TryCreate(SKManagedStream stream, Action onFrameChanged, int? targetWidth, int? targetHeight, [NotNullWhen(true)] out IFrameProvider? provider)
 	{
 		using var codec = SKCodec.Create(stream);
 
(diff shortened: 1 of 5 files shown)


src/Uno.UI.Composition/Composition/FrameProviderFactory.skia.cs
```diff
@@ -78,10 +105,132 @@ public static bool TryCreate(SKManagedStream stream, Action onFrameChanged, [Not
...
+		if (targetWidth is <= 0) targetWidth = null;
+		if (targetHeight is <= 0) targetHeight = null;
+
+		var codecWidth = codec.Info.Width;
+		var codecHeight = codec.Info.Height;
+
+		if (targetWidth is null && targetHeight is null)
+		{
+			return new SKSizeI(codecWidth, codecHeight);
+		}
```
Xiaoy312, line 147:
```suggestion
		var codecWidth = codec.Info.Width;
		var codecHeight = codec.Info.Height;
		
		// Normalize non-positive dimensions to null so downstream branches
		// only deal with genuinely positive values.
		if (targetWidth is <= 0) targetWidth = null;
		if (targetHeight is <= 0) targetHeight = null;
		if (targetWidth is null && targetHeight is null)
		{
			return new SKSizeI(codecWidth, codecHeight);
		}
```

src/Uno.UI.Composition/Composition/FrameProviderFactory.skia.cs
```diff
@@ -78,10 +105,132 @@ public static bool TryCreate(SKManagedStream stream, Action onFrameChanged, [Not
...
+
+		if (targetWidth is > 0 && targetHeight is > 0)
+		{
+			targetCodecW = swaps ? targetHeight.Value : targetWidth.Value;
+			targetCodecH = swaps ? targetWidth.Value : targetHeight.Value;
+		}
+		else if (targetWidth is > 0)
+		{
+			int targetDisplayW = targetWidth.Value;
+			int targetDisplayH = (int)Math.Max(1, (long)displayHeight * targetDisplayW / displayWidth);
```
Xiaoy312: `displayHeight` is already swapped from above
yet we are multiplying it to unswapped value `targetDisplayW`(aka `targetWidth`)
thats fine if we ends here(which means they are all swapped), but this value is further swapped again in the next lines
is that even correct?

i think we should either compute with all swapped values, or
compute with pre-swapped (need to swap target values to the same coordinate-system first) values and swap the final result

Xiaoy312 (approved):

PR #22791, opened 2026-03-06
fix(docs): add docs menu active visual cue

agneszitte:
closes https://github.com/unoplatform/uno-website/issues/1500

Adds a Docs top-menu active visual cue with the requested inset underline styling consistent with https://platform.uno/. 
It also updates docs path detection so the active cue works for both hosted and local docs routes (`/docs/articles/...` and `/articles/...`). This is needed to make the current Docs section clearly identifiable and to ensure local validation matches hosted behavior.


MartinZikmund:
@agneszitte Should be reopened?

PR #22823, opened 2026-03-12
fix: prevent Tab key from inserting character in TextBox on Skia

### LINKED ISSUE #22818: TextBox does not handle tab correctly on skia/macos
### Current behavior 🐛

When you tab from a TextBox on Mac using Skia desktop, it adds a tab character as well as tabbing to the next tab index.

### Expected behavior 🎯

No tab character is inserted into the textbox

_No response_

### Workaround 🛠️

None.

### Renderer 🎨

- [x] Skia
- [ ] Native

### Affected platforms 📱💻🖥️

Desktop (macOS)

6.5.31

### IDE version 🧑‍💻

_No response_

_No response_

MartinZikmund:
## Summary

Fixes #22818

On several Skia backends (macOS, FrameBuffer/Linux, Android), pressing **Tab** in a `TextBox` inserts a `\t` character into the text in addition to moving focus. The expected WinUI behavior is that Tab should **only** move focus and never insert a character — `TextBox` has no `AcceptsTab` property (only `RichEditBox` does).

### Root cause

In `TextBox.OnKeyDownSkia()`, there was no explicit `case VirtualKey.Tab:` in the second switch statement (text input/movement). When a backend passes the tab unicode character (`0x09`) through `KeyRoutedEventArgs.UnicodeKey`, the unhandled key fell into the `default` case which inserts any non-null `UnicodeKey` as text. Focus navigation happens later in `UnoFocusInputHandler` (on the root element's KeyDown), but by then the tab character was already inserted.

**Affected backends:** macOS, FrameBuffer (Linux), Android — all pass unicode `0x09` for Tab.
**Unaffected backends:** Win32, WPF, X11, and WebAssembly already filter it out at the platform layer.

### WinUI comparison

WinUI blocks Tab insertion through multiple layers:
1. Windows doesn't generate `WM_CHAR` for Tab, so no character input event exists
2. `KeyboardInputProcessor` triggers `UIElement_TabProcessing` for focus navigation before `CharacterReceived` fires
3. RichEdit single-line mode ignores Tab

In Uno/Skia, layer 1 is inconsistent across backends, layer 2 (`UnoFocusInputHandler`) runs after TextBox processes the key, and there's no RichEdit layer. This fix mirrors WinUI's intent by preventing Tab from reaching character insertion.

### Fix

Added `case VirtualKey.Tab:` with a `return;` before the `default` case in `TextBox.skia.cs`:

- **Tab does NOT insert a character** — we skip the default case entirely
- **The event is NOT marked as `Handled`** — we use `return` (not `break`), so the event bubbles up
- **Focus navigation still works** — `UnoFocusInputHandler.OnKeyDown()` handles Tab for focus navigation as before

Using `return` (not `break`) is important because `break` would fall through to `ProcessTextInput(text)`, whereas `return` exits the method immediately.

## Changes

- `src/Uno.UI/UI/Xaml/Controls/TextBox/TextBox.skia.cs` — Add `case VirtualKey.Tab: return;` before the default case
- `src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.skia.cs` — Add runtime test `When_Tab_Pressed_Should_Not_Insert_Character`

## Test plan

- [x] Compile validation: `Uno.UI.Skia.csproj` and `Uno.UI.RuntimeTests.Skia.csproj` build with 0 errors
- [ ] Runtime test: `When_Tab_Pressed_Should_Not_Insert_Character` — creates a TextBox, simulates Tab with unicode `\t`, asserts no tab character is inserted
- [ ] Manual: Open SamplesApp on macOS/Linux, focus a TextBox, press Tab — verify no `\t` character appears and focus moves to the next control

### DIFF
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.skia.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.skia.cs
@@ -5717,6 +5717,34 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 
 		#endregion
 
+		[TestMethod]
+		public async Task When_Tab_Pressed_Should_Not_Insert_Character()
+		{
+			using var _ = new TextBoxFeatureConfigDisposable();
+
+			var SUT = new TextBox();
+			var button = new Button { Content = "Next" };
+			var panel = new StackPanel
+			{
+				Children = { SUT, button }
+			};
+
+			WindowHelper.WindowContent = panel;
+
+			await WindowHelper.WaitForIdle();
+			await WindowHelper.WaitForLoaded(SUT);
+
+			SUT.Focus(FocusState.Programmatic);
+			await WindowHelper.WaitForIdle();
+
+			// Simulate Tab key press with unicode '\t' (as macOS/FrameBuffer/Android backends do)
+			SUT.SafeRaiseEvent(UIElement.KeyDownEvent, new KeyRoutedEventArgs(SUT, VirtualKey.Tab, VirtualKeyModifiers.None, unicodeKey: '\t'));
+			await WindowHelper.WaitForIdle();
+
+			Assert.AreEqual(string.Empty, SUT.Text, "Tab should not insert a character in TextBox");
+			Assert.IsFalse(SUT.Text.Contains('\t'), "TextBox.Text should not contain a tab character");
+		}
+
 		private class TextBoxFeatureConfigDisposable : IDisposable
 		{
 			private bool _useOverlay;
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.skia.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.skia.cs
@@ -5743,6 +5743,10 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 
 			Assert.AreEqual(string.Empty, SUT.Text, "Tab should not insert a character in TextBox");
 			Assert.IsFalse(SUT.Text.Contains('\t'), "TextBox.Text should not contain a tab character");
+
+			// Ensure that Tab key still moves focus to the next control
+			var focusedElement = FocusManager.GetFocusedElement(SUT.XamlRoot) as Control;
+			Assert.AreEqual(button, focusedElement, "Tab should move focus to the next control (button)");
 		}
 
 		private class TextBoxFeatureConfigDisposable : IDisposable
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.skia.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.skia.cs
@@ -5718,6 +5718,7 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 		#endregion
 
 		[TestMethod]
+		[GitHubWorkItem("https://github.com/unoplatform/uno/issues/22818")]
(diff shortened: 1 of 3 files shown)


morning4coffe-dev (approved):

ajpinedam (approved):

ramezgerges:
I don't see this on SamplesApp inside any textbox. Tab will just move focus. Is there a specific scenario that needs this?

ajpinedam:
> I don't see this on SamplesApp inside any textbox. Tab will just move focus. Is there a specific scenario that needs this?

@ramezgerges  are you on mac? I believe this is specific to macOS. I was able to reproduce it, but on mac.

MartinZikmund:
@ramezgerges yes, it is only happening on some targets, Win32 for example already prevented it above in the native layer I think

MartinZikmund:
## Review comments addressed — commit `2c26921fc4`

All 4 unresolved threads were judged against the current code on `dev/mazi/tab-input`.

### Fixed (2)

|---|---|---|
| `Given_TextBox.skia.cs:5721` | Annotate the issue-covering test with `[GitHubWorkItem]` (convention in this file / `runtime-tests.md`) | Added `[GitHubWorkItem("https://github.com/unoplatform/uno/issues/22818")]` above `When_Tab_Pressed_Should_Not_Insert_Character`. |
| `Given_TextBox.skia.cs:5745` | `Assert.IsFalse(SUT.Text.Contains('\t'))` is redundant after asserting `SUT.Text == string.Empty` | Removed the redundant assertion. The `AreEqual(string.Empty, …)` assert plus the focus-moved assert remain. |

### No change needed (2)

Both remaining threads (one from Copilot, one from the ultrareview bot) raise the *same* concern: the `case VirtualKey.Tab: return;` path skips `TrySetCurrentlyTyping(false)`, so an undo/redo typing run could be merged across a Tab focus change.

| file:line | Point | Why the code is already right |
|---|---|---|
| `TextBox.skia.cs:1051-1054` | Copilot: "Tab is a navigation/focus move — call `TrySetCurrentlyTyping(false)` before returning, otherwise `_currentlyTyping` stays true across the focus change." | The typing run is **already closed by the focus-loss path**. `UnoFocusInputHandler` subscribes to the root element's `KeyDown` (`UnoFocusInputHandler.cs:19`, `:36-39`), so it runs *synchronously* later in the same bubbling dispatch of this very key event, right after `TextBox.OnKeyDownSkia` returns. It calls `FindAndSetNextFocus`, focus leaves the TextBox, and `TextBox.OnFocusStateChangedPartial(FocusState.Unfocused, …)` calls `TrySetCurrentlyTyping(false)` (`TextBox.skia.cs:452-455`). The undo unit is therefore already sealed at the Tab press — the described "type → Tab away → come back → resumed run merges with the pre-Tab run" scenario cannot happen. |
| `TextBox.skia.cs:1054` | github-actions (ultrareview): same, with a `suggestion` block adding `TrySetCurrentlyTyping(false);` before the `return;` | Same reasoning. The one case the suggestion *would* change is a Tab that **does not** move focus (no other tab stop, or `TabNavigation` traps focus): today the typing run continues untouched, which is the correct no-op behavior and consistent with how the other no-op keys in this switch (`Shift`/`Control`/`Menu`/… at `:981-993`) return without ending the run. Adding the call there would split the undo history at a point where the user observed *nothing at all* happening — a small behavioral regression for no benefit. The `VirtualKey.A when ctrl` precedent the bot cites is different: select-all *does* mutate selection state within the TextBox, so it legitimately needs its own undo boundary. |

### Validation

Changes were **reviewed by code inspection only** — nothing was built, run, or tested locally (disk-constrained environment). Azure Pipelines CI will verify the build and the runtime test.

morning4coffe-dev (approved):
Code review assessment of the current head: returning for Tab prevents insertion of the backend-supplied control character without marking the key handled, leaving routed focus navigation available. The separate character-release path already rejects control characters. This is consistent with TextBox/PasswordBox not exposing AcceptsTab; it does not prevent programmatic/pasted text containing tabs. No blocking issue found in the insertion fix. Ending the typing transaction when Tab cannot transfer focus is a useful non-blocking undo-boundary follow-up. This reviews the updated head rather than merely repeating the old approval.

PR #22873, opened 2026-03-23
WIP feat: Intermediate scrolling updates mode

MartinZikmund:
**GitHub Issue:** closes #

Scrolling is stuttering on Android and WASM

Should no longer jitter

## PR Checklist ✅

### DIFF
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_ScrollViewer.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_ScrollViewer.cs
@@ -2087,5 +2087,201 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 			// The scroll offset should remain 0 — there's nothing to scroll
 			Assert.AreEqual(0d, sut.VerticalOffset, "Should not have scrolled");
 		}
+
+		#region Skia intermediate scrolling mode
+
+		[TestMethod]
+		[RunsOnUIThread]
+		[PlatformCondition(ConditionMode.Include, RuntimeTestPlatforms.Skia | RuntimeTestPlatforms.NativeWinUI)]
+		public async Task When_ChangeView_DisableAnimation_Then_SingleViewChangedEvent()
+		{
+			var content = new Border
+			{
+				Height = 2000,
+				Background = new SolidColorBrush(Colors.DeepPink)
+			};
+			var sut = new ScrollViewer
+			{
+				Height = 200,
+				Width = 200,
+				Content = content,
+			};
+
+			await UITestHelper.Load(sut);
+
+			var viewChangedCount = 0;
+			sut.ViewChanged += (s, e) => viewChangedCount++;
+
+			sut.ChangeView(null, 100, null, disableAnimation: true);
+
+			// On both WinUI and Skia, ChangeView stores offsets and defers DP updates to the arrange pass.
+			// ViewChanged fires during the layout pass. Exactly one event should be raised.
+			await WindowHelper.WaitFor(() => viewChangedCount >= 1);
+			await WindowHelper.WaitForIdle();
+
+			Assert.AreEqual(1, viewChangedCount, "Exactly one ViewChanged event should be raised for ChangeView with disableAnimation.");
+		}
+
+		[TestMethod]
+		[RunsOnUIThread]
+		[PlatformCondition(ConditionMode.Include, RuntimeTestPlatforms.Skia | RuntimeTestPlatforms.NativeWinUI)]
+		public async Task When_ChangeView_DisableAnimation_Then_ViewChangedIsNotIntermediate()
+		{
+			var content = new Border
+			{
+				Height = 2000,
+				Background = new SolidColorBrush(Colors.DeepPink)
+			};
+			var sut = new ScrollViewer
+			{
+				Height = 200,
+				Width = 200,
+				Content = content,
+			};
+
+			await UITestHelper.Load(sut);
+
+			var records = new List<bool>();
+			sut.ViewChanged += (s, e) => records.Add(e.IsIntermediate);
+
+			sut.ChangeView(null, 100, null, disableAnimation: true);
+
+			// On both WinUI and Skia, ViewChanged fires during the layout pass.
+			// Exactly one non-intermediate event should be raised.
+			await WindowHelper.WaitFor(() => records.Count >= 1);
+			await WindowHelper.WaitForIdle();
+
+			Assert.AreEqual(1, records.Count, "Exactly one ViewChanged event should be raised.");
+			Assert.IsFalse(records[0], "The ViewChanged event from ChangeView should have IsIntermediate=false.");
+		}
+
+		[TestMethod]
+		[RunsOnUIThread]
+		[PlatformCondition(ConditionMode.Include, RuntimeTestPlatforms.Skia | RuntimeTestPlatforms.NativeWinUI)]
+		public async Task When_ChangeView_DisableAnimation_Then_OffsetUpdatedAfterArrange()
+		{
+			var content = new Border
+			{
+				Height = 2000,
+				Background = new SolidColorBrush(Colors.DeepPink)
+			};
+			var sut = new ScrollViewer
+			{
+				Height = 200,
+				Width = 200,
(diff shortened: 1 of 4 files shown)


MartinZikmund:
## Review comments addressed

Pushed commit **`e02d938a3d`** — `refactor(scrollviewer): Drop redundant isIntermediate check`.

All 8 open review threads were re-checked against the **current** code on `dev/mazi/scrolling-improve`. Seven of them were already satisfied by commits pushed earlier on this branch (`9994515efb` and `de3219e9d3`) — the bot threads were simply never resolved. One was still live and is fixed here.

### Fixed (1)

| File:line | Reviewer point | What changed |
|---|---|---|
| `src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.cs:1319` | `!isIntermediate` in the `else if` is always true — redundant, and reads as if a non-obvious condition is being tested | Adopted the suggested restructure: the redundant `!isIntermediate &&` is gone and the snap-points check moved into the `else if` condition, so the inner `if` disappears. On Apple/Android the `UseNativeSnapping != true` guard is `&&`-ed in front of it; on other platforms the condition is just the snap-points check. Semantics are unchanged on every platform. |

### No change needed — already handled (7)

| File:line | Reviewer point | Why no change |
|---|---|---|
| `ScrollViewer.cs:1420` / `:1448` (2 threads) | Automation peer gets the *new* offsets as the "old" args, so UIA scroll-percent `PropertyChanged` never fires | Already fixed by `9994515efb`. `OnPresenterArranged` now captures `oldHorizontalOffset`/`oldVerticalOffset` **before** the DP assignments and threads them into `RaiseViewChanged(bool, double, double)` → `peer.RaiseAutomationEvents`. The delayed path also preserves the pre-batch offsets (`m_delayedOldHorizontalOffset`/`m_delayedOldVerticalOffset`), so a batched event still reports the full delta. |
| `ScrollViewer.cs:1677` / `:1735` (2 threads) | Comment claims the `SCP → OnPresenterScrolled` chain is synchronous when `disableAnimation` is true — untrue in the default `Deferred` mode | Comment already rewritten (now at `ScrollViewer.cs:1766-1769`) and it says exactly what was asked: batching covers zoom (`OnPresenterZoomed`) and `Synchronous` `UpdatesMode`, and in `Deferred` mode `OnPresenterScrolled` only stores pending offsets with the event raised later during arrange, so nothing is batched there. |
| `Given_ScrollViewer.cs:2177` / `:2195` / `:2251` (3 threads) | Platform skip should use `[PlatformCondition]`, not `#if` on `[Ignore]` | Already fixed by `de3219e9d3`. Both `When_TouchScroll_Then_IntermediateEventsFollowedByFinal` and `When_TouchPressReleaseWithoutScroll_Then_NoViewChangedEvents` now carry `[PlatformCondition(ConditionMode.Include, RuntimeTestPlatforms.Skia)]`, with the `#if !HAS_INPUT_INJECTOR` `[Ignore]` guard kept as `#if` (compile-time constant, no runtime equivalent) — precisely the shape `runtime-tests.md` and the suggestions call for. |

None — nothing ambiguous was left open.

---

**Validation level:** this change was reviewed **by inspection only** — nothing was built, run, or tested locally. Azure Pipelines CI will verify. The edit is a pure condition restructure with no behavior change, and the `#if` shape compiles on both the Apple/Android and the non-mobile branches.

src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.cs
```diff
@@ -1180,45 +1180,55 @@ internal void OnPresenterScrolled(double horizontalOffset, double verticalOffset
...
+				VerticalOffset = verticalOffset;
+				RaiseViewChanged(isIntermediate || m_isInIntermediateViewChangedMode, oldHorizontalOffset, oldVerticalOffset);
 			}
 			else
 			{
-				Update(isIntermediate);
+				// WinUI-aligned behavior: store pending offsets and defer DP updates to ArrangeOverride.
+				// On WinUI, SetOffsetsWithExtents stores offsets in ScrollData + InvalidateArrange;
+				// the DPs are updated during ArrangeOverride → VerifyScrollData → put_HorizontalOffset/VerticalOffset.
+				_hasPendingScrollUpdate = true;
```
morning4coffe-dev, line 1204: Deferring the DP update here makes `ChangeView`'s existing no-op check use stale state. From offset 0, `ChangeView(null, 100, null, true)` queues 100, then a same-tick `ChangeView(null, 0, null, true)` is rejected because the public `VerticalOffset` is still 0; the presenter keeps the first target and produces a spurious move/event. WinUI compares against the current/imminent target rather than only the displayed DP. Please use the pending/presenter target for no-op detection and add a back-to-back revert test.

src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.cs
```diff
@@ -1231,56 +1241,101 @@ internal void OnPresenterZoomed(float zoomFactor)
 			// Recalculate scrollable dimensions with new zoom factor
 			UpdateDimensionProperties();
 
-			// Note: We should also defer the intermediate zoom changes
-			Update(isIntermediate: false);
+			RaiseViewChanged(isIntermediate: false);
```
morning4coffe-dev, line 1244: `OnPresenterZoomed` also runs from touch `OnUpdated` while intermediate mode is active, so hard-coding `false` emits a terminal `ViewChanged` in the middle of a pinch, followed by more intermediate events. WinUI's `OnZoomFactorChanged` passes `m_isInIntermediateViewChangedMode`. Please preserve that state here and add pinch sequencing coverage.

src/Uno.UI/UI/Xaml/Controls/ScrollContentPresenter/ScrollContentPresenter.Managed.cs
```diff
@@ -804,13 +805,15 @@ void IDirectManipulationHandler.OnCompleted(GestureRecognizer _, ManipulationCom
 			if (args?.IsInertial is true && _touchInertia is null)
 			{
 				// Inertia has been aborted (external ChangeView request?) or was not even allowed, do not try to apply the final value.
+				Scroller?.LeaveIntermediateViewChangedMode(raiseFinalViewChanged: false);
```
morning4coffe-dev, line 808: This suppresses the terminal event whenever inertia is unclaimed. A fling released outward at a boundary sets `_touchInertia = null`; the coordinator completes with `IsInertial=true`, so this branch runs after earlier intermediate drag events and no later `Set(...false)` occurs. WinUI suppresses the immediate final only when an offset update is still pending. Please distinguish that case from unclaimed inertia and guarantee one final non-intermediate event.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_ScrollViewer.cs
```diff
@@ -2154,5 +2154,199 @@ public async Task When_TouchFlickOnNonScrollable_Then_InertiaDoesNotAssert()
...
+			};
+
+			await UITestHelper.Load(sut);
+
+			var viewChangedCount = 0;
+			sut.ViewChanged += (s, e) => viewChangedCount++;
+
+			sut.ChangeView(null, 100, null, disableAnimation: true);
+
+			// On both WinUI and Skia, ChangeView stores offsets and defers DP updates to the arrange pass.
```
morning4coffe-dev, line 2184: This deferred contract requires updating `VerifyUniformGridLayoutDoesntHangWhenTryingToScrollToStart`. On WASM, it calls `ChangeView` and then synchronously `WaitOne()`s on the UI thread, so the arrange required to raise `ViewChanged` can never run. Builds 229592, 231025, and 231727 all stop at that exact test and hit the 35-minute timeout. Please make that wait asynchronous or explicitly use synchronous updates for the test.

morning4coffe-dev (commented):
Draft review: four correctness blockers remain before this is ready. The deferred offset model can drop a later same-tick `ChangeView`, pinch zoom can emit a terminal event while manipulation is still active, boundary flings can omit the terminal event, and the deferred completion deterministically deadlocks an existing single-threaded WASM test. The WASM shard failure is directly related; the Android crash occurred during test enumeration and appears unrelated.

PR #22933, opened 2026-03-31
fix: native clipping never cleared on AppleUIKit

rajamatt:
related to https://github.com/unoplatform/dispatchscience-private/issues/71

## PR Type:

- 🐞 Bugfix

In `RootViewController.cs` on AppleUIKit (Skia iOS), when the SVG clip path becomes `null` (meaning all native areas are covered by Skia content), `ClearNativeClipping()` is never called. This happens because `_lastSvgClipPath` is set to `null` *before* the `NativeDispatcher.Main.Enqueue` lambda executes. The lambda then checks `_lastSvgClipPath is not null`, which is always `false`, so the `ClearNativeClipping()` branch is dead code.

As a result, the `CAShapeLayer` mask on `NativeOverlayLayer` is never removed, causing `HitTest` to continue returning non-nil values. This prevents touches from passing through to the `TopViewLayer` for Skia input handling (e.g., MAUI Picker popups fail to receive touch input).

The fix removes the captured `oldPath` variable and changes the `else if (_lastSvgClipPath is not null)` guard to a plain `else`. Now when `svgPath` is `null`, `ClearNativeClipping()` is always called, correctly removing the `CAShapeLayer` mask so touches pass through to Skia.

## PR Checklist ✅

## Other information ℹ️

The bug was a classic closure-over-mutable-state issue: `_lastSvgClipPath` was assigned `null` synchronously, then the async lambda captured the field (not a snapshot), so by the time the lambda ran the guard `_lastSvgClipPath is not null` was always false.

### DIFF
--- a/src/Uno.UI.Runtime.Skia.AppleUIKit/UI/Xaml/Window/RootViewController.cs
+++ b/src/Uno.UI.Runtime.Skia.AppleUIKit/UI/Xaml/Window/RootViewController.cs
@@ -140,13 +140,6 @@ internal class RootViewController : UINavigationController, IAppleUIKitXamlRootH
 				}
 				else
 				{
-					// svgPath is null meaning the clip path became empty (all native areas
-					// are covered by Skia content). Clear the CAShapeLayer mask so that
-					// NativeOverlayLayer.HitTest returns nil and touches pass through to
-					// the TopViewLayer for Skia input handling.
-					// NOTE: Previously this checked `_lastSvgClipPath is not null`, but
-					// _lastSvgClipPath was already set to null before the lambda ran,
-					// so ClearNativeClipping was never called.
 					ClearNativeClipping();
 				}
 			});
--- a/src/Uno.UI.Runtime.Skia.AppleUIKit/UI/Xaml/Window/RootViewController.cs
+++ b/src/Uno.UI.Runtime.Skia.AppleUIKit/UI/Xaml/Window/RootViewController.cs
@@ -130,7 +130,6 @@ internal class RootViewController : UINavigationController, IAppleUIKitXamlRootH
 
 		if (svgPath != _lastSvgClipPath)
 		{
-			var oldPath = _lastSvgClipPath;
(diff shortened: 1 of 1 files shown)


src/Uno.UI.Runtime.Skia.AppleUIKit/UI/Xaml/Window/RootViewController.cs
```diff
@@ -139,8 +138,15 @@ private void UpdateNativeClipping(SKPath path)
 				{
 					ClipBySvgPath(svgPath);
 				}
-				else if (_lastSvgClipPath is not null)
+				else
 				{
+					// svgPath is null meaning the clip path became empty (all native areas
+					// are covered by Skia content). Clear the CAShapeLayer mask so that
+					// NativeOverlayLayer.HitTest returns nil and touches pass through to
```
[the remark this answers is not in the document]
rajamatt: Remove the comment because it was just bloating code

src/Uno.UI.Runtime.Skia.AppleUIKit/UI/Xaml/Window/RootViewController.cs
```diff
@@ -139,7 +138,7 @@ private void UpdateNativeClipping(SKPath path)
 				{
 					ClipBySvgPath(svgPath);
 				}
-				else if (_lastSvgClipPath is not null)
```
jeromelaban, line 142: @MartinZikmund would you know of a good way to test this part?

ramezgerges (approved):

MartinZikmund (approved):

src/Uno.UI.Runtime.Skia.AppleUIKit/UI/Xaml/Window/RootViewController.cs
MartinZikmund: No really testable well, unfortunately

PR #23028, opened 2026-04-13
test: Verify custom Dictionary subclass compiles when used in XAML resources

### LINKED ISSUE #13373: Cannot create custom `Dictionary` in XAML
### Current behavior

Code generation fails when attempting to instantiate a custom Dictionary class in XAML

### Expected behavior

Code should generate fine

Create a custom Dictionary class such as

```cs
public class MyDictionary : Dictionary<string, string> { }
```

Attempt to create a instance of MyDictionary in your XAML

```xml
<Page.Resources>
	<local:MyDictionary x:Key="MyDict">
		<x:String x:Key="One">One</x:String>
		<x:String x:Key="Two">Two</x:String>
	</local:MyDictionary>
</Page.Resources>
```

Note the build error:

> error CS7036: There is no argument given that corresponds to the required parameter 'value' of 'Dictionary<string, string>.Add(string, string)'

Sample repro app attached. Just remove the `win:` prefix in the Page.Resources` of the `MainPage.xaml` to see the error
[UnoApp57.zip](https://github.com/unoplatform/uno/files/12421373/UnoApp57.zip)

### Workaround

_No response_

### Works on UWP/WinUI

Yes

### Environment

_No response_

### NuGet package version(s)

_No response_

### Affected platforms

Build tasks

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

MartinZikmund:
Closes #13373

## Summary

Issue #13373 reports that a custom `Dictionary<string, string>` subclass used as a XAML resource produced `error CS7036: There is no argument given that corresponds to the required parameter 'value' of 'Dictionary<string, string>.Add(string, string)'` from the generated code.

The test sets up exactly that scenario (`MyDictionary : Dictionary<string, string>` filled with `<x:String x:Key="...">` entries) and asserts the XAML generator produces compiling code.

The test **passes on current master** — the generator now emits a valid `Add(key, value)` call for custom dictionary subclasses.

### Test(s) added
- `src/SourceGenerators/Uno.UI.SourceGenerators.Tests/XamlCodeGeneratorTests/Given_ResourceDictionary.cs` → `Given_ResourceDictionary.When_Custom_Dictionary_Class_Used_In_XAML_13373`

### DIFF
--- a/src/SourceGenerators/Uno.UI.SourceGenerators.Tests/XamlCodeGeneratorTests/Given_ResourceDictionary.cs
+++ b/src/SourceGenerators/Uno.UI.SourceGenerators.Tests/XamlCodeGeneratorTests/Given_ResourceDictionary.cs
@@ -185,4 +185,60 @@ public class Given_ResourceDictionary
 
 		await Verify.AssertXamlGenerator(test);
 	}
+
+	// Reproduction for https://github.com/unoplatform/uno/issues/13373
+	// Custom Dictionary subclass declared in XAML produces invalid Add(...) calls
+	// because the generator picks the IDictionary base method that requires
+	// (key, value) but emits only the value, surfacing as CS7036.
+	[TestMethod]
+	public async Task When_Custom_Dictionary_Class_Used_In_XAML_13373()
+	{
+		var xamlFile = new XamlFile(
+			"MainPage.xaml",
+			"""
+			<Page
+				x:Class="TestRepro.MainPage"
+				xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+				xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+				xmlns:local="using:TestRepro">
+
+				<Page.Resources>
+					<local:MyDictionary x:Key="MyDict">
+						<x:String x:Key="One">One</x:String>
+						<x:String x:Key="Two">Two</x:String>
+					</local:MyDictionary>
+				</Page.Resources>
+			</Page>
+			""");
+
+		var test = new Verify.Test(xamlFile)
+		{
+			TestState =
+			{
+				Sources =
+				{
+					"""
+					using System.Collections.Generic;
+					using Microsoft.UI.Xaml.Controls;
+
+					namespace TestRepro
+					{
+						public sealed partial class MainPage : Page
+						{
+							public MainPage()
+							{
+								this.InitializeComponent();
+							}
+						}
+
+						public class MyDictionary : Dictionary<string, string> { }
+					}
+					"""
+				}
+			}
+		};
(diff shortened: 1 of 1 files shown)


MartinZikmund:
## PR review comments addressed

Pushed commit `3c83f641ec` on `dev/mazi/repro/13373`. All three open threads point at the same repro comment (`Given_ResourceDictionary.cs:266`) and make the same request. Reviewed **by inspection only** (no local build — CI will verify); the change is comment-text only in a test file, so it cannot affect compilation.

### Investigation

I traced the actual failure against the fix that resolved the issue (commit `8a208e86a4` *"fix: Properly generate code for items in custom dictionary"*). Pre-fix, `XamlFileGenerator` emitted, per item:

```csharp
MyDict.Add(
    /* BuildChild(item) */   // the element's VALUE only
);
```

`BuildChild` builds the `<x:String x:Key="One">One</x:String>` **element (the value)**; the `x:Key` is fetched separately via `GetDictionaryResourceKey` and was **dropped**. So the broken call was `MyDict.Add(<value>)` — a single argument that is the **value**, with the **key omitted** → CS7036.

So the reviewers' suggested wording `emits a one-argument Add(key) call` is **technically inaccurate**: the emitted argument was the *value*, not the key. CS7036 names the `value` parameter only because that lone value argument binds positionally to the first (`key`) slot. Applying the literal suggestion would have introduced a factual error.

The reviewers' underlying *clarity* concern is valid, though, so I rewrote the clause to be unambiguous while staying factually correct — naming that the **key** is omitted and that it's a **single-argument** `Add(value)` call.

### Fixed / addressed (3)

|---|---|---|
| `Given_ResourceDictionary.cs:266` | copilot — `emits only the value` is misleading; suggest `Add(key)` | Reworded to `…targets the IDictionary Add that requires (key, value) but omits the key, emitting a single-argument Add(value) call, surfacing as CS7036.` |
| `Given_ResourceDictionary.cs:266` | github-actions (ultrareview) — same ambiguity; suggest `Add(key)` | Same edit — resolves the ambiguity without the `Add(key)` inaccuracy |
| `Given_ResourceDictionary.cs:266` | github-actions (ultrareview) — duplicate of above; suggest `Add(key)` | Same edit |

### No change needed (0)

_Note: I deliberately did **not** apply the exact `Add(key)` suggestion because `BuildChild` emits the element value (the key is what got dropped); the reworded comment addresses the same ambiguity accurately. Reviewed by inspection; CI validates on push._

morning4coffe-dev (approved):
This is a meaningful compilation regression for the custom Dictionary<string,string> subclass and keyed XAML entries. SkipGeneratedSourcesCheck suppresses golden-source comparison, not the generator/compiler diagnostics checked by the verifier, so an invalid one-argument Add still fails the case. No product behavior is changed and no new WinUI collection-population discrepancy is introduced.

PR #23065, opened 2026-04-16
feat: Animate on keyboard scroll

MartinZikmund:
**GitHub Issue:** closes #

## PR Checklist ✅

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/ScrollContentPresenter/ScrollContentPresenter.Managed.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/ScrollContentPresenter/ScrollContentPresenter.Managed.cs
@@ -327,7 +327,10 @@ namespace Microsoft.UI.Xaml.Controls
 				var easing = CompositionEasingFunction.CreatePowerEasingFunction(compositor, CompositionEasingFunctionMode.Out, 10);
 				var animation = compositor.CreateVector2KeyFrameAnimation();
 				animation.InsertKeyFrame(1.0f, target, easing);
-				animation.Duration = TimeSpan.FromSeconds(1);
+				// Distance-based duration: snappy for small motions (keyboard arrows ~30 px → 200 ms),
+				// smoother for larger jumps (wheel notch ~200 px → ~450 ms, big ChangeView → capped at 600 ms).
+				var distance = Vector2.Distance(visual.AnchorPoint, target);
+				animation.Duration = TimeSpan.FromMilliseconds(Math.Clamp(150 + distance * 1.5, 200, 600));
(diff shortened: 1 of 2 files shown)


MartinZikmund:
## Review-comment pass — pushed `f84ca7205b`

Automated pass over the 6 unresolved review threads. **Validation level: code inspection only — nothing was built, run, or tested locally. Azure Pipelines CI is the verifier.**

### Fixed (2)

|---|---|---|
| `ScrollViewer.cs:1852` (github-actions, ×2 duplicate threads) | Comment narrates what `disableAnimation: false` already says; trim to the WinUI attribution only | Applied the suggestion verbatim — the second line is gone, leaving `// WinUI routes keyboard scrolls through ChangeViewWithOptionalAnimation(disableAnimation=FALSE).` This matches `.claude/rules/code-style.md` (comments state the non-obvious *why*). |

### Needs @MartinZikmund input (4) — left open

| file:line | the point | why it needs you |
|---|---|---|
| PR description (copilot) | `**GitHub Issue:** closes #` is incomplete — needs a real issue number | Correct per AGENTS.md ("every PR must reference an associated issue"), but only you know which issue this animates-on-keyboard-scroll work closes. The PR body also still has every template section unfilled. I can't pick the number for you. |
| `ScrollViewer.cs:1859` (github-actions ×2, copilot ×1 — same point, 3 threads) | Keyboard scroll now fires many `ViewChanged` with `IsIntermediate=true` followed by one final event, instead of a single instant `IsIntermediate=false`. Asks for a runtime test asserting ≥1 intermediate event and a final non-intermediate one at the expected offset. | The point is **technically valid** — the `ViewChanged` contract for keypresses genuinely changed. But a test that races a 200–600 ms composition animation to catch an intermediate event is exactly the kind of timing-sensitive test that needs to be *run* to be trusted, and I am not permitted to build or run here. Writing it blind would be a coin-flip on CI flakiness, so I'm escalating rather than guessing. |

### Heads-up while you're in there (not a review thread)

Worth checking what CI says about the **existing** keyboard tests in `src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_ScrollViewer.cs`:

- `When_ArrowKeys_Pressed` (line ~153) and the PageDown/Home/End test (line ~421) press a key, `await WindowHelper.WaitForIdle()`, then assert an **exact** `SUT.VerticalOffset` / `HorizontalOffset`.
- `VerticalOffset` is fed by `OnPresenterScrolled` including its *intermediate* values, and `WaitForIdle` pumps the dispatcher — it does not wait out a 200–600 ms composition animation.

By inspection that looks like it could make those assertions observe a mid-animation offset. I did **not** touch them (unvalidated change, and CI is the right oracle) — but if they go red, that's the likely reason, and the fix would pair naturally with the new-test request above.

*Threads judged individually against the current code; the two comment-style threads were duplicates of one another, as were three of the runtime-test threads.*

PR #23113, opened 2026-04-22
fix: Resource resolution

MartinZikmund:

When a cached app-level `Style` containing `{ThemeResource}` setters is retrieved via `Application.Current.Resources[...]` and assigned programmatically to an already-loaded control (`control.Style = cachedStyle`), the `ThemeResourceReference` tracking that setter is created with a **null pinned `ResourceDictionary`**.

This regressed with #22887 (Lazy theme resource evaluation), which introduced a pinned-dictionary fast path matching WinUI's `CThemeResource::RefreshValue` — but the two resolution paths that can fire for programmatic `Style.ApplyTo` both used **3-parameter** overloads of `TryStaticRetrieval` / `TryTopLevelRetrieval` that don't capture the providing dictionary:

1. `ResourceResolver.ApplyResource` — when `Setter.ApplyTo` runs during `Style.ApplyTo`, this is the primary resolution entry. For app-level themed resources it resolved the value but left `ThemeResourceReference.targetDictionary = null`.
2. `DependencyObjectStore.InnerUpdateResourceBindingsUnsafe`'s `!wasSet` fallback — the load-time fallback for keys not found in the element's ancestor dictionaries also used the 3-param overload, so even the deferred re-resolution path didn't pin.

Net effect: the `_themeResources` entry on the control held an unpinned reference, defeating the O(1) pinned-dictionary fast path that `ThemeResourceReference.RefreshValue` was designed to use. Refreshes fell back to `TryTopLevelRetrieval` on every theme change.

In WinUI (`ThemeResourceExtension.cpp:394`, `PreserveThemeResourceExtension`), Style Setters store the `CThemeResource` markup extension and create a live binding *when the Style is applied to a target control*, which is the point where the providing dictionary is known and pinned. The intent of #22887 was to mirror this via a deferred `Setter.ThemeResourceRef` hook-up path, but that hook-up is currently incomplete (no code path assigns `Setter.ThemeResourceRef`), so every programmatic Style application goes through `ApplyResource` and must pin there.

Both resolution paths now use the **4-parameter** overload that returns the providing `ResourceDictionary`, and construct the `ThemeResourceReference` with that dictionary pinned:

- `src/Uno.UI/UI/Xaml/ResourceResolver.cs` — `ApplyResource` switches to `TryStaticRetrieval(..., out value, out providingDictionary)` and passes `providingDictionary` to the new `ThemeResourceReference`.
- `src/Uno.UI/UI/Xaml/DependencyObjectStore.Theming.cs` — `InnerUpdateResourceBindingsUnsafe`'s `!wasSet` branch switches to `TryTopLevelRetrieval(..., out value, out providingDict)` and pins the returned dictionary on the matching `_themeResources` entry via `themeRef.SetTargetDictionary(providingDict)`.

The pinned dictionary is the owner of the `ThemeDictionaries` (per the WinUI pinning contract — see `ResourceDictionary.TryGetValue`'s comment *"pin THIS dictionary (the owner of ThemeDictionaries), not the inner theme sub-dictionary"*), so subsequent theme changes resolve correctly via `GetActiveThemeDictionary(GetActiveTheme())` without having to re-walk top-level resources.

### Why this matters

- **WinUI parity**: restores the pinned-dictionary contract on the programmatic-Style path. Without pinning, `RefreshValue` falls back to `TryTopLevelRetrieval`, which works for simple app-theme changes but doesn't mirror the WinUI architecture and can diverge in edge cases (custom themes pushed onto the per-subtree theme stack, scenarios where top-level resolution isn't available during re-entrance, etc.).
- **Performance**: theme changes on an app with many themed controls no longer re-walk top-level resources once per binding; the pinned lookup is O(1).
- **Correctness on theme re-entry**: `ThemeResourceReference.RefreshValue` short-circuits on the pinned dictionary before the top-level fallback, which is the path `UpdateAllThemeReferences` Phase B relies on when the element's ancestor walk (Phase A) doesn't include application resources.

## PR Checklist ✅

- [x] 📝 Commits follow [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary)
- [x] 🧪 Added runtime test and unit test coverage for programmatic Style assignment with `{ThemeResource}` setters
- [ ] 📚 Docs
- [ ] 🖼️ Validated PR `Screenshots Compare Test Run` results

## Tests

- **Runtime test** — `Given_ThemeResource.When_AppLevel_Style_With_ThemeResource_Applied_Programmatically` (`src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_ThemeResource.cs`): sets up an app-level `ResourceDictionary` with Light/Dark `ThemeDictionaries` and a Style referencing it, loads a `TextBlock`, assigns the cached Style programmatically, asserts the correct themed value.
- **Unit test** — `Given_ResourceDictionary.When_AppLevel_Style_With_ThemeResource_Applied_Programmatically` (`src/Uno.UI.Tests/Windows_UI_Xaml/Given_ResourceDictionary.cs`): programmatically assigns a Style with a themed Setter, toggles `Application.RequestedTheme` through Dark → Light → Dark, manually triggers the store's `UpdateResourceBindings(ThemeResource)` (unit tests have no visual tree walk) and asserts the setter re-resolves to the correct themed brush on each transition.

## WinUI References

- `dxaml/xcp/core/core/elements/ThemeResourceExtension.cpp:394` — `PreserveThemeResourceExtension()` check for Style Setters, describing the pattern these fixes align with.
- `Theming.cpp:349-400` — `CDependencyObject::SetThemeResourceBinding`.
- `Theming.cpp:288-346` — `CDependencyObject::UpdateThemeReference` (Phase A ancestor walk, Phase B `RefreshValue` on pinned dictionary).

### DIFF
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_ThemeResource.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_ThemeResource.cs
@@ -344,6 +344,81 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml
 			Assert.AreEqual(Colors.Blue, (border.Background as SolidColorBrush)?.Color,
 				"After hot reload, brush should be updated to Blue even for ThemeResource-only bindings");
 		}
+
+		// When an app-level Style containing {ThemeResource} setters is retrieved from
+		// Application.Current.Resources and assigned programmatically to an already-loaded control,
+		// the setter must resolve against the current theme AND stay in sync with theme changes.
+		// Prior to the ResourceResolver / DependencyObjectStore.Theming fixes, the programmatic path
+		// created a ThemeResourceReference without pinning the providing dictionary.
+		//
+		// Uses element-level RequestedTheme on a container we own, which triggers the per-element
+		// NotifyThemeChanged walk and avoids any host-chrome/Frame RequestedTheme that would
+		// short-circuit an app-level PropagateResourcesChanged walk in the runtime-test harness.
+		[TestMethod]
+		public async Task When_AppLevel_Style_With_ThemeResource_Applied_Programmatically()
+		{
+			var themedBrushes = new ResourceDictionary();
+			var lightDict = new ResourceDictionary();
+			lightDict["Issue455_Foreground"] = new SolidColorBrush(Colors.Red);
+			var darkDict = new ResourceDictionary();
+			darkDict["Issue455_Foreground"] = new SolidColorBrush(Colors.Blue);
+			themedBrushes.ThemeDictionaries["Light"] = lightDict;
+			themedBrushes.ThemeDictionaries["Default"] = darkDict;
+
+			var styleHost = (ResourceDictionary)XamlReader.Load(
+				"""
+				<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+									xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">
+					<Style x:Key="Issue455_TextBlockStyle" TargetType="TextBlock">
+						<Setter Property="Foreground" Value="{ThemeResource Issue455_Foreground}" />
+					</Style>
+				</ResourceDictionary>
+				""");
+
+			using var brushesCleanup = StyleHelper.UseAppLevelResources(themedBrushes);
+			using var styleCleanup = StyleHelper.UseAppLevelResources(styleHost);
+
+			var host = new Border
+			{
+				RequestedTheme = ElementTheme.Dark,
+				Child = new TextBlock { Text = "Test" },
+			};
+			var textBlock = (TextBlock)host.Child;
+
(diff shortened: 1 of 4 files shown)


MartinZikmund:
Closing as superseded.

The theme-resource pinning mechanism this PR introduces has since shipped to `master` in a more evolved form. `master` now contains:

- `src/Uno.UI/UI/Xaml/Data/ThemeResourceReference.cs`
- `src/Uno.UI/UI/Xaml/DependencyObjectStore.Theming.cs` (`SetThemeResourceBinding`)
- the `TryStaticRetrieval(..., out value, out providingDictionary)` overload and the pinned-dictionary resolution path in `ResourceResolver`, plus `ApplyThemeResource` for Setter-applied theme resources

This PR is an earlier iteration of that same mechanism, so rebasing it onto `master` would re-introduce a duplicate, older implementation of code that is already there.

If any scenario from this PR is still not covered by the shipped pinning path, it should be re-filed as a focused issue/PR against the current implementation rather than resurrecting this branch.

PR #23213, opened 2026-05-06
fix(xaml): Apply x:Uid resw values to non-string properties (#17727)

### LINKED ISSUE #17727: [skia] HyperlinkButton does not work when the link is specified in NavigateUri a resource
### Current behavior

Clicking does nothing

### Expected behavior

Webpage should open

Please see the NavigationUri is specified in resw file.

[UnoAppHyperlink.zip](https://github.com/user-attachments/files/16396591/UnoAppHyperlink.zip)

### Workaround

_No response_

### Works on UWP/WinUI

Yes

### Environment

_No response_

### NuGet package version(s)

_No response_

### Affected platforms

Skia (WPF), Skia (Linux X11), Skia (Linux Framebuffer)

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

MartinZikmund:
## Summary

- The XAML source generator dropped `x:Uid` resource values for any property whose type wasn't `String`/`Object` (e.g., `HyperlinkButton.NavigateUri` of type `Uri`), because `BuildInlineLocalizedProperties` filtered candidate properties via `IsLocalizablePropertyType`. The user-reported case in #17727 (clicking a `HyperlinkButton` with `NavigateUri` set via resw did nothing on Skia) is exactly this: the property was never assigned at runtime.
- The fix iterates resw entries by uid (mirroring how `BuildStatementLocalizedProperties` handles attached properties) and wraps non-string values with `XamlBindingHelper.ConvertValue` so the runtime conversion (e.g., `string` → `Uri`) is applied. For string/object properties the generated code is unchanged.

## Test plan

- [x] Unit-test snapshots for the existing `When_Xuid_Basic` scenario remain unaffected (string property → identical generated code)
- [x] New runtime test in `Given_xUid.When_xUid` asserts `HyperlinkButton.NavigateUri` is set from resw (`When_xUid_NavigateUri.NavigateUri` → `https://platform.uno/`)
- [x] `Given_xUid.When_xUid` and `Given_xUid.When_xUid_On_Root` pass on Skia Desktop runtime tests
- [ ] WinUI parity is the issue's premise — the bug report itself confirms "Works on UWP/WinUI: Yes". The runtime test is gated `#if !WINAPPSDK` (uses Uno-specific `WindowHelper`), so `/winui-runtime-tests` cannot exercise it; WinUI's documented behavior was the target of this fix.

Closes #17727

### DIFF
--- a/src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.cs
+++ b/src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.cs
@@ -5116,13 +5116,22 @@ namespace Uno.UI.SourceGenerators.XamlGenerator
 
 			string Inner()
 			{
-				if (IsLocalizedString(propertyType, objectUid))
+				if (!objectUid.IsNullOrEmpty())
 				{
 					var resourceValue = BuildLocalizedResourceValue(FindType(owner.Member.DeclaringType), memberName, objectUid);
 
 					if (resourceValue != null)
 					{
-						return resourceValue;
+						// Match WinUI: an x:Uid resw entry overrides the XAML literal
+						// for the same property, regardless of property type.
+						// (ObjectWriter.cpp: InitiatePropertyReplacementIfNeeded)
+						if (IsLocalizablePropertyType(propertyType))
+						{
+							return resourceValue;
+						}
+
+						var typeName = propertyType.GetFullyQualifiedTypeIncludingGlobal();
+						return $"({typeName})global::Microsoft.UI.Xaml.Markup.XamlBindingHelper.ConvertValue(typeof({typeName}), {resourceValue})";
 					}
 				}
 
--- a/src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.cs
+++ b/src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.cs
@@ -3983,24 +3983,73 @@ namespace Uno.UI.SourceGenerators.XamlGenerator
 			var objectUid = GetObjectUid(objectDefinition);
 
 			var ret = false;
-			if (objectUid != null)
+			if (objectUid != null && objectDefinitionType != null)
 			{
-				var candidateProperties = FindLocalizableProperties(objectDefinitionType)
-					.Except(objectDefinition.Members.Select(m => m.Member.Name));
-				foreach (var prop in candidateProperties)
+				var explicitMembers = new HashSet<string>(objectDefinition.Members.Select(m => m.Member.Name));
+
+				foreach (var resource in _resourceDetailsCollection.FindByUId(objectUid))
 				{
-					var localizedValue = BuildLocalizedResourceValue(null, prop, objectUid);
-					if (localizedValue != null)
+					// The resource key is in the form "<uid>.<member>", or
+					// "<uid>.[using:<ns>]<type>.<property>" for attached properties.
+					var firstDotIndex = resource.Key.IndexOf('.');
+					if (firstDotIndex == -1)
 					{
-						ret = true;
-						if (isInInitializer)
-						{
-							writer.AppendLineInvariantIndented("{0} = {1},", prop, localizedValue);
-						}
-						else
-						{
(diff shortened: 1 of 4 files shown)


MartinZikmund:
## Review comments addressed

Pushed commit `bc54c25473` to `dev/mazi/navigateuriresource`.

All changes below were **reviewed by inspection only — I did not build or run anything; Azure Pipelines CI will verify.**

### Fixed (1)

|---|---|---|
| `Given_xUid.cs:28` | Issue-repro tests should carry `[GitHubWorkItem]` for traceability. | Added `[GitHubWorkItem("https://github.com/unoplatform/uno/issues/17727")]` to the `Given_xUid` class. Resolves via the project's implicit usings (matches every sibling test). |

### No change needed (1)

| file:line | point | why |
|---|---|---|
| `Given_xUid.cs:2` | `#if !WINAPPSDK` guard was removed; PR description still claims the test is gated. | The removal is **intentional** — dedicated commit `8d86fcf753` ("Run x:Uid runtime tests on WinAppSDK as well") deliberately drops the guard so the test validates WinUI parity for `x:Uid`, since `WindowHelper.WindowContent` is supported on WinAppSDK. The sibling Copilot thread on the same guard is already resolved. The reviewer's `[PlatformCondition(..., NativeWinUI)]` alternative is explicitly conditioned on *wanting* to exclude WinAppSDK, which is the opposite of the author's intent. The PR **description** text can be reconciled, but no code change is warranted. |

### Needs @MartinZikmund input (2) — left open

Both threads report the **same** real defect and are technically correct, but the proper fix is subtle, touches shared hot-path source-generator code, and needs a compile + `SourceGenerators.Tests` run to trust — which is beyond this inspection-only pass, so I left them open rather than risk a partial fix.

| file:line | point | why it needs you |
|---|---|---|
| `XamlFileGenerator.cs:4001` (Copilot) | `BuildInlineLocalizedProperties` now uses `FindByUId` + `IndexOf('.')`, which breaks for dotted `x:Uid` (e.g. `x:Uid="MainPage.Button.Sync"`, which **exist** in `SourceGenerators.Tests` scenarios). | Confirmed regression. Pre-PR this method iterated the type's localizable properties and looked up the resw key via **exact-match** `FindByKey(objectUid + "." + prop)`, which handled dotted UIDs. Post-PR it iterates `FindByUId(objectUid)`. But `ResourceDetailsCollection.AddRange` groups resources by the substring **before the first `.`** (`_resourcesByFileNameThenUid`), so for a dotted UID `FindByUId` returns **empty** and the loop never runs — standard `Text`/`Content` localization silently stops working for dotted UIDs. Note the bot's minimal in-loop suggestion (`resource.Key.Substring(uidName.Length + 1)`) is *necessary but not sufficient*: it won't help while `FindByUId` returns nothing. A correct fix must also change the grouping/parsing in `ResourceDetailsCollection` (e.g. expose/centralize `ParseXUid`, split on the `.[using:` sentinel or last `.`), which is a change to shared, perf-sensitive generator code that I'm not confident landing without building + running the generator tests. |
| `XamlFileGenerator.cs:4000` (ultrareview) | Same dotted-UID regression: `resource.Key.IndexOf('.')` picks the wrong property segment; suggests deriving the member from the known UID length or exposing `ParseXUid`. | Same root cause and same conclusion as above — escalating rather than applying a partial fix. |

---
*Reviewed by inspection; CI will verify. `#if !WINAPPSDK` guard removal and the dotted-UID analysis were checked against the branch's own commit history and the `ResourceDetailsCollection` grouping logic.*

morning4coffe-dev (changes requested):
[P2] The dotted-x:Uid regression reported in the existing threads is confirmed. The old ordinary-property discovery could use FindByKey with the complete UID; the new enumeration goes through FindByUId, whose ResourceDetailsCollection grouping splits at the first dot. For x:Uid="A.B" and resource "A.B.Text", no resources are returned for "A.B". The new propertyPath extraction also splits at the first dot, so both sides must be corrected together, with a dotted-UID regression case.

The broader direction is aligned with WinUI: [ObjectWriter::InitiatePropertyReplacementIfNeeded](https://github.com/microsoft/microsoft-ui-xaml/blob/d0c6705ec46dbddb853cdfdb2eb3b42849cf304e/dxaml/xcp/core/Parser/ObjectWriter.cpp#L2196-L2218) substitutes localized property values rather than limiting replacement to CLR string properties. Keep that behavior without losing existing UID names.

PR #23214, opened 2026-05-06
feat: Detect duplicate property assignments in XAML

### LINKED ISSUE #17563: Invalid XAML in HR test
### Current behavior

This is setting Content property twice, which is wrong and should actually be disallowed by Uno.

https://github.com/unoplatform/uno/blob/d334b5d1c700e9ddfcbcaa152ea0ebbc9488fb3e/src/Uno.UI.RuntimeTests/Tests/HotReload/Frame/HRApp/Tests/Pages/HR_Frame_Pages_AppResources.xaml#L9-L11

### Expected behavior

_No response_

_No response_

### Workaround

_No response_

### Works on UWP/WinUI

None

### Environment

_No response_

### NuGet package version(s)

_No response_

### Affected platforms

_No response_

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

MartinZikmund:
**GitHub Issue:** closes #17563

## PR Type:

🐞 Bugfix

## What changed? 🚀

Aligns Uno Platform with WinUI's `AG_E_PARSER2_OW_DUPLICATE_MEMBER` parse-time check: setting the same property on a XAML element more than once — through any combination of attribute syntax, property-element syntax, and implicit child content (resolved via `[ContentProperty]`) — now produces a clear error instead of being silently accepted with a last-writer-wins outcome.

### Source generator (`XamlFileGenerator`)

- New `XamlFileGenerator.DuplicatePropertyValidation.cs` runs a recursive validation pass over the parsed XAML tree before code generation. Each duplicate produces a `UXAML0001` diagnostic with the WinUI-aligned message: `The property 'X' is set more than once`.
- The validator records the offending members so `BuildProperties` and `GetExtendedProperties` skip them during emission. Without this, the generated C# would itself raise `CS1912 Duplicate initialization of member 'X'`, drowning the actual XAML error.
- Whitespace-only literal content (including text wrapped into a `Run` by the `XamlFileParser` inline-text handling) is treated as no content, so formatting whitespace around a property element does not get flagged.

### Runtime XAML reader (`XamlObjectBuilder`)

- A recursive duplicate-property walk now runs at the start of `XamlObjectBuilder.Build`, raising `Uno.Xaml.XamlParseException` with the same message. Previously the broader `ValidateContent` pass only ran on `FrameworkTemplate` materialization, so dynamically-loaded XAML never surfaced this error.

### Bug fixes surfaced by the new validator

- `mergedstyles` (`Generic.xaml`): `MediaTransportControls.PlaybackRateButton` had two `<FontIcon Glyph="EC57"/>` children (the second one a copy-paste duplicate that flanked the `<Button.Flyout>` property element).
- `HR_Frame_Pages_AppResources.xaml`: the original repro from #17563 — `<ContentControl Content="{StaticResource ...}">` wrapped a `<TextBlock Text="{Binding}" />` child, setting `Content` twice. Replaced with a direct `<TextBlock Text="{StaticResource ...}" />` that preserves the test intent (Hot Reload of string resources updates the displayed text) and the descendant `TextBlock` count expected by the existing assertions.

## PR Checklist ✅

### Test coverage

Source generator (`Given_Parser.Errors.cs`, `Given_Grid.cs`):
- Attribute + implicit content / + property element / two property elements (Content)
- Same scenarios on a non-content property (`TextBlock.Text`)
- `DatePicker` `Header` attribute + implicit content (real `[ContentProperty]` validation)
- Negative regressions confirming WinUI parity: `<TextBlock Text="Hi">Something</TextBlock>` does not throw, `<TextBlock Text="Hi"><Run/><Run/></TextBlock>` does not throw, `<DatePicker Header="Test"> </DatePicker>` (whitespace-only content) does not throw.
- Existing `When_Grid_Uses_Both_Syntaxes` was previously asserting the trailing `CS1912` — updated to expect the meaningful `UXAML0001` instead.

Runtime (`Given_XamlReader.cs`, `TextBlockTextAndContentPage.xaml`):
- Same six positive cases via `XamlHelper.LoadXaml<…>` + `Assert.ThrowsExactly<XamlParseException>`.
- Same four negative regressions, including `Assert.AreEqual("HiSomething", sut.Text)` to verify `<TextBlock Text="Hi">Something</TextBlock>` materializes correctly through the runtime reader.
- A compiled-XAML page (`TextBlockTextAndContentPage`) verifies the source-generator path produces the same `Text == "HiSomething"` result for parity.

### Behavior alignment with WinUI

WinUI throws `XamlParseException("The property 'X' is set more than once.")` from `ObjectWriter::Logic_DuplicatePropertyCheck` (`xamlcore\Parser\ObjectWriter.cpp`) keyed on the resolved `XamlProperty`, so attribute, property-element, and implicit-content forms all funnel through the same per-frame "this property already had a `SetMember`" guard. The implementations in this PR follow that same model: a single effective-name resolution maps `_UnknownContent` to the type's `[ContentProperty]` (no per-type special casing required, since `TextBlock`/`Span`/`Run`/`Border`/`SolidColorBrush`/`RowDefinition`/`ColumnDefinition` all declare it).


MartinZikmund:
## Review threads addressed — pushed `bc552e5442`

All 7 unresolved threads triaged against the current code. **Validation level: code inspection only** — I did not build, run, or test anything locally; Azure Pipelines CI will verify.

### Fixed (1)

|---|---|---|
| `Given_XamlReader.cs:182` | New duplicate-property tests should carry `[GitHubWorkItem]` for traceability (per `.claude/rules/runtime-tests.md`) | Added `[GitHubWorkItem("https://github.com/unoplatform/uno/issues/17563")]` to all 10 new tests (`When_DuplicateProperty_*`, `When_TextBlock_TextAttribute_*`, `When_DatePicker*`). No `using` needed — the attribute resolves through the enclosing `Uno.UI.RuntimeTests` namespace, same as `Given_Canvas.cs`. |

### No change needed (6)

| file:line | reviewer point | why |
|---|---|---|
| `XamlObjectBuilder.cs:1855` (github-actions) + `XamlFileGenerator.DuplicatePropertyValidation.cs:105` (Copilot) | Runtime message ends with `.`, generator message doesn't — standardize | **Both are already correct for their surface, and the divergence is deliberate.** WinUI's own resource string is `AG_E_PARSER2_OW_DUPLICATE_MEMBER "The property '%0' is set more than once."` — *with* the period (`dxaml/xcp/win/agcore.debug/CommonErrors.rc`). The runtime `XamlParseException` therefore matches WinUI verbatim, and matches every sibling parse error in `XamlObjectBuilder` (`"The type 'X' was not found."`, …). The source-generator message must **not** end with a dot — that's the documented convention in the `XamlFileGenerator.Errors.cs` header, because the dot is appended when the error is emitted into the generated code. Dropping the runtime period would break WinUI parity *and* intra-file consistency to satisfy a convention that only governs the generator. Added a comment at the runtime site recording this so it doesn't get "fixed" later. |
| `XamlObjectBuilder.cs:1842` + `:1845`, `XamlFileGenerator.DuplicatePropertyValidation.cs:88` + `:93` (4 threads: 2× github-actions, 2× Copilot) | Attached properties are skipped, so `Grid.Row="0"` + `<Grid.Row>1</Grid.Row>` isn't caught; WinUI keys on the resolved `XamlProperty` (declaring type + name) | **Real parity gap — acknowledged, deliberately deferred, now documented in code.** The reviewing bot itself classifies it as "a known limitation worth tracking rather than a blocker". Note the only reachable duplicate here is attribute + property-element (XML already forbids duplicate attributes), so the practical exposure is small, while a wrong key or a mis-detected attached member would emit a **build-breaking error** on valid XAML. That is not a change I'm willing to make on inspection alone in this PR. The misleading "cannot collide" comment in both files is replaced with a `// TODO Uno:` marker that names WinUI's `Logic_DuplicatePropertyCheck` and the `"{DeclaringType}.{Name}"` keying a follow-up should use. |

Also corrected a stale doc comment in `XamlFileGenerator.DuplicatePropertyValidation.cs` that still referenced `UXAML0001` after the diagnostic moved to the WinUI-aligned `XLS0501`.

### Follow-up for @MartinZikmund (no thread left open)

Attached-property duplicate detection needs its own issue (I couldn't file one — GitHub secondary rate limit). Suggested scope: key `seenMembers` on `{DeclaringType}.{Name}` for attached members in **both** validators, plus tests for `Grid.Row` attribute + property-element on one element.

morning4coffe-dev (commented):
The ordinary-property/implicit-content mapping is aligned with WinUI, including TextBlock.Text versus its distinct Inlines content property. The specific diagnostic and suppression of secondary CS1912 errors are useful.

This is not yet full duplicate-member parity: [WinUI validates the resolved XamlProperty](https://github.com/microsoft/microsoft-ui-xaml/blob/d0c6705ec46dbddb853cdfdb2eb3b42849cf304e/dxaml/xcp/core/Parser/ObjectWriter.cpp#L1388-L1398), whereas both new passes skip attached properties and exempt several types by simple name rather than resolved namespace/type identity. Also, the runtime pass only accumulates errors: Build still calls LoadObject at line 86 before inspecting them, so invalid duplicate assignments can run constructors/setters before the already-known parse error is returned. WinUI [rejects the duplicate property before applying it](https://github.com/microsoft/microsoft-ui-xaml/blob/d0c6705ec46dbddb853cdfdb2eb3b42849cf304e/dxaml/xcp/core/Parser/ObjectWriterErrorService.cpp#L212-L244). Please retain these limitations explicitly and address the pre-materialization error boundary.

PR #23290, opened 2026-05-19
fix(x11): crash on first frame when software rendering

### LINKED ISSUE #23275: [Skia X11] X11SoftwareRenderer.Flush() throws "Nullable object must have a value" on first frame when UpdateSize() has not yet run
**Current behavior**

When `X11SoftwareRenderer` is the selected renderer (i.e. Vulkan/OpenGL/EGL initialization all failed and the fallback path in `X11XamlRootHost` instantiates the software renderer), the app can crash on the very first render with:

```
Unhandled exception. System.InvalidOperationException: Nullable object must have a value.
   at System.Nullable`1.get_Value()
   at Uno.WinUI.Runtime.Skia.X11.X11SoftwareRenderer.Flush()
   at Uno.WinUI.Runtime.Skia.X11.X11Renderer.Render()
   at Uno.WinUI.Runtime.Skia.X11.X11XamlRootHost.RenderLoop()
```

The root cause is in [`src/Uno.UI.Runtime.Skia.X11/Rendering/X11SoftwareRenderer.cs`](https://github.com/unoplatform/uno/blob/master/src/Uno.UI.Runtime.Skia.X11/Rendering/X11SoftwareRenderer.cs):

```csharp
protected override void Flush()
{
    _ = X11Helper.XPutImage(
        display: _x11Window.Display,
        drawable: _x11Window.Window,
        gc: _gc,
        image: _xImage!.Value,   // <-- throws if _xImage is still null
        ...);
}
```

`_xImage` is only initialized inside `UpdateSize(int, int)`, which is called from the lambda passed to `CompositionTarget.OnNativePlatformFrameRequested` in `X11Renderer.Render()`. If that lambda is not invoked on the first render pass (e.g. the requested size is 0×0 or the composition target otherwise short-circuits), `Flush()` still runs unconditionally and immediately throws on `_xImage!.Value`.

The same code is present on \`master\` and in our local \`Uno.UI.Runtime.Skia.X11\` 6.7.0-dev.57 build.

**Expected behavior**

\`Flush()\` should be a no-op when \`_xImage\` (or equivalently \`_surface\`) has not yet been initialized, just like the GL/Vulkan renderers tolerate being called before they have a backing surface. Falling back to the software renderer should never crash the process — currently it is the last-resort fallback in \`X11XamlRootHost\`, so when it crashes the app has no further fallback.

**How to reproduce it**

Anything that takes the renderer-selection path in \`X11XamlRootHost\` down to \`new X11SoftwareRenderer(...)\` can trigger this. In our case it was reproduced with a packaged \`net9.0-desktop\` Uno 6.7.0-dev.36 app on Ubuntu 24.04 (Wayland session, XWayland \`DISPLAY=:0\`) where the loader could not locate \`libGL.so.1\` / \`libEGL.so.1\`, so Vulkan + OpenGL + EGL initialization all failed and the software renderer was chosen. The crash is then 100% reproducible on launch.

A minimal logic-level repro: in any X11 desktop app, force the software path (e.g. by hiding \`libGL.so.*\`, \`libEGL.so.*\`, \`libGLX.so.*\` from \`LD_LIBRARY_PATH\` so the GL renderers throw and \`X11XamlRootHost\` falls through to \`_renderer = new X11SoftwareRenderer(...)\`). The first frame will throw.

I can put together a zipped minimal repro if helpful, but the issue is visible directly from the source.

**Workaround**

Make sure the hardware renderers can initialize so the software fallback is never used:
- ensure \`libGL.so.1\`, \`libEGL.so.1\`, etc. are reachable via \`LD_LIBRARY_PATH\` / \`ldconfig\`

A user-side workaround for the actual software-renderer bug isn't possible since the failure is on the first frame, before any application code gets a chance to intervene.

**Suggested fix**

Guard \`Flush()\` (and likewise rework \`Dispose()\` which is already guarded) so it short-circuits when the renderer hasn't been sized yet:

\`\`\`csharp
protected override void Flush()
{
    if (_xImage is not { } xImage)
    {
        return;
    }

    _ = X11Helper.XPutImage(
        display: _x11Window.Display,
        drawable: _x11Window.Window,
        gc: _gc,
        image: xImage,
        srcx: 0, srcy: 0, destx: 0, desty: 0,
        width: (uint)_width, height: (uint)_height);
}
\`\`\`

**Renderer**

- [x] Skia

**Affected platforms**

- Desktop (X11)

**Uno.Sdk version**

\`Uno.Sdk\` 6.7.0-dev.36 (also visible in source on \`master\`); \`Uno.WinUI.Runtime.Skia.X11\` package version 6.7.0-dev.57.

**IDE version**

N/A — repro is at the runtime/packaging level, IDE-agnostic.

**Anything else we need to know?**

Happy to send a PR with the null-guard if it would help.

ramezgerges:
**GitHub Issue:** closes https://github.com/unoplatform/uno/issues/23275

## PR Checklist ✅

### DIFF
--- a/src/Uno.UI.Runtime.Skia.X11/Rendering/X11SoftwareRenderer.cs
+++ b/src/Uno.UI.Runtime.Skia.X11/Rendering/X11SoftwareRenderer.cs
@@ -63,17 +63,20 @@ namespace Uno.WinUI.Runtime.Skia.X11
 
 		protected override void Flush()
 		{
-			_ = X11Helper.XPutImage(
-				display: _x11Window.Display,
-				drawable: _x11Window.Window,
-				gc: _gc,
-				image: _xImage!.Value,
-				srcx: 0,
-				srcy: 0,
-				destx: 0,
-				desty: 0,
-				width: (uint)_width,
-				height: (uint)_height);
+			if (_xImage is { } xImage)
+			{
+				_ = X11Helper.XPutImage(
+					display: _x11Window.Display,
+					drawable: _x11Window.Window,
+					gc: _gc,
+					image: xImage,
+					srcx: 0,
+					srcy: 0,
+					destx: 0,
+					desty: 0,
+					width: (uint)_width,
+					height: (uint)_height);
+			}
 		}
 
 		public override void Dispose()


MartinZikmund (approved):

ajpinedam (approved):

HakanL:
Can we have this backported to v6.6?

agneszitte:

PR #23300, opened 2026-05-21
test: Add ItemsRepeater scroll sample

MartinZikmund:
## Overview

Adds two SamplesApp samples (under `Windows_UI_Xaml_Controls/Repeater`) for exercising `ItemsRepeater` behavior with mixed item templates and high per-item height variance.

### New sample: `ItemsRepeaterChatConversation`

A chat-style conversation backed by a single `ItemsRepeater` + `ScrollViewer`, with a `DataTemplateSelector` dispatching to seven message kinds:

- User bubble (right-aligned)
- Assistant reply (left body text)
- Reasoning section (the only `Expander`-based item)
- Tool-call row (status icon + optional file pill + diff/entry counts)
- Outcome card (Success / Warning / Critical)
- Thinking indicator (spinner + text)
- Benchmark line (accent text)

Notable characteristics, chosen to surface virtualization/layout behavior:

- **Nested image-attachment repeaters inside items** (horizontal row for user bubbles, wrapping `UniformGridLayout` grid for assistant messages).
- **Wide per-item height variance** — cooking copy mixed with lorem ipsum (1–10 sentences) and variable image grids.
- **Seeded, deterministic generation** — reloading the sample and clicking "Add" produce identical content every time, so behavior is reproducible.
- **`ChatAutoScroll`** — a small self-contained attached behavior that scrolls to the bottom on item add (deferred `UpdateLayout()` + `ChangeView`), while respecting the user's scroll position and pointer-over state.

Buttons: Top · Bottom · Add 20 items · Remove expanders · Clear · Reset.

### Updated sample: `ItemsRepeaterMultiTemplate`

Adds a **Remove expanders** button that removes all `Expander`-based items, to exercise item removal.

## Notes

- Both samples are `IsManualTest = true`.
- No runtime/control code is touched — SamplesApp only.

## Validation

- **Compile**: `SamplesApp.Skia.Generic` (net10.0) built green with the new sample (0 warnings, 0 errors).
- **Runtime**: launched Skia Desktop (Win32) and confirmed the sample loads and renders all message kinds; manual scroll/add behavior to be reviewed.

### DIFF
--- a/src/SamplesApp/SamplesApp.Samples/Windows_UI_Xaml_Controls/Repeater/ChatAutoScroll.cs
+++ b/src/SamplesApp/SamplesApp.Samples/Windows_UI_Xaml_Controls/Repeater/ChatAutoScroll.cs
@@ -1,0 +1,253 @@
+using System;
+using System.Collections.Specialized;
+using System.Runtime.CompilerServices;
+using Microsoft.UI.Dispatching;
+using Microsoft.UI.Xaml;
+using Microsoft.UI.Xaml.Controls;
+using Microsoft.UI.Xaml.Input;
+using Microsoft.UI.Xaml.Media;
+
+namespace UITests.Windows_UI_Xaml_Controls.Repeater;
+
+/// <summary>
+/// Self-contained auto-scroll behavior for an ItemsRepeater inside a ScrollViewer, modelled on a
+/// typical chat conversation panel. When enabled it scrolls to the bottom as items are added, but
+/// respects the user's scroll position (won't fight a scroll-up) and skips while the pointer is
+/// over the chat. Used by <see cref="ItemsRepeaterChatConversation"/> to reproduce the dynamic
+/// layout churn (UpdateLayout + ChangeView on add) that a static list lacks.
+/// </summary>
+public static class ChatAutoScroll
+{
+	private const double ScrollTolerance = 100.0;
+
+	private static readonly ConditionalWeakTable<ItemsRepeater, ScrollState> _states = new();
+
+	public static readonly DependencyProperty IsEnabledProperty =
+		DependencyProperty.RegisterAttached(
+			"IsEnabled",
+			typeof(bool),
+			typeof(ChatAutoScroll),
+			new PropertyMetadata(false, OnIsEnabledChanged));
+
+	public static bool GetIsEnabled(DependencyObject obj) => (bool)obj.GetValue(IsEnabledProperty);
+
+	public static void SetIsEnabled(DependencyObject obj, bool value) => obj.SetValue(IsEnabledProperty, value);
+
+	private static void OnIsEnabledChanged(DependencyObject d, DependencyPropertyChangedEventArgs e)
+	{
+		if (d is not ItemsRepeater repeater)
+		{
+			return;
+		}
+
+		if ((bool)e.NewValue)
+		{
+			AttachBehavior(repeater);
+		}
+		else
+		{
+			DetachBehavior(repeater);
+		}
+	}
+
+	private static void AttachBehavior(ItemsRepeater repeater)
+	{
+		var state = new ScrollState(repeater);
+		_states.AddOrUpdate(repeater, state);
+
+		repeater.Unloaded += state.OnUnloaded;
+
+		state.ItemsSourceCallbackToken =
+			repeater.RegisterPropertyChangedCallback(ItemsRepeater.ItemsSourceProperty, state.OnItemsSourceChanged);
+
+		if (repeater.ItemsSource is INotifyCollectionChanged incc)
+		{
+			state.SubscribeCollection(incc);
+		}
+
+		if (repeater.IsLoaded)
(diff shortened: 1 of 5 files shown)


MartinZikmund:
## Review comments addressed — commit `8849512bd0`

All 6 unresolved threads triaged. 4 fixed, 2 declined with reasoning below.

> **Validation level: code inspection only.** Nothing here was built, run, or tested locally — Azure Pipelines CI will verify.

### Fixed (4)

|---|---|---|
| `ItemsRepeaterChatConversation.xaml.cs:20` | Missing trailing space concatenates to `andclicking "Add"` | Added the trailing space. Correct as reported. |
| `ItemsRepeaterChatConversation.xaml.cs:169` | `:N0` is culture-sensitive, contradicting the "seeded, deterministic" claim | Fixed — but the suggested patch was **incomplete**: the same line also has `{_rng.Next(8, 120) / 10.0:0.0}`, whose *decimal separator* is equally culture-sensitive (`8.4` vs `8,4`). Wrapping the whole interpolation in `FormattableString.Invariant(...)` fixes both in one go and keeps the RNG evaluation order (and therefore the seeded sequence) untouched. |
| `ItemsRepeaterChatConversation.xaml.cs:292` | `"h:mm tt"` is locale-sensitive | Now `_clock.ToString("h:mm tt", CultureInfo.InvariantCulture)`. Added `using System.Globalization;`. |
| `ItemsRepeaterChatConversation.xaml.cs:480` | Supporting types should be `internal`, not `public` | Done for all six (`ChatMessageKind`, `ToolCallState`, `OutcomeSeverity`, `ChatImageAttachment`, `ChatMessage`, `ChatMessageTemplateSelector`). Verified safe before applying: the XAML uses classic `{Binding}` (zero `x:Bind`), so it depends on generated bindable metadata — and `BindableTypeProvidersGenerationTask` filters on `IsLocallyPublic`, which is *public **or** internal-declared-in-the-same-module*, so metadata is still emitted. The `x:DataType`/`local:` references resolve in generated code in the same assembly. The page itself stays `public`. |

### No change needed (2)

| file:line | Reviewer point | Why declined |
|---|---|---|
| `ChatAutoScroll.cs:10` | "New files must have `#nullable enable` per `code-style.md`" | `code-style.md` scopes itself to *"Uno.UI / Uno.UWP / Uno.Foundation source"* and says new files **should** enable it — it is not a hard rule for SamplesApp. The established convention in this folder is the opposite: **11 of 1608** sample `.cs` files use `#nullable enable` (0.7%). Flipping these two files is also not the 2-line change the bot suggests — it cascades into CS8618 on the 7 `DataTemplate` properties and `Brush Fill`, CS8603 on the `SelectTemplateCore` return, and CS8622 `sender` nullability mismatches against BCL-annotated delegates (`NotifyCollectionChangedEventHandler`, `EventHandler<T>`). That is a warning-churn pass I cannot validate without a build, so per the repo's debugging-discipline rules I did not guess at it. |
| `ItemsRepeaterChatConversation.xaml.cs:12` | Same `#nullable enable` request | Same reasoning as above. |

Happy to do the nullable pass as a follow-up if you'd rather samples start adopting it — it just needs a real compile to land safely.

src/SamplesApp/SamplesApp.Samples/Windows_UI_Xaml_Controls/Repeater/ChatAutoScroll.cs
```diff
@@ -0,0 +1,253 @@
...
+
+		public ScrollState(ItemsRepeater repeater) => _repeater = repeater;
+
+		public void OnLoaded(object sender, RoutedEventArgs e)
+		{
+			_repeater.Loaded -= OnLoaded;
+			FindAndAttachScrollViewer();
+		}
+
+		public void OnUnloaded(object sender, RoutedEventArgs e) => DetachBehavior(_repeater);
```
morning4coffe-dev, line 122: [P2] This permanently detaches an otherwise enabled behavior. DetachBehavior removes the state from the ConditionalWeakTable, and Detach unregisters Loaded, Unloaded and the ItemsSource callback. IsEnabledProperty remains true, so reinserting the same repeater/page does not fire OnIsEnabledChanged and nothing subscribes again. Auto-scroll silently stops after that normal lifecycle transition. Keep a reload hook for as long as the attached property is enabled, while detaching collection/ScrollViewer handlers during unload, and cover unload -> reload -> add.

morning4coffe-dev (changes requested):
The seeded mixed-height/mixed-template sample is useful, but the attached auto-scroll behavior loses its enabled state operationally after an unload/reload, as described inline. This is SamplesApp behavior, not a fix or proof of WinUI ItemsRepeater virtualization.

The existing OneWay Expander comment also matters when interpreting this repro: user expansion does not update IsReasoningExpanded, so recycling can restore the model's old value. If persistent user expansion is intended, use TwoWay rather than attributing that reset to the repeater.

PR #23301, opened 2026-05-21
`AnimatedIcon`

MartinZikmund:
* Lexer accepts identifiers starting with '_' (Lottie generates names like
  '_theme', '_').
* AnimationExpressionSyntax.Evaluate now takes CompositionAnimation so
  KeyFrameAnimations can also evaluate expression keyframes (not just
  ExpressionAnimation).
* Add Pow, Square, Vector4 and ColorRGB function specifications, plus
  mixed-type binary multiplication for float * Vector2 / Vector3 (used by
  Lottie-generated bezier expressions).

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>**GitHub Issue:** closes #

## PR Checklist ✅


src/Uno.UI.Composition/Composition/KeyFrameAnimations/KeyFrameEvaluator.cs
```diff
@@ -41,8 +41,12 @@ public KeyFrameEvaluator(
...
-		if (elapsed >= _totalDuration)
+		// While paused, freeze elapsed time at the pause point — otherwise the evaluator would
+		// keep advancing as wall-clock time passes (and report "shouldStop" once it exceeds the
+		// duration), even though the animation is being held in place by an AnimationController.
+		var nowTimestamp = _pauseTimestamp ?? _compositor.TimestampInTicks;
+		var elapsed = new TimeSpan(nowTimestamp - _totalPause - _startTimestamp);
+		if (_pauseTimestamp is null && elapsed >= _totalDuration)
 		{
 			return (_finalValue.Value, true);
 		}
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in 589115b — the held/final/exact-hit shortcuts in KeyFrameEvaluator now route through an injected `resolve` delegate (same pattern as `_lerp`), so an expression keyframe at an endpoint is evaluated instead of returning its placeholder `Value`. The `Evaluate(float)` completion branch also returns the resolved value rather than the boxed tuple.

src/Uno.UI.Composition/Composition/KeyFrameAnimations/KeyFrameEvaluator.cs
```diff
@@ -64,7 +68,17 @@ public object Evaluate(float progress)
...
-		var nextKeyFrame = _keyFrames.Keys.FirstOrDefault(k => k >= currentFrame, _keyFrames.Keys.Last());
+		var lastKey = _keyFrames.Keys.Last();
+		// Past the final keyframe: hold the last value. Without this the math below collapses
+		// to "previousKeyFrame == nextKeyFrame", producing a divide-by-zero in the lerp ratio
+		// and returning NaN — which would make any animated property (Opacity, Scale, …) drop
+		// off into invisibility.
+		if (currentFrame >= lastKey)
+		{
+			return (_keyFrames[lastKey].Value, false);
+		}
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in 589115b — the held/final/exact-hit shortcuts in KeyFrameEvaluator now route through an injected `resolve` delegate (same pattern as `_lerp`), so an expression keyframe at an endpoint is evaluated instead of returning its placeholder `Value`. The `Evaluate(float)` completion branch also returns the resolved value rather than the boxed tuple.

src/Uno.UI.Composition/Composition/KeyFrameAnimations/KeyFrameEvaluator.cs
```diff
@@ -41,8 +41,12 @@ public KeyFrameEvaluator(
...
-		var elapsed = new TimeSpan(_compositor.TimestampInTicks - _totalPause - _startTimestamp);
-		if (elapsed >= _totalDuration)
+		// While paused, freeze elapsed time at the pause point — otherwise the evaluator would
+		// keep advancing as wall-clock time passes (and report "shouldStop" once it exceeds the
+		// duration), even though the animation is being held in place by an AnimationController.
+		var nowTimestamp = _pauseTimestamp ?? _compositor.TimestampInTicks;
+		var elapsed = new TimeSpan(nowTimestamp - _totalPause - _startTimestamp);
+		if (_pauseTimestamp is null && elapsed >= _totalDuration)
 		{
 			return (_finalValue.Value, true);
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in 589115b — the held/final/exact-hit shortcuts in KeyFrameEvaluator now route through an injected `resolve` delegate (same pattern as `_lerp`), so an expression keyframe at an endpoint is evaluated instead of returning its placeholder `Value`. The `Evaluate(float)` completion branch also returns the resolved value rather than the boxed tuple.

src/Uno.UI.Composition/Composition/KeyFrameAnimations/KeyFrameEvaluator.cs
```diff
@@ -64,7 +68,17 @@ public object Evaluate(float progress)
...
 	{
-		var nextKeyFrame = _keyFrames.Keys.FirstOrDefault(k => k >= currentFrame, _keyFrames.Keys.Last());
+		var lastKey = _keyFrames.Keys.Last();
+		// Past the final keyframe: hold the last value. Without this the math below collapses
+		// to "previousKeyFrame == nextKeyFrame", producing a divide-by-zero in the lerp ratio
+		// and returning NaN — which would make any animated property (Opacity, Scale, …) drop
+		// off into invisibility.
+		if (currentFrame >= lastKey)
+		{
+			return (_keyFrames[lastKey].Value, false);
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in 589115b — the held/final/exact-hit shortcuts in KeyFrameEvaluator now route through an injected `resolve` delegate (same pattern as `_lerp`), so an expression keyframe at an endpoint is evaluated instead of returning its placeholder `Value`. The `Evaluate(float)` completion branch also returns the resolved value rather than the boxed tuple.

src/Uno.UI.Composition/Composition/KeyFrameAnimations/ScalarKeyFrameAnimation.cs
```diff
@@ -47,12 +52,27 @@ public void InsertKeyFrame(float normalizedProgressKey, float value, Composition
...
+	private static float ResolveScalarKeyFrameValue(KeyFrameAnimation animation, AnimationKeyFrame<float> frame)
+	{
+		if (frame.Expression is null)
+		{
+			return frame.Value;
+		}
+
+		var parsed = new ExpressionAnimationParser(frame.Expression).Parse();
+		var evaluation = parsed.Evaluate(animation);
+		return SubPropertyHelpers.ValidateValue<float>(evaluation);
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in 589115b — each expression is parsed once at `InsertExpressionKeyFrame` time and cached on the keyframe (`ParsedExpression`); the per-frame resolver evaluates the cached syntax instead of re-parsing the string.

src/Uno.UI.Composition/Composition/KeyFrameAnimations/Vector2KeyFrameAnimation.cs
```diff
@@ -41,11 +47,25 @@ public void InsertKeyFrame(float normalizedProgressKey, Vector2 value)
...
+	private static Vector2 ResolveVector2KeyFrameValue(KeyFrameAnimation animation, AnimationKeyFrame<Vector2> frame)
+	{
+		if (frame.Expression is null)
+		{
+			return frame.Value;
+		}
+
+		var parsed = new ExpressionAnimationParser(frame.Expression).Parse();
+		var evaluation = parsed.Evaluate(animation);
+		return SubPropertyHelpers.ValidateValue<Vector2>(evaluation);
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in 589115b — each expression is parsed once at `InsertExpressionKeyFrame` time and cached on the keyframe (`ParsedExpression`); the per-frame resolver evaluates the cached syntax instead of re-parsing the string.

src/Uno.UI.Composition/Composition/KeyFrameAnimations/ScalarKeyFrameAnimation.cs
```diff
@@ -47,12 +52,27 @@ public void InsertKeyFrame(float normalizedProgressKey, float value, Composition
...
+	}
+
+	private static float ResolveScalarKeyFrameValue(KeyFrameAnimation animation, AnimationKeyFrame<float> frame)
+	{
+		if (frame.Expression is null)
+		{
+			return frame.Value;
+		}
+
+		var parsed = new ExpressionAnimationParser(frame.Expression).Parse();
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in 589115b — each expression is parsed once at `InsertExpressionKeyFrame` time and cached on the keyframe (`ParsedExpression`); the per-frame resolver evaluates the cached syntax instead of re-parsing the string.

src/Uno.UI.Composition/Composition/Compositor.skia.cs
```diff
@@ -33,29 +33,54 @@ static partial void Initialize()
...
+		{
+			Visual visual => visual.CompositionTarget,
+			CompositionPropertySet { Owner: Visual ownerVisual } => ownerVisual.CompositionTarget,
+			_ => null,
+		};
+
+		if (target is null)
+		{
+			return;
+		}
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in 64a91c6 — the comment now states that a standalone `CompositionPropertySet` must have its `Owner` set to a `Visual` to tick (which `AnimatedIcon` does); there is no compositor-wide `CompositionTarget` fallback.

src/Uno.UI.Composition/Composition/Compositor.skia.cs
```diff
@@ -33,29 +33,54 @@ static partial void Initialize()
...
+		// Resolve the CompositionTarget that needs invalidation. For Visuals it's the visual's
+		// own target; for a CompositionPropertySet that was created via Visual.Properties, fall
+		// back to the owning Visual's target so animations on `someVisual.Properties.Foo` still
+		// get ticked. For property sets created standalone via Compositor.CreatePropertySet
+		// (e.g. AnimatedIcon's progress property set) fall back to any active CompositionTarget
+		// the compositor already knows about — without one the animation would never tick.
+		ICompositionTarget? target = host switch
+		{
+			Visual visual => visual.CompositionTarget,
+			CompositionPropertySet { Owner: Visual ownerVisual } => ownerVisual.CompositionTarget,
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in 64a91c6 — the comment now states that a standalone `CompositionPropertySet` must have its `Owner` set to a `Visual` to tick (which `AnimatedIcon` does); there is no compositor-wide `CompositionTarget` fallback.

src/Uno.UI.Composition/Composition/SubPropertyHelpers.cs
```diff
@@ -11,17 +11,36 @@ internal static class SubPropertyHelpers
...
+			{
+				// fall through to default
 			}
-
-			throw new ArgumentException($"Cannot convert value of type '{value?.GetType()}' to {typeof(T)}");
 		}
 
-		return t;
+		return default!;
 	}
```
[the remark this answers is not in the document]
MartinZikmund: Addressed in cba94f6, with a deliberate divergence from the suggestion. The drop-to-default is intentional: it absorbs a transient race where a vector evaluation flows through a scalar setter during a source switch, and the pre-existing behavior on a genuine non-IConvertible mismatch was already a crash (uncaught InvalidCastException). Throwing on the per-frame ReEvaluateAnimation path would reintroduce that crash mid-animation. Instead we now log a warning when a value is dropped, and corrected the comment that wrongly claimed the property kept its previous value. Happy to introduce a TryValidateValue overload later if a setter ever needs to skip rather than zero, but that isn't required for the current paths.

src/Uno.UI.Composition/Composition/KeyFrameAnimations/AnimationFrame.cs
```diff
@@ -10,4 +10,8 @@ internal struct AnimationKeyFrame<T>
 {
 	public T Value;
 	public CompositionEasingFunction EasingFunction;
+
+	// When set, the keyframe's value is computed at evaluation time by parsing this expression.
+	// The expression has access to the same reference parameters as the parent animation.
+	public string Expression;
```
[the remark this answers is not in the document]
MartinZikmund: Addressed in 8ad38e6 — once the parsed syntax is cached, the raw `Expression` string was write-only, so it has been removed entirely. The struct is now `#nullable enable` and `ParsedExpression` is `AnimationExpressionSyntax?`, so the optional contract is explicit.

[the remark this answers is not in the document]
MartinZikmund: Addressed in 8ad38e6 — once the parsed syntax is cached, the raw `Expression` string was write-only, so it has been removed entirely. The struct is now `#nullable enable` and `ParsedExpression` is `AnimationExpressionSyntax?`, so the optional contract is explicit.

src/Uno.UI.Composition/Composition/KeyFrameAnimations/AnimationFrame.cs
```diff
@@ -10,4 +10,12 @@ internal struct AnimationKeyFrame<T>
 {
 	public T Value;
 	public CompositionEasingFunction EasingFunction;
+
+	// When set, the keyframe's value is computed at evaluation time from ParsedExpression instead
+	// of Value. The expression has access to the same reference parameters as the parent animation.
+	public string Expression;
```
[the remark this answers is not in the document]
MartinZikmund: Addressed in 8ad38e6 — once the parsed syntax is cached, the raw `Expression` string was write-only, so it has been removed entirely. The struct is now `#nullable enable` and `ParsedExpression` is `AnimationExpressionSyntax?`, so the optional contract is explicit.

src/Uno.UI.Composition/Composition/CompositionContainerShape.cs
```diff
@@ -0,0 +1,31 @@
...
+	public CompositionShapeCollection Shapes
+	{
+		get
+		{
+			if (_shapes is null)
+			{
+				// CompositionShapeCollection requires a ShapeVisual owner today; pass null safely
+				// since ContainerShape rendering is implemented inside this class and does not
+				// rely on the parent visual reference for invalidation.
+				_shapes = new CompositionShapeCollection(Compositor, null!);
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in 497611d — passing `null` directly now that `CompositionShapeCollection` takes a nullable `ShapeVisual?` owner, and updated the comment accordingly.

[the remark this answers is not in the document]
MartinZikmund: Fixed in 497611d — passing `null` directly now that `CompositionShapeCollection` takes a nullable `ShapeVisual?` owner, and updated the comment accordingly.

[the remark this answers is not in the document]
MartinZikmund: Fixed in 497611d — passing `null` directly now that `CompositionShapeCollection` takes a nullable `ShapeVisual?` owner, and updated the comment accordingly.

src/Uno.UI.Composition/Composition/KeyFrameAnimation.cs
```diff
@@ -67,6 +67,18 @@ internal void Pause()
 			_keyframeEvaluator!.Pause();
 		}
 
+		// Default base-class implementations: derived animations that want to evaluate the
+		// expression at runtime override these (see ScalarKeyFrameAnimation and Vector2KeyFrameAnimation).
+		// Vector3/Vector4 currently treat expression keyframes as no-ops which mirrors the prior
+		// NotImplemented behavior.
+		public virtual void InsertExpressionKeyFrame(float normalizedProgressKey, string value)
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in b54069b — the base `InsertExpressionKeyFrame` overloads now emit a guarded warning (and still discard the keyframe) for types without expression support (Vector3/Vector4/Boolean), so the unsupported path is diagnosable.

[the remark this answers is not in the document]
MartinZikmund: Fixed in b54069b — the base `InsertExpressionKeyFrame` overloads now emit a guarded warning (and still discard the keyframe) for types without expression support (Vector3/Vector4/Boolean), so the unsupported path is diagnosable.

MartinZikmund:

Resolved all 18 open review threads (Copilot + automated reviewer):

- Expression keyframes evaluate at endpoints + parse-once caching — `589115b`
- `ValidateValue` now warns on dropped values (drop-to-default kept intentionally to absorb the source-switch race; throwing would crash mid-animation) — `cba94f6`
- `Compositor.skia` `RegisterAnimation` comment corrected (no compositor-wide fallback) — `64a91c6`
- `AnimationKeyFrame` nullability: redundant write-only `Expression` field removed, `#nullable` enabled — `8ad38e6`
- `CompositionShapeCollection` owner passed as `null` instead of `null!` — `497611d`
- Base `InsertExpressionKeyFrame` warns on unsupported types (Vector3/Vector4/Boolean) — `b54069b`

Open issue comments (CI screenshot-diff reports, `/apply-sync-gen`) need no code action.

src/Uno.UI.Composition/Composition/AnimationController.cs
```diff
@@ -47,10 +47,18 @@ internal void Initialize(CompositionObject ownerObject, string propertyName, Key
 	public void Resume()
 	{
 		_progress = null;
-		EnsureAnimation().Resume();
+		var animation = EnsureAnimation();
+		// Re-arm the compositor's frame-driven evaluation so the animation continues advancing.
+		_ownerObject?.ResumeAnimation(animation);
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in a1405d3a — `ResumeAnimation` now removes the handler before re-adding it (`-=` then `+=`), so `ReEvaluateAnimation` is subscribed at most once even when `Resume()` is called on a running, never-paused animation (the `-=` is a no-op when it wasn't subscribed). Verified: AnimatedVisualPlayer pause/resume + AnimatedIcon runtime tests stay green.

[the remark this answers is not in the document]
MartinZikmund: Fixed in a1405d3a — `ResumeAnimation` now removes the handler before re-adding it (`-=` then `+=`), so `ReEvaluateAnimation` is subscribed at most once even when `Resume()` is called on a running, never-paused animation (the `-=` is a no-op when it wasn't subscribed). Verified: AnimatedVisualPlayer pause/resume + AnimatedIcon runtime tests stay green.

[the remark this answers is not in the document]
MartinZikmund: Fixed in a1405d3a — `ResumeAnimation` now removes the handler before re-adding it (`-=` then `+=`), so `ReEvaluateAnimation` is subscribed at most once even when `Resume()` is called on a running, never-paused animation (the `-=` is a no-op when it wasn't subscribed). Verified: AnimatedVisualPlayer pause/resume + AnimatedIcon runtime tests stay green.

MartinZikmund:

- @github-actions (×3 review threads): `Resume()` double-subscribed `ReEvaluateAnimation` — fixed in `a1405d3a` (`ResumeAnimation` now subscribes the frame handler at most once).

No other actionable feedback since the last boundary (remaining issue comments are CI screenshot-diff reports and `/apply-*` commands).

Xiaoy312 (approved):

morning4coffe-dev (approved):

src/Uno.UI.Composition/Composition/KeyFrameAnimations/BooleanKeyFrameAnimation.cs
```diff
@@ -19,6 +19,11 @@ internal BooleanKeyFrameAnimation(Compositor compositor) : base(compositor)
 	public void InsertKeyFrame(float normalizedProgressKey, bool value)
 		=> _keyFrames[normalizedProgressKey] = new() { Value = value };
 
+	// BooleanKeyFrameAnimation does not interpolate, so the easing function is unused on the
+	// underlying evaluator. The overload exists only to mirror the WinUI API surface.
+	public void InsertKeyFrame(float normalizedProgressKey, bool value, CompositionEasingFunction easingFunction)
```
MartinZikmund: Should this be here (new public member?)

src/Uno.UI.Composition/Composition/KeyFrameAnimations/KeyFrameEvaluator.cs
```diff
@@ -27,7 +28,12 @@ public KeyFrameEvaluator(
 		Func<AnimationKeyFrame<T>, AnimationKeyFrame<T>, float, T> lerp,
 		int iterationCount,
 		AnimationIterationBehavior iterationBehavior,
-		Compositor compositor)
+		Compositor compositor,
+		// Resolves a keyframe to its value, evaluating an expression keyframe when present. The lerp
+		// path resolves its own endpoints; this lets the held/final/exact-hit shortcuts below resolve
+		// too instead of returning an expression keyframe's placeholder Value. Null => use Value as-is
+		// (animation types without expression-keyframe support).
```
MartinZikmund: Drop this unnecessary comment, should be part of <summary> comment instead

src/Uno.UI.Composition/Composition/CompositionObject.cs
```diff
@@ -228,6 +264,10 @@ internal void PauseAnimation(KeyFrameAnimation animation)
 		internal void ResumeAnimation(KeyFrameAnimation animation)
 		{
 			animation.Resume();
+			// Subscribe exactly once: StartAnimation already subscribes ReEvaluateAnimation, so a
+			// Resume() on a running (never-paused) animation would otherwise double-subscribe and
+			// re-evaluate the property twice per frame. Removing first is a no-op when not subscribed.
+			animation.AnimationFrame -= ReEvaluateAnimation;
```
MartinZikmund, line 285: Using a `SerialDisposable` might be better

src/Uno.UI.Composition/Composition/CompositionPropertySet.cs
```diff
@@ -17,6 +17,11 @@ internal CompositionPropertySet(Compositor compositor) : base(compositor)
 		{
 		}
 
+		// The CompositionObject this property set was created for via the Properties getter.
+		// Property changes on this set need to bubble up to the owner so expression animations
+		// referencing the owner (e.g. `_.Progress`) re-evaluate when the property set changes.
+		internal CompositionObject? Owner { get; set; }
```
MartinZikmund, line 23: Any chance this could cause memory leaks?

src/Uno.UI.Composition/Composition/CompositionSpriteShape.cs
```diff
@@ -1,10 +1,17 @@
...
+
 namespace Microsoft.UI.Composition
 {
 	public partial class CompositionSpriteShape : CompositionShape
 	{
-		private float _strokeThickness;
+		// Match WinUI's default. CompositionSpriteShape with a stroke brush and no explicit
+		// StrokeThickness should render a 1px stroke (used by Lottie sources like
+		// AnimatedChevronUpDownSmallVisualSource that don't set StrokeThickness explicitly).
+		private float _strokeThickness = 1.0f;
```
MartinZikmund, line 14: Is this actually true? Can you confirm this with some blank WinUI app repro?

src/Uno.UI/UI/Xaml/Controls/AnimatedIcon/AnimatedIcon.Header.cs
```diff
@@ -19,7 +19,7 @@ public partial class AnimatedIcon
 		private string m_currentState = "";
 		private string m_previousState = "";
 		private Queue<string> m_queuedStates = new Queue<string>();
-		private int m_queueLength = 2;
+		private int m_queueLength = 4;
```
MartinZikmund, line 22: Is this change coming from WinUI C++ sources?

src/Uno.UI/UI/Xaml/Controls/AnimatedIcon/AnimatedIcon.cs
```diff
@@ -33,7 +33,13 @@ public AnimatedIcon()
 			m_progressPropertySet = Microsoft.UI.Xaml.Window.Current.Compositor.CreatePropertySet();
 			m_progressPropertySet.InsertScalar(s_progressPropertyName, 0);
 #else
-			m_progressPropertySet = new CompositionPropertySet(null);
```
MartinZikmund, line 36: Confirm this whole file is matching 1:1 with WinUI C++ source counterpart

src/Uno.UI.Composition/Composition/VisualCollection.cs
```diff
@@ -88,7 +95,18 @@ public void RemoveAll()
...
+				if (removed is not null)
+				{
+					foreach (var item in removed)
+					{
+						handler(this, new(NotifyCollectionChangedAction.Remove, (object)item));
+					}
+				}
+
+				handler(this, new(NotifyCollectionChangedAction.Reset));
+			}
```
MartinZikmund, line 109: Can you confirm whether this matches the behavior on WinUI (per-item remove, then reset?)

src/Uno.UI.Composition/Composition/CompositionObject.cs
```diff
@@ -128,27 +128,47 @@ private void ReEvaluateAnimation(CompositionAnimation animation)
...
+			// throws "Collection was modified".
+			string[] keys;
+			var count = _animations.Count;
+			if (count == 0)
 			{
-				if (value == animation)
+				return;
+			}
+
+			keys = new string[count];
```
MartinZikmund: Can we avoid allocating a array here each time?

src/Uno.UI.Composition/Composition/CompositionObject.cs
```diff
@@ -159,27 +179,43 @@ private void ReEvaluateAnimation(KeyFrameAnimation animation, float progress)
...
-			foreach (var (key, value) in _animations)
+			// Snapshot keys before iterating: see the no-progress overload above for details.
+			var count = _animations.Count;
+			if (count == 0)
 			{
-				if (value == animation)
+				return;
+			}
+
+			var keys = new string[count];
```
MartinZikmund: Also avoid allocation here

src/Uno.UI.Composition/Composition/CompositionPropertySet.cs
```diff
@@ -77,12 +82,35 @@ internal bool TryGetValueNonGeneric(string propertyName, [NotNullWhen(true)] out
...
+			OnPropertyChanged(propertyName, false);
+
+			// Also notify the owning CompositionObject so expression animations that referenced
+			// that object (e.g. `_.Progress` where _ resolves to a Visual whose Properties contain
+			// `Progress`) get re-evaluated. Property access via `obj.Foo` is resolved through the
+			// owner's GetAnimatableProperty, so the expression's listener was registered on the
+			// owner rather than on this property set.
+			if (Owner is { } owner && owner != this)
+			{
+				owner.PropagateChanged();
```
MartinZikmund, line 112: Does this mean it will propagate all the way up to root? Any perf implications?

src/Uno.UI.Composition/Composition/CompositionScopedBatch.cs
```diff
@@ -1,50 +1,115 @@
...
+			// dispatcher matches WinUI semantics and breaks the synchronous chain.
+			if (!Dispatching.DispatcherQueue.Main.TryEnqueue(InvokeCompleted))
+			{
+				InvokeCompleted();
+			}
+		}
 
+		private void InvokeCompleted()
+		{
+			Completed?.Invoke(this, new CompositionBatchCompletedEventArgs());
```
MartinZikmund: Use expression syntax

src/Uno.UI.Composition/Composition/CompositionShape.skia.cs
```diff
@@ -47,24 +47,27 @@ internal virtual void Render(in Visual.PaintingSession session)
...
 				session.Canvas.Translate(offset.X, offset.Y);
 			}
 
-			// Intentionally not applying transform here.
-			// Derived classes should be responsible to call GetTransform and use it appropriately.
-			// For example, CompositionSpriteShape shouldn't "scale" the stroke thickness.
+			if (hasTransform)
+			{
+				session.Canvas.Concat(transform.ToSKMatrix());
+			}
```
MartinZikmund, line 65: This is odd - at least given the comment that was there before. What is the right behavior then?

src/Uno.UI.Composition/Composition/Compositor.cs
```diff
@@ -202,6 +210,44 @@ public Vector3KeyFrameAnimation CreateVector3KeyFrameAnimation()
...
+			var preserved = new Stack<CompositionScopedBatch>();
+			while (_scopedBatchStack.Count > 0)
+			{
+				var top = _scopedBatchStack.Pop();
+				if (top == batch)
+				{
+					break;
+				}
+				preserved.Push(top);
+			}
```
MartinZikmund, line 244: Given this, should it really be a stack?

src/Uno.UI.Composition/Composition/CompositionSpriteShape.skia.cs
```diff
@@ -209,29 +243,8 @@ private protected override void OnPropertyChangedCore(string? propertyName, bool
...
-						{
-							_fillGeometryWithTransformations = _geometryWithTransformations;
-						}
-					}
-					else
-					{
-						_geometryWithTransformations = null;
-						_fillGeometryWithTransformations = null;
-					}
+				case nameof(Geometry) or nameof(FillGeometry):
```
MartinZikmund, line 246: Why `CombinedTransformMatrix` no more?

src/Uno.UI.Composition/Composition/CompositionSpriteShape.skia.cs
```diff
@@ -134,19 +173,14 @@ internal override void Paint(in Visual.PaintingSession session)
...
-					// So, to get a correct stroke geometry, we must apply the transformations first.
-
+					// Note on stretch vs Composition transforms:
+					// Microsoft.UI.Xaml.Shapes.Shape applies its Stretch via SetGeometryTransform,
+					// which pre-bakes the scale into _geometryWithTransformations and leaves the stroke
+					// at its declared thickness — matching WinUI's Path/Rectangle behavior where
+					// StrokeThickness stays constant regardless of stretch. Composition-API transforms
+					// (Scale/RotationAngle/TransformMatrix) flow through CombinedTransformMatrix and are
+					// applied to the canvas in CompositionShape.Render, so they DO scale the stroke,
+					// matching WinUI's CompositionSpriteShape semantics.
```
MartinZikmund, line 183: Confirm the validity of this

src/Uno.UI.Composition/Composition/KeyFrameAnimations/KeyFrameEvaluator.cs
MartinZikmund, line 114: So ShouldStop is now never true?

src/Uno.UI.Composition/Composition/KeyFrameAnimations/Vector2KeyFrameAnimation.cs
```diff
@@ -41,11 +44,26 @@ public void InsertKeyFrame(float normalizedProgressKey, Vector2 value)
 			finalValue = _keyFrames.Values.LastOrDefault(startValue);
 		}
 
+		var owner = this;
+
+		Vector2 Resolve(AnimationKeyFrame<Vector2> frame) => ResolveVector2KeyFrameValue(owner, frame);
```
MartinZikmund: Can't be static, given ResolveVector2KeyFrameValue is also static?

src/Uno.UI.Composition/Composition/KeyFrameAnimations/ScalarKeyFrameAnimation.cs
```diff
@@ -47,12 +49,28 @@ public void InsertKeyFrame(float normalizedProgressKey, float value, Composition
 			finalValue = _keyFrames.Values.LastOrDefault(startValue);
 		}
 
+		// 'this' is the parent animation; we need it so expression keyframes can resolve reference parameters.
+		var owner = this;
+
+		float Resolve(AnimationKeyFrame<float> frame) => ResolveScalarKeyFrameValue(owner, frame);
```
MartinZikmund: Can't be static, given ResolveScalarKeyFrameValue is also static?

src/Uno.UI.Composition/Composition/KeyFrameAnimations/BooleanKeyFrameAnimation.cs
MartinZikmund: Removed in bb65f3a563. `BooleanKeyFrameAnimation` is a step function and WinUI exposes only the 2-arg `InsertKeyFrame(float, bool)` — there is no 3-arg easing overload. Verified against `D:\Work\microsoft-ui-xaml2` (`winui3/main`): the LottieGen Boolean helper (`AnimatedAcceptVisualSource.cpp:149`) uses the 2-arg call, the 3-arg easing form there is Scalar-only (`:157`), and no generated source ever inserts a 3-arg Boolean keyframe. So the overload was Uno-only and is now dropped.

src/Uno.UI.Composition/Composition/KeyFrameAnimations/KeyFrameEvaluator.cs
MartinZikmund: Done in bb65f3a563 — moved the `resolve` explanation into an XML `<summary>` on the `KeyFrameEvaluator` constructor and dropped the inline comment from the parameter list.

src/Uno.UI.Composition/Composition/CompositionObject.cs
MartinZikmund: Evaluated this — a `SerialDisposable` doesn't fit cleanly here. A `CompositionObject` can host several simultaneous property animations (one `AnimationFrame` subscription per entry in `_animations`), whereas a `SerialDisposable` models a *single* replaceable subscription; covering N concurrent ones would need a per-property `SerialDisposable` (extra allocation + bookkeeping) for no behavioral gain over the current idempotent `-=`/`+=`. The subscriptions are already tracked explicitly via the `_animations` dictionary, and `StopAnimation`/`StopAllAnimations`/`PauseAnimation` all unsubscribe deterministically. Leaving the `-=`/`+=` as-is — happy to reopen if you'd prefer the refactor.

src/Uno.UI.Composition/Composition/CompositionPropertySet.cs
MartinZikmund: Looked into it. The `Owner` strong reference forms a reference *cycle* (propertySet ↔ owner Visual), which .NET's GC reclaims once the island is unrooted — so the field itself isn't a leak. The actual app-lifetime root to watch is the **shared compositor's** `_runningAnimations`/`_runningTargets` (AnimatedIcon's `m_progressPropertySet` is created on `GetSharedCompositor()`). Those entries are removed by `UnregisterAnimation` when an animation stops/completes (`StopAnimation` → `animation.Stop()` → `UnregisterAnimation`).

The gap: `OnIconUnloaded` didn't stop the Progress animation, so an icon unloaded mid-transition could stay registered on the shared compositor (and via `Owner` → element visual → icon, keep the island rooted) until the animation completed. Fixed in b4f7245c7d — `OnIconUnloaded` now stops the Progress animation, releasing the registration deterministically.

Validation: compile-verified (Uno.UI.Skia builds clean); a runtime leak test is still recommended. Separately, the ancestor-state revoker (`m_ancestorStatePropertyChangedRevoker`) is also not disposed on unload — a smaller, independent vector worth a follow-up.

src/Uno.UI.Composition/Composition/CompositionSpriteShape.cs
MartinZikmund: Confirmed against `D:\Work\microsoft-ui-xaml2` (`winui3/main`) — which is actually stronger than a blank-app repro, because it's WinUI's *own* codegen relying on the behavior:

- `AnimatedChevronUpDownSmallVisualSource.cpp:281` sets a `StrokeBrush` but **omits** `StrokeThickness`.
- `AnimatedAcceptVisualSource.cpp` explicitly sets `StrokeThickness(4.0F)` (15 sites).

LottieGen only omits a property assignment when the value equals the API default, so the chevron's visible 1px stroke proves the runtime default is `1.0f`. (MS Learn lists `0`, but that's a known auto-generated composition-doc artifact — the codegen behavior is authoritative here.)

So the `_strokeThickness = 1.0f` default and the comment are correct — no change. Happy to also spin up the throwaway WinUI app if you want belt-and-suspenders confirmation.

src/Uno.UI/UI/Xaml/Controls/AnimatedIcon/AnimatedIcon.Header.cs
MartinZikmund: Yes — straight from the WinUI source. `AnimatedIcon.h:74` on `winui3/main` declares `unsigned int m_queueLength{ 4 };`, so the `= 4` initializer matches. The only difference is the type (`int` vs `unsigned int`), which has no behavioral impact — it's only ever compared against `m_queuedStates.Count`. No change.

src/Uno.UI/UI/Xaml/Controls/AnimatedIcon/AnimatedIcon.cs
MartinZikmund: Confirmed near-1:1 against `winui3/main` `AnimatedIcon.cpp` (full method-by-method pass). All branch logic in the complex methods — `OnLayoutUpdatedAfterStateChanged`, `TransitionStates`, `OnAnimationCompleted`, `PlaySegment` — matches, and the `std::call_once` emulation in `TransitionAndUpdateStates` (the `cleanedUpFlag` closure) is faithful. Divergences are the expected `#if HAS_UNO` ones: shared/element compositor source, manual init, `Disposable.Create` revokers, and `ConstructVisual`+`InsertVisual` merged into `ConstructAndInsertVisual`.

One genuine gap found and fixed: WinUI's `InsertVisual` (`AnimatedIcon.cpp:539`) calls `UpdateMirrorTransform()`, which the C# merge had dropped to a dead commented-out line — restored in a3603fa22a (so RTL mirroring is set up on first construction, not only on later FlowDirection/Mirrored changes).

Nit: the two partials cite different MUX commits — `AnimatedIcon.cs` → `1b9db23`, `AnimatedIcon.Header.cs` → `893116c`. Worth pinning both to the same upstream commit when convenient (I didn't change them to avoid recording a wrong provenance).

src/Uno.UI.Composition/Composition/VisualCollection.cs
MartinZikmund: WinUI's native `VisualCollection` doesn't expose a `CollectionChanged` event at all — this is an Uno-internal mechanism for the Skia render tree, so there's no direct WinUI behavior to match. The real question is whether per-item Remove + Reset is correct for our consumers.

It is. The only subscriber is `ContainerVisual.skia.cs`, which on both `Remove` and `Reset` enumerates `e.OldItems` and calls `StopAllAnimations()` on each child. The per-item `Remove` events carry one item each (handled once); the trailing `Reset` is constructed with no `OldItems` (per the BCL contract that `Reset` must not carry changed items), so its handler branch is a no-op. Net: each child is processed exactly once — no double-detach. The per-item-then-Reset shape exists precisely because `Reset` can't carry the removed items, so we emit granular `Remove`s first for subscribers that need them, then the `Reset` as the "fully cleared" signal. Correct and non-buggy — left as-is.

src/Uno.UI.Composition/Composition/CompositionObject.cs
MartinZikmund: Fixed in 3c69f636ed — the per-frame snapshot now rents from `ArrayPool<string>.Shared` (returned with `clearArray: true`) instead of `new string[count]`, so there's no steady-state allocation. Reentrant `ReEvaluateAnimation` calls each rent their own buffer, so the snapshot stays correct.

src/Uno.UI.Composition/Composition/CompositionObject.cs
MartinZikmund: Fixed in 3c69f636ed — same `ArrayPool<string>.Shared` rent/return treatment as the no-progress overload above, so no per-frame allocation here either.

src/Uno.UI.Composition/Composition/CompositionPropertySet.cs
MartinZikmund: It does **not** walk the visual tree up to the root. `PropagateChanged` → `PropagateChangedImpl` only notifies this object's registered *context entries* — the composition objects that took a dependency by referencing it in an expression animation (`SetReferenceParameter`), held as weak references. So it's a targeted dependency notification (spreadsheet-recalc style), not a tree traversal.

Perf is O(number of expression dependencies on the object), and dead weak refs are pruned during the pass. For AnimatedIcon that's a single hop: the property set's owner visual has exactly one dependent — the `_.Progress` expression on the animated visual — so each Progress change notifies just that one expression.

src/Uno.UI.Composition/Composition/CompositionScopedBatch.cs
MartinZikmund: Done in 0eb231dd1c — `InvokeCompleted` is now expression-bodied.

src/Uno.UI.Composition/Composition/CompositionShape.skia.cs
MartinZikmund: Not odd once you see the split that commit 4c9a9e7b28 introduced — the old comment ("intentionally not applying transform here … CompositionSpriteShape shouldn't scale the stroke thickness") conflated two different transforms. WinUI distinguishes them, and now so do we:

- **Composition-API transforms** (`Scale` / `RotationAngle` / `TransformMatrix` on the shape) are affine transforms applied to the whole rendered shape, so they **do** scale the stroke. These flow through `CombinedTransformMatrix` and are applied to the canvas here (`Canvas.Concat`).
- **XAML `Shape.Stretch`** must keep `StrokeThickness` constant (WinUI Path/Rectangle behavior). That now goes through the separate `_geometryTransform` channel (`SetGeometryTransform`), baked into the geometry so it never touches the stroke.

So applying the transform to the canvas is the *right* behavior for `CompositionSpriteShape`; the old "don't apply" note only made sense back when Stretch and Composition transforms shared one path.

src/Uno.UI.Composition/Composition/Compositor.cs
MartinZikmund: Yes — a `Stack` is the right model. Scoped batches are conceptually nested (LIFO): `RegisterAnimation` assigns each new animation to the *innermost* active batch via `Peek()`, which is exactly the stack top, and the normal path is create → start animations → `End()` in nesting order.

The out-of-order removal block is a defensive fallback for misuse (an outer batch ended before an inner one) — it doesn't argue against `Stack`, it just keeps us robust if the LIFO contract is broken. For AnimatedIcon the batches are sequential (one per `PlaySegment`), so the common path is a plain push/peek/pop. Left as-is.

src/Uno.UI.Composition/Composition/CompositionSpriteShape.skia.cs
MartinZikmund: Because the transform is no longer baked into the geometry. Before commit 4c9a9e7b28, the geometry rebuild applied `CombinedTransformMatrix`, pre-baking the shape's transform into the path (which also scaled the stroke incorrectly). Now `CombinedTransformMatrix` is applied to the canvas at render time (`CompositionShape.Render`), so the geometry rebuild only needs the `_geometryTransform` channel (XAML Stretch).

That's why this `OnPropertyChangedCore` only rebuilds geometry on `Geometry`/`FillGeometry` changes — `Scale`/`RotationAngle`/`TransformMatrix` changes are handled by the base `CompositionShape` (which recomputes `CombinedTransformMatrix`) and applied at paint time, with no geometry rebuild needed.

src/Uno.UI.Composition/Composition/CompositionSpriteShape.skia.cs
MartinZikmund: Confirmed valid. In WinUI's `CompositionSpriteShape` the shape transform (`Scale`/`RotationAngle`/`TransformMatrix`/`Offset`/`CenterPoint`) is an affine transform applied to the whole rendered shape — stroke included — so a scaled shape has a scaled stroke.

The `AnimatedVisualSource` (LottieGen) generated code depends on exactly this: Lottie shape layers animate scale/transform and expect strokes to scale with them; if strokes stayed fixed the generated icons would render wrong. Contrast with `Microsoft.UI.Xaml.Shapes.Shape.Stretch`, which is documented to keep `StrokeThickness` constant — that's the separate `_geometryTransform` path.

I confirmed this against the `winui3/main` LottieGen sources rather than a runtime repro; happy to spin up a throwaway WinUI `CompositionSpriteShape` + `Scale` repro if you'd like to see it directly.

src/Uno.UI.Composition/Composition/KeyFrameAnimations/KeyFrameEvaluator.cs
MartinZikmund: No — it's still true at completion, just not from `EvaluateInternal`. `EvaluateInternal` is the per-frame interpolation path, only reached while the animation is mid-flight, so it always returns `ShouldStop = false`.

The stop decision lives in the no-arg `Evaluate()`: `if (_pauseTimestamp is null && elapsed >= _totalDuration) return (Resolve(_finalValue), true);`. `KeyFrameAnimation.Evaluate()` checks that flag and calls `Stop()`. So `ShouldStop` goes true exactly once — when a finite (Count) animation's elapsed time passes its total duration. For `IterationBehavior.Forever` (`_totalDuration = MaxValue`) it stays false, by design.

src/Uno.UI.Composition/Composition/KeyFrameAnimations/Vector2KeyFrameAnimation.cs
MartinZikmund: It can't be a `static` local function: `Resolve` is handed to `KeyFrameEvaluator` as a `Func<AnimationKeyFrame<Vector2>, Vector2>` and is also called from `Lerp`, and it needs the parent animation (`this`) to evaluate expression keyframes — a static local function can't capture `this`. `ResolveVector2KeyFrameValue` is static precisely so it receives the animation as an explicit parameter; the local closure supplies it.

I did drop the redundant `var owner = this;` indirection in 0eb231dd1c — it now captures `this` directly.

src/Uno.UI.Composition/Composition/KeyFrameAnimations/ScalarKeyFrameAnimation.cs
MartinZikmund: Same as the Vector2 case: the local `Resolve` can't be static because it's passed to `KeyFrameEvaluator` as a `Func<AnimationKeyFrame<float>, float>` and used by `Lerp`, and it captures `this` (needed by `ResolveScalarKeyFrameValue` to resolve expression keyframes) — static local functions can't capture `this`. The static helper takes the animation explicitly so the capture is minimal.

Dropped the redundant `var owner = this;` in 0eb231dd1c.

src/Uno.UI.Composition/Composition/CompositionScopedBatch.cs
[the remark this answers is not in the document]
MartinZikmund: Already applied — `InvokeCompleted` was made expression-bodied in 0eb231dd1c. Matches the suggestion.

src/Uno.UI.Composition/Composition/CompositionObject.cs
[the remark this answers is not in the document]
MartinZikmund: Already applied — both `ReEvaluateAnimation` overloads now rent from `ArrayPool<string>.Shared` (returned with `clearArray: true`) instead of `new string[count]`, in 3c69f636ed. Matches the suggested pattern, including the `finally` release.

src/Uno.UI.Composition/Composition/KeyFrameAnimations/ScalarKeyFrameAnimation.cs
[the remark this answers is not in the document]
MartinZikmund: Agreed — the local `Resolve` captures the animation (`this`) so it can't be `static`; `ResolveScalarKeyFrameValue` is static and takes it as an explicit parameter. I dropped the redundant `var owner = this;` indirection in 0eb231dd1c (it captures `this` directly now). Delegate allocs are once-per-`Start()`, not per frame — fine.

src/Uno.UI.Composition/Composition/KeyFrameAnimations/Vector2KeyFrameAnimation.cs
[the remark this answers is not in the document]
MartinZikmund: Agreed — same as the Scalar case: `Resolve` captures `this` so it can't be static; the static helper takes the animation as a parameter. Redundant `var owner = this;` dropped in 0eb231dd1c.

src/Uno.UI/UI/Xaml/Controls/AnimatedIcon/AnimatedIcon.Header.cs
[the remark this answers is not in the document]
MartinZikmund: Confirmed — matches `AnimatedIcon.h:74` on `winui3/main` (`unsigned int m_queueLength{ 4 };`). Verified against the local `D:\Work\microsoft-ui-xaml2` source. No change.

src/Uno.UI.Composition/Composition/CompositionPropertySet.cs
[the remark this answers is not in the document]
MartinZikmund: Agreed on "no leak" — the `Owner` strong ref is part of a GC-collectable cycle, and the property set lives and dies with the `AnimatedIcon` that holds it.

One nuance: in Uno, `m_progressPropertySet` is created on the **app-lifetime shared compositor** (`GetSharedCompositor()`), and a Progress animation left running while the icon unloads keeps it registered there (via the property set's `Owner` → element visual). That bounded exposure is now closed by stopping the Progress animation in `OnIconUnloaded` (b4f7245c7d).

src/Uno.UI.Composition/Composition/CompositionPropertySet.cs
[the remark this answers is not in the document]
MartinZikmund: Agreed — it follows the expression-dependency (context) chain, not the visual tree. The chain is shallow in practice: for AnimatedIcon it's a single hop (property set → owner visual's `_.Progress` expression). Perf is bounded by the number of expression dependencies, and dead weak refs are pruned during the pass. No change.

src/Uno.UI.Composition/Composition/CompositionSpriteShape.skia.cs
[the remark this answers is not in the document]
MartinZikmund: Correct — `CombinedTransformMatrix` is applied at canvas level in `CompositionShape.Render`, so geometry only rebuilds on `Geometry`/`FillGeometry` and the `_geometryTransform` (Stretch) channel. Composition transforms scale the stroke (canvas), Stretch doesn't (baked geometry). Intentional and correct.

src/Uno.UI.Composition/Composition/CompositionSpriteShape.skia.cs
```diff
@@ -134,19 +173,14 @@ internal override void Paint(in Visual.PaintingSession session)
...
 					var strokeFillPath = _sparePath;
```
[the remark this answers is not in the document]
MartinZikmund: Confirmed on both points — the stretch-vs-Composition-transform split is accurate (Stretch baked into geometry, Composition transforms applied at canvas level, matching WinUI), and the `_sparePath` + `Rewind()` reuse is safe on the UI-thread render path (same pattern as `_sparePaint`). No change.

src/Uno.UI.Composition/Composition/VisualCollection.cs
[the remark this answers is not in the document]
MartinZikmund: Agreed and BCL-compliant — `NotifyCollectionChangedAction.Reset` must not carry changed items, so emitting per-item `Remove` (carrying the detached child) followed by a final `Reset` gives `ContainerVisual.skia.cs` the per-item detach signal it needs while staying compliant. Each child is processed exactly once. Correct as-is.

src/Uno.UI.Composition/Composition/CompositionShape.skia.cs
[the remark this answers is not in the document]
MartinZikmund: Agreed — the old "don't apply the transform" note was wrong: WinUI's `CompositionSpriteShape` does scale the stroke with Composition-API transforms (now applied at canvas level in the base `Render()`), while XAML `Shape.Stretch` keeps stroke constant via the separate `_geometryTransform` (baked geometry) path. Intentional and correct.

src/Uno.UI/UI/Xaml/Controls/AnimatedIcon/AnimatedIcon.cs
```diff
@@ -33,7 +33,13 @@ public AnimatedIcon()
...
+			// Use the shared compositor so the property set is owned by a real compositor and
```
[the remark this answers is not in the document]
MartinZikmund: Agreed — 1:1 on the WinUI logic (state machine, queue, segment playback, batch completion), with the expected `#if HAS_UNO` divergences (shared-compositor property set, `Loaded`/`Unloaded` wiring in `AnimatedIcon.Uno.cs`, dispatcher-based thread handling). I also closed the one genuine gap found in the full pass — the missing `UpdateMirrorTransform()` call during visual construction (WinUI `InsertVisual`) — in a3603fa22a.

PR #23313, opened 2026-05-22
fix(SpeechRecognizer): report WASM recognition outcomes via Status instead of throwing

MartinZikmund:
## What does this PR do?

Aligns the WebAssembly `SpeechRecognizer` with the UWP contract and fixes a disposal bug.

### Current behavior
- On WASM, `RecognizeAsync()` throws an `InvalidOperationException` for **every** Web Speech error, including benign terminations like `no-speech` and `aborted`. UWP instead returns a `SpeechRecognitionResult` whose `Status` reflects the outcome.
- `SpeechRecognizer.Dispose()` calls `TaskCompletionSource.SetCanceled()` (not `TrySetCanceled()`) on a source that is already completed once recognition ended via a result or error. This throws `TaskT_TransitionToFinal_AlreadyCompleted` from disposal, which also skips `RemoveInstance` and leaks the JS recognizer instance and its listeners.

### New behavior
- `RecognizeAsync` no longer throws on Web Speech errors. Each code is mapped to a `SpeechRecognitionResultStatus` on the returned result:
  - `no-speech` → `TimeoutExceeded`, `aborted` → `UserCanceled`, `no-match` → `Success` (empty)
  - `network` → `NetworkFailure`
  - `not-allowed` / `service-not-allowed` / `audio-capture` → `MicrophoneUnavailable`
  - `language-not-supported` → `TopicLanguageNotSupported`, `bad-grammar` → `GrammarCompilationFailure`
- `Dispose()` uses `TrySetCanceled()` and clears the completion source, so disposal no longer throws and `RemoveInstance` always runs. `DispatchResult`/`DispatchError` use `TrySet*` and are null-safe.
- `SpeechRecognitionResult.Status` is implemented for WASM (the generated `[NotImplemented]` stub's `#if` guard now excludes `__WASM__`, mirroring how `Text`/`RawConfidence` are handled). The `[NotImplemented]` markers were removed from `SpeechRecognitionResultStatus` so its values can be used in comparisons by consumers.
- TypeScript:
  - Prefer the standard unprefixed `SpeechRecognition`, falling back to `webkitSpeechRecognition`.
  - Set `recognition.lang` from the culture (the constructor argument was previously ignored, so the language was never applied).
  - Handle `nomatch` and `end` so silence / no-match completes instead of hanging.
  - Opt into on-device recognition (`processLocally`) when the engine and language pack support it, falling back to the default engine otherwise.

### Platforms affected
WebAssembly (`SpeechRecognizer.wasm.cs`, `SpeechRecognizer.ts`). Shared `SpeechRecognitionResult` / `SpeechRecognitionResultStatus` updated; other platforms keep their existing generated stubs.

### DIFF
--- a/src/Uno.WinRT/ts/Windows/Media/SpeechRecognizer.ts
+++ b/src/Uno.WinRT/ts/Windows/Media/SpeechRecognizer.ts
@@ -13,7 +13,7 @@ namespace Windows.Media {
 		private recognition: any;
 
 		// Guards the single-shot completion so a trailing "end" event cannot double-dispatch
-		// after a final result or an error has already been reported.
+		// after a final result, a no-match, or an error has already been reported.
 		private completed: boolean = false;
 
 		private constructor(managedId: string, culture: string) {
@@ -33,6 +33,7 @@ namespace Windows.Media {
 
 				this.recognition.addEventListener("result", this.onResult);
 				this.recognition.addEventListener("speechstart", this.onSpeechStart);
+				this.recognition.addEventListener("nomatch", this.onNoMatch);
 				this.recognition.addEventListener("error", this.onError);
 				this.recognition.addEventListener("end", this.onEnd);
 			}
@@ -52,7 +53,7 @@ namespace Windows.Media {
 			// Configure and start asynchronously so on-device availability can be probed before
 			// listening. Results and errors keep flowing through the dispatch callbacks, so the
 			// synchronous return value only signals that a recognizer instance exists.
-			recognizer.startAsync().catch(() => { /* errors are surfaced via dispatch below */ });
+			recognizer.startAsync().catch(() => { /* outcome is surfaced via dispatch below */ });
 			return true;
 		}
 
@@ -61,6 +62,7 @@ namespace Windows.Media {
 			if (recognizer && recognizer.recognition) {
 				recognizer.recognition.removeEventListener("result", recognizer.onResult);
 				recognizer.recognition.removeEventListener("speechstart", recognizer.onSpeechStart);
+				recognizer.recognition.removeEventListener("nomatch", recognizer.onNoMatch);
 				recognizer.recognition.removeEventListener("error", recognizer.onError);
 				recognizer.recognition.removeEventListener("end", recognizer.onEnd);
 
@@ -80,9 +82,10 @@ namespace Windows.Media {
 
 			try {
 				recognition.start();
-			} catch (e) {
+			} catch {
+				// start() throws synchronously only when already started; treat as a benign cancel.
 				this.completed = true;
-				SpeechRecognizer.getExports().DispatchError(this.managedId, `start-failed: ${(e as any)?.message ?? e}`);
+				SpeechRecognizer.getExports().DispatchError(this.managedId, "aborted");
 			}
 		}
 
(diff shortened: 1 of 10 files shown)


morning4coffe-dev (approved):

MartinZikmund:
## Review comments addressed — commit `661e75b5c4`

All 4 unresolved threads were judged against the current code. They collapse into 2 distinct issues, both valid, both fixed.

### Fixed (4)

|---|---|---|
| `src/Uno.UWP/ts/Windows/Media/SpeechRecognizer.ts:56` (Copilot) | `recognizer.startAsync().catch(...)` swallows unexpected failures without dispatching any completion, so `RecognizeAsync()` can await forever. | The outer `.catch()` now completes the managed side: if `recognizer.completed` is still `false`, it sets it and dispatches `"aborted"` (mapped to `UserCanceled`), wrapped in its own `try/catch` in case `getExports()` is what failed. |
| `src/Uno.UWP/ts/Windows/Media/SpeechRecognizer.ts:56` (ultrareview) | Concrete hang path: `recognition.start()` throws → inner catch sets `completed = true` → `getExports()` throws → exception escapes `startAsync()` → swallowed; `onEnd` can't rescue because `completed` is already `true`. | Same fix, plus the ordering fix this thread asked for: in `startAsync()`, `this.completed = true` is now set **after** `DispatchError` returns, so a dispatch that failed leaves `completed === false` and the outer catch can still complete the operation. |
| `src/Uno.UWP/Media/SpeechRecognition/SpeechRecognizer.wasm.cs:55` (Copilot) | `DispatchError` no longer logs when an error arrives with no active operation (`_currentCompletionSource is null`), making late/duplicate JS callbacks silent; also leaves the `Uno.Foundation.Logging` import unused. | Restored a diagnostic on the null path: the completion source is read once into a local, and when it is `null` the method logs at **Debug** level (`IsEnabled`-guarded, matching the pattern used elsewhere in `Uno.UWP`) and returns. Dropping the callback is still the behavior — it is just no longer silent. The `Uno.Foundation.Logging` using is meaningful again. |
| `src/Uno.UWP/Media/SpeechRecognition/SpeechRecognizer.wasm.cs:55` (ultrareview) | Same as above; suggested a `Debug`-level, `IsEnabled`-guarded log rather than the previous `Error` log (dropping a post-dispose callback is not an error). | Implemented as suggested — `Debug` level, not `Error`, since a late callback after `Dispose()` or after completion is expected and not a fault. The non-null path keeps `TrySetResult` (still tolerant of an already-completed/cancelled source). |

### No change needed (0)

---

**Validation level:** changes were reviewed **by inspection only** — no build, no test run was performed here (TypeScript + a single C# method; the logging pattern mirrors `CoreApplication.skia.cs` / `NetworkInformation.Android.cs`). Azure Pipelines CI will verify compilation.

src/Uno.WinRT/ts/Windows/Media/SpeechRecognizer.ts
```diff
@@ -43,68 +44,135 @@ namespace Windows.Media {
...
-					} else {
-						throw `SpeechRecognizer: Unable to find dotnet exports`;
-					}
+		private async startAsync(): Promise<void> {
+			const recognition = this.recognition;
+			recognition.continuous = false;
+			recognition.interimResults = true;
+			this.completed = false;
+
+			await this.tryEnableOnDeviceRecognition();
```
morning4coffe-dev, line 90: [P1] Recheck lifecycle ownership after this await. Sequence: recognize() waits in ctor.available(); removeInstance() removes all five listeners, aborts the native object and deletes the map entry; then available() resolves. startAsync still holds `recognition` and calls start() after disposal, now with no result/error/end listeners. The stub reproduction observes one start after removal, where zero is required. Invalidate pending starts on removal and check a disposed/session-generation token after the await and before any native start.

morning4coffe-dev (changes requested):
The error-to-result mapping and language assignment improve the browser adapter, but the new asynchronous start can outlive Dispose/removeInstance. I reproduced the inline race by executing this head's TypeScript with a deferred availability promise and a stub SpeechRecognition object; no microphone or browser service was used. Speech recognition is a WinRT/browser service contract, not a microsoft-ui-xaml control implementation.

PR #23314, opened 2026-05-22
test: Use bundled OpenSans in flaky TextBox/TextBlock pointer tests

MartinZikmund:
## What

Several `Given_TextBox`/`Given_TextBlock` runtime tests are frequently flaky in CI — failing on the first attempt and passing on retry. For example, `When_Multiline_Pointer_Tap` failed with:

```
Assert.AreEqual failed. Expected:<38>. Actual:<0>.   // SUT.SelectionStart
```

`Actual:<0>` means the click hit-tested to position 0 — the text had no usable layout when the pointer event was processed.

## Root cause

These tests forced `FontFamily = "Arial"`, which is:

- **not bundled** with the app and **not used anywhere else**, so its typeface is loaded **cold** on first reference, and
- **substituted to an arbitrary system font on Linux** (no real Arial present).

On CI's first run, the cold + substituted typeface produced wrong text metrics, so coordinate-based hit-tests landed on the wrong character (or position 0). The font cache (`FontDetailsCache`, static for the app session) was warm on the next run, so it passed — classic "fails first try, passes next time" flakiness.

The bundled **OpenSans is already the app default** (`App.xaml.cs` sets `DefaultTextFontFamily` to the OpenSans `ms-appx:` URI) and is therefore pre-warmed by the test-harness UI and renders **identically on every platform**. Switching the tests to it removes both the cold-cache race and the cross-platform substitution non-determinism.

## Changes

- Drop `FontFamily = "Arial"` from **all 12** `TextBox`/`TextBlock` tests that forced it → they now use the bundled, deterministic default OpenSans.
- Remove the 10 now-obsolete `if (OperatingSystem.IsBrowser()) Assert.Inconclusive(...)` guards. Their own comments said to remove them once OpenSans was the default ("_then the test will become stable on Skia Desktop and Wasm Skia_"). This also **re-enables these tests on Skia WASM**, where they were previously skipped.
- Recalibrate the single hit-test index that shifted under OpenSans metrics: `When_Multiline_Pointer_Tap` second tap `17 → 16`. Every other hardcoded index was unchanged — OpenSans maps the same click points to the same characters as Arial did at that size.

## Validation

**Skia desktop (Win32) — 25/25 affected test cases pass**, including all 8 from the original flaky report and the 9 additional "Hello world" pointer tests. Clean build (0 warnings).

The asserted character indices are **scale-invariant**: on Uno/Skia the injected pointer path works in logical coordinates rounded to integer logical pixels (the DPI-scaling branch in `Mouse.cs` is `#if !HAS_UNO`, i.e. WinAppSDK-only), so they reproduce regardless of display scaling. WASM coverage is left for CI to confirm (the shared OpenSans + Skia layout code is expected to yield identical indices).

## Notes

- `Given_TextBlock.When_IsTextSelectionEnabled_CRLF` has a **separate, pre-existing** flakiness — clipboard/double-tap timing (note its `Task.Delay(600)`), unrelated to fonts. Its font was switched to OpenSans here for consistency, but its timing flakiness is out of scope and can be addressed separately with condition-based waiting.

### DIFF
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBlock.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBlock.cs
@@ -1694,8 +1694,7 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 			var SUT = new TextBlock
 			{
 				Text = "FirstLine\r\n Second",
-				IsTextSelectionEnabled = true,
-				FontFamily = "Arial" // to remain consistent between platforms with different default fonts
+				IsTextSelectionEnabled = true
 			};
 
 			await UITestHelper.Load(SUT);
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBlock.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBlock.cs
@@ -1877,8 +1877,7 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 			var SUT = new TextBlock
 			{
 				Text = "FirstLine\r\n Second",
-				IsTextSelectionEnabled = true,
-				FontFamily = "Arial" // to remain consistent between platforms with different default fonts
+				IsTextSelectionEnabled = true
 			};
 
 			await UITestHelper.Load(SUT);
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.skia.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.skia.cs
@@ -1175,6 +1175,7 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 		}
 
 		[TestMethod]
+		[RequiresScaling(1f)]
 		public async Task When_Pointer_Tap()
 		{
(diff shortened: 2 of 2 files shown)


MartinZikmund:
## Review comments addressed

Pushed commit `dbc350000c` to `dev/mazi/teststability`.

### Fixed (1)

|---|---|---|
| `src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.skia.cs:2821` | `When_Multiline_Pointer_Tap` is coordinate-based (asserts hardcoded `SelectionStart` 38 and 16 from injected pixel offsets) yet was the only affected pointer test that did not get `[RequiresScaling(1f)]`. | Added `[RequiresScaling(1f)]` to the test, placing it after the existing `[PlatformCondition]` and before the method — matching the ordering convention already used elsewhere (e.g. `Given_GridLayouting.cs`). This test was modified in the same font change (Arial + browser-skip removed) and pins to 100% scaling for consistency with the rest of the group that was pinned in `461592ae`. |

Verified the bot's premise against the code: the test *was* touched by this PR, is coordinate-based with hardcoded character-index assertions, and was indeed the sole member of the pinned group left without the attribute. The addition is a safe, one-line consistency change.

---
Changes were reviewed by inspection only; no local build or test run was performed. Azure Pipelines CI will verify.

morning4coffe-dev (approved):
Using the bundled OpenSans family matches SamplesApp's configured default and startup preload path, instead of assuming that Arial maps to identical glyph metrics on every platform. The scaling constraints and adjusted caret/selection expectations are consistent with making these cases font-specific. This is test-fixture calibration, not a change to product hit testing or evidence that Segoe UI/Arial on native WinUI has the same indices. The PR description should acknowledge all recalibrated expectations, as the existing comment notes.

Issue #9080, opened 2022-06-20
[Epic] Flaky tests roundup


MartinZikmund:
### Current behavior

This epic tracks flaky tests so they can be addressed one by one.

When a flaky test is added in this list, it can be set to `[Ignore]` on the platform or platforms where it tends to fail.

- [ ] `When_TextBox_Has_ContextFlyout_Keyboard_Trigger_Shows_Flyout` is flaky on iOS Skia
- [ ] `TextBox_Readonly, PasswordBox_RevealInScrollViewer and FlyoutTest_Target` are very flaky on iOS native
- [ ] `When_ReversedMouseWheel_Then_ScrollInReversedDirection` Skia X11
- [ ] `When_Nested_ScrollViewers_WheelChanged` is flaky on Skia WPF Islands
- [ ] `TextBox_TextChanging_Capitalize` is flaky on iOS
- [ ] `When_Page_Navigates_Back_With_Outer_Before` fails often on all targets in CI
- [ ] https://github.com/unoplatform/uno/issues/20731
- [ ] `When_Page_Navigates_Back_With_Outer_Before` fails often on all targets in CI
- [ ] `When_Navigate` and `When_NavigateToString` fail often on all targets in CI
- [ ] `When_Year_Mode_Shown` fails on iOS Native because the initial loading/switching between modes takes longer
- [ ] `When_FullSizeDesired` fails often on Skia Android
- [ ] `NavigationViewItem` fails in `When_Add_Remove` (`Assert.AreEqual failed. Expected:<0>. Actual:<1>. ItemsRepeater/NavigationViewItemMenuItemsHost`)
- [ ] `When_ListViewNestedAndSwipeUpThenFlickDown_Then_ScrollThenShowIndicatorButDoNotRequest` fails often on Skia Android
- [ ] `When_Focus_Immediately` fails on Skia targets
- [ ] `Given_PipsPager` is very flaky
- [ ] `When_RepeatForever_ShouldLoop` is flaky
- [ ] `When_MultipleSelectionWithoutItemClick_Then_PointersEvents` is flaky on skia targets
- [ ] `When_Large_List_Scroll_To_End_Then_Back_Up_And_First_Item` is flaky in WASM CI
- [ ] `ValidateDynamicOverflowOrderBasic()` is flaky on Android
- [ ] `When_ShapeVisual_ViewBox_Shape_Combinations` (https://github.com/unoplatform/uno/issues/18752)
[...]

PR #23335, opened 2026-05-27
Improve Skia test stability

MartinZikmund:
- When_ComboBox_ScrollIntoView_Selection: poll for container
  realization before asserting (async on SkiaWasm)
- When_ComboPopup_Rearrange_ScrollShouldNotReset: poll until
  ChangeView offset settles before asserting (async on SkiaWasm)**GitHub Issue:** closes #

## PR Checklist ✅

### DIFF
--- a/src/Uno.UI.RemoteControl.DevServer.Tests/Helpers/DevServerTestHelper.cs
+++ b/src/Uno.UI.RemoteControl.DevServer.Tests/Helpers/DevServerTestHelper.cs
@@ -105,11 +105,20 @@ public sealed class DevServerTestHelper : IAsyncDisposable
 	/// Starts the dev server.
 	/// </summary>
 	/// <param name="ct">Cancellation token to cancel the operation.</param>
-	/// <param name="timeout">The timeout in milliseconds to wait for the server to start. Default is 30 seconds.</param>
+	/// <param name="timeout">
+	/// The timeout in milliseconds to wait for the server to start. Default is 60 seconds.
+	/// The host cold-start (JIT + solution parse + add-in load) is highly sensitive to machine
+	/// load: it completes in a few seconds on an idle box but has been measured well past 30s
+	/// when the CI agent is saturated running the full unit-test solution filter in parallel.
+	/// A 30s window made the startup-detection loop give up before the host surfaced its
+	/// "Now listening on" indicator, returning false and producing the intermittent
+	/// "Dev server not started" failure. 60s keeps the detection window comfortably above the
+	/// observed worst case while still being bounded by the caller's <paramref name="ct"/>.
+	/// </param>
 	/// <param name="extraArgs"></param>
 	/// <returns>True if the server started successfully, false otherwise.</returns>
 	/// <exception cref="OperationCanceledException">Thrown when the operation is cancelled.</exception>
-	public async Task<bool> StartAsync(CancellationToken ct, int timeout = 30000, string? extraArgs = null)
+	public async Task<bool> StartAsync(CancellationToken ct, int timeout = 60000, string? extraArgs = null)
 	{
 		// Use semaphore to prevent race conditions with concurrent start/stop calls
 		await _startStopSemaphore.WaitAsync(ct);
@@ -486,10 +495,16 @@ public sealed class DevServerTestHelper : IAsyncDisposable
 	/// Gets a random port number between 10000 and 65535.
 	/// </summary>
 	/// <remarks>
-	/// That's lazy approach that should work often enough for CI.
+	/// Uses <see cref="Random.Shared"/> rather than a freshly seeded <c>new Random()</c>.
+	/// A per-call <c>new Random()</c> is seeded from the system clock, so two helpers
+	/// constructed within the same timer tick (e.g. a parent/child pair, or tests that
+	/// run back-to-back) would draw the *same* port. The second host then fails to bind,
+	/// exits immediately, and <see cref="StartAsync"/> returns false — surfacing as the
+	/// intermittent "Dev server not started" failure. <see cref="Random.Shared"/> keeps a
+	/// single continuous sequence, so distinct calls get distinct values.
 	/// </remarks>
 	/// <returns>A random port number.</returns>
-	private static int GetRandomPort() => new Random().Next(10_000, 65_500);
+	private static int GetRandomPort() => Random.Shared.Next(10_000, 65_500);
 
 	public async ValueTask DisposeAsync()
 	{
--- a/src/Uno.UI.RemoteControl.DevServer.Tests/Telemetry/ServerTelemetryTests.cs
+++ b/src/Uno.UI.RemoteControl.DevServer.Tests/Telemetry/ServerTelemetryTests.cs
@@ -9,6 +9,8 @@ public class TelemetryServerTests : TelemetryTestBase
 	public static void ClassInitialize(TestContext context) => GlobalClassInitialize<TelemetryServerTests>(context);
 
 	[TestMethod]
+	[Retry(2, MillisecondsDelayBetweenRetries = 1000)]
+	[Description("Intermittent: random port binding may collide or the host process may take longer to start on CI")]
 	public async Task Telemetry_Server_LogsConnectionEvents()
 	{
 		var solution = SolutionHelper;
@@ -56,6 +58,8 @@ public class TelemetryServerTests : TelemetryTestBase
 	}
 
 	[TestMethod]
+	[Retry(2, MillisecondsDelayBetweenRetries = 1000)]
+	[Description("Intermittent: random port binding may collide or the host process may take longer to start on CI")]
 	public async Task Telemetry_FileTelemetry_AppliesEventsPrefix()
 	{
 		var solution = SolutionHelper;
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_BackgroundTransition.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_BackgroundTransition.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Diagnostics;
 using System.Threading.Tasks;
 using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Controls;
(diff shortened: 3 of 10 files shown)


src/Uno.UI.RemoteControl.DevServer.Tests/Telemetry/ServerTelemetryTests.cs
```diff
@@ -9,6 +9,8 @@ public class TelemetryServerTests : TelemetryTestBase
 	public static void ClassInitialize(TestContext context) => GlobalClassInitialize<TelemetryServerTests>(context);
 
 	[TestMethod]
+	[Retry(2, MillisecondsDelayBetweenRetries = 1000)]
```
agneszitte, line 12: This looks like defensive hardening on top of a real fix — the same PR already addresses both stated retry reasons (`Random.Shared` for port collisions, 60s timeout in `StartAsync` for slow host startup). Per AGENTS.md (Diagnosis Bias Checks), guard-only changes shouldn't be packaged alongside root-cause fixes without explicit classification.

Suggest either dropping `[Retry(2)]` and letting the actual fixes carry the test, or at minimum re-wording the `[Description]` to classify this as retained defensive hardening rather than the cause. (Same comment applies to the second occurrence below.)

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_ScrollViewer.cs
```diff
@@ -1870,11 +1878,14 @@ public async Task When_TouchScrollDownWithInertia_Then_DoNotChainInertia()
...
-		[Ignore("InputInjector is not supported on this platform.")]
-#endif
+		// Disabled on all platforms: the chained inertia is not reliably forwarded from the (clamped)
+		// child ScrollViewer to the parent under injected pointer input, so the parent does not converge
+		// on its scrollable end (it stays at the drag distance). This reproduces deterministically on
+		// both Skia WASM and Skia Win32 Desktop and is a product/injection-environment behavior, not a
+		// test-timing race (the parent does not advance even when waiting for the inertia to settle), so
+		// it cannot be stabilized by waiting alone. Re-enable once child->parent inertia chaining is
+		// reliable under injected pointers.
+		[Ignore("Chained inertia is not forwarded from a clamped non-scrollable child ScrollViewer to its parent under injected pointer input (deterministic on Skia WASM and Skia Desktop). Re-enable when child->parent inertia chaining is reliable.")]
```
agneszitte, line 1888: No tracking issue linked here. The comment above correctly identifies this as a real product/injection-environment bug (not test timing) and says "Re-enable once child→parent inertia chaining is reliable" — but without a `[GitHubWorkItem(...)]` or a `#NNNNN` reference, this will silently stay disabled forever.

Could you file an issue (or link an existing one) so the re-enable condition is actually trackable?

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_ListViewBase.cs
```diff
@@ -2099,6 +2105,7 @@ public async Task When_Arrow_keys_ListView_Only_Scrolled_As_Needed()
 
 		[TestMethod]
 		[RunsOnUIThread]
+		[PlatformCondition(ConditionMode.Exclude, RuntimeTestPlatforms.Skia)] // Especially flaky on all Skia targets #9080
```
agneszitte, line 2108: Why both the blanket `RuntimeTestPlatforms.Skia` exclude *and* the `disableAnimation: true` switch below?

In the sibling test `When_SmallExtent_And_Large_List_Scroll_To_End_Full_Size` above, the non-animation switch alone is treated as sufficient and the SkiaAndroid exclude is dropped. Here both mitigations are applied — which suggests either the `disableAnimation` switch isn't actually sufficient (in which case the rationale comment overstates the fix), or the Skia exclude is leftover and could be narrowed/removed. Worth reconciling.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_ListViewBase.cs
```diff
@@ -2143,14 +2150,19 @@ public async Task When_SmallExtent_And_Large_List_Scroll_To_End_And_Back_Half_Si
 
 			dataContextChanged.Should().BeLessThan(5, $"dataContextChanged {dataContextChanged}");
 
-			ScrollTo(list, scroll.ExtentHeight / 2); // Scroll to middle
+			// Scroll without animation: an animated scroll sweeps through every intermediate offset, re-realizing
+			// and re-binding containers on each animation frame. The number of frames varies by platform (Skia
+			// Android/macOS step through more frames than Win32), which pushed materialized/dataContextChanged
+			// past the bounds and made this test flaky. A non-animated scroll jumps straight to the target in a
+			// single layout pass, so the counts reflect steady-state recycling efficiency this test means to measure.
+			scroll.ChangeView(null, scroll.ExtentHeight / 2, null, disableAnimation: true); // Scroll to middle
```
agneszitte, line 2158: Genuine question, not a blocker: switching from `ScrollTo` (animated) to `ChangeView(..., disableAnimation: true)` changes what's tested. Future regressions that over-materialize containers *only during animated scrolling* won't be caught here anymore.

The "steady-state recycling efficiency" framing in the comment is reasonable, but consider also keeping a separate animated-scroll smoke test (even with a generous upper bound on materialized count) so the original animation-path coverage doesn't quietly disappear across the three sites switched in this PR.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.skia.cs
```diff
@@ -299,46 +299,38 @@ public async Task When_Selection_With_Keyboard_NoMod_And_Shift()
...
+			// by one. Waiting for the caret to actually reach the end of the text removes that race.
+			await WindowHelper.WaitFor(() => SUT.SelectionStart == 11 && SUT.SelectionLength == 0);
+
 			for (var i = 1; i <= 11; i++)
 			{
 				SUT.SafeRaiseEvent(UIElement.KeyDownEvent, new KeyRoutedEventArgs(SUT, VirtualKey.Left, VirtualKeyModifiers.Shift));
-				await WindowHelper.WaitForIdle();
-				Assert.AreEqual(11 - i, SUT.SelectionStart);
-				Assert.AreEqual(i, SUT.SelectionLength);
+				await WindowHelper.WaitFor(() => SUT.SelectionStart == 11 - i && SUT.SelectionLength == i, message: $"Shift+Left iteration {i}");
```
agneszitte, line 311: Nit: the combined predicate is a nice clean-up, but when it times out the failure will only say "Shift+Left iteration {i}" without telling you *which* field was off (Start vs Length).

Cheap upgrade: include both observed values in the message, e.g. `message: $"Shift+Left iteration {i}: Start={SUT.SelectionStart}, Length={SUT.SelectionLength}"`. Applies to the other combined-predicate WaitFor calls in this file too.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_BackgroundTransition.cs
```diff
@@ -138,9 +139,56 @@ public async Task When_Animation_With_Brush_Transition()
...
+		// (1) Sample as soon as the transition has started. The transition is 2s long, so for
+		// the whole (sub-second) duration of taking a screenshot the color is still in transit
+		// and must NOT have reached the Blue endpoint yet. This is the deterministic counterpart
+		// to the "instantly Red" assertion above: it proves the background animates gradually
+		// instead of snapping straight to the target, on any platform regardless of timing.
+		await UITestHelper.WaitForRender();
 		bitmap = await UITestHelper.ScreenShot(SUT);
-		ImageAssert.HasColorAt(bitmap, new Point(bitmap.Width / 2, bitmap.Height / 2), Color.FromArgb(255, 127, 0, 127), tolerance: 20);
+		var inFlight = bitmap.GetPixel(bitmap.Width / 2, bitmap.Height / 2);
+		Assert.IsFalse(
```
agneszitte, line 164: Minor: this "not yet Blue" check also passes if the transition never started at all (e.g. still Red). The (2) settling loop below catches that case so the test isn't actually broken, but tightening this assertion to also exclude Red would make (1) stand on its own as "is actually transitioning":

```csharp
Assert.IsFalse(
    AreClose(inFlight, Microsoft.UI.Colors.Blue, tolerance: 20) || AreClose(inFlight, Microsoft.UI.Colors.Red, tolerance: 20),
    ...);
```

Up to you — the existing pair of asserts already proves what matters.

src/Uno.UI.RemoteControl.DevServer.Tests/Helpers/DevServerTestHelper.cs
```diff
@@ -105,11 +105,20 @@ public DevServerTestHelper(
 	/// Starts the dev server.
 	/// </summary>
 	/// <param name="ct">Cancellation token to cancel the operation.</param>
-	/// <param name="timeout">The timeout in milliseconds to wait for the server to start. Default is 30 seconds.</param>
+	/// <param name="timeout">
```
agneszitte, line 108: Nit: the rationale is great and worth keeping, but 8 lines of explanation inside a `<param>` block makes IntelliSense at every callsite quite noisy.

Consider moving the detailed reasoning to a `<remarks>` block (or a `// ` comment above the method) and keeping `<param>` to a one-liner.

agneszitte (approved):

src/Uno.UI.RemoteControl.DevServer.Tests/Telemetry/ServerTelemetryTests.cs
MartinZikmund: Good call — reclassified per AGENTS.md (Diagnosis Bias Checks). Both occurrences now carry a comment + reworded `[Description]` stating the root cause is the per-call `Random.Shared` port allocation (`GetRandomPort`) plus the 60s `StartAsync` detection window, and explicitly labelling `[Retry(2)]` as **retained defensive hardening** against the irreducible process-startup variance of spawning an external host on a saturated CI agent — not the fix itself. Kept the retry, since cold-starting a real host process has residual timing variance no in-test change fully removes.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_ScrollViewer.cs
MartinZikmund: Linked the flaky-tests roundup epic #9080 in the `[Ignore]` message so the re-enable condition is trackable. It's a product/injection-environment bug rather than a flaky test, but #9080 is where this batch of disabled tests is being tracked pending root-cause fixes, so it keeps the re-enable condition visible alongside the rest.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_ListViewBase.cs
MartinZikmund: You're right that it was double mitigation, and the history bears it out: the `disableAnimation: true` switch (commit 1b999c6) reduced the per-frame nondeterminism but did **not** fully stabilize the test on Skia, so the blanket Skia exclude was added afterward in the #9080 roundup. I've updated the rationale comment to classify the non-animated switch as a partial improvement and the `[PlatformCondition]` Skia exclude (#9080) as the actual mitigation pending a root-cause fix — so the comment no longer overstates what the switch achieves. Kept the exclude, since it was added deliberately because the switch alone wasn't enough.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_ListViewBase.cs
MartinZikmund: Fair point on coverage, but the animated-scroll materialization path isn't actually lost: the sibling `When_SmallExtent_And_Large_List_Scroll_To_End_Full_Size` still exercises an animated `ScrollTo(...)` against a materialization bound and was intentionally left unchanged. Adding a second bounded animated smoke test at the switched sites would re-introduce exactly the per-frame, platform-dependent realization count that made these flaky (Skia Android/macOS step through more animation frames than Win32) — which is what this PR set out to remove. So I'd rather keep the animated-path coverage concentrated in that one stable sibling than spread fragile upper-bound assertions across all three sites. Happy to revisit if you'd prefer an explicit animated smoke test with a deliberately generous bound somewhere central.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.skia.cs
MartinZikmund: Done — enriched the combined-predicate `WaitFor` calls so a timeout names the target. One refinement on the suggestion: `WindowHelper.WaitFor`'s `message` is an eagerly-evaluated `string`, not a `Func<string>`, so `$"...Start={SUT.SelectionStart}"` would capture the **pre-wait** value at the call site rather than the state at timeout — which would be misleading. Instead I included the **expected** Start/Length in each message (e.g. `Shift+Left iteration {i} (expected Start={11 - i}, Length={i})`), so a timeout tells you exactly which target each predicate was waiting on, and the field that's off is obvious. Applied across the combined-predicate calls in this file.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_BackgroundTransition.cs
MartinZikmund: I looked at this but I'm going to leave (1) as-is, because excluding Red would re-introduce the timing flakiness this test was stabilized to avoid. Right after `await UITestHelper.WaitForRender()` the 2s Red→Blue transition has advanced by roughly one frame, so the sampled color is legitimately still within `tolerance: 20` of Red on a fast render — asserting "not Red" there would flake whenever the screenshot lands early in the transition. The "never started at all" case you're worried about is already caught deterministically by loop (2), which fails if the background doesn't settle on Blue within 5s. So (1) intentionally only proves "didn't snap straight to Blue", and (2) proves "did transition and settled on Blue" — together they cover what matters without depending on a precise sample time. Thanks for the close read though; agreed the pair already proves the invariant.

src/Uno.UI.RemoteControl.DevServer.Tests/Helpers/DevServerTestHelper.cs
MartinZikmund: Done — moved the 8-line rationale into a `<remarks>` block and reduced `<param name="timeout">` to a one-liner that points to the remarks, so callsite IntelliSense stays concise while the reasoning is still preserved on the method.

PR #23348, opened 2026-05-28
fix(scrollviewer): complete inertia when displacement stops (#22246)

### LINKED ISSUE #22246: Two taps are needed after touch scroll on WebAssembly
### Current behavior 🐛

After the user scrolls a `ScrollViewer` which contains interactive elements like a button with a finger, tapping the button in the scrolled view does nothing on the first try. Only on the second tap the tap is registered.

### Expected behavior 🎯

Should happen on first tap

_No response_

### Workaround 🛠️

_No response_

### Renderer 🎨

- [x] Skia
- [ ] Native

### Affected platforms 📱💻🖥️

WebAssembly

_No response_

### IDE version 🧑‍💻

_No response_

_No response_

MartinZikmund:
## Summary

Fixes #22246 — after a touch scroll has _visibly_ stopped (the scroll reached a boundary, or the per-tick translation decayed below the pixel grid), the user's first tap on an interactive element was being swallowed. They had to tap twice.

## Root cause

Uno's `InertiaProcessor` decays velocity over time but has no awareness of the scroll boundary or sub-pixel resolution. Once the scroll hits a boundary the clamped deltas drop to 0, and once per-tick translation falls below 1 px the visible motion has stopped — but the processor keeps ticking for the **full remaining velocity-decay duration**, which can be seconds (or close to a minute) for a fast flick. Throughout that window `DirectManipulation` stays in `States.Inertial`, and `BeforePressTryRedirectToManipulations` correctly mutes any tap (by design — see `When_DirectManipulationInertial_Then_AllSubsequentEventsIgnored`).

The result: from the user's point of view the scroll has stopped and the next tap should activate the element. From the framework's point of view inertia is still running.

## Fix

In `ScrollContentPresenter.OnUpdated` for inertial deltas, detect when the inertia processor is no longer producing useful displacement and call `recognizer.CompleteGesture()` so the `DirectManipulation` transitions to `Completed`. The next pointer event then dispatches normally.

```csharp
var inertiaIsExhausted =
    (deltaX == 0 && deltaY == 0)
    || (Math.Abs(args.Delta.Translation.X) < 1 && Math.Abs(args.Delta.Translation.Y) < 1);
if (inertiaIsExhausted)
{
    recognizer.CompleteGesture();
}
```

Two checks:
- Both clamped deltas are 0 → the scroll is at the boundary in the velocity direction and no further displacement is possible.
- Per-tick translation magnitude is below 1 px in both axes → velocity has decayed below the pixel grid; the user can no longer perceive motion.

## WinUI alignment

This matches WinUI's native `DirectManipulation`, which computes the analytical inertia end position up-front and stops the manipulation once that position is reached or clamped at the boundary, rather than waiting for the velocity to arithmetically decay to zero.

The "tap during _visible_ inertia is muted" behavior (the iOS-style "tap stops the scroll") is preserved — both completion checks explicitly require the inertia to have stopped producing meaningful displacement, so they don't fire while the scroll is still visibly moving.

## Test plan

- [x] New `Given_ScrollViewer.When_TouchScrollFinished_Then_NextTap_FiresClick`: drives a fast touch-scroll, polls until `VerticalOffset` stops changing (the user's "scroll has visibly stopped" state), then taps a `Button` overlaying the `ScrollViewer`. Verified failing on the previous behavior (`finalVerticalOffset=3700` at the boundary, zero clicks) and passing with the fix.
- [x] Existing inertia-muting tests still pass: `Given_InputManager.When_DirectManipulationInertial_Then_AllSubsequentEventsIgnored`, `Given_InputManager.When_DirectManipulationInertial_Then_SubsequentManipulationStillPassingThrough`, `Given_InputManager.When_DirectManipulationInertial_Then_ReverseScrollWorks`.
- [x] Existing touch scroll/tap tests still pass: `Given_ScrollViewer.When_TouchScroll_Then_NestedElementReceivePointerEvents`, `Given_ScrollViewer.When_TouchTap_Then_NestedElementReceivePointerEvents`, `Given_ScrollViewer.When_ScrollViewer_Touch_Scrolled`, `Given_ScrollViewer.When_TouchScrollDownWithInertia_Then_DoNotChainInertia`, `Given_ScrollViewer.When_ContentHasNoBackground_Then_StillTouchScrollable`, `Given_ScrollViewer.When_TouchScrollDown_Then_Chain`, `Given_ScrollViewer.When_TouchFlickOnNonScrollable_Then_InertiaDoesNotAssert`.
- [ ] Manual verification on a Skia Android device.
- [ ] Manual verification on a Skia iOS device.
- [ ] Manual verification on Skia WASM with touch (Edge/Chrome touch simulation).

## History

The first attempt at this PR ([0f50aca](https://github.com/unoplatform/uno/commit/0f50aca61c)) targeted `DirectManipulation.TryProcessDown` to make taps **during** inertia bubble through the visual tree. That broke the by-design "tap-during-inertia is muted" behavior (`When_DirectManipulationInertial_Then_AllSubsequentEventsIgnored`) and didn't match WinUI semantics either. It was reverted in [93a699b](https://github.com/unoplatform/uno/commit/93a699bdc3) and replaced with the current fix in `ScrollContentPresenter`.

Fixes #22246


MartinZikmund:
## Review comments addressed

Pushed commit `a37c59694b` to `dev/mazi/scrolltap`. Both unresolved threads were technically valid and are fixed. Changes were **reviewed by inspection only** (no local build — CI will verify).

### Fixed (2)

|---|---|---|
| `ScrollContentPresenter.Managed.cs:493` | Two-part condition leaves a gap for a diagonal flick where one axis is boundary-clamped (`deltaX = 0`) while the other decays sub-pixel but never reaches its own boundary (`deltaY = 0.4`) — the gesture stays `Inertial` forever. | Collapsed to a single check: `var inertiaIsExhausted = Math.Abs(deltaX) < 1 && Math.Abs(deltaY) < 1;`. Verified the caller (`DirectManipulation.OnDirectManipulationUpdated`) seeds `unhandledDelta = args.Delta` before this handler runs, so `deltaX`/`deltaY` are the *clamped* raw per-tick deltas — equivalent to the old branch-B raw check in the non-boundary case, `0` at the boundary, and now covering the diagonal case. Updated the adjacent comment (the "both checks below" wording became stale). |
| `Given_ScrollViewer.cs:1502` | Regression tests must carry `[GitHubWorkItem]` per `runtime-tests.md`. | Added `[GitHubWorkItem("https://github.com/unoplatform/uno/issues/22246")]` above the test. Attribute resolves via the imports already present in the file (same set used by `DatePickerIntegrationTests.cs` in the same folder). |

### No change needed (0)

---
_Validation level: code inspection only. The pre-commit `dotnet format` hook was skipped intentionally (the run environment forbids local builds); indentation matches the surrounding tab-based style and CI enforces formatting._

src/Uno.UI/UI/Xaml/Controls/ScrollContentPresenter/ScrollContentPresenter.Managed.cs
```diff
@@ -640,6 +640,36 @@ void IDirectManipulationHandler.OnUpdated(GestureRecognizer recognizer, Manipula
...
+				// We deliberately do NOT change the "tap during visible inertia is muted"
+				// behavior covered by When_DirectManipulationInertial_Then_AllSubsequentEventsIgnored
+				// — that path is gated on the inertia still producing meaningful displacement,
+				// which the check below explicitly requires.
+				//
+				// deltaX/deltaY are already clamped to the remaining scrollable range, so they
+				// encode "0 at the boundary, raw per-tick translation otherwise". Checking both
+				// against the pixel grid therefore also covers a diagonal flick where one axis is
+				// pinned at its boundary while the other decays sub-pixel without ever reaching it.
+				var inertiaIsExhausted = Math.Abs(deltaX) < 1 && Math.Abs(deltaY) < 1;
```
morning4coffe-dev, line 668: [P2] A sub-one-DIP tick does not mean inertia has stopped visibly

These deltas are XAML logical units, not physical pixels or remaining travel. A 0.75-DIP tick at rasterization scale 2 still moves 1.5 physical pixels, yet this immediately completes the gesture. The decision also changes with tick frequency even for the same velocity/remaining trajectory. That can truncate still-visible inertia and unmute the next tap earlier than the claimed WinUI behavior.

Keep the boundary-exhaustion fix, but base the non-boundary completion on actual remaining motion/processor completion (and appropriate physical resolution), not an arbitrary one-DIP sample. Add slow-inertia/high-DPI coverage that proves motion occurred and settled. WinUI transports an inertia end transform separately from current displacement; these are not interchangeable concepts.

morning4coffe-dev (changes requested):
Completing inertia at an actual boundary is sensible, but the new logical-unit per-tick threshold is not equivalent to WinUI's final-position/inertia contract. The high-DPI/slow-motion case needs to be addressed before approval.

PR #23397, opened 2026-06-04
fix(theming): honor owner theme in VisualState setter

### LINKED ISSUE #23396: VisualState.Setter {ThemeResource} resolves against the application theme instead of the element's RequestedTheme
### Current behavior 🐛

A control that lives in a subtree pinned to an explicit `RequestedTheme` that differs from the application theme (for example a `RequestedTheme="Light"` subtree while the application/OS theme is Dark) can paint the wrong theme's brush for a property set through a **`VisualState.Setter` whose value is a `{ThemeResource}`**.

When the visual state is applied while no theme walk is active — e.g. when a virtualizing/recycling control re-applies a recycled container's visual state, outside the element's load-time theme walk — `ResourceResolver.ApplyVisualStateSetter` resolves the `{ThemeResource}` against the *ambient active theme* (the application theme) rather than the element's inherited (Actual) theme. The result is the Dark brush painted on a Light-themed element (e.g. a dark hover/pointer-over background appearing on a control inside a Light subtree).

The equivalent direct `{ThemeResource}` binding path (`ResourceResolver.ApplyResource`) already pushes the owner element's inherited theme before resolving; the VisualState.Setter path did not, so the two paths disagree.

### Expected behavior 🎯

A `VisualState.Setter` whose value is a `{ThemeResource}` resolves the resource against the element's inherited theme (its `ActualTheme`), consistent with direct `{ThemeResource}` bindings — regardless of whether the setter is applied inside or outside a theme walk.

Place a templated control whose visual state sets a property to a `{ThemeResource}` (e.g. `<Setter Target="Root.Background" Value="{ThemeResource ...}" />`) inside a `RequestedTheme="Light"` subtree under a Dark application, then cause that visual state to be (re)applied outside a theme walk (e.g. through a virtualizing container that recycles the element after navigation). The setter paints the Dark value instead of the Light one.

A runtime regression test covering the scenario is added in this PR: `Given_CheckBox_ThemeResource_Regression.When_VisualStateSetter_PointerOver_Keeps_Light_After_Frame_Navigation_Under_Dark_App`.

### Workaround 🛠️

_No response_

### Renderer 🎨

- [x] Skia
- [ ] Native

### Affected platforms 📱💻🖥️

Desktop (Windows), WebAssembly

6.7.0-dev (reproduced on Uno.WinUI 6.7.0-dev.263)

### IDE version 🧑‍💻

_No response_

_No response_

ajpinedam:
**GitHub Issue:** closes #23396

## PR Type:
Bugfix

## What changed? 🚀

A property set through a `VisualState.Setter` whose value is a `{ThemeResource}` could paint the wrong theme's brush for an element living in a subtree pinned to a `RequestedTheme` that differs from the application theme (e.g. a `RequestedTheme="Light"` subtree under a Dark application). When the visual state is applied outside a theme walk — for instance when a virtualizing/recycling control re-applies a recycled container's visual state after navigation — `ResourceResolver.ApplyVisualStateSetter` resolved the resource against the ambient active (application) theme instead of the element's inherited theme, so the element rendered the Dark brush while its subtree was Light.

The fix pushes the owner element's inherited theme around the dictionary lookup in `ApplyVisualStateSetter` (via `ThemeResourceReference.PushOwnerThemeIfDifferent`), mirroring the push the direct `{ThemeResource}` binding path (`ApplyResource`) already performs, so the setter selects the owner's Light/Dark sub-dictionary rather than falling back to `Themes.Active`.

A runtime regression test is added in `Given_CheckBox_ThemeResource_Regression` (`When_VisualStateSetter_PointerOver_Keeps_Light_After_Frame_Navigation_Under_Dark_App`) exercising a flat-button-style `VisualState.Setter` PointerOver `{ThemeResource}` in a Light-pinned `Frame` under a Dark application across navigation. The fix was additionally validated end-to-end in a downstream consumer app whose items-grid icon buttons rendered a dark hover square under a Light-pinned subtree; with the fix the hover background resolves the correct Light value (confirmed via instrumented theme-resolution logging: pre-fix the setter resolved with `active=Dark`/`ownerTheme=Light`, post-fix it pushes the owner theme and resolves the Light value).

## PR Checklist ✅

- [x] 🧪 Added Runtime tests (for bug fixes / features, if applicable)

### DIFF
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/CheckBoxNavReproOtherPage.xaml
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/CheckBoxNavReproOtherPage.xaml
@@ -1,0 +1,6 @@
+﻿<Page x:Class="Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls.CheckBoxNavReproOtherPage"
+      xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+      xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">
+	<!-- The "other tab": navigating here forces the checkbox page to fully unload before we navigate back. -->
+	<TextBlock Text="Other" />
+</Page>
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/CheckBoxNavReproOtherPage.xaml.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/CheckBoxNavReproOtherPage.xaml.cs
@@ -1,0 +1,11 @@
+﻿using Microsoft.UI.Xaml.Controls;
+
+namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls;
+
+public sealed partial class CheckBoxNavReproOtherPage : Page
+{
+	public CheckBoxNavReproOtherPage()
+	{
+		this.InitializeComponent();
+	}
+}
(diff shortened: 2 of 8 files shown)


MartinZikmund (approved):

agneszitte:
@Mergify backport release/stable/6.6

Issue #4563, opened 2020-11-19
Paths in a canvas not rendering correctly


mikernet:
## Current behavior

The rendering of the path objects is not done correctly in WASM. It looks like this:

![image](https://user-images.githubusercontent.com/1700111/99605613-09421d00-29d6-11eb-841c-a0b091786736.png)

## Expected behavior

It should render the same as UWP, i.e. like this:

![image](https://user-images.githubusercontent.com/1700111/99605656-1e1eb080-29d6-11eb-8e1f-15bb16e0c4d9.png)

## How to reproduce it (as minimally and precisely as possible)

Create a user with the following content:

```xml
<UserControl
    x:Class="iRatio.Client.Icon"
    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
    mc:Ignorable="d"
    d:DesignHeight="300"
    d:DesignWidth="400">

    <Grid>
        <Viewbox>
            <Canvas Name="layer1" Width="53" Height="53">
                <Canvas Name="g958" Opacity="1">
                    <Canvas.RenderTransform>
                        <MatrixTransform Matrix="1.1048446, 0, 0, 1.1959561, -67.035883, -96.098854"/>
                    </Canvas.RenderTransform>
                    <Canvas Name="g920" Opacity="1">
                        <Canvas.RenderTransform>
                            <TranslateTransform X="3.6170566" Y="26.121871"/>
                        </Canvas.RenderTransform>
                        <Path Name="path882" Fill="#FFFFFFFF" Data="m 89.844925 86.908932 -5.484915 -9.50015 -5.484913 -9.500149 10.969828 -10e-7 10.969825 0 -5.484911 9.50015 z">
                            <Path.RenderTransform>
                                <MatrixTransform Matrix="0.99974084, 0, 0, 1.0590767, 0.02096173, -5.1400261"/>
                            </Path.RenderTransform>
                        </Path>
                        <Path Name="path8822" Fill="#FFFFFFFF" Data="m 71.890129 -67.818339 -5.484914 -9.500151 -5.484915 -9.50015 10.969829 0 10.969829 -10e-7 -5.484915 9.500151 z">
                            <Path.RenderTransform>
[...]

PR #23401, opened 2026-06-05
test: Reproduce paths in Canvas with MatrixTransform rendering issue

MartinZikmund:
Relates to https://github.com/unoplatform/uno/issues/4563

**GitHub Issue:** closes #

## PR Checklist ✅

### DIFF
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Shapes/Given_Path.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Shapes/Given_Path.cs
@@ -470,64 +470,72 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Shapes
 		}
 #endif
 
-		// Repro tests for https://github.com/unoplatform/uno/issues/4563
+		// Repro test for https://github.com/unoplatform/uno/issues/4563
 		[TestMethod]
 		[GitHubWorkItem("https://github.com/unoplatform/uno/issues/4563")]
 		[PlatformCondition(ConditionMode.Exclude, RuntimeTestPlatforms.NativeWinUI)]
-		public async Task When_Paths_In_Canvas_With_MatrixTransform_Render_Without_Exception()
+		public async Task When_Path_In_Nested_Canvas_With_MatrixTransform_Loads_Without_Exception()
 		{
-			// Issue: Path objects inside nested Canvases with MatrixTransform and TranslateTransform
-			// don't render correctly on WASM (visual mismatch with UWP).
-			// This test verifies the layout doesn't throw and produces non-zero bounds.
+			// Issue: a Path inside nested Canvases with MatrixTransform and TranslateTransform
+			// doesn't render correctly on WASM (visual mismatch with UWP).
+			// This only guards the layout pass: loading must not throw and the Canvas must keep its
+			// explicit bounds. Pixel-level validation of the transformed geometry is not covered here.
 
 			Exception caught = null;
 			Microsoft.UI.Xaml.Controls.Canvas canvas = null;
 
 			try
 			{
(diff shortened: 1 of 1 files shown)


ajpinedam (approved):

MartinZikmund:
## Review comments addressed

Pushed commit **`3fde0aff38`** (`test: Drop comment duplicating GitHubWorkItem attribute`).

Most threads on this PR were already satisfied by the earlier cleanup commit `64a64faaa2` and are now stale — I re-read the current code on the branch rather than trusting the thread text. Breakdown below.

### Fixed (1)

| file:line | The point | What changed |
|---|---|---|
| `Given_Path.cs:473` (github-actions) | `// Repro test for .../issues/4563` duplicates the `[GitHubWorkItem]` attribute on the next line; per the code-style rule the default is no comment. | Correct — the attribute is the authoritative traceability link and the comment added nothing. Removed it. |

### No change needed (6)

| file:line | The point | Why no change |
|---|---|---|
| `Given_Path.cs` (copilot) | `caught?.Message` drops stack trace / inner exception; use the full exception. | Already done in `64a64faaa2`. Current line 529 reads `$"...but got: {caught}"` — the full exception, `ToString()`-formatted by interpolation. |
| `Given_Path.cs` (github-actions) | Same `caught?.Message` point, with a suggestion block. | Duplicate of the above; already satisfied by the current code. |
| `Given_Path.cs` (copilot) | Comment claims "non-zero bounds" but only `ActualWidth` is asserted, and the `$` string has no placeholders. | Already done. Current lines 532-533 assert `canvas.ActualWidth > 0 && canvas.ActualHeight > 0` and the message interpolates both values. |
| `Given_Path.cs` (github-actions) | Same two points (unnecessary `$`, assert both width and height). | Duplicate of the above; already satisfied. |
| `Given_Path.cs` (copilot) | Test name says "Paths"/"Render" but the XAML has one `Path` and the assertions only cover load/layout — rename or extend. | Already done. The test is now named `When_Path_In_Nested_Canvas_With_MatrixTransform_Loads_Without_Exception`, which states exactly what it asserts, and the body comment scopes it explicitly to the layout pass. |
| `Given_Path.cs:538` (github-actions) | `WindowHelper.WindowContent` is never reset, polluting later tests — needs a `finally`. | Already done. The scenario is wrapped in `try { … } finally { WindowHelper.WindowContent = null; }` (lines 535-538). |

### Needs @MartinZikmund input (1) — left open

| file:line | The point | Why it needs you |
|---|---|---|
| `Given_Path.cs` (copilot) | #4563 is a **visual** mismatch, so this test passes even when rendering is wrong; suggests `UITestHelper.ScreenShot(...)` + `ImageAssert` to assert known pixels once the transforms are applied. | The point is technically valid — the test currently guards only the layout pass, and the in-code comment says so honestly. But turning it into a pixel assertion means picking concrete sample coordinates for a geometry that is nested-`MatrixTransform`-ed inside a `Viewbox`, and those can only be chosen and trusted by *running* the test on each target. I can't do that here (no build/run in this pass), and guessing pixel coordinates would produce a test that is either flaky or vacuous — worse than the honest layout-only guard. Your call on whether this repro PR should grow real pixel validation. |

---

**Validation level:** the change above was reviewed **by inspection only** — I did not build, run, or test anything in this pass. Azure Pipelines CI will verify. The edit is a comment deletion, so the compile risk is nil, but the assertion that the other six threads are already satisfied rests on reading the current branch code (quoted line numbers above), not on a test run.

morning4coffe-dev (commented):
This is useful as a load/layout smoke test, but it does not establish transformed-Path rendering parity with WinUI: Canvas.ActualWidth/ActualHeight can be nonzero while the path pixels are missing, misplaced or clipped. The current description acknowledges that limitation. Please keep it explicit when discussing the linked rendering issue, and add a transformed-geometry/pixel assertion before treating this as proof that the visual regression is covered. I am not treating an Uno-only layout assertion as a native WinUI rendering comparison.

PR #23438, opened 2026-06-11
fix(alc): clear cross-ALC InheritanceContext associations on teardown (1/3)

### LINKED ISSUE #23437: Window-level events, composition-target registrations and ContentControl's template memo pin collectible ALCs
## Current behavior

Three further pieces of host-side static state pin a secondary collectible `AssemblyLoadContext` after its app unloads (each confirmed via `gcroot` on full memory dumps):

1. **Window-level events** — `Activated` / `SizeChanged` / `VisibilityChanged` / `Closed` subscriptions live on the window *implementation* object, which is not a visual-tree element. A secondary-ALC subscriber on a host window is never reached by the visual-tree event prune, and a closed ALC window is removed from `ApplicationHelper` early, so no sweep visits it either. Chains like `CoreDragDropManager map → window wrapper → DesktopWindow → Closed → per-ALC subscriber → LoaderAllocator` keep the ALC alive.
2. **`NativeDispatcher` composition-target render registrations** — never removed; a closed secondary window's `CompositionTarget` (and its content tree) is retained by the dispatcher's registration map.
3. **`ContentControl.HasDefaultTemplate`** — a static memoized `Func<Type, bool>` keyed by DefaultStyleKey types; keys from the unloaded app's controls pin the ALC through the memoization dictionary.

## Expected behavior

`Application.CleanupNonDefaultAlcCaches()` prunes window-level event subscriptions with collectible targets (for registered windows, and closed ALC windows prune in their close path), removes composition-target registrations whose ContentRoot is unregistered, and resets the default-template memo.

Depends on the event-prune infrastructure from #23431.

nickrandolph:
**GitHub Issue (If applicable):** closes #23437

> **Part 1 of 3** of the ALC window-teardown work, now split into a reviewable stack for easier review and future regression isolation:
> - **Part 1 (this PR)** — foundation + cross-ALC InheritanceContext associations
> - **Part 2 — #23495** — window events / orphaned composition targets (stacks on this)
> - **Part 3 — #23496** — Type-keyed cache pruning (stacks on Part 2)

## PR Type
🐞 Bugfix

## What changed? 🚀
A shared (host-lifetime) resource first consumed by a secondary-ALC element records that element as its InheritanceContext parent (`DependencyObjectStore._associatedParent`); nothing un-associates it when the element's ALC unloads, so the host resource pins the collectible ALC. `CleanupNonDefaultAlcCaches` now sweeps every dictionary reachable from the host application and the master theme set, clearing those associations.

Also adds the shared foundation used by Parts 2–3: `AlcStateHelper` (reads ALC unload state), `FeatureConfiguration.Alc.ThrowOnUnloadStateReadFailure`, and the `RunCleanupStep` best-effort step helper.

## PR Checklist ✅
- [x] 🧪 Runtime test added (`Given_CollectibleAlcAssociations`)

### DIFF
--- a/src/Uno.UI.UnitTests/Windows_UI_Xaml/Given_AlcStateHelper.cs
+++ b/src/Uno.UI.UnitTests/Windows_UI_Xaml/Given_AlcStateHelper.cs
@@ -1,0 +1,34 @@
+using System.Runtime.CompilerServices;
+using System.Runtime.Loader;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Uno.UI.Xaml.Core;
+
+namespace Uno.UI.Tests.Windows_UI_Xaml
+{
+	[TestClass]
+	public class Given_AlcStateHelper
+	{
+		[TestMethod]
+		public void When_Alive_Then_Not_UnloadInitiated_And_When_Unloaded_Then_UnloadInitiated()
+		{
+			if (!RuntimeFeature.IsDynamicCodeSupported)
+			{
+				Assert.Inconclusive("Collectible AssemblyLoadContext is not supported on this target.");
+			}
+
+			var alc = new AssemblyLoadContext("AlcStateHelperProbe", isCollectible: true);
+
+			// valueIfUnknown is set to the OPPOSITE of the expected result so a failed read (e.g. the
+			// reflected _state field no longer resolving) would surface as a test failure, not a false pass.
+			Assert.IsFalse(
+				AlcStateHelper.IsUnloadInitiated(alc, valueIfUnknown: true),
+				"A live ALC must report that its unload has not been initiated.");
+
+			alc.Unload();
+
+			Assert.IsTrue(
+				AlcStateHelper.IsUnloadInitiated(alc, valueIfUnknown: false),
+				"An ALC whose Unload() has been called must report that its unload has been initiated.");
+		}
+	}
+}
--- a/src/Uno.UI/FeatureConfiguration.cs
+++ b/src/Uno.UI/FeatureConfiguration.cs
@@ -17,6 +17,21 @@ namespace Uno.UI
 {
 	public static class FeatureConfiguration
 	{
+		/// <summary>
+		/// Configuration for collectible AssemblyLoadContext (secondary-app) teardown.
+		/// </summary>
+		public static class Alc
+		{
+			/// <summary>
+			/// When true, a failure to read an <see cref="System.Runtime.Loader.AssemblyLoadContext"/>'s
+			/// unload state (e.g. a future runtime renaming the private state field) throws instead of
+			/// silently falling back. The fallback keeps teardown leak-safe, but it would also let a
+			/// broken read silently degrade every ALC cleanup scenario — enable this in tests/CI so such
+			/// a regression fails loudly rather than going unnoticed. Defaults to false.
+			/// </summary>
+			public static bool ThrowOnUnloadStateReadFailure { get; set; }
+		}
+
 		public static class ApiInformation
 		{
 			/// <summary>
--- a/src/Uno.UI/UI/Xaml/AlcStateHelper.cs
+++ b/src/Uno.UI/UI/Xaml/AlcStateHelper.cs
@@ -1,0 +1,48 @@
+#nullable enable
+
+using System;
+using System.Reflection;
+using System.Runtime.Loader;
+
+namespace Uno.UI.Xaml.Core;
+
+/// <summary>
+/// Reads <see cref="AssemblyLoadContext"/> unload state. The runtime exposes no public unload-state
+/// API, so this reads the private <c>_state</c> field. Verified against .NET 9 and .NET 10: the field
+/// exists and <c>0 == Alive</c> (any other value means Unloading/Unloaded). If a future runtime renames
+/// the field this resolves to <c>null</c> and <see cref="IsUnloadInitiated"/> returns the caller-supplied
+/// fallback, so each call site keeps its own leak-safe default.
+/// </summary>
+internal static class AlcStateHelper
+{
+	private static readonly FieldInfo? _alcStateField =
+		typeof(AssemblyLoadContext).GetField("_state", BindingFlags.NonPublic | BindingFlags.Instance);
+
+	/// <summary>
+	/// Whether <paramref name="alc"/>'s unload has been initiated (private state != Alive). Returns
+	/// <paramref name="valueIfUnknown"/> when the state cannot be read (field absent on a future runtime,
+	/// or the read throws).
+	/// </summary>
+	internal static bool IsUnloadInitiated(AssemblyLoadContext alc, bool valueIfUnknown)
+	{
+		if (_alcStateField is null)
+		{
+			if (global::Uno.UI.FeatureConfiguration.Alc.ThrowOnUnloadStateReadFailure)
+			{
+				throw new InvalidOperationException(
+					"AssemblyLoadContext unload-state field '_state' was not found; the runtime layout may have changed.");
+			}
+
+			return valueIfUnknown;
+		}
+
+		try
+		{
+			return (int)_alcStateField.GetValue(alc)! != 0;
+		}
+		catch (Exception) when (!global::Uno.UI.FeatureConfiguration.Alc.ThrowOnUnloadStateReadFailure)
+		{
+			return valueIfUnknown;
+		}
+	}
+}
--- a/src/Uno.UI/UI/Xaml/Application.Alc.cs
+++ b/src/Uno.UI/UI/Xaml/Application.Alc.cs
@@ -371,6 +371,12 @@ partial class Application
 		// ResourceResolver — remove Func delegates whose Target is from a non-default ALC
 		ResourceResolver.ClearNonDefaultAlcRegistrations();
 
+		// Shared resources (theme brushes etc.) first consumed by a secondary-ALC element record
+		// it as their InheritanceContext parent (DependencyObjectStore._associatedParent); nothing
+		// clears that association on unload, so host-lifetime resources pin the collectible ALC.
+		// Sweep every dictionary reachable from the host application and the master theme set.
+		RunCleanupStep(nameof(ClearCollectibleResourceAssociations), ClearCollectibleResourceAssociations);
+
 		// ResourceDictionary — invalidate _keyNotFoundCache on the host Application.
 		// The cache accumulates ResourceKey entries with TypeKey referencing ALC types.
 		// These entries pin RuntimeType → LoaderAllocator → ALC.
@@ -395,6 +401,32 @@ partial class Application
 				typeof(Application).Log().Trace($"[ALC-SCAN] Error during deep scan: {ex.GetType().Name}: {ex.Message}");
 			}
 		}
+	}
+
+	// Runs a teardown cleanup step in isolation: teardown is best-effort, so a failing step must not
+	// abort the rest. A failure may indicate a real defect, so it is logged as a warning with the
+	// full exception.
+	private static void RunCleanupStep(string name, Action step)
+	{
+		try
+		{
+			step();
+		}
+		catch (Exception ex)
+		{
+			if (typeof(Application).Log().IsEnabled(LogLevel.Warning))
+			{
+				typeof(Application).Log().Warn($"[ALC-CLEANUP] {name} failed", ex);
+			}
+		}
+	}
+
+	private static void ClearCollectibleResourceAssociations()
+	{
+#if !__NETSTD_REFERENCE__
+		Uno.UI.GlobalStaticResources.MasterDictionary.ClearCollectibleAssociatedParents();
+#endif
+		_current?.Resources?.ClearCollectibleAssociatedParents();
 	}
 
 	private static void ClearNonDefaultAlcApplications()
--- a/src/Uno.UI/UI/Xaml/DependencyObjectStore.Binder.cs
+++ b/src/Uno.UI/UI/Xaml/DependencyObjectStore.Binder.cs
@@ -1,4 +1,4 @@
-﻿#nullable enable
+#nullable enable
 
 #if ENABLE_LEGACY_TEMPLATED_PARENT_SUPPORT
 // fallback option for legacy DepObj from external library generated before templated-parent rework.
@@ -711,6 +711,7 @@ namespace Microsoft.UI.Xaml
 			if (_associatedParent == null)
 			{
 				_associatedParent = parent;
+				RegisterCollectibleParentAssociation(parent);
 			}
 			else
 			{
--- a/src/Uno.UI/UI/Xaml/DependencyObjectStore.alc.cs
+++ b/src/Uno.UI/UI/Xaml/DependencyObjectStore.alc.cs
@@ -1,0 +1,174 @@
+#nullable enable
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.UI.Xaml
+{
+	public partial class DependencyObjectStore
+	{
+		/// <summary>
+		/// Drops <see cref="_associatedParent"/> — and any DataContext it propagated into this store —
+		/// when the parent is owned by an unloading collectible ALC, is an unloaded
+		/// <see cref="FrameworkElement"/>, or is orphaned from the content-root coordinator, so a
+		/// host-lifetime resource cannot pin a secondary app's ALC. A live host element re-associates
+		/// on its next value assignment.
+		/// </summary>
+		internal void ClearCollectibleAssociatedParent()
+		{
+			var associationCleared = false;
+
+			// Two flavors of secondary-app parents: objects whose TYPE lives in the collectible
+			// ALC, and instances of SHARED types (Grid, Border, …) owned by the unloaded app —
+			// the latter are indistinguishable by type, but after unload they are detached
+			// (unloaded, parentless), so a detached-element association is also dropped. A live
+			// host element re-associates naturally on its next value assignment.
+			// IsLoaded == false also matches host elements that are merely unloaded at sweep time
+			// (e.g. a collapsed pane); clearing their association only resets InheritanceContext
+			// DataContext propagation for the shared resource, which re-establishes on the next
+			// value assignment — an acceptable cost for guaranteeing the unloaded app's detached
+			// elements (shared types, indistinguishable by ALC) release their hold. An element
+			// whose ContentRoot has been unregistered from the coordinator (a closed secondary
+			// app's tree — popups included, which keep IsLoaded) is orphaned and also released.
+			// The collectible branch is gated on the parent's ALC unload having actually been
+			// initiated: a still-live session-lifetime add-in ALC (e.g. a designer host) is also
+			// collectible, but its associations must be preserved until it really unloads.
+			var collectibleAndUnloading = false;
+			if (_associatedParent is { } collectibleCandidate && collectibleCandidate.GetType().IsCollectible)
+			{
+				var parentAlc = global::System.Runtime.Loader.AssemblyLoadContext.GetLoadContext(collectibleCandidate.GetType().Assembly);
+				// Conservative when the unload state can't be read: do NOT clear, so a still-live add-in
+				// ALC's associations (and inherited DataContext) are never dropped on an unreadable state.
+				// FeatureConfiguration.Alc.ThrowOnUnloadStateReadFailure surfaces such a runtime change in dev.
+				collectibleAndUnloading = parentAlc is not null
+					&& global::Uno.UI.Xaml.Core.AlcStateHelper.IsUnloadInitiated(parentAlc, valueIfUnknown: false);
+			}
+
+			if (_associatedParent is { } parent
+				&& (collectibleAndUnloading
+					|| parent is FrameworkElement { IsLoaded: false }
+					|| (parent is FrameworkElement orphanCandidate && IsOrphanedFromContentRoots(orphanCandidate))))
+			{
+				_associatedParent = null;
+				associationCleared = true;
+			}
+
+			// The association may already have propagated the parent's DataContext into this
+			// store at Inheritance precedence; that propagated value (e.g. the secondary app's
+			// view model) survives the parent's removal and pins the value's ALC on its own.
+			if (associationCleared
+				|| GetValue(_dataContextProperty)?.GetType().IsCollectible == true)
+			{
+				ClearInheritedDataContext();
+			}
+		}
+
+		/// <summary>
+		/// Returns whether the element's visual tree no longer has a registered ContentRoot —
+		/// i.e. it belonged to a closed (secondary-app) tree whose root was unregistered from
+		/// the <see cref="Uno.UI.Xaml.Core.ContentRootCoordinator"/> on Window close.
+		/// </summary>
+		private static bool IsOrphanedFromContentRoots(FrameworkElement element)
+		{
+			try
+			{
+				var contentRoot = element.XamlRoot?.VisualTree?.ContentRoot;
+				if (contentRoot is null)
+				{
+					return true;
+				}
+
+				return !Uno.UI.Xaml.Core.CoreServices.Instance.ContentRootCoordinator.ContentRoots.Contains(contentRoot);
+			}
+			catch (Exception)
+			{
+				// Fail leak-safe: if orphan state can't be determined, treat the element as orphaned
+				// so its association is cleared rather than left to pin a potentially-unloaded ALC.
+				return true;
+			}
+		}
+
+		// Stores that recorded a collectible-ALC object as their associated parent, grouped by
+		// that parent's ALC. Entries are swept (associated parent cleared) when the ALC unloads,
+		// so a host-lifetime shared resource can never outlive-pin a secondary app's ALC.
+		// CWT-keyed by the ALC so the registry itself never extends the ALC's lifetime.
+		private static readonly global::System.Runtime.CompilerServices.ConditionalWeakTable<global::System.Runtime.Loader.AssemblyLoadContext, List<WeakReference<DependencyObjectStore>>> _collectibleParentAssociations = new();
+
+		private void RegisterCollectibleParentAssociation(object parent)
+		{
+			// Collectible-ALC association tracking is only meaningful when secondary ALCs exist.
+			// Gate on the substitutable Application.HasSecondaryApps so that, for the overwhelmingly
+			// common single-ALC app, this whole path stays off and is linkable away.
+			if (!Application.HasSecondaryApps)
+			{
+				return;
+			}
+
+			// Fast path: Type.IsCollectible is a cached runtime flag; non-collectible parents
+			// (the overwhelmingly common case) exit here without touching the registry.
+			if (!parent.GetType().IsCollectible)
+			{
+				return;
+			}
+
+			var alc = global::System.Runtime.Loader.AssemblyLoadContext.GetLoadContext(parent.GetType().Assembly);
+			if (alc is null || !alc.IsCollectible)
+			{
+				return;
+			}
+
+			var list = _collectibleParentAssociations.GetValue(alc, static a =>
+			{
+				a.Unloading += static unloading =>
+				{
+					// ClearCollectibleAssociatedParent reads/writes DependencyProperty values, which is
+					// only valid on the UI thread. Unloading fires on whichever thread called
+					// AssemblyLoadContext.Unload(); marshal to the UI thread so the clear actually runs
+					// instead of being swallowed as an off-thread DP-access failure.
+					static void Sweep(global::System.Runtime.Loader.AssemblyLoadContext unloading)
+					{
+						if (_collectibleParentAssociations.TryGetValue(unloading, out var stores))
+						{
+							lock (stores)
+							{
+								foreach (var weakStore in stores)
+								{
+									if (weakStore.TryGetTarget(out var store))
+									{
+										try
+										{
+											store.ClearCollectibleAssociatedParent();
+										}
+										catch (global::System.Exception)
+										{
+											// Defensive: one store throwing must not abort the sweep for the rest.
+										}
+									}
+								}
+
+								stores.Clear();
+							}
+
+							_collectibleParentAssociations.Remove(unloading);
+						}
+					}
+
+					if (global::Uno.UI.Dispatching.NativeDispatcher.Main.HasThreadAccess)
+					{
+						Sweep(unloading);
+					}
+					else
+					{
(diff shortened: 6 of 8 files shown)


src/Uno.UI.RuntimeTests/Tests/AssemblyLoadContext/Given_WindowCollectibleEventPrune.cs
```diff
@@ -0,0 +1,102 @@
...
+				"lives in a collectible assembly; otherwise the host window pins the unloaded ALC.");
+		}
+		finally
+		{
+			// In the failure case the subscription is still attached — detach to not leak into
+			// later tests. (In the success case the invocation entry is already gone and this
+			// is a no-op.)
+			if (handler is not null)
+			{
+				window!.Closed -= handler;
```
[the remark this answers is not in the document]
nickrandolph: Good catch - the staging helper now returns only a WeakReference (the subscription is the sole strong reference), and the failure-cleanup path re-creates an equivalent delegate from weakTarget.Target to detach (delegate equality is target+method, so -= matches). Also added a RuntimeFeature.IsDynamicCodeSupported guard while in there.

src/Uno.UI/UI/Xaml/ResourceResolver.cs
```diff
@@ -975,6 +975,32 @@ internal static void ClearNonDefaultAlcRegistrations()
...
+				foreach (var kvp in _registeredDictionariesByUriByAlc)
+				{
+					toRemove.Add(kvp.Key);
+				}
+
+				foreach (var alc in toRemove)
+				{
+					_registeredDictionariesByUriByAlc.Remove(alc);
+				}
+			}
```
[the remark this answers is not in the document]
nickrandolph: Fixed (via the base-branch merge in 342e567): RemoveNonDefaultAlcScopedRegistrations now prunes only entries whose key ALC has begun unloading, so a still-live secondary ALC keeps its registrations (matching the multi-ALC lookups this file supports).

src/Uno.UI.Dispatching/Native/NativeDispatcher.cs
```diff
@@ -35,6 +35,35 @@ internal sealed partial class NativeDispatcher
...
+
+				if (toRemove is not null)
+				{
+					foreach (var key in toRemove)
+					{
+						_compositionTargets.Remove(key);
+					}
+				}
+			}
+		}
```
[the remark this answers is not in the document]
nickrandolph: Fixed (342e567) - the predicate is now evaluated outside the gate on a snapshot of the keys, with a brief re-lock for the removals.

src/Uno.UI/UI/Xaml/Application.Alc.cs
```diff
@@ -332,6 +401,255 @@ internal static void CleanupNonDefaultAlcCaches()
...
+
+		try
+		{
+			return _alcStateField is not null && (int)_alcStateField.GetValue(alc) != 0;
+		}
+		catch (Exception)
+		{
+			// Fail leak-safe: an unreadable unload state must not leave the pin in place.
+			return true;
+		}
```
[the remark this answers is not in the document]
nickrandolph: Fixed in df72763ad6. `IsFromUnloadInitiatedAlc` no longer treats every default-ALC collectible type as prunable:

- **Constructed generics** (`HostType<TAddIn>`): I now re-evaluate against the generic type *definition* (`type.GetGenericTypeDefinition()`) and recompute *its* load context. If the definition lives in the default/null ALC and its assembly is non-dynamic, it is a host-owned definition that is collectible only because of a generic argument — returns `false` (not prunable), so live add-in subscriptions on `HostType<TAddIn>` survive. Otherwise the definition's own load context is judged like any other type.
- **Default/unknown ALC, non-generic**: now prunable only when `type.Assembly.IsDynamic` (genuine dynamic/RunAndCollect builder, which legitimately maps to a collectible context). A static host assembly that is collectible here would be a false positive and is left alone.
- **Unreadable `_state`**: the read is consolidated into the new `AlcStateHelper.IsUnloadInitiated(alc, valueIfUnknown)`. The destructive event-prune path passes `valueIfUnknown: true` (leak-safe for that path: it only reaches the helper after the type already resolved to a *non-default, non-dynamic* collectible ALC, i.e. a real secondary ALC), while the ResourceResolver sweep passes `valueIfUnknown: false` so an unreadable state never wipes a possibly-live secondary ALC's registrations.

src/Uno.UI/UI/Xaml/ResourceResolver.cs
```diff
@@ -975,6 +975,39 @@ internal static void ClearNonDefaultAlcRegistrations()
...
+						toRemove.Add(kvp.Key);
+					}
+				}
+
+				foreach (var alc in toRemove)
+				{
+					_registeredDictionariesByUriByAlc.Remove(alc);
+				}
+			}
 		}
```
jeromelaban, line 981: A later pattern that appeared when building ALC features is to move the ALC code into a `.alc.cs` file so it's properly isolated. Let's do that here.

src/Uno.UI/UI/Xaml/ResourceDictionary.cs
```diff
@@ -485,6 +485,45 @@ private void Set(in ResourceKey resourceKey, object value, bool throwIfPresent)
...
+
+			_themeDictionaries?.ClearCollectibleAssociatedParents();
+
+			// The MATERIALIZED active theme dictionary may be reachable only through this cache:
+			// when the theme entry in _themeDictionaries is still a lazy stub, the stub is skipped
+			// above and the materialized dictionary it produced — holding the live brushes whose
+			// stores record cross-ALC associations — would never be visited.
+			_activeThemeDictionary?.ClearCollectibleAssociatedParents();
+		}
+
```
jeromelaban: Move the ALC code into a `.alc.cs` file so it's properly isolated.

src/Uno.UI/UI/Xaml/DependencyProperty.Dictionary.TypeNullableDictionary.cs
```diff
@@ -45,6 +56,33 @@ internal void Add(Type key, bool isNullable)
...
+					{
+						keysToRemove.Add(key);
+					}
+				}
+
+				foreach (var key in keysToRemove)
+				{
+					_entries.Remove(key);
+				}
+			}
```
jeromelaban: Move the ALC code into a `.alc.cs` file so it's properly isolated.

src/Uno.UI/UI/Xaml/DependencyProperty.cs
```diff
@@ -61,6 +61,10 @@ internal static void ClearCachesForNonDefaultAlc()
 			_registry.RemoveNonDefaultAlcEntries();
 			_getInheritedPropertiesForType.RemoveNonDefaultAlcEntries();
 			_getPropertyCache.RemoveNonDefaultAlcEntries();
+			_isTypeNullableDictionary.RemoveNonDefaultAlcEntries();
+
+			// The pooled lookup key retains the last-queried Type past the cache prunes.
+			_searchPropertyCacheEntry.Update(typeof(object), "");
 		}
```
jeromelaban, line 64: Move the ALC code into a `.alc.cs` file so it's properly isolated.

src/Uno.UI/UI/Xaml/DependencyObjectStore.Binder.cs
```diff
@@ -685,6 +686,161 @@ private void UnassociateParent(object? parent)
 			}
 		}
 
+		/// <summary>
```
jeromelaban, line 740: Move the ALC code into a `.alc.cs` file so it's properly isolated.

src/Uno.UI/UI/Xaml/DependencyObjectStore.Binder.cs
```diff
@@ -662,6 +662,7 @@ private void AssociateParent(object? parent)
 			if (_associatedParent == null)
 			{
 				_associatedParent = parent;
+				RegisterCollectibleParentAssociation(parent);
```
jeromelaban, line 714: Let's make this entirely conditional to the presence of secondary ALCs. It must be using an explicitly substitutable property for proper linkability.

src/Uno.UI/UI/Xaml/Application.Alc.cs
```diff
@@ -306,6 +307,74 @@ internal static void CleanupNonDefaultAlcCaches()
...
+#endif
+			_current?.Resources?.ClearCollectibleAssociatedParents();
+		}
+		catch (Exception ex)
+		{
+			if (typeof(Application).Log().IsEnabled(LogLevel.Debug))
+			{
+				typeof(Application).Log().Debug($"[ALC-CLEANUP] ClearCollectibleAssociatedParents error: {ex.GetType().Name}: {ex.Message}");
+			}
+		}
```
jeromelaban: Let's move the try/catches inside the called methods being added, it's getting hard to read through.

src/Uno.UI/UI/Xaml/Application.Alc.cs
```diff
@@ -306,6 +307,74 @@ internal static void CleanupNonDefaultAlcCaches()
...
+
+		// NativeDispatcher render registrations are never removed; prune entries whose
+		// CompositionTarget's ContentRoot is no longer registered (a closed secondary-app
+		// surface). Runs AFTER the tree prune above so orphaned roots are already unregistered.
+		try
+		{
+			var rootCoordinator = Uno.UI.Xaml.Core.CoreServices.Instance.ContentRootCoordinator;
+			Uno.UI.Dispatching.NativeDispatcher.Main.RemoveCompositionTargets(target =>
+				target is Media.CompositionTarget compositionTarget
+				&& !rootCoordinator.ContentRoots.Contains(compositionTarget.ContentRoot));
```
jeromelaban: This entire block should be in a separate method, for readability.

src/Uno.UI/UI/Xaml/Application.Alc.cs
```diff
@@ -332,6 +401,274 @@ internal static void CleanupNonDefaultAlcCaches()
 		}
 	}
 
+	// Per-type cache of delegate-typed instance fields, to keep the teardown tree walk cheap.
+	// Concurrent: the cleanup can be triggered from multiple teardown paths at once
+	// (Window.CloseAlcWindows, Application.RemoveAlcApplication, host sweeps).
+	private static readonly global::System.Collections.Concurrent.ConcurrentDictionary<Type, global::System.Reflection.FieldInfo[]> _delegateFieldsCache = new();
```
jeromelaban: Move to the top of this file.

jeromelaban: I'm not seeing any pruning happening for this collection, particularly for the key.

src/Uno.UI/UI/Xaml/Application.Alc.cs
```diff
@@ -332,6 +401,274 @@ internal static void CleanupNonDefaultAlcCaches()
...
+			}
+			catch (Exception)
+			{
+				// One window refusing reflection must not abort pruning of the remaining windows.
+			}
+		}
+	}
+
+	[ThreadStatic]
+	private static bool _treeContainsUnloadingAlcElement;
```
jeromelaban: Unclear why this needs to be thread static, let's explain why.

jeromelaban: Move to the top of the file.

src/Uno.UI/UI/Xaml/Application.Alc.cs
```diff
@@ -332,6 +401,274 @@ internal static void CleanupNonDefaultAlcCaches()
...
+		}
+
+		try
+		{
+			PruneCollectibleDelegateFields(element);
+		}
+		catch (Exception)
+		{
+			// One element refusing reflection must not abort the sweep of the rest of the tree.
+		}
```
jeromelaban: We should log warnings for exceptions in debug builds. Those might indicate proper errors.

src/Uno.UI/UI/Xaml/AlcStateHelper.cs
```diff
@@ -0,0 +1,42 @@
...
+		}
+
+		try
+		{
+			return (int)_alcStateField.GetValue(alc)! != 0;
+		}
+		catch (Exception)
+		{
+			return valueIfUnknown;
+		}
```
jeromelaban: We should be able to setup for a forced exception when this is failing through a feature flag. Hiding the error will break all the scenarios silently otherwise.

src/Uno.UI/UI/Xaml/Application.Alc.cs
```diff
@@ -332,6 +401,274 @@ internal static void CleanupNonDefaultAlcCaches()
...
+	[UnconditionalSuppressMessage("Trimming", "IL2075", Justification = "ALC cleanup reflection over live instances")]
+	internal static void PruneCollectibleDelegateFields(object instance, AssemblyLoadContext dyingAlc = null)
+	{
+		var type = instance.GetType();
+		if (type.IsCollectible)
+		{
+			// Elements that themselves live in the collectible ALC die with it — and their
+			// subtree is theirs to keep; pruning is only needed on host-lifetime elements.
+			return;
+		}
```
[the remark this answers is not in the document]
nickrandolph: Fixed in 962dcb72c8 — the early-return now uses IsFromUnloadInitiatedAlc(type) (the same discriminator the per-invocation prune uses), so a host-lifetime constructed generic like HostType<TAddIn> is no longer skipped.

src/Uno.UI/UI/Xaml/DependencyObjectStore.Binder.cs
```diff
@@ -685,6 +686,161 @@ private void UnassociateParent(object? parent)
...
 
+		/// <summary>
+		/// Clears <see cref="_associatedParent"/> when it references an object from a collectible
+		/// AssemblyLoadContext. A shared (e.g. theme-dictionary) resource first consumed by a
+		/// secondary-ALC element records that element as its associated parent; nothing
+		/// unassociates it when the element's ALC unloads (UnassociateParent only runs on value
+		/// changes), so the host-lifetime resource pins the collectible ALC. Invoked from
+		/// <see cref="Microsoft.UI.Xaml.ResourceDictionary.ClearCollectibleAssociatedParents"/>
+		/// during ALC teardown and from the per-ALC Unloading sweep below.
+		/// </summary>
```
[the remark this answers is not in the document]
nickrandolph: Addressed in 962dcb72c8 — added an accurate summary covering all three clear paths (unloading collectible ALC, unloaded FrameworkElement, orphaned content root) plus the inherited-DataContext clearing.

src/Uno.UI/UI/Xaml/ResourceResolver.cs
nickrandolph: Moved to ResourceResolver.alc.cs in 523899be80 (class made partial).

src/Uno.UI/UI/Xaml/ResourceDictionary.cs
nickrandolph: Moved to ResourceDictionary.alc.cs in 523899be80.

src/Uno.UI/UI/Xaml/DependencyProperty.Dictionary.TypeNullableDictionary.cs
nickrandolph: Moved to DependencyProperty.Dictionary.TypeNullableDictionary.alc.cs in 523899be80 (nested class made partial).

src/Uno.UI/UI/Xaml/DependencyProperty.cs
nickrandolph: This file only adds statements inside an existing teardown method (no standalone ALC member to extract), so there is nothing to relocate here. The standalone ALC members in the sibling files were moved to .alc.cs partials in 523899be80.

src/Uno.UI/UI/Xaml/DependencyObjectStore.Binder.cs
nickrandolph: Moved to DependencyObjectStore.alc.cs in 523899be80.

src/Uno.UI/UI/Xaml/DependencyObjectStore.Binder.cs
nickrandolph: Done in 92af0a0be3 — RegisterCollectibleParentAssociation now early-returns unless Application.HasSecondaryApps, the existing substitutable gate used across the ALC resource paths, so the single-ALC common case is linkable away.

src/Uno.UI/UI/Xaml/Application.Alc.cs
nickrandolph: Done in c9eefc9c93 — the repeated inline try/catch blocks are replaced by a RunCleanupStep(name, action) helper, so the orchestration reads as a clean named sequence and the error handling lives in one place.

src/Uno.UI/UI/Xaml/Application.Alc.cs
nickrandolph: Extracted into PruneOrphanedCompositionTargets in c9eefc9c93.

src/Uno.UI/UI/Xaml/Application.Alc.cs
nickrandolph: Moved _delegateFieldsCache to the top of the file in c9eefc9c93.

nickrandolph: Good catch — fixed in c9eefc9c93. PruneCollectibleDelegateFieldsCache now drops collectible Type keys (incl. host generics over a collectible argument) during teardown; entries re-cache on demand.

src/Uno.UI/UI/Xaml/Application.Alc.cs
nickrandolph: Documented in c9eefc9c93 — it is per-walk transient state and teardown sweeps can run concurrently on multiple threads, so it must stay thread-local.

nickrandolph: Moved to the top of the file in c9eefc9c93.

src/Uno.UI/UI/Xaml/Application.Alc.cs
nickrandolph: Done in c9eefc9c93 — RunCleanupStep now logs failures as warnings (debug builds) instead of debug.

src/Uno.UI/UI/Xaml/AlcStateHelper.cs
nickrandolph: Done in 92af0a0be3 — added FeatureConfiguration.Alc.ThrowOnUnloadStateReadFailure; when set, IsUnloadInitiated throws instead of silently falling back, so a runtime-layout regression fails loudly in tests/CI.

src/Uno.UI.UnitTests/ContentControlTests/Given_ContentControl_DefaultTemplateMemo.cs
```diff
@@ -0,0 +1,59 @@
...
+			// The fresh memo still answers lookups.
+			after(typeof(ContentControl));
+		}
+
+		private static Type EmitCollectibleControlType()
+		{
+			var typeBuilder = AssemblyBuilder
+				.DefineDynamicAssembly(new AssemblyName("CollectibleDefaultTemplateProbe"), AssemblyBuilderAccess.RunAndCollect)
+				.DefineDynamicModule("main")
+				.DefineType("CollectibleControl", TypeAttributes.Public, typeof(ContentControl));
```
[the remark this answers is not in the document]
nickrandolph: Fixed in 221435dccf — guarded behind RuntimeFeature.IsDynamicCodeSupported with Assert.Inconclusive, matching the runtime-test sibling.

src/Uno.UI/UI/Xaml/Window/Window.alc.cs
```diff
@@ -298,6 +298,17 @@ private bool CloseAlcWindow()
...
+		// Capture the dying ALC BEFORE the content reference is cleared: at window-close time the
+		// secondary app's Unload() has not been initiated yet, so the delegate prune below cannot
+		// discriminate by unload state and needs the ALC identity explicitly.
+		AssemblyLoadContext? dyingAlc = null;
+		if (_secondaryAlcContent is { } secondaryContent
+			&& AssemblyLoadContext.GetLoadContext(secondaryContent.GetType().Assembly) is { IsCollectible: true } contentAlc
+			&& contentAlc != AssemblyLoadContext.Default)
+		{
+			dyingAlc = contentAlc;
+		}
```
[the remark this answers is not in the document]
nickrandolph: Fixed in 221435dccf — dyingAlc now comes from the authoritative OwnerAssemblyLoadContext (correct even when the secondary app rooted a shared framework type), falling back to the content type's ALC only when no owner was captured.

src/Uno.UI/UI/Xaml/Application.Alc.cs
```diff
@@ -397,6 +466,275 @@ internal static void CleanupNonDefaultAlcCaches()
...
+		catch (Exception)
+		{
+			// A subtree that cannot be enumerated mid-teardown must not abort the rest of the sweep.
+		}
+	}
+
+	[UnconditionalSuppressMessage("Trimming", "IL2070", Justification = "ALC cleanup reflection over live instances")]
+	[UnconditionalSuppressMessage("Trimming", "IL2075", Justification = "ALC cleanup reflection over live instances")]
+	internal static void PruneCollectibleDelegateFields(object instance, AssemblyLoadContext dyingAlc = null)
```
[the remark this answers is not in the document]
nickrandolph: Application.Alc.cs is under #nullable disable, so the non-nullable 'AssemblyLoadContext dyingAlc = null' default is correct and produces no warning; adding '?' would itself raise CS8632 (nullable annotation in a disabled context) here. Left as-is intentionally.

src/Uno.UI/UI/Xaml/DependencyObjectStore.Binder.cs
```diff
@@ -893,6 +894,160 @@ internal void RepropagateMentoredChildrenDataContext()
...
+			{
+				var parentAlc = global::System.Runtime.Loader.AssemblyLoadContext.GetLoadContext(collectibleCandidate.GetType().Assembly);
+				collectibleAndUnloading = parentAlc is not null
+					&& global::Uno.UI.Xaml.Core.AlcStateHelper.IsUnloadInitiated(parentAlc, valueIfUnknown: true);
+			}
+
+			if (_associatedParent is { } parent
+				&& (collectibleAndUnloading
+					|| parent is FrameworkElement { IsLoaded: false }
+					|| (parent is FrameworkElement orphanCandidate && IsOrphanedFromContentRoots(orphanCandidate))))
```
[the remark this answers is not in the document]
nickrandolph: This is an intentional, documented trade-off: ClearCollectibleAssociatedParent runs only during ALC teardown (CleanupNonDefaultAlcCaches), and clearing an unloaded element's association merely resets InheritanceContext DataContext propagation, which re-establishes on the next value assignment — the cost of guaranteeing that the unloaded app's detached shared-type elements (indistinguishable by ALC) release their hold. Style/setter-binding DataContext propagation is covered by runtime tests on the related branch. Flagging for your call on whether to tighten this to orphaned-only.

src/Uno.UI/UI/Xaml/ResourceResolver.alc.cs
```diff
@@ -0,0 +1,41 @@
+using System;
+using System.Collections.Generic;
+using System.Runtime.Loader;
```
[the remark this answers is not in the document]
nickrandolph: Fixed in e5b52810df — removed the unused 'using System;' and 'using System.Runtime.Loader;' from ResourceResolver.alc.cs (only List<> and the fully-qualified AssemblyLoadContext are used).

src/Uno.UI/UI/Xaml/Application.Alc.cs
```diff
@@ -397,6 +403,31 @@ internal static void CleanupNonDefaultAlcCaches()
...
+			step();
+		}
+		catch (Exception ex)
+		{
+			if (typeof(Application).Log().IsEnabled(LogLevel.Warning))
+			{
+				typeof(Application).Log().Warn($"[ALC-CLEANUP] {name} failed: {ex.GetType().Name}: {ex.Message}");
+			}
+		}
+	}
```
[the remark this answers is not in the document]
nickrandolph: Fixed in 791740f523 — corrected the comment (it logs whenever Warning logging is enabled, not debug-only) and now passes the exception to Log().Warn(message, ex) so the stack trace is captured.

src/Uno.UI/UI/Xaml/DependencyObjectStore.alc.cs
```diff
@@ -0,0 +1,171 @@
...
+			}
+
+			// The association may already have propagated the parent's DataContext into this
+			// store at Inheritance precedence; that propagated value (e.g. the secondary app's
+			// view model) survives the parent's removal and pins the value's ALC on its own.
+			if (hadCollectibleAssociation
+				|| GetValue(_dataContextProperty)?.GetType().IsCollectible == true)
+			{
+				ClearInheritedDataContext();
+			}
```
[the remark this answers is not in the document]
nickrandolph: Done in 791740f523 — renamed to associationCleared.

src/Uno.UI/UI/Xaml/AlcStateHelper.cs
```diff
@@ -0,0 +1,48 @@
...
+/// <summary>
+/// Reads <see cref="AssemblyLoadContext"/> unload state. The runtime exposes no public unload-state
+/// API, so this reads the private <c>_state</c> field. Verified against .NET 9 and .NET 10: the field
+/// exists and <c>0 == Alive</c> (any other value means Unloading/Unloaded). If a future runtime renames
+/// the field this resolves to <c>null</c> and <see cref="IsUnloadInitiated"/> returns the caller-supplied
+/// fallback, so each call site keeps its own leak-safe default.
+/// </summary>
+internal static class AlcStateHelper
+{
+	private static readonly FieldInfo? _alcStateField =
```
ebariche, line 18: We could use UnsafeAccessors here.

src/Uno.UI/FeatureConfiguration.cs
```diff
@@ -17,6 +17,21 @@ namespace Uno.UI
...
+		public static class Alc
+		{
+			/// <summary>
+			/// When true, a failure to read an <see cref="System.Runtime.Loader.AssemblyLoadContext"/>'s
+			/// unload state (e.g. a future runtime renaming the private state field) throws instead of
+			/// silently falling back. The fallback keeps teardown leak-safe, but it would also let a
+			/// broken read silently degrade every ALC cleanup scenario — enable this in tests/CI so such
+			/// a regression fails loudly rather than going unnoticed. Defaults to false.
+			/// </summary>
+			public static bool ThrowOnUnloadStateReadFailure { get; set; }
```
ebariche, line 32: If this is not exercised on non-ALC apps, it should always throw.

ebariche:
Looks ok to me as long as all tests pass.

ebariche:
cc @MartinZikmund @Xiaoy312

ebariche (approved):

src/Uno.UI/UI/Xaml/ResourceDictionary.alc.cs
```diff
@@ -0,0 +1,46 @@
...
+		/// </summary>
+		internal void ClearCollectibleAssociatedParents()
+		{
+			foreach (var value in _values.Values)
+			{
+				if (value is IDependencyObjectStoreProvider { Store: { } store })
+				{
+					store.ClearCollectibleAssociatedParent();
+				}
+				else if (value is ResourceDictionary nested)
```
ebariche: Should this come first ?

nickrandolph: Good catch — yes. Fixed in ee05bb4107: a nested ResourceDictionary is itself an IDependencyObjectStoreProvider, so it was matching the provider branch first and never being recursed into (its values' associations were missed). Reordered so the ResourceDictionary check comes first and recurses.

src/Uno.UI/FeatureConfiguration.cs
nickrandolph: We're keeping the opt-in flag here rather than always-throwing. It was added per an earlier review request for a feature-flag-gated forced exception, and the intent is: the production default stays leak-safe (fall back instead of throwing out of a teardown path if a future runtime changes the field), while FeatureConfiguration.Alc.ThrowOnUnloadStateReadFailure lets tests/CI opt into failing loudly so the incompatibility is caught in dev. Happy to revisit if you'd prefer always-throw.

src/Uno.UI/UI/Xaml/AlcStateHelper.cs
nickrandolph: Tried this, but UnsafeAccessor can't bind to this field: _state is the private InternalState enum, and [UnsafeAccessor(Field)] requires the accessor's ref-return type to match the field's exact type — which isn't nameable/expressible outside CoreLib (a 'ref int' accessor throws MissingFieldException at runtime; verified with a test). Kept the reflection read, but added a Given_AlcStateHelper unit test (ee05bb4107) that creates and unloads a collectible ALC to validate the read and guard against a future runtime field change. Open to a working UnsafeAccessor form if you know one.

src/Uno.UI.UnitTests/Windows_UI_Xaml/Given_AlcStateHelper.cs
```diff
@@ -0,0 +1,34 @@
...
+		{
+			if (!RuntimeFeature.IsDynamicCodeSupported)
+			{
+				Assert.Inconclusive("Collectible AssemblyLoadContext is not supported on this target.");
+			}
+
+			var alc = new AssemblyLoadContext("AlcStateHelperProbe", isCollectible: true);
+
+			// valueIfUnknown is set to the OPPOSITE of the expected result so a failed read (the
+			// UnsafeAccessor not binding) would surface as a test failure rather than a false pass.
```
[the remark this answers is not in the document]
nickrandolph: Fixed in 8914f32b92 — the comment no longer mentions UnsafeAccessor; it now refers to the reflected _state field failing to resolve, matching the helper's actual implementation.

MartinZikmund (approved):

MartinZikmund (approved):

PR #23447, opened 2026-06-12
fix(skia): suspend recycle-pooled element visuals

### LINKED ISSUE #23446: ItemsRepeater: animated content (ProgressRing/AnimatedVisualPlayer) in recycled elements keeps rendering — render loop never idles (high CPU on Skia/WASM)
## Summary

On the **Skia** render path (Desktop and WebAssembly), an `ItemsRepeater` keeps a **recycled element parented in the visual tree until it is reused**. If that element hosts continuously-animating content — e.g. an indeterminate `ProgressRing` (which is an `AnimatedVisualPlayer` + Lottie) — the animation keeps running on the recycled-but-still-parented element. The Skia render loop therefore **never goes idle**: the whole canvas repaints every frame at ~100% of a CPU core, even though nothing animated is visible on screen.

This happens whenever a list item with active animated content is removed/replaced and **no later item reuses its element** (for example, the replacement item resolves to a different `DataTemplate`).

## Environment

- Uno Platform 6.7 (also present on current `master`).
- .NET 10.
- Reproduces on the Skia render path: Desktop (X11 / macOS / Windows Skia) and **WebAssembly (Skia)**.

## Root cause

1. **Recycled elements stay parented.** `RecyclePool.PutElementCore` adds the element to the pool **without detaching it from its parent**. It is only removed from the parent later, in `RecyclePool.TryGetElementCore`, *when the element is reused* for another item. → If it is never reused, it stays a child of the `ItemsRepeater` indefinitely.
   - `src/Uno.UI/UI/Xaml/Controls/Repeater/RecyclePool.cs` — `PutElementCore` (no detach) vs `TryGetElementCore` (detaches on reuse).
2. **DataContext is not cleared on recycle (Uno-specific divergence).** On Uno, it is set on *prepare* (`ViewManager` → `elementAsFE.DataContext = elementDataContext`) and only overwritten on reuse. A recycled element keeps its **stale** DataContext, so a data-bound `ProgressRing.IsActive` stays `true` and **no `DataContextChanged` fires**. Current WinUI behaves differently: `ViewManager::ClearElementToElementFactory` nulls the DataContext on recycle when the repeater set it ([`MustClearDataContext`](https://github.com/microsoft/microsoft-ui-xaml/blob/d4aeb088aeeb17e023e83e00bc8e95ee20733059/controls/dev/Repeater/ViewManager.cpp#L144-L167)); Uno's Repeater port predates that change. This divergence is **not** what makes the bug Skia-specific, though: measured on real WinUI, the nulled DataContext leaves the `{Binding}`-driven `IsActive` at its last value (`true`) — the pooled ring keeps animating there too.
3. **The animation self-schedules while parented + active.** For the indeterminate `ProgressRing` the Skia source (`src/AddIns/Uno.UI.Lottie/LottieVisualSource.Skottie.cs`) re-invalidates from its own `Render()` every frame while playing (`if (_stopwatch.IsRunning) _skCanvasElement?.Invalidate();`), and the looped animation restarts the stopwatch forever. Because the recycled element is still in the visual tree and the ring is still active, it keeps invalidating the canvas → the render loop runs at full frame rate continuously.
4. `ElementClearing` does fire on the granular clear path, but the framework does nothing to stop/pause animations in the cleared element, and (per 1–2) the element is not unloaded — and on Uno its DataContext is not cleared either.

Net effect: a removed item's animated element is left parented and animating forever — high idle CPU, no visible animation.

## Minimal reproduction

A minimal UI with **two `DataTemplate`s** (so the spinner element is recycled but *not* reused). On Skia desktop or WASM:

**MainPage.xaml**
```xml
<Page x:Class="RepeaterSpinnerRepro.MainPage"
      xmlns:muxc="using:Microsoft.UI.Xaml.Controls"
      xmlns:local="using:RepeaterSpinnerRepro">

    <Page.Resources>
        <DataTemplate x:Key="SpinnerTemplate">
            <!-- Active indeterminate spinner. -->
            <muxc:ProgressRing IsActive="True" Width="24" Height="24" Margin="8" />
        </DataTemplate>

        <DataTemplate x:Key="TextTemplate">
            <TextBlock Text="{Binding Title}" Margin="8" />
        </DataTemplate>

        <local:ReproTemplateSelector x:Key="Selector"
                                     SpinnerTemplate="{StaticResource SpinnerTemplate}"
                                     TextTemplate="{StaticResource TextTemplate}" />
    </Page.Resources>

    <StackPanel Padding="24" Spacing="12">
        <Button Content="Remove the spinner item" Click="OnRemoveSpinner" />
        <ScrollViewer Height="200">
            <muxc:ItemsRepeater x:Name="Repeater"
                                ItemTemplate="{StaticResource Selector}" />
        </ScrollViewer>
    </StackPanel>
</Page>
```

**MainPage.xaml.cs**
```csharp
using System.Collections.ObjectModel;
using Microsoft.UI.Xaml;
using Microsoft.UI.Xaml.Controls;

namespace RepeaterSpinnerRepro;

public sealed record Item(string Title, bool IsSpinner);

{
    private readonly ObservableCollection<Item> _items = new()
    {
        new Item("a plain row", IsSpinner: false), // TextTemplate (a different template)
        new Item("the spinner",  IsSpinner: true),  // SpinnerTemplate (active ProgressRing)
    };

    public MainPage()
    {
        InitializeComponent();
        Repeater.ItemsSource = _items;
    }

    private void OnRemoveSpinner(object sender, RoutedEventArgs e)
    {
        // Remove the spinner item. Its element is recycled, but nothing reuses it
        // (the only remaining item uses TextTemplate), so it stays parented + animating.
        var spinner = _items.FirstOrDefault(i => i.IsSpinner);
        if (spinner is not null) _items.Remove(spinner);
    }
}

public sealed class ReproTemplateSelector : DataTemplateSelector
{
    public DataTemplate? SpinnerTemplate { get; set; }
    public DataTemplate? TextTemplate { get; set; }

    protected override DataTemplate SelectTemplateCore(object item)
        => (item is Item { IsSpinner: true } ? SpinnerTemplate : TextTemplate)!;
}
```

**Steps**
1. Run on Skia desktop or WebAssembly. Both rows render; the spinner animates.
2. Click **“Remove the spinner item.”** The spinner disappears from view (only the plain row remains).
3. Observe CPU: it stays pegged at ~100% of one core **indefinitely**, with no visible spinner.
   - WASM: DevTools → Rendering → **Paint flashing** shows the *whole canvas* repainting every frame; the Performance monitor shows continuous CPU.
   - Desktop: a system CPU monitor (or `dotnet-counters monitor`) shows sustained high CPU; the render loop never idles.

To confirm the mechanism programmatically: capture the realized spinner element via `Repeater.TryGetElement(index)` before removal, then after removal observe that it is **still parented** (`VisualTreeHelper.GetParent(...) != null`), its **`DataContext` is unchanged** on Uno (no `DataContextChanged`; on real WinUI the DataContext is cleared on recycle — see root cause 2), and the nested `ProgressRing.IsActive` is **still `true`** (measured on both Uno/Skia and WinUI).

> Contrast: if the replacement item uses the **same** template (so the element is *reused*), the DataContext updates on reuse and the spinner deactivates — no leak. The bug only manifests when a recycled element is **not** reused.

## What is (and is not) Uno's to fix

Of the behaviors described above, the first is **WinUI-parity by design and should NOT change**; the second is a real divergence but orthogonal to this bug:

- **Recycled elements staying parented** (`RecyclePool.PutElementCore` not detaching) matches WinUI's MUX implementation; same-owner elements are deliberately kept attached "so that we don't incur the enter/leave cost during recycling". Detaching on recycle would regress recycle performance and diverge from WinUI.
- **Stale DataContext on pooled elements** turns out to be Uno-specific, not parity: current WinUI clears the DataContext on recycle when it set it ([`MustClearDataContext`](https://github.com/microsoft/microsoft-ui-xaml/blob/d4aeb088aeeb17e023e83e00bc8e95ee20733059/controls/dev/Repeater/ViewManager.cpp#L144-L167)); Uno's port predates that change. Porting the clear is a separate parity work item (tracked in #23624) — but measurably **not a fix for this bug**: on real WinUI the ring's bound `IsActive` keeps its last value (`true`) after the clear, so the pooled spinner keeps animating there as well. Either way, **data bindings are not a viable shutdown mechanism for a recycled element**; the documented WinUI contract remains that the *app* stops animations in `ItemsRepeater.ElementClearing`, and that contract holds for Uno too.

The Uno-owned defect is the **render-loop cost model** for animated visuals that are not effectively rendered:

- On WinUI, forgetting the `ElementClearing` hygiene is near-free — composition animations run on the compositor with dirty-region rendering, so an invisible animated visual does not pin the UI thread.
- On the Skia path, the same omission **permanently pins a CPU core with full-canvas repaints from content that contributes zero visible pixels** (and on WebAssembly that is effectively fatal). A render loop that can be pinned forever by a non-rendered animation is a platform defect regardless of app hygiene.

## Expected

1. **Render layer (the fix):** a recycle-pooled element that stays parented should not keep the render loop awake. Its visual should be excluded from composition while it sits in the pool, and restored when it is reused — so continuously-animating content it hosts cannot keep invalidating the canvas once its item has been removed.
2. **Ideally, partial invalidation:** one small animating visual should dirty its own region, not force a whole-canvas repaint every frame.
3. **Docs:** the `ElementClearing` stop-animations pattern should be documented as a workaround on Skia targets for `ItemsRepeater` content hosting continuous animations on versions without the fix, since the failure mode there is catastrophic rather than benign.

## Actual

The recycled element remains parented with its `ProgressRing` active; the Skia render loop runs at full frame rate continuously (~100% of a core), invisibly.

## Notes / downstream workaround

Consumers can subscribe to `ItemsRepeater.ElementClearing` and stop animations in the cleared element (deactivate any `ProgressRing`, and/or collapse the element so it leaves the render pass), restoring on `ElementPrepared`. Two practical traps for anyone applying this workaround:

- A plain `IsActive = false` write on the cleared element can be overridden while the stale DataContext keeps a visual state / attached-property driver asserting "active" — collapsing the element is the robust part of the workaround.
- The restore on `ElementPrepared` must **re-drive the animation state from the new DataContext**, not just restore visibility: value-change-driven mechanisms (bindings, attached-property walks, `StateTrigger`s) do not re-fire when the new item produces the same values as the stale one (true→true), which otherwise leaves a visible row with a permanently stopped spinner.

DerekRaven:
## What

On the Skia composition path, an element recycled by `ItemsRepeater` (via `RecyclePool`) stays **parented and composed** until it is reused. If it hosts continuously-animating content — e.g. an indeterminate `ProgressRing`, whose Lottie re-invalidates the canvas from its own `Render()` — the render loop never idles again after the item is removed: ~100% of one core, full-canvas repaints, with nothing visible on screen.

Fixes #23446.

## How

"Elements in the pool are not composed", encapsulated in the pool itself:

- `RecyclePool.PutElementCore` suspends the element's visual; `TryGetElementCore` restores it on reuse — covering every pool consumer and every reuse path.
- The suspension is a new internal `UIElement.IsVisualSuspendedForRecyclePooling` (Skia-only) that drives `Visual.IsVisible` **without touching the element's public `Visibility`** — no app-observable state changes, no interference with bindings or visual states.
- The restore is eager rather than left to the reuse arrange (which can early-out and skip `ShowVisual`), and `ShowVisual` respects the flag so an arrange while pooled cannot resurrect the visual.
- Intentionally **not** done: detaching pooled elements (WinUI-parity behavior the pool relies on for cheap reuse) or clearing their DataContext. Note: current WinUI *does* clear the DataContext on recycle when the repeater set it ([`ViewManager::ClearElementToElementFactory` / `MustClearDataContext`](https://github.com/microsoft/microsoft-ui-xaml/blob/d4aeb088aeeb17e023e83e00bc8e95ee20733059/controls/dev/Repeater/ViewManager.cpp#L144-L167)) — Uno's Repeater port predates that change. That is a separate parity gap (tracked in #23624), and measurably not a fix for this bug: running the repro sample on the WinAppSDK head (real WinUI) reports `stillParented=True / dataContextUnchanged=False / ringIsActive=True` — the pooled ring keeps animating on WinUI even after the clear; WinUI just renders it for free.

## Validation

- **Repro**: `ItemsRepeater_RecycledSpinner` SamplesApp sample — two templates so the removed spinner element is recycled but never reused; includes a mechanism readout, a passive frames-per-second probe (the primary render-loop signal — full frame rate while pinned vs ~1–2 frames/s once idle), and an in-app CPU meter (meaningful on software-rasterized targets only; GPU-accelerated desktops read near-idle CPU even while the loop is pinned).
- **Measured on Skia X11** (external 1 Hz sampling of `/proc/<pid>/stat`):

  | Phase | Before | After |
  |---|---|---|
  | Spinner visible (animating) | ~1.4 cores | ~1.3 cores |
  | After spinner item removed | **~1.4 cores, forever** | **0.00 cores within ~1 s** |
  | Spinner item re-added (element reused) | — | ~1.3 cores (spinner animates again) |

  Control: the same app idling on a static `ItemsRepeater` sample measures 0.00 cores.
- **Runtime test**: `When_Item_Removed_Then_Pooled_Element_Visual_Suspended_Until_Reuse` — realized → composed; pooled → still parented, suspended, public `Visibility` untouched; reused → same element instance, composed again. Fails without the `RecyclePool` change, passes with it.
- **Existing suites**: Repeater + MUX repeater runtime tests pass (108/108).

## Docs

`Uno-UI-Performance.md` (Skia specifics) documents the new behavior, plus the `ElementClearing`/`ElementPrepared` workaround guidance for earlier versions — including the trap that re-activation on reuse cannot be left to bindings (no re-fire when the new item produces identical values).

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/Repeater/RecyclePool.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Repeater/RecyclePool.cs
@@ -50,6 +50,14 @@ namespace Microsoft.UI.Xaml.Controls
 			var winrtOwner = owner;
 			var winrtOwnerAsPanel = EnsureOwnerIsPanelOrNull(winrtOwner);
 
+#if __SKIA__
+			// Uno specific: a pooled element stays parented (and would stay composed) until it is
+			// reused. Suspend its visual so it neither renders at its stale position nor keeps the
+			// render loop awake through continuously-animating content (e.g. an indeterminate
+			// ProgressRing self-invalidating the canvas every frame). Restored in TryGetElementCore.
+			element.IsVisualSuspendedForRecyclePooling = true;
+#endif
+
 			ElementInfo elementInfo = new ElementInfo(element, winrtOwnerAsPanel);
 
 			if (m_elements.TryGetValue(winrtKey, out var infos))
@@ -111,6 +119,11 @@ namespace Microsoft.UI.Xaml.Controls
 							panel.Children.RemoveAt(childIndex);
 						}
 					}
+
+#if __SKIA__
+					// Uno specific: undo the composition suspension applied in PutElementCore.
+					elementInfo.Element.IsVisualSuspendedForRecyclePooling = false;
+#endif
 
 					return elementInfo.Element;
 				}
--- a/doc/articles/Uno-UI-Performance.md
+++ b/doc/articles/Uno-UI-Performance.md
@@ -164,6 +164,7 @@ You'll find below other known memory leak patterns on iOS Native:
   - It's possible to set `DebugSettings.EnableFrameRateCounter` in `App.OnLaunched` in order to view a top-left indicator. It indicates the current frames per second, as well as the time spent rendering a composition frame, in milliseconds.
   - If the indicator does not change, this means that the UI is not refreshing.
   - If it is, but nothing is changing visually, it could be that a XAML or Composition animation is still running, see the `ProgressRing` section in this document.
+- Elements recycled by an `ItemsRepeater` (or any `RecyclePool` consumer) stay parented in the visual tree until they are reused. Their visuals are automatically excluded from composition while pooled, so continuously-animating content they host (e.g. an indeterminate `ProgressRing`) cannot keep the render loop awake once its item is removed. When targeting Uno Platform versions without this behavior, stop animations in `ItemsRepeater.ElementClearing` and restore them on `ElementPrepared` — and note that re-activating cannot be left to bindings alone, as a reused element whose new item produces the same values does not re-fire them.
 - Avoid the use of SkiaSharp's `SKXamlCanvas`, instead use [`SKCanvasElement`](xref:Uno.Controls.SKCanvasElement) which is hardware accelerated.
 
 ## Advanced performance Tracing
--- a/src/SamplesApp/SamplesApp.Samples/Windows_UI_Xaml_Controls/Repeater/ItemsRepeater_RecycledSpinner.xaml
+++ b/src/SamplesApp/SamplesApp.Samples/Windows_UI_Xaml_Controls/Repeater/ItemsRepeater_RecycledSpinner.xaml
@@ -1,0 +1,103 @@
+<Page
+    x:Class="UITests.Windows_UI_Xaml_Controls.Repeater.ItemsRepeater_RecycledSpinner"
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:local="using:UITests.Windows_UI_Xaml_Controls.Repeater"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    xmlns:mux="using:Microsoft.UI.Xaml.Controls"
+    Background="{ThemeResource ApplicationPageBackgroundThemeBrush}"
+    mc:Ignorable="d">
+
+    <Page.Resources>
+        <!--  Template 1: row hosting an active indeterminate ProgressRing (Lottie/AnimatedVisualPlayer on Skia)  -->
+        <DataTemplate x:Key="SpinnerRowTemplate" x:DataType="local:RecycledSpinnerItem">
+            <StackPanel
+                Height="40"
+                Padding="16,0"
+                Orientation="Horizontal"
+                Spacing="8">
+                <mux:ProgressRing
+                    Width="24"
+                    Height="24"
+                    VerticalAlignment="Center"
+                    IsActive="{Binding IsSpinning}" />
+                <TextBlock VerticalAlignment="Center" Text="{Binding Title}" />
+            </StackPanel>
+        </DataTemplate>
+
+        <!--  Template 2: plain text row. Removing the spinner item leaves only rows of THIS template,
+              so the recycled spinner element is never reused (different template = different pool key).  -->
+        <DataTemplate x:Key="TextRowTemplate" x:DataType="local:RecycledSpinnerItem">
+            <TextBlock
+                Height="32"
+                Padding="16,0"
+                VerticalAlignment="Center"
+                Text="{Binding Title}" />
+        </DataTemplate>
+
+        <local:RecycledSpinnerTemplateSelector
+            x:Key="ItemTemplateSelector"
+            SpinnerTemplate="{StaticResource SpinnerRowTemplate}"
+            TextTemplate="{StaticResource TextRowTemplate}" />
+    </Page.Resources>
+
+    <Grid Padding="12" RowSpacing="8">
+        <Grid.RowDefinitions>
+            <RowDefinition Height="Auto" />
+            <RowDefinition Height="Auto" />
+            <RowDefinition Height="Auto" />
+            <RowDefinition Height="*" />
+        </Grid.RowDefinitions>
+
+        <TextBlock
+            Grid.Row="0"
+            MaxWidth="680"
+            HorizontalAlignment="Left"
+            Style="{ThemeResource BodyTextBlockStyle}"
+            TextWrapping="Wrap">
+            Repro: an active ProgressRing inside an ItemsRepeater item keeps the Skia render loop pinned after its item is removed.
+            Steps: (1) press "Measure CPU (2 s)" for a baseline while the spinner is visible (expected: high — the spinner is animating);
+            (2) press "Remove spinner item" — the spinner disappears, its element is recycled but never reused (remaining rows use a different template);
+            (3) press "Measure CPU (2 s)" again. Expected: near-idle. Actual: still ~100% of one core, indefinitely, with nothing animating on screen.
+            The mechanism line shows the recycled element is still parented, with its stale DataContext and an active ring.
+        </TextBlock>
+
+        <StackPanel
+            Grid.Row="1"
+            Orientation="Horizontal"
+            Spacing="6">
+            <Button
+                x:Name="RemoveSpinnerButton"
+                Click="OnRemoveSpinnerClick"
+                Content="Remove spinner item" />
+            <Button Click="OnResetClick" Content="Reset" />
+            <Button Click="OnMeasureCpuClick" Content="Measure CPU (2 s)" />
+        </StackPanel>
+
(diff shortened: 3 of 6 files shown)


ajpinedam (approved):

Xiaoy312 (approved):

src/Uno.UI/UI/Xaml/Controls/Repeater/RecyclePool.cs
```diff
@@ -112,6 +120,11 @@ protected virtual UIElement TryGetElementCore(string key, UIElement owner)
 						}
 					}
 
+#if __SKIA__
+					// Uno specific: undo the composition suspension applied in PutElementCore.
+					elementInfo.Element.IsVisualSuspendedForRecyclePooling = false;
+#endif
```
MartinZikmund, line 126: I would prefer avoiding any new Uno specific code in ItemsRepeater, as Skia should ideally be aligned with WinUI sources - does the animation keep rendering active even in WinUI? I would wager that should not be the case 🤔

DerekRaven: @MartinZikmund  After some testing it appears that WinUI has the same behavior keeping ``IsActive`` true when the animation is suspended, however, the difference is that WinUI doesn't render anything when the suspended element has a size of zero, so it doesn't incur any performance issue. [Documented on the issue itself ](https://github.com/unoplatform/uno/issues/23446#issuecomment-4867985710).

I found one other inconsistency where WinUI is handling clearing the DataContext on recycle, a behavior that isn't present in Uno, but that alone won't fix the original issue. [Issue created](https://github.com/unoplatform/uno/issues/23624).

Do you still think we can move forward with the fix in ItemsRepeater or do you think it should be moved elsewhere?

morning4coffe-dev (approved):
The suspension is paired at pool put/take, leaves public Visibility intact, survives an arrange while pooled, and restores eagerly when reuse does not trigger another arrange. The Skia render path skips invisible visuals, addressing the render-driven spinner loop without detaching pool entries.

Retain the explicit remaining WinUI difference: [ViewManager clears DataContext when the repeater assigned it](https://github.com/microsoft/microsoft-ui-xaml/blob/d0c6705ec46dbddb853cdfdb2eb3b42849cf304e/controls/dev/Repeater/ViewManager.cpp#L153-L167). This patch intentionally does not port that ownership cleanup, so visual suspension is not full recycle-lifecycle parity. Source approval; conflicts must still be reconciled.

PR #23450, opened 2026-06-12
fix: recover layout when MeasureOverride throws

### LINKED ISSUE #23449: Element layout permanently frozen after its MeasureOverride throws
### Current behavior 🐛

When a control's `MeasureOverride` throws, the element is left permanently unable to be measured again. Its `DesiredSize` (and therefore `ActualWidth`/`ActualHeight`) freeze at the last successfully measured value and never update again — even after the condition that caused the exception is gone.

Once in this state:

- Subsequent `InvalidateMeasure()` calls have no effect.
- `UpdateLayout()` cannot recover the element.
- Because the measure-dirtiness no longer propagates to ancestors, the freeze affects the whole subtree: e.g. a parent with an explicit `Width` can no longer be resized.

The frozen state persists for the lifetime of the element.

**Cause:** in the crossruntime layouter (`UIElement.Layout.crossruntime.cs`), `DoMeasure` sets the per-element `LayoutFlag.MeasuringSelf` flag before calling `MeasureCore`, and clears it afterwards — but not in a `finally`. When `MeasureCore` throws, `MeasuringSelf` is left set. `InvalidateMeasure()` early-returns whenever `MeasuringSelf` is set, so the element can never be marked measure-dirty again, and never re-measured.

### Expected behavior 🎯

A `MeasureOverride` that throws should not permanently disable layout for the element. Once the throwing condition is resolved and the element is re-invalidated, it should measure normally again — the same as an element that never threw.

An element whose `MeasureOverride` throws once, is then re-invalidated and measured again, keeps its pre-throw `DesiredSize` instead of updating to the new value.

Reproduced by the runtime tests added in the associated PR:

- `Given_Layouter_MeasureException.When_MeasureThrowsOnce_Then_Recovers_On_Reinvalidate`
- `Given_Layouter_MeasureException.When_ChildThrows_DuringResize_Then_Parent_Recovers`

Both fail before the fix (the element/parent stays at its pre-throw size) and pass after.

### Workaround 🛠️

None from application code — once `MeasuringSelf` is stuck, neither `InvalidateMeasure()` nor `UpdateLayout()` can recover the element.

### Renderer 🎨

- [x] Skia

### Affected platforms 📱💻🖥️

All platforms using the crossruntime layouter (Skia, WebAssembly, Reference).

Fix: clear `MeasuringSelf` in a `finally` in `DoMeasure` so a transient measure exception cannot permanently disable an element's layout.

DerekRaven:
**GitHub Issue:** closes #23449

## PR Type:

🐞 Bugfix

## What changed? 🚀

When a control's `MeasureOverride` throws, the element was left permanently unable to be measured again: its `DesiredSize` froze at the last successful value and never updated, `InvalidateMeasure()`/`UpdateLayout()` had no effect, and the freeze propagated to ancestors (a parent with an explicit size could no longer resize). The state persisted for the lifetime of the element.

Cause: `DoMeasure` (in `UIElement.Layout.crossruntime.cs`) set `LayoutFlag.MeasuringSelf` before calling `MeasureCore` and cleared it afterwards, but not in a `finally`. On a throw the flag stayed set, and `InvalidateMeasure()` early-returns whenever `MeasuringSelf` is set — so the element could never be marked measure-dirty again.

This PR clears `MeasuringSelf` in a `finally`, so a transient measure exception no longer permanently disables layout: once the throwing condition is resolved and the element is re-invalidated, it measures normally again.

Added runtime tests (`Given_Layouter_MeasureException`):
- `When_MeasureThrowsOnce_Then_Recovers_On_Reinvalidate` — a single element recovers after a throwing measure.
- `When_ChildThrows_DuringResize_Then_Parent_Recovers` — a parent with an explicit `Width` re-measures after a child throws during a resize.

Both fail before the change and pass after (validated on Skia Desktop).

## PR Checklist ✅

- [x] 🧪 Added Runtime tests, UI tests, or a manual test sample (for bug fixes / features, if applicable)

### DIFF
--- a/src/Uno.UI/UI/Xaml/UIElement.Layout.crossruntime.cs
+++ b/src/Uno.UI/UI/Xaml/UIElement.Layout.crossruntime.cs
@@ -249,8 +249,19 @@ namespace Microsoft.UI.Xaml
 						}
 
 						SetLayoutFlags(LayoutFlag.MeasuringSelf);
-						MeasureCore(availableSize);
-						ClearLayoutFlags(LayoutFlag.MeasuringSelf);
+						try
+						{
+							MeasureCore(availableSize);
+						}
+						finally
+						{
+							// Clear MeasuringSelf even if MeasureCore throws. Otherwise the flag stays
+							// set and InvalidateMeasure() permanently no-ops for this element (see the
+							// guard in InvalidateMeasure), which freezes its DesiredSize until the
+							// process restarts — a transient measure exception becomes a permanent
+							// layout lock for the whole subtree.
+							ClearLayoutFlags(LayoutFlag.MeasuringSelf);
+						}
 						InvalidateArrange();
 					}
 #if DEBUG
--- a/src/Uno.UI/UI/Xaml/UIElement.Layout.crossruntime.cs
+++ b/src/Uno.UI/UI/Xaml/UIElement.Layout.crossruntime.cs
@@ -255,11 +255,7 @@ namespace Microsoft.UI.Xaml
 						}
 						finally
 						{
-							// Clear MeasuringSelf even if MeasureCore throws. Otherwise the flag stays
-							// set and InvalidateMeasure() permanently no-ops for this element (see the
-							// guard in InvalidateMeasure), which freezes its DesiredSize until the
-							// process restarts — a transient measure exception becomes a permanent
-							// layout lock for the whole subtree.
+							// Clear even on throw, otherwise InvalidateMeasure() permanently no-ops and the subtree's layout freezes.
 							ClearLayoutFlags(LayoutFlag.MeasuringSelf);
 						}
 						InvalidateArrange();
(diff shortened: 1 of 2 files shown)


MartinZikmund (approved):

src/Uno.UI/UI/Xaml/UIElement.Layout.crossruntime.cs
```diff
@@ -249,8 +249,19 @@ private void DoMeasure(Size availableSize)
...
 						SetLayoutFlags(LayoutFlag.MeasuringSelf);
-						MeasureCore(availableSize);
-						ClearLayoutFlags(LayoutFlag.MeasuringSelf);
+						try
+						{
+							MeasureCore(availableSize);
+						}
+						finally
+						{
+							// Clear MeasuringSelf even if MeasureCore throws. Otherwise the flag stays
```
MartinZikmund: Unnecessarily verbose comment

MartinZikmund (approved):

ajpinedam (approved):

PR #23464, opened 2026-06-15
refactor!: Remove IgnoreINPCSameReferences binding flag

MartinZikmund:
**GitHub Issue:** #8339

## PR Type:

## What changed? 🚀

Part of the breaking-changes rollout (epic #8339), item **BC35**.

Removes the public `FeatureConfiguration.Binding.IgnoreINPCSameReferences` flag and the now-empty `FeatureConfiguration.Binding` nested class, then collapses the flag's three conditional branches in `BindingPath.BindingItem` to the always-taken default-`false` ("correct") path:

1. **`DataContext` setter** (`BindingPath.BindingItem.cs`): dropped the early-return short-circuit
   `if (IgnoreINPCSameReferences && !AreDifferent(DataContext, value)) return;`. With the flag at its default `false`, this condition was always `false`, so the early return never happened. The setter now always rents a weak reference and calls `SetWeakDataContext`.

2. **`OnPropertyChanged`**: removed the `&& (!IgnoreINPCSameReferences || previousValue != newValue)` conjunct (with the flag `false`, `!flag` was always `true`, so the conjunct was always satisfied). `RaiseValueChanged` now fires whenever `shouldRaiseValueChanged` is set. The stale explanatory comment that referenced the flag (including the "In Uno 6, we should remove the following line" note) was removed.

3. **`SubscribeToPropertyChanged`**: removed the `if (IgnoreINPCSameReferences) { valueHandler.PreviousValue = GetSourceValue(); }` block. With the flag `false`, this `GetSourceValue()` seeding never ran. The dispose path still reads `valueHandler.PreviousValue` (now `null`) and passes it with `shouldRaiseValueChanged: false`, so behaviour is unchanged.

The flag was never set to `true` anywhere in the repo (verified across the integration branch and `master`) and had no test references, so behaviour on the default is **neutral** — this is purely the removal of dead configurability plus its public surface.

### Validation

- **Compile**: `dotnet build src/Uno.UI/Uno.UI.Skia.csproj -p:UnoTargetFrameworkOverride=net10.0 -p:UnoFastDevBuild=true` — **Build succeeded, 0 errors** (`Uno.UI.dll` produced). The only warnings are pre-existing binding-redirect warnings in the unrelated `Uno.UI.Tasks` build tool.
- **Code-review**: confirmed via grep that the only consumers of the symbol were the declaration, the three usages collapsed here, and the spec checklist entry — no external callers, no tests reference it, and it is never assigned `true`. Each branch collapse matches the default-`false` semantics.
- **Runtime**: not run locally; CI on this PR exercises the full data-binding / INPC runtime test suite, which is the runtime backstop for this cross-platform binding-engine change.

## PR Checklist ✅

- [ ] 🧪 Added [Runtime tests, UI tests, or a manual test sample](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) (for bug fixes / features, if applicable) — N/A, behaviour is neutral on the default path; existing binding/INPC suites cover the kept path.
- [ ] 📚 Docs have been added/updated following the [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features) — N/A.

**Breaking change — impact & migration:** The public static property `FeatureConfiguration.Binding.IgnoreINPCSameReferences` (and the now-empty `FeatureConfiguration.Binding` class) is removed. **Impact:** code that read or set this flag will no longer compile. **Migration:** remove any references — the engine now permanently uses the default (`false`) behaviour, which is the correct INPC processing path and was already the behaviour for every app that never set the flag (i.e. all of them).

### DIFF
--- a/src/Uno.UI/DataBinding/BindingPath.BindingItem.cs
+++ b/src/Uno.UI/DataBinding/BindingPath.BindingItem.cs
@@ -170,7 +170,7 @@ namespace Uno.UI.DataBinding
 				}
 			}
 
-			private void OnPropertyChanged(object? previousValue, object? newValue, bool shouldRaiseValueChanged)
+			private void OnPropertyChanged(object? newValue, bool shouldRaiseValueChanged)
 			{
 				if (_isDataContextChanging && newValue == DependencyProperty.UnsetValue)
 				{
@@ -375,11 +375,9 @@ namespace Uno.UI.DataBinding
 						// weak with regards to the delegates that are provided.
 						disposables.Add(() =>
 						{
-							var previousValue = valueHandler.PreviousValue;
-
 							valueHandler = null;
 							handlerDisposable.Dispose();
-							OnPropertyChanged(previousValue, DependencyProperty.UnsetValue, shouldRaiseValueChanged: false);
+							OnPropertyChanged(DependencyProperty.UnsetValue, shouldRaiseValueChanged: false);
 						});
 					}
 				}
@@ -465,8 +463,6 @@ namespace Uno.UI.DataBinding
 					_self = WeakReferencePool.RentSelfWeakReference(this);
 				}
 
-				public object? PreviousValue { get; set; }
-
 				public ManagedWeakReference WeakReference
 					=> _self;
 
@@ -474,9 +470,7 @@ namespace Uno.UI.DataBinding
 				{
 					var newValue = _owner.GetSourceValue();
 
-					_owner.OnPropertyChanged(PreviousValue, newValue, shouldRaiseValueChanged: true);
-
-					PreviousValue = newValue;
+					_owner.OnPropertyChanged(newValue, shouldRaiseValueChanged: true);
 				}
 
 				public void NewValue(DependencyObject dependencyObject, DependencyPropertyChangedEventArgs args)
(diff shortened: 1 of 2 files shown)


MartinZikmund:
Addressed the open review thread on this PR. Pushed commit `24089a38b4`.

### Fixed (1)

|-----------|----------------|--------------|
| `src/Uno.UI/DataBinding/BindingPath.BindingItem.cs:188` | Nit (ultrareview): after the `OnPropertyChanged` collapse, `previousValue` is dead — a complete clean-up would drop the parameter, its call sites, and the `PreviousValue` plumbing on `PropertyChangedValueHandler`. | Did the full clean-up now rather than deferring: removed `previousValue` from `OnPropertyChanged`, updated both call sites, and removed the `PreviousValue` property plus its `= newValue` assignment. Verified `PreviousValue` is **not** part of the `IPropertyChangedValueHandler` interface (only the two `NewValue` overloads are), `OnPropertyChanged` is `private` and `BindingItem` is not `partial`, and no reference survives repo-wide — so this is a fully self-contained change with zero API impact (no BC entry needed). |

### No change needed (0)

---
Changes were reviewed by inspection only (no local build/test run per the rate-limit/disk constraints); Azure Pipelines CI will verify the compile.

PR #23495, opened 2026-06-18
fix(alc): prune window events and orphaned composition targets on teardown (2/3)

nickrandolph:
> **Depends on #23438 (Part 1).** This PR's branch is stacked on Part 1, so it contains Part 1's commit too — **review the last commit only**. Will be rebased onto master once #23438 merges.

**GitHub Issue (If applicable):** Part of #23437

> **Part 2 of 3** of the ALC window-teardown work, split out for reviewability and regression isolation. **Stacks on #23438 (Part 1)** — this PR is based on Part 1's branch, so review only its own commit. Will be retargeted to `master` once Part 1 merges.

## PR Type
🐞 Bugfix

## What changed? 🚀
Window-level event and render pruning during collectible-ALC teardown. Secondary-ALC code subscribes to events on host visual-tree elements and on the window implementation; those handlers are never removed when the secondary app unloads, and each pins the collectible ALC. This walks every live content root and prunes delegate-field invocation entries whose owner lives in an *unload-initiated* ALC (`IsFromUnloadInitiatedAlc`, which re-evaluates constructed generics against their definition so host types are never stripped), prunes the Type-keyed delegate cache, prunes orphaned `NativeDispatcher` composition targets, and clears window-level subscriptions at close time using the authoritative `OwnerAssemblyLoadContext`.

## PR Checklist ✅
- [x] 🧪 Runtime test added (`Given_WindowCollectibleEventPrune`)

### DIFF
--- a/src/Uno.UI.Dispatching/Native/NativeDispatcher.cs
+++ b/src/Uno.UI.Dispatching/Native/NativeDispatcher.cs
@@ -35,6 +35,57 @@ namespace Uno.UI.Dispatching
 
 		private readonly Dictionary<object, (Action? renderAction, int normalItemsToProcessBeforeNextRenderAction)> _compositionTargets = new();
 
+		/// <summary>
+		/// Removes composition-target render registrations matching the predicate. Nothing else
+		/// ever removes entries, so a closed secondary-app window's CompositionTarget otherwise
+		/// stays registered forever — pinning its visual tree (and its collectible
+		/// AssemblyLoadContext) through this process-lifetime dispatcher.
+		/// </summary>
+		internal void RemoveCompositionTargets(Predicate<object> shouldRemove)
+		{
+			ArgumentNullException.ThrowIfNull(shouldRemove);
+
+			// The predicate is caller-provided: evaluate it OUTSIDE the gate so arbitrary work
+			// (or a predicate that re-enters dispatcher APIs taking the gate) cannot deadlock or
+			// extend the lock's hold time.
+			object[] snapshot;
+			lock (_gate)
+			{
+				snapshot = new object[_compositionTargets.Keys.Count];
+				_compositionTargets.Keys.CopyTo(snapshot, 0);
+			}
+
+			List<object>? toRemove = null;
+			foreach (var key in snapshot)
+			{
+				if (shouldRemove(key))
+				{
+					(toRemove ??= new()).Add(key);
+				}
+			}
+
+			if (toRemove is not null)
+			{
+				lock (_gate)
+				{
+					foreach (var key in toRemove)
+					{
+						// A pending renderAction was counted in _globalCount by EnqueueRender and is
+						// normally decremented when TryGetRenderAction consumes it. Removing the entry
+						// here would skip that decrement, leaving _globalCount permanently elevated and
+						// breaking the 0→1 transition future scheduling relies on — so account for it.
+						if (_compositionTargets.TryGetValue(key, out var details)
+							&& details.renderAction is not null)
+						{
+							Interlocked.Decrement(ref _globalCount);
+						}
+
+						_compositionTargets.Remove(key);
+					}
+				}
+			}
+		}
+
 		private NativeDispatcherPriority _currentPriority;
 
 		private int _globalCount;
--- a/src/Uno.UI.UnitTests/Windows_UI_Xaml/Given_AlcStateHelper.cs
+++ b/src/Uno.UI.UnitTests/Windows_UI_Xaml/Given_AlcStateHelper.cs
@@ -1,0 +1,34 @@
+using System.Runtime.CompilerServices;
+using System.Runtime.Loader;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Uno.UI.Xaml.Core;
+
+namespace Uno.UI.Tests.Windows_UI_Xaml
+{
+	[TestClass]
+	public class Given_AlcStateHelper
+	{
+		[TestMethod]
+		public void When_Alive_Then_Not_UnloadInitiated_And_When_Unloaded_Then_UnloadInitiated()
+		{
+			if (!RuntimeFeature.IsDynamicCodeSupported)
+			{
+				Assert.Inconclusive("Collectible AssemblyLoadContext is not supported on this target.");
+			}
+
+			var alc = new AssemblyLoadContext("AlcStateHelperProbe", isCollectible: true);
+
+			// valueIfUnknown is set to the OPPOSITE of the expected result so a failed read (e.g. the
+			// reflected _state field no longer resolving) would surface as a test failure, not a false pass.
+			Assert.IsFalse(
+				AlcStateHelper.IsUnloadInitiated(alc, valueIfUnknown: true),
+				"A live ALC must report that its unload has not been initiated.");
+
+			alc.Unload();
+
+			Assert.IsTrue(
+				AlcStateHelper.IsUnloadInitiated(alc, valueIfUnknown: false),
+				"An ALC whose Unload() has been called must report that its unload has been initiated.");
+		}
+	}
+}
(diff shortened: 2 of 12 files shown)


src/Uno.UI.Dispatching/Native/NativeDispatcher.cs
```diff
@@ -35,6 +35,45 @@ internal sealed partial class NativeDispatcher
...
+		/// Removes composition-target render registrations matching the predicate. Nothing else
+		/// ever removes entries, so a closed secondary-app window's CompositionTarget otherwise
+		/// stays registered forever — pinning its visual tree (and its collectible
+		/// AssemblyLoadContext) through this process-lifetime dispatcher.
+		/// </summary>
+		internal void RemoveCompositionTargets(Predicate<object> shouldRemove)
+		{
+			// The predicate is caller-provided: evaluate it OUTSIDE the gate so arbitrary work
+			// (or a predicate that re-enters dispatcher APIs taking the gate) cannot deadlock or
+			// extend the lock's hold time.
```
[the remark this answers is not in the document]
nickrandolph: Done in 87cf111309 — added ArgumentNullException.ThrowIfNull(shouldRemove) at the top of RemoveCompositionTargets.

src/Uno.UI/UI/Xaml/Application.Alc.cs
```diff
@@ -428,6 +456,289 @@ private static void ClearCollectibleResourceAssociations()
...
+		var alc = AssemblyLoadContext.GetLoadContext(type.Assembly);
+		if (alc is null || alc == AssemblyLoadContext.Default)
+		{
+			// Default/unknown ALC is only "dying" when the assembly is a dynamic/RunAndCollect
+			// builder (which legitimately maps to a collectible context). A genuinely static
+			// host assembly that is collectible here would be a false positive, so guard on it.
+			return type.Assembly.IsDynamic;
+		}
+
+		return global::Uno.UI.Xaml.Core.AlcStateHelper.IsUnloadInitiated(alc, valueIfUnknown: true);
```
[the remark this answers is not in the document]
nickrandolph: Done in 87cf111309 — IsFromUnloadInitiatedAlc now uses valueIfUnknown:false, so an unreadable unload state no longer triggers destructive pruning of still-live collectible ALCs; FeatureConfiguration.Alc.ThrowOnUnloadStateReadFailure surfaces the incompatibility in dev as you suggested.

src/Uno.UI/UI/Xaml/Application.Alc.cs
```diff
@@ -397,6 +431,319 @@ internal static void CleanupNonDefaultAlcCaches()
...
+		var alc = AssemblyLoadContext.GetLoadContext(type.Assembly);
+		if (alc is null || alc == AssemblyLoadContext.Default)
+		{
+			// Default/unknown ALC is only "dying" when the assembly is a dynamic/RunAndCollect
+			// builder (which legitimately maps to a collectible context). A genuinely static
+			// host assembly that is collectible here would be a false positive, so guard on it.
+			return type.Assembly.IsDynamic;
+		}
+
+		// Conservative when the unload state can't be read: do NOT treat the ALC as dying, so this
```
ebariche, line 533: This is why throwing matters. We should fix that post PR.

src/Uno.UI/UI/Xaml/Application.Alc.cs
```diff
@@ -397,6 +431,319 @@ internal static void CleanupNonDefaultAlcCaches()
...
+	private static void PruneCollectibleAlcEventSubscriptions()
+	{
+		// Window-owned content roots are NEVER removed here, no matter what their trees contain:
+		// the host main window's tree legitimately hosts a previewed app's (collectible-typed)
+		// content, and unregistering the host root would take down the host window.
+		var windows = global::Uno.UI.ApplicationHelper.WindowsInternal.ToArray();
+		var windowRoots = new HashSet<object>();
+		foreach (var window in windows)
+		{
+			try
```
ebariche: Try/catch shouldn't be necessary here.

src/Uno.UI/UI/Xaml/Application.Alc.cs
```diff
@@ -397,6 +431,319 @@ internal static void CleanupNonDefaultAlcCaches()
...
+				}
+			}
+			catch (Exception)
+			{
+			}
+		}
+
+		var coordinator = Uno.UI.Xaml.Core.CoreServices.Instance.ContentRootCoordinator;
+		var roots = coordinator.ContentRoots;
+		for (var i = roots.Count - 1; i >= 0; i--)
```
ebariche, line 568: Can we explain why we're iterating in reverse here.

src/Uno.UI/UI/Xaml/DependencyObjectStore.alc.cs
```diff
@@ -0,0 +1,171 @@
...
+			// The collectible branch is gated on the parent's ALC unload having actually been
+			// initiated: a still-live session-lifetime add-in ALC (e.g. a designer host) is also
+			// collectible, but its associations must be preserved until it really unloads.
+			var collectibleAndUnloading = false;
+			if (_associatedParent is { } collectibleCandidate && collectibleCandidate.GetType().IsCollectible)
+			{
+				var parentAlc = global::System.Runtime.Loader.AssemblyLoadContext.GetLoadContext(collectibleCandidate.GetType().Assembly);
+				collectibleAndUnloading = parentAlc is not null
+					&& global::Uno.UI.Xaml.Core.AlcStateHelper.IsUnloadInitiated(parentAlc, valueIfUnknown: true);
+			}
```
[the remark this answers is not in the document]
nickrandolph: Fixed in 08e9085283 — ClearCollectibleAssociatedParent now uses valueIfUnknown:false, matching the delegate-prune path, so an unreadable unload state no longer clears associations for still-live collectible (add-in) ALCs. Surfaced in dev via FeatureConfiguration.Alc.ThrowOnUnloadStateReadFailure.

src/Uno.UI.UnitTests/Windows_UI_Xaml/Given_AlcStateHelper.cs
```diff
@@ -0,0 +1,34 @@
...
+		{
+			if (!RuntimeFeature.IsDynamicCodeSupported)
+			{
+				Assert.Inconclusive("Collectible AssemblyLoadContext is not supported on this target.");
+			}
+
+			var alc = new AssemblyLoadContext("AlcStateHelperProbe", isCollectible: true);
+
+			// valueIfUnknown is set to the OPPOSITE of the expected result so a failed read (the
+			// UnsafeAccessor not binding) would surface as a test failure rather than a false pass.
```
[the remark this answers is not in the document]
nickrandolph: Already addressed — the comment no longer references UnsafeAccessor (it refers to the reflected _state field). Resolving.

src/Uno.UI/UI/Xaml/Application.Alc.cs
nickrandolph: Agreed — leaving the throw-by-default change as post-PR follow-up per your note. The conservative fallback here keeps teardown leak-safe in the meantime, and the FeatureConfiguration.Alc.ThrowOnUnloadStateReadFailure flag lets it be exercised loudly in tests/CI.

src/Uno.UI/UI/Xaml/Application.Alc.cs
nickrandolph: Removed in 8b7094da82 — the WindowImplementation?.XamlRoot?.VisualTree?.ContentRoot access is fully null-conditional (the XamlRoot getters return null rather than throw), so the try/catch wasn't needed.

src/Uno.UI/UI/Xaml/Application.Alc.cs
nickrandolph: Added a comment in 8b7094da82 — we iterate in reverse because RemoveContentRoot mutates the collection during the loop, so walking from the end keeps the remaining indices valid.

MartinZikmund (approved):

PR #23511, opened 2026-06-20
fix(macos): Skip native window creation for secondary ALC windows

ajpinedam:
**GitHub Issue:** closes #

## PR Type:

🐞 Bugfix

## What changed? 🚀

In the hosted secondary-ALC scenario on **macOS** desktop, the secondary app's `new Window()` still went through `DesktopWindow.Initialize()` → `BaseWindowImplementation.InitializeNativeWindow()` → `NativeWindowFactory.CreateWindow()`, creating a real native macOS window and registering it in `MacOSWindowHost`. Because the secondary window collided with the host app's already-registered window, this raised a duplicate-key `ArgumentException`, leaving the inner app in a broken state (no DevServer bridge, no screenshots, blank UI). This was also the root cause of the "unknown core dump" that previously forced these ALC runtime tests to be disabled on macOS.

This PR makes `DesktopWindow.Initialize()` return early — skipping native window, chrome, and `XamlSource` creation — when `Window.ContentHostOverride` is set **and** the platform is macOS. Content redirection to the `AlcContentHost` is already handled by `Window.Content` at the framework level, so no native window is needed. The skip is intentionally scoped to macOS: Win32/X11 must continue through the normal `Initialize` path because their `DisplayInformation` extensions dereference `XamlRoot`/`RootElement`. The `Content` setter is also guarded to gracefully handle the intentionally-null `WindowChrome` in hosted mode.

With the collision fixed, macOS is re-enabled across the existing ALC runtime tests, plus three new tests covering the skip behavior directly:

- `When_ContentHostOverride_Set_Then_NativeWindowNotCreated`
- `When_AlcWindow_ContentSetToNull_Then_DoesNotThrow`
- `When_SecondaryAlcApp_OnMacOS_Then_LoadsWithoutNativeWindowConflict`

## PR Checklist ✅

### DIFF
--- a/src/Uno.UI.RuntimeTests/Tests/AssemblyLoadContext/Given_AlcContentHost.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/AssemblyLoadContext/Given_AlcContentHost.cs
@@ -1551,6 +1551,25 @@ public class Given_AlcContentHost
 	}
 
 	/// <summary>
+	/// A macOS hosted-ALC window has no native window (DesktopWindow.Initialize skips it), so the
+	/// Window ctor must skip DisplayInformation.GetOrCreateForWindowId for it. Otherwise the macOS
+	/// DisplayInformation extension subscribes to MacOSWindowNative.NativeWindowReady and never
+	/// unsubscribes (no native window is coming) — pinning the extension/DisplayInformation (and the
+	/// collectible ALC) and binding to the next unrelated native window that gets created.
+	/// </summary>
+	[TestMethod]
+	[PlatformCondition(ConditionMode.Include, RuntimeTestPlatforms.SkiaMacOS)]
+	public async Task When_MacOSAlcWindow_Then_NoDisplayInformationRegistered()
+	{
+		var (_, alcWindow) = await StartSecondaryAlcAppWithWindowAsync();
+
+		Assert.IsFalse(
+			Windows.Graphics.Display.DisplayInformation.IsRegisteredForWindowId(alcWindow.AppWindow.Id),
+			"A macOS hosted-ALC window must not register a DisplayInformation — doing so leaves " +
+			"MacOSDisplayInformationExtension subscribed to NativeWindowReady forever, pinning the ALC.");
+	}
+
+	/// <summary>
 	/// Boots the AlcApp test application into a secondary ALC, assuming the caller has already set up
 	/// <see cref="WindowHelper.ContentHostOverride"/> and the window content. Returns the registered
 	/// secondary <see cref="Application"/>. Use <see cref="StartSecondaryAlcAppAsync"/> instead when the
--- a/src/Uno.UI.RuntimeTests/Tests/AssemblyLoadContext/Given_AlcContentHost.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/AssemblyLoadContext/Given_AlcContentHost.cs
@@ -1478,7 +1478,7 @@ public class Given_AlcContentHost
 	[PlatformCondition(ConditionMode.Include, RuntimeTestPlatforms.SkiaMacOS)]
 	public async Task When_ContentHostOverride_Set_Then_NativeWindowNotCreated()
 	{
-		var (contentHost, alcWindow) = await StartSecondaryAlcAppWithWindowAsync();
+		var (_, alcWindow) = await StartSecondaryAlcAppWithWindowAsync();
 
 		// The ALC window should NOT have a native window wrapper — DesktopWindow.Initialize()
 		// must skip native window creation when ContentHostOverride is set on macOS.
--- a/src/Uno.UI.RuntimeTests/Tests/AssemblyLoadContext/Given_AlcContentHost.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/AssemblyLoadContext/Given_AlcContentHost.cs
@@ -80,7 +80,7 @@ public class Given_AlcContentHost
 	}
 
 	[TestMethod]
-	[PlatformCondition(ConditionMode.Include, RuntimeTestPlatforms.SkiaWin32 | RuntimeTestPlatforms.SkiaX11)] // Disabled on macOS for unknown core dump
+	[PlatformCondition(ConditionMode.Include, RuntimeTestPlatforms.SkiaWin32 | RuntimeTestPlatforms.SkiaX11 | RuntimeTestPlatforms.SkiaMacOS)]
 	public async Task When_AlcContentHost_Then_ResourcesInherited()
 	{
 		var contentHost = await StartSecondaryAlcAppAsync();
@@ -98,7 +98,7 @@ public class Given_AlcContentHost
 	}
 
 	[TestMethod]
-	[PlatformCondition(ConditionMode.Include, RuntimeTestPlatforms.SkiaWin32 | RuntimeTestPlatforms.SkiaX11)] // Disabled on macOS for unknown core dump
+	[PlatformCondition(ConditionMode.Include, RuntimeTestPlatforms.SkiaWin32 | RuntimeTestPlatforms.SkiaX11 | RuntimeTestPlatforms.SkiaMacOS)]
 	public async Task When_AlcContentHost_Then_MergedDictionariesInherited()
 	{
 		var contentHost = await StartSecondaryAlcAppAsync();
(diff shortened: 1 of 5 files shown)


src/Uno.UI/UI/Xaml/Window/Implementations/DesktopWindow.cs
```diff
@@ -21,6 +21,16 @@ public DesktopWindow(Window window) : base(window)
...
 	{
+		// macOS-only: the secondary ALC scenario collides with MacOSWindowHost's
+		// native window registration. Skipping native window/chrome/XamlSource here
+		// avoids the duplicate registration. Win32/X11 require XamlRoot/RootElement
+		// to be set up (DisplayInformation extensions dereference them), so they
+		// must continue through the normal Initialize path.
+		if (Window.ContentHostOverride is not null && OperatingSystem.IsMacOS())
+		{
+			return;
+		}
```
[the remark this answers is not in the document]
ajpinedam: Fixed in 87a9a45523 — the `Initialize()` guard is now scoped to actual ALC windows: `OperatingSystem.IsMacOS() && Window.ContentHostOverride is not null && Window.IsAlcWindow`.

src/Uno.UI/UI/Xaml/Window/Implementations/DesktopWindow.cs
```diff
@@ -41,6 +51,13 @@ public override UIElement? Content
 		{
 			if (_windowChrome is null)
 			{
+				// On macOS hosted ALC scenarios the chrome is intentionally not created;
+				// content is redirected at the Window.Content level via ContentHostOverride.
+				if (Window.ContentHostOverride is not null && OperatingSystem.IsMacOS())
+				{
+					return;
+				}
```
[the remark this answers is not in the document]
ajpinedam: Fixed in dc4ea0f9cd — the `Content` setter no-op guard is now scoped to ALC windows (`OperatingSystem.IsMacOS() && Window.ContentHostOverride is not null && Window.IsAlcWindow`), so non-ALC windows still fail fast on premature content assignment.

src/Uno.UI.RuntimeTests/Tests/AssemblyLoadContext/Given_AlcContentHost.cs
```diff
@@ -1241,6 +1241,90 @@ public async Task When_SecondaryAlcWindow_RootContentIsSharedOrNull_Then_OwnedBy
...
+	/// a native window that conflicted with the host's window. Win32/X11 are intentionally
+	/// excluded — they continue to set up XamlRoot/RootElement because their
+	/// DisplayInformation extensions dereference them.
+	/// </summary>
+	[TestMethod]
+	[PlatformCondition(ConditionMode.Include, RuntimeTestPlatforms.SkiaMacOS)]
+	public async Task When_ContentHostOverride_Set_Then_NativeWindowNotCreated()
+	{
+		var (contentHost, alcWindow) = await StartSecondaryAlcAppWithWindowAsync();
+
```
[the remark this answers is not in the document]
ajpinedam: Fixed in c492d65d60 — the unused tuple element is now a discard (`var (_, alcWindow) = ...`).

src/Uno.UI/UI/Xaml/Window/Implementations/DesktopWindow.cs
```diff
@@ -21,6 +21,11 @@ public DesktopWindow(Window window) : base(window)
 
 	public override void Initialize()
 	{
+		if (OperatingSystem.IsMacOS() && Window.ContentHostOverride is not null && Window.IsAlcWindow)
```
MartinZikmund: Why is this only for macOS? I assume we don't "throwaway" want native windows created even on other targets, even if it does not throw

ajpinedam: Let me investigate. This PR was intended only to move this other PR https://github.com/unoplatform/uno-private/pull/1850 into this repository.

ajpinedam: Indeed, the fix would only validate macOS but it will throw away for Windows Desktop. I just updated the validation and even though (as mentioned above) the fix was meant to be only for macOS it will also take into account Windows Desktop.

ajpinedam: Following up on my last reply here — the cross-platform version I pushed actually **broke Win32/X11 in CI**, so I've reverted to the macOS-only scope. The CI failure turned out to be the real answer to your question:

On **Win32/X11 the native window is not a throwaway** — creating it is what registers the `X11XamlRootHost` (and the Win32 host equivalent), and `DisplayInformation.GetOrCreateForWindowId(...)` — called unconditionally from the `Window` constructor (`Window.cs:125`) — dereferences it. With native-window creation skipped, the secondary ALC's `new Window()` crashes during construction:

- **X11**: `InvalidOperationException: X11DisplayInformationExtension couldn't find a X11XamlRootHost` → `Aborted (core dumped)` (exit 134).
- **Win32**: the same failure in `Win32Host`'s DisplayInformation init, on every ALC `new Window()`, until the run hangs to the 1h timeout.

`InitializeAlcWindowMode()` only *closes* that native window afterward — so on Win32/X11 it's needed transiently for `XamlRootHost`/`DisplayInformation` registration, not pointless. **macOS is the special case**: it both can't create the native window without colliding in `MacOSWindowHost`'s registration *and* its `DisplayInformation` extension tolerates the missing host — which is exactly why the skip is correctly scoped to macOS.

Avoiding the transient window on Win32/X11 too would mean making those `DisplayInformation` extensions tolerate a missing host (or deferring the ctor's DisplayInformation call for ALC windows) — a deeper platform change I'd rather not fold into this fix. So I'm keeping it macOS-only.

src/Uno.UI/UI/Xaml/Window/Implementations/DesktopWindow.cs
```diff
@@ -21,6 +21,11 @@ public DesktopWindow(Window window) : base(window)
...
+		{
+			return;
+		}
```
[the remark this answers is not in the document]
ajpinedam: Good catch — fixed in 5353b188b7 (+ test in 8a3af404c3).

You're right: the macOS hosted-ALC window still went through the ctor's unconditional `DisplayInformation.GetOrCreateForWindowId(AppWindow.Id)` (`Window.cs:125`), and `MacOSDisplayInformationExtension` subscribes to the **static** `MacOSWindowNative.NativeWindowReady` in its ctor, only unsubscribing once a native window fires it — which never happens here. So the subscription dangles: it pins the extension → `DisplayInformation` → the collectible ALC, and the next unrelated native window would bind this stale extension. Worth noting the existing `CloseAlcWindow` → `DisplayInformation.DestroyForWindowId` cleanup didn't cover it — that only clears the `_windowIdMap` entry, not the static-event subscription.

Fix: the macOS hosted-ALC window has no native window, so it now **skips DisplayInformation registration entirely**. I centralized the condition into `Window.IsMacOSHostedAlcWindow` (`OperatingSystem.IsMacOS() && ContentHostOverride is not null && IsAlcWindow`), reused by `DesktopWindow.Initialize`/the `Content` setter, and gated the ctor's `GetOrCreateForWindowId` on `!IsMacOSHostedAlcWindow`. No native window ⇒ no extension created ⇒ no subscription, so nothing to leak or mis-bind.

Added an internal `DisplayInformation.IsRegisteredForWindowId` seam and a regression test (`When_MacOSAlcWindow_Then_NoDisplayInformationRegistered`) asserting the macOS ALC window registers no `DisplayInformation`. `Given_AlcContentHost` runtime-validated on macOS: 28 passed / 0 failed.

src/Uno.UI/UI/Xaml/Window/Window.alc.cs
```diff
@@ -467,6 +418,16 @@ private void UnsubscribeFromContentHostEvents()
...
 
+	/// <summary>
+	/// True for a hosted secondary-ALC window on macOS. <see cref="Uno.UI.Xaml.Controls.DesktopWindow.Initialize"/>
+	/// skips native window creation in this case (it would collide with MacOSWindowHost's native-window
+	/// registration), so the window has no native backing — callers that depend on one must skip their work
+	/// too (the native window itself, and the DisplayInformation registration whose macOS extension subscribes
+	/// to <c>MacOSWindowNative.NativeWindowReady</c> and would never unsubscribe).
+	/// </summary>
+	internal bool IsMacOSHostedAlcWindow
+		=> OperatingSystem.IsMacOS() && ContentHostOverride is not null && IsAlcWindow;
```
MartinZikmund, line 429: Why specific to macOS? Can't all the changes apply generally for ALC windows on any platform?

ajpinedam: The issue is only happening on macOS, Windows is working fine. Tried to add a xplatform implementation (following your other comment) and it breaks the app. More information here: https://github.com/unoplatform/uno/pull/23511#discussion_r3467987944

cc: @MartinZikmund

Xiaoy312 (approved):

PR #23555, opened 2026-06-25
fix(dp): DependencyProperty.Register returns existing property on exact-duplicate registration

### LINKED ISSUE #23554: DependencyProperty.Register throws on an exact-duplicate registration (WinUI returns the existing property)
## Current behavior

`DependencyProperty.Register(name, propertyType, ownerType, metadata)` **throws** when the same dependency property `(name, ownerType)` is registered a second time, even when the registration is identical (same `name`, `ownerType`, and `propertyType`):

```
System.InvalidOperationException: The dependency property Substitutions already exists in type ...ControlExample
 ---> System.ArgumentException: SR.Argument_AddingDuplicate__   (Dictionary duplicate-key add)
```

This surfaces as a `TypeInitializationException` the first time the owning type is used, taking down the whole feature that depends on it.

## Expected behavior

WinUI/WinAppSDK does **not** throw on an exact-duplicate registration — it returns the already-registered `DependencyProperty`. The official WinUI 3 Gallery relies on this: both `ControlExample` and `SampleCodePresenter` call `DependencyProperty.Register("Substitutions", typeof(IList<...>), typeof(ControlExample), ...)` (same name + owner + type) and it runs fine on WinUI. Uno should match: an exact-duplicate `(name, ownerType, propertyType)` registration should return the existing `DependencyProperty` instead of throwing. (A *different* `propertyType` for the same `name`+`owner` is a genuine conflict and may still throw.)

## How to reproduce

```csharp
public class A : Control
{
    public static readonly DependencyProperty FooProperty =
        DependencyProperty.Register("Foo", typeof(object), typeof(A), new PropertyMetadata(null));
}

public class B : Control
{
    // Same name + ownerType(A) + propertyType — re-registration.
    public static readonly DependencyProperty FooProperty =
        DependencyProperty.Register("Foo", typeof(object), typeof(A), new PropertyMetadata(null));
}
```

Touching both static fields throws on Uno; works on WinUI.

## Affected area

`src/Uno.UI/UI/Xaml/DependencyProperty.cs` — `Register` (the registry duplicate-key add).

## Environment

- Uno.Sdk `6.7.0-dev.93` (also expected on current `master`)
- Target: net10.0 (all targets, Skia renderer); discovered on `net10.0-desktop`

MartinZikmund:
**GitHub Issue:** closes #23554

## PR Type:

🐞 Bugfix

## What changed? 🚀

`DependencyProperty.Register` / `RegisterAttached` threw `InvalidOperationException` ("…already exists in type…") when the same dependency property was registered twice with the **same `name`, `ownerType`, and `propertyType`**. WinUI does not throw here — the official WinUI Gallery registers `"Substitutions"` on `typeof(ControlExample)` from **both** `ControlExample` and `SampleCodePresenter`, and it runs fine on WinUI. On Uno the second registration threw a `TypeInitializationException`, taking down every feature that used those types.

I verified WinUI's behavior against the WinUI source — `microsoft/microsoft-ui-xaml`, `dxaml/xcp/dxaml/lib/MetadataAPI.cpp` (`MetadataAPI::RegisterDependencyProperty`): WinUI never validates or de-duplicates; every call allocates a fresh DP (there's even a `// TODO: Validate that people don't register the same DP multiple times…` comment). Allocating a brand-new DP each time is unsafe in Uno because its lookup relies on a unique `(ownerType, name)` registry, so this PR reproduces WinUI's **observable** behavior conservatively:

- `RegisterProperty` (now shared by `Register` and `RegisterAttached`) returns the **already-registered** `DependencyProperty` when an exact duplicate `(name, ownerType, propertyType)` is registered, instead of throwing.
- A same `(name, ownerType)` registration with a **different `propertyType`** is still a genuine conflict and keeps throwing `InvalidOperationException`.
- The previously-duplicated duplicate-key try/catch in `Register`/`RegisterAttached` is consolidated into `RegisterProperty`.

### ⚠️ Reviewer note (behavior change)

There was an existing test `When_Property_RegisterTwice_then_Fail` that **asserted the throw** on an exact duplicate — i.e. Uno intentionally threw before. To match WinUI it is repurposed to `When_Property_RegisterTwice_With_Same_Type_Then_Returns_Existing` (`Assert.AreSame`). The genuine-conflict case (different `propertyType`) is still covered by a new test. Please confirm this is the desired semantics. Nothing should rely on a DP *registration throwing*, so this is a permissive change rather than a functional break.

## Tests

`src/Uno.UI.UnitTests/DependencyProperty/Given_DependencyProperty.cs`:
- `When_Property_RegisterTwice_With_Same_Type_Then_Returns_Existing` — exact duplicate returns the same instance.
- `When_Property_RegisterTwice_With_Different_Type_Then_Fail` — different `propertyType` still throws.
- `When_AttachedProperty_RegisterTwice_With_Same_Type_Then_Returns_Existing` — same for `RegisterAttached`.

`dotnet test src/Uno.UI.UnitTests --filter FullyQualifiedName~Given_DependencyProperty` → **127/127 passed**.

## PR Checklist ✅

- [x] 🧪 Added unit tests
- [ ] 📚 Docs (n/a — XML doc comments updated on `Register`/`RegisterAttached`)
- [ ] 🖼️ Screenshots compare (n/a)
- [x] ❗ Contains **NO** breaking changes (permissive change — see reviewer note above; genuine conflicts still throw)

### DIFF
--- a/src/Uno.UI/UI/Xaml/DependencyProperty.cs
+++ b/src/Uno.UI/UI/Xaml/DependencyProperty.cs
@@ -163,8 +163,8 @@ namespace Microsoft.UI.Xaml
 		/// <param name="propertyType">The type of the property</param>
 		/// <param name="ownerType">The owner type of the property</param>
 		/// <param name="typeMetadata">The metadata to use when creating the property</param>
-		/// <returns>A dependency property instance. If a property with the same name and property type has already been registered for the ownerType, that existing instance is returned.</returns>
-		/// <exception cref="InvalidOperationException">A property with the same name but a different property type has already been declared for the ownerType</exception>
+		/// <returns>A dependency property instance. If a non-attached property with the same name and property type has already been registered for the ownerType, that existing instance is returned.</returns>
+		/// <exception cref="InvalidOperationException">A property with the same name but a different property type, or an attached property, has already been declared for the ownerType</exception>
 		public static DependencyProperty Register(
 			string name,
 			[DynamicallyAccessedMembers(BindableType.TypeRequirements)] Type propertyType,
@@ -225,8 +225,8 @@ namespace Microsoft.UI.Xaml
 		/// <param name="propertyType">The type of the property</param>
 		/// <param name="ownerType">The owner type of the property</param>
 		/// <param name="defaultMetadata">The metadata to use when creating the property</param>
-		/// <returns>A dependency property instance. If a property with the same name and property type has already been registered for the ownerType, that existing instance is returned.</returns>
-		/// <exception cref="InvalidOperationException">A property with the same name but a different property type has already been declared for the ownerType</exception>
+		/// <returns>A dependency property instance. If an attached property with the same name and property type has already been registered for the ownerType, that existing instance is returned.</returns>
+		/// <exception cref="InvalidOperationException">A property with the same name but a different property type, or a non-attached property, has already been declared for the ownerType</exception>
 		public static DependencyProperty RegisterAttached(
 			string name,
 			[DynamicallyAccessedMembers(BindableType.TypeRequirements)] Type propertyType,
@@ -438,16 +438,12 @@ namespace Microsoft.UI.Xaml
 			[DynamicallyAccessedMembers(BindableType.TypeRequirements)] Type propertyType,
 			DependencyProperty newProperty)
 		{
-			// WinUI does not validate or de-duplicate registrations: every call to
-			// MetadataAPI::RegisterDependencyProperty creates a fresh DP, so registering the
-			// same (name, ownerType, propertyType) twice is allowed there. The official WinUI
-			// Gallery relies on this (ControlExample and SampleCodePresenter both register
-			// "Substitutions" on typeof(ControlExample)). To stay faithful while keeping Uno's
-			// (ownerType, name) registry unique, return the already-registered property on an
-			// exact duplicate instead of throwing. A same-name registration with a different
-			// propertyType is still a genuine conflict and keeps throwing.
+			// WinUI does not de-duplicate registrations (MetadataAPI::RegisterDependencyProperty creates a fresh
+			// DP each time) and apps rely on it, so an exact duplicate returns the already-registered property
+			// instead of throwing. A different propertyType or attached-ness is still a genuine conflict.
 			if (_registry.TryGetValue(ownerType, name, out var existingProperty) &&
-				existingProperty!.Type == propertyType)
+				existingProperty!.Type == propertyType &&
+				existingProperty.IsAttached == newProperty.IsAttached)
 			{
 				return existingProperty;
 			}
--- a/src/Uno.UI/UI/Xaml/DependencyProperty.cs
+++ b/src/Uno.UI/UI/Xaml/DependencyProperty.cs
@@ -163,8 +163,8 @@ namespace Microsoft.UI.Xaml
 		/// <param name="propertyType">The type of the property</param>
 		/// <param name="ownerType">The owner type of the property</param>
 		/// <param name="typeMetadata">The metadata to use when creating the property</param>
-		/// <returns>A dependency property instance</returns>
-		/// <exception cref="InvalidOperationException">A property with the same name has already been declared for the ownerType</exception>
+		/// <returns>A dependency property instance. If a property with the same name and property type has already been registered for the ownerType, that existing instance is returned.</returns>
+		/// <exception cref="InvalidOperationException">A property with the same name but a different property type has already been declared for the ownerType</exception>
 		public static DependencyProperty Register(
 			string name,
 			[DynamicallyAccessedMembers(BindableType.TypeRequirements)] Type propertyType,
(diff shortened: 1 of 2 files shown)


MartinZikmund:
@pr23555-comment.md

MartinZikmund:
## Review comments addressed — pushed `b33cdc4d94`

All 10 open threads triaged against the current code on `dev/mazi/dp-register-idempotent`. Several were already satisfied by the earlier commit `34d8f8e10c` ("Guard duplicate DP registration on attached-ness") and by the latest test revision, so only two distinct points needed code.

### Fixed (5 threads / 2 distinct points)

|---|---|---|
| `src/Uno.UI/UI/Xaml/DependencyProperty.cs:176` (×3 threads) | `new DependencyProperty(...)` is allocated (burning an `Interlocked.Increment` global id) *before* `RegisterProperty` detects the duplicate | Valid — fixed, but not with the suggested diff. Instead of duplicating the fast-path guard into both `Register` and `RegisterAttached` (two copies of the same predicate), `RegisterProperty` now takes `(ownerType, name, propertyType, typeMetadata, attached)` and constructs the `DependencyProperty` **after** the duplicate check. A duplicate registration now allocates nothing, burns no unique id, and skips `FixMetadataIfNeeded` (its result was discarded anyway). The guard stays in exactly one place. |
| `src/Uno.UI.UnitTests/DependencyProperty/Given_DependencyProperty.cs:398` (×2 threads) | New tests should carry `[GitHubWorkItem(...)]` per `.claude/rules/unit-tests.md` | Added `[GitHubWorkItem("https://github.com/unoplatform/uno/issues/23554")]` to all four new tests. (`GitHubWorkItemAttribute` ships in MSTest.TestFramework 4.x — the repo pins `MSTestVersion` 4.2.1 — and the file already has the `Microsoft.VisualStudio.TestTools.UnitTesting` using.) |

### No change needed (5 threads)

| file:line | the point | why no change |
|---|---|---|
| `DependencyProperty.cs:450` (×2 threads) | Duplicate guard should also compare `IsAttached`, not just `Type` | Already implemented in `34d8f8e10c`: `RegisterProperty` compares `existingProperty.Type == propertyType && existingProperty.IsAttached == attached`, and `When_Property_Registered_Then_RegisterAttached_With_Same_Type_Then_Fail` covers the cross-register case. |
| `Given_DependencyProperty.cs:405` (×2 threads) | Test reuses `PropertyMetadata("42")` for both calls, so "first registration's metadata wins" isn't exercised | Already implemented: the second `Register`/`RegisterAttached` call passes `new PropertyMetadata("99")` and the tests assert `Assert.AreEqual("42", …GetMetadata(typeof(MockDependencyObject)).DefaultValue)`. |
| `DependencyProperty.cs:448` | The 7-line comment on the guard is a wall of text (`code-style.md`) | Already trimmed to three lines stating only the non-obvious *why* (WinUI does not de-duplicate; exact duplicate returns the existing DP; a differing type/attached-ness is still a conflict). This change adds one further line noting the check now precedes construction — kept to the minimum that earns its place. |

### Needs @MartinZikmund input (0) — none left open

Nothing was left unaddressed. The only outstanding question is the one already in the PR description: the intentional behavior change from *throwing* to *returning the existing DP* on an exact duplicate (the old `When_Property_RegisterTwice_then_Fail` test asserted the throw). That's a product decision for a human reviewer, not a code fix.

---

**Validation level:** changes were reviewed **by inspection only** — no build, no test run was performed in this pass (`dotnet build`/`dotnet test` were intentionally not executed). Azure Pipelines CI will verify. The trimming annotations were carried across with the moved construction (`RegisterProperty`'s `ownerType` now carries `[DynamicallyAccessedMembers(BindableType.TypeRequirements)]`, matching the `DependencyProperty` constructor's requirement), but CI is the authority on that.

Xiaoy312 (approved):

src/Uno.UI/UI/Xaml/DependencyProperty.cs
```diff
@@ -450,11 +414,41 @@ internal static DependencyProperty[] GetInheritedPropertiesForType(Type type)
...
 		{
+			// WinUI does not de-duplicate registrations (MetadataAPI::RegisterDependencyProperty creates a fresh
+			// DP each time) and apps rely on it, so an exact duplicate returns the already-registered property
+			// instead of throwing. A different propertyType or attached-ness is still a genuine conflict.
+			// Checked before the property is created, so a duplicate neither allocates nor burns a unique id.
+			if (_registry.TryGetValue(ownerType, name, out var existingProperty) &&
+				existingProperty!.Type == propertyType &&
+				existingProperty.IsAttached == attached)
+			{
+				return existingProperty;
```
morning4coffe-dev, line 432: [P2] Do not conflate separate WinUI dependency-property registrations

WinUI's MetadataAPI::RegisterDependencyProperty allocates a fresh property index, creates a new CCustomDependencyProperty, and queues the metadata supplied on that call: https://github.com/microsoft/microsoft-ui-xaml/blob/439bda90582f7892ca595a193f55dbfc6ae2010c/dxaml/xcp/dxaml/lib/MetadataAPI.cpp#L226-L269 . Register/RegisterAttached forward directly to it.

Returning the first object here aliases independent registrations and discards the second default value/callback. The new Assert.AreSame and first-default-wins assertions therefore test an Uno-specific workaround, not WinUI's observable contract. The registry's current one-name/one-entry design explains the constraint but does not make the semantics equivalent. Preserve separate identities/metadata while defining name lookup, or explicitly separate this compatibility compromise from a WinUI-parity fix; add a native-reference identity/default-value case.

morning4coffe-dev (changes requested):
The important WinUI compatibility issue is identity/metadata, not only whether a duplicate registration throws. The new first-registration-wins behavior is observably different from WinUI and the tests currently lock in that difference.

PR #23557, opened 2026-06-25
fix(xaml): escape quotes in generated string literals for object-typed properties

### LINKED ISSUE #23556: XAML source generator emits invalid C# for a quote-containing string assigned to an object-typed property
## Current behavior

The XAML source generator emits **invalid C#** when a XAML attribute assigns a string value that **contains double-quote characters** to a property whose type is `object` (or is otherwise taken as a raw string). It produces a C# *verbatim* string literal `@"..."` without doubling the embedded `"`.

Given a property `Value` of type `object`:

```xml
<local:MyItem Value="IsEnabled=&quot;False&quot; " />
```

the generator emits (paraphrased):

```csharp
new local::MyItem { Value = @"IsEnabled="False" " }
```

`@"IsEnabled="False" "` is not a valid verbatim string literal — the embedded `"` must be doubled. The consuming compilation then fails with errors such as:

- `CS0117: '...' does not contain a definition for 'False'`
- `CS0747: Invalid initializer member declarator`
- `CS1003: Syntax error, ',' expected`

The same happens for any quote-containing literal assigned to an `object`/raw-string target (e.g. `Value="link=&quot;https://...&quot;"`). String values without embedded quotes are emitted correctly, so only quote-containing literals on object/non-string targets are affected.

## Expected behavior

The generated code should be a valid C# string literal regardless of content — embedded `"` (and `\`, newlines, etc.) must be escaped. e.g. `@"IsEnabled=""False"" "` or `"IsEnabled=\"False\" "`. WinUI/WinAppSDK compiles this XAML correctly.

## How to reproduce

Assign a `&quot;`-containing string to an `object`-typed dependency/CLR property of a control via XAML and build. Real-world repro: the WinUI 3 Gallery's `ControlExample` samples set `ControlExampleSubstitution.Value` (typed `object`) to values like `Value="IsEnabled=&quot;False&quot; "`.

## Affected area

`src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.cs` (string-literal emission for object-typed / raw-string property values, e.g. `BuildLiteralValue`).

## Environment

- Uno.Sdk `6.7.0-dev.93` (also expected on current `master`)
- Target: net10.0 (all targets, Skia renderer); discovered on `net10.0-desktop`

MartinZikmund:
**GitHub Issue:** closes #23556

## PR Type:

🐞 Bugfix

## What changed? 🚀

The XAML source generator emitted **invalid C#** when a string value containing double-quotes was assigned to an `object`-typed (or implicit-`ToString`) property. In `XamlFileGenerator.BuildLiteralValue`, the `hasImplictToString || SpecialType.System_Object` branch wrapped the value in a C# **verbatim** literal `@"..."` without doubling embedded `"`, so `Value="IsEnabled=&quot;False&quot; "` produced `Value = @"IsEnabled="False" "` — which doesn't compile (CS0117/CS0747/CS1003). Quoted URLs etc. broke the same way.

Fix: route that branch through the generator's existing `DoubleEscape` helper (the same one the `System_String` path already uses), emitting a properly-escaped regular literal:

```diff
- return "@\"(" + memberValue + ")\"";        + return "\"(" + DoubleEscape(memberValue) + ")\"";
- return "@\"" + memberValue + "\"";           + return "\"" + DoubleEscape(memberValue) + "\"";
```

`DoubleEscape` handles `"`, `\`, `\n`, `\r`, `\t` and is null-safe. So e.g. `Value="IsEnabled=&quot;False&quot; "` now generates `Value = "IsEnabled=\"False\" ",`.

### Snapshot churn (expected, no behavior change)

This normalizes 18 existing golden snapshots for object-typed values from `Member = @"x"` to `Member = "x"` (verbatim → regular literal; **values are byte-for-byte identical** — verified every changed line is purely the `@"…"` → `"…"` form). Scope was kept to the object/implicit-string branch only; other unescaped `@"..."` sites (Geometry path data, FontFamily, URI/ImageSource) receive values that don't contain `"` and were left untouched to keep the diff minimal — happy to harden those too if preferred.

## Tests

`src/SourceGenerators/Uno.UI.SourceGenerators.Tests/XamlCodeGeneratorTests/Given_LiteralValue.cs` — `When_Object_Property_With_Quotes`: assigns `Value="IsEnabled=&quot;False&quot; "` to an `object` property; the verifier compiles the generated C# (so it asserts it compiles) and snapshot-compares (golden = `Value = "IsEnabled=\"False\" ",`). Confirmed it fails without the fix and passes with it.

Full test project (snapshot check on): **480 passed, 2 skipped** (the 2 skips are pre-existing "known to crash roslyn" cases); no existing test broken.

## PR Checklist ✅

- [x] 🧪 Added a source-generator test (+ updated golden snapshots)
- [ ] 📚 Docs (n/a)
- [ ] 🖼️ Screenshots compare (n/a)


MartinZikmund:
## PR review comments addressed

No code changes were needed. Both open threads are the same low-priority coverage suggestion from the ultrareview bot, and the code is already correct by inspection. Current PR head is `70529ed5a6d7eb09a0ba30842d4d667cfecb792c` (no new commit was pushed).

### No change needed (2)

| file:line | the point | why the code is already right |
|---|---|---|
| `src/SourceGenerators/Uno.UI.SourceGenerators.Tests/XamlCodeGeneratorTests/Given_LiteralValue.cs:12` | Add a companion test for the template-binding attached-property branch (`isTemplateBindingAttachedProperty = true`, line 5273 wrapping in `"(…)"`) to make coverage explicit. | The two branches in `XamlFileGenerator.cs` (line 5273 template-binding, line 5277 non-template-binding) call the byte-identical `DoubleEscape(memberValue)`; only the `(…)` wrapper differs. `DoubleEscape` (line 5460) escapes `\`, `"`, `\n`, `\r`, `\t`. The escaping mechanism under test is the same for both branches, so there is no correctness gap — the bot agrees ("no correctness gap", "Low priority given the identical code path"). |
| `src/SourceGenerators/Uno.UI.SourceGenerators.Tests/XamlCodeGeneratorTests/Given_LiteralValue.cs:58` | Same suggestion — a companion test exercising an attached property assigned inside a `ControlTemplate` would make both arms explicit. | Duplicate of the thread above; same reasoning applies. Optional coverage for an identical code path, not a defect. |

### Why the suggested test was not added

Authoring a new source-generator test that reliably hits the template-binding branch (an attached property with a quote-containing string inside a `ControlTemplate`) and produces the exact generated output is a change that needs a compile to trust. This run operated under a strict no-build constraint, so adding it blindly risked landing a broken/false-coverage test and turning CI red for a change the bot itself rates low priority with no correctness gap. If explicit two-arm coverage is desired, it is a safe follow-up to add once it can be built and verified locally.

---

Changes reviewed by inspection only; I did not build, run, or test anything here. Azure Pipelines CI will verify on the branch.

morning4coffe-dev (approved):
The production change uses the existing ordinary-string escape helper in the object-valued literal branch, matching the emitted quote style. DoubleEscape preserves backslashes, embedded quotes, CR/LF and tabs, and the attached TemplateBinding path keeps its parentheses. The representative snapshot changes follow the same transformation. This preserves the XAML string value rather than introducing a new WinUI-visible conversion rule; no blocking source-level issue found.

PR #23593, opened 2026-06-29
fix(android): Avoid white flash on first frame

MartinZikmund:
Avoids the white flash between Android splash-screen dismissal and the first Skia paint.

The splash is held (`OnPreDraw` keeps returning `false`) until the GL thread has flushed the first Skia frame. A `volatile bool FirstFrameRendered` is set on the GL thread after the first `_context.Flush()`, which posts an invalidation so `OnPreDraw` re-evaluates and dismisses the splash only once pixels are actually ready.

### Validation
- **Compile:** `Uno.UI.Runtime.Skia.Android.csproj` (net10.0-android) builds clean — 0 errors (pre-existing XA0101 `@(Content)` warning only).
- **Runtime:** pending device / CI.

https://claude.ai/code/session_01KBAigLEAf3csGwnWE22Yr5

### DIFF
--- a/src/Uno.UI.Runtime.Skia.Android/ApplicationActivity.cs
+++ b/src/Uno.UI.Runtime.Skia.Android/ApplicationActivity.cs
@@ -44,6 +44,12 @@ namespace Microsoft.UI.Xaml
 		private bool _isContentViewSet;
 
 		/// <summary>
+		/// Set by the GL thread after the first Skia frame has been rendered.
+		/// Used to keep the splash screen visible until pixels are ready.
+		/// </summary>
+		internal volatile bool FirstFrameRendered;
+
+		/// <summary>
 		/// The windows model implies only one managed activity.
 		/// </summary>
 		internal static ApplicationActivity Instance { get; private set; } = null!;
--- a/src/Uno.UI.Runtime.Skia.Android/ApplicationActivity.cs
+++ b/src/Uno.UI.Runtime.Skia.Android/ApplicationActivity.cs
@@ -43,10 +43,7 @@ namespace Microsoft.UI.Xaml
 		private static bool _started;
 		private bool _isContentViewSet;
 
-		/// <summary>
-		/// Set by the GL thread after the first Skia frame has been rendered.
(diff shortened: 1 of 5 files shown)


Xiaoy312 (approved):

MartinZikmund:
### Addressed review comments

Pushed as `2c6d054bb2`. All changes were **reviewed by inspection only** — no local build/test was run (CI on Azure Pipelines will verify).

### Fixed (3)

|---|---|---|
| `src/Uno.UI.Runtime.Skia.Android/Rendering/UnoSKVulkanView.cs` (RenderFrame) | Vulkan render path never sets `FirstFrameRendered`, so with `UseVulkanOnSkiaAndroid = true` the `OnPreDraw` gate stays `false` and the splash never dismisses. | Added the same signal-and-`Post(Invalidate)` block to `UnoSKVulkanView.RenderFrame()`. |
| `src/Uno.UI.Runtime.Skia.Android/Rendering/UnoSKCanvasView.cs:231` | Duplicate of the above (still-blocking follow-up). | Resolved by the same Vulkan fix. |
| `src/Uno.UI.Runtime.Skia.Android/ApplicationActivity.cs:50` | Style nit: `internal` field shouldn't carry an XML-doc (`///`) comment; use a short `//` instead. | Demoted to a one-line `//` comment, and corrected "GL thread" → "GL/Vulkan render thread" since the field is now written from both paths. |

**Placement note (deviation from the bot's literal suggestion):** the suggestion placed the flag-set *after* `_vulkanContext.RenderFrame(...)` returns. I instead placed it *inside* the composition callback, right after the frame content is drawn and past the `compositionTarget == null` early-return. Reason: on the Vulkan path `InvalidateRender()` can fire from `SurfaceChanged` before WinUI content is ready; in that case the callback early-returns (nothing composed) but `RenderFrame` still presents an empty swapchain image and returns `true`. Setting the flag unconditionally after `RenderFrame` would dismiss the splash on that blank frame — reintroducing the exact flash this PR fixes. Guarding on "content was actually composed" mirrors the GL path's semantics (`OnDrawFrame` only runs with a live `RootElement`).

### No change needed (0)

PR #23596, opened 2026-06-29
[WIP] fix: Cancel click on touch long-press menu

### LINKED ISSUE #22229: [WASM] Long press on clicks before opening context menu
### Current behavior 🐛

Long pressing on an element (like say a button) clicks that button as well as opening an attached context flyout.

### Expected behavior 🎯

Long pressing should open the context flyout ***instead of*** clicking the element, treating it the same as a right click.

Only tested right now on WASM Skia, not entirely sure how the other platforms behave.

```xml
<Button Command="{x:Bind ...}" CommandParameter="{x:Bind}">
  <TextBlock Text="Button content"/>

  <Button.ContextFlyout>
    <MenuFlyout>
      <MenuFlyoutItem Text="Some Menu Item" Command="{x:Bind}"/>
    </MenuFlyout>
  </Button.ContextFlyout>
</Button>
```

Observe that long pressing on the button opens the menu but also executes the button command.

Also, I believe the context menu should open as soon as the long-press is held long enough, not wait until you release the long press, but idk.

### Workaround 🛠️

_No response_

### Renderer 🎨

- [x] Skia
- [ ] Native

### Affected platforms 📱💻🖥️

WebAssembly

6.4.53, 6.5.0-dev.100

### IDE version 🧑‍💻

_No response_

_No response_

MartinZikmund:
On Skia targets, long-pressing a control (e.g. a Button) with a
ContextFlyout both opened the menu AND raised Click, instead of
behaving like a right-tap. WinUI suppresses the click because opening
the menu's popup steals the OS pointer capture (PointerCaptureLost);
Uno's in-canvas popups don't, so the pressed state was never cleared.

When a context menu is shown on a touch press-and-hold, release the
pressing element's pointer capture and raise an unhandled
PointerCaptureLost so the control clears its pressed state and does not
click on release.

fixes #22229

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01WQEkhvePgnJMns2KhVNUxr**GitHub Issue:** closes #

## PR Checklist ✅

### DIFF
--- a/src/Uno.UI.Runtime.Skia.WebAssembly.Browser/WasmCSS/uno.css
+++ b/src/Uno.UI.Runtime.Skia.WebAssembly.Browser/WasmCSS/uno.css
@@ -16,6 +16,17 @@ body {
     Remarks 2: If applied only on the root of the app, this won't break the scrolling (unlike if applied on all UIElement).
   */
   touch-action: none;
+  /*
+    Disable the browser's text-selection and long-press callout gestures.
+    On touch devices the browser starts a text-selection gesture on long-press (firing a haptic
+    buzz and showing selection/callout UI) before the cancelable 'contextmenu' event. Uno draws all
+    text and selection with Skia, so the browser selection is never useful and only conflicts with
+    the app's own gestures (e.g. press-and-hold to open a ContextFlyout). The hidden IME input
+    re-enables selection on itself so text editing keeps working.
+  */
+  user-select: none;
+  -webkit-user-select: none;
+  -webkit-touch-callout: none;
 }
 
 .uno-root-element {
--- a/src/Uno.UI.Runtime.Skia.WebAssembly.Browser/ts/Runtime/BrowserInvisibleTextBoxViewExtension.ts
+++ b/src/Uno.UI.Runtime.Skia.WebAssembly.Browser/ts/Runtime/BrowserInvisibleTextBoxViewExtension.ts
@@ -74,6 +74,10 @@
 			input.style.textShadow = "none";
 			input.style.overflow = "hidden";
 			input.style.pointerEvents = "none";
+			// Re-enable text selection on the IME input: the root disables it (uno.css) to suppress the
+			// browser's long-press selection gesture, but the input still needs it for editing/IME.
+			input.style.userSelect = "text";
+			(input.style as any).webkitUserSelect = "text";
 			input.style.zIndex = "99";
 			input.style.top = "0px";
 			input.style.left = "0px";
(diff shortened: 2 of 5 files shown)


MartinZikmund:
## Review comments addressed — `baec68336893cc2780ac97061fe9db3f9bd07b02`

All 5 unresolved threads are handled. **Validation level: code inspection only — nothing was built, run, or tested locally. CI (Azure Pipelines) will verify.**

### Fixed (5)

|---|---|---|
| `ContextMenuProcessor.cs:122` | `_contextMenuOnHoldingPointerId` is `0` when `SetContextMenuOnHoldingPointer` was never called; add an explicit early-return to make the invariant clear | Fixed, **but not with the suggested `== 0` sentinel** — see the note below. The field is now `uint?` and the method early-returns on `is not { } holdingPointerId`. Same clarity, no sentinel hazard. |
| `ContextMenuProcessor.cs:116` | drop `(issue #22229)` from the doc comment | Removed. |
| `UIElement.Pointers.cs:1656` | drop `(issue #22229)` from the doc comment | Removed. |
| `Given_ContextRequested.Injection.cs:423` | drop `(issue #22229)` from the assertion message | Removed. |
| `Given_ContextRequested.Injection.cs:479` | drop `(issue #22229)` from the assertion message | Removed. |

Also swept one unflagged occurrence of the same nit in the prose comment at `Given_ContextRequested.Injection.cs:449`. The `[GitHubWorkItem]` attributes remain as the traceability link, per the reviewer's own reasoning.

### Note on the `== 0` guard — suggestion amended, not taken as written

The suggested guard assumed *"no real pointer uses ID 0"*. That does not hold on Android, where `PointerHelpers.GetPointerId` composes the id as:

```csharp
return ((uint)nativeEvent.GetPointerId(pointerIndex) & _pointerIdsCount) << _pointerIdsShift | (uint)nativeEvent.DeviceId;
```

The first finger has `GetPointerId(0) == 0`, so a device reporting `DeviceId == 0` produces a **legitimate pointer id of `0`**. Treating `0` as "unset" would silently skip the capture release there and reintroduce the very bug this PR fixes — on the exact platform (touch) it targets.

So the reviewer's *intent* (make "no holding pointer recorded" explicit rather than relying on a scan that happens to find nothing) is implemented, with a nullable field that carries the distinction unambiguously:

```csharp
// Nullable as 0 is a valid pointer id (e.g. first finger on an Android device with deviceId 0).
private uint? _contextMenuOnHoldingPointerId;

private void ReleaseContextMenuHoldingPointerCapture()
{
    // No holding pointer was recorded (e.g. ContextRequested raised without a holding gesture).
    if (_contextMenuOnHoldingPointerId is not { } holdingPointerId)
    {
        return;
    }
    ...
}
```

This path is reachable: `RaiseContextRequestedEvent(..., isTouchInput: true)` can be called without a preceding `SetContextMenuOnHoldingPointer` (the setter has exactly one caller, the `HoldingState.Started` branch in `UIElement.Pointers.cs`), so the guard is worth having.

### No change needed (0) / Needs @MartinZikmund input (0)

Nothing escalated; no threads left open.

Issue #23142, opened 2026-04-27
[Native AOT] [Gallery App] The webview component does not load in the Gallery app on Desktop when native aot is enabled. Other elements will no longer load in the app


khaskoul:
### Current behavior 🐛

https://github.com/user-attachments/assets/4daeb848-19c8-4870-9a30-7ec1004b1ddb

The webview component does not load and all other elements will no longer become visible

### Expected behavior 🎯

Webviews load the user can still access all other components

### How to reproduce it (as minimally and precisely as possible) 🔬

1. Launch Uno Gallery with Native AOT enabled
2. Expand the UI features section
3. Select any UI Component (Other than webview)
4. Go to Webview
5. Observe that the webview element is not loaded
6. Attempt to navigate to any other page
7. Observe that the elements no longer load
8. Attempt to collapse sections in the navigation bar
9. Observe that the sections do not appear to fully collapse but their child items are not visible

### Renderer 🎨

- [ ] Skia
- [ ] Native

### Uno.Sdk version (and other relevant versions) 📦

uno.sdk: 6.6.0-dev.216

### IDE version 🧑‍💻

VS Code 1.117.0

jonpryor:
[khaskoul](https://github.com/khaskoul) mentioned that this happens with Uno.Sdk 6.6.0-dev.216.

In PowerShell terms:

```pshell
git clone https://github.com/unoplatform/Uno.Gallery.git
cd Uno.Gallery
git checkout 54a08b1587aa020a5651d61eb3a3023a79e89d72
sed -i '' 's/"Uno.Sdk": ".*"/"Uno.Sdk": "6.6.0-dev.216"/g' global.json
dotnet publish -c Release -r win-arm64 -f net10.0-desktop -p:TargetFrameworkOverride=net10.0-desktop `
  -bl Uno.Gallery/Uno.Gallery.csproj -p:SelfContained=true -p:UseSkiaRendering=true `
  -p:SkiaPublishAot=true -p:IsAotCompatible=true
```

(Which is a mismatch of Unix `sed`… but you get the idea).

What's interesting is that if we run `Uno.Gallery.exe` *on PowerShell* and pipe output to `Out-Default` so that it's visible, we see an error message after clicking on **UI components > WebView**:

```powershell
> .\Uno.Gallery\bin\Release\net10.0-desktop\win-arm64\publish\Uno.Gallery.exe | Out-Default
fail: Uno.UI.Dispatching.NativeDispatcher[0]
[...]

PR #23604, opened 2026-06-30
feat(win32): WebView2Aot backend for <WebView2/>

jonpryor:
Fixes: https://github.com/unoplatform/uno/issues/23142

Context: https://github.com/MicrosoftEdge/WebView2Feedback/issues/5238
Context: https://github.com/smourier/WebView2Aot
Context: https://github.com/smourier/DirectNAot/issues/9

On Windows, the <WebView2/> XAML control is implemented in terms of the [Microsoft.Web.WebView2 NuGet package][0].

The problem: Microsoft.Web.WebView2, when used for "normal" .NET, uses COM Interop.  Native AOT does not support COM interop. The result is that things fail, a'la
MicrosoftEdge/WebView2Feedback#5238 (which is for WinForms, but the same fundamental idea).

In the case of #23142, if we build Uno.Gallery for Native AOT:

	git clone https://github.com/unoplatform/Uno.Gallery.git
	cd Uno.Gallery
	git checkout 54a08b1587aa020a5651d61eb3a3023a79e89d72
	sed -i '' 's/"Uno.Sdk": ".*"/"Uno.Sdk": "6.6.0-dev.216"/g' global.json
	dotnet publish -c Release -r win-arm64 -f net10.0-desktop -p:TargetFrameworkOverride=net10.0-desktop `
	  -bl Uno.Gallery/Uno.Gallery.csproj -p:SelfContained=true -p:UseSkiaRendering=true `
	  -p:SkiaPublishAot=true -p:IsAotCompatible=true

Run it:

	.\Uno.Gallery\bin\Release\net10.0-desktop\win-arm64\publish\Uno.Gallery.exe | Out-Default

and access the **UI components > WebView** entry in the left tree, then the console contains the following output:

	fail: Uno.UI.Dispatching.NativeDispatcher[0]
	      NativeDispatcher unhandled exception
	      System.NotSupportedException: COM Interop requires ComWrapper instance registered for marshalling.
	         at System.Runtime.InteropServices.ComWrappers.ComInterfaceForObject(Object) + 0xa4
	         at System.Runtime.InteropServices.ComWrappers.ComInterfaceForObject(Object, Guid) + 0x24
	         at Microsoft.Web.WebView2.Core.CoreWebView2Environment.CreateCoreWebView2EnvironmentWithOptions(String, String, ICoreWebView2EnvironmentOptions, ICoreWebView2CreateCoreWebView2EnvironmentCompletedHandler) + 0xdc
	         at Microsoft.Web.WebView2.Core.CoreWebView2Environment.<CreateAsync>d__102.MoveNext() + 0xe4
	      --- End of stack trace from previous location ---
	         at Uno.UI.Runtime.Skia.Win32.Win32NativeWebView.<>c__DisplayClass15_0.<<-ctor>b__0>d.MoveNext() + 0x60
	      --- End of stack trace from previous location ---
	         at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__124_0(Object state) + 0x1c
	         at Uno.UI.Dispatching.NativeDispatcher.RunAction(NativeDispatcher, Action) + 0x8c

The Microsoft.Web.WebView2 package *does* support Native AOT, but *only if* you're building for WinRT.  (Uno is not built on WinRT.)

A way to use Microsoft.Web.WebView2 while supporting Native AOT in a "normal" (non-WinRT) Windows environment is to instead use smourier/WebView2Aot, in the [WebView2Aot NuGet package][1]. WebView2Aot uses the [DirectN AOT generator][2] to emit Native AOT- compatible COM interop infrastrucure.

Update `Uno.UI.Runtime.Skia.Win32` as follows:

 1. Split out the "native" bits of `Win32NativeWebView` into `Win32NativeWebViewBase.cs`.  (Anything dealing with a Windows WndProc, etc.)

 2. Update `Win32NativeWebView` to inherit `Win32NativeWebViewBase`.

 3. Add a new `Win32NativeAotWebView` type, which also inherits from `Win32NativeWebViewBase`.  This implements `<WebView2/>` support in terms of WebView2Aot.

However, WebView2Aot requires .NET 10.  Conseuqently, `Uno.UI.Runtime.Skia.Win32.csproj` should only reference WebView2Aot when targeting .NET 10 and later, and `Win32NativeAotWebView` can only exist in .NET 10 and later.

Additionally, this is a potential significant change.  We want WebView2Aot to be the default in all Windows configurations for consistency -- nobody wants to debug a scenario of "WebView2 works when I'm not building for Native AOT!" -- but there should be a way to use the previous Microsoft.Web.WebView2 implementation in case WebView2Aot breaks things, or for testing, or…

Update `Win32NativeWebViewProvider` as follows:

 1. When built for .NET 9, only Win32NativerWebView / the Microsoft.Web.WebView2 backend is supported.

 2. When built for .NET 10+, the new WebView2Aot backend is used by default.  This can be overridden by setting the `UNO_WEBVIEW2_BACKEND` environment variable to `microsoft.web.webview2`.

    `UNO_WEBVIEW2_BACKEND` can also be set to `webview2aot` to be explicit.

The `UNO_WEBVIEW2_BACKEND` environment variable is always read, and will log an error message when using an unsupported value, e.g. when `UNO_WEBVIEW2_BACKEND=webview2aot` on .NET 9.

Aside: add BannedApiAnalyzers to `Uno.UI.Runtime.Skia.Win32.csproj` which make it an error to use `PWSTR.From(string)`. See also smourier/DirectNAot#9.

[0]: https://www.nuget.org/packages/Microsoft.Web.WebView2
[1]: https://www.nuget.org/packages/WebView2Aot
[2]: https://github.com/smourier/DirectNAot

## PR Type:
🐞 Bugfix
✨ Feature

## PR Checklist ✅


src/Uno.UI.Runtime.Skia.Win32/Uno.UI.Runtime.Skia.Win32.csproj
```diff
@@ -26,6 +26,9 @@
 
 	<ItemGroup>
 		<PackageReference Include="Microsoft.Web.WebView2" Aliases="mswebview2" PrivateAssets="all" />
+		<PackageReference Include="WebView2Aot" Condition="$([MSBuild]::IsTargetFrameworkCompatible($(TargetFramework), 'net10.0'))" />
+
+		<PackageReference Include="Microsoft.CodeAnalysis.BannedApiAnalyzers" Version="3.3.4" PrivateAssets="All" />
```
Xiaoy312: move this and `AdditionalFiles` to a separate `ItemGroup`?
(or, do as Uno.UI + .UWP, putting it into directory.build.targets)

src/Uno.UI.Runtime.Skia.Win32/UI/Xaml/Controls/WebView/Win32NativeWebView.cs
```diff
@@ -18,21 +15,48 @@
...
 
 namespace Uno.UI.Runtime.Skia.Win32;
 
-// Heavily inspired/copied from WpfNativeWebView
-
 internal class Win32NativeWebViewProvider(CoreWebView2 owner) : INativeWebViewProvider
 {
 	public INativeWebView CreateNativeWebView(ContentPresenter contentPresenter)
+	{
+		var backend = Environment.GetEnvironmentVariable("UNO_WEBVIEW2_BACKEND")?.ToLowerInvariant();
```
Xiaoy312, line 29: this needs to be explained in a md file in the doc, not just pr description

src/Uno.UI.Runtime.Skia.Win32/UI/Xaml/Controls/WebView/Win32NativeWebView.cs
```diff
@@ -18,21 +15,48 @@
...
+#if NET10_0_OR_GREATER
+	private INativeWebView CreateWin32NativeAotWebView(ContentPresenter contentPresenter)
+		=> new Win32NativeAotWebView(owner, contentPresenter);
+
+	private INativeWebView CreateDefaultWebView(ContentPresenter contentPresenter)
+		=> CreateWin32NativeAotWebView(contentPresenter);
+#else   // !NET10_0_OR_GREATER
+	private INativeWebView CreateDefaultWebView(ContentPresenter contentPresenter)
+		=> CreateWin32NativeWebView(contentPresenter);
+#endif  // !NET10_0_OR_GREATER
```
Xiaoy312:
```suggestion
#else
	private INativeWebView CreateDefaultWebView(ContentPresenter contentPresenter)
		=> CreateWin32NativeWebView(contentPresenter);
#endif
```

Xiaoy312: or, minimally fix the the whitespaces

src/Uno.UI.Runtime.Skia.Win32/UI/Xaml/Controls/WebView/Win32NativeAotWebView.cs
Xiaoy312, line 1: not sure where to put those extra classes,
but maybe we can move them into: Win32NativeAotWebView.(deps/misc/etc).cs

src/Uno.UI.Runtime.Skia.Win32/UI/Xaml/Controls/WebView/Win32NativeWebView.cs
```diff
@@ -18,21 +15,48 @@
...
+				return CreateWin32NativeAotWebView(contentPresenter);
+#endif  // NET10_0_OR_GREATER
+			case "microsoft.web.webview2":
+				return CreateWin32NativeWebView(contentPresenter);
+			case "":
+			case null:
+				break;
+			default:
+				typeof(Win32Host).LogError()?.Error($"Unsupported `UNO_WEBVIEW2_BACKEND` value `{backend}!");
+				break;
```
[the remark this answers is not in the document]
Xiaoy312:
```suggestion
			default:
				typeof(Win32Host).LogError()?.Error(
#if NET10_0_OR_GREATER
					$"Unsupported `UNO_WEBVIEW2_BACKEND` value `{backend}`. Supported values: `webview2aot`, `microsoft.web.webview2`."
#else
					$"Unsupported `UNO_WEBVIEW2_BACKEND` value `{backend}`. Supported values: `microsoft.web.webview2`."
#endif
				);
				break;
```

src/Uno.UI.Runtime.Skia.Win32/UI/Xaml/Controls/WebView/Win32NativeAotWebView.cs
jonpryor: Split up into `Win32NativeAotWebView.{Request,Response,Helpers}.cs` in 0ed4f5d71c4411200569e34dc3e61ed4194c47da.

src/Uno.UI.Runtime.Skia.Win32/UI/Xaml/Controls/WebView/Win32NativeWebView.cs
jonpryor: Fixed the whitespace in 0ed4f5d71c4411200569e34dc3e61ed4194c47da.

src/Uno.UI.Runtime.Skia.Win32/UI/Xaml/Controls/WebView/Win32NativeWebView.cs
jonpryor: Docs updated in 0ed4f5d71c4411200569e34dc3e61ed4194c47da.

src/Uno.UI.Runtime.Skia.Win32/Uno.UI.Runtime.Skia.Win32.csproj
jonpryor: Used a separate item group in 0ed4f5d71c4411200569e34dc3e61ed4194c47da.

src/Uno.UI.Runtime.Skia.Win32/UI/Xaml/Controls/WebView/Win32NativeWebView.cs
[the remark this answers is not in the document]
jonpryor: Added a "Supported values: …" message in 0ed4f5d71c4411200569e34dc3e61ed4194c47da.

Xiaoy312 (approved):

ajpinedam (approved):

src/Directory.Build.targets
```diff
@@ -156,6 +156,7 @@
 		<PackageReference Update="CommunityToolkit.Mvvm" Version="$(CommunityToolkitMvvmVersion)" />
 		<PackageReference Update="System.CommandLine" Version="2.0.0-beta4.22272.1" />
 		<PackageReference Update="Microsoft.Web.WebView2" Version="1.0.2592.51" Condition="'$(MSBuildProjectName)' != 'Uno.WinAppSDKSyncGenerator.References'" />
+		<PackageReference Update="WebView2Aot" Version="1.5.0" />
```
MartinZikmund, line 159: The new package will need to be added to Uno.Sdk and uno.templates

jonpryor: @MartinZikmund: I don't understand "added to Uno.Sdk and uno.templates".

A `.nuspec` is generated for Uno.UI.Runtime.Skia.Win32 in Release builds: https://github.com/unoplatform/uno/blob/feature/breakingchanges/src/Uno.UI.Runtime.Skia.Win32/Uno.UI.Runtime.Skia.Win32.csproj#L10

```xml
		<GeneratePackageOnBuild Condition="'$(Configuration)'=='Release'">true</GeneratePackageOnBuild>
```

Indeed, when I build locally, the resulting generated `obj/Release/Uno.WinUI.Runtime.Skia.Win32.255.255.255.255.nuspec` contains:

```xml
      <group targetFramework="net10.0">
        <dependency id="Uno.WinUI.Runtime.Skia" version="255.255.255.255" exclude="Build,Analyzers" />
        <dependency id="Uno.WinUI" version="255.255.255.255" exclude="Build,Analyzers" />
        <dependency id="HarfBuzzSharp" version="8.3.1.1" exclude="Build,Analyzers" />
        <dependency id="Microsoft.Windows.CsWin32" version="0.3.106" exclude="Build,Analyzers" />
        <dependency id="Uno.icu-win" version="77.2.1" exclude="Build,Analyzers" />
        <dependency id="WebView2Aot" version="1.5.0" exclude="Build,Analyzers" />
      </group>
```

Is more needed?

Additionally, what needs to be done in uno.templates? Is there an example to copy?

MartinZikmund: I was under the impression that there was a prop to override WebView2 version, but it seems there is not currently. In that case we can probably skip that for now, maybe it will not be necessary then.

MartinZikmund: @jonpryor however, because the environment variable decides whether the AOT or non-AOT version is used, does it mean we now bundle both versions? In that case, it would probably be best to default to AOT version and have a prop in Uno.Sdk that would allow people, if they chose to, to switch to the non-AOT one, so that the choice happens before build, instead of having the option to switch at runtime

jonpryor: @MartinZikmund [asked](https://github.com/unoplatform/uno/pull/23604#discussion_r3505432294):

> does it mean we now bundle both versions?

Yes.

> it would probably be best to default to AOT version

That is the default, which is also mentioned in the updated WebView docs: https://github.com/unoplatform/uno/blob/dev/jonpryor/jonp-Windows%2BNativeAOT%2BWebView2/doc/articles/controls/WebView.md#windows-specifics

> instead of having the option to switch at runtime

The benefit to switching at runtime via environment variable is with testing: you can build once, then test either/both backends without requiring an app rebuild. (You would need to relaunch the app if setting the env var from the launching Console, but this is required anyway…)

The other benefit to the env var approach is that it adds no new public API, so we can change things without *API* breaks in the future.

Additionally, my hope is that for Uno *8* we drop the Microsoft.Web.WebView2 backend entirely, and use only WebView2Aot.

MartinZikmund: Runtime switching is nice for us, but I think for end users it will mean bigger assembly size, which is not ideal and is probably more important than the convenience... So if it is possible, it would be best to make sure only the selected version is bundled, via Uno.Sdk. As long as WebView2Aot seems stable, we can just keep it default and in case someone needs to, they could set some `.csproj` prop to swap for the old backend.

jonpryor:
#23621 -- this same idea but for master -- has been merged.

TODO: when feature/breakingchanges is updated to contain #23621, rebase this PR atop it to *remove* the old backend.

jonpryor:
@jonpryor [wrote](https://github.com/unoplatform/uno/pull/23604#issuecomment-4895101047):

> TODO: when feature/breakingchanges is updated to contain https://github.com/unoplatform/uno/pull/23621, rebase this PR atop it to remove the old backend.

I forgot I had that idea, and went off and created a new PR instead: #23757.

Closing this PR.

PR #23620, opened 2026-07-02
test: Migrate SamplesApp UI tests to runtime tests (batch 1)

MartinZikmund:
**GitHub Issue:** Related to #9811 (first batch of an ongoing effort; does not fully close it)

## PR Type:

💬 Other — 🧪 Test infrastructure: migrate `SamplesApp.UITests` (Selenium/Xamarin) into equivalent Uno runtime tests.

## What changed? 🚀

The Selenium/Xamarin-based `SamplesApp.UITests` **do not run on the Skia targets**. This PR migrates the first tractable batch of them into equivalent **runtime tests** (`Uno.UI.RuntimeTests`), which run on Skia (Desktop + WASM), native, and WinUI.

For each UITest class/method we: (1) judged whether it's meaningful on Skia, (2) translated it to a runtime test if so, and (3) removed the original. Non-translatable tests are **kept in place with a short explainer comment** for manual review.

**This batch:**
- **30 new runtime-test files** added under `src/Uno.UI.RuntimeTests/Tests/**` (~114 test cases, all green on **Skia Desktop / Win32**).
- **33 original UITest files removed** (migrated or already-covered by existing runtime tests).
- **5 UITest files kept with an explainer** (not applicable to Skia): `GyrometerTests` (sensor returns `null` on Skia), `TestFrameworkTests` (tests the Xamarin UITest harness itself), `Wasm/Given_CustomEvents` (WASM DOM), `BindableDrawerLayoutTests`, and `UnoSamples_Tests.XBind` (its `XBind_Validation` half relies on `x:Uid`, not resolvable by the runtime query shim — its `XBind_Functions` half was migrated).

**Translation approach (per test):**
- **Reconstruct inline** for simple, self-contained scenarios (build the controls directly, drive them, assert).
- **Load the real sample by name** (`SampleControlUITestBase.RunAsync`) for elaborate sample/harness pages.
- Real pointer taps via `InputInjector` (`#if HAS_INPUT_INJECTOR || WINAPPSDK`, gated off native heads with `[PlatformCondition]`).
- Pixel checks via `UITestHelper.ScreenShot` + `ImageAssert` (gated `#if !HAS_RENDER_TARGET_BITMAP [Ignore]`, i.e. skipped on WASM/Reference).
- Where a UITest was `[Ignore]`d/flaky, we tried to **enable** it as a runtime test (the flakiness was often a Selenium/native/screenshot artifact) rather than blindly carrying the skip forward.

Each wave was built and run via the `/runtime-tests` skill (Skia Desktop) and committed only when green. This is a foundation batch; further batches (the larger `TextBox`/`ListView`/`TreeView`/`Manipulation`/input suites) will follow.

New runtime tests span: AutomationId, Control (disabled input suppression), x:Bind functions, ToggleSwitch header, RadioMenuFlyoutItem, VisualStateManager, StaticResource, ContentControl, InfoBar, CheckBox, RadioButton(s), Pivot, Popup (light-dismiss + placement), NumberBox, Enability, ResourceLoader, App.xaml global resources, FocusManager (direction + input), NativeButtons, SplitView, SolidColorBrush/LinearGradientBrush/Geometry/Path/Shapes.

## PR Checklist ✅

- [x] 🧪 Added [Runtime tests, UI tests, or a manual test sample](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) (this PR *is* the tests)
- [ ] 📚 Docs have been added/updated following the [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (N/A — test-only change)
- [ ] 🖼️ Validated PR `Screenshots Compare Test Run` results (to validate on CI)

https://claude.ai/code


src/Uno.UI.RuntimeTests/Tests/Microsoft_UI_Xaml_Controls/Given_NumberBox_UITest.cs
```diff
@@ -0,0 +1,263 @@
...
+	{
+		var numberBox = new NumberBox
+		{
+			SpinButtonPlacementMode = NumberBoxSpinButtonPlacementMode.Inline,
+			Minimum = double.MinValue,
+			Maximum = double.MaxValue,
+			Value = 0,
+		};
+
+		await UITestHelper.Load(numberBox);
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in 1cf907601c — all six test methods in `Given_NumberBox_UITest` now wrap the body in `try { … } finally { WindowHelper.WindowContent = null; }`. Note the engine only clears content between tests when `IsUnloadingTestContent` is set, and `SampleChooserViewModel` builds a bare `new UnitTestEngineConfig()` on CI where that flag defaults to `false` — so on CI the `finally` is what actually resets it. Reviewed by inspection; CI is the validation gate.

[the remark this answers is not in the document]
MartinZikmund: Fixed in 1cf907601c (duplicate of the thread above) — all six methods now reset `WindowHelper.WindowContent` in a `finally`: `When_SpinButtons_Clicked_Changes_Value_By_SmallChange`, `When_Wrap_Enabled_SpinButtons_Wrap_At_Bounds`, `When_SpinButton_Clicked_Validates_Pending_Text_First`, `When_SpinButtons_IsEnabled_Reflects_Value_State`, `When_Minimum_Maximum_Value_Are_Coerced`, `When_AcceptsExpression_Evaluates`.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_RadioButton.Pointer.cs
```diff
@@ -0,0 +1,167 @@
...
+			if (TestServices.WindowHelper.IsXamlIsland)
+			{
+				return;
+			}
+
+			var rb1 = new RadioButton { Content = "RadioButton 1" };
+			var rb2 = new RadioButton { Content = "RadioButton 2" };
+			var panel = new StackPanel { Children = { rb1, rb2 } };
+
+			await UITestHelper.Load(panel);
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in 1cf907601c — all three methods in `Given_RadioButton.Pointer.cs` now wrap their body in `try { … } finally { TestServices.WindowHelper.WindowContent = null; }`. The `using var mouse` and the local `Tap` function stay inside the `try`, so the mouse is still disposed before the content reset.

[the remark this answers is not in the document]
MartinZikmund: Fixed in 1cf907601c (duplicate of the thread above) — `When_Tapped_Selects_And_Respects_IsEnabled`, `When_Tapped_Twice_Does_Not_Uncheck` and `When_IsEnabled_Toggled_State_Is_Preserved` all reset window content in a `finally` now.

src/Uno.UI.RuntimeTests/Tests/Microsoft_UI_Xaml_Controls/Given_NumberBox_Uno_UITest.cs
```diff
@@ -0,0 +1,93 @@
...
+	public async Task When_Header_Is_Custom_Content()
+	{
+		var headerContent = new TextBlock { Text = "This is a NumberBox Header" };
+		var numberBox = new NumberBox
+		{
+			SpinButtonPlacementMode = NumberBoxSpinButtonPlacementMode.Inline,
+			Header = headerContent
+		};
+
+		await UITestHelper.Load(numberBox);
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in 1cf907601c — `When_Header_Is_Custom_Content` and `When_Description_Is_Custom_Content` now use the same `try/finally { WindowHelper.WindowContent = null; }` shape that `When_Text_Parsed_With_Custom_DecimalFormatter` already had.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Input/Given_Enability_UITest.cs
```diff
@@ -0,0 +1,99 @@
...
+		{
+			return;
+		}
+
+		var output = new TextBlock();
+		var button = new Button { Content = "Disabled button", IsEnabled = false };
+		RegisterEvents(button, output);
+
+		var panel = new StackPanel { Children = { output, button } };
+		await UITestHelper.Load(panel);
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in 1cf907601c — both `When_ButtonDisabled_Then_NoPointerEvents` and `When_ButtonDisabling_Then_NoMorePointerEvents` now wrap the body in `try { … } finally { TestServices.WindowHelper.WindowContent = null; }`.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_RadioButton.Pointer.cs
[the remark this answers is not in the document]
MartinZikmund: Fixed in 1cf907601c — all three methods in `Given_RadioButton.Pointer.cs` now wrap their body in `try { … } finally { TestServices.WindowHelper.WindowContent = null; }`. The `using var mouse` and the local `Tap` function stay inside the `try`, so the mouse is still disposed before the content reset.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_ContentControl_UITest.cs
```diff
@@ -0,0 +1,101 @@
...
+using Uno.UITest.Helpers.Queries;
+
+namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls;
+
+/// <summary>
+/// Migrated from SamplesApp.UITests UnoSamples_Tests.ContentControl: exercises ContentPresenter /
+/// ContentControl content-template rendering and toggling by loading the real sample pages.
+/// The samples self-initialize (DataContext / template are set in their constructors), so
+/// RunAsync (which only calls the parameterless ctor) is sufficient — no ViewModel wiring needed.
+/// </summary>
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in 1cf907601c — the `<summary>` block is now a two-line `//` comment, matching the migration notes used by the other files in this PR (`Given_Pivot_UITest`, `Given_Enability_UITest`).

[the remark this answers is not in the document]
MartinZikmund: Fixed in 1cf907601c (duplicate of the thread above) — collapsed to a short `//` migration note.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_RadioButton.Pointer.cs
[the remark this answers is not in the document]
MartinZikmund: Fixed in 1cf907601c (duplicate of the thread above) — `When_Tapped_Selects_And_Respects_IsEnabled`, `When_Tapped_Twice_Does_Not_Uncheck` and `When_IsEnabled_Toggled_State_Is_Preserved` all reset window content in a `finally` now.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Automation/Given_AutomationId_UITest.cs
```diff
@@ -0,0 +1,83 @@
...
+using Uno.UI.RuntimeTests.Helpers;
+using Uno.UITest.Helpers.Queries;
+
+namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Automation;
+
+/// <summary>
+/// Migrated from SamplesApp.UITests AutomationId_Tests: verifies that a bound
+/// AutomationProperties.AutomationId flows onto ListView item containers and that
+/// selecting an item updates the bound result text.
+/// </summary>
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in 1cf907601c — replaced with the single-line `//` migration note you suggested.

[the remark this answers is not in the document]
MartinZikmund: Fixed in 1cf907601c (duplicate of the thread above) — now a single `//` line.

src/Uno.UI.RuntimeTests/Tests/Microsoft_UI_Xaml_Controls/Given_NumberBox_Uno_UITest.cs
```diff
@@ -0,0 +1,93 @@
...
+using SamplesApp.UITests;
+using Uno.UI.RuntimeTests.Helpers;
+using Uno.UITest.Helpers.Queries;
+using static Private.Infrastructure.TestServices;
+
+namespace Uno.UI.RuntimeTests.Tests.Microsoft_UI_Xaml_Controls;
+
+/// <summary>
+/// Migrated from SamplesApp.UITests Microsoft_UI_Xaml_Controls.NumberBoxTests.Given_NumberBox.Uno.
+/// </summary>
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in 1cf907601c — the `<summary>` block is now a single `//` migration line.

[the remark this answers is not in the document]
MartinZikmund: Fixed in 1cf907601c — the `<summary>` block is now a single `//` migration line.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_Pivot_UITest.cs
```diff
@@ -0,0 +1,108 @@
...
+		};
+
+		return (root, title, content, pivot);
+	}
+
+	private static DataTemplate CreateTemplate(string inner)
+		=> (DataTemplate)XamlReader.Load(
+			$"<DataTemplate xmlns=\"http://schemas.microsoft.com/winfx/2006/xaml/presentation\">{inner}</DataTemplate>");
+
+	public class MyCustomPivotItem
```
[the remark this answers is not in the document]
MartinZikmund: No change — declining this one on two counts.

1. The stated rationale does not hold: `MyCustomPivotItem` is a **nested** type, so it is already scoped to `Given_Pivot_UITest` and never enters the `Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls` namespace. There is nothing to un-pollute.
2. It is a data-binding source: it backs `{Binding Title}` / `{Binding Content}` inside a `XamlReader.Load`-ed `DataTemplate` plus the `SelectedItem.Title` / `SelectedItem.Content` `PropertyPath` bindings. These tests also run against native WinUI (`Uno.UI.RuntimeTests.Windows`), whose binding engine resolves CLR sources through XAML type metadata rather than plain reflection, so demoting the type to `private` is a potential behavioural change there, not a cosmetic one. This PR pass was reviewed by inspection only, so I am not taking a change I would need a WinUI run to trust for zero benefit.

On the nullable sub-point: the file is nullable-oblivious (no `#nullable enable`, and the project does not set `<Nullable>enable</Nullable>`), so the un-annotated `string` properties raise no warning and are consistent with the rest of the suite.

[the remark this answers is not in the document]
MartinZikmund: No change — same reasoning as the earlier thread on this type. A nested type does not live in the namespace, so there is nothing to pollute; and `MyCustomPivotItem` is the binding source for the `XamlReader`-built `DataTemplate`s and the `SelectedItem.Title`/`SelectedItem.Content` `PropertyPath` bindings, which also execute on the native-WinUI test head. Demoting a binding source to `private` is a runtime-behaviour change there, not a cosmetic one, and is not worth taking without a WinUI run to back it.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Shapes/Given_Path_UITest.cs
```diff
@@ -0,0 +1,109 @@
...
+
+namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Shapes;
+
+[TestClass]
+public class Given_Path_UITest
+{
+	// Matches SamplesApp.UITests Path_Tests.Test_FillRule / PathTestsControl.Path_FillRule sample:
+	// a self-intersecting outline with an inner "hole" figure, rendered with EvenOdd vs. Nonzero
+	// fill rules, expressed via the three ways Path.Data can be authored.
+	public enum DataStyle
```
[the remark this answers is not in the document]
MartinZikmund: No change — this suggestion would not compile. `DataStyle` is the parameter type of the `public` `[TestMethod] When_Path_FillRule(DataStyle style, FillRule fillRule)` and is used in its `[DataRow]`s. Making the enum `private` (or `internal`) on a `public` class gives **CS0051: Inconsistent accessibility: parameter type `Given_Path_UITest.DataStyle` is less accessible than method `When_Path_FillRule`**. MSTest requires the test method to stay `public`, so the enum has to stay `public` too.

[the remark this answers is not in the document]
MartinZikmund: No change — duplicate of the earlier `DataStyle` thread. Making it `private` breaks the build: it is the parameter type of the `public` `[TestMethod] When_Path_FillRule(DataStyle, FillRule)`, so a less-accessible enum triggers CS0051. It must stay `public`.

MartinZikmund:
Addressed the 2 remaining unresolved review threads. Both are re-flags of suggestions that are already settled on the branch, so **no new commit was needed** — the current branch HEAD `59504099b6` already contains the correct state for both. Changes were reviewed **by inspection only**; Azure Pipelines CI is the validation gate.

### No change needed (2)

| file:line | reviewer point | why the code is already right |
|---|---|---|
| `Given_Pivot_UITest.cs:104` | `MyCustomPivotItem` should be `private` (flagged 3 rounds) | **Already applied.** Commit `59504099b6` ("test: Make Pivot test data item private") already demoted the nested type to `private class MyCustomPivotItem`, keeping `Title`/`Content` `public` so reflection-based `{Binding}` (including the native-WinUI head) still resolves them. The file now carries an inline note documenting exactly that. Nothing left to change. |
| `Given_Path_UITest.cs:21` | `DataStyle` enum should be `private` (flagged 3 rounds) | **Bot suggestion is incorrect — it would not compile.** `DataStyle` is the parameter type of the `public [TestMethod] When_Path_FillRule(DataStyle style, FillRule fillRule)` and its `[DataRow]`s. MSTest requires the test method to stay `public`, so a `private`/`internal` enum triggers **CS0051 (inconsistent accessibility)**. The enum must remain `public`; the code already documents this with an inline comment on line 20. |

No threads required escalation. Both are marked handled for the parent to resolve.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Markup/Given_StaticResource_UITest.cs
```diff
@@ -0,0 +1,36 @@
...
+public class Given_StaticResource_UITest : SampleControlUITestBase
+{
+	// The sample self-initializes: its constructor registers the C# resource and the
+	// converter into Application.Current.Resources and assigns its own DataContext,
+	// so RunAsync (Activator.CreateInstance) yields a fully wired page.
+	[TestMethod]
+	[DataRow("XAMLResource_Text", "This resource was registered in XAML", DisplayName = "XAML-declared resource")]
+	[DataRow("CSharpResource_Text", "This resource was registered in C#", DisplayName = "C#-registered resource")]
+	[DataRow("ConverterResource_Text", "Hello Converter!", DisplayName = "Converter resource")]
+	public async Task When_StaticResource_Resolves(string elementName, string expected)
```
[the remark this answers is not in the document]
MartinZikmund: Good catch — fixed in `e78eeff792`. Confirmed `RuntimeTestsApp.RunAsync` throws `PlatformNotSupportedException` outside `__SKIA__`/`__WASM__`, so the guard was genuinely missing here. Added `[PlatformCondition(ConditionMode.Exclude, RuntimeTestPlatforms.Native)]` (placed above the `[DataRow]`s to match the attribute ordering used elsewhere in this batch). Also swept every remaining `await RunAsync` call site in the PR — this and `Given_XBind_UITest` were the only two missing the guard.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Data/Given_XBind_UITest.cs
```diff
@@ -0,0 +1,58 @@
...
+using static Private.Infrastructure.TestServices;
+
+namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Data;
+
+[TestClass]
+[RunsOnUIThread]
+public class Given_XBind_UITest : SampleControlUITestBase
+{
+	[TestMethod]
+	public async Task When_Functions()
```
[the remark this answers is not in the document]
MartinZikmund: Correct — fixed in `e78eeff792`. `RuntimeTestsApp.RunAsync` hits the `#else` branch (`throw new PlatformNotSupportedException()`) on the native heads, so `[PlatformCondition(ConditionMode.Exclude, RuntimeTestPlatforms.Native)]` was required here too.

src/Uno.UI.RuntimeTests/Tests/Windows_ApplicationModel_Resources/Given_ResourceLoader_UITest.cs
```diff
@@ -0,0 +1,81 @@
...
+	{
+		try
+		{
+			await RunAsync("UITests.Shared.Windows_ApplicationModel_Resources_ResourceLoader.ResourceLoader_Simple");
+
+			// The sample's TextBlocks carry only x:Uid (no x:Name), so App.Marked (Name-based) can't find
+			// them; walk the tree instead. Each label TextBlock is immediately followed by the TextBlock
+			// whose Text was resolved via its x:Uid.
+			var textBlocks = new List<TextBlock>();
+			CollectTextBlocks(TestServices.WindowHelper.WindowContent, textBlocks);
```
Xiaoy312, line 33: the textblock by index approach is quite fragile...
we should give the x:Name to the textboxes tb01...05 and use:
a. `FindFirstDescendant(name);` or
b. `EnumerateDescendant().OfType<TextBlock>().ToArray()` with `FirstOrDefault(x => MarkupHelper.GetXUid(x) is ...)`

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_AppXamlResources_UITest.cs
```diff
@@ -0,0 +1,59 @@
...
+
+				var themedBorder = (Border)root.FindName("ThemedBorder");
+				var staticBorder = (Border)root.FindName("StaticBorder");
+
+				Assert.AreEqual(Colors.Yellow, ((SolidColorBrush)themedBorder.Background).Color);
+				Assert.AreEqual(Colors.Purple, ((SolidColorBrush)staticBorder.Background).Color);
+			}
+			finally
+			{
+				WindowHelper.WindowContent = null;
```
Xiaoy312, line 55: isnt the clearing of WindowContent sort of implied now?

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_FocusManagerDirection_UITest.cs
```diff
@@ -0,0 +1,170 @@
...
+[TestClass]
+public class Given_FocusManagerDirection_UITest
+{
+	// 3x3 grid of TextBoxes, mirroring FocusManager_FocusDirection sample layout.
+	private static async Task<TextBox[]> SetupGrid()
+	{
+		var boxes = new TextBox[9];
+		var root = new StackPanel();
+
+		for (var row = 0; row < 3; row++)
```
Xiaoy312, line 20: add a comment to show the setup/layout of tboxes
// 012
// 345
// 678

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_VisualStateManager_UITest.cs
```diff
@@ -0,0 +1,89 @@
...
+	[PlatformCondition(ConditionMode.Exclude, UnsupportedPlatforms)]
+	public async Task When_Testing_ComplexSetters()
+	{
+		try
+		{
+			await RunAsync("UITests.Shared.Windows_UI_Xaml.VisualStateTests.VisualState_ComplexSetters_Automated");
+
+			// The sample wires its "changeState" button to VisualStateManager.GoToState(this, "State01", true).
+			// Drive that state change directly (no pointer injection) so the test is valid on Skia and WebAssembly.
+			DependencyObject node = App.Query("changeState").Single().Element;
```
Xiaoy312, line 34: lets not use this Xamarin remnant...
the test page(`node`) should be the WindowContent here
and we can use FindFirstDescendant for the borders below

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Automation/Given_AutomationId_UITest.cs
```diff
@@ -0,0 +1,79 @@
...
+	[TestMethod]
+	[PlatformCondition(ConditionMode.Exclude, RuntimeTestPlatforms.Native)] // RunAsync + pointer injection are only supported on Skia/WASM runtime hosts.
+	[DataRow("Item01", "Item 01", DisplayName = "Item01")]
+	[DataRow("Item02", "Item 02", DisplayName = "Item02")]
+	[DataRow("Item03", "Item 03", DisplayName = "Item03")]
+	public async Task When_Item_Tapped_Then_Result_Updated(string automationId, string expectedText)
+	{
+		try
+		{
+			await RunAsync(SampleName);
```
Xiaoy312, line 55: RunAsync by default takes a screenshot whenever it loads anything
are we should we want to loop(with DataRow) on that?
@MartinZikmund and, more specificially, do we even want to take a screeny for all these converted runtime tests?

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_CheckBox_UITest.cs
```diff
@@ -0,0 +1,138 @@
...
+// fire with the expected IsChecked value (as the CheckBox_Automated sample surfaced via a TextBlock).
+[TestClass]
+[RunsOnUIThread]
+public class Given_CheckBox_UITest
+{
+	[TestMethod]
+	[PlatformCondition(ConditionMode.Exclude, RuntimeTestPlatforms.NativeWinUI | RuntimeTestPlatforms.NativeAndroid | RuntimeTestPlatforms.NativeIOS | RuntimeTestPlatforms.NativeWasm)]
+	public async Task When_TwoState_Tapped_Toggles()
+	{
+		if (TestServices.WindowHelper.IsXamlIsland)
```
Xiaoy312, line 27: could be part of PlatformCondition: `existing_filter | RuntimeTestPlatforms.SkiaIslands`

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_CheckBox_UITest.cs
```diff
@@ -0,0 +1,138 @@
...
+			Assert.AreEqual(false, checkBox.IsChecked);
+			Assert.AreEqual("Unchecked twoState01 False", result);
+
+			await Tap();
+			Assert.AreEqual(true, checkBox.IsChecked);
+			Assert.AreEqual("Checked twoState01 True", result);
+		}
+		finally
+		{
+			WindowContent = null;
```
Xiaoy312, line 74: whether to keep the reset is another question (asked before)
but the try-finally-reset could be just: using _ = ResetWindowContent();
it would make the code a lot cleaner

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_ContentControl_UITest.cs
```diff
@@ -0,0 +1,97 @@
...
+	[PlatformCondition(ConditionMode.Exclude, RuntimeTestPlatforms.Native)] // RunAsync + pointer injection are only supported on Skia/WASM runtime hosts.
+	public async Task When_ContentPresenter_Template_Binds_And_Updates_On_Content_Change()
+	{
+		try
+		{
+			await RunAsync("Uno.UI.Samples.Content.UITests.ContentPresenter.ContentPresenter_Template");
+
+			// Content is null, so the template's {Binding} resolves to the inherited DataContext ("DataContext").
+			Assert.AreEqual("ContentPresenter:  DataContext", App.Marked("innerText").GetDependencyPropertyValue<string>("Text"));
+			Assert.AreEqual("ContentControl:  DataContext", App.Marked("innerText2").GetDependencyPropertyValue<string>("Text"));
```
Xiaoy312, line 26: more xamarin era vestige...: WindowContent>FindFirstDescendantOrThrow<TextBlock>().Text

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_Pivot_UITest.cs
```diff
@@ -0,0 +1,109 @@
...
+public class Given_Pivot_UITest
+{
+	// Migrated from SamplesApp.UITests UnoSamples_Pivot_Tests, backed by the
+	// Pivot_CustomContent_Automated sample: a Pivot populated with non-PivotItem
+	// objects, whose SelectedItem.Title/Content are surfaced through bindings.
+
+	[TestMethod]
+	public async Task When_Non_PivotItem_Items()
+	{
+		var sut = BuildSut();
```
Xiaoy312, line 23: there is some inconsistency with previous "setup" method
where `await UITestHelper.Load` is bundled within the setup method
pick a lane

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_Pivot_UITest.cs
```diff
@@ -0,0 +1,109 @@
...
+				new RowDefinition { Height = GridLength.Auto },
+				new RowDefinition { Height = new GridLength(1, GridUnitType.Star) },
+			},
+			Children = { header, pivot },
+		};
+
+		return (root, title, content, pivot);
+	}
+
+	private static DataTemplate CreateTemplate(string inner)
```
Xiaoy312, line 99: inline under BuildSut()

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_Popup_HVAlign_UITest.cs
```diff
@@ -0,0 +1,77 @@
...
+using Microsoft.UI.Xaml.Media;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Uno.UI.RuntimeTests.Helpers;
+using Windows.Foundation;
+using static Private.Infrastructure.TestServices;
+
+namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls;
+
+[TestClass]
+public class Given_Popup_HVAlign_UITest
```
Xiaoy312, line 14:
```suggestion
public class Given_Popup
```
we can probably do a finally pass to drop the _UITest suffix, once they are all migrated

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_Popup_LightDismiss_UITest.cs
```diff
@@ -0,0 +1,603 @@
...
+// an (open) popup are NOT reachable by descending from WindowHelper.WindowContent — the App.Marked/
+// App.Query shim used by RunAsync-based tests can't find them. Holding direct C# references to the
+// popups and their content sidesteps that entirely, and lets each scenario use small, deliberately
+// non-overlapping coordinates instead of depending on the original sample's StackPanel/font layout.
+[TestClass]
+[RunsOnUIThread]
+public class Given_Popup_LightDismiss_UITest
+{
+	private const RuntimeTestPlatforms InjectionUnsupported =
+		RuntimeTestPlatforms.NativeWinUI | RuntimeTestPlatforms.NativeAndroid | RuntimeTestPlatforms.NativeIOS | RuntimeTestPlatforms.NativeWasm;
```
Xiaoy312, line 35: IsXamlIsland seems like to be in all the tests?
```suggestion
	private const RuntimeTestPlatforms InjectionUnsupported =
		RuntimeTestPlatforms.NativeWinUI | RuntimeTestPlatforms.NativeAndroid | RuntimeTestPlatforms.NativeIOS | RuntimeTestPlatforms.NativeWasm | RuntimeTestPlatforms.SkiaIslands;
```

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_Popup_LightDismiss_UITest.cs
```diff
@@ -0,0 +1,603 @@
...
+			popup4.IsOpen = true;
+			popup3.IsOpen = true;
+			popup2.IsOpen = true;
+			popup1.IsOpen = true;
+			await TestServices.WindowHelper.WaitForIdle();
+
+			Assert.IsTrue(popup1.IsOpen);
+			Assert.IsTrue(popup2.IsOpen);
+			Assert.IsTrue(popup3.IsOpen);
+			Assert.IsTrue(popup4.IsOpen);
```
Xiaoy312, line 185: lets do a array compare here, and add message "initial state" "tapped once/twice/thrice/quatro"

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_Popup_LightDismiss_UITest.cs
```diff
@@ -0,0 +1,603 @@
...
+			Assert.IsFalse(popup2.IsOpen);
+			Assert.IsFalse(popup3.IsOpen);
+			Assert.IsFalse(popup4.IsOpen);
+		}
+		finally
+		{
+			popup1.IsOpen = false;
+			popup2.IsOpen = false;
+			popup3.IsOpen = false;
+			popup4.IsOpen = false;
```
Xiaoy312, line 216: using cleanupPopups = Disposable.Create(() => { ... });

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_RadioButton.Pointer.cs
```diff
@@ -0,0 +1,188 @@
...
+using Microsoft.UI.Xaml.Controls;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Private.Infrastructure;
+using Uno.UI.RuntimeTests.Helpers;
+using Uno.UI.Toolkit.DevTools.Input;
+using Windows.Foundation;
+using Windows.UI.Input.Preview.Injection;
+using static Private.Infrastructure.TestServices.WindowHelper;
+
+namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
```
Xiaoy312, line 13:
```suggestion
namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls;
```

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_SplitView_UITest.cs
```diff
@@ -0,0 +1,84 @@
...
+[TestClass]
+public class Given_SplitView_UITest
+{
+	// Mirrors UITests.Windows_UI_Xaml_Controls.SplitView.SplitViewClip: a right-placed, compact-overlay
+	// SplitView whose compact pane must occupy exactly CompactPaneLength, leaving the content area to its left unclipped.
+	[TestMethod]
+	[RunsOnUIThread]
+#if !HAS_RENDER_TARGET_BITMAP
+	[Ignore("Cannot take screenshot on this platform.")]
+#endif
```
Xiaoy312, line 19: could we create a helper for these? 
something like [RequiredCapability(RuntimeCapabilities.RenderTargetBitmap)]

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Input/Given_FocusInput_UITest.cs
```diff
@@ -0,0 +1,83 @@
...
+			Assert.AreEqual(FocusState.Unfocused, contentControl.FocusState);
+			Assert.AreEqual(FocusState.Pointer, textBox.FocusState);
+		}
+		finally
+		{
+			TestServices.WindowHelper.WindowContent = null;
+		}
+	}
+
+	private static Point GetCenter(FrameworkElement element) =>
```
Xiaoy312, line 78: thats like the 5th occurrence already
lets create a helper method for this

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Media/Given_Geometry_UITest.cs
```diff
@@ -0,0 +1,83 @@
...
+		{
+			Background = new SolidColorBrush(Colors.Yellow),
+			Child = path
+		};
+
+		try
+		{
+			await UITestHelper.Load(border);
+
+			var screenshot = await UITestHelper.ScreenShot(border);
```
Xiaoy312, line 41: not gated by `#if HAS_RENDER_TARGET_BITMAP [Ignore]`?

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Shapes/Given_ShapesClearData_UITest.cs
```diff
@@ -0,0 +1,109 @@
...
+			var center = new Point(border.ActualWidth / 2, border.ActualHeight / 2);
+			ImageAssert.HasColorAt(screenshot, center, Microsoft.UI.Colors.White);
+		}
+		finally
+		{
+			WindowHelper.WindowContent = null;
+		}
+	}
+
+	private static Border CreateHost(out Path path)
```
Xiaoy312, line 96:
```suggestion
	private static Border CreateHost(Size borderSize, Size pathSize, out Path path)
```
and pass the context-dependent pathSize in

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Media/Given_LinearGradientBrush_UITest.cs
```diff
@@ -0,0 +1,105 @@
...
+{
+	[TestMethod]
+	[RunsOnUIThread]
+#if !HAS_RENDER_TARGET_BITMAP
+	[Ignore("Cannot take screenshot on this platform.")]
+#endif
+	public async Task When_GradientStops_Changed()
+	{
+		// Colors mirror the LinearGradientBrush_Change_Stops sample's UnoGreen/UnoBlue/UnoPurple/UnoRed resources.
+		var brush = new LinearGradientBrush
```
Xiaoy312, line 23: var unoGreen = Color.FromArgb(0xFF, ...);
...
new GradientStop { Offset = 0.0, Color = unoGreen },
...

Xiaoy312 (approved):

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_CheckBox_UITest.cs
```diff
@@ -0,0 +1,138 @@
+#if HAS_INPUT_INJECTOR || WINAPPSDK
+#nullable enable
+
+using System.Threading.Tasks;
+using Microsoft.UI.Xaml.Controls;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Private.Infrastructure;
+using Uno.UI.RuntimeTests.Helpers;
+using Uno.UI.Extras.DevTools.Input;
```
morning4coffe-dev, line 9: [P1] Update the input-helper namespace in all seven new files. The current tree and exact merge base declare `InputInjectorExtensions` under `Uno.UI.DevTools.Input`, not `Uno.UI.Extras.DevTools.Input`. An explicitly restored exact-head SamplesApp build fails with seven CS0234 errors, and CI 231702's WinAppSDK build reports the same errors. Replacing only these seven imports in isolated diagnostic copies makes the build succeed and all 114 migrated Windows Skia cases pass. The same correction is needed in Control, NativeButtons, Popup_LightDismiss, RadioButton.Pointer, Enability and FocusInput.

morning4coffe-dev (changes requested):
The migration is currently blocked by a reproducible build error at `eb9e9d309634`, not merely an unclassified CI failure.

**P1: seven new files import the removed `Uno.UI.Extras.DevTools.Input` namespace.** After an explicit Release restore, the exact source-based net11.0-desktop SamplesApp build fails with seven CS0234 errors in CheckBox, Control, NativeButtons, Popup_LightDismiss, RadioButton.Pointer, Enability and FocusInput. `InputInjectorExtensions.cs` declares `Uno.UI.DevTools.Input` in both this head and its true merge base `d083fba2df58`. CI 231702 includes this head and reports the same seven errors in the native WinAppSDK build as well. Please update all seven imports rather than treating this as a feed or runner problem.

I isolated the fix without editing the PR source: review-only Compile overrides substitute copies differing solely in those seven using directives. That diagnostic app builds with zero warnings/errors and runs all **114 migrated cases successfully on actual Windows Skia**. This is useful confirmation of the immediate correction, but it is not a passing build of the submitted head; the seven corrections are not in the PR.

I also read the migration/retained-test changes and existing discussions. The public DataStyle enum is appropriate for a public test-method parameter, and the Pivot helper type is already private with public bindable properties; those previously settled points should not be reopened. Existing reviewers' conversations were left unchanged.

Validation scope: CI-pinned .NET 11 SDK, diagnostic analyzers disabled, actual source build and import-corrected Skia execution. No local native WinUI, mobile or browser execution, full-suite clearance, or complete equivalence proof for every retired platform-specific scenario is claimed. The submitted head still needs the namespace correction and normal CI gates.

src/Uno.UI.RuntimeTests/Tests/Windows_ApplicationModel_Resources/Given_ResourceLoader_UITest.cs
MartinZikmund: Fixed in 9cf21dde5bd: the sample's value TextBlocks now carry `x:Name` tb01..tb05, and the test reads each one with `WindowContent.FindFirstDescendantOrThrow<TextBlock>("tbNN")`, so tree order no longer matters.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_AppXamlResources_UITest.cs
MartinZikmund: Not implied, unfortunately: the engine's `GeneralCleanupAsync` only clears `WindowContent` when `IsUnloadingTestContent` is set, and the headless CI runs don't set it (same caveat noted in `Given_TextBox.skia.cs`). So the reset stays, but it's now a one-liner: `using var _ = UITestHelper.ResetWindowContent();` (9cf21dde5bd).

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_FocusManagerDirection_UITest.cs
MartinZikmund: Added the 0 1 2 / 3 4 5 / 6 7 8 index layout comment above `SetupGrid` (9cf21dde5bd).

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_VisualStateManager_UITest.cs
MartinZikmund: Done in 9cf21dde5bd: no more `App.Query` or parent walking. The page is `WindowContent.FindFirstDescendantOrThrow<UserControl>()` and the borders, button and log TextBlock come from `page.FindFirstDescendantOrThrow<T>(name)`.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Automation/Given_AutomationId_UITest.cs
MartinZikmund: No screenshot is taken here: the runtime-test `SampleControlUITestBase.RunAsync` defaults to `skipInitialScreenshot: true` and throws `NotSupportedException` if you pass false. So each DataRow only loads the sample, and none of the migrated tests take a screenshot through `RunAsync`. Only the explicit `UITestHelper.ScreenShot` pixel assertions do.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_CheckBox_UITest.cs
MartinZikmund: Done in 9cf21dde5bd for every migrated test: the `IsXamlIsland` early returns are gone and `RuntimeTestPlatforms.SkiaIslands` is now part of each `PlatformCondition` (for Popup_LightDismiss, part of the shared `InjectionUnsupported` const).

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_CheckBox_UITest.cs
MartinZikmund: Done in 9cf21dde5bd: added `UITestHelper.ResetWindowContent()`, which returns an `IDisposable`. All migrated tests now use `using var _ = UITestHelper.ResetWindowContent();` instead of try/finally. Tests that also close popups use `Disposable.Create` for the cleanup.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_ContentControl_UITest.cs
MartinZikmund: Done in 9cf21dde5bd: the text and border lookups now use `WindowContent.FindFirstDescendant(OrThrow)<T>(name)`, and waits go through `UITestHelper.WaitFor`. `App.Tap` is still used for the button taps.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_Pivot_UITest.cs
MartinZikmund: Picked the SetupGrid lane (9cf21dde5bd): Pivot's helper is now `async LoadSut()`, which builds the tree and awaits `UITestHelper.Load`, just like `Given_FocusManagerDirection_UITest.SetupGrid`.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_Pivot_UITest.cs
MartinZikmund: Done: `CreateTemplate` is now a static local function inside `LoadSut()` (9cf21dde5bd).

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_Popup_HVAlign_UITest.cs
MartinZikmund: I'm leaving the rename for the final pass, as you suggested, so this batch keeps one consistent `_UITest` suffix. One note for that pass: `Given_Popup` already exists under `Tests/Windows_UI_Xaml_Controls_Primitives`, so the popup suites should probably be folded into it rather than creating a second `Given_Popup` in `Windows_UI_Xaml_Controls`.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_Popup_LightDismiss_UITest.cs
MartinZikmund: Done in 9cf21dde5bd: `InjectionUnsupported` now includes `RuntimeTestPlatforms.SkiaIslands`, and the per-test `IsXamlIsland` returns are removed. The native Android/iOS/WASM flags are gone because master dropped them in the native-target removal.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_Popup_LightDismiss_UITest.cs
MartinZikmund: Done in 9cf21dde5bd for both multi-popup order tests: `AssertOpenStates("tapped once", popups, false, true, true, true)` uses `CollectionAssert.AreEqual` on the IsOpen array, with the step (initial state / all opened / tapped once/twice/thrice/four times) in the message.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_Popup_LightDismiss_UITest.cs
MartinZikmund: Done in 9cf21dde5bd: every popup test now uses `using var cleanup = Disposable.Create(() => { ...IsOpen = false; WindowContent = null; });` (Popup_HVAlign too).

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_RadioButton.Pointer.cs
MartinZikmund: Done, the file now uses a file-scoped namespace (9cf21dde5bd).

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_SplitView_UITest.cs
MartinZikmund: This one went away with the rebase: master dropped every `HAS_RENDER_TARGET_BITMAP` gate from the runtime tests (34731cdf339), because every remaining runtime-test head (Skia and WinUI) supports RenderTargetBitmap. So I removed the `#if !HAS_RENDER_TARGET_BITMAP [Ignore]` blocks here and in the SolidColorBrush/LinearGradientBrush tests instead of adding a capability attribute (17569078ca8).

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Input/Given_FocusInput_UITest.cs
MartinZikmund: Done in 9cf21dde5bd: added a `GetAbsoluteCenter()` extension next to `GetAbsoluteBounds()` in `UITestHelper.cs`. The per-file `GetCenter` helpers and the inline `TransformToVisual(...).TransformPoint(center)` copies in all seven pointer test files now use it.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Media/Given_Geometry_UITest.cs
MartinZikmund: The `NativeWasm` exclusion was acting as the RenderTargetBitmap gate. After the rebase, native WASM no longer exists, and every remaining runtime-test head supports RTB (master removed all `HAS_RENDER_TARGET_BITMAP` gates in 34731cdf339). So no gate is needed, and the exclusion is gone (17569078ca8).

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Shapes/Given_ShapesClearData_UITest.cs
MartinZikmund: Done in 9cf21dde5bd: `CreateHost(Size borderSize, Size pathSize, out Path path)`, and each test passes its own path size.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Media/Given_LinearGradientBrush_UITest.cs
MartinZikmund: Done: `unoGreen` / `unoBlue` / `unoPurple` / `unoRed` locals now feed the gradient stops (9cf21dde5bd).

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_CheckBox_UITest.cs
MartinZikmund: Fixed in 9047173c783: all seven files now import `Uno.UI.DevTools.Input`. The rebased branch builds (Uno.UI.RuntimeTests.Skia net10.0 and the SamplesApp net11.0-desktop head), and all 30 migrated test classes (127 cases) pass on Skia Desktop (Win32).

MartinZikmund:
@morning4coffe-dev Thanks for isolating it. The seven `Uno.UI.Extras.DevTools.Input` imports now point to `Uno.UI.DevTools.Input` (9047173c783). The branch is rebased on master, the native-only platform flags master removed are adapted (17569078ca8), and all 30 migrated classes (127 cases) pass on Skia Desktop.

MartinZikmund:

Rebased onto master (9cf21dde5bd), then worked through the feedback:

- @morning4coffe-dev (changes requested, namespace P1): fixed in 9047173c783
- @Xiaoy312: 20 review threads, replied inline and resolved. Code changes are in 9cf21dde5bd. The RTB-gate threads were settled by the rebase adaptation in 17569078ca8.

Issue #2080, opened 2019-11-13
[iOS] Applying a style on custom control make pointer events propagation erratic


baptiste-nv:
## Current behavior
2 instances of the same custom controls
- The first one is hidden
- The second one is on top of the 1st one

The 1st control has a custom style. the 2nd one uses the default style.

Both have a button at the same place.

Clicking on the button works as expected on UWP and Android (execute the event handler for 2nd control only).
On iOS, when I click on the visible button (2nd control), it executes the event handler on the 1st control, THEN on the 2nd one.

If I use TranslateTransform on the button of the 1st control to put it out of the screen, the handler is still executed.

If I remove the custom style on the 1st control, it works as expected.

## Expected behavior
On iOS, when I click on the visible button (2nd control), it executes the event handler for 2nd control only.

## How to reproduce it (as minimally and precisely as possible)
See internal project 
https://feedback.nventive.com/communities/1/topics/3675-

## Environment

Package Version(s):
Uno.UI 1.46.196-dev.2440
Uno.UI 2.0.512-dev.3874

Affected platform(s):

- [X] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] WebAssembly renderers for Xamarin.Forms
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

Visual Studio:

- [X] 2017 (version: )
- [X] 2019 (version: )
- [ ] for Mac (version: )

Relevant plugins:

- [ ] Resharper (version: )

dr1rrb:
Tested on iOS, Android, Skia and WASM with 

```xaml
	<Grid>
	  <TextBlock Text="Hello, world!" Margin="20" FontSize="30" />

		<Button Width="300" Height="300" x:Name="Container" Background="Red" Click="OnButtonClicked">
			<Button Width="200" Height="200" x:Name="Nested" Background="Orange" Click="OnButtonClicked"/>
		</Button>

		<Button Width="100" Height="100" x:Name="Above" Background="Purple" Click="OnButtonClicked"/>

	</Grid>
```

Works as expected by now.

PR #23654, opened 2026-07-06
refactor!: Align parameter names with WinUI (sync-gen)

MartinZikmund:
**GitHub Issue:** https://github.com/unoplatform/uno-private/issues/2080 (part of #8339)

## PR Type:

## What changed? 🚀

The WinAppSDK sync generator (`Uno.WinAppSDKSyncGenerator`) carried a workaround,
`SymbolMatchingHelpers.IgnoreParameterName`, that treated method/constructor **parameter-name**
mismatches between the WinAppSDK/CsWinRT metadata and Uno's hand-written members as "matching", to
avoid a breaking change. For 7.0 (a major) we remove that workaround so the generated public surface
mirrors the WinAppSDK metadata.

Concretely this:

1. Removes `IgnoreParameterName` and its use in `AreParametersMatching` (matching now requires an
   exact parameter-name match). `IgnoreRefKind` is intentionally left in place (separate sweep item).
2. Renames the hand-written Uno parameters to the WinAppSDK metadata names for exactly the members the
   generator would otherwise re-emit as duplicate `[NotImplemented]` stubs:
   - `Windows.UI.Text.FontWeight(ushort weight)` → `_Weight`
   - `Windows.Foundation.Rect(Point point, Size size)` → `location`
   - `Microsoft.UI.Xaml.Duration.Compare(Duration first, Duration second)` → `t1`, `t2`
   - `System.IO.WindowsRuntimeStreamExtensions` (6 overloads): `windowsRuntimeStream` → `windowsruntimeStream`
   - `Windows.Graphics.PointInt32 / RectInt32 / SizeInt32` internal ctors: `x/y/width/height` → `_X/_Y/_Width/_Height`
3. Adds `PackageDiffIgnore.xml` entries for the intentionally-renamed **public** members (the
   graphics-struct ctors are internal and not in the public diff).

The underscore-prefixed names (`_Weight`, `_X`, …) and the lowercase `windowsruntimeStream` are the
actual WinAppSDK/CsWinRT metadata names — surprising, but 1:1 parity with WinUI is exactly what the sync
generator is for. Short comments were added on those members so a future `dotnet format --severity info`
or a well-meaning "fix naming" pass doesn't silently revert them (which would break the generator match).

### Scope: why some documented cases are not renamed

The removed workaround also documented `Equals`-family divergences (`CornerRadius`/`Thickness`/`Point`/
`Duration.Equals`, e.g. `other`/`obj`/`first,second`). The generator **skips `Equals` and `GetHashCode`
by name** (`Generator.cs`), so it never emits those members — removing the workaround has no effect on
them and regeneration is byte-identical. They are intentionally left unchanged to keep this scoped to the
generator's enforced surface. (Note: because the sync generator only enforces param names on the members
it processes, a handful of Uno-only / skipped-type members remain unverified against WinUI — a pre-existing
gap, not introduced here.)

## Validation

- **Regeneration** — Ran the sync generator. Before the renames it emitted new stubs into exactly 7
  `Generated/` files; after the renames the regenerated `Generated/` tree is **byte-identical to
  baseline** (empty `git diff` on `*Generated*`). This proves the members now match the WinAppSDK names
  with no new stubs. The reference set is pinned (`Microsoft.WindowsAppSDK` 2.1.3), so the result
  transfers to CI.
- **Compile** — `Uno.UI.Skia` plus the Reference and WASM variants of `Uno.UWP`/`Uno.Foundation` build
  clean (0/0). A CI-equivalent analyzer + `EnforceCodeStyleInBuild` + `TreatWarningsAsErrors` build of
  `Uno.UWP` (which carries every underscore param) is also clean, confirming the naming-suggestion rule
  does not break CI. The native Android/iOS heads were not compiled locally, but the edits are
  platform-agnostic (no `#if`) and a repo-wide grep confirms no named-argument callers anywhere.

> Note: the `PackageDiffIgnore.xml` signatures were authored to match the tool's format; if the CI
> PackageDiff step reports a slightly different rendering, the exact expected string in its output can be
> pasted in.

## PR Checklist ✅

- [x] 🧪 Added test / sample (N/A — no behavior change; validated by generator regeneration + builds)
- [x] 📚 Docs have been added/updated (N/A)
- [x] 🖼️ Validated PR `Screenshots Compare Test Run` results (N/A — no visual change)
- [x] 👀 Reviewed 2 other open pull requests (optional)

### Breaking change impact

Public parameter names change on `FontWeight(ushort)`, `Rect(Point, Size)`, `Duration.Compare`, and the
`WindowsRuntimeStreamExtensions.AsStream*` overloads. This affects only callers that pass these arguments
**by name** (named arguments); it is a source-level change (parameter names are not part of the binary
signature) and is intended for 7.0 to align with the WinAppSDK/CsWinRT metadata.


Xiaoy312 (approved):

ajpinedam (approved):

MartinZikmund:
## Review comments addressed

Pushed commit `85bc05d0fb768a4eb3049dab69d9a3e2c215aa9e`. All changes were **reviewed by inspection only** (no local build/test run); Azure Pipelines CI will do the authoritative verification.

### Fixed (1)

|---|---|---|
| `build/PackageDiffIgnore.xml:2961` | The `#2080` comment + nine `reason` attributes link a private-repo issue, which AGENTS.md forbids in a repo-tracked public file. | Replaced all 10 `#2080` occurrences with the public breaking-changes epic **`#8339`** (confirmed OPEN, "[Epic] Breaking changes roundup", already the umbrella this PR rolls under). |

### No change needed (2)

| file:line | reviewer point | why no change |
|---|---|---|
| `build/PackageDiffIgnore.xml:2973` (copilot) | The `fullName` signatures use pre-rename param names (`weight`, `point`, `first/second`, `windowsRuntimeStream`); add post-rename entries too or the ignores won't match. | **Suggestion is based on a misreading of how Uno.PackageDiff works.** The tool (`AssemblyComparer.CompareTypes(base, target)`) reports *removed* members from the **base = latest published NuGet** assembly, matching the ignore `fullName` against that base member's `ToSignature()`, which renders the **pre-rename** parameter names. Param renames are detected because matching keys on `p.Name + p.ParameterType.FullName`, but the reported/diffed side is always the base assembly — the post-rename names (`_Weight`, `location`, `t1/t2`, `windowsruntimeStream`) never appear in the report. Post-rename entries would be dead, never-matched noise. The current pre-rename entries are exactly what the tool matches. |
| `src/Uno.UWP/UI/Text/FontWeight.cs:10` (ultrareview) | CA1707 ("no underscores") could fire on the `_Weight` public ctor param in a CI-equivalent build. | **CA1707 does not fire here.** It is opt-in — *"Enabled by default in .NET 10: No"* (MS docs) — and `src/Directory.Build.props` elevates only the **Performance** and **Globalization** categories to `AllEnabledByDefault`, not **Naming** (CA1707's category). No `.editorconfig` in the repo enables CA1707 or the Naming category (grepped: 0 hits). The name is also mandated by the sync generator (it mirrors the WinAppSDK/CsWinRT metadata), so it must stay `_Weight`. Adding a speculative `#pragma warning disable CA1707` would be an *unnecessary* suppression, so it was deliberately not added. The bot itself framed this as a conditional heads-up ("no action needed if confirmed clean"); CI's analyzer-strict build is the authority.|

None.

---

⚠️ This PR still **conflicts with its base** (`feature/breakingchanges`) and needs a **manual rebase** — that was intentionally left untouched here.

PR #23655, opened 2026-07-06
refactor!: Align divergent property types with WinUI (sync-gen)

MartinZikmund:
**GitHub Issue:** https://github.com/unoplatform/uno-private/issues/2086 (part of #8339)

## PR Type:

💬 Other: API alignment for 7.0 (source/binary-breaking, no behavior change)

## What changed? 🚀

The WinAppSDK sync generator (`Uno.WinAppSDKSyncGenerator`) carried three by-name exceptions in its property-type matcher (`SymbolMatchingHelpers.ArePropertiesMatching`) that treated a **property-type** mismatch between the WinAppSDK metadata and Uno's hand-written members as "matching", to avoid a breaking change. For 7.0 (a major) we remove those exceptions so the generated public surface mirrors the WinAppSDK metadata, and align Uno's declared types to match.

The three cases (`Content`, `EasingFunction`, `ElementName`):

1. **`Content` — the exception was stale.** Every WinAppSDK type that declares a `Content` property already has a matching Uno type (`UIElement`↔`UIElement`, `object`↔`object`). Removing the `Content` clause changes nothing in the generated output; no Uno type change was needed.

2. **`DoubleAnimation.EasingFunction`: `IEasingFunction` → `EasingFunctionBase`.** WinAppSDK metadata declares this property (and the sibling `ColorAnimation`/`PointAnimation`/`Easing*KeyFrame` ones) as `EasingFunctionBase`; the `IEasingFunction` typing was a WPF holdover. `EasingFunctionBase : IEasingFunction`, so the internal `IAnimation<T>` contract and the animator pipeline are unaffected. `DoubleAnimation` was the only type the exception actually masked — `ColorAnimation`/`PointAnimation` have no hand-written public `EasingFunction` and already emitted the correct `EasingFunctionBase` stub.
   - `SwipeControl` assigned the internal, non-`EasingFunctionBase` `LinearEase.Instance` to the public property. It now assigns `null`, which the animator interpolates linearly (`_easing = easingFunction ?? LinearEase.Instance`) — identical behavior, and closer to WinUI (linear = no easing function).

3. **`Binding.ElementName`: `object` → `string`.** The WinUI IDL declares `String ElementName`. Uno used `object` to carry either the element-name string **or** an `ElementNameSubject` used for late binding. The subject is relocated to a new Uno-only `[EditorBrowsable(Never)]` **`Binding.ElementNameSubject`** property — structurally identical to the existing public Uno-only `CompiledSource` seam on the same class. It must be public because the compiled-XAML output lives in the consuming assembly (no access to Uno.UI internals). The XAML code generator, the runtime XAML reader, `BindingExpression`, and the cross-store binding copy all route the subject through the new property. `Binding` is the only WinAppSDK type with an `ElementName` property, so no other declaration is affected.

The by-name exceptions for these three are removed; only the genuine `ContentTemplateRoot` divergence (native-view typed) is kept.

## Validation

- **Regeneration (authoritative)** — Ran the sync generator against the pinned reference set (`Microsoft.WindowsAppSDK` 2.1.3). After the type alignments and with all three exceptions removed, the regenerated `Generated/` tree is **byte-identical to baseline** (empty `git diff` on `*Generated*`, exit 0). This proves the members now match the WinAppSDK metadata with no new `[NotImplemented]` stubs — the result transfers to CI.
- **Compile** — `Uno.UI.Skia` (net10.0) and `Uno.UI.SourceGenerators` build clean (0 errors). The `SwipeControl` consumer break was caught here and fixed.
- **Source-generator tests** — `Uno.UI.SourceGenerators.Tests` `Given_Binding` + `Given_Parser` pass; the 4 XAML goldens with `{Binding ElementName=…}` now emit `ElementNameSubject = _…Subject` (updated in this PR).
- **Runtime** — Ran the two applicable ElementName binding tests on Skia Desktop — `Given_Control.When_Refresh_Setter_BindingOnInvocation_ElementName` and `Given_ListViewBase.When_Outer_ElementName_Binding` — both **pass** (2/2), confirming the `ElementNameSubject` relocation resolves `{Binding ElementName=…}` end-to-end. `Given_DoubleAnimation.When_EasingFunction` is `#if __ANDROID__`-gated (native-Android-only assertion) and does not run on Skia; the `EasingFunctionBase` narrowing is behavior-identical (a null easing is still interpolated linearly) and compile-validated.

> Note: the `PackageDiffIgnore.xml` signatures were authored to match the tool's format; if the CI PackageDiff step reports a slightly different rendering, the exact expected string in its output can be pasted in.

## PR Checklist ✅

- [x] 🧪 Added test / sample (updated source-generator goldens; behavior-preserving so no new runtime test)
- [x] 📚 Docs have been added/updated (N/A)
- [x] 🖼️ Validated PR `Screenshots Compare Test Run` results (N/A — no visual change)
- [x] 👀 Reviewed 2 other open pull requests (optional)

### Breaking change impact

- `Binding.ElementName` changes from `object` to `string` (source + binary break). Code-behind assigning a subject to `ElementName` no longer compiles (undocumented Uno-internal pattern); assigning a string still compiles and matches WinUI. Precompiled third-party XAML libraries using `{Binding ElementName=…}` generated against an older Uno.UI must be recompiled — appropriate within the 7.0 breaking-changes window.
- `DoubleAnimation.EasingFunction` changes from `IEasingFunction` to `EasingFunctionBase` (source + binary break). Callers assigning a custom `IEasingFunction` that does not derive from `EasingFunctionBase` no longer compile — this matches WinUI, where `EasingFunctionBase` is the only valid public value.
- A new public `Binding.ElementNameSubject` property is added (`[EditorBrowsable(Never)]`, Uno-only); it is ignored by the sync generator (WinUI has no such member).

### DIFF
--- a/build/PackageDiffIgnore.xml
+++ b/build/PackageDiffIgnore.xml
@@ -2492,6 +2492,9 @@
 		  <Member fullName="System.Boolean Uno.UI.FeatureConfiguration/Control::UseDeferredOnApplyTemplate()" reason="Removed Control.UseDeferredOnApplyTemplate legacy flag; OnApplyTemplate deferral is now a compile-time platform decision (#2053, breaking changes for 7.0)" />
 		  <Member fullName="System.Boolean Uno.UI.FeatureConfiguration/Popup::EnableLightDismissByDefault()" reason="Removed Popup.EnableLightDismissByDefault legacy flag; Popup.IsLightDismissEnabled always defaults to false (WinUI-aligned) (#2053, breaking changes for 7.0)" />
 		  <Member fullName="System.Boolean Uno.UI.FeatureConfiguration/TextBox::UseOverlayOnSkia()" reason="Removed TextBox.UseOverlayOnSkia legacy flag; the Skia-based TextBox is always used (#2053, breaking changes for 7.0)" />
+		  <!-- Property types aligned to the WinUI metadata (#2086, breaking changes for 7.0). -->
+		  <Member fullName="System.Object Microsoft.UI.Xaml.Data.Binding::ElementName()" reason="WinUI sync: Binding.ElementName narrowed from object to string to match WinUI metadata (#2086, breaking changes for 7.0)" />
+		  <Member fullName="Microsoft.UI.Xaml.Media.Animation.IEasingFunction Microsoft.UI.Xaml.Media.Animation.DoubleAnimation::EasingFunction()" reason="WinUI sync: DoubleAnimation.EasingFunction changed from IEasingFunction to EasingFunctionBase to match WinUI metadata (#2086, breaking changes for 7.0)" />
 		  <!-- Residual dead native-renderer FeatureConfiguration flags removed (#2125/#2052, breaking changes for 7.0). -->
 		  <Member fullName="System.Boolean Uno.UI.FeatureConfiguration/UIElement::AssignDOMXamlName()" reason="Removed UIElement.AssignDOMXamlName flag; the WASM DOM annotation it gated no longer exists (#2125/#2052, breaking changes for 7.0)" />
 		  <Member fullName="System.Boolean Uno.UI.FeatureConfiguration/UIElement::FailOnNSObjectExtensionsValidateDispose()" reason="Removed UIElement.FailOnNSObjectExtensionsValidateDispose flag; NSObjectExtensions.ValidateDispose it gated is gone (BC47) (#2125, breaking changes for 7.0)" />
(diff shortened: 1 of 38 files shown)


Xiaoy312 (approved):

MartinZikmund:
Addressed the open review threads on this `Binding.ElementName` (`object` → `string`) type-alignment PR. Pushed commit **`8b0a35ea8e`**. All changes were **reviewed by inspection only — CI will verify** (no local build/test was run).

### Fixed (2)

|---|---|---|
| `src/Uno.UI/UI/Xaml/Markup/Reader/XamlObjectBuilder.cs`:1207 | Runtime reader sets `ElementNameSubject` + calls `AddElementName` but never writes `binding.ElementName = value`, so `binding.ElementName` reads back `null` for `{Binding ElementName=Foo}` — breaking the WinUI string-roundtrip contract. | Added `binding.ElementName = value;` inside the `if` block so the public string roundtrips. Resolution still flows through the subject; the string is purely diagnostic/tooling state. |

The two github-actions threads at this location (original nit + the follow-up-pass suggestion block) are the same point and are both covered by this one change.

### No change needed (2)

| file:line | reviewer point | why the code is already right / the suggestion is off |
|---|---|---|
| `src/Uno.UI/UI/Xaml/Data/Binding.cs`:100 (copilot) | Make the `ElementName` setter build/update `ElementNameSubject` so `new Binding { ElementName = "Foo" }` from code-behind resolves the element. | **Not a regression.** Pre-PR, `ApplyElementName` required `ElementName is ElementNameSubject`; a code-behind *string* never matched, so programmatic element-name bindings never resolved in Uno either — only the XAML reader/generator (which build a namescope-registered subject) ever worked. Making code-behind resolve is a **feature enhancement out of scope** for a type-alignment refactor, and it would **conflict** with the reader path: the reader now also writes `ElementName = value`, which would clobber a setter-built subject. That interaction needs design + runtime validation, so it's deliberately not done here. Noted as a possible follow-up below. |
| `src/Uno.UI/UI/Xaml/DependencyPropertyDetailsCollection.cs`:79 (copilot) | Guard `newBinding.ElementNameSubject = binding.ElementNameSubject` with a non-null check so it doesn't overwrite a subject created by copying `ElementName`. | Premise doesn't hold: `ElementName` is a plain auto-property — assigning the string does **not** build a subject, so there's nothing for line 78 to overwrite. `newBinding` is freshly constructed (line 67), so its `ElementNameSubject` is already `null`; copying `null`-over-`null` is a no-op. The unconditional copy faithfully mirrors the source binding. The guard would be inert. |

### Follow-up (noted, not blocking)
- Code-behind `new Binding { ElementName = "Foo" }` resolving against the target's namescope is a **pre-existing** Uno limitation (not introduced here). If desired, worth a separate issue rather than growing this refactor.
- The **compiled-XAML** path (`XamlFileGenerator`) emits only `ElementNameSubject = _subject`, so `binding.ElementName` also doesn't roundtrip there. Left untouched — touching the generator regenerates golden files and needs a build to trust; the flagged threads only concern the runtime reader.

*This single comment replaces per-thread replies. Thread resolution is handled centrally by the parent process.*

morning4coffe-dev (commented):
The public types align with WinUI: Binding.ElementName is String and the animation easing property is EasingFunctionBase (see [the IDL](https://github.com/microsoft/microsoft-ui-xaml/blob/d0c6705ec46dbddb853cdfdb2eb3b42849cf304e/dxaml/xcp/dxaml/idl/winrt/core/microsoft.ui.xaml.coretypes.idl#L5250) and [easing signature](https://github.com/microsoft/microsoft-ui-xaml/blob/d0c6705ec46dbddb853cdfdb2eb3b42849cf304e/dxaml/xcp/dxaml/idl/winrt/core/microsoft.ui.xaml.coretypes.idl#L3294)). The generator/clone/runtime-reader subject plumbing has also been updated.

Remaining behavioral difference: code-created `new Binding { ElementName = "source" }` still does not perform namescope resolution; the patch warns and avoids DataContext fallback. WinUI's [BindingExpression::GetSourceElement](https://github.com/microsoft/microsoft-ui-xaml/blob/d0c6705ec46dbddb853cdfdb2eb3b42849cf304e/dxaml/xcp/dxaml/lib/BindingExpression_Partial.cpp#L1030-L1076) actually calls FindNameInPage. That is a pre-existing/incomplete capability, not evidence that the type correction is wrong, but this PR should not be described as full ElementName behavior parity.

morning4coffe-dev (approved):
Re-reviewed the resolved feedback and the stated API-alignment scope. Binding.ElementName is now String and DoubleAnimation.EasingFunction is EasingFunctionBase, matching the WinUI IDL. The runtime reader preserves the public ElementName string, the generated/runtime subject bridge and hot-reload copy are wired, and SwipeControl retains linear interpolation through a null easing function. With the threads resolved and another reviewer approval present, I am approving this scoped type correction.

The separate limitation noted in my earlier review remains: programmatic ElementName bindings still lack WinUI's namescope lookup. That pre-existing capability gap is not a blocker for correcting these public property types, and this approval should not be read as claiming complete ElementName behavior parity.

MartinZikmund:
### Addressed PR comments up to this point

Rebased onto current `master` (clean, no conflicts) - new tip `2d8f8ce7920`.

- @morning4coffe-dev (review, Sep 8 + approval, Sep 10): namescope lookup for code-created `new Binding { ElementName = "..." }` is still missing. Agreed, that is a pre-existing gap and out of scope here - this PR only fixes the public property types and does not claim full `ElementName` behavioral parity. No code change.

PR #23659, opened 2026-07-06
fix(sdk): align Linux SkiaSharp native asset on desktop

### LINKED ISSUE #23658: SkiaSharpVersion override does not retarget SkiaSharp.NativeAssets.Linux on Skia Desktop (managed/native mismatch → startup crash)
### Current behavior 🐛

On a Skia Desktop single-project head (`net*-desktop`), overriding `<SkiaSharpVersion>` to a version above the one the Uno.Sdk ships does **not** retarget `SkiaSharp.NativeAssets.Linux`. The managed `SkiaSharp` assembly and the Win32/macOS native assets move to the requested version, but the **Linux** native `libSkiaSharp.so` stays at the lower version that the `Uno.WinUI.Runtime.Skia.X11` / `Uno.WinUI.Runtime.Skia.Linux.FrameBuffer` runtime packs reference as a floor.

Because managed and native SkiaSharp must match, the app **crashes at startup on Linux** (native/managed ABI mismatch). There is **no build or restore warning** — the failure is silent until runtime, and only on Linux (Windows and macOS self-align).

The docs (*Using the Skia Desktop* → "Upgrading to a later version of SkiaSharp") state that setting `SkiaSharpVersion` retargets "the SkiaSharp packages the Uno.Sdk manages, including the platform native-asset packages … in one place" — but this is not true for the Linux native asset.

#### Root cause

- The base `SkiaSharp` NuGet package auto-depends on `SkiaSharp.NativeAssets.Win32` and `SkiaSharp.NativeAssets.macOS` (in every desktop/generic TFM group, at its own managed version), but has **no dependency on `SkiaSharp.NativeAssets.Linux`** in any group. So bumping managed `SkiaSharp` self-aligns Win32/macOS native, but never Linux — Linux native assets are opt-in by SkiaSharp's design.
- The Uno.Sdk injects `SkiaSharp.NativeAssets.Linux` as an implicit top-level reference only for the legacy `*.Skia.Linux.FrameBuffer` multi-head, **not** for the modern single-project desktop head. For that head the only reference to it is the transitive `>= <floor>` dependency from the X11 / FrameBuffer runtime packs, which NuGet happily satisfies with the floor version (no conflict → no diagnostic).
- `SkiaSharpVersion` retargets implicit references, but since `SkiaSharp.NativeAssets.Linux` is never injected for the desktop head, the property (and the documented `PackageReference Update` fallback) has nothing to bind to — the transitive floor wins.

### Expected behavior 🎯

Setting `<SkiaSharpVersion>` on a Skia Desktop head retargets **all** managed and native SkiaSharp packages, including `SkiaSharp.NativeAssets.Linux`, so managed and native stay aligned on every desktop RID. Moving to SkiaSharp 4 (`4.148.0`) should work from the single `SkiaSharpVersion` property, as documented.

Minimal, SkiaSharp-only repro (no full app required) that shows the NuGet-resolution outcome — `repro.csproj`:

```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <OutputType>Library</OutputType>
    <TargetFramework>net9.0</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Uno.WinUI.Runtime.Skia.X11" Version="6.5.237" />
    <PackageReference Include="SkiaSharp" Version="4.148.0" />
  </ItemGroup>
</Project>
```

```
dotnet restore
dotnet list package --include-transitive
```

Resolved versions:

| Package | Resolved |
|---|---|
| SkiaSharp | 4.148.0 |
| SkiaSharp.NativeAssets.Win32 | 4.148.0 |
| SkiaSharp.NativeAssets.macOS | 4.148.0 |
| **SkiaSharp.NativeAssets.Linux** | **3.119.0** ← stuck at the X11 pack floor |

Restore completes with **no `NU1605`/`NU1608` warning or error**. In a real Skia Desktop app this ships a `3.119.0` `libSkiaSharp.so` alongside a `4.148.0` managed `SkiaSharp`, crashing on Linux at startup.

Full-app path: `dotnet new unoapp`, set `<SkiaSharpVersion>4.148.0</SkiaSharpVersion>`, run on Linux/X11 → startup crash; Windows and macOS are unaffected.

### Workaround 🛠️

Explicitly reference the Linux native asset at the same version as managed SkiaSharp so it gets a top-level constraint:

```xml
<PropertyGroup>
  <SkiaSharpVersion>4.148.0</SkiaSharpVersion>
</PropertyGroup>
<ItemGroup>
  <PackageReference Include="SkiaSharp.NativeAssets.Linux" Version="$(SkiaSharpVersion)" />
</ItemGroup>
```

Under Central Package Management (without transitive pinning) add a direct `PackageReference Include` in the head plus a matching `PackageVersion` — a `PackageVersion` / `Update` alone is inert against a transitive-only reference.

### Renderer 🎨

Skia

### Affected platforms 📱💻🖥️

Desktop (X11), Desktop (Linux Framebuffer)

Reproduced with public stable `Uno.WinUI.Runtime.Skia.X11 6.5.237`; the gap is present on `master`, 6.6 and 6.5.

Related latent gap (not yet triggered): `HarfBuzzSharp.NativeAssets.Linux` is referenced by the same runtime packs via the same floor mechanism and would show the identical issue if a consumer overrides the HarfBuzzSharp version above Uno's floor. The WebAssembly native asset is structurally similar (managed `SkiaSharp` has no wasm dependency), though it links statically at publish so it may surface as a link/AOT failure rather than a startup crash. A focused fix for the Linux SkiaSharp native asset is proposed in the linked PR; the HarfBuzz/WASM siblings can be tracked as follow-ups.

MartinZikmund:
**GitHub Issue:** closes #23658

## PR Type:

🐞 Bugfix

## What changed? 🚀

On a Skia Desktop single-project head (`net*-desktop`), overriding `<SkiaSharpVersion>` above the version the Uno.Sdk ships did **not** retarget `SkiaSharp.NativeAssets.Linux`. Managed `SkiaSharp` and the Win32/macOS native assets moved to the requested version, but the Linux native `libSkiaSharp.so` stayed at the transitive floor referenced by the `Uno.WinUI.Runtime.Skia.X11` / `Uno.WinUI.Runtime.Skia.Linux.FrameBuffer` runtime packs — a **silent** managed/native ABI mismatch that crashed Skia Desktop on Linux at startup (no build or restore diagnostic).

**Root cause:** the base `SkiaSharp` package auto-depends on `SkiaSharp.NativeAssets.Win32` and `.macOS` (in every desktop/generic TFM group, at its own version), so those self-align on a managed bump — but it has **no `SkiaSharp.NativeAssets.Linux` dependency** in any group (Linux native is opt-in). Meanwhile the Uno.Sdk only injected the Linux native asset for the legacy `*.Skia.Linux.FrameBuffer` head, never for the modern single-project desktop head. With no top-level reference to bind to, `SkiaSharpVersion` (and the documented `PackageReference Update` fallback) had nothing to retarget, so the transitive floor won.

**Fix:** inject `SkiaSharp.NativeAssets.Linux` as an implicit package for executable desktop heads — in the same `$(_IsExecutable)` item group that already pulls the X11/FrameBuffer runtime packs. Because the package is already part of the Uno.Sdk `SkiaSharp` version group, `SkiaSharpVersion` now retargets it *in one place*, making the *Using the Skia Desktop* docs' promise actually hold, and mirroring both the legacy FrameBuffer head and SkiaSharp's own Win32/macOS behavior. This also aligns the default (non-overridden) case, where the Linux native previously resolved one patch below managed.

**Validation:**

- **NuGet-resolution repro (root cause + fix mechanism):** a minimal `Uno.WinUI.Runtime.Skia.X11 6.5.237` + `SkiaSharp 4.148.0` project resolves managed + `NativeAssets.Win32` + `NativeAssets.macOS` to `4.148.0` while `NativeAssets.Linux` stays at `3.119.0`, with **no** restore warning/error. Adding a top-level `SkiaSharp.NativeAssets.Linux 4.148.0` reference — exactly what this change injects — realigns it to `4.148.0`.
- **Code review:** the injection mirrors the existing `Uno.Implicit.Packages.ProjectSystem.Legacy.targets` pattern, and `ImplicitPackagesResolver` already routes `SkiaSharpVersion` into the `SkiaSharp` group so the injected reference is emitted at the overridden version.
- The full Uno.Sdk-pack + consumer-restore end-to-end was not run in this environment (heavy Sdk packaging pipeline); happy to run it on request.

> Related latent gaps tracked for follow-up (out of scope here to keep the backport minimal): `HarfBuzzSharp.NativeAssets.Linux` shares the same floor mechanism, and the WebAssembly native asset is structurally similar (managed `SkiaSharp` has no wasm dependency).

## PR Checklist ✅

- [ ] 🧪 Added Runtime/UI tests or a manual sample — build-system (MSBuild targets) change; there is no runtime-test harness for implicit-package resolution, so it is validated via the NuGet-resolution repro above.
- [ ] 📚 Docs updated — the existing *Using the Skia Desktop* → "Upgrading to a later version of SkiaSharp" guidance becomes accurate with this fix; no doc change required.

### DIFF
--- a/src/Uno.Sdk/targets/Uno.Implicit.Packages.ProjectSystem.Desktop.targets
+++ b/src/Uno.Sdk/targets/Uno.Implicit.Packages.ProjectSystem.Desktop.targets
@@ -30,5 +30,13 @@
 		<_UnoProjectSystemPackageReference Include="Uno.WinUI.Runtime.Skia.MacOS" ProjectSystem="true" />
 		<_UnoProjectSystemPackageReference Include="Uno.WinUI.Runtime.Skia.Win32" ProjectSystem="true" />
 		<_UnoProjectSystemPackageReference Include="Uno.WinUI.Runtime.Skia.X11" ProjectSystem="true" />
+
+		<!--
+			The managed SkiaSharp package auto-references its Win32/macOS native assets but not the Linux one,
+			which only arrives as a transitive floor from the X11 / FrameBuffer runtime packs above. Inject it
+			so overriding SkiaSharpVersion retargets the Linux native asset too; otherwise a managed SkiaSharp
+			bumped past Uno's floor keeps the old libSkiaSharp on Linux and crashes Skia Desktop at startup.
+		-->
+		<_UnoProjectSystemPackageReference Include="SkiaSharp.NativeAssets.Linux" ProjectSystem="true" />
 	</ItemGroup>
 </Project>
--- a/build/test-scripts/run-net7-template-linux.ps1
+++ b/build/test-scripts/run-net7-template-linux.ps1
@@ -131,3 +131,25 @@ for($i = 0; $i -lt $projects.Length; $i++)
         dotnet clean $release "$projectPath"
     }
 }
+
+# Regression guard for issue #23658: overriding SkiaSharpVersion on a desktop head must retarget
+# SkiaSharp.NativeAssets.Linux too. The base SkiaSharp package pulls its Win32/macOS native assets
+# but not the Linux one, so without the Uno.Sdk injecting it the Linux native stays at the runtime
+# packs' floor while managed SkiaSharp moves up — a silent mismatch that crashes Skia Desktop on Linux.
+if ($TestGroup -eq '1')
+{
+    $skiaAlignProject = "5.6/uno56netcurrent/uno56netcurrent/uno56netcurrent.csproj"
+    $skiaAlignVersion = "4.148.0"
+    $skiaAlignAssets = "5.6/uno56netcurrent/uno56netcurrent/obj/project.assets.json"
+
+    Write-Host "Validating SkiaSharp.NativeAssets.Linux aligns with an overridden SkiaSharpVersion ($skiaAlignVersion) - issue #23658"
+    dotnet restore $skiaAlignProject "-p:SkiaSharpVersion=$skiaAlignVersion" "-p:RestoreConfigFile=$env:NUGET_CI_CONFIG" -p:EnableWindowsTargeting=true -bl:binlogs/skiasharp-linux-alignment/restore.binlog
+    Assert-ExitCodeIsZero
+
+    $assets = Get-Content $skiaAlignAssets -Raw
+    if ($assets -notmatch [regex]::Escape("SkiaSharp.NativeAssets.Linux/$skiaAlignVersion"))
+    {
+        throw "SkiaSharp.NativeAssets.Linux did not resolve to $skiaAlignVersion to match the overridden managed SkiaSharp. The Uno.Sdk must inject the Linux native asset for desktop heads (issue #23658)."
+    }
+    Write-Host "SkiaSharp.NativeAssets.Linux correctly aligned to $skiaAlignVersion"
+}


MartinZikmund:

MartinZikmund:
From https://www.nuget.org/packages/SkiaSharp#dependencies-body-tab

image

MartinZikmund:
Added a CI regression guard (2nd commit): the Linux template-app job now restores the `uno56netcurrent` `net10.0-desktop` head with `-p:SkiaSharpVersion=4.148.0` and asserts `SkiaSharp.NativeAssets.Linux` resolves to `4.148.0` (matching managed SkiaSharp). This fails on `master` today and passes with the SDK injection in this PR — locally reproduced with a public-package project (`Uno.WinUI.Runtime.Skia.X11` + `SkiaSharp 4.148.0`) where the Linux native otherwise stays stuck at the runtime-pack floor `3.119.0` with no restore warning. The full Uno.Sdk-pack + consumer-restore path is exercised by this job on the built SDK.

ramezgerges (approved):

src/Uno.Sdk/targets/Uno.Implicit.Packages.ProjectSystem.Desktop.targets
```diff
@@ -30,5 +30,13 @@
...
 		<_UnoProjectSystemPackageReference Include="Uno.WinUI.Runtime.Skia.Win32" ProjectSystem="true" />
 		<_UnoProjectSystemPackageReference Include="Uno.WinUI.Runtime.Skia.X11" ProjectSystem="true" />
+
+		<!--
+			The managed SkiaSharp package auto-references its Win32/macOS native assets but not the Linux one,
+			which only arrives as a transitive floor from the X11 / FrameBuffer runtime packs above. Inject it
+			so overriding SkiaSharpVersion retargets the Linux native asset too; otherwise a managed SkiaSharp
+			bumped past Uno's floor keeps the old libSkiaSharp on Linux and crashes Skia Desktop at startup.
+		-->
+		<_UnoProjectSystemPackageReference Include="SkiaSharp.NativeAssets.Linux" ProjectSystem="true" />
```
ramezgerges, line 40: It would be better if we made that fix on SkiaSharp's side.

MartinZikmund: @ramezgerges Can you do a PR on their side please? We can then undo this afterwards transparently after they merge into stable, as it is in Uno.Sdk anyway.

Issue #22927, opened 2026-03-31
chore: Align UWPSyncGenerator library routing with WinUI structure


MartinZikmund:
## Summary

The `Uno.UWPSyncGenerator` has accumulated special-case routing overrides in `GetNamespaceBasePath()` (the `HAS_UNO_WINUI` path) to avoid breaking changes when types were previously introduced in different Uno libraries than where WinUI defines them. These workarounds should be resolved by properly aligning the library structure with WinUI.

## Current workarounds

All located in `src/Uno.UWPSyncGenerator/Generator.cs`, method `GetNamespaceBasePath()`.

### Type-specific overrides

| Type(s) | Current target | Correct target | Reason |
|---------|---------------|----------------|--------|
| `CreateFromStringAttribute` | `Uno.UWP` | Assembly-based (would be `Uno.UI` from `Microsoft.WinUI`) | Type-specific override before namespace routing |
| `Microsoft.UI.Colors`, `Microsoft.UI.ColorHelper`, `Microsoft.UI.FontWeights` | `Uno.UI` | Assembly-based (`Uno.UI` from `Microsoft.WinUI`) | Happens to match correct target but uses name-matching instead of assembly routing |
| `Microsoft.UI.IClosableNotifier`, `Microsoft.UI.ClosableNotifierHandler` | `Uno.UI.Composition` | `Uno.UWP` | Was previously introduced in `Uno.UI.Composition` |
| `WinRT.Interop.*` (`WindowNative`, `InitializeWithWindow`) | `Uno.UI` | `Uno.Foundation` (from `WinRT.Runtime` assembly) | Hand-written implementations in `Uno.UI` depend on `Microsoft.UI.Xaml.Window` — can't move to lower-level assembly |

### Namespace-level backwards compatibility redirects

| Namespace | Current target | Correct target | Reason |
|-----------|---------------|----------------|--------|
| `Microsoft.UI.Content.*` | `Uno.UI` | `Uno.UWP` (from `Microsoft.WinUI` assembly) | Was previously generated in `Uno.UI` |
| `Microsoft.UI.System.*` | `Uno.UI` | `Uno.UWP` | Kept in `Uno.UI` for consistency with other `Microsoft.UI.*` namespaces |
[...]

PR #23669, opened 2026-07-07
refactor(sync-gen)!: Relocate 4 namespaces to WinUI-aligned assemblies

MartinZikmund:
**GitHub Issue:** #22927 (partial — removes 4 of the back-compat redirects; `Content` / `WebView2` / `WindowNative` remain as follow-ups)

## PR Type:

## What changed? 🚀

The WinAppSDK sync generator (`Uno.WinAppSDKSyncGenerator`) carried a block of **"BACKWARDS COMPATIBILITY REDIRECTS"** in `GetNamespaceBasePath` that forced several namespaces into a legacy assembly instead of the one matching their WinAppSDK metadata. This removes the redirects for the **four namespaces that route cleanly**, so each now generates into its WinUI-aligned assembly:

| Namespace | WinAppSDK assembly | Was | Now |
|---|---|---|---|
| `Microsoft.UI.System` (`ThemeSettings`) | InteractiveExperiences | `Uno.UI` | **`Uno.UWP`** |
| `Microsoft.Graphics.DirectX` / `.Display` | InteractiveExperiences | `Uno.UI.Composition` | **`Uno.UWP`** |
| `Microsoft.UI.IClosableNotifier` / `ClosableNotifierHandler` | InteractiveExperiences | `Uno.UI.Composition` | **`Uno.UWP`** |
| `Microsoft.Windows.ApplicationModel.Resources` (MRT-Core) | Resources.Projection | `Uno.UI` | **`Uno.UWP`** |

- **Hand-written partials are moved next to their regenerated stubs.** Partials are relocated *before* regeneration so the generator's per-head "already declared" dedup keeps the stubs skipped on every target — otherwise the Wasm/mobile heads emit duplicate definitions (`CS0101`/`CS0102`).
- **Fixes a name-shadowing break** the `Microsoft.UI.System` move exposes: the namespace becomes newly visible in `Uno.UI.Composition`, shadowing the BCL `System`, so `SkiaExtensions` now qualifies `global::System.Drawing.Color`.
- **Documents the intentionally-retained redirects** (`Content`, `WebView2`, `WindowNative`) and the `Input` / `Xaml.Automation` force-routes, whose WinUI-correct assembly cannot host their hand-written implementations without a dedicated seam.
- **`PackageDiff` 6.6 baseline updated** for the relocated public types (they move from package `Uno.WinUI` to `Uno.WinRT`).

### Why the remaining redirects aren't in this PR

Each is a genuine architectural constraint of Uno's layered assemblies (`Uno.Foundation → Uno.UI.Dispatching → Uno.UWP → Uno.UI.Composition → Uno.UI`), not just caution:

- **`Microsoft.UI.Content`** — WinUI assembly is InteractiveExperiences (→ Uno.UWP), but `ContentIsland`/`ContentSite` require `Compositor`/`Visual` from `Uno.UI.Composition`, a layer *above* Uno.UWP. Its reachable home is `Uno.UI.Composition`, not Uno.UWP.
- **`WinRT.Interop.WindowNative` / `InitializeWithWindow`** — defined in `Microsoft.Windows.SDK.NET` (→ Uno.UWP), but the impl depends on `Microsoft.UI.Xaml.Window` (Uno.UI); needs an `ApiExtensibility` seam.
- **`Microsoft.Web.WebView2.Core`, `Microsoft.UI.Input`** — kept in `Uno.UI`; their implementations are welded to the Uno.UI visual tree / windowing.

### Validation

- ✅ **Compile** — full Skia stack (`Uno.UI.Skia` → Foundation / Dispatching / UWP / Composition / UI), 0 errors.
- ✅ **Compile** — `Uno.UWP` Wasm head, 0 errors (the native head that surfaced, and now passes, the dedup case).
- Generated tree regenerated from the sync generator; diff is surgical — only the 5 relocated namespaces change.
- **No runtime behavior change**: relocated types are `[Uno.NotImplemented]` stubs plus their existing hand-written partials — only the declaring assembly changes.

## PR Checklist ✅

- [ ] 🧪 Added Runtime tests / UI tests / sample — *n/a: assembly relocation of `NotImplemented` stubs; validated by cross-head compile.*
- [ ] 📚 Docs have been added/updated — *n/a.*
- [ ] 🖼️ Validated PR `Screenshots Compare Test Run` results — *n/a (no visual change).*
- [ ] ❗ Contains **NO** breaking changes — *intentional binary break for 7.0; targets `feature/breakingchanges` (see below).*

> **Breaking change (intentional, 7.0).** Relocating a public type across assemblies is binary-breaking: `Microsoft.UI.System.ThemeSettings`, `Microsoft.Graphics.DirectX.*`, `Microsoft.Graphics.Display.*`, `Microsoft.UI.IClosableNotifier` / `ClosableNotifierHandler`, and `Microsoft.Windows.ApplicationModel.Resources.*` move from `Uno.UI` / `Uno.UI.Composition` (package `Uno.WinUI`) to the `Uno` assembly / Uno.UWP (package `Uno.WinRT`). **Source-compatible** — the namespaces are unchanged, so ordinary `using`-based code keeps compiling; only assembly-identity-sensitive code (`typeof`/reflection/`InternalsVisibleTo`) is affected. This is one of the intended 7.0 major-version corrections to align Uno's generated surface with WinAppSDK.

https://claude.ai/code/session_01GDb8ugFEpgnj4B2Z2E5psM


build/PackageDiffIgnore.xml
Xiaoy312, line 1: you should move these under 6.5; 6.6 package isnt out yet on nuget
not sure why 6.6 are even there to begin with, let alone merged?

src/Uno.UWP/Generated/3.0.0.0/Microsoft.Graphics.Display/DisplayAdvancedColorInfo.cs
```diff
@@ -3,18 +3,18 @@
 #pragma warning disable 114 // new keyword hiding
 namespace Microsoft.Graphics.Display
 {
-#if __SKIA__ || __NETSTD_REFERENCE__
```
Xiaoy312, line 6: if we are just move them around (read: regenerated),
shouldnt the #if-conditionals remain the same?

MartinZikmund:
## Review comments addressed — pushed `692fb77c8f`

Two commits on top of `00ac445dbe`:

| commit | what |
|---|---|
| `8900bfec2a` | `fix(build): Move relocation ignores to 6.5 baseline` |
| `692fb77c8f` | `docs(sync-gen): Clarify Microsoft.UI.Content redirect` |

All 6 unresolved threads are covered below.

### Fixed (5)

|---|---|---|
| `build/PackageDiffIgnore.xml` (@Xiaoy312) | These belong under 6.5 — 6.6 isn't on NuGet yet | Moved all **22** type-relocation entries from the `baseVersion="6.6"` `IgnoreSet` into the existing `baseVersion="6.5"` one. You're right, and the file's own CDATA guidance agrees: *"The 'baseVersion' should be the current latest stable version published on nuget, NOT what will be published by the PR in progress."* With a 6.6 baseline the diff tool has no baseline package to resolve, so the ignores would silently never apply. |
| `build/PackageDiffIgnore.xml:3015` (github-actions) | Same point — use the 6.5 baseline that already exists at line 2113 | Same fix. The pre-existing 6.6 `FeatureConfiguration` entries (#2053) have the same defect, but they aren't mine and I left them alone rather than widen this PR — flagging below. |
| `src/Uno.WinAppSDKSyncGenerator/Generator.cs:443` (github-actions) | Comment says the WinUI-correct home is `Uno.UI.Composition`, but the code returns `Uno.UI` | Comment reworded to separate *current* home from *eventual* target. |
| `src/Uno.WinAppSDKSyncGenerator/Generator.cs:443` (copilot) | Same comment/code mismatch | Same fix. |
| `src/Uno.WinAppSDKSyncGenerator/Generator.cs:443` (github-actions) | Same — a future reader could "fix" the return to `Uno.UI.Composition` and break the build | Same fix. New text: <br>`// which Uno.UWP cannot reference. The stubs therefore stay in Uno.UI for now; the eventual`<br>`// Uno home is Uno.UI.Composition, which requires a layering seam not yet in place.` |

### No change needed (1)

| file:line | the point | why the code is already right |
|---|---|---|
| `.../Microsoft.Graphics.Display/DisplayAdvancedColorInfo.cs:6` (@Xiaoy312) | *"if we are just moving them around (read: regenerated), shouldn't the `#if`-conditionals remain the same?"* | They can't, and the widening is the correct output. The conditional set isn't a property of the **type** — it's a function of the **heads the destination assembly is generated for**. `Uno.UI.Composition` has 2 heads (`Uno.UI.Composition.Reference`, `.Skia`) → `#if __SKIA__ \|\| __NETSTD_REFERENCE__`. `Uno.UWP` has 4 (`Uno.Reference`, `Uno.Skia`, `Uno.Wasm`, `Uno.netcoremobile`) → `#if __ANDROID__ \|\| __IOS__ \|\| __TVOS__ \|\| __WASM__ \|\| __SKIA__ \|\| __NETSTD_REFERENCE__`. So no head *loses* the type; Android/iOS/tvOS/WASM simply gain the `[NotImplemented]` stub they never had — which is the intent, since in WinAppSDK these live in the WinRT projection that is available on every platform. The only real change is the assembly (`Uno.WinUI` → `Uno.WinRT`), i.e. exactly the intentional binary break the ignore entries above cover. |

src/Uno.WinAppSDKSyncGenerator/Generator.cs
Xiaoy312, line 1: have we not updated sync-gen to remove the first 4 constants, yet?:
```c#
#if __ANDROID__ || __IOS__ || __TVOS__ || __WASM__ || __SKIA__ || __NETSTD_REFERENCE__
```

Xiaoy312 (approved):

Issue #14478, opened 2023-11-20
Remove `AdjustArrange` as a breaking change


Youssef1313:
### Current behavior

`AdjustArrange` exists and always returns the same size.

### Expected behavior

It shouldn't exist.

### Works on UWP/WinUI

None

PR #23676, opened 2026-07-07
refactor(layout)!: Remove no-op AdjustArrange (BC70)

Xiaoy312:
**GitHub Issue:** https://github.com/unoplatform/uno-private/issues/2125

## PR Type:

> ⚠️ Contains a **breaking change** (`refactor!`) — removes public API. See the breaking-changes section below.

## What changed? 🚀

Implements **BC70** from breaking-changes Phase 3 (#8339, tracked publicly as #14478): removes the no-op `AdjustArrange(Size)` member.

- Deletes the `AdjustArrange` declaration from the internal `IFrameworkElement` interface and both implementations (`FrameworkElement.Interface.skia.cs`, `FrameworkElement.Interface.reference.cs`), each of which just returned its input.
- Inlines the sole call site in the legacy `Layouter` arrange path (a no-op assignment `layoutFrame.Size = frameworkElement.AdjustArrange(layoutFrame.Size)`).
- Drops the `AdjustArrange` stub from the `NativeViewIFrameworkElement` test double.
- Registers the publicly-surfaced `FrameworkElement.AdjustArrange` in `build/PackageDiffIgnore.xml` (6.5 set).

The companion `AdjustArrangePartial` hook (which had zero implementations) and the native mixin emitters that once exposed a `public virtual AdjustArrange` on the native heads are already gone with the native-rendering drop, leaving only the pieces removed here.

**On default settings runtime behavior is unchanged** — `AdjustArrange` always returned its input unchanged, so inlining the call is behavior-preserving on every target.

## PR Checklist ✅

### Breaking changes & migration

- **`AdjustArrange(Size)` (BC70):** declared on the internal `IFrameworkElement`, it was publicly surfaced only as the implicit `FrameworkElement.AdjustArrange(Size)` implementation. That public method is removed (binary break). It was a pure no-op that always returned its input, so there is no behavioral replacement. Migration: none — the only theoretical break was a native-iOS subclass calling/overriding the public-virtual member, and that native target is being dropped this major version.

### Validation evidence

- **Compile:** `Uno.UI.Skia` + `Uno.UI.Reference` + `Uno.UI.RuntimeTests.Skia` (net10.0) all build clean — 0 warnings, 0 errors.
- **Code review:** `AdjustArrange` is a no-op on every implementation, so inlining the call site is behavior-preserving; a repo-wide grep finds zero remaining references to the identifier.
- **Not validated locally (CI-gated):** native heads (iOS/Android) were not built locally; the native `AdjustArrange` implementation came from mixin generators that no longer exist, so this removes an interface member whose native impl source is already gone. The `PackageDiff` allow-list runs CI-only.

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/Layouter/Layouter.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/Layouter/Layouter.cs
@@ -712,9 +712,6 @@ namespace Microsoft.UI.Xaml.Controls
 				// Calculate Create layoutFrame and apply child's margins
 				var layoutFrame = new Rect(x, y, width, height).DeflateBy(childMargin);
 
-				// Give opportunity to element to alter arranged size
-				layoutFrame.Size = frameworkElement.AdjustArrange(layoutFrame.Size);
-
 				// Calculate clipped frame.
 				var clippedFrameWithParentOrigin =
 					layoutFrame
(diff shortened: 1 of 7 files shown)


src/Uno.UI/UI/Xaml/Controls/Layouter/Layouter.cs
```diff
@@ -712,9 +712,6 @@ public void ArrangeChild(View view, Rect frame)
 				// Calculate Create layoutFrame and apply child's margins
 				var layoutFrame = new Rect(x, y, width, height).DeflateBy(childMargin);
 
-				// Give opportunity to element to alter arranged size
-				layoutFrame.Size = frameworkElement.AdjustArrange(layoutFrame.Size);
-
```
ramezgerges, line 717: `Layouter` can be dropped entirely

ramezgerges (approved):

PR #23689, opened 2026-07-08
fix(skia): Paint TextDecorations underline/strikethrough on TextBlock

### LINKED ISSUE #23687: [Skia] `TextDecorations` (underline/strikethrough) not rendered on `TextBlock`
### Current behavior 🐛

`TextDecorations` (Underline / Strikethrough) set on a `TextBlock` or `Run` are not painted by the Skia text renderer, so the text shows without any decoration line.

### Expected behavior 🎯

Underline and strikethrough should be rendered, matching WinUI.

### Renderer 🎨

- [x] Skia
- [ ] Native

### Affected platforms 📱💻🖥️

All platforms 🌐

morning4coffe-dev:
**GitHub Issue:** closes #23687

## PR Type:

🐞 Bugfix

## What changed? 🚀

The Skia text renderer (`UnicodeText`) ignored `TextDecorations`, so `Underline` and `Strikethrough` set on a `TextBlock` or `Run` were never painted. This draws the underline/strikethrough lines using the font metrics (position + thickness) and the run's foreground, matching WinUI's DWrite text renderer (`DrawUnderline` / `DrawStrikethrough`). Line-trailing whitespace is not decorated, consistent with WinUI. This also restores `Hyperlink` underlines, which relied on the same path.

## PR Checklist ✅


src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBlock_TextDecorations.cs
```diff
@@ -0,0 +1,134 @@
...
+using Windows.UI.Text;
+using Microsoft.UI.Xaml;
+using Microsoft.UI.Xaml.Controls;
+using Microsoft.UI.Xaml.Documents;
+using Microsoft.UI.Xaml.Media;
+
+namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
+{
+	[TestClass]
+	[RunsOnUIThread]
```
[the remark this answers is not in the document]
morning4coffe-dev: Addressed in `23f534c78a`: the test class now carries `[GitHubWorkItem("https://github.com/unoplatform/uno/issues/23687")]`, and the Skia-only coverage is expressed with `[PlatformCondition(...)]` rather than `#if`-gated test declarations.

MartinZikmund:
@ramezgerges can you check this one please?

MartinZikmund:
@ramezgerges ping on this one

src/Uno.UI/UI/Xaml/Documents/UnicodeText.cs
```diff
@@ -113,7 +114,7 @@ public static Cluster Create(string _text, int indexStart, int indexEnd, LinkedL
 	private readonly List<LinkedListNode<Cluster>> _clustersInLogicalOrder;
 	private readonly LinkedList<Glyph> _glyphs;
 	private readonly List<(int end, FlowDirection direction)> _bidiBreaks;
-	private readonly List<(int end, Brush? foreground, FlowDirection direction)> _runBreaks;
+	private readonly List<(int end, Brush? foreground, FlowDirection direction, global::Windows.UI.Text.TextDecorations decorations)> _runBreaks;
```
Xiaoy312: using ...;
or if that brings in collision:
using TextDecorations =...;

src/Uno.UI/UI/Xaml/Documents/UnicodeText.cs
```diff
@@ -1023,6 +1025,55 @@ highlighter.Value.background is not null
...
+
+				if (decorationRightX > decorationLeftX)
+				{
+					var decorationBaseline = y + line.baselineOffset;
+					var decorationColor = BrushToColor(_runBreaks[runBreakIndex].foreground, session.Opacity);
+					var decorationMetrics = fontDetails.SKFontMetrics;
+
+					if ((runDecorations & global::Windows.UI.Text.TextDecorations.Underline) != 0)
+					{
+						var underlineThickness = decorationMetrics.UnderlineThickness ?? Math.Max(1f, fontDetails.SKFontSize / 20f);
```
Xiaoy312: add something like:
```cs
// Fallbacks for fonts that don't publish underline/strikeout metrics, as a fraction of
// the em size so they track the font size.
private const float FallbackDecorationThicknessRatio = 1f / 20f;   // 0.05 em
private const float FallbackUnderlinePositionRatio = 1f / 10f;     // 0.1 em below the baseline
private const float FallbackStrikeoutPositionRatio = -1f / 3f;     // ~0.33 em above the baseline (y-down), approximating mid x-height
```

Xiaoy312 (approved):

src/Uno.UI/UI/Xaml/Documents/UnicodeText.cs
morning4coffe-dev: Done in `1448400`:

```cs
// Fallbacks for fonts that don't publish underline/strikeout metrics, as a fraction of the em
// size so they track the font size. They approximate what real fonts publish (measured on
// Segoe UI/Arial/Times New Roman/Consolas: ~0.05 em thickness, ~0.1 em under and ~0.25 em over
// the baseline). Positions are the top edge of the line, positive downwards.
private const float FallbackDecorationThicknessRatio = 1f / 20f;
private const float FallbackUnderlinePositionRatio = 1f / 10f;
private const float FallbackStrikethroughPositionRatio = -1f / 4f;
```

Two deliberate differences from the suggested wording:

- The strikethrough ratio moved from `-1/3` to `-1/4` em. It is the **top edge** of the line, not its centre, and real fonts publish ~0.25–0.26 em there: reading the raw tables through SkiaSharp gives `OS/2.yStrikeoutPosition` = 530/2048 = 0.259 em for Segoe UI, Arial and Times New Roman, and 512/2048 = 0.25 em for Consolas. `-1/3` em placed the fallback line noticeably too high.
- The comment says the values are top edges rather than centres, which is the same convention the metric-driven path now uses (`DWRITE_UNDERLINE.offset` semantics — see the geometry change in the same commit).

src/Uno.UI/UI/Xaml/Documents/UnicodeText.cs
```diff
@@ -1023,6 +1025,55 @@ highlighter.Value.background is not null
...
+					contentRightX = contentLeftX + line.widthWithoutTrailingSpaces;
+				}
+
+				var decorationLeftX = Math.Max(unalignedX + alignmentOffset, contentLeftX);
+				var decorationRightX = Math.Min(unalignedX + alignmentOffset + cluster.Value.width, contentRightX);
+
+				if (decorationRightX > decorationLeftX)
+				{
+					var decorationBaseline = y + line.baselineOffset;
+					var decorationColor = BrushToColor(_runBreaks[runBreakIndex].foreground, session.Opacity);
```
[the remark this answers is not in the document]
morning4coffe-dev: Both points were valid and are fixed in `1448400`.

1. **`effectiveOpacity`** — `decorationColor` now uses `effectiveOpacity`, so a decoration is as opaque as the glyphs it belongs to under a high-contrast backplate.

2. **High-contrast palette** — honoured, but with `highContrastForeground` only, without the selection branch from the suggestion. WinUI resolves the brush for decoration lines in `D2DTextDrawingContext::HWRenderLines` through `GetAlternativeForegroundBrush(m_lines[i].BrushSource)`, which returns `SystemColorWindowText` (or `SystemColorHotlight` for hyperlinks) and has no selection-foreground variant — decoration lines live in `m_lines` and keep the run's brush, unlike glyph runs which are re-emitted with the highlight foreground. Uno's own `ParsedText` renderer already encodes the same rule (`// We don't need to see where selection starts and ends in this case, as the decoration color doesn't get affected by the selection`), so mirroring the composition-underline branch here would have diverged from both. The composition underline is an IME affordance with no WinUI counterpart, which is why it does branch on the selection.

Upstream reference: `microsoft-ui-xaml@439bda90582f7892ca595a193f55dbfc6ae2010c`, `dxaml/xcp/core/native/text/Common/D2DTextDrawingContext.cpp:1422-1440` and `:2708-2752`.

src/Uno.UI/UI/Xaml/Documents/UnicodeText.cs
morning4coffe-dev: Done in `1448400`. No alias was needed: `using Windows.UI.Text;` is already imported at the top of `UnicodeText.cs` and nothing else in scope is named `TextDecorations`, so the `global::Windows.UI.Text.` qualification was simply dropped from all six occurrences — the `_runBreaks` field declaration and initialisation, the `EnumerateShapingRuns` parameter, and the three enum comparisons in `Draw`.

morning4coffe-dev:
@pr-comment.md

morning4coffe-dev:
CI is green on `f242f356`.

The first run (build 229939) had two failures in `Tests - Desktop Skia Windows`, both unrelated to this change and both already auto-retried twice by the pipeline:

- `Given_MediaPlayerElement.When_MediaPlayerElement_SetIsFullWindow_Check_Fullscreen` — `Assert.AreNotEqual failed ... 'notExpected' expression: 'mpp.ActualWidth'`, i.e. the presenter measured 0 after going full-window. The test is already tracked by `[GitHubWorkItem(#14735)]` and passes locally on this build.
- `Given_HotReloadWorkspace.When_HotReloadScenario` — the child HRApp run failed after two rebuild/relaunch retries, with `wglCreateContext failed: The pixel format is invalid.` on the agent.

Re-running the same commit (build 229957) completed **succeeded** with `Tests - Desktop Skia Windows` green, so both were flakes rather than regressions. Neither touches the text pipeline: the decoration code only runs when a run has `TextDecorations != None`.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBlock_TextDecorations.cs
```diff
@@ -0,0 +1,266 @@
...
+				VerticalAlignment = VerticalAlignment.Top,
+				Child = textBlock,
+			};
+
+			await UITestHelper.Load(container);
+
+			var metrics = run.FontInfo.SKFontMetrics;
+			if (metrics.UnderlinePosition is not { } underlinePosition || metrics.UnderlineThickness is not { } underlineThickness)
+			{
+				Assert.Inconclusive("The resolved font does not publish underline metrics.");
```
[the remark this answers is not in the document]
morning4coffe-dev: Valid — fixed at the root in `1001999`, since the underlying problem was that the test relied on whatever font the host resolved.

The geometry test now pins a font that is guaranteed to be present on every Skia target:

```cs
FontFamily = new FontFamily("ms-appx:///Uno.UI.RuntimeTests/Assets/Fonts/Roboto-Regular.ttf"),
```

and waits for the asset to actually be resolved for the run before reading metrics (font assets load asynchronously, so `Inline.FontInfo` can briefly return the fallback details until `OnFontLoaded` invalidates the cache):

```cs
await UITestHelper.WaitFor(
    () => run.FontInfo.SKFont.Typeface?.FamilyName == RobotoFamily,
    5000,
    $"Timed out waiting for {RobotoAsset} to be resolved for the run.");
```

`Assert.Inconclusive` is gone; a missing metric is now a hard failure, exactly as suggested:

```cs
if (metrics.UnderlinePosition is not { } underlinePosition || metrics.UnderlineThickness is not { } underlineThickness)
{
    Assert.Fail($"{RobotoFamily} publishes underline metrics, so their absence means the run did not resolve to the bundled asset.");
    return;
}
```

Evidence for the choices:

- **The asset is reachable on every Skia target.** `Given_TextBlock.When_FontFamily_Changed` already exercises `ms-appx:///Uno.UI.RuntimeTests/Assets/Fonts/Roboto-Regular.ttf` under the same `[PlatformCondition(ConditionMode.Include, RuntimeTestPlatforms.Skia)]` and asserts the desired size *changes* after switching to it, so it only passes when the font really loads. It is green on Desktop Windows/Linux/Linux FrameBuffer/macOS, Android 0-4, iOS 0-3 and WASM 0-3 in build 229957.
- **Roboto-Regular publishes the metrics.** Reading the bundled file directly: `post.underlinePosition = -150`, `underlineThickness = 100`, `unitsPerEm = 2048` → at the test's 200 px that is `UnderlinePosition = 14.648`, `UnderlineThickness = 9.766`, `Ascent = -190.039`, so `expectedTop = 204.688`.
- **The guard still fails before / passes after.** Re-ran the fails-before experiment with the pinned font: with the previous centred rendering the test fails with `Underline top row is 200, expected 204.6875` (off by ~half the thickness, 4.88 px), and passes with the fix. Pinning actually widens the margin against the ±2 tolerance from ~2.5 px to ~4.9 px, so the guard is now less font-dependent than before.

Local validation on `1001999`: analyzer-enabled build (no `UnoFastDevBuild`) of `SamplesApp.Skia.Generic` at `net10.0` — 0 warnings, 0 errors; all 6 `Given_TextBlock_TextDecorations` tests pass on Skia Desktop (Win32).

PR #23692, opened 2026-07-08
fix(hot-reload): scope the dev-server workspace to the running app's target framework

dr1rrb:
closes https://github.com/unoplatform/uno.hotdesign/issues/7130#top
linked https://github.com/unoplatform/uno-private/issues/2211#top

## PR Type:

🐞 Bugfix

## What changed? 🚀

Two robustness fixes for the hot-reload dev-server, plus a supporting refactor.

**1. Benign assembly-resolution misses were logged as errors.** The host's default-load-context resolver is a safety net: when it cannot resolve an assembly, the request is still satisfied downstream by an add-in's or the hot-reload processors' own resolver (for example, dependencies that ship only under the processors folder). These *expected* misses were written to the error stream and showed up as red `[ERROR]` lines. They are now logged at debug level on the standard output, with reworded text making clear they are expected.

**2. Hot reload could compile the wrong target framework.** When the application head is multi-targeted and the evaluated `TargetFramework` is empty, Roslyn's `MSBuildWorkspace` loads one project per `TargetFrameworks` entry. Debugging one target could then pull in another target's flavor, producing duplicate-type / overload / signature errors (`CS0433`, `CS1503`, `CS1729`, …) that block every hot-reload update — or fail initialization outright with a null `CompilationOutputInfo.AssemblyPath` for a flavor that was never built.

The dev-server now scopes the workspace to the single target framework the running application actually reports:

- The application composes its runtime target framework from information available at run time — the platform is compile-time knowledge of the (per-flavor) client, the framework version comes from the assembly's `TargetFrameworkAttribute` — and sends it in a new `ConfigureServer.RuntimeTargetFramework` field. This removes the dependency on the IDE promoting a build-captured `TargetFramework` global, which was the fragile link.
- The server drops the build-time property promotion in favour of a small, explicit global-property whitelist (`Configuration`, `Platform`, `Solution*`, and the hot-reload host marker); everything else captured at build time is intentionally not re-applied.
- Once the workspace has loaded, the head flavors that do not match the reported target framework — and every project only reachable from them — are removed from the solution snapshot. The pass is conservative: if the head cannot be found, the reported framework is unknown, or nothing matches, the solution is left intact and the situation is reported.
- Older clients that do not send the runtime target framework fall back to the build-captured `TargetFramework`.

**Supporting refactor.** The workspace provider now returns the raw `MSBuildWorkspace` (the caller applies the target-framework filter, since it is what knows the runtime framework), and the hot-reload pipeline operates on `Solution` snapshots throughout. See `specs/047-hotreload-workspace-single-tfm` for the full design, decisions and matching rules.

## PR Checklist ✅

- [x] 🧪 Added [Runtime tests, UI tests, or a manual test sample](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) (for bug fixes / features, if applicable) — unit tests for target-framework parsing, matching and snapshot filtering, plus the hot-reload workspace runtime scenario.
- [ ] 📚 Docs have been added/updated following the [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features) — internal design spec added under `specs/`; no user-facing documentation change.
- [ ] 🖼️ Validated PR `Screenshots Compare Test Run` results. — n/a, no UI change.
- [ ] ❗ Contains **NO** breaking changes — see below.

### Breaking change

The `Workspace`-based `HotReloadManager.CreateAsync` and `WatchHotReloadService.CreateAsync` overloads are removed; the hot-reload pipeline now operates exclusively on `Solution` snapshots. This is an internal tooling API used to host the hot-reload engine, not part of the app-facing Uno API surface. External hosts should pass `workspace.CurrentSolution` (or a `Func<CancellationToken, ValueTask<Solution>>`) instead of the workspace. `CompilationWorkspaceProvider.CreateWorkspaceAsync` now returns `MSBuildWorkspace`; existing `(MSBuildWorkspace)` casts on its result remain valid.


src/Uno.HotReload/Utils/RoslynExtensions.TryGetTargetFramework.cs
```diff
@@ -0,0 +1,233 @@
...
+	/// <summary>
+	/// Recognised platform-specific ref pack name prefixes mapped to the corresponding TFM
+	/// platform identifier. The actual on-disk pack name may carry a TFM-encoding suffix
+	/// (e.g. <c>Microsoft.iOS.Ref.net10.0_26.0</c>); the matcher accepts both shapes.
+	/// </summary>
+	private static readonly Dictionary<string, string> _platformPackPrefixes = new(StringComparer.OrdinalIgnoreCase)
+	{
+		["Microsoft.iOS.Ref"] = "ios",
+		["Microsoft.tvOS.Ref"] = "tvos",
+		["Microsoft.MacCatalyst.Ref"] = "maccatalyst",
```
ajpinedam, line 114: I believe martin already remove catalyst (and macos) both from the supported platforms.

ajpinedam (approved):

agneszitte (commented):
Two inline notes, both verified against the code (not just the spec).

specs/047-hotreload-workspace-single-tfm/spec.md
```diff
@@ -0,0 +1,288 @@
...
+
+The resolver derives a project's TFM from the project's **metadata references** (ref-pack
+path classification: `Microsoft.<Platform>.Ref` / base packs) with **preprocessor symbols**
+(`__WASM__`, `__DESKTOP__`) as tiebreaker — deliberately never consulting output paths
+(user-customizable) nor project names. An existing, field-proven implementation is ported
+into the product (`Uno.HotReload/Utils/`) together with its unit tests; the DevServer test
+project (linked-source convention) validates the end-to-end behavior against multi-TFM
+fixtures.
+
+Known gap to address during the port: the classifier does not recognize
```
agneszitte, line 191: Verified against the code — this D5 "known gap" reads as stale: `TryGetTargetFramework` already classifies `Microsoft.Windows.SDK.NET.Ref` → `windows` (in `_platformPackPrefixes`), and it's covered by `TryParseRefPackPath_WindowsSdkRefPack_ParsesPlatform` (plus `TryParseRefPackPath_WindowsDesktopAppRef_BasePackRecognised` for the base pack). 

Could you update the spec so this no longer reads as an open gap? If there's a residual path/format case the map entry doesn't cover, a one-line note on what's still missing would help.

src/Uno.UI.RemoteControl/HotReload/ClientHotReloadProcessor.cs
```diff
@@ -141,6 +142,54 @@ private async Task ConfigureServer()
...
+			{
+				frameworkVersion = parsed.Version;
+			}
+		}
+		catch (Exception)
+		{
+			// Malformed attribute — fall back to the runtime version below.
+		}
+
+		frameworkVersion ??= Environment.Version;
```
agneszitte, line 170: Minor / non-blocking: `GetRuntimeTargetFramework` is `#if`-gated on platform and this malformed-`TargetFrameworkAttribute` → `Environment.Version` fallback isn't unit-tested — the suite covers the server-side parse/match/filter but not the client-side composition. That's largely inherent to the `#if` code; just flagging the one untested branch in case it's worth extracting the version-composition (non-`#if`) part into a tiny testable helper.

agneszitte:

src/Uno.HotReload/Utils/RoslynExtensions.TryGetTargetFramework.cs
dr1rrb: OK, but this allows us to support any kind of projects and cost nothing here

specs/047-hotreload-workspace-single-tfm/spec.md
dr1rrb: will do in a following PR, don't  want to re-trigger a new build

src/Uno.UI.RemoteControl/HotReload/ClientHotReloadProcessor.cs
dr1rrb: Same

src/Uno.HotReload.Tests/Utils/Given_TryGetTargetFramework.cs
```diff
@@ -0,0 +1,448 @@
...
+			projectId,
+			VersionStamp.Default,
+			name: "Test",
+			assemblyName: "Test",
+			language: LanguageNames.CSharp,
+			parseOptions: parseOptions,
+			metadataReferences: metadataReferences);
+
+		var solution = workspace.AddProject(projectInfo);
+		return solution;
```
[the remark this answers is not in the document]
dr1rrb: Addressed in https://github.com/unoplatform/uno/pull/23696 — renamed the misleading `solution` local to `project` (`AdhocWorkspace.AddProject` returns a `Project`).

specs/047-hotreload-workspace-single-tfm/spec.md
dr1rrb: Addressed in https://github.com/unoplatform/uno/pull/23696 — the D5 note no longer reads as an open gap: it now states the `Microsoft.Windows.SDK.NET.Ref` → `windows` mapping (covered by `TryParseRefPackPath_WindowsSdkRefPack_ParsesPlatform` and `…_WindowsDesktopAppRef_BasePackRecognised`), with the one-line residual note.

src/Uno.UI.RemoteControl/HotReload/ClientHotReloadProcessor.cs
dr1rrb: Addressed in https://github.com/unoplatform/uno/pull/23696 — extracted the non-`#if` version composition into `ResolveFrameworkVersion(string?)` (catch narrowed to `ArgumentException`), with unit tests covering the missing / malformed / non-`.NETCoreApp` fallback to `Environment.Version`.

PR #23710, opened 2026-07-09
fix(hot-reload): only request an explicit IDE hot reload inside Visual Studio

dr1rrb:
closes https://github.com/unoplatform/uno.hotdesign/issues/7130
related https://github.com/unoplatform/uno-private/issues/2211

second step of https://github.com/unoplatform/uno/pull/23692
NOT related to the already existing follow up PR https://github.com/unoplatform/uno/pull/23696

## PR Type:

🐞 Bugfix

## What changed? 🚀

Editing a file from a non-Visual-Studio IDE (VS Code, Rider) broke hot reload: after applying the
edit, the dev-server sent a `ForceHotReloadIdeMessage` to the IDE and **waited for a reply that never
came**. Those IDEs register no callback for that message — they drive hot reload through the
dev-server's own pipeline, not an explicit IDE force-reload — so the IDE side only logged
`No callback registered for method: …ForceHotReloadIdeMessage` and sent nothing back. The dev-server
then blocked the update path for the full 25s IDE-result timeout, and `ProcessingFiles` was never
reported.

`RequestHotReloadToIde` is now a no-op outside Visual Studio:

- no `ForceHotReloadIdeMessage` is sent (no more 25s stall, no more "no callback" error);
- no `ProcessingFiles` is reported (the IDE never acknowledges, and the dev-server's own pipeline
  reports hot-reload status independently);
- it returns `false`, so the no-changes auto-retry (`HotReloadOperation`) completes the operation
  instead of deferring while waiting for an IDE acknowledgement that never comes.

Behaviour inside Visual Studio is unchanged (VS registers the handler and replies).

## PR Checklist ✅

- [ ] 🧪 Added [Runtime tests, UI tests, or a manual test sample](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) (for bug fixes / features, if applicable) — server-side IDE-channel timing behaviour; validated by running the hot-reload scenario from VS Code / Rider (no automated coverage for the IDE-channel round-trip).
- [ ] 📚 Docs have been added/updated following the [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features) — n/a.
- [ ] 🖼️ Validated PR `Screenshots Compare Test Run` results. — n/a, no UI change.
- [x] ❗ Contains **NO** breaking changes — internal behaviour only, no API change.

### DIFF
--- a/src/Uno.UI.RemoteControl.Server.Processors/HotReload/ServerHotReloadProcessor.cs
+++ b/src/Uno.UI.RemoteControl.Server.Processors/HotReload/ServerHotReloadProcessor.cs
@@ -337,6 +337,18 @@ namespace Uno.UI.RemoteControl.Host.HotReload
 
 		private async ValueTask<bool> RequestHotReloadToIde()
 		{
+			// Only Visual Studio handles the explicit force-hot-reload request. Other IDEs (VS Code,
+			// Rider) register no callback for ForceHotReloadIdeMessage, so sending it there only logs a
+			// "no callback" error on the IDE side and blocks here until the IDE-result timeout — and they
+			// drive hot reload through the dev-server's own pipeline anyway. Outside VS this is a no-op:
+			// no request is sent and no ProcessingFiles is reported, and we return false so the
+			// no-changes auto-retry (HotReloadOperation) completes the operation instead of deferring
+			// while waiting for an IDE acknowledgement that will never come.
+			if (!_isRunningInsideVisualStudio)
+			{
+				return false;
+			}
+
 			var result = await SendAndWaitForResult(new ForceHotReloadIdeMessage(GetNextIdeCorrelationId()));
 
 			if (result.IsSuccess)


ajpinedam (approved):

dr1rrb:

kazo0 (approved):

src/Uno.UI.RemoteControl.Server.Processors/HotReload/ServerHotReloadProcessor.cs
```diff
@@ -337,6 +337,18 @@ await _remoteControlServer.SendFrame(new UpdateFileResponse(
...
 		private async ValueTask<bool> RequestHotReloadToIde()
 		{
+			// Only Visual Studio handles the explicit force-hot-reload request. Other IDEs (VS Code,
+			// Rider) register no callback for ForceHotReloadIdeMessage, so sending it there only logs a
+			// "no callback" error on the IDE side and blocks here until the IDE-result timeout — and they
+			// drive hot reload through the dev-server's own pipeline anyway. Outside VS this is a no-op:
+			// no request is sent and no ProcessingFiles is reported, and we return false so the
+			// no-changes auto-retry (HotReloadOperation) completes the operation instead of deferring
+			// while waiting for an IDE acknowledgement that will never come.
+			if (!_isRunningInsideVisualStudio)
```
agneszitte, line 347: Non-blocking: now that this is a pure `_isRunningInsideVisualStudio` guard, it's cheap to lock in with a unit test , assert that outside VS, `RequestHotReloadToIde()` returns `false` and sends no `ForceHotReloadIdeMessage` (a spy `IRemoteControlServer` recording frames is enough; no IDE round-trip needed). 
Can be a separate follow-up PR

dr1rrb: Thanks — agreed in principle, but locking this in with a unit test is more involved than it looks here. `DevServer.Tests` doesn't compile-reference `Server.Processors` (the processors are loaded in a collectible ALC at runtime, and a few tests assert exactly that), so we can't just `new ServerHotReloadProcessor(...)` and call the method: it would need either a reflection-based test (brittle — a rename wouldn't fail at compile time) or a dedicated `Server.Processors.Tests` project. I'd rather not add a fragile reflection test, so I'm deferring the coverage as a follow-up.

More importantly, the `_isRunningInsideVisualStudio` guard is a deliberate stop-gap, not the end state. The clean design is capability-based: the IDE should tell the dev-server it doesn't handle `ForceHotReloadIdeMessage` — i.e. reply "not supported" instead of staying silent — rather than the server hard-coding "is this VS?". As written, the guard bakes per-IDE support knowledge into the server and doesn't scale: every new IDE, or any change in what an IDE supports, means editing this check. The real fix belongs on the IDE-channel protocol (an explicit capability / "unsupported" response), and this guard should be removed once that exists.

Issue #23041, opened 2026-04-14
test: Add runtime tests for ItemsRepeater fast-scroll and scroll-to-top with high-variance item heights


MartinZikmund:
## Current behavior

The existing runtime test suite for `ItemsRepeater` in `src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Repeater/Given_ItemsRepeater.cs` covers a handful of virtualization scenarios (`When_ItemSignificantlyTaller_Then_VirtualizeProperly`, `When_NestedIRSlowlyChangeViewport_Then_MaterializedNeededItems`, etc.), but does **not** cover the class of problems that are visible to users when `ItemsRepeater` hosts items with very high size variance and the user scrolls quickly:

1. **Fast-scroll jumps with variable-height items**: when the user rapidly navigates (e.g. `ChangeView` to far-apart offsets) through a list where a few items are significantly taller than the running average, newly realized items can end up visually overlapping, appearing above the viewport, or positioned at the wrong `ActualOffset.Y`. This is a direct consequence of `StackLayout.GetAnchorForRealizationRect` / `GetExtent` using average-size estimation and the (Uno-specific) clamp of the extent origin to `0` blocking the correction pass. See #23042 (`fix: StackLayout.GetExtent clamps extent origin to 0`).

2. **Scroll-to-top after variable-height realization**: after scrolling down through a list with tall items and scrolling back up, users can observe that the first item is not flush at offset `0` — some content is inaccessible. `ScrollableHeight` / `ExtentHeight` end up inconsistent with the real positions because the clamp and the missing anchor provider on `ScrollViewer` combine to lose track of the true layout extent.

3. **Anchor maintenance across a measure pass that expands items in place**: when a realized item grows (e.g. late-loading content replaces a short placeholder), `ItemsRepeater` / `ScrollViewer` should keep the currently-visible items visually stable. Today nothing pins them — the visible content jumps.

We do not have targeted regression tests for any of the three scenarios above, so regressions in virtualization can ship undetected.

## Expected behavior

Runtime tests under `Uno.UI.RuntimeTests` exercise and assert the following invariants on Skia Desktop and WebAssembly (and ideally on WinUI for parity):

### Test A — `When_FastScrollWithHighVarianceItems_Then_NoOverlap`

- Sources: 200 items, heights drawn from `{ 40, 40, 40, 1200, 40, 40, 40, ... }` (every 10th item is ~30× taller).
[...]

PR #23719, opened 2026-07-10
fix(scrollviewer): Clamp offset intent to the request-time extent

### LINKED ISSUE #23695: Regression: ScrollViewer.ChangeView beyond the extent chases the growing virtualized extent (breaks WinUI parity; over-fetches incremental loading)
## Summary

On the current **dev** `Uno.WinUI`, `ScrollViewer.ChangeView(...)` to a vertical offset **beyond the current extent** now **chases the virtualized extent as it grows**, instead of **clamping to the live extent** the way **native WinUI** (and Uno **stable 6.5**) do.

For content that grows during/after the scroll (e.g. an `ItemsRepeater` with incremental loading), the preserved "scroll to a far offset" intent keeps the viewport pinned to the moving bottom, so any *"load more when the viewport reaches the end"* consumer fires repeatedly and **over-fetches**.

> [!NOTE]
>  **Affected versions — 6.7-dev only.** Verified by testing each runtime: **6.7-dev is affected**; **6.6 is NOT** (tested `Uno.Sdk 6.6.16` / `Uno.WinUI 6.6.134` — behaves correctly), and neither is 6.5 nor native WinUI. The regressing commit is on `master` and was **not** backported to `release/stable/6.6`, so this can be fixed before 6.7 GA and no released version is impacted.

## Reproduction (core-only, no Uno.Toolkit)

An `ItemsRepeater` (vertical `StackLayout`, 29px items) in a **210×210 `ScrollViewer`**, with a source that adds a 25-item batch each time the viewport reaches the end (the standard incremental-loading trigger: `distanceToEnd = ItemsRepeater.ActualHeight − EffectiveViewport.Bottom`). Scrolling **to the bottom via `ChangeView(null, 10_000, null, disableAnimation: true)`** then counts how many batches load.

Full runnable repro attached: **[IncrementalLoadRepro.zip](https://github.com/user-attachments/files/29829412/IncrementalLoadRepro.zip)** (switch Uno.Sdk in `global.json`; switch scroll mode via the `IRREPRO_MODE` env var; a `net10.0-windows10.0.26100.0` target builds against native WinUI).

### Results — batches loaded

| Scroll mode | **native WinUI** | **Uno 6.5.36 (stable)** | **Uno 6.7.0-dev** |
|---|:---:|:---:|:---:|
| `ChangeView(10_000)` — **beyond extent** | **3** | **3** | **15** ⬅ |
| `ChangeView(ScrollableHeight)` — live max | 7 | 7 | 7 |
| gradual/wheel-like steps | 9 | 9 | 9 |

- **Only the beyond-extent `ChangeView` differs, and only on dev.** Native WinUI and Uno stable both **clamp** (3 batches); dev **chases** the growing extent (15 batches → 375 items).
- The dev run loads until the content height reaches the requested offset (`375 × 29 ≈ 10,000`) — i.e. the beyond-extent offset intent is preserved and chased as the extent grows.
- `ScrollableHeight`/gradual scrolling are **identical** across all three → **natural user scrolling is not affected**.

## Expected vs actual

- **Expected (WinUI parity):** `ChangeView` to an offset beyond the current extent clamps to the live extent; later async extent growth does not re-drive the old intent. → bounded loading (3 batches).
- **Actual (dev):** the beyond-extent offset intent is preserved and chased as the virtualized extent grows → unbounded/over-fetch (15 batches / 375 items).

## Root cause

Regressed by **`385b6146a9` — "fix(scrollviewer): Preserve offset intent across virtualization extent fluctuations (fixes #23041, #23042)"** (part of the wheel-scroll-flicker work). Confirmed:
- **not** in `release/stable/6.5` (→ stable matches WinUI), and
- **in** `master` (→ dev regresses).

## Impact

- `Uno.Toolkit.UI` `ItemsRepeaterExtensionTests.When_Incremental_Load` / `_ShouldStop` / `_IsLoading` fail on nightly (loads 375 items instead of the expected 50; `IsLoading` never settles).
- Any app that **programmatically scrolls to the bottom** (`ChangeView` to a large/`double.MaxValue`-ish offset) **and** loads content incrementally — e.g. chat/feed "jump to latest" then load — will over-fetch.

## Suggested fix direction

`ChangeView` to a beyond-extent offset should **clamp to the live extent** (WinUI parity) rather than persist and chase the requested offset as the extent later grows via async content — or bound the offset-intent preservation to the extent as measured at request time. cc the #23041/#23042 change author.

## Environment

- Uno.Sdk `6.7.0-dev.*` (regressed) vs `6.5.36` (stable, matches WinUI); native WinUI via WinAppSDK on the Windows target.
- `net10.0-desktop` (Skia) + `net10.0-windows10.0.26100.0` (native WinUI); .NET 10; Windows 11.

MartinZikmund:
**GitHub Issue:** closes #23695

## PR Type:

🐞 Bugfix

## What changed? 🚀

`ScrollViewer.ChangeView(...)` to an offset **beyond the current extent** chased the virtualized extent as it grew, instead of clamping to the live extent the way native WinUI (and Uno 6.5/6.6) do. With content that grows during the scroll — an `ItemsRepeater` with incremental loading — the viewport stayed pinned to the moving bottom edge, so every "load more when the viewport reaches the end" consumer kept re-firing and over-fetched: **15 batches / 375 items where WinUI loads 3.**

**Root cause.** `385b6146a9` introduced an offset "intent" to preserve a `ChangeView` target across the extent fluctuations that a virtualized `StackLayout` produces while realizing (#23041). The intent was stored **raw**, and `RecomputeOffsetsFromIntent` re-applied it as `clamp(intent, 0, ScrollableHeight)` on every arrange. As `ScrollableHeight` grew, the clamp grew with it, so a request for `10_000` kept pulling the offset down until the content actually reached `10_000`.

WinUI bounds the request **when it is made**: `ScrollViewer::AdjustTargetVerticalOffset` clamps against `ComputePixelExtentHeight()` with no intervening layout pass, and `ScrollContentPresenter::SetOffsetsWithExtents` stores only the `ValidateInputOffset`-clamped value. Later extent growth can never carry the offset past what the caller could reach at request time.

**The fix.** Clamp in `SetVerticalOffsetIntent` / `SetHorizontalOffsetIntent`, and route every arm site through them — public `ChangeView`, both scrollbar handlers, and the `ScrollContentPresenter` `IScrollInfo` setters, which previously assigned the fields directly. Also skip arming on a `NaN` offset or `NaN` bound: WinUI's `ValidateInputOffset` rejects those outright, and either one would otherwise freeze the recompute or store the request unbounded.

The intent still survives a *transient* shrink, so the #23041 behavior this regressed out of is preserved. Only requests that exceed the request-time maximum change behavior.

### Validation

**Runtime — Skia Desktop.** The new tests fail before and pass after. The end-to-end repeater test reproduces the issue's exact reported numbers pre-fix (15 batches, 375 items, `VerticalOffset=10000`).

**Runtime — native WinUI (WinAppSDK).** `When_ChangeView_BeyondExtent_*` and `When_ChangeView_BeforeFirstLayout_*` pass unmodified against native WinUI, confirming the clamped semantics are parity rather than a guess.

Worth calling out: `When_ExtentShrinksThenGrowsBack_Then_RequestedOffset_Is_Restored` is `[PlatformCondition(Exclude, NativeWinUI)]` because native WinUI was *measured* to **not** restore the offset (it lands at 100, not 200) — its DirectManipulation pipeline drops the request permanently. Uno's restore is a deliberate divergence covering the non-monotonic extent estimate of the virtualized `StackLayout`. The test documents that.

**Regression sweep — Skia Desktop.** `Given_ScrollViewer` + `ScrollViewerTests` + `Repeater` (92 tests) and `Given_ListViewBase` + `Given_ScrollContentPresenter` + `Given_ItemsStackPanel` + `ListViewTests` (152 tests). Failure sets with the fix stashed and applied are **identical** — zero new failures. (The 4 pre-existing failures on my machine are 150%-DPI rounding artifacts.)

## PR Checklist ✅

- [x] 📚 Docs have been added/updated following the [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)  — n/a, internal behavior fix restoring WinUI parity

The regressing commit is on `master` only and was never backported to `release/stable/6.6`, so no released version is affected.

https://claude.ai/code/session_01XbEqMr6QpVhzH9acLwTgnP

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.cs
@@ -819,19 +819,11 @@ namespace Microsoft.UI.Xaml.Controls
 			UpdateComputedVerticalScrollability(invalidate: false);
 			UpdateComputedHorizontalScrollability(invalidate: false);
 
-			// Prefer the WinUI-parity recompute when the caller has expressed an offset intent (most
-			// recent programmatic ChangeView). The recompute keeps VerticalOffset/HorizontalOffset
-			// aligned with the user's intent (clamped to the current ScrollableHeight) across the
-			// realization cascades that virtualizing content (ItemsRepeater) produces. If the
-			// recompute applied an adjustment, skip the legacy TrimOverscroll so we don't double-clamp
-			// in the same arrange cycle.
-			// When no intent is armed (e.g. wheel-driven or touch-driven scrolling), only run the
-			// legacy TrimOverscroll if the SV's *viewport* (its own size) actually changed. Pure
-			// extent shrinkage from virtualizing content's realization must NOT be auto-clamped —
-			// that's the path that produces the "wheel fight" the user reports on high-variance
-			// ItemsRepeater content (issue #23041 / studio.live#816). The viewport-resize case
-			// (When_SizeChanged_Offsets_Adjusted) still runs because that pass sees ViewportHeight
-			// or ViewportWidth different from its previous snapshot.
+			// The recompute realigns the offset with the armed intent, so a TrimOverscroll on top of it
+			// would double-clamp in the same arrange cycle.
+			// Without an intent (wheel- or touch-driven scrolling), TrimOverscroll must only run when the
+			// SV's own viewport changed: auto-clamping the pure extent shrinkage that virtualizing content
+			// produces while realizing is what makes the wheel fight the user's input (#23041).
 			var viewportChanged =
 				!NumericExtensions.AreClose(vpHeight, _lastTrimViewportHeight) ||
 				!NumericExtensions.AreClose(vpWidth, _lastTrimViewportWidth);
@@ -1227,7 +1219,7 @@ namespace Microsoft.UI.Xaml.Controls
 
 			// Scrollbar interactions are explicit offset requests: arm the intent so the post-layout
 			// recompute coerces toward this offset instead of snapping back to a stale programmatic one.
-			_verticalOffsetIntent = offset;
+			SetVerticalOffsetIntent(offset);
 
 			ChangeViewCore(
 				horizontalOffset: null,
@@ -1255,7 +1247,7 @@ namespace Microsoft.UI.Xaml.Controls
 			};
 
 			// Arm the intent — see OnVerticalScrollBarScrolled.
-			_horizontalOffsetIntent = offset;
+			SetHorizontalOffsetIntent(offset);
 
 			ChangeViewCore(
 				horizontalOffset: offset,
@@ -1525,21 +1517,18 @@ namespace Microsoft.UI.Xaml.Controls
 			}
 
 			// Capture the caller's offset intent so the post-layout recompute can preserve it across
-			// realization-driven extent fluctuations. The recompute (RecomputeOffsetsFromIntent) clamps
-			// the intent to the current ScrollableHeight without overwriting the intent itself, mirroring
-			// WinUI's m_Offset.Y vs m_ComputedOffset.Y separation in ScrollContentPresenter
-			// (ScrollContentPresenter_Partial.cpp:902 SetVerticalOffsetPrivate vs :3170 CoerceOffsets).
+			// realization-driven extent fluctuations. See SetVerticalOffsetIntent for the bound it carries.
 			// Intent is only updated when this is an external/programmatic ChangeView; internal calls
 			// (TrimOverscroll, the recompute itself) set _isInternalOffsetAdjustment to suppress this.
 			if (!_isInternalOffsetAdjustment)
 			{
 				if (horizontalOffset is double hh)
 				{
-					_horizontalOffsetIntent = hh;
+					SetHorizontalOffsetIntent(hh);
 				}
 				if (verticalOffset is double vv)
 				{
-					_verticalOffsetIntent = vv;
+					SetVerticalOffsetIntent(vv);
 				}
 			}
 
@@ -1567,11 +1556,8 @@ namespace Microsoft.UI.Xaml.Controls
 			}
 		}
 
-		// Tracks the caller-requested offset (the user's intent) separately from VerticalOffset/
-		// HorizontalOffset (the displayed/clamped value). Mirrors WinUI's pattern in
-		// ScrollContentPresenter where m_Offset is the requested offset (validated only against the
-		// CURRENT ScrollableHeight at request time) and m_ComputedOffset is the one returned to
-		// callers (recoerced after every layout pass).
+		// Tracks the caller-requested offset (the user's intent), bounded at request time, separately from
+		// VerticalOffset/HorizontalOffset (the value displayed after coercion against the live extent).
 		// Cleared on user input (mouse wheel, touch press) so wheel-driven scrolling never gets
 		// auto-pushed back toward a stale programmatic intent.
 		private double? _verticalOffsetIntent;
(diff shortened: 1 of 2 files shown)


ajpinedam (approved):

MartinZikmund:
## Review comments addressed

Pushed commit `034c5240f1b0dbb778e680c1cac8c10f15650f2f` to `dev/mazi/extent`. The four open threads reduce to two distinct issues (each reported by two bots), both in `Given_ScrollViewer_OffsetIntent.cs`. Both are now fixed.

### Fixed (4)

|---|---|---|
| `Given_ScrollViewer_OffsetIntent.cs:16` (copilot) | Unused usings will trip CS8019 under `TreatWarningsAsErrors`. | Removed the five unused usings — `System.Collections.Generic`, `System.Linq`, `Microsoft.UI.Xaml.Media`, `Windows.Foundation`, `Windows.UI`. Kept `System` (Math.Abs), `System.Collections.ObjectModel` (ObservableCollection), `System.Threading.Tasks`, the three `Microsoft.UI.Xaml*`, `Microsoft.VisualStudio.TestTools.UnitTesting`, `Private.Infrastructure`, and `Uno.UI.RuntimeTests.Helpers`. |
| `Given_ScrollViewer_OffsetIntent.cs:16` (github-actions) | Same five unused usings; build-breaking. | Same removal — landed on exactly the keep-list the bot proposed. |
| `Given_ScrollViewer_OffsetIntent.cs:171` (github-actions) | Prefer `[PlatformCondition]` over `#if` + `[Ignore]` so the skip reason/platforms show in test metadata. | Dropped the `#if __ANDROID__ \|\| __IOS__ \|\| __WASM__` / `[Ignore]` block and merged the native skips into the existing attribute: `[PlatformCondition(ConditionMode.Exclude, RuntimeTestPlatforms.NativeWinUI \| NativeAndroid \| NativeIOS \| NativeWasm)] // async native scrolling`. |
| `Given_ScrollViewer_OffsetIntent.cs:171` (github-actions) | Same convention point (the additive one-line variant). | Used the merged single-attribute form rather than adding a second `[PlatformCondition]`, since a duplicate attribute would not compile. This covers both line-171 threads. |

Notes on the second fix: the two suggested variants differed — one added a separate `[PlatformCondition]` alongside the existing `NativeWinUI` one, the other merged them. Only the merged form is valid (two `[PlatformCondition]` attributes on one method won't compile), so I used that. The switch from `#if` (which would also gate the Skia mobile/WASM heads) to the `Native*` flags is intentional and matches the test's stated reason — "async native scrolling" is a native-platform concern, and Skia scrolling is managed — consistent with `.claude/rules/runtime-tests.md`.

Changes were reviewed by inspection only; I did not build or run anything. Azure Pipelines CI will verify.

Xiaoy312 (approved):

src/Uno.UI/UI/Xaml/Controls/ScrollViewer/ScrollViewer.cs
```diff
@@ -1494,14 +1480,43 @@ internal void ClearOffsetIntents()
...
+		// Every offset request — ChangeView, the ISCP setters, scrollbar interactions — is stored clamped
+		// to the scrollable extent known at request time, as WinUI does (ScrollViewer::AdjustTargetVerticalOffset
+		// clamps against ComputePixelExtentHeight before ScrollContentPresenter::SetOffsetsWithExtents
+		// stores it). Bounding the stored value is what keeps later extent growth from carrying the offset
+		// past what the caller could reach. Re-coercing it against the live extent every pass
+		// (RecomputeOffsetsFromIntent) is Uno-specific: it restores the offset after the transient
+		// realization shrink that a virtualized StackLayout's extent estimate produces (#23041).
+		internal void SetVerticalOffsetIntent(double offset)
+		{
+			if (TryClampOffsetIntent(offset, ScrollableHeight, out var clamped))
```
morning4coffe-dev, line 1492: [P2] Clamp combined ChangeView requests using the target zoom

ChangeView arms this intent before applying its zoomFactor, so ScrollableHeight/Width still describe the old zoom. With unzoomed extent 400, viewport 200, current zoom 1, ChangeView(null, 400, 2, true) stores an intent of 200 even though 400 is valid in the target zoom's 600-pixel scrollable range. RecomputeOffsetsFromIntent can subsequently pull the offset back to 200.

WinUI's AdjustTargetVerticalOffset explicitly computes unzoomed extent * targetZoomFactor before clamping (and the horizontal path does likewise): https://github.com/microsoft/microsoft-ui-xaml/blob/439bda90582f7892ca595a193f55dbfc6ae2010c/dxaml/xcp/dxaml/lib/ScrollViewer_Partial.cpp#L12230-L12264 . Thread the effective target zoom into intent capture, and add combined zoom/offset regression coverage.

morning4coffe-dev (changes requested):
Request-time bounding is the right direction, but the new clamp misses WinUI's target-zoom behavior for a combined ChangeView request. Also, retain the explicit compatibility note already in the test: restoring the old intent after an extent shrinks and regrows is an intentional Uno-specific behavior, not native WinUI parity.

PR #23728, opened 2026-07-12
fix(Uno.Sdk): wire tvOS single-project + guard RetargetAssets against a missing app manifest

ZachariahAyers:
## Problem

Building a **tvOS single-project** head fails in the asset pipeline with an opaque:

```
error MSB4018: The "RetargetAssets" task failed unexpectedly.
System.ArgumentNullException: Value cannot be null. (Parameter 'url')
   at System.Xml.XmlDocument.Load(String filename)
   at Uno.UI.Tasks.Assets.RetargetAssets.EnumerateFontsFromPList(String iosAppManifest)
```

## Root cause

There are two related issues:

1. **`RetargetAssets.EnumerateFontsFromPList` loads an optional input unconditionally.**
   `IosAppManifest` is declared without `[Required]`, so callers may legitimately pass nothing — but the method does `new XmlDocument().Load(iosAppManifest)` with no guard, throwing on `null`/empty. It's invoked for every `uikit` `TargetPlatform` (iOS, tvOS, macCatalyst).

2. **tvOS single-project has no target wiring an `Info.plist`.**
   `Uno.Sdk` ships `Uno.SingleProject.iOS.targets`, `.MacCatalyst.targets`, `.MacOS.targets` (each wires `Platforms/<head>/Info.plist` + entitlements/assets/privacy), and imports the matching `Uno.Common.<head>.targets` per `TargetPlatformIdentifier`. **There is no tvOS pair**, even though `$(tvOSProjectFolder)` (= `Platforms/tvOS/`) and `$(IsTvOS)` are already defined. So a tvOS head never gets an app manifest — which is exactly what makes (1) throw.

## Fix

- **`RetargetAssets.Fonts.cs`** — return an empty set when there is no manifest to read (no manifest ⇒ no pre-existing fonts to preserve), and null-guard `SelectNodes`. This respects the property's optional contract and turns the crash into a no-op for heads without a manifest.
- **`Uno.SingleProject.tvOS.targets`** (new) — mirrors the iOS/macCatalyst target: wires `$(tvOSProjectFolder)Info.plist` (+ entitlements, `.xcassets`, `PrivacyInfo.xcprivacy`).
- **`Uno.Common.tvOS.targets`** (new) — mirrors `Uno.Common.iOS.targets` (sets `IsTvOS`, a default `SupportedOSPlatformVersion`, imports the SingleProject target).
- **`Uno.Common.targets`** — import `Uno.Common.tvOS.targets` for `TargetPlatformIdentifier == 'tvos'`, alongside the existing per-head imports.

Either change alone stops the crash; together they make tvOS single-project a first-class head like the other Apple targets. The new `.targets` are picked up by the existing `targets\**\*.targets` pack glob.

## Testing

Reproduced and validated downstream on a real single-project app (Uno.Sdk 6.5.x, .NET 10, macOS 26 / Xcode 26.5): before, the tvOS head crashed in `RetargetAssets`; after supplying `Platforms/tvOS/Info.plist` (now wired by the new target), the tvOS head builds past the task and **compiles + native-links for `tvos-arm64`**. iOS and macCatalyst heads are unaffected.

One deliberate omission vs. `Uno.SingleProject.iOS.targets`: the `TrimmerRootAssembly` workaround for `Microsoft.Extensions.DependencyInjection` is not carried over, since I could not verify the underlying issue reproduces on tvOS - easy to add if you know it does.

Happy to adjust the `SupportedOSPlatformVersion` default or split this into two PRs (the task guard vs. the tvOS wiring) if preferred — they're independent but causally linked.

### DIFF
--- a/src/SourceGenerators/Uno.UI.Tasks/Assets/RetargetAssets.Fonts.cs
+++ b/src/SourceGenerators/Uno.UI.Tasks/Assets/RetargetAssets.Fonts.cs
@@ -72,14 +72,23 @@ public partial class RetargetAssets_v0
 
 	private string[] EnumerateFontsFromPList(string iosAppManifest)
 	{
+		// IosAppManifest is optional (not [Required]); some heads (e.g. a tvOS single-project that has no
+		// Info.plist wired) provide no manifest. With nothing to read there are no pre-existing fonts to
+		// preserve, so return an empty set rather than throwing from XmlDocument.Load(null).
+		if (string.IsNullOrEmpty(iosAppManifest) || !File.Exists(iosAppManifest))
+		{
+			return Array.Empty<string>();
+		}
+
 		XmlDocument doc = new();
 		doc.Load(iosAppManifest);
 
 		// Get the list of registered fonts in the info.plist
 		return doc
 			.SelectNodes("//key[text()='UIAppFonts']/following-sibling::array[1]/string")
-			.OfType<XmlNode>()
+			?.OfType<XmlNode>()
 			.Select(n => n.InnerText)
-			.ToArray();
+			.ToArray()
+			?? Array.Empty<string>();
 	}
 }
--- a/src/Uno.Sdk/targets/Uno.Common.targets
+++ b/src/Uno.Sdk/targets/Uno.Common.targets
@@ -79,6 +79,8 @@
 		Condition=" $([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'ios' " />
 	<Import Project="Uno.Common.MacCatalyst.targets"
 		Condition=" $([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'maccatalyst' " />
+	<Import Project="Uno.Common.tvOS.targets"
+		Condition=" $([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'tvos' " />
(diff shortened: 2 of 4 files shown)


morning4coffe-dev (approved):
The tvOS import, platform identity and single-project manifest/resource/entitlement items follow the existing Apple-target scheme. The optional-manifest guard avoids loading a nonexistent input, while malformed supplied XML still surfaces as an error rather than being broadly swallowed. This is Uno SDK/platform packaging, not a WinUI UI contract. The branch conflicts and platform-floor alignment with the current base remain integration gates.

ZachariahAyers:
@morning4coffe-dev thank you for the review. Rebased on master (`eb70172ff`) and cut down to what master does not already have:

- The tvOS single-project wiring is gone from this branch. master carries `Uno.Common.tvOS.targets`, `Uno.SingleProject.tvOS.targets` and the import in `Uno.Common.targets` since `0a3d18224e`, with the 15.0 floor and the MEDI trimming workaround, so the platform-floor point is settled by using master's files as they are.
- What remains is the `RetargetAssets` guard alone: on master `EnumerateFontsFromPList` still calls `XmlDocument.Load(iosAppManifest)` unconditionally while `IosAppManifest` is optional, so a UIKit head with no `Info.plist` still fails the way the description shows.

Conflicts are cleared; +11/-2 in one file.

PR #23745, opened 2026-07-13
refactor(window)!: Retype WindowActivationState to WinUI (BC13)

Xiaoy312:
**GitHub Issue:** #8339

## PR Type:

💬 Other... Breaking change (API alignment with WinUI)

## What changed? 🚀

`Microsoft.UI.Xaml.WindowActivatedEventArgs.WindowActivationState` was typed as the Uno-only **`Windows.UI.Core.CoreWindowActivationState`**. WinUI types it as **`Microsoft.UI.Xaml.WindowActivationState`**. This PR (**BC13** of the Phase 5 breaking-changes wave) retypes the property (and the internal constructor) to the WinUI enum.

The two enums are value-identical (`CodeActivated=0`, `Deactivated=1`, `PointerActivated=2`), so no runtime behaviour changes — only the public type.

Details:
- **New hand-written enum** `Microsoft.UI.Xaml.WindowActivationState` (`src/Uno.UI/UI/Xaml/Window/WindowActivationState.cs`), available on **all** platforms. The previous definition was a generated stub gated to `__SKIA__ || __NETSTD_REFERENCE__` only, so it couldn't back a cross-platform property. Following the "move out of `Generated/`" convention, the enum is now hand-written; the generated stub is set to the disabled `#if false` form the sync generator emits for hand-written enums (mirrors `ElementTheme`/`FocusState`).
- **`WindowActivatedEventArgs`** — property + internal ctor retyped.
- **Producers** (`BaseWindowImplementation`, `Window.alc.cs`) — the internal/native plumbing stays on `CoreWindowActivationState`; the value is cast to `WindowActivationState` at the public-args boundary (value-preserving).
- **Consumers** — `WindowChrome`, `DatePicker` (Uno-specific activation hook), `MacOSWindowHost`, the `Given_AlcContentHost` ALC tests, and the `WindowActivationTests` sample updated to the new type / value-preserving cast.
- **Parity shim removed** — the `SymbolMatchingHelpers` special-case that forced the WinUI-parity check to pass for this property (`// TODO: Match API with WinUI`) is deleted; the check now validates the real (matching) shape.

**Validation:**
- **Compile:** `SamplesApp.Skia.Generic` (net10.0 Skia) builds clean, transitively compiling `Uno.UI`, `Uno.UI.RuntimeTests`, samples, and the Skia runtimes (incl. `Uno.UI.Runtime.Skia.MacOS`). Reference build validated separately (the API-contract compilation). Native heads are safe **by construction** — the enum is now defined unconditionally, so no native head loses the type. (Native Android/iOS not built locally; stated, not implied.)
- **Runtime (Skia Desktop):** `Given_AlcContentHost` activation tests pass — they raise `Window.Activated`, read `args.WindowActivationState`, and assert `CodeActivated`, exercising the producer cast end-to-end. Value-identical retype ⇒ pass-after regression guards.
- The `Uno.WinAppSDKSyncGenerator` project targets .NET 11 (SDK not available locally); the shim removal is a mechanical block deletion — CI's parity gate validates it.

`PackageDiffIgnore.xml` (6.5 baseline) gets one entry for the retyped getter (`get_WindowActivationState()` return type changed ⇒ old signature reads as removed).

## PR Checklist ✅

- [ ] 🧪 Added Runtime tests, UI tests, or a manual test sample — **N/A**: not a behaviour change; existing `Given_AlcContentHost` activation tests exercise the path and were run as regression guards.
- [ ] 📚 Docs have been added/updated — **N/A** (migration note below).
- [ ] 🖼️ Validated PR `Screenshots Compare Test Run` results — **N/A** (no visual change).
- [ ] ❗ Contains **NO** breaking changes — intentionally left unchecked; this **is** a breaking change (see below).

### Breaking change — impact & migration path

- **Impact (source-breaking):** `WindowActivatedEventArgs.WindowActivationState` now returns `Microsoft.UI.Xaml.WindowActivationState` instead of `Windows.UI.Core.CoreWindowActivationState`. Code that assigned it to a `CoreWindowActivationState` variable or compared against `CoreWindowActivationState.*` members no longer compiles.
- **Migration:** Use `Microsoft.UI.Xaml.WindowActivationState` and its `CodeActivated`/`Deactivated`/`PointerActivated` members. The enum values are unchanged, so a value-preserving cast (`(CoreWindowActivationState)args.WindowActivationState`) bridges any code that must stay on the legacy type. This matches WinUI.

### DIFF
--- a/build/PackageDiffIgnore.xml
+++ b/build/PackageDiffIgnore.xml
@@ -2546,6 +2546,9 @@
 		  <Member fullName="Microsoft\.UI\.Xaml\.DependencyProperty .+(::|\.)TemplatedParentProperty\(\)" reason="DependencyObject as class: member inherited from base" isRegex="true" />
 		  <!-- ColorKeyFrameCollection carried a binary-compat 'new'-shadow block (Size/Count/IsReadOnly/indexer) that its zero-shadow siblings Double/ObjectKeyFrameCollection never had; the get-only base DependencyObjectCollection<T> members show through with identical behavior. Accessor methods are paired in <Methods> below. -->
 		  <Member fullName=".+ Microsoft\.UI\.Xaml\.Media\.Animation\.ColorKeyFrameCollection(::|\.)(Size|Count|IsReadOnly|Item)\(.*\)" reason="Removed ColorKeyFrameCollection binary-compat shadow members; provided by base DependencyObjectCollection&lt;T&gt; (BC67, breaking changes for 7.0)" isRegex="true" />
+
+		  <!-- WindowActivatedEventArgs.WindowActivationState retyped from the Uno-only Windows.UI.Core.CoreWindowActivationState to the WinUI Microsoft.UI.Xaml.WindowActivationState; the old-typed property reads as removed (getter accessor paired in <Methods> below). -->
+		  <Member fullName="Windows.UI.Core.CoreWindowActivationState Microsoft.UI.Xaml.WindowActivatedEventArgs::WindowActivationState()" reason="Retyped WindowActivationState to Microsoft.UI.Xaml.WindowActivationState (BC13, breaking changes for 7.0)" />
 	  </Properties>
 
 	  <Methods>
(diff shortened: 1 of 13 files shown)


MartinZikmund (approved):

MartinZikmund:
@Xiaoy312 CI has some issue, but otherwise good to go

PR #23796, opened 2026-07-16
fix(skia): Native elements in Win32 never receive focus (WS_EX_LAYERED)

ramezgerges:
**GitHub Issue:** closes https://github.com/unoplatform/kahua-private/issues/518

## PR Type:

🐞 Bugfix

## What changed? 🚀

On Skia Desktop (Win32), a hosted native element such as a `WebView2` never accepted keyboard/pointer focus, did not register clicks, and showed the wrong (stuck) cursor instead of its own (e.g. the I-beam over a text field). This was most reliably reproduced with a `WebView2` inside a `ContentDialog`.

`Win32NativeElementHostingExtension.AttachNativeElement` applied `WS_EX_LAYERED` unconditionally to every hosted native child window (to support opacity via `SetLayeredWindowAttributes`). A layered *child* window is excluded from the OS's normal hit-testing and message routing — `WindowFromPoint`, mouse buttons, `WM_SETCURSOR`, etc. all resolve to the parent instead — so the hosted content never received input.

This PR:
- Stops setting `WS_EX_LAYERED` on hosted native windows in `AttachNativeElement`, and **actively clears it** (covering a window re-attached after a prior attach set it).
- Makes `ChangeNativeElementOpacity` a no-op: per-window alpha for a classic Win32 child window is only possible via `WS_EX_LAYERED`, which is exactly what breaks input. Native-element fractional opacity is therefore unsupported on Win32 — the same stance the X11 head already takes.
- Adds a manual repro sample (`WebView → WebView2_ContentDialogFocus`).

> A runtime test was prototyped but removed: reliably asserting the hosted child's ex-style requires spawning an external native window (`CreateSampleComponent` → powershell/Windows Terminal), which carries its *own* `WS_EX_LAYERED` and produces false positives on CI. The manual sample covers the scenario instead.

## PR Checklist ✅

- [x] 🧪 Added Runtime tests, UI tests, or a manual test sample (for bug fixes / features, if applicable)

### DIFF
--- a/src/Uno.UI.Runtime.Skia.Win32/Native/Win32NativeElementHostingExtension.cs
+++ b/src/Uno.UI.Runtime.Skia.Win32/Native/Win32NativeElementHostingExtension.cs
@@ -90,7 +90,13 @@ internal class Win32NativeElementHostingExtension : ContentPresenter.INativeElem
 
 		var oldExStyleVal = PInvoke.GetWindowLong((HWND)window.Hwnd, WINDOW_LONG_PTR_INDEX.GWL_EXSTYLE);
 		var oldStyleVal = PInvoke.GetWindowLong((HWND)window.Hwnd, WINDOW_LONG_PTR_INDEX.GWL_STYLE);
-		PInvoke.SetWindowLong((HWND)window.Hwnd, WINDOW_LONG_PTR_INDEX.GWL_EXSTYLE, oldExStyleVal | (int)WINDOW_EX_STYLE.WS_EX_LAYERED);
+		// WS_EX_LAYERED is intentionally never set — and actively cleared — on the hosted child window. It is
+		// the only way to give a classic Win32 child window per-window alpha, but a layered child is excluded
+		// from the OS's normal hit-testing and input routing (WindowFromPoint, mouse buttons, WM_SETCURSOR all
+		// resolve to the parent), which prevents hosted content such as WebView2 from receiving focus/clicks.
+		// Clearing it also covers a window that arrives already layered (e.g. re-attached after a prior attach
+		// set it). Native element opacity is therefore unsupported here (as on X11); see ChangeNativeElementOpacity.
+		PInvoke.SetWindowLong((HWND)window.Hwnd, WINDOW_LONG_PTR_INDEX.GWL_EXSTYLE, oldExStyleVal & ~(int)WINDOW_EX_STYLE.WS_EX_LAYERED);
 		PInvoke.SetWindowLong((HWND)window.Hwnd, WINDOW_LONG_PTR_INDEX.GWL_STYLE, (oldStyleVal | (int)WINDOW_STYLE.WS_CLIPSIBLINGS) & ~(int)WINDOW_STYLE.WS_CAPTION); // removes the title bar and borders
 
 		_ = PInvoke.ShowWindow((HWND)window.Hwnd, SHOW_WINDOW_CMD.SW_HIDE);
@@ -458,13 +464,10 @@ internal class Win32NativeElementHostingExtension : ContentPresenter.INativeElem
 
 	public void ChangeNativeElementOpacity(object content, double opacity)
 	{
-		if (content is not Win32NativeWindow window)
-		{
-			throw new ArgumentException($"content is not a {nameof(Win32NativeWindow)} instance.", nameof(content));
-		}
-
-		var success = PInvoke.SetLayeredWindowAttributes((HWND)window.Hwnd, new COLORREF(0), (byte)Math.Round(opacity * 255), LAYERED_WINDOW_ATTRIBUTES_FLAGS.LWA_ALPHA);
-		if (!success) { this.LogError()?.Error($"{nameof(PInvoke.SetLayeredWindowAttributes)} failed: {Win32Helper.GetErrorMessage()}"); }
+		// Native element opacity is not supported on Win32. Per-window alpha for a classic Win32 child
+		// window requires WS_EX_LAYERED, which excludes the child from the OS's normal hit-testing and
+		// input routing and thus breaks focus/clicks for hosted content such as WebView2 (see
+		// AttachNativeElement). X11 likewise does not support opacity for hosted child windows.
 	}
 
 	public bool SupportsZIndex() => true;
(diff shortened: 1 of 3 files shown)


ajpinedam (approved):

Xiaoy312 (approved):

src/Uno.UI.Runtime.Skia.Win32/Native/Win32NativeElementHostingExtension.cs
```diff
@@ -90,7 +90,13 @@ public void AttachNativeElement(object content)
...
 		var oldExStyleVal = PInvoke.GetWindowLong((HWND)window.Hwnd, WINDOW_LONG_PTR_INDEX.GWL_EXSTYLE);
 		var oldStyleVal = PInvoke.GetWindowLong((HWND)window.Hwnd, WINDOW_LONG_PTR_INDEX.GWL_STYLE);
-		PInvoke.SetWindowLong((HWND)window.Hwnd, WINDOW_LONG_PTR_INDEX.GWL_EXSTYLE, oldExStyleVal | (int)WINDOW_EX_STYLE.WS_EX_LAYERED);
+		// WS_EX_LAYERED is intentionally never set — and actively cleared — on the hosted child window. It is
+		// the only way to give a classic Win32 child window per-window alpha, but a layered child is excluded
+		// from the OS's normal hit-testing and input routing (WindowFromPoint, mouse buttons, WM_SETCURSOR all
+		// resolve to the parent), which prevents hosted content such as WebView2 from receiving focus/clicks.
+		// Clearing it also covers a window that arrives already layered (e.g. re-attached after a prior attach
+		// set it). Native element opacity is therefore unsupported here (as on X11); see ChangeNativeElementOpacity.
+		PInvoke.SetWindowLong((HWND)window.Hwnd, WINDOW_LONG_PTR_INDEX.GWL_EXSTYLE, oldExStyleVal & ~(int)WINDOW_EX_STYLE.WS_EX_LAYERED);
```
dr1rrb, line 99: Did you consider applying `WS_EX_LAYERED` lazily instead of dropping opacity support outright? I.e. keep it cleared at attach (as done here), and in `ChangeNativeElementOpacity`: set the style + `SetLayeredWindowAttributes(alpha)` when `opacity < 1`, and clear the style again when opacity returns to 1.

That keeps input intact in the common fully-opaque case while preserving opacity for hosts that can render under it. The clear-on-return-to-1 part is critical: `UpdateNativeHostContentPresentersOpacities` multiplies ancestor opacities, so e.g. a `ContentDialog` entrance fade would transiently set the style on the hosted WebView2, and it must come off afterwards or the focus bug comes back.

If this was tried and rejected (e.g. because WebView2 doesn't render/behave correctly under a layered window anyway), it would be worth capturing that in the PR description for posterity.

src/Uno.UI.Runtime.Skia.Win32/Native/Win32NativeElementHostingExtension.cs
```diff
@@ -458,13 +464,10 @@ public unsafe object CreateSampleComponent(string text)
...
-		{
-			throw new ArgumentException($"content is not a {nameof(Win32NativeWindow)} instance.", nameof(content));
-		}
-
-		var success = PInvoke.SetLayeredWindowAttributes((HWND)window.Hwnd, new COLORREF(0), (byte)Math.Round(opacity * 255), LAYERED_WINDOW_ATTRIBUTES_FLAGS.LWA_ALPHA);
-		if (!success) { this.LogError()?.Error($"{nameof(PInvoke.SetLayeredWindowAttributes)} failed: {Win32Helper.GetErrorMessage()}"); }
+		// Native element opacity is not supported on Win32. Per-window alpha for a classic Win32 child
+		// window requires WS_EX_LAYERED, which excludes the child from the OS's normal hit-testing and
+		// input routing and thus breaks focus/clicks for hosted content such as WebView2 (see
+		// AttachNativeElement). X11 likewise does not support opacity for hosted child windows.
```
dr1rrb, line 470: The silent no-op worries me a bit: someone setting `Opacity=0.5` on a hosted native element (or fading an ancestor) now gets nothing, with no clue why. A `LogDebug`/`LogWarning` along the lines of "native element opacity is not supported on Win32" would save that user hours of head-scratching. Dropping the `ArgumentException` type check is fine for a no-op, but a log costs nothing.

Also note this doesn't only affect WebView2: `Win32MediaPlayerPresenterExtension` hosts the VLC surface through the same mechanism (`ContentPresenter { Content = new Win32NativeWindow(_hwnd) }`), so any opacity/fade over a `MediaPlayerElement` lands here too. In practice I suspect opacity never actually worked there either — VLC renders with D3D straight into the HWND, which largely bypasses the layered-window redirection — but it would be good to validate that on master (e.g. `Opacity=0.5` on a playing `MediaPlayerElement`) before concluding nothing is lost.

dr1rrb (commented):
Makes sense overall — the diagnosis is right and input working clearly beats fractional opacity on a native host. A few non-blocking comments below.

One more thing: since this is a behavioral divergence from WinUI (native element opacity silently unsupported on Win32 Skia Desktop), it deserves a line in the Skia Desktop limitations docs — the docs checkbox in the PR template is unchecked.

src/Uno.UI.Runtime.Skia.Win32/Native/Win32NativeElementHostingExtension.cs
ramezgerges: doing it lazily is fragile as also doesn't results in the real issue of partially transparent windows not receiving input.

agneszitte, line 99: 🔀 **CI: rebase/merge latest `master` to clear the failing Analyze (CodeQL) check**

This branch is ~17 commits behind `master` and still pins `System.Security.Cryptography.Xml` `8.0.3` in `Uno.HotReload.csproj` and `Uno.UI.RemoteControl.Server.Processors.csproj`, which is what the failing **Analyze (csharp)** security scan is flagging.

`master` already bumped it to `10.0.10` via #23830 (merged 2026-07-21, commit `6de04d0`). 
Note: #23828 was the equivalent fix but it targeted `feature/breakingchanges`, not `master`, so #23830 is the one that applies here.

**Action:** merge (or rebase onto) latest `master`,  that pulls in the bump and should also refresh the other stale build/test jobs.

@ramezgerges you can resolve this thread once the branch is updated, thanks in advance.

PR #23804, opened 2026-07-17
fix: re-create unmaterialized Frame content on hr element update

rajamatt:
**GitHub Issue:** related to https://github.com/unoplatform/uno.extensions/issues/3130

## PR Type:

🐞 Bugfix

## What changed? 🚀

**Current behavior:** the hot-reload UI-update phase replaces live instances of updated types by walking the visual tree (`VisualTreeHelper` children only). Content assigned to a `ContentControl` whose subtree never got a layout pass — e.g. the host sits under a collapsed ancestor, or the hosting view is not composited yet — exists only as the `Content` property: the control's template never applied, so the content is never materialized as a visual child. The walk cannot see it, so the stale pre-update instance is what gets shown once the subtree finally lays out. The most common shape is a `Page` navigated into a `Frame` (`FrameElementMetadataUpdateHandler` already re-creates page instances for the back/forward stack, but not the current page), but any `ContentControl` hosting an updated view type has the same blind spot.

**New behavior:** stranded content is re-created by the element-update handlers, for all ContentControls:

- New `ContentControlElementMetadataUpdateHandler` (registered for `ContentControl`) owns the shared detection/re-creation (`CreateReplacementForStrandedContent`): only when the content is a `FrameworkElement` (data-item content rendered through a `ContentTemplate` is never touched), only when it is **not** materialized (`VisualTreeHelper.GetParent` is null — a materialized element is still handled by the visual-tree walk itself, so there is no double replacement), and only when its type was updated — either a CNOMU replacement type (`GetReplacementType()`) or an in-place EnC update where the type identity is unchanged but present in the updated list. Local values are transferred with the walk's own `ClonePropertiesToAnotherStoreForHotReload`, so an inherited DataContext keeps flowing from the host.
- `FrameElementMetadataUpdateHandler` reuses that shared helper and adds the Frame-specific part: `newPage.Frame` wiring and keeping the navigation history consistent (the current `PageStackEntry`'s `Instance`/`SourcePageType` — explicitly in WinUI-behavior mode; legacy mode syncs through `OnContentChanged`). The ContentControl handler skips Frames so the two handlers never double-patch the same instance.
- Known not-covered (follow-up material): standalone `ContentPresenter` (not a `ContentControl`) and unrealized `ItemsControl` containers have the same theoretical blind spot; no known repro for either.

**Review feedback addressed:**

- Content re-creation now goes through `Frame.CreatePageInstance` (the centralized creation path also used by `NavigationCache` and `PagePool`) instead of a raw `Activator.CreateInstance`, so hot-reload re-creation matches normal navigation — including `BindableMetadataProvider` support and the replacement-type indirection. Despite its name the helper is type-agnostic, which is what the shared ContentControl path needs.
- Consequently, the parameterless-constructor pre-check was removed: a bindable metadata activator may construct types a reflection probe would reject. Creation failure is handled instead of predicted.
- **State transfer now reuses the walk's `ClonePropertiesToAnotherStoreForHotReload`** instead of a hand-rolled DataContext copy. That helper moves **local** values only, and only where the new instance does not already have one, so page-XAML values win, an inherited DataContext keeps flowing from the host, and local bindings carry over as expressions rather than pinned resolved values. It also stops silently dropping attached properties — navigation frameworks set region/route properties on content right after creating it, so that loss was real rather than theoretical. The previous thin copy was inherited from the pre-existing back-stack re-creation loop rather than a deliberate decision for this path, so it is replaced rather than rationalized.
- **Creation failures are diagnosable**: both the exception path and a null result from `CreatePageInstance` log at Warning (with the exception) and keep the stale instance — "hot reload silently kept stale UI" should never require a debugger to notice.
- **`SetContent` runs before the history-entry sync** in the Frame handler, so a failed content assignment cannot leave the navigation history pointing at a page that was never displayed.
- Tests: back-navigation round-trip added (pins the history-entry sync — landing back on the stale instance would show pre-update text); negative test added asserting stranded content of a **non-updated** type keeps its identity and data-item content is never touched; unused usings removed; `CancellationTokenSource` disposed via `using var`; `Assert.AreNotSame` for identity assertions.

**Design notes (scope and semantics):**

- **Scope**: the handler intentionally covers every `ContentControl` subclass — the blind spot is structural (template-mediated content), not Frame-specific. The guards keep it inert for the common cases: data-item content (most Buttons/items) exits on the first check, materialized content on the second.
- **Eager re-creation** (rather than deferring the swap to materialization) deliberately mirrors the visual-tree walk's semantics for materialized instances of updated types: replace at update time. The discarded instance was never loaded, so no `Loaded`/`Unloaded` pairing is broken; references held by outer code go stale exactly as they already do when the walk replaces a materialized instance.
- **Trigger parity**: re-creating on `liveType ∈ updatedTypes` (in-place EnC) in addition to CNOMU mappings matches the walk's own replacement trigger for materialized elements.
- **Replacement notifications are intentionally not raised** for this swap, and it is not counted in the operation's replaced-element total. `Before`/`AfterElementReplaced` are the walk's contract for elements it owns; raising them from a handler would mean either exposing the agent's handler actions to `ElementUpdate` or moving this patch into the agent's dispatch, i.e. giving up the layer this fix deliberately sits at. Impact was checked rather than assumed: `Page.HotReload.cs` is the only in-box handler implementing either method, and its whole body (`newPage.Frame = oldPage.Frame` plus re-assigning `Frame.Content`) is already what `PatchStrandedContent` does — so notifying it here would be a no-op. The count feeds only `TotalElementCount` on the client-operation event; nothing branches on it, and HR success is not count-derived. What remains is third-party handlers registered for a replaced content type, which is why this is documented rather than left implicit. `AfterVisualTreeUpdate` already exists for exactly this — a once-per-phase reconciliation point — and is what a consumer tracking a content instance should use. Worth noting the downstream navigation fix needs that hook regardless of notifications: it also has to recover from `FrameNavigator` tracking a dead instance (no `Frame.Navigated` is raised on either replacement path), rebuild the view model (an in-place EnC update keeps type identity, so the "wrong VM type" check sees the stale VM as valid), and re-parent nested regions. Documented in `doc/articles/uno-development/uno-internals-hotreload.md`.
- **`CaptureState`/`RestoreState` is not run** for this content: it was never materialized, so there is no visual state to preserve, and the walk keys its capture on a visual-tree position this content does not have. Also documented.
- **Not covered (follow-ups)**: standalone `ContentPresenter` and unrealized `ItemsControl` containers (same theoretical blind spot, no repro); nested stranded content of a *non-updated* type containing an updated child (one-level scope, same as what the walk can reach today); page-cache eviction under CNOMU (pre-existing "does not support CNOMUA" note on the cache-removal path).

**Red/green coverage:**

- `Given_Frame_StrandedContent.When_Page_Updated_While_Unmaterialized_Then_Content_Recreated` — navigates a page into a frame under a collapsed ancestor (asserting the live-but-unmaterialized precondition), applies a server-side XAML update, and verifies the stranded page is re-created and shows the updated content once the subtree is revealed. Red without the fix, green with it.
- `Given_ContentControl_StrandedContent.When_Content_Updated_While_Unmaterialized_Then_Content_Recreated` — same scenario for a plain `ContentControl` hosting a `UserControl`.
- `Given_ContentControl_StrandedContent.When_Content_Updated_While_Unmaterialized_Then_Local_Values_Are_Transferred` — a locally-set value (`Tag`) and a locally-set attached property (`AutomationProperties.Name`) survive the re-creation, while an inherited DataContext is not pinned as a local value on the new instance.
- Also verified end-to-end by a deterministic runtime test in uno.extensions (`Given_HotReload.When_PageXamlUpdatedWhileUnmaterialized_Then_RevealShowsUpdatedContent`): red on current packages, green with this change, and no regressions across the navigation + TabBar hot-reload runtime suites there (29 tests).

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/ContentControl/ContentControl.HotReload.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/ContentControl/ContentControl.HotReload.cs
@@ -16,9 +16,8 @@ internal static partial class ContentControlElementMetadataUpdateHandler
 {
 	public static void ElementUpdate(FrameworkElement element, Type[] updatedTypes)
 	{
-		// Frame has its own element-update handler which additionally keeps the navigation
-		// history in sync with the re-created page — let it own its content patching
-		// entirely so the two handlers never double-replace the same instance.
+		// Frame's own handler patches its content and syncs the navigation history;
+		// skipping it here prevents the two handlers from double-replacing.
 		if (element is not ContentControl contentControl || contentControl is Frame)
 		{
 			return;
@@ -31,30 +30,22 @@ internal static partial class ContentControlElementMetadataUpdateHandler
 	}
 
 	/// <summary>
-	/// Creates a replacement for the host's current content when that content's type was
-	/// hot-reloaded but the content was never materialized in the visual tree — e.g. the
-	/// host lives under an ancestor that has not had a layout pass, so its template was
-	/// never applied and the content exists only as <see cref="ContentControl.Content"/>.
-	/// The hot-reload visual-tree walk enumerates materialized children only, so it cannot
-	/// replace such an instance itself; without this, the stale pre-update content is what
-	/// gets displayed once the subtree is finally materialized.
-	/// Returns <see langword="null"/> when there is nothing to patch (data-item content,
-	/// materialized content, type not updated, or no parameterless constructor).
+	/// Re-creates content whose type was hot-reloaded while the content was never
+	/// materialized in the visual tree (host not laid out, template never applied) —
+	/// the HR visual-tree walk only enumerates materialized children and cannot
+	/// replace such an instance itself. Returns null when there is nothing to patch.
 	/// </summary>
 	[UnconditionalSuppressMessage("Trimming", "IL2072")]
-	// Hot reload is only expected to work when trimming is disabled; TypeMappings
-	// preserves the replacement types' constructors.
+	// Hot reload requires trimming disabled; TypeMappings preserves replacement ctors.
 	internal static FrameworkElement? CreateReplacementForStrandedContent(ContentControl host, Type[] updatedTypes)
 	{
-		// Content can be a plain data item rendered through a ContentTemplate — only view
-		// instances can be re-created here.
+		// Data-item content rendered through a ContentTemplate is never re-created.
 		if (host.Content is not FrameworkElement currentContent)
 		{
 			return null;
 		}
 
-		// Materialized content is enumerated (and replaced) by the hot-reload visual-tree
-		// walk itself — patching it here as well would double-replace it.
+		// Materialized content is replaced by the visual-tree walk itself.
 		if (VisualTreeHelper.GetParent(currentContent) is not null)
 		{
 			return null;
@@ -63,8 +54,7 @@ internal static partial class ContentControlElementMetadataUpdateHandler
 		var liveType = currentContent.GetType();
 		var expectedType = liveType.GetReplacementType();
 
-		// CNOMU update: the live type maps to a replacement type.
-		// In-place (EnC) update: same type identity, but present in the updated list.
+		// Same triggers as the walk: a CNOMU replacement type or an in-place (EnC) update.
 		var isUpdated = expectedType != liveType || Array.IndexOf(updatedTypes, liveType) >= 0;
 		if (!isUpdated)
 		{
@@ -80,22 +70,16 @@ internal static partial class ContentControlElementMetadataUpdateHandler
 		FrameworkElement? newContent;
 		try
 		{
-			// Despite its name, Frame.CreatePageInstance is the centralized, type-agnostic
-			// hot-reload-aware creation path (replacement-type resolution + bindable metadata
-			// provider support) also used by NavigationCache and PagePool. No constructor
-			// pre-check here: a bindable metadata activator may construct types a reflection
-			// probe would reject, so creation failure is handled instead of predicted.
+			// Centralized HR-aware creation (replacement-type resolution + bindable
+			// metadata provider), also used by NavigationCache and PagePool.
 			newContent = Frame.CreatePageInstance(liveType) as FrameworkElement;
 		}
 		catch (Exception e)
 		{
-			// The replacement type cannot be constructed (e.g. no parameterless constructor
-			// and no bindable activator) — keep the stale content rather than failing the
-			// hot-reload pass; the visual-tree walk applies the same tolerance.
-			if (typeof(ContentControlElementMetadataUpdateHandler).Log().IsEnabled(LogLevel.Debug))
+			if (typeof(ContentControlElementMetadataUpdateHandler).Log().IsEnabled(LogLevel.Warning))
 			{
-				typeof(ContentControlElementMetadataUpdateHandler).Log().Debug(
-					$"Could not re-create stranded content {liveType.Name} as {expectedType.Name}: {e.Message}");
+				typeof(ContentControlElementMetadataUpdateHandler).Log().Warn(
+					$"Could not re-create stranded content {liveType.Name} as {expectedType.Name}; keeping the stale instance.", e);
 			}
 
 			return null;
@@ -103,13 +87,18 @@ internal static partial class ContentControlElementMetadataUpdateHandler
 
 		if (newContent is null)
 		{
+			if (typeof(ContentControlElementMetadataUpdateHandler).Log().IsEnabled(LogLevel.Warning))
+			{
+				typeof(ContentControlElementMetadataUpdateHandler).Log().Warn(
+					$"Could not re-create stranded content {liveType.Name} as {expectedType.Name} (no instance produced); keeping the stale instance.");
+			}
+
 			return null;
 		}
 
-		// Carry the DataContext over only when the old content had its own value —
-		// an inherited DataContext must keep flowing from the host rather than being
-		// pinned onto the new instance.
-		if (!ReferenceEquals(currentContent.DataContext, host.DataContext))
+		// Copy the DataContext only when it was locally set; an inherited value must
+		// keep flowing from the host instead of being pinned on the new instance.
+		if (currentContent.ReadLocalValue(FrameworkElement.DataContextProperty) != DependencyProperty.UnsetValue)
 		{
 			newContent.DataContext = currentContent.DataContext;
 		}
--- a/src/Uno.UI/UI/Xaml/Controls/ContentControl/ContentControl.HotReload.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/ContentControl/ContentControl.HotReload.cs
@@ -1,0 +1,96 @@
+#if HAS_UNO
+#nullable enable
+
+using System;
+using System.Diagnostics.CodeAnalysis;
+using System.Reflection.Metadata;
+using Microsoft.UI.Xaml.Media;
+using Uno.Foundation.Logging;
+using Uno.UI.Helpers;
+
+[assembly: ElementMetadataUpdateHandlerAttribute(typeof(Microsoft.UI.Xaml.Controls.ContentControl), typeof(Microsoft.UI.Xaml.Controls.ContentControlElementMetadataUpdateHandler))]
+
+namespace Microsoft.UI.Xaml.Controls;
+
+internal static partial class ContentControlElementMetadataUpdateHandler
+{
+	public static void ElementUpdate(FrameworkElement element, Type[] updatedTypes)
+	{
+		// Frame has its own element-update handler which additionally keeps the navigation
+		// history in sync with the re-created page — let it own its content patching
+		// entirely so the two handlers never double-replace the same instance.
+		if (element is not ContentControl contentControl || contentControl is Frame)
+		{
+			return;
+		}
+
+		if (CreateReplacementForStrandedContent(contentControl, updatedTypes) is { } newContent)
+		{
+			contentControl.Content = newContent;
+		}
+	}
+
+	/// <summary>
+	/// Creates a replacement for the host's current content when that content's type was
+	/// hot-reloaded but the content was never materialized in the visual tree — e.g. the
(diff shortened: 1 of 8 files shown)


autocarl:
@dr1rrb, could you please give this PR a review before it is merged? Thanks!

src/Uno.UI/UI/Xaml/Controls/Frame/Frame.HotReload.cs
```diff
@@ -68,6 +71,39 @@ public static void ElementUpdate(FrameworkElement element, Type[] updatedTypes)
...
+		/// </summary>
+		private static void PatchStrandedContent(Frame frame, Type[] updatedTypes)
+		{
+			if (ContentControlElementMetadataUpdateHandler.CreateReplacementForStrandedContent(frame, updatedTypes) is not Page newPage)
+			{
+				return;
+			}
+
+			newPage.Frame = frame;
+			frame.SetContent(newPage);
```
kazo0, line 92: **Replacement-notification contract — should be settled before merge.**

When the visual-tree walk replaces a materialized instance, `ReplaceViewInstance` brackets the swap with `BeforeElementReplaced` / `AfterElementReplaced` on the registered element handlers and reports it via `hrOp.ReportElementReplaced()`. This path does neither. That gives the platform two replacement mechanisms with different observability: handlers registered for the replaced type (e.g. `Page.HotReload.cs`, or downstream navigation frameworks) are never told the instance changed, and the HR operation's element-replaced telemetry under-reports.

The consequence is already visible downstream: unoplatform/uno.extensions#3131 has to detect this swap by reference-diffing `Frame.Content` from `AfterVisualTreeUpdate` instead of listening on `AfterElementReplaced` — and every other consumer that tracks a content instance will need the same workaround.

Can we decide deliberately which contract we're shipping?

- **(a)** The stranded re-creation raises `Before/AfterElementReplaced` like the walk does (would need the agent to expose handler actions to `ElementUpdate`, or the patch to move into the agent's dispatch), so consumers keep a single notification channel; or
- **(b)** "Handler-patched swaps are unobservable; consumers must diff in `AfterVisualTreeUpdate`" is the intended contract — in which case it should be documented in `doc/articles/uno-development/uno-internals-hotreload.md`, since downstream fixes are now being built on top of that behavior.

src/Uno.UI/UI/Xaml/Controls/ContentControl/ContentControl.HotReload.cs
```diff
@@ -0,0 +1,109 @@
...
+
+			return null;
+		}
+
+		// Copy the DataContext only when it was locally set; an inherited value must
+		// keep flowing from the host instead of being pinned on the new instance.
+		if (currentContent.ReadLocalValue(FrameworkElement.DataContextProperty) != DependencyProperty.UnsetValue)
+		{
+			newContent.DataContext = currentContent.DataContext;
+		}
```
kazo0: **State-transfer parity with the walk — should be an explicit decision before merge.**

The walk's `ReplaceViewInstance` transfers the whole DP store to the replacement (`ClonePropertiesToAnotherStoreForHotReload`), carries `BaseUri` / `DebugParseContext`, and brackets the swap with `CaptureState` / `RestoreState`. This path transfers exactly one thing: a locally-set `DataContext`. Every other locally-set value on the stranded instance — `Tag`, local `Visibility`, and notably **attached properties** (e.g. region/route attached properties set by navigation frameworks on the page instance) — is silently dropped, and `CaptureState`/`RestoreState` handlers never run for it.

The downstream compensation already exists: unoplatform/uno.extensions#3131's re-hook re-applies its region-name attached property after this swap. But that only covers that one framework — a plain app that set local values on stranded content loses them, with no equivalent on the materialized-swap path.

If the thin copy is intentional (never-materialized content usually carries little meaningful local state, and a wholesale store clone risks pinning stale values — presumably why the `DataContext` carry-over was made conditional), can that rationale be captured in the comment here and the limitation added to the PR's design notes? If not, should this reuse `ClonePropertiesToAnotherStoreForHotReload` for parity with the walk? Either answer works — it just shouldn't be incidental, given downstream code is already patching around it.

kazo0 (changes requested):
The fix itself looks right — correct layer (the element-update handlers, not the walker), sound guards against double-replacement and data-item content, and solid red/green coverage including the back-navigation consistency case.

Two contract-level questions are anchored inline, and I'd like both answered (and either addressed or explicitly documented as intended behavior) before this merges, since downstream consumers — starting with unoplatform/uno.extensions#3131 — are already building on top of the answers:

1. The stranded-content swap bypasses the `Before/AfterElementReplaced` notification (and element-replaced telemetry) that the visual-tree walk provides for materialized swaps.
2. The stranded-content swap transfers only a locally-set `DataContext`, while the walk clones the full DP store and runs `CaptureState`/`RestoreState` — locally-set values and attached properties on the stranded instance are dropped.

Neither is necessarily a defect — but each should be a deliberate, documented semantic rather than an incidental difference between the two replacement paths.

src/Uno.UI/UI/Xaml/Controls/ContentControl/ContentControl.HotReload.cs
rajamatt: Now uses `ClonePropertiesToAnotherStoreForHotReload`

src/Uno.UI/UI/Xaml/Controls/Frame/Frame.HotReload.cs
rajamatt: I think both options make sense, we should see if @dr1rrb has an opinion on which is better (a) or (b). 

Here's what Opus thinks:

> Going with **(b)** — documented as the intended contract. I wanted to check the actual impact rather than just assert it:
> 
> **In-box:** `Page.HotReload.cs` is the only file in the repo implementing either method, and its whole body is `newPage.Frame = oldPage.Frame` plus re-assigning `Frame.Content`. `PatchStrandedContent` already does both explicitly, so raising the notification here would be a no-op for the one in-box consumer.
> 
> **Telemetry:** `_elementsReplaced` feeds exactly one thing — `TotalElementCount` on `HotReloadClientOperationEvent`. Nothing branches on it, and HR success (`uiUpdating`) isn't count-derived, so the under-report is a metrics inaccuracy rather than behavior.
> 
> **On unoplatform/uno.extensions#3131:** it needs its `AfterVisualTreeUpdate` hook regardless of notifications — it also has to recover from `FrameNavigator` tracking a dead instance (no `Frame.Navigated` is raised on *either* replacement path), rebuild the view model through `InitializeCurrentView(refresh: true)` because an in-place EnC update keeps type identity and the "wrong VM type" check treats the stale VM as valid, and re-parent nested regions. So **(a)** wouldn't let it drop that handler.
> 
> Against that, **(a)** costs what you flagged: exposing the agent's handler actions to `ElementUpdate`, or moving the patch into the agent's dispatch — i.e. giving up the layer this fix sits at. `AfterVisualTreeUpdate` is already the once-per-phase reconciliation point for exactly this.
> 
> Documented in `doc/articles/uno-development/uno-internals-hotreload.md` as a `###` under "Intercepting the UI Update", directly below the handler-method list, plus a note on the helper and in the PR design notes.
> 
> What genuinely remains is that a *third-party* handler registered for a replaced content type isn't notified — that's now a stated contract rather than an accident. If you'd rather unify the channel anyway, I have **(a)** implemented locally and can push it: the mechanism sits in its own `ClientHotReloadProcessor` partial with a single assignment added to `DoUpdateVisualTreeCore`, so the walk stays untouched. Your call.

src/Uno.UI/UI/Xaml/Controls/ContentControl/ContentControl.HotReload.cs
```diff
@@ -0,0 +1,111 @@
...
+			typeof(ContentControlElementMetadataUpdateHandler).Log().Debug(
+				$"Re-creating stranded (unmaterialized) content {liveType.Name} of {host.GetType().Name} as {expectedType.Name}");
+		}
+
+		FrameworkElement? newContent;
+		try
+		{
+			// Centralized HR-aware creation (replacement-type resolution + bindable
+			// metadata provider), also used by NavigationCache and PagePool.
+			newContent = Frame.CreatePageInstance(liveType) as FrameworkElement;
```
dr1rrb: It's weird that `ContentControl` have a dependency on `Frame`, should be the opposite

src/Uno.UI/UI/Xaml/Controls/ContentControl/ContentControl.HotReload.cs
```diff
@@ -0,0 +1,111 @@
...
+			}
+
+			return null;
+		}
+
+		// Same transfer the walk's ReplaceViewInstance does. It moves local values only — including
+		// attached properties (region/route properties set by navigation frameworks) and bindings —
+		// so an inherited DataContext keeps flowing from the host instead of being pinned here.
+		((IDependencyObjectStoreProvider)currentContent).Store
+			.ClonePropertiesToAnotherStoreForHotReload(((IDependencyObjectStoreProvider)newContent).Store);
```
dr1rrb, line 106: Please make sure this is not a duplicate with what's being performed by the default HR engine (would suspect it's not because the element is not in the view, but this method has not been designed to be used anywhere else)

dr1rrb (approved):

src/Uno.UI/UI/Xaml/Controls/Frame/Frame.HotReload.cs
kazo0: (b) is fine so this should be good as-is then

kazo0 (approved):

src/Uno.UI/UI/Xaml/Controls/ContentControl/ContentControl.HotReload.cs
rajamatt: Yeah i just confirmed it by adding logs for the clone and both replacement paths: each instance gets cloned exactly once per cycle. It would seem that `VisualTreeHelper.GetParent` falls back to the visual parent, and the walk only reaches elements via visual children

PR #23816, opened 2026-07-20
build: Move accepted API removals to the 6.6 PackageDiff ignore set

ramezgerges:
**GitHub Issue:** _No related issue._ This is a CI/infra fix: the **Managed Binaries Build** (public-API diff) broke repo-wide once Uno 6.6 became the latest-stable comparison base; there is no separate tracking issue.

## PR Type:

## What changed? 🚀

The **Managed Binaries Build** (`generatepkgdiff` public-API diff) started failing on essentially every PR once **Uno 6.6 went stable**, with hundreds of `Removed <api> not found in ignore set` errors.

**Root cause.** `generatepkgdiff` diffs the built package against the **latest _stable_** `Uno.WinUI` on nuget.org, and applies **only the single** `<IgnoreSet>` whose `baseVersion` Major.Minor matches that base (`DiffIgnore.CompareVersion` → `FirstOrDefault`, non-cumulative). When the stable base moved `6.5.x → 6.6.x`, it switched from the `6.5` set to the `6.6` set, so the accepted WinUI-sync (WinAppSDK 1.8) removals — which lived in the `6.5` set — stopped being ignored:

- `ElementAnimator` / `AnimationContext` / `ElementAnimationCompleted` (types)
- `ItemsRepeater.Animator` / `AnimatorProperty` (property + accessors)
- the `AnimatedVisuals.*VisualSource` player surface (`Load`/`Measure`/`Play`/`Pause`/`Resume`/`Stop`/`SetProgress`/`Unload`/`Update`)

**Fix — move, not duplicate (and reuse the existing regex).** Because only the latest-stable set is ever consulted, every set below it (incl. `6.5`) is dead. So this **moves** the still-applicable removals into the now-active `6.6` set, **adds** the newer `AnimatedVisuals` removals (recent, never previously ignored), and **removes the now-redundant `6.5`↔`6.6` duplicates** (incl. pre-existing StackLayout/NavigationView copies and their orphaned comments). Each accepted removal is now recorded once, against the current base.

Net: +87 / −24 in `build/PackageDiffIgnore.xml`; `6.5`↔`6.6` overlap is 0.

## Validation
Reproduced and validated locally with the exact CI tool:
```
dotnet tool install Uno.PackageDiff --version 1.1.0-dev.29
generatepkgdiff --base=Uno.WinUI --other=<Uno.WinUI 6.7.0-dev.815> \
    --diffignore=build/PackageDiffIgnore.xml --outfile=diff.md
```
- Before: base resolves to `Uno.WinUI 6.6.166`, **830 errors**.
- After: **0 errors**, exit 0. XML validated well-formed.

> Note: the underlying repetition/per-release migration burden is a `Uno.PackageDiff` design issue (non-cumulative ignore-set selection). Applying all sets with `baseVersion ≤ base` cumulatively would remove the need to migrate on every stable release — worth a follow-up against the tool.

## PR Checklist ✅
- [x] 🧪 Validated locally with `generatepkgdiff` (0 errors)
- [x] ❗ Contains **NO** breaking changes (ignore-set only; documents already-accepted API removals)

### DIFF
--- a/build/PackageDiffIgnore.xml
+++ b/build/PackageDiffIgnore.xml
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8" ?>
+<?xml version="1.0" encoding="utf-8" ?>
 <DiffIgnore>
   <IgnoreSets>
 	<IgnoreSet baseVersion="4.0.13">
@@ -2262,10 +2262,6 @@
 
 		  <!-- END WinUI sync: Windows.* legacy / preview namespaces removed -->
 
-		  <!-- ItemsRepeater animator surface removed by WinAppSDK 1.8 (replaced by TransitionManager / ItemCollectionTransitionProvider). No references remain in WinUI microsoft-ui-xaml sources. -->
-		  <Member fullName="Microsoft.UI.Xaml.Controls.AnimationContext" reason="WinUI sync: ElementAnimator surface removed by WinAppSDK 1.8 (replaced by ItemCollectionTransitionProvider)" />
-		  <Member fullName="Microsoft.UI.Xaml.Controls.ElementAnimationCompleted" reason="WinUI sync: ElementAnimator surface removed by WinAppSDK 1.8 (replaced by ItemCollectionTransitionProvider)" />
-		  <Member fullName="Microsoft.UI.Xaml.Controls.ElementAnimator" reason="WinUI sync: ElementAnimator surface removed by WinAppSDK 1.8 (replaced by ItemCollectionTransitionProvider)" />
 	  </Types>
 
 	  <Fields>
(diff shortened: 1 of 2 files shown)


jonpryor (approved):

Xiaoy312 (approved):

dr1rrb (approved):
Validated the root cause on both sides: the ignore-set move matches how generatepkgdiff selects the single IgnoreSet for the base Major.Minor (base moved to 6.6.166 when 6.6 went stable on July 17), and the pinned `--no-http-cache` CLI install is confirmed by the green DevServer CLI jobs on this very build with the packaging SDK unchanged — the failing runs show an unpinned `--prerelease` resolution on a polluted feed setup, exactly what the pin removes. Thanks for unblocking everyone!

PR #23833, opened 2026-07-21
fix(devserver): surface settings app failure reason

agneszitte:
**GitHub Issue:** Related to internal tracker [unoplatform/uno.app-mcp#122](https://github.com/unoplatform/uno.app-mcp/issues/122)

Closes https://github.com/unoplatform/uno.app-mcp/issues/122

## PR Type:

🐞 Bugfix

## What changed? 🚀

**Scoped down per review ([@carldebilly](https://github.com/unoplatform/uno.app-mcp/issues/122)).** This PR is now just the diagnostics fix.

When the sign‑in ("Settings") app exits within the launch window, its own message — e.g. *"Another instance of this app is running"* — was logged only at `Debug`, so users saw a bare **`Settings application exited with code 2`** and assumed a failure. This surfaces the app's `stderr`/`stdout` as the reason on the error line.

### Dropped from this PR (superseded)

The earlier **project‑less‑login fallback** commit (cache‑scan / restore‑latest `uno.settings.devserver` when no `global.json`) was **removed**. Per review, the right fix for project‑less login is not to rebuild the `global.json → Uno.Sdk → packages.json` resolution chain, but to **publish `Uno.Settings` as a real dotnet tool** from the licensing stack (it already ships as a self‑contained publish) so the CLI invokes it via `dnx`. That gives official NuGet acquisition/integrity/execution, version coherence by construction, and pinning/kill‑switch for free — and removes the hand‑rolled resolution entirely. **That work is tracked on the licensing side; it is out of scope here.**

## Validation

- **Compile** — `Uno.UI.DevServer.Cli` builds clean on current master.
- **Runtime** — re‑running `login` with a sign‑in window already open now prints `Settings application exited with code 2: Another instance of this app is running, exiting` instead of a bare exit code.

## PR Checklist ✅

- [x] 🧪 Behavior verified at runtime (diagnostics message)
- [x] 📚 Docs — N/A (log message only)
- [x] 🖼️ N/A — no UI surface

### DIFF
--- a/src/Uno.UI.DevServer.Cli/CliManager.cs
+++ b/src/Uno.UI.DevServer.Cli/CliManager.cs
@@ -361,17 +361,18 @@ internal class CliManager
 
 		if (exitCode is not null)
 		{
-			// Display output for debugging purposes
-			if (!string.IsNullOrWhiteSpace(stdOut))
-			{
-				_logger.LogDebug("Settings application stdout:\n{Stdout}", stdOut);
-			}
-			if (!string.IsNullOrWhiteSpace(stdErr))
-			{
-				_logger.LogError("Settings application stderr:\n{Stderr}", stdErr);
-			}
+			// The settings app writes its exit reason to its output; surface it so the cause is
+			// visible instead of a bare exit code.
+			var reason = FormatSettingsExitReason(stdOut, stdErr);
 
-			_logger.LogError("Settings application exited with code {ExitCode}", exitCode);
+			if (!string.IsNullOrWhiteSpace(reason))
+			{
+				_logger.LogError("Settings application exited with code {ExitCode}: {Reason}", exitCode, reason);
+			}
+			else
+			{
+				_logger.LogError("Settings application exited with code {ExitCode} (no diagnostic output captured).", exitCode);
+			}
 
 			return 1;
 		}
@@ -380,6 +381,28 @@ internal class CliManager
 			_logger.LogInformation("Settings application started successfully");
 			return 0;
 		}
+	}
+
+	/// <summary>
+	/// Labels stdout/stderr (stdout first — the app writes its exit reason there) and collapses
+	/// internal whitespace so the reason stays a single log line.
+	/// </summary>
+	internal static string FormatSettingsExitReason(string? stdOut, string? stdErr)
+	{
+		var parts = new List<string>();
+		if (!string.IsNullOrWhiteSpace(stdOut))
+		{
+			parts.Add($"stdout: {CollapseWhitespace(stdOut)}");
+		}
+		if (!string.IsNullOrWhiteSpace(stdErr))
+		{
+			parts.Add($"stderr: {CollapseWhitespace(stdErr)}");
+		}
+
+		return string.Join("; ", parts);
+
+		static string CollapseWhitespace(string value)
+			=> string.Join(" ", value.Split((char[]?)null, StringSplitOptions.RemoveEmptyEntries));
 	}
 
 	private async Task<int> RunMcpProxyAsync(string[] args, string requestedWorkingDirectory, WorkspaceResolution workspaceResolution)
--- a/src/Uno.UI.DevServer.Cli.Tests/Given_SettingsExitReason.cs
+++ b/src/Uno.UI.DevServer.Cli.Tests/Given_SettingsExitReason.cs
@@ -1,0 +1,37 @@
+using AwesomeAssertions;
+using Uno.UI.DevServer.Cli;
+
+namespace Uno.UI.DevServer.Cli.Tests;
+
+[TestClass]
+public class Given_SettingsExitReason
+{
+	[TestMethod]
+	public void WhenOnlyStdout_LabelsStdout()
+		=> CliManager.FormatSettingsExitReason("Another instance is running", null)
+			.Should().Be("stdout: Another instance is running");
+
+	[TestMethod]
+	public void WhenOnlyStderr_LabelsStderr()
+		=> CliManager.FormatSettingsExitReason(null, "boom")
+			.Should().Be("stderr: boom");
+
+	[TestMethod]
+	public void WhenBothStreams_StdoutFirstSemicolonSeparated()
+		=> CliManager.FormatSettingsExitReason("out", "err")
+			.Should().Be("stdout: out; stderr: err");
+
+	[TestMethod]
+	[DataRow(null, null)]
+	[DataRow("", "   ")]
+	public void WhenNoOutput_ReturnsEmpty(string? stdOut, string? stdErr)
+		=> CliManager.FormatSettingsExitReason(stdOut, stdErr).Should().BeEmpty();
+
+	[TestMethod]
+	public void WhenMultilineOutput_CollapsedToSingleLine()
+	{
+		var result = CliManager.FormatSettingsExitReason("line1\r\nline2\n  line3  ", null);
+
+		result.Should().Be("stdout: line1 line2 line3");
+	}
+}


src/Uno.UI.DevServer.Cli/Helpers/UnoToolsLocator.cs
```diff
@@ -759,6 +777,114 @@ internal static bool TryParseNetTfm(string? name, out int major, out int minor)
...
+		// PackageReference (it ships as a tools package, not as an MSBuild SDK).
+		var csprojContents =
+			"""
+			<Project Sdk="Microsoft.NET.Sdk">
+				<PropertyGroup>
+					<TargetFramework>net9.0</TargetFramework>
+					<EnableDefaultItems>false</EnableDefaultItems>
+				</PropertyGroup>
+				<ItemGroup>
+					<PackageReference Include="uno.settings.devserver" Version="*" />
```
carldebilly: **HIGH — supply chain.** This floating `Version="*"` restore runs from `%TEMP%`, so it resolves against the user/machine `NuGet.Config` — every configured feed, with no package-source mapping, no source pinning and no signature check — and the resulting `Uno.Settings.dll` is then executed. Whoever can publish a higher `uno.settings.devserver` version to *any* configured feed (stale internal feed, team feed, mirror) gets code execution on the next project-less `uno.devserver login`. Suggest dropping a `nuget.config` with `<clear/>` + the explicit trusted feed(s) next to the temp csproj (or passing `--source`), and resolving a concrete version rather than `*`. A pinned version also enables a kill switch: consider an env var / `--settings-version` option to force a specific version when the latest published one is broken.

src/Uno.UI.DevServer.Cli/Helpers/UnoToolsLocator.cs
```diff
@@ -759,6 +777,114 @@ internal static bool TryParseNetTfm(string? name, out int major, out int minor)
...
+	}
+
+	private string? GetLatestCachedSettingsExecutable()
+	{
+		string? bestStableDir = null;
+		Version? bestStable = null;
+		string? bestAnyDir = null;
+		Version? bestAny = null;
+
+		foreach (var cache in GetNuGetCachePaths(_workDirectory))
```
carldebilly: **HIGH — cache-scan trust and a dropped parameter.** Two issues with `GetNuGetCachePaths(_workDirectory)` here:
- The highest-version-wins scan includes machine-wide locations such as `%ProgramData%\NuGet\packages`, and the winner is executed with no integrity check. On default Windows ACLs a local user can plant `uno.settings.devserver\999.0.0\tools\manager\Uno.Settings.dll` there, and unlike the packages.json-pinned normal path, this max-version policy will select it — cross-user code execution. Consider restricting the project-less fallback to `NUGET_PACKAGES` + the user-profile cache, and requiring `.nupkg.metadata`/signature presence before executing.
- `_workDirectory` is only assigned by `DiscoverAsync`, so on a pure login run it is null and a repo-local `nuget.config` redirecting `globalPackagesFolder` is silently ignored. Pass the `workDirectory` parameter down from `ResolveSettingsExecutableAsync`.

src/Uno.UI.DevServer.Cli/Helpers/UnoToolsLocator.cs
```diff
@@ -759,6 +777,114 @@ internal static bool TryParseNetTfm(string? name, out int major, out int minor)
...
+			if (!Directory.Exists(pkgRoot))
+			{
+				continue;
+			}
+
+			foreach (var dir in Directory.EnumerateDirectories(pkgRoot))
+			{
+				var name = Path.GetFileName(dir);
+				// NuGet folder names are the package version; strip any pre-release suffix to parse.
+				if (!Version.TryParse(name.Split('-')[0], out var v))
```
carldebilly: **HIGH — prerelease ordering is discarded.** `Version.TryParse(name.Split('-')[0])` drops the prerelease label, so `5.6.0-dev.1` and `5.6.0-dev.200` compare equal and `v > bestAny` keeps whichever directory the filesystem enumerates first — the "latest" pick becomes nondeterministic across machines whenever only prereleases are cached (often selecting the oldest build). Compare with NuGet version semantics (prerelease label as tie-break). This is also the 4th copy of this parsing idiom in the CLI (`ManifestAddInResolver.cs:34,113`, `UnoToolsLocator.cs:1021`) — worth promoting to a shared, tested helper.

src/Uno.UI.DevServer.Cli/Helpers/UnoToolsLocator.cs
```diff
@@ -759,6 +777,114 @@ internal static bool TryParseNetTfm(string? name, out int major, out int minor)
...
+		}
+
+		var dll = Path.Combine(chosen, "tools", "manager", "Uno.Settings.dll");
+		if (File.Exists(dll))
+		{
+			_logger.LogDebug("Found fallback Settings App: {Path}", dll);
+			return dll;
+		}
+
+		_logger.LogWarning("uno.settings.devserver resolved at {Dir} but Uno.Settings.dll is missing.", chosen);
```
carldebilly: **HIGH — no fallback to the next candidate.** The selection commits to the single newest directory and only then checks `File.Exists`; a corrupted/partial extraction of the newest version (with `.nupkg.metadata` intact, so `dotnet restore` no-ops) permanently breaks project-less login even when an intact older version sits in the same cache. Check for `tools/manager/Uno.Settings.dll` per candidate during the scan and skip directories that lack it. Related: this is the third copy of the `tools/manager/Uno.Settings.dll` layout literal (also lines 180 and 910) — extract a single helper/const so a package-layout change is a one-place edit.

src/Uno.UI.DevServer.Cli/Helpers/UnoToolsLocator.cs
```diff
@@ -759,6 +777,114 @@ internal static bool TryParseNetTfm(string? name, out int major, out int minor)
...
+		{
+			return cached;
+		}
+
+		_logger.LogInformation("No uno.settings.devserver found in cache; restoring the latest stable from NuGet.");
+		InstallLatestSettingsPackage();
+		return GetLatestCachedSettingsExecutable();
+	}
+
+	private string? GetLatestCachedSettingsExecutable()
```
carldebilly: **HIGH — tests (echoing the PR's own note).** The stable-preferred/newest-wins selection encoded here is pure, easily-testable policy, but as written it is fused with the filesystem walk, so it can't be unit-tested without a real NuGet cache on disk. Suggest extracting an internal static "pick best version dir from candidate names" function (names in → chosen dir out) and covering: stable-vs-prerelease preference, equal base versions with different prerelease labels, empty cache, corrupted directory, and restore failure — plus the three reason branches (stderr / stdout / none) in `CliManager.OpenSettings`. One practical note: `Uno.UI.RemoteControl.DevServer.Tests` has no `InternalsVisibleTo` into `Uno.UI.DevServer.Cli`, so the planned tests can't reach these internal types — they need to live in `Uno.UI.DevServer.Cli.Tests` (already IVT'd) or the IVT needs adding.

src/Uno.UI.DevServer.Cli/Helpers/UnoToolsLocator.cs
```diff
@@ -759,6 +777,114 @@ internal static bool TryParseNetTfm(string? name, out int major, out int minor)
...
+			</Project>
+			""";
+
+		var tempPath = Path.Combine(Path.GetTempPath(), $"uno-settings-installer-{Guid.NewGuid()}");
+		Directory.CreateDirectory(tempPath);
+		File.WriteAllText(Path.Combine(tempPath, "uno-settings-installer.csproj"), csprojContents);
+
+		_logger.LogDebug("Restoring latest uno.settings.devserver using temporary project in {TempPath}", tempPath);
+
+		var result = ProcessHelpers.RunProcess("dotnet", "restore", tempPath);
```
carldebilly: **HIGH — blocking call, no timeout, no cancellation.** `ProcessHelpers.RunProcess` blocks on `WaitForExit` with no timeout inside an async chain. On an unreachable or slow feed, or a credential provider that prompts interactively, `dotnet restore` hangs the login indefinitely with zero output, and Ctrl+C won't cleanly reach the hidden child process. Run the restore async with a `CancellationToken` + `CancelAfter` (~60–120 s), kill the child on cancel, and flow the token down from `ResolveSettingsExecutableAsync`.

src/Uno.UI.DevServer.Cli/Helpers/UnoToolsLocator.cs
```diff
@@ -759,6 +777,114 @@ internal static bool TryParseNetTfm(string? name, out int major, out int minor)
...
+		if (result.exitCode != 0)
+		{
+			_logger.LogWarning("uno.settings.devserver restore returned exit code {ExitCode}. Error: {Error}", result.exitCode, result.error);
```
carldebilly: **HIGH — restore diagnostics lost exactly where they're needed.** On failure only `result.error` is logged (at Warning), but NuGet writes its NU1101/NU1301 diagnostics to stdout — behind a corporate feed missing the package, the user sees `exit code 1. Error:` (empty) followed by the generic "could not resolve" error. The cold-cache fresh-machine scenario is the one this fallback exists for. Log `result.output` too (as `InstallUnoSdk` does at line 425) and consider Error level, since the login flow terminally fails here.

src/Uno.UI.DevServer.Cli/Helpers/UnoToolsLocator.cs
```diff
@@ -759,6 +777,114 @@ internal static bool TryParseNetTfm(string? name, out int major, out int minor)
...
 	}
 
+	/// <summary>
+	/// Project-less fallback for <see cref="ResolveSettingsExecutableAsync"/>: when no
+	/// global.json/SDK is in scope, use the latest available <c>uno.settings.devserver</c>
+	/// (newest already in the NuGet cache, else restored from the configured feeds) and
+	/// return its <c>tools/manager/Uno.Settings.dll</c>. The sign-in app does not depend on
+	/// a project, so this lets login run during environment setup, before any app exists.
+	/// </summary>
+	private string? ResolveLatestSettingsExecutable()
```
carldebilly: **HIGH — no compatibility contract between the CLI host and the settings app.** "Newest in cache, else floating `*`" has no floor or ceiling: a future `uno.settings.devserver` requiring a newer host, or a stale years-old cached copy, launches silently and breaks login in hard-to-diagnose ways. Consider pinning a minimum known-good version (ideally with a max-major aligned to the CLI's own release channel) and documenting the contract in this XML doc.

src/Uno.UI.DevServer.Cli/Helpers/UnoToolsLocator.cs
```diff
@@ -759,6 +777,114 @@ internal static bool TryParseNetTfm(string? name, out int major, out int minor)
...
+	{
+		var cached = GetLatestCachedSettingsExecutable();
+		if (cached is not null)
+		{
+			return cached;
+		}
+
+		_logger.LogInformation("No uno.settings.devserver found in cache; restoring the latest stable from NuGet.");
```
carldebilly: **MEDIUM — inconsistent prerelease/feed policy.** This log promises "the latest stable", but the cache scan above happily falls back to a prerelease, while the restore below is stable-only (`Version="*"` excludes prereleases) and runs from `%TEMP%`, so it doesn't inherit a working-directory `nuget.config` (private feeds). Net effect: cold cache + prerelease-only or private feed — exactly the setup-phase early-adopter case — leaves login broken with a confusing message. Decide the prerelease policy once (e.g. retry with `*-*` when `*` resolves nothing), restore with the working directory's NuGet context (or `--configfile`), and align the log message with the actual behavior.

src/Uno.UI.DevServer.Cli/CliManager.cs
```diff
@@ -361,18 +361,23 @@ private async Task<int> OpenSettings(string[] originalArgs, string workingDirect
...
+			// stdout — previously logged only at Debug, so users saw a bare exit code and assumed
+			// a failure. Prefer stderr, then stdout, so the actual cause is always visible.
+			var reason = !string.IsNullOrWhiteSpace(stdErr) ? stdErr.Trim()
+				: !string.IsNullOrWhiteSpace(stdOut) ? stdOut.Trim()
+				: null;
+
+			if (reason is not null)
 			{
-				_logger.LogDebug("Settings application stdout:\n{Stdout}", stdOut);
+				_logger.LogError("Settings application exited with code {ExitCode}: {Reason}", exitCode, reason);
```
carldebilly, line 370: **MEDIUM — every exit inside the launch window is classified as an error.** Any exit within the window — including exit code 0, and the "another instance of this app is running" handoff the comment above names as the common case — is logged at Error and returns 1. A second `uno.devserver login` while the settings window is already open reports failure (exit 1) to setup scripts even though the sign-in UI is up. Consider treating exit code 0 (and the documented single-instance exit code) as success/informational. Minor hardening while here: the echoed reason is unbounded and unsanitized — truncating (~1 KB) and stripping control/ANSI sequences would prevent terminal-escape injection and accidental leakage of auth output into logs.

src/Uno.UI.DevServer.Cli/Helpers/UnoToolsLocator.cs
```diff
@@ -297,12 +297,20 @@ public async Task<DiscoveryInfo> DiscoverAsync(string workDirectory, WorkspaceRe
...
+			// No global.json / Uno SDK in scope (e.g. signing in during environment setup,
+			// before any app exists). The sign-in app is project-independent, so fall back to
+			// the latest uno.settings.devserver instead of refusing to launch.
+			_logger.LogInformation("No Uno SDK resolved from global.json; using the latest uno.settings.devserver for project-less login.");
+			var fallback = ResolveLatestSettingsExecutable();
+			if (fallback is null)
+			{
+				_logger.LogError("Could not resolve a uno.settings.devserver package for project-less login.");
+			}
+			return fallback;
```
carldebilly: **MEDIUM — a stale global.json still hard-fails login.** The fallback only triggers when no global.json/Uno.Sdk is found; a global.json pinning an unresolvable SDK version still fails the login with no fallback, which contradicts the rationale that sign-in is project-independent. Either fall back after `EnsureNugetPackage` failure on the login path too, or document why pinned-version fidelity must win there.

src/Uno.UI.DevServer.Cli/Helpers/UnoToolsLocator.cs
```diff
@@ -759,6 +777,114 @@ internal static bool TryParseNetTfm(string? name, out int major, out int minor)
...
+	/// <summary>
+	/// Project-less fallback for <see cref="ResolveSettingsExecutableAsync"/>: when no
+	/// global.json/SDK is in scope, use the latest available <c>uno.settings.devserver</c>
+	/// (newest already in the NuGet cache, else restored from the configured feeds) and
+	/// return its <c>tools/manager/Uno.Settings.dll</c>. The sign-in app does not depend on
+	/// a project, so this lets login run during environment setup, before any app exists.
+	/// </summary>
+	private string? ResolveLatestSettingsExecutable()
+	{
+		var cached = GetLatestCachedSettingsExecutable();
```
carldebilly: **MEDIUM — unhandled exceptions escape the fallback path.** Unlike the normal path (`GetSettingsExecutable` wraps everything in try/catch), nothing here catches `Directory.EnumerateDirectories` (inaccessible `NUGET_PACKAGES` path) or `ProcessHelpers.RunProcess` ("Failed to start") — the exception escapes `ResolveSettingsExecutableAsync` and breaks `OpenSettings`' "errors already logged, return null" contract. Wrap the scan/restore mirroring the normal path's exception handling.

src/Uno.UI.DevServer.Cli/Helpers/UnoToolsLocator.cs
```diff
@@ -759,6 +777,114 @@ internal static bool TryParseNetTfm(string? name, out int major, out int minor)
...
+					<TargetFramework>net9.0</TargetFramework>
+					<EnableDefaultItems>false</EnableDefaultItems>
+				</PropertyGroup>
+				<ItemGroup>
+					<PackageReference Include="uno.settings.devserver" Version="*" />
+				</ItemGroup>
+			</Project>
+			""";
+
+		var tempPath = Path.Combine(Path.GetTempPath(), $"uno-settings-installer-{Guid.NewGuid()}");
```
carldebilly: **MEDIUM — temp directory is never cleaned up.** Every cold-cache project-less login leaks a `uno-settings-installer-{guid}` directory, and failed restores retry on each invocation, so a broken feed produces unbounded `%TEMP%` growth. Wrap in try/finally with a best-effort recursive delete. (`InstallUnoSdk` has the same leak — a shared helper would fix both, see the comment on the duplication below.)

src/Uno.UI.DevServer.Cli/Helpers/UnoToolsLocator.cs
```diff
@@ -759,6 +777,114 @@ internal static bool TryParseNetTfm(string? name, out int major, out int minor)
...
+		}
+
+		_logger.LogWarning("uno.settings.devserver resolved at {Dir} but Uno.Settings.dll is missing.", chosen);
+		return null;
+	}
+
+	private void InstallLatestSettingsPackage()
+	{
+		// Restore the latest stable uno.settings.devserver into the global NuGet cache via a
+		// throwaway project. Mirrors InstallUnoSdk, but references the settings app as a
```
carldebilly: **MEDIUM — copy-paste of `InstallUnoSdk`'s restore scaffolding.** The temp-dir/csproj/restore/warn-on-fail mechanism is duplicated (as this comment itself notes), and the two copies already diverge: this one pins `net9.0` where `InstallUnoSdk` uses `net10.0` (silent drift at the next TFM bump), and it logs less on failure. Extract a shared `RestoreViaTempProject(projectName, csprojContents)` helper — also the single place to own the TFM choice and the temp-dir cleanup.

src/Uno.UI.DevServer.Cli/Helpers/UnoToolsLocator.cs
```diff
@@ -759,6 +777,114 @@ internal static bool TryParseNetTfm(string? name, out int major, out int minor)
...
+		if (chosen is null)
+		{
+			_logger.LogDebug("No uno.settings.devserver package found in any NuGet cache location.");
+			return null;
+		}
+
+		var dll = Path.Combine(chosen, "tools", "manager", "Uno.Settings.dll");
+		if (File.Exists(dll))
+		{
+			_logger.LogDebug("Found fallback Settings App: {Path}", dll);
```
carldebilly: **MEDIUM — chosen fallback version only visible at Debug.** Support reports for "login launched a broken settings app" won't say which version was picked at default verbosity. Log the chosen version (parsed from the directory name) at Information on the fallback path so failure reports include it.

src/Uno.UI.DevServer.Cli/Helpers/UnoToolsLocator.cs
```diff
@@ -354,12 +362,22 @@ public async Task<DiscoveryInfo> DiscoverAsync(string workDirectory, WorkspaceRe
...
+			const string message = "No global.json found in current directory or parent directories. Please run this command from within a project that uses Uno SDK.";
+			// Callers with a project-less fallback (login) pass false so this reads as an
+			// informational Debug line rather than a spurious error the user should act on.
+			if (logMissingGlobalJsonAsError)
+			{
+				_logger.LogError(message);
+			}
+			else
+			{
+				_logger.LogDebug(message);
```
carldebilly: **LOW — misleading guidance on the fallback path.** The downgraded Debug message still says "Please run this command from within a project that uses Uno SDK" on a path that deliberately proceeds without one. Use a neutral message when `logMissingGlobalJsonAsError` is false.

carldebilly (approved):
Reviewed across security, architecture, robustness, operability, behavior-contract and quality dimensions. The feature is sound and in the right place: `UnoToolsLocator` already owns cache probing and restore, the cache-first fallback mirrors the established SDK-resolution flow, and the stderr/stdout surfacing in `CliManager` is a genuine diagnostics improvement. Approving to support the direction — this is a draft, and the inline comments (8 high, 7 medium, 1 low) should be addressed before undrafting, most importantly pinning the feed/version on the restore path and adding the planned tests.

Verified as safe:
- The error→debug downgrade is correctly scoped to the login path (`logMissingGlobalJsonAsError` defaults to `true`; every other caller keeps the error).
- No public API surface is touched (both classes are `internal`; the only `InternalsVisibleTo` is `Uno.UI.DevServer.Cli.Tests`, which has no test on this path today).
- Nothing in the repo or docs asserts the old exit-1 "run from within a project" message.
- Concurrent logins are safe (GUID temp dirs + NuGet global-cache locking).

Additional low/info notes, not worth an inline thread each:
- The `uno.settings.devserver` package-id literal is now repeated (lines 809, 870, plus pre-existing 904) — hoist a private const.
- The fallback skips the normal path's `GetDotNetVersion` gate, so the newest settings app may require a runtime the machine lacks; at least the new stderr capture will surface the failure.
- The `Settings application exited with code {ExitCode}` log template forks into two variants — anything keyed on the exact template will see new shapes.
- The login exit-code contract intentionally flips outside a project (exit 1 + error → exit 0). Nothing asserts the old behavior, but it is an observable CLI behavior change — worth a line in the PR description rather than "Contains NO breaking changes".
- Style nits: the comment block at `CliManager.cs:364-367` narrates prior behavior ("previously logged only at Debug…") — trim to the why; `name.Split('-')[0]` allocates an array and `Contains('-')` re-scans the same string (use `IndexOf` once + `AsSpan`); the new block mixes `== null` (line 302) with `is null` (line 309) — converge on `is null`.

src/Uno.UI.DevServer.Cli/CliManager.cs
```diff
@@ -361,18 +361,23 @@ private async Task<int> OpenSettings(string[] originalArgs, string workingDirect
 
 		if (exitCode is not null)
 		{
-			// Display output for debugging purposes
-			if (!string.IsNullOrWhiteSpace(stdOut))
+			// Surface the app's own message as the reason (it exited before the launch window).
+			// The settings app reports things like "Another instance of this app is running" on
+			// stdout — previously logged only at Debug, so users saw a bare exit code and assumed
+			// a failure. Prefer stderr, then stdout, so the actual cause is always visible.
+			var reason = !string.IsNullOrWhiteSpace(stdErr) ? stdErr.Trim()
```
autocarl: **MEDIUM — the new diagnostic can still lose the actual reason because output capture is not drained.** On `4f6522a`, `RunGuiProcessAsync` returns when the process exits and snapshots its `StringBuilder`s while the asynchronous `OutputDataReceived` / `ErrorDataReceived` callbacks may still be appending. A one-line early-exit probe lost the final reason in 13/1,000 runs; a 2,000-line probe also reproduced `ArgumentOutOfRangeException` in `StringBuilder.ToString()`. This path can therefore fall into “no diagnostic output captured” (or crash) instead of delivering the reason this change promises. Please make the helper await both process exit and stream EOF/drain before snapshotting the buffers, synchronize buffer access, and add a regression test that writes its reason immediately before exit.

autocarl (commented):
One additional diagnostics issue reproduced on the pinned head; see the inline comment.

src/Uno.UI.DevServer.Cli/Helpers/UnoToolsLocator.cs
autocarl: **HIGH — the temporary restore project can import attacker-controlled ancestor MSBuild files.** On Unix-like hosts, `Path.GetTempPath()` is commonly a shared directory. `dotnet restore` automatically walks ancestor directories for `Directory.Build.props` / `.targets`, so another local user can plant one in the shared temp root and have its targets run as the developer during this documented cold-cache login path. I reproduced the exact import behavior safely: an ordinary restore of a nested scratch project executed an inert target from its parent; the marker disappeared with `ImportDirectoryBuildProps=false` and `ImportDirectoryBuildTargets=false`. Create the project below a user-private root and disable both directory-build imports on the restore command (or avoid evaluating an MSBuild project by restoring through NuGet APIs).

autocarl (commented):
One additional security issue reproduced on the pinned head; see the inline comment.

src/Uno.UI.DevServer.Cli/CliManager.cs
agneszitte: This PR is now scoped to just surfacing the settings app's output (per the scope-down). It doesn't change the exit-code/return semantics — every within-window exit still returns 1. The "exit code 0 should be success" point is valid, but it's a behavior change beyond this diagnostics-only PR; it's better handled with the broader login work — and cleanly once `Uno.Settings` is a real dotnet tool with a defined exit contract (rather than coupling to specific magic exit codes here). Resolving here and tracking it there.

src/Uno.UI.DevServer.Cli/CliManager.cs
agneszitte: Confirmed valid and pre-existing: the race is in `DevServerProcessHelper.RunGuiProcessAsync` (the `StringBuilder` snapshot can run before the async `OutputDataReceived`/`ErrorDataReceived` callbacks finish draining), not in this PR's changed code. Per the review's own cross-cutting note, the fix belongs in `DevServerProcessHelper.cs` and is a separate follow-up; this diagnostics improvement is still a net gain in the meantime. Resolving here and tracking the drain fix separately.

autocarl (approved):
Re-reviewed the exact current head `c5eda0570b829c1b99bd01de2301a76ef22207d4`. No material findings.

The earlier project-less package-resolution changes—and the associated security, restore, versioning, and operability concerns—are no longer in this PR. The remaining diagnostics-only change is correctly scoped and covered: the 6 targeted exit-reason cases pass, the DevServer CLI builds in Release with 0 warnings/errors, and the full CLI project yields 390/393 locally with the same three unrelated environment-specific failures reproduced on the exact base SHA. All hosted checks are passing and no current unresolved review thread remains.

morning4coffe-dev (approved):
The settings-process failure now exposes both diagnostic streams at the visible error level, while preserving the existing exit-code/result behavior. The helper has deterministic stdout/stderr ordering and handles whitespace-only and multiline output without turning process text into a logging format string. No WinUI runtime behavior is involved.

PR #23835, opened 2026-07-21
fix(x11): Disable GDK frame sync on embedded webview window

### LINKED ISSUE #23159: [X11] WebView2 crash on Linux
### Current behavior 🐛

Splitting out from: https://github.com/unoplatform/uno/issues/23142#issuecomment-4329613200

Use of `WebView2` within Uno.Gallery crashes:

```sh
git clone https://github.com/unoplatform/Uno.Gallery.git
cd Uno.Gallery
git checkout 54a08b1587aa020a5651d61eb3a3023a79e89d72
sed -i '' 's/"Uno.Sdk": ".*"/"Uno.Sdk": "6.6.0-dev.216"/g' global.json
dotnet publish -c Release -r linux-arm64 -f net10.0-desktop -p:TargetFrameworkOverride=net10.0-desktop \
  -bl Uno.Gallery/Uno.Gallery.csproj -p:SelfContained=true -p:UseSkiaRendering=true
./Uno.Gallery/bin/Release/net10.0-desktop/linux-arm64/publish/Uno.Gallery 
```

Navigate to **UI components > WebView** in the left-hand tree, and the app immediately crashes:

```
fail: Uno.WinUI.Runtime.Skia.X11.X11NativeOverlappedPresenter[0]
      Couldn't get OverlappedPresenterState: _NET_WM_STATE does not exist on the window. Make sure you use an EWMH-compliant WM.
fail: Uno.WinUI.Runtime.Skia.X11.X11NativeOverlappedPresenter[0]
      Couldn't get OverlappedPresenterState: _NET_WM_STATE does not exist on the window. Make sure you use an EWMH-compliant WM.

(Uno.Gallery:8586): Gdk-CRITICAL **: 14:46:16.851: gdk_x11_window_get_xid: assertion 'GDK_IS_X11_WINDOW (window)' failed
X Error of failed request:  BadWindow (invalid Window parameter)
  Major opcode of failed request:  3 (X_GetWindowAttributes)
  Resource id in failed request:  0x0
  Serial number of failed request:  103
  Current serial number in output stream:  104
```

### Expected behavior 🎯

No crash. WebView2 works.

_No response_

### Workaround 🛠️

_No response_

### Renderer 🎨

- [x] Skia
- [ ] Native

### Affected platforms 📱💻🖥️

Desktop (X11)

Uno.Sdk 6.6.0-dev.216.

### IDE version 🧑‍💻

_No response_

This occurs on a Parallels VM, running an "Ubuntu 24.04.3 ARM64" image.

I also needed to run:

```sh
sudo apt install libwebkit2gtk-4.1-dev
```

in order to get the "unversioned" `libwebkit2gtk-4.1.so` which is used by a `[DllImport]`: https://github.com/unoplatform/uno/blob/c507882bcc1b6c40ad37c86ae5599e3d2d63f45e/src/AddIns/Uno.UI.WebView.Skia.X11/X11NativeWebView.cs#L56

ramezgerges:
**GitHub Issue:** closes #23159

## PR Type:

🐞 Bugfix

## What changed? 🚀

On GNOME (Mutter — both Xorg and XWayland sessions), `WebView2` on Skia X11 renders a permanently blank panel. Historically it was flaky (blank until scrolling forced a repaint); since 6.7.0-dev.123 (#23736) it stopped rendering entirely.

### Root cause

Mutter advertises `_NET_WM_FRAME_DRAWN` (compositor frame synchronization), so GDK paints a frame only after the WM acknowledges the previous one. The WebKitGTK window hosting the webview is reparented into the application's X11 window and is therefore never managed by the WM — no acknowledgement ever arrives, GDK's frame clock stalls, and WebKit stops painting. The X11 side is entirely healthy (correct parent, mapped, viewable, correct geometry); the window is just never drawn into.

Weston (WSLg) and wlroots-based compositors (sway, etc.) don't support frame sync, so GTK falls back to a timer-driven frame clock there — which is why the issue only reproduces on GNOME.

#23736 didn't regress this at the X11 level: before it, the marshaling bug in `AttachSubWindow` meant override-redirect was effectively never applied, so the window sometimes got *briefly* managed by Mutter during the map/attach race, letting a few frames through — the "blank until you scroll" flakiness reported in #23159. Making "never managed" deterministic made the stall permanent.

### Fix

Call `gdk_x11_window_set_frame_sync_enabled(window, FALSE)` on the GDK window right after `Realize()` — GDK's opt-out for exactly this scenario. The override-redirect handling from #23736 is unchanged (still required for Weston's XWM).

### Validation

- **Runtime, Mutter 46.2 (Ubuntu 24.04) headless XWayland**: `SamplesApp` → `WebView2_Basic` renders the page with the fix; a negative control build (same everything minus the one call) reproduces the permanently blank webview. Fails-before / passes-after, screenshot-verified.
- **Runtime, WSLg (Weston XWM)**: page renders with the fix — no regression of the #23736 scenario.
- **GTK-level repro matrix** replicating the exact embedding flow (realize → override-redirect → reparent → map → layout), across four XWMs (Weston, sway/wlroots, Mutter-on-XWayland, Mutter-on-Xorg) × both map/attach orderings × old/new attach variants: only frame-sync-disabled variants render everywhere; the current code fails only under Mutter, matching the field reports.

## PR Checklist ✅

- [x] 🧪 Added [Runtime tests, UI tests, or a manual test sample](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) (for bug fixes / features, if applicable) — covered by the existing `WebView2_Basic` manual-test sample; the failure depends on the window manager (GNOME/Mutter), which CI test environments don't exercise.

### DIFF
--- a/src/AddIns/Uno.UI.WebView.Skia.X11/X11NativeWebView.cs
+++ b/src/AddIns/Uno.UI.WebView.Skia.X11/X11NativeWebView.cs
@@ -180,7 +180,17 @@ public class X11NativeWebView : INativeWebView
 			// painting the next one. Our window gets reparented into the app's window and
 			// is never managed by the WM, so no acknowledgement ever comes and GTK (and
 			// therefore WebKit) stops painting entirely, leaving the webview blank.
-			gdk_x11_window_set_frame_sync_enabled(_window.Window.Handle, false);
+			try
+			{
+				gdk_x11_window_set_frame_sync_enabled(_window.Window.Handle, false);
+			}
+			catch (EntryPointNotFoundException e) // GTK < 3.8
+			{
+				if (this.Log().IsEnabled(LogLevel.Warning))
+				{
+					this.Log().Warn("Couldn't disable GDK frame sync. The webview might not render under Mutter-based window managers (e.g. GNOME).", e);
+				}
+			}
 
 			// Set override-redirect before the window is first mapped, so that the WM never
 			// starts managing it as a toplevel. Some WMs (e.g. Weston's XWM on WSLg) would
--- a/src/AddIns/Uno.UI.WebView.Skia.X11/X11NativeWebView.cs
+++ b/src/AddIns/Uno.UI.WebView.Skia.X11/X11NativeWebView.cs
@@ -66,7 +66,7 @@ public class X11NativeWebView : INativeWebView
 	private static extern IntPtr gdk_x11_window_get_xid(IntPtr window);
 
 	[DllImport("libgdk-3.so", CallingConvention = CallingConvention.Cdecl)]
-	private static extern void gdk_x11_window_set_frame_sync_enabled(IntPtr window, bool frameSyncEnabled);
+	private static extern void gdk_x11_window_set_frame_sync_enabled(IntPtr window, int frameSyncEnabled);
 
 	static X11NativeWebView()
 	{
@@ -182,7 +182,7 @@ public class X11NativeWebView : INativeWebView
 			// therefore WebKit) stops painting entirely, leaving the webview blank.
 			try
 			{
-				gdk_x11_window_set_frame_sync_enabled(_window.Window.Handle, false);
+				gdk_x11_window_set_frame_sync_enabled(_window.Window.Handle, /* False */ 0);
 			}
 			catch (EntryPointNotFoundException e) // GTK < 3.8
 			{
--- a/src/AddIns/Uno.UI.WebView.Skia.X11/X11NativeWebView.cs
+++ b/src/AddIns/Uno.UI.WebView.Skia.X11/X11NativeWebView.cs
@@ -65,6 +65,9 @@ public class X11NativeWebView : INativeWebView
 	[DllImport("libgdk-3.so", CallingConvention = CallingConvention.Cdecl)]
 	private static extern IntPtr gdk_x11_window_get_xid(IntPtr window);
 
+	[DllImport("libgdk-3.so", CallingConvention = CallingConvention.Cdecl)]
+	private static extern void gdk_x11_window_set_frame_sync_enabled(IntPtr window, bool frameSyncEnabled);
+
 	static X11NativeWebView()
 	{
 		try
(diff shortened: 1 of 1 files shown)


src/AddIns/Uno.UI.WebView.Skia.X11/X11NativeWebView.cs
```diff
@@ -172,6 +175,13 @@ public X11NativeWebView(CoreWebView2 coreWebView2, ContentPresenter presenter)
 			_webview.ShowAll();
 			_window.Realize(); // creates the Gdk window (and the X11 window) without showing it
 
+			// WMs that support _NET_WM_FRAME_DRAWN (e.g. Mutter/GNOME, on both X11 and
+			// XWayland) make GDK wait for the WM's acknowledgement of each frame before
+			// painting the next one. Our window gets reparented into the app's window and
+			// is never managed by the WM, so no acknowledgement ever comes and GTK (and
+			// therefore WebKit) stops painting entirely, leaving the webview blank.
+			gdk_x11_window_set_frame_sync_enabled(_window.Window.Handle, false);
```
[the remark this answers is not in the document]
ramezgerges: Guarded in 597c56d3ac — the call is now wrapped in try/catch on `EntryPointNotFoundException` with a warning log. Note this entry point has existed since GTK 3.8 (2013) and WebKitGTK 4.x requires a much newer GTK3, so this is defensive hardening only; also, since #23736 an exception here would propagate as a WebView init failure rather than crash the process.

src/AddIns/Uno.UI.WebView.Skia.X11/X11NativeWebView.cs
```diff
@@ -65,6 +65,9 @@ public class X11NativeWebView : INativeWebView
 	[DllImport("libgdk-3.so", CallingConvention = CallingConvention.Cdecl)]
 	private static extern IntPtr gdk_x11_window_get_xid(IntPtr window);
 
+	[DllImport("libgdk-3.so", CallingConvention = CallingConvention.Cdecl)]
+	private static extern void gdk_x11_window_set_frame_sync_enabled(IntPtr window, bool frameSyncEnabled);
```
jonpryor: I would suggest *not* using `bool`, as the marshaling rules for `bool` are quite complicated, and it's not a blittable type.

You should use `int frameSyncEnabled` instead.

ramezgerges: Done in 31a2fbba66 — `int frameSyncEnabled`, passing `/* False */ 0` at the call site.

jonpryor (approved):

PR #23874, opened 2026-07-24
fix(GLCanvasElement): blank canvas on physical iOS devices — probe GLES BGRA readback support

### LINKED ISSUE #23873: GLCanvasElement renders a blank canvas on physical iOS devices (works in the iOS simulator)
### Current behavior 🐛

On a **physical iOS device**, a `GLCanvasElement` subclass shows a permanently blank/empty canvas. There is no exception and nothing in the logs: `IsGLInitialized` is `true`, `Init()` and `RenderOverride()` run normally, and an `Invalidate()`-driven render loop keeps ticking. The same app renders correctly in the **iOS simulator** and on **Skia Desktop**.

### Expected behavior 🎯

Content drawn in `RenderOverride` is displayed on physical devices, the same way it is in the iOS simulator and on desktop.

1. Create an app with the `GLCanvas` UnoFeature and add an explicit `PackageReference` to `Uno.WinUI.Graphics3DGL` for the iOS head (the SDK feature mapping currently wires it only for Desktop/WinAppSDK).
2. Subclass `GLCanvasElement` and draw anything — the triangle sample from the [GLCanvasElement docs](https://platform.uno/docs/articles/controls/GLCanvasElement.html) is enough.
3. Run on the iOS **simulator** → the canvas renders.
4. Deploy the same build to a **physical device** (reproduced on iPad Pro 11" 3rd gen (M1), iPadOS 26.5) → the canvas is blank.

I can attach a zipped minimal repro project if helpful.

### Renderer 🎨

Skia

### Affected platforms 📱💻🖥️

iOS (physical devices only — the simulator is not affected)

- Uno.Sdk `6.7.0-dev.125` (resolves `Uno.WinUI.Graphics3DGL` `6.7.0-dev.815`)
- `net10.0-ios`, .NET SDK `10.0.201`, iOS workload `26.2.10217`

### IDE version 🧑‍💻

None (dotnet CLI on macOS; deployed with `xcrun devicectl`)

I instrumented the element on the device to narrow down where the pipeline stops:

- The GL context is created fine on the real GPU: `Vendor: Apple Inc.`, `Renderer: Apple M1 GPU`, `Version: OpenGL ES 3.0 Metal - 104.1` (on the simulator it is `Apple Software Renderer`).
- `Init()` and `RenderOverride()` execute every frame with `glGetError() == GL_NO_ERROR` after drawing.
- Reading the FBO's center pixel with `glReadPixels(GL_RGBA, GL_UNSIGNED_BYTE)` inside `RenderOverride` returns the expected drawn color — **the offscreen framebuffer content is correct**.
- Inspecting the `WriteableBitmap` back buffer that feeds the element's `ImageBrush` shows it stays **all zeros** — the FBO → bitmap copy never writes anything.

**Possible cause:** `GLCanvasElement.Render()` reads the framebuffer back with `glReadPixels(…, GL_BGRA, GL_UNSIGNED_BYTE, …)`. On OpenGL ES, BGRA readback is only legal with the optional `GL_EXT_read_format_bgra` extension, which Apple's GLES-on-Metal driver on physical hardware does not expose (and with the FBO's unsized `GL_RGB` color attachment, the implementation-defined read-format pair is not BGRA either). The existing RGBA-with-swap readback fallback (`NeedsRgbaReadbackSwap`) is only engaged when `OperatingSystem.IsAndroid()`, so iOS always takes the BGRA path. Since a failed `glReadPixels` only records a GL error and never throws, the destination buffer is left untouched and the element silently presents an empty bitmap. The simulator's software renderer happens to accept BGRA readback, which is why this only reproduces on physical devices.

ajpinedam:
**GitHub Issue:** closes #23873

## PR Type:

🐞 Bugfix

## What changed? 🚀

On **physical iOS devices**, `GLCanvasElement` presented a permanently blank canvas while the same build worked in the iOS simulator and on desktop. The scene was actually rendered correctly into the offscreen FBO every frame — the failure was the FBO → `WriteableBitmap` copy: `Render()` reads back with `glReadPixels(GL_BGRA, GL_UNSIGNED_BYTE)`, which Apple's GLES-on-Metal driver rejects for the element's framebuffer. Since a failed `glReadPixels` only sets a GL error and never throws, the back buffer stayed zeroed and the element silently presented nothing.

Two changes, one per commit:

1. **Probe GLES BGRA readback support with a real read** — `NeedsRgbaReadbackSwap` previously engaged the RGBA+swap fallback only on Android, via an extension-string check. Neither is sufficient: on-device probing shows Apple's driver **advertises `GL_EXT_read_format_bgra` but still rejects BGRA reads** from the element's FBO (unsized `GL_RGB` color attachment; `GL_IMPLEMENTATION_COLOR_READ_FORMAT/TYPE` reports `GL_RGBA`/`GL_UNSIGNED_BYTE`). The detection now runs on any GLES context (identified by the mandated `"OpenGL ES"` version-string prefix) and performs an actual 1×1 BGRA `glReadPixels` against the element's framebuffer, which is bound at the call site: if it errors, readback falls back to `GL_RGBA` + R/B swap (always legal on GLES). Desktop GL (BGRA is core) and WASM (the JS shim swaps internally) keep their current behavior.

2. **Surface `glReadPixels` failures** — `Render()` now drains pre-existing GL errors before the readback (bounded, since `GetError` can return `GL_CONTEXT_LOST` forever) and throws `InvalidOperationException` naming the GL error if the readback failed. This lands in the existing catch that logs and sets `IsGLInitialized = false`, so this class of failure can never again present as a silent blank canvas.

**Validation** (iPad Pro 11" 3rd gen (M1), iPadOS 26.5, `net10.0-ios`, instrumented app inspecting both the FBO and the presented back buffer; the issue also reproduced on an iPhone 16 Pro-class device, A18 Pro):

| Configuration | Readback path chosen | Presented back buffer | Result |
|---|---|---|---|
| Current package (`6.7.0-dev.815`) | BGRA | all zeros | blank canvas, no error ❌ |
| Extension-string detection (rejected approach) | BGRA (extension is advertised!) | all zeros | fails — extension can't be trusted ❌ |
| **This PR (1×1 probe read)** | RGBA + swap | correct pixels | renders ✅ |

## PR Checklist ✅

- [x] 🧪 Validated manually on physical hardware (iPad Pro M1 / iPadOS 26.5) and on the iOS simulator + Skia Desktop (no regression); existing GLCanvasElement samples apply

https://claude.ai/code/session_014BdoX4X8eXiY1CuyXdhjqC

### DIFF
--- a/src/AddIns/Uno.WinUI.Graphics3DGL/GLCanvasElement.cs
+++ b/src/AddIns/Uno.WinUI.Graphics3DGL/GLCanvasElement.cs
@@ -55,6 +55,9 @@ public abstract partial class GLCanvasElement : Grid, INativeContext
 	// internally), so this stays false there.
 	private bool _readbackAsRgbaWithSwap;
 
+	// Rate-limits the pending-GL-error warning in Render to once per element.
+	private bool _warnedPendingGlError;
+
 	// valid if and only if GLCanvasElement was loaded at least once and OpenGL is available on the running platform
 	private INativeOpenGLWrapper? _nativeOpenGlWrapper;
 	// These are valid if and only if IsLoaded and _nativeOpenGlWrapper is not null
@@ -100,7 +103,7 @@ public abstract partial class GLCanvasElement : Grid, INativeContext
 	/// OpenGL errors raised by user code should be checked and handled within <see cref="Init"/>
 	/// and <see cref="RenderOverride"/>. Errors still pending before the framebuffer readback are
 	/// drained (GL errors are sticky and would otherwise be misattributed to the readback check)
-	/// and logged as warnings.
+	/// and logged — as a warning on the first occurrence, then at Debug level.
 	/// </remarks>
 	protected abstract void RenderOverride(GL gl);
 
@@ -510,9 +513,16 @@ public abstract partial class GLCanvasElement : Grid, INativeContext
 					break;
 				}
 
-				if (this.Log().IsEnabled(LogLevel.Warning))
+				// Warn once per element: with Invalidate() called from RenderOverride, a persistent
+				// user error would otherwise flood the logs at frame rate.
+				if (!_warnedPendingGlError && this.Log().IsEnabled(LogLevel.Warning))
 				{
-					this.Log().Warn($"GL error {pendingError} was pending before the framebuffer readback, most likely raised by {nameof(RenderOverride)} (or {nameof(Init)} on the first frame); GL errors should be checked and handled within user code.");
+					_warnedPendingGlError = true;
+					this.Log().Warn($"GL error {pendingError} was pending before the framebuffer readback, most likely raised by {nameof(RenderOverride)} (or {nameof(Init)} on the first frame); GL errors should be checked and handled within user code. Further occurrences will be logged at Debug level.");
+				}
+				else if (this.Log().IsEnabled(LogLevel.Debug))
+				{
+					this.Log().Debug($"GL error {pendingError} was pending before the framebuffer readback.");
 				}
 			}
 
--- a/src/AddIns/Uno.WinUI.Graphics3DGL/GLCanvasElement.cs
+++ b/src/AddIns/Uno.WinUI.Graphics3DGL/GLCanvasElement.cs
@@ -587,6 +587,10 @@ public abstract partial class GLCanvasElement : Grid, INativeContext
 		// lost context.
 		for (var i = 0; i < 8 && gl.GetError() is not GLEnum.NoError; i++) { }
 
+		// Select the read buffer explicitly so the probe mirrors the readback in Render and
+		// doesn't depend on the caller having just bound a freshly created framebuffer.
+		gl.ReadBuffer(GLEnum.ColorAttachment0);
+
 		Span<byte> probe = stackalloc byte[BytesPerPixel];
 		gl.ReadPixels(0, 0, 1, 1, GLEnum.Bgra, GLEnum.UnsignedByte, probe);
 		return gl.GetError() is not GLEnum.NoError;
--- a/src/AddIns/Uno.WinUI.Graphics3DGL/GLCanvasElement.cs
+++ b/src/AddIns/Uno.WinUI.Graphics3DGL/GLCanvasElement.cs
@@ -493,6 +493,13 @@ public abstract partial class GLCanvasElement : Grid, INativeContext
 
 			_gl.ReadBuffer(GLEnum.ColorAttachment0);
 
+			// glReadPixels doesn't throw on failure (e.g. an unsupported readback format); it only
+			// sets a GL error and leaves the destination buffer untouched, which shows up as a
+			// permanently blank canvas. Drain errors left over from user code first so the check
+			// after the readback is attributable to the readback itself. The drain is bounded
+			// because GetError can keep returning GL_CONTEXT_LOST forever on a lost context.
+			for (var i = 0; i < 8 && _gl.GetError() is not GLEnum.NoError; i++) { }
+
 #if WINAPPSDK
 			_gl.ReadPixels(0, 0, (uint)RenderSize.Width, (uint)RenderSize.Height, GLEnum.Bgra, GLEnum.UnsignedByte, (void*)_pixels);
 			using (var stream = _backBuffer.PixelBuffer.AsStream())
@@ -514,6 +521,13 @@ public abstract partial class GLCanvasElement : Grid, INativeContext
 			});
 			_backBuffer.PixelBuffer.Length = (uint)RenderSize.Width * (uint)RenderSize.Height * BytesPerPixel;
 #endif
+
+			if (_gl.GetError() is var readbackError && readbackError is not GLEnum.NoError)
+			{
+				throw new InvalidOperationException(
+					$"glReadPixels failed with {readbackError} while copying the framebuffer to the back buffer.");
+			}
+
 			_backBuffer.Invalidate();
 		}
 		catch (Exception e)
--- a/src/AddIns/Uno.WinUI.Graphics3DGL/GLCanvasElement.cs
+++ b/src/AddIns/Uno.WinUI.Graphics3DGL/GLCanvasElement.cs
@@ -96,6 +96,12 @@ public abstract partial class GLCanvasElement : Grid, INativeContext
 	/// the end if you choose to enable it.
 	/// Some of this may be done for you automatically.
 	/// </remarks>
+	/// <remarks>
+	/// OpenGL errors raised by your rendering logic should be checked and handled within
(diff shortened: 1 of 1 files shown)


ramezgerges (approved):

src/AddIns/Uno.WinUI.Graphics3DGL/GLCanvasElement.cs
```diff
@@ -490,6 +493,13 @@ private unsafe void Render()
 
 			_gl.ReadBuffer(GLEnum.ColorAttachment0);
 
+			// glReadPixels doesn't throw on failure (e.g. an unsupported readback format); it only
+			// sets a GL error and leaves the destination buffer untouched, which shows up as a
+			// permanently blank canvas. Drain errors left over from user code first so the check
+			// after the readback is attributable to the readback itself. The drain is bounded
+			// because GetError can keep returning GL_CONTEXT_LOST forever on a lost context.
+			for (var i = 0; i < 8 && _gl.GetError() is not GLEnum.NoError; i++) { }
```
ramezgerges: This feels a bit hacky. It should probably be the user's job to drain their own errors.

src/AddIns/Uno.WinUI.Graphics3DGL/GLCanvasElement.cs
```diff
@@ -524,25 +541,34 @@ private unsafe void Render()
...
+			return false;
 		}
 
-		return true;
+		// The drain is bounded because GetError can keep returning GL_CONTEXT_LOST forever on a
+		// lost context.
+		for (var i = 0; i < 8 && gl.GetError() is not GLEnum.NoError; i++) { }
+
+		Span<byte> probe = stackalloc byte[BytesPerPixel];
+		gl.ReadPixels(0, 0, 1, 1, GLEnum.Bgra, GLEnum.UnsignedByte, probe);
```
ramezgerges, line 605: Why not use the existing GL_EXT_read_format_bgra probe?

src/AddIns/Uno.WinUI.Graphics3DGL/GLCanvasElement.cs
ajpinedam: Fair — the silent swallow was the hacky part. The drain itself is hard to avoid: GL errors are sticky and context-global, so without it a leftover error from `RenderOverride` would be misattributed to the readback check and permanently kill the canvas for what's actually a user bug. Changed it in e24235840c so the drain runs right after `RenderOverride` returns and logs a warning naming each drained error, and documented on `RenderOverride` that handling GL errors is the implementation's job.

src/AddIns/Uno.WinUI.Graphics3DGL/GLCanvasElement.cs
ajpinedam: That's what the previous code did (on Android), and it's exactly what fails on physical iOS devices: Apple's GLES-on-Metal driver advertises `GL_EXT_read_format_bgra` but still rejects BGRA reads from the element's FBO — see the "extension-string detection (rejected approach)" row in the PR description's validation table (iPad Pro M1 + A18 Pro iPhone). A real 1×1 read against the element's framebuffer is the only probe that reflects what the readback will actually do, and it subsumes the extension check on Android too (extension absent ⇒ the probe read errors ⇒ RGBA+swap).

src/AddIns/Uno.WinUI.Graphics3DGL/GLCanvasElement.cs
```diff
@@ -524,25 +562,34 @@ private unsafe void Render()
...
+		return gl.GetError() is not GLEnum.NoError;
```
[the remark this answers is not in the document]
ajpinedam: The scenario can't occur as sequenced: `READ_BUFFER` is per-FBO state, the probe runs against a freshly created FBO (whose read buffer is `COLOR_ATTACHMENT0` by spec-mandated default, GLES 3.0 §4.3.2 / state table 6.24) right after it's bound in `OnLoaded`, and before `Init` — so no user code can have changed it and `FrameBufferDetails` never touches it. That said, making the probe self-contained and identical to the real readback path in `Render` (`ReadBuffer` → `ReadPixels`) is worth it, so the explicit `glReadBuffer(GL_COLOR_ATTACHMENT0)` was added in 41d7418e3e as hardening.

src/AddIns/Uno.WinUI.Graphics3DGL/GLCanvasElement.cs
```diff
@@ -488,6 +497,25 @@ private unsafe void Render()
...
+			{
+				var pendingError = _gl.GetError();
+				if (pendingError is GLEnum.NoError)
+				{
+					break;
+				}
+
+				if (this.Log().IsEnabled(LogLevel.Warning))
+				{
+					this.Log().Warn($"{nameof(RenderOverride)} left GL error {pendingError} unhandled; GL errors should be checked and handled within {nameof(RenderOverride)}.");
```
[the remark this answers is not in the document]
ajpinedam: The claimed sources can't set the flag as sequenced: `glBindFramebuffer`/`glViewport` cannot error with the arguments passed here (a live generated FBO name; width/height are never negative), and a prior-frame error can't survive — every frame ends with drain → readback → a `GetError` check that throws and permanently stops rendering on any error, so any frame that runs started from a clean flag. The one real imprecision was first-frame leftovers from `Init` (user code, just not `RenderOverride`), so the warning and docs were hedged accordingly in fc6031b225 — no extra drains, since that would reintroduce silent mop-up for calls that can't realistically fail.

src/AddIns/Uno.WinUI.Graphics3DGL/GLCanvasElement.cs
```diff
@@ -488,6 +497,25 @@ private unsafe void Render()
...
+				if (pendingError is GLEnum.NoError)
+				{
+					break;
+				}
+
+				if (this.Log().IsEnabled(LogLevel.Warning))
+				{
+					this.Log().Warn($"GL error {pendingError} was pending before the framebuffer readback, most likely raised by {nameof(RenderOverride)} (or {nameof(Init)} on the first frame); GL errors should be checked and handled within user code.");
+				}
+			}
```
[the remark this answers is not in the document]
ajpinedam: Agreed on the flood risk, but a per-frame summary attacks the wrong multiplier: GL keeps one sticky flag per error *category*, so a frame typically drains 1–2 distinct errors, and the real driver is frame rate — with `Invalidate()` called from `RenderOverride`, one warning per frame is still 60–120/sec. Addressed in c68ca7c100 by throttling per element instead: the first drained error logs the full warning (which now notes that further occurrences move to Debug), subsequent ones log at Debug level, so persistence stays traceable without drowning the log.

PR #23895, opened 2026-07-28
fix(skia): Restore Compositor.IsSoftwareRenderer

### LINKED ISSUE #23896: [Skia] Compositor.IsSoftwareRenderer is never set, so CompositionCapabilities.AreEffectsFast() is always true
### Current behavior 🐛

`Compositor.IsSoftwareRenderer` is never set by any Skia render backend, so it stays `null` for the whole app lifetime and every consumer treats the target as hardware-rendered regardless of the actual renderer:

- **`CompositionCapabilities.AreEffectsFast()` always returns `true`**, even when the scene is rasterized on the CPU (software fallback on Win32/X11, `UseMetalOnMacOS = false`, WASM without WebGL, framebuffer without DRM, Android without OpenGL). Apps consulting it to decide whether to enable effects (as recommended on WinUI) get the wrong answer, and so does anything keying off `CompositionCapabilities.Changed`.
- `CompositionEffectBrush` and the acrylic brush implementation consult the flag while recording the scene to generate `SKImageFilter`s the target surface can rasterize; on software targets they currently take the GPU path.

This is a regression: before the rendering refactor to the record/replay architecture (b96ad06bba), each backend wrapped its paint with `Compositor.IsSoftwareRenderer = renderer.IsSoftware()`. The refactor dropped all setters — only `RenderTargetBitmap` still sets the flag, temporarily, around software re-records. The Win32 backend even still carries a per-renderer `IRenderer.IsSoftware()` that nothing consumes.

### Expected behavior 🎯

`Compositor.IsSoftwareRenderer` accurately reflects whether the selected renderer rasterizes on the CPU, on every Skia target and including GPU-init fallback paths, so `CompositionCapabilities.AreEffectsFast()` and the effect brushes behave according to the actual rendering mode.

On any Skia target, force software rendering (e.g. `FeatureConfiguration.Rendering.UseOpenGLOnWin32 = false` on Windows) and evaluate `CompositionCapabilities.GetForCurrentView().AreEffectsFast()` — it returns `true` where it should return `false`. Internally, `Compositor.GetSharedCompositor().IsSoftwareRenderer` remains `null` no matter the backend or renderer.

### Workaround 🛠️

None from app code — the flag is internal and nothing sets it.

### Renderer 🎨

Skia

### Affected platforms 📱💻🖥️

All platforms 🌍

master

### IDE version 🧑‍💻

N/A

Regression introduced by b96ad06bba (`chore: XamlRoot.OnNativePlatformFrameRequested`).

ramezgerges:
**GitHub Issue:** closes #23896

## PR Type:

🐞 Bugfix

## What changed? 🚀

`Compositor.IsSoftwareRenderer` has been permanently `null` since the rendering refactor to the record/replay architecture (b96ad06bba) dropped the per-paint setters, so every consumer — `CompositionEffectBrush`, `SkiaAcrylicBrush`, `CompositionCapabilities.AreEffectsFast()` — has been treating all targets as hardware-rendered, even when the scene is rasterized on the CPU. Only `RenderTargetBitmap` still set it, temporarily, around software re-records.

This PR makes every Skia backend set the flag at the point it commits to a renderer, including GPU-init fallback paths:

| Backend | Set where | Value |
|---|---|---|
| Win32 | `Win32WindowWrapper` after the Vulkan → OpenGL → Software cascade | existing `IRenderer.IsSoftware()` (previously unconsumed) |
| X11 | `X11XamlRootHost` after the Vulkan → GLX/EGL → Software cascade | `renderer is X11SoftwareRenderer` |
| macOS | `MacSkiaHost.InitializeMac` once `RenderSurfaceType` resolves | `RenderSurfaceType == Software` |
| WASM | `BrowserRenderer` ctor after WebGL/Software selection | `renderer is SoftwareBrowserRenderer` |
| Linux FrameBuffer | `FrameBufferHost` after DRM/Software selection | `renderer is SoftwareRenderer` |
| Android (GL) | `UnoSKCanvasView` ctor | `!UseOpenGLOnSkiaAndroid` — the scene is CPU-rasterized then blitted through GL, so it reports software |
| Android (Vulkan) | `UnoSKVulkanView` ctor | `false` |
| Apple UIKit | `UnoSKMetalView` ctor after Metal context creation | `false` (Metal-only) |

The flag matters at SKPicture **recording** time on the UI thread (effect brushes consult it to generate filters the target surface can rasterize), so setting it once at renderer selection is sufficient — renderer choice is fixed after startup on all backends. The `bool?` type and `RenderTargetBitmap`'s save/restore semantics are preserved, and the property now carries doc comments describing the contract.

Also adds a runtime test asserting the flag is populated on all Skia targets (fails before this change everywhere, passes after).

## PR Checklist ✅

### DIFF
--- a/src/Uno.UI.Composition/Composition/Compositor.skia.cs
+++ b/src/Uno.UI.Composition/Composition/Compositor.skia.cs
@@ -29,6 +29,12 @@ public partial class Compositor
 		UnoSkiaApi.Initialize();
 	}
 
+	/// <summary>
+	/// Whether the scene is rasterized on the CPU rather than by a GPU-backed surface.
+	/// Set by the active render backend once its renderer is selected; null until then.
+	/// Consulted while recording (e.g. by effect brushes to generate filters the target
+	/// surface can rasterize) and temporarily overridden by RenderTargetBitmap.
+	/// </summary>
 	internal bool? IsSoftwareRenderer { get; set; }
 
 	internal static bool SkipVisualTreePainting { get; set; }
--- a/src/Uno.UI.Runtime.Skia.Android/Rendering/UnoSKCanvasView.cs
+++ b/src/Uno.UI.Runtime.Skia.Android/Rendering/UnoSKCanvasView.cs
@@ -34,6 +34,9 @@ internal sealed partial class UnoSKCanvasView : GLSurfaceView, IUnoSkiaRenderVie
 		SetEGLContextClientVersion(2);
 		SetEGLConfigChooser(8, 8, 8, 8, 0, 8);
 		SetRenderer(_renderer = new InternalRenderer());
+		// The scene is still presented through GL, but when not hardware-accelerated it is
+		// rasterized on the CPU first, which is what IsSoftwareRenderer reflects.
+		Microsoft.UI.Composition.Compositor.GetSharedCompositor().IsSoftwareRenderer = !_renderer.HardwareAccelerated;
 		ExploreByTouchHelper = new UnoExploreByTouchHelper(this);
 		TextInputPlugin = new TextInputPlugin(this);
 		ViewCompat.SetAccessibilityDelegate(this, ExploreByTouchHelper);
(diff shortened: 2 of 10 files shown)


ajpinedam (approved):
My only comment/concern is that when a new platform/host is added this might get forgotten and will default to false.

ramezgerges:
To cover this, the PR includes a runtime test that will fail if the property is unset.

PR #23928, opened 2026-07-31
refactor(xaml)!: Remove `clr-namespace:` XAML leniency (BC42)

Xiaoy312:
**GitHub Issue:** closes #

## PR Type:

## What changed? 🚀

**BC42 — remove the `clr-namespace:` XAML leniency.**

WinUI/UWP only supports the `using:MyApp.Controls` xmlns form. Uno additionally accepted the
WPF/Silverlight `clr-namespace:MyApp.Controls;assembly=MyLib` form, silently stripping the prefix
and the `;assembly=` token before resolving types. That let non-portable XAML build on Uno and then
fail when the same markup was compiled for WinAppSDK.

The declaration is now **rejected outright**, with no deprecation cycle:

- **Build time** — the new `UXAML0006` diagnostic, reported with the file and line of the
  declaration.
- **Run time** — `XamlReader.Load` (and therefore Hot Reload) throws a `XamlParseException`.

The only exemption is a prefix listed in `mc:Ignorable` on the root element.

### Notable implementation details

- **The check fires on the declaration, not on first use.** A `clr-namespace:` xmlns is invalid
  WinUI markup whether or not the prefix is referenced, and rejecting at the declaration gives an
  actionable message from a single choke point per parser instead of threading diagnostic locations
  through four resolution call sites.
- **Validation is deferred to the end of the parse.** The XAML reader yields `NamespaceDeclaration`
  nodes *before* it reads the element's `mc:Ignorable` attribute, so declarations are collected
  during the visit and reported once the ignorable set is known.
- **Simply deleting the stripping would not have been a hard error.** Unresolved prefixed types fall
  through to `SearchWithFuzzyMatching`, which can still resolve by simple name — and fuzzy matching
  is enabled in `XamlGenerationTests`. An explicit diagnostic is required.
- **Conditional-XAML prefixes are deliberately not exempt** (e.g. `xmlns:android="clr-namespace:…"`).
  On the included platform `IsIncluded` already rewrites the URI to the presentation namespace,
  discarding the CLR namespace; on the excluded platform the elements are dropped. That markup never
  did what its author intended, so the error is a true positive.
- **`src/SourceGenerators/System.Xaml/` is untouched.** Its `clr-namespace:` handling is already
  unreachable on Uno's paths (both entry points construct `XamlSchemaContext` with an empty assembly
  list, which gates `ResolveXamlTypeName` off). `Uno.Xaml.Tests` stays green.
- Repo XAML migrated to `using:`. Only two files actually used the prefix; the rest declared one they
  never referenced. `themeresources_v2.xaml` is generated by `FluentMerge.targets` and was
  regenerated, not hand-edited.

### Validation

| | Result |
|---|---|
| Compile — `Uno.UI-UnitTests-only.slnf` | 0 errors |
| Compile — `SamplesApp.Skia.Generic` (Debug + Release) | 0 errors |
| Test — `Uno.Xaml.Tests` (vendored parser) | 791 total, 0 failed |
| Test — `Uno.UI.SourceGenerators.Tests` | 479 total, 0 failed (4 new `UXAML0006` cases) |
| Test — `Uno.UI.UnitTests` | 4036 total, 13 failed — all pre-existing timezone-dependent `Windows_Globalization` Calendar tests, unrelated |
| **Runtime — `Given_XamlReader`, Skia Desktop** | **54/54 passed**, including both new cases |

**Fails-before / passes-after:** with the two runtime files stashed and rebuilt, the two rejection
tests fail and the `mc:Ignorable` test still passes; with the change applied all three pass.

Test coverage spans a used prefix, an unused prefix, an inline-scope declaration (xmlns on an element
inside a member, e.g. a `DataTemplate`), and an `mc:Ignorable` prefix that must keep working.

## PR Checklist ✅

### Breaking change — impact and migration path

This is an intentional breaking change, part of the 7.0 breaking-change rollup (Phase 5), targeting
the `feature/breakingchanges` branch.

**Impact:** any app or library whose XAML declares a `clr-namespace:` xmlns now fails to build with
`UXAML0006`, and fails at run time if the markup reaches `XamlReader.Load`. Unused declarations are
rejected too.

**Migration:**

```diff
- xmlns:local="clr-namespace:MyApp.Controls;assembly=MyLib"
+ xmlns:local="using:MyApp.Controls"
```

The assembly is inferred from the compilation, so the `;assembly=` token has no replacement — drop
it. Prefixes listed in `mc:Ignorable` on the root element are unaffected.

Documented in `doc/articles/migrating-to-uno-7.md` (new *XAML changes* section plus a migration
checklist step) and `doc/articles/uno-build-error-codes.md` (`UXAML0006`).


src/SamplesApp/SamplesApp.Samples/Windows_UI_Xaml_Media/ImageBrushTests/EllipsemaskingEllipseGrid.xaml
```diff
@@ -9,7 +9,6 @@
     xmlns:local="using:Uno.UI.Samples.UITests"
     xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
     xmlns:not_ios="http://uno.ui/not_ios"
-    xmlns:shapes="clr-namespace:Microsoft.UI.Xaml.Shapes;assembly=Uno"
```
ajpinedam, line 12: should these be replaced with their `using:` equivalent instead of simply removing it?

src/SamplesApp/SamplesApp.Samples/Windows_UI_Xaml_Media/ImageBrushTests/ImageBrushAlignmentXY.xaml
```diff
@@ -9,7 +9,6 @@
     xmlns:local="using:Uno.UI.Samples.UITests"
     xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
     xmlns:not_ios="http://nventive.com/not_ios"
-    xmlns:shapes="clr-namespace:Uno.UI.Xaml.Shapes;assembly=nVentive.Umbrella"
```
ajpinedam, line 12: same here?

src/SamplesApp/SamplesApp.Samples/Windows_UI_Xaml_Media/ImageBrushTests/ImageBrushStretch.xaml
```diff
@@ -9,7 +9,6 @@
     xmlns:local="using:Uno.UI.Samples.UITests"
     xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
     xmlns:not_ios="http://nventive.com/not_ios"
-    xmlns:shapes="clr-namespace:Uno.UI.Xaml.Shapes;assembly=nVentive.Umbrella"
```
ajpinedam, line 12: and here?

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlCodeBehindParser.cs
```diff
@@ -13,6 +13,8 @@ namespace Uno.UI.SourceGenerators.XamlGenerator;
 /// </summary>
 internal static class XamlCodeBehindParser
 {
+	private const string UsingPrefix = "using:";
```
ajpinedam, line 16: `static readonly` to keep consistency?

ajpinedam (approved):
Good!

src/SamplesApp/SamplesApp.Samples/Windows_UI_Xaml_Media/ImageBrushTests/EllipsemaskingEllipseGrid.xaml
Xiaoy312: these are dead xmlns

src/SamplesApp/SamplesApp.Samples/Windows_UI_Xaml_Media/ImageBrushTests/ImageBrushAlignmentXY.xaml
Xiaoy312: unsused xmlns

src/SamplesApp/SamplesApp.Samples/Windows_UI_Xaml_Media/ImageBrushTests/ImageBrushStretch.xaml
Xiaoy312: unsused xmlns

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlCodeBehindParser.cs
Xiaoy312: we should use `const`; 
reason why we cant use `const` below is because they are doing a implicit-cast from string to XNamespace:
```cs
XNamespace XamlNamespace = "http://schemas.microsoft.com/winfx/2006/xaml";
```
and, in our case here, there is absolutely no reason, to not use `const`
in fact, analyzers might flag it if we did otherwise

PR #23931, opened 2026-07-31
feat(skills): Add wasm-device-test skill for on-device WASM testing

### LINKED ISSUE #23930: Add a skill for testing WebAssembly apps on physical devices over the LAN

A repository skill (`/wasm-device-test`) that publishes an Uno Platform WebAssembly head and serves the published output on the local network, so it can be opened on a phone, tablet, or another machine.

Today there is no convenient path for this:

- `dotnet run` serves on `localhost` only, so a phone cannot reach it.
- `python -m http.server` (what the WASM runtime-test script uses) binds loopback only, *and* does not know the `.wasm` MIME type, so the browser cannot use `WebAssembly.instantiateStreaming`.

### Why is this needed? 🤔

A significant class of WebAssembly behaviour cannot be reproduced in a desktop browser:

- touch and gesture handling, inertia and fling behaviour
- device refresh rate — many phones idle at 60 Hz and boost to 120 Hz only while touch is active, so an animation can run at a different cadence than the interaction that started it
- mobile GPU rasterization cost, which is where per-frame CPU/GPU work actually becomes visible
- small viewports, safe-area insets, and on-screen keyboard behaviour

Investigating any of these currently means re-deriving the same publish-and-serve setup by hand, including the MIME-type and firewall pitfalls, which are easy to mistake for the app itself being broken.

### Renderer 🎨

- [x] Skia

### For which platforms 📱💻🖥️

WebAssembly

Scope is contributor tooling only — no product code is affected. The skill:

- publishes **Release** by default, since a Debug WebAssembly build is dominated by interpreter overhead and any conclusion about responsiveness drawn from it is meaningless
- registers the .NET WebAssembly MIME types and sends `Cache-Control: no-store`, so a republished app is picked up without clearing the service worker
- probes the LAN address before reporting a URL, because a blocking host firewall is the usual failure mode and is otherwise indistinguishable from the app not loading

MartinZikmund:
**GitHub Issue:** closes #23930

## PR Type:

🤖 Project automation

## What changed? 🚀

Adds a `/wasm-device-test` skill that publishes an Uno Platform WebAssembly head and serves the published output on the local network, so it can be opened on a phone, tablet, or another machine.

Neither existing option works for this:

- `dotnet run` serves on `localhost` only, so a device cannot reach it.
- `python -m http.server` (used by `build/test-scripts/wasm-run-skia-runtime-tests.sh`) binds loopback only, *and* does not register the `.wasm` MIME type, so the browser cannot use `WebAssembly.instantiateStreaming`.

The skill covers the whole loop — publish, serve, verify, report:

- **Publishes Release by default.** A Debug WebAssembly build is dominated by interpreter overhead, so any conclusion about responsiveness or frame rate drawn from it is meaningless. Debug is opt-in.
- **Serves on all interfaces** with the .NET WebAssembly MIME types registered, and `Cache-Control: no-store` so a republished app is picked up without clearing the service worker.
- **Probes the LAN address before reporting a URL.** A blocking host firewall is the usual failure and is otherwise indistinguishable from the app not loading. The skill reports the address that actually answered rather than listing every adapter, and gives the `New-NetFirewallRule` command rather than silently elevating.

It documents the limitations that bite in practice: HTTP only (so secure-context APIs are unavailable), AP isolation on guest and corporate Wi-Fi, and that this is for driving an app by hand — automated WASM runtime tests remain the `/runtime-tests` skill's job.

Contributor tooling only; no product code is touched. `AGENTS.md` gains a row in the skills table.

### Verification

The bundled `scripts/serve-wasm.py` was exercised against a published `SamplesApp.Skia.WebAssembly.Browser` output:

- `index.html` → 200 on both `localhost` and the LAN address
- `_framework/*.wasm` → `Content-Type: application/wasm`
- `Cache-Control: no-store, must-revalidate` present
- the app was loaded and driven by hand from a phone browser on the same network

## PR Checklist ✅

- [x] 🧪 Added [Runtime tests, UI tests, or a manual test sample](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) (for bug fixes / features, if applicable) — N/A, contributor tooling; the script was validated manually as described above
- [x] 📚 Docs have been added/updated following the [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features) — the skill is self-documenting and is registered in `AGENTS.md`
- [x] 🖼️ Validated PR `Screenshots Compare Test Run` results — N/A, no product code

### DIFF
--- a/.claude/skills/wasm-device-test/SKILL.md
+++ b/.claude/skills/wasm-device-test/SKILL.md
@@ -1,0 +1,157 @@
+---
+description: Publish an Uno WebAssembly app and serve it on the local network so it can be opened on a phone, tablet, or another machine. Use when testing WASM behaviour that cannot be reproduced on a desktop browser — touch input, mobile GPUs, device refresh rates, small viewports.
+---
+
+## User Input
+
+```text
+$ARGUMENTS
+```
+
+You **MUST** consider the user input before proceeding (if not empty).
+
+---
+
+## Overview
+
+You are executing the **WASM Device Test Skill**. It publishes an Uno Platform WebAssembly app and
+serves the published output over HTTP **bound to all network interfaces**, so a phone or tablet on
+the same network can load it.
+
+This exists because the two obvious approaches both fail for device testing:
+
+- `dotnet run` serves on `localhost` only — a phone cannot reach it.
+- `python -m http.server` binds loopback only *and* does not know the `.wasm` MIME type, so the
+  browser cannot use `WebAssembly.instantiateStreaming`.
+
+Use it when the thing under test is device-specific: touch and gesture handling, inertia and fling
+behaviour, display refresh rate (many phones idle at 60 Hz and boost to 120 Hz only while touch is
+active), GPU rasterization, safe-area insets, or on-screen keyboard behaviour.
+
+---
+
+## Execution Workflow
+
+### Phase 0: Parse User Input
+
+| Input | Meaning | Default |
+|---|---|---|
+| a project path or name | the WASM head to publish | auto-detect (below) |
+| `debug` | publish Configuration=Debug | **Release** |
+| `port <n>` | HTTP port | `8123` |
+| `no-build` / `serve-only` | skip publish, serve the existing output | publish first |
+
+**Always publish Release unless the user explicitly asks for Debug.** A Debug WebAssembly build is
+dominated by interpreter overhead, so any conclusion about smoothness, frame rate, or responsiveness
+drawn from it is meaningless.
+
+**Auto-detecting the project.** Look for a WebAssembly head in this order:
+
+1. A project the user named.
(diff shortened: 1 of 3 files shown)


morning4coffe-dev (commented):
There is no microsoft-ui-xaml equivalent for this contributor workflow. A couple of the guarantees need narrowing: modern `python -m http.server` defaults to all interfaces, and the Python installation here already maps .wasm to application/wasm, so those are not universal deficiencies of the built-in server. More importantly, Cache-Control: no-store controls HTTP caching, not an existing service worker's Cache Storage response strategy; document clearing/unregistering stale service workers when needed. A host successfully requesting its own LAN address also does not prove phone-to-host reachability through the firewall/AP. I reviewed the workflow without starting its LAN server.

PR #23863, opened 2026-07-23
fix(hr): align workspace analyzer flavors and Roslyn with the SDK compiler (spec 053)

dr1rrb:
**GitHub Issue:** closes #23860

## PR Type:

🐞 Bugfix

## What changed? 🚀

Implements [`specs/053-hotreload-workspace-analyzer-flavor-skew/spec.md`](https://github.com/unoplatform/uno/blob/dev/dr/hr-workspace-analyzer-flavor-skew/specs/053-hotreload-workspace-analyzer-flavor-skew/spec.md) (carried by this PR):

- **Analyzer flavor pin (R1)** — the hot-reload workspace now passes `CompilerApiVersion=roslyn{X.Y}` of its **embedded** Microsoft.CodeAnalysis as a global property (computed at runtime — `EmbeddedRoslyn.CompilerApiVersion`), so the standard analyzer multi-targeting (`analyzers/dotnet/roslyn{X.Y}` folders) selects flavors the workspace can actually load. Without it, an SDK newer than the embedded Roslyn (SDK 10 → `roslyn5.6`) selects flavors that fail to type-load and `AnalyzerFileReference.GetGenerators()` **silently returns zero generators** — every workspace compile of the affected projects then misses its generated code (measured: 229 phantom errors on a single CommunityToolkit.Mvvm 8.4.2 library, CS0759/CS0103/CS1061 walls).
- **Embedded Roslyn bumps (R2)** — net9 flavor `4.13.0` → `4.14.0` (latest 4.x, SDK 9 compiler line), net10 flavor `4.14.0` → `5.6.0` (latest 5.x, SDK 10 line; brings C# 14 parsing and native `roslyn5.x` flavor loading). Companion compile-time pins in the net10 groups: `Microsoft.Build*` 18.0.2 (`ExcludeAssets="runtime"`), `Microsoft.Extensions.Logging*` 10.0.1.
- **EnC shim retarget** — Roslyn removed `ExternalAccess.Watch` from `Microsoft.CodeAnalysis.Features` between 5.0 and 5.3; the reflection shim now targets the shape-identical `ExternalAccess.UnitTesting.Api.UnitTestingHotReloadService` (identical from 4.14 to 5.6 — capabilities move to `StartSessionAsync`, and `commitUpdates: true` reproduces Watch's implicit commit-on-ready, confirmed against the Roslyn sources). The obsoleted `Workspace.WorkspaceFailed` event is replaced by `RegisterWorkspaceFailedHandler` on the net10 flavor.
[...]

PR #23947, opened 2026-08-03
fix(hr): harden the collectible analyzer loader (review follow-up of #23863)

dr1rrb:
Follow-up to #23863 — the review-round-4 changes: the PR auto-merged the moment its last review conversations were resolved, before these fixes were pushed, so the "Done" replies on those threads describe the changes delivered here.

## Changes (all on `CollectibleAnalyzerAssemblyLoader`)

- **Delegate resolution to the compiler context** (https://github.com/unoplatform/uno/pull/23863#discussion_r3692410010): `Load` now tries `_compilerContext.LoadFromAssemblyName(assemblyName)` (bare-catch fall-through, the exact shape of Roslyn's own loader) instead of probing the context's already-loaded `Assemblies`. Dependencies the compiler context can resolve but has not loaded yet materialize in it — no split-identity analyzer-local copy — and resolution stays correct when assemblies keep appearing at runtime (e.g. NuGet packages added to the app through hot reload). This also removes the per-`Load` linear scan discussed in https://github.com/unoplatform/uno/pull/23863#discussion_r3686034480 without any cache to invalidate: the ALC's internal by-name lookup is authoritative and re-probed by the runtime on every miss.
- **Shadow copies only ever become visible complete** (https://github.com/unoplatform/uno/pull/23863#discussion_r3692410015): the copy lands in a unique staging file and is published with an atomic `File.Move`; the PDB is published before the DLL so a visible DLL means the bundle is whole. Rename-publish was chosen over an in-process `ConcurrentDictionary<string, Task<string>>` because the shadow key is deterministic and the temp root shared: the race also exists across dev-server processes, and a process killed mid-copy would otherwise poison the key with a permanent partial file.
- **Best-candidate dependency resolution** (https://github.com/unoplatform/uno/pull/23863#discussion_r3692410019): `AddDependencyLocation` retains every location registered for a file name (distinct projects may carry distinct versions of the same analyzer package), and resolution mirrors Roslyn's `GetBestPath`: exact requested version wins, otherwise the highest version with a compatible identity (simple name + public key token when requested).
- **`Path.Join` throughout** the file and the test helper (CodeQL findings).

## Validation

- `Uno.HotReload.Tests`: 120/120 passing, including two new regression tests — `When_CompilerContextResolvesLazily_Then_AnalyzerRequestsUnifyToIt` (lazily-resolving compiler context + same-directory copy of the assembly: the request must materialize in the compiler context; fails on the previous enumeration-based code) and `When_MultipleVersionsRegistered_Then_RequestedVersionWins` (two registered versions: the exact requested one must win over the first registered).
- `Uno.HotReload` compiles warning-free on both flavors: net9.0 (Roslyn 4.14) and net10.0 (Roslyn 5.6).

### DIFF
--- a/src/Uno.HotReload.Tests/Roslyn/Given_CollectibleAnalyzerAssemblyLoader.cs
+++ b/src/Uno.HotReload.Tests/Roslyn/Given_CollectibleAnalyzerAssemblyLoader.cs
@@ -39,6 +39,7 @@ public sealed class Given_CollectibleAnalyzerAssemblyLoader
 			Assert.IsNotNull(context);
 			Assert.IsTrue(context.IsCollectible, "analyzer contexts must be collectible");
 			Assert.AreSame(context, AssemblyLoadContext.GetLoadContext(assembly2), "same directory must share one context");
+			Assert.AreSame(assembly1, loader.LoadFromPath(payload1), "same path must keep yielding the same assembly instance");
 		}
 		finally
 		{
@@ -141,7 +142,7 @@ public sealed class Given_CollectibleAnalyzerAssemblyLoader
 	{
 		var root = Directory.CreateTempSubdirectory("uno-hr-tests").FullName;
 		var csharpPath = typeof(CSharpCompilation).Assembly.Location;
-		var compilerContext = new OnDemandContext(new() { ["Microsoft.CodeAnalysis.CSharp"] = csharpPath });
+		var compilerContext = new OnDemandContext(new Dictionary<string, string> { ["Microsoft.CodeAnalysis.CSharp"] = csharpPath });
 		try
 		{
 			var payload = CopyTo(root, typeof(Given_CollectibleAnalyzerAssemblyLoader).Assembly.Location);
@@ -224,6 +225,66 @@ public sealed class Given_CollectibleAnalyzerAssemblyLoader
 		}
 	}
 
+	[TestMethod]
+	[Description(
+		"A registered candidate whose identity does not match the request (public key token) " +
+		"must be skipped — not force-loaded because its file name matches — leaving the " +
+		"request to the runtime's own resolution.")]
+	public void When_RequestedTokenDoesNotMatch_Then_CandidatesAreSkipped()
+	{
+		var root = Directory.CreateTempSubdirectory("uno-hr-tests").FullName;
+		try
+		{
+			var v1 = EmitAssembly(Path.Join(root, "v1", "Dep.dll"), "Dep", new Version(1, 0, 0, 0));
+			var v2 = EmitAssembly(Path.Join(root, "v2", "Dep.dll"), "Dep", new Version(2, 0, 0, 0));
+			var payload = CopyTo(Directory.CreateDirectory(Path.Join(root, "analyzer")).FullName, typeof(Given_CollectibleAnalyzerAssemblyLoader).Assembly.Location);
+
+			var loader = new CollectibleAnalyzerAssemblyLoader();
+			loader.AddDependencyLocation(v1);
+			loader.AddDependencyLocation(v2);
+			var analyzerContext = AssemblyLoadContext.GetLoadContext(loader.LoadFromPath(payload))!;
+
+			// Both candidates are unsigned: a request carrying a public key token must skip them
+			// (and then fail in the runtime's own resolution, since nobody can provide it).
+			Assert.ThrowsExactly<FileNotFoundException>(() =>
+				analyzerContext.LoadFromAssemblyName(new AssemblyName("Dep, Version=2.0.0.0, PublicKeyToken=b03f5f7f11d50a3a")));
+		}
+		finally
+		{
+			Directory.Delete(root, recursive: true);
+		}
+	}
+
+	[TestMethod]
+	[Description(
+		"A shadow directory holding partial state — an orphaned staging file and a pdb " +
+		"published without its dll, i.e. a process killed mid-publish — must not prevent the " +
+		"load: the dll's atomic publish is what completes the bundle.")]
+	public void When_ShadowDirectoryHoldsPartialState_Then_LoadRecovers()
+	{
+		var root = Directory.CreateTempSubdirectory("uno-hr-tests").FullName;
+		try
+		{
+			var payload = CopyTo(root, typeof(Given_CollectibleAnalyzerAssemblyLoader).Assembly.Location);
+
+			// Re-create the debris a kill between the pdb and dll publishes would leave behind.
+			var shadowPath = CollectibleAnalyzerAssemblyLoader.GetShadowCopyPath(payload);
+			Assert.IsFalse(File.Exists(shadowPath), "premise: this (path, timestamp, size) key must be fresh");
+			Directory.CreateDirectory(Path.GetDirectoryName(shadowPath)!);
+			File.WriteAllText($"{shadowPath}.{Guid.NewGuid():N}.staging", "debris from a killed copy");
+			File.WriteAllBytes(Path.ChangeExtension(shadowPath, ".pdb"), [1, 2, 3]);
+
+			var loaded = new CollectibleAnalyzerAssemblyLoader().LoadFromPath(payload);
+
+			Assert.IsNotNull(loaded);
+			Assert.IsTrue(File.Exists(shadowPath), "the dll must have been published despite the pre-existing debris");
+		}
+		finally
+		{
+			Directory.Delete(root, recursive: true);
+		}
+	}
+
 	private static string CopyTo(string directory, string sourcePath, string? fileName = null)
 	{
 		var target = Path.Join(directory, fileName ?? Path.GetFileName(sourcePath));
@@ -254,7 +315,7 @@ public sealed class Given_CollectibleAnalyzerAssemblyLoader
 	/// A compiler-side context that — like the dev-server's per-application context — can
 	/// RESOLVE assemblies it has not loaded yet (lazy, on demand).
 	/// </summary>
-	private sealed class OnDemandContext(Dictionary<string, string> resolvable)
+	private sealed class OnDemandContext(IReadOnlyDictionary<string, string> resolvable)
 		: AssemblyLoadContext("test-on-demand-compiler", isCollectible: true)
 	{
 		protected override Assembly? Load(AssemblyName assemblyName)
--- a/src/Uno.HotReload.Tests/Roslyn/Given_CollectibleAnalyzerAssemblyLoader.cs
+++ b/src/Uno.HotReload.Tests/Roslyn/Given_CollectibleAnalyzerAssemblyLoader.cs
@@ -1,6 +1,7 @@
 using System.Reflection;
 using System.Runtime.Loader;
 using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.Diagnostics;
 using Uno.HotReload.Roslyn;
 using Uno.HotReload.Tests.TestUtils;
@@ -131,11 +132,103 @@ public sealed class Given_CollectibleAnalyzerAssemblyLoader
 		Assert.IsTrue(AssemblyLoadContext.GetLoadContext(loaded)!.IsCollectible, "the swapped loader must load collectible");
 	}
 
+	[TestMethod]
+	[Description(
+		"Resolution must be DELEGATED to the compiler's context (LoadFromAssemblyName) so its " +
+		"load policy can materialize dependencies it has not loaded YET: probing only its " +
+		"already-loaded assemblies would fall through and load a split-identity local copy.")]
+	public void When_CompilerContextResolvesLazily_Then_AnalyzerRequestsUnifyToIt()
+	{
+		var root = Directory.CreateTempSubdirectory("uno-hr-tests").FullName;
+		var csharpPath = typeof(CSharpCompilation).Assembly.Location;
+		var compilerContext = new OnDemandContext(new() { ["Microsoft.CodeAnalysis.CSharp"] = csharpPath });
+		try
+		{
+			var payload = CopyTo(root, typeof(Given_CollectibleAnalyzerAssemblyLoader).Assembly.Location);
+			CopyTo(root, csharpPath); // A same-directory copy that must NOT win over the compiler's.
+			var loader = new CollectibleAnalyzerAssemblyLoader(compilerContext);
+			var analyzerContext = AssemblyLoadContext.GetLoadContext(loader.LoadFromPath(payload))!;
+
+			Assert.IsFalse(compilerContext.Assemblies.Any(), "premise: the compiler context resolves lazily, nothing loaded yet");
+
+			var resolved = analyzerContext.LoadFromAssemblyName(new AssemblyName("Microsoft.CodeAnalysis.CSharp"));
+
+			Assert.AreSame(compilerContext, AssemblyLoadContext.GetLoadContext(resolved),
+				"the request must materialize the dependency IN the compiler's context, not load the analyzer-local copy");
+		}
+		finally
+		{
+			compilerContext.Unload();
+			Directory.Delete(root, recursive: true);
+		}
+	}
+
+	[TestMethod]
+	[Description(
+		"AddDependencyLocation must retain EVERY location registered for a file name (distinct " +
+		"projects may use distinct versions of the same analyzer package): resolution picks the " +
+		"version matching the request, not whichever project registered that file name first.")]
+	public void When_MultipleVersionsRegistered_Then_RequestedVersionWins()
+	{
+		var root = Directory.CreateTempSubdirectory("uno-hr-tests").FullName;
+		try
+		{
+			var v1 = EmitAssembly(Path.Join(root, "v1", "Dep.dll"), "Dep", new Version(1, 0, 0, 0));
+			var v2 = EmitAssembly(Path.Join(root, "v2", "Dep.dll"), "Dep", new Version(2, 0, 0, 0));
+			var payload = CopyTo(Directory.CreateDirectory(Path.Join(root, "analyzer")).FullName, typeof(Given_CollectibleAnalyzerAssemblyLoader).Assembly.Location);
+
+			var loader = new CollectibleAnalyzerAssemblyLoader();
+			loader.AddDependencyLocation(v1); // Registered FIRST: "first file name wins" would pick this one.
+			loader.AddDependencyLocation(v2);
+			var analyzerContext = AssemblyLoadContext.GetLoadContext(loader.LoadFromPath(payload))!;
+
+			var resolved = analyzerContext.LoadFromAssemblyName(new AssemblyName("Dep, Version=2.0.0.0"));
+
+			Assert.AreEqual(new Version(2, 0, 0, 0), resolved.GetName().Version, "the exact requested version must win over the first registered one");
+		}
+		finally
+		{
+			Directory.Delete(root, recursive: true);
+		}
+	}
+
 	private static string CopyTo(string directory, string sourcePath, string? fileName = null)
 	{
-		var target = Path.Combine(directory, fileName ?? Path.GetFileName(sourcePath));
+		var target = Path.Join(directory, fileName ?? Path.GetFileName(sourcePath));
 		File.Copy(sourcePath, target);
 		return target;
+	}
+
+	/// <summary>
+	/// Emits a minimal assembly with the requested identity — the best-candidate resolution
+	/// reads the real <see cref="AssemblyName"/> from disk, so the fixture needs real PE files.
+	/// </summary>
+	private static string EmitAssembly(string path, string name, Version version)
+	{
+		Directory.CreateDirectory(Path.GetDirectoryName(path)!);
+
+		var compilation = CSharpCompilation.Create(
+			name,
+			[CSharpSyntaxTree.ParseText($"""[assembly: System.Reflection.AssemblyVersion("{version}")]""")],
+			references: [MetadataReference.CreateFromFile(typeof(object).Assembly.Location)],
+			options: new CSharpCompilationOptions(OutputKind.DynamicallyLinkedLibrary));
+
+		var emit = compilation.Emit(path);
+		Assert.IsTrue(emit.Success, $"Fixture assembly emit failed: {string.Join(", ", emit.Diagnostics)}");
+		return path;
+	}
+
+	/// <summary>
+	/// A compiler-side context that — like the dev-server's per-application context — can
+	/// RESOLVE assemblies it has not loaded yet (lazy, on demand).
+	/// </summary>
+	private sealed class OnDemandContext(Dictionary<string, string> resolvable)
+		: AssemblyLoadContext("test-on-demand-compiler", isCollectible: true)
+	{
+		protected override Assembly? Load(AssemblyName assemblyName)
+			=> assemblyName.Name is { } name && resolvable.TryGetValue(name, out var path)
+				? LoadFromAssemblyPath(path)
+				: null;
 	}
 
 	private sealed class PassthroughLoader : IAnalyzerAssemblyLoader
--- a/src/Uno.HotReload.Tests/Roslyn/Given_CollectibleAnalyzerAssemblyLoader.cs
+++ b/src/Uno.HotReload.Tests/Roslyn/Given_CollectibleAnalyzerAssemblyLoader.cs
@@ -192,6 +192,38 @@ public sealed class Given_CollectibleAnalyzerAssemblyLoader
 		}
 	}
 
+	[TestMethod]
+	[Description(
+		"When the exact requested version is not registered, resolution must fall back to the " +
+		"highest registered version with a compatible identity — a fresh context is required " +
+		"here: once a context has loaded a simple name, its own by-name cache answers (or " +
+		"rejects) later requests without ever calling Load() again.")]
+	public void When_RequestedVersionNotRegistered_Then_HighestCompatibleWins()
+	{
+		var root = Directory.CreateTempSubdirectory("uno-hr-tests").FullName;
+		try
+		{
+			var v1 = EmitAssembly(Path.Join(root, "v1", "Dep.dll"), "Dep", new Version(1, 0, 0, 0));
+			var v2 = EmitAssembly(Path.Join(root, "v2", "Dep.dll"), "Dep", new Version(2, 0, 0, 0));
+			var payload = CopyTo(Directory.CreateDirectory(Path.Join(root, "analyzer")).FullName, typeof(Given_CollectibleAnalyzerAssemblyLoader).Assembly.Location);
+
+			var loader = new CollectibleAnalyzerAssemblyLoader();
+			loader.AddDependencyLocation(v1);
+			loader.AddDependencyLocation(v2);
+			var analyzerContext = AssemblyLoadContext.GetLoadContext(loader.LoadFromPath(payload))!;
+
+			// 1.5 is not registered: the scan must pick the highest compatible version (2.0) —
+			// which also satisfies the runtime's own version validation on the returned assembly.
+			var resolved = analyzerContext.LoadFromAssemblyName(new AssemblyName("Dep, Version=1.5.0.0"));
+
+			Assert.AreEqual(new Version(2, 0, 0, 0), resolved.GetName().Version, "the highest compatible registered version must win");
+		}
+		finally
+		{
+			Directory.Delete(root, recursive: true);
+		}
+	}
+
 	private static string CopyTo(string directory, string sourcePath, string? fileName = null)
 	{
 		var target = Path.Join(directory, fileName ?? Path.GetFileName(sourcePath));
--- a/src/Uno.HotReload/Roslyn/CollectibleAnalyzerAssemblyLoader.cs
+++ b/src/Uno.HotReload/Roslyn/CollectibleAnalyzerAssemblyLoader.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Buffers;
 using System.Collections.Concurrent;
 using System.Collections.Immutable;
 using System.IO;
@@ -8,6 +9,7 @@ using System.Runtime.Loader;
 using System.Security.Cryptography;
 using System.Text;
 using Microsoft.CodeAnalysis;
+using Uno.HotReload.Tracking;
 using Uno.HotReload.Utils;
 
 namespace Uno.HotReload.Roslyn;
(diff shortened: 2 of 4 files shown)


src/Uno.HotReload/Roslyn/CollectibleAnalyzerAssemblyLoader.cs
```diff
@@ -40,8 +41,8 @@ internal sealed class CollectibleAnalyzerAssemblyLoader : IAnalyzerAssemblyLoade
 {
 	private readonly AssemblyLoadContext _compilerContext;
 	private readonly ConcurrentDictionary<string, DirectoryLoadContext> _contexts = new(PathComparer.Comparer);
-	private readonly ConcurrentDictionary<string, string> _knownPathsByFileName = new(StringComparer.OrdinalIgnoreCase);
-	private static readonly string _shadowCopyRoot = Path.Combine(Path.GetTempPath(), "uno-devserver", "analyzers");
+	private readonly ConcurrentDictionary<string, ImmutableHashSet<string>> _knownPathsByFileName = new(StringComparer.OrdinalIgnoreCase);
+	private static readonly string _shadowCopyRoot = Path.Join(Path.GetTempPath(), "uno-devserver", "analyzers");
```
carldebilly: **MEDIUM (security)** — The shadow-copy root is a process-shared, well-known path with no per-user isolation: the first writer of the tree owns it. On a shared machine (CI agent, multi-user dev box) a hostile local process can pre-create the root permissive or as a symlink, plant assemblies at the deterministic keys (SHA-256 of `path|mtime`, derivable from a readable build directory), or swap a file between `File.Move` and `LoadFromAssemblyPath` (TOCTOU) — arbitrary code in the dev-server host. Consider rooting the cache per user (user name/SID in the path, `UnixFileMode` user-only on Unix) and rejecting a pre-existing root that is a symlink or not owned by the current user.

**LOW (operability)** — The shared root also has no retention story: keyed directories and orphaned `*.staging` files accumulate forever. A best-effort startup sweep (e.g. `*.staging` older than a day, delete failures ignored) would keep long dev sessions bounded.

src/Uno.HotReload/Roslyn/CollectibleAnalyzerAssemblyLoader.cs
```diff
@@ -66,37 +74,122 @@ public Assembly LoadFromPath(string fullPath)
...
+			return null;
+		}
+
+		if (candidates.Count == 1)
+		{
+			// Single registered location: load it by name like the same-directory probe does
+			// (identity inspection only matters when there are alternatives to pick between).
+			var single = candidates.First();
+			return File.Exists(single) ? single : null;
+		}
```
carldebilly: **MEDIUM** — The single-candidate fast path skips the name/PKT identity checks the multi-candidate path enforces, which contradicts the doc comment above (Roslyn's loader validates identity even for one candidate). One stale registration with a mismatched identity gets force-loaded (or throws `FileLoadException`) instead of returning null and letting default-context probing take over. Suggest applying the name/PKT filter to the single candidate too (skipping only the version selection), or documenting the deliberate divergence and covering it with a single-wrong-candidate test.

src/Uno.HotReload/Roslyn/CollectibleAnalyzerAssemblyLoader.cs
```diff
@@ -66,37 +74,122 @@ public Assembly LoadFromPath(string fullPath)
...
+
+		var requestedToken = requested.GetPublicKeyToken();
+		string? bestPath = null;
+		Version? bestVersion = null;
+		foreach (var path in candidates.OrderBy(static p => p, StringComparer.Ordinal))
```
carldebilly, line 106: **LOW** — This ordinal sort is load-bearing (deterministic winner when best versions tie, since `ImmutableHashSet` enumeration order is unspecified) but reads as cosmetic. A one-line comment naming the determinism intent would keep a future edit from dropping it.

src/Uno.HotReload/Roslyn/CollectibleAnalyzerAssemblyLoader.cs
```diff
@@ -66,37 +74,122 @@ public Assembly LoadFromPath(string fullPath)
...
+		{
+			AssemblyName candidate;
+			try
+			{
+				candidate = AssemblyName.GetAssemblyName(path);
+			}
+			catch (Exception e) when (e is IOException or BadImageFormatException)
+			{
+				continue; // Deleted since it was registered, or not a loadable assembly.
+			}
```
carldebilly, line 116: **LOW** — Two points on this candidate probe: (1) the filter misses `UnauthorizedAccessException` (unreadable file) and `ArgumentException` (directory-shaped path) — either escapes the ALC `Load` override and turns one bad file in a registered location into a hard analyzer-resolution failure; adding them keeps the probe best-effort. (2) `AssemblyName.GetAssemblyName` re-reads PE headers from disk on every resolution miss — Roslyn caches identities at registration; a small lock-free path→`AssemblyName` cache keyed on (path, timestamp) would avoid repeated I/O for hot names.

src/Uno.HotReload/Roslyn/CollectibleAnalyzerAssemblyLoader.cs
```diff
@@ -66,37 +74,122 @@ public Assembly LoadFromPath(string fullPath)
...
+
+			if (bestVersion is null || candidateVersion > bestVersion)
+			{
+				bestPath = path;
+				bestVersion = candidateVersion;
+			}
+		}
+
+		return bestPath;
```
carldebilly, line 151: **MEDIUM (diagnosability)** — Best-candidate resolution picks a winner (exact / highest compatible) with no record of the candidates, the winner, or the skip reasons. When the highest-version fallback turns out binary-incompatible, the developer only sees a downstream binding error and cannot reconstruct the decision. Threading the project's `IReporter` here and Verbose-logging winner + skipped candidates would make those failures diagnosable.

src/Uno.HotReload/Roslyn/CollectibleAnalyzerAssemblyLoader.cs
```diff
@@ -66,37 +74,122 @@ public Assembly LoadFromPath(string fullPath)
...
+
 	private static string GetShadowCopyPath(string fullPath)
 	{
 		// One shadow directory per (path, timestamp): reused while the file is unchanged, naturally
 		// re-copied after a rebuild. The original file is never memory-mapped, so it never locks.
 		var timestamp = File.GetLastWriteTimeUtc(fullPath).Ticks;
 		var key = Convert.ToHexString(SHA256.HashData(Encoding.UTF8.GetBytes($"{fullPath}|{timestamp}")))[..16];
-		var directory = Path.Combine(_shadowCopyRoot, key);
-		var shadowPath = Path.Combine(directory, Path.GetFileName(fullPath));
+		var directory = Path.Join(_shadowCopyRoot, key);
```
carldebilly: **LOW** — The shadow key is derived from `path|lastWriteTimeUtc` only, so a timestamp-preserving rebuild or cache restore that changes content silently reuses a stale shadow copy. Rare, but worth a note (or folding size/content into the key) now that the key is a cross-process contract.

src/Uno.HotReload/Roslyn/CollectibleAnalyzerAssemblyLoader.cs
```diff
@@ -66,37 +74,122 @@ public Assembly LoadFromPath(string fullPath)
...
-				// Keep the symbols next to the shadow copy so analyzer stack traces stay resolvable.
-				if (Path.ChangeExtension(fullPath, ".pdb") is { } pdb && File.Exists(pdb))
-				{
-					File.Copy(pdb, Path.ChangeExtension(shadowPath, ".pdb"), overwrite: false);
-				}
-			}
-			catch (IOException) when (File.Exists(shadowPath))
+			// Keep the symbols next to the shadow copy so analyzer stack traces stay resolvable.
+			// Published before the assembly: once the .dll is visible the bundle is complete.
+			if (Path.ChangeExtension(fullPath, ".pdb") is { } pdb && File.Exists(pdb))
```
carldebilly: Nit: `Path.ChangeExtension(fullPath, ".pdb") is { } pdb` — `ChangeExtension` never returns null for a non-null input, so the pattern arm is dead; `var pdb = Path.ChangeExtension(...)` + `File.Exists(pdb)` says the same thing.

src/Uno.HotReload/Roslyn/CollectibleAnalyzerAssemblyLoader.cs
```diff
@@ -66,37 +74,122 @@ public Assembly LoadFromPath(string fullPath)
...
+	/// </summary>
+	private static void Publish(string sourcePath, string targetPath)
+	{
+		var staging = $"{targetPath}.{Guid.NewGuid():N}.staging";
+		File.Copy(sourcePath, staging);
+		try
+		{
+			File.Move(staging, targetPath);
+		}
+		catch (IOException) when (File.Exists(targetPath))
```
carldebilly: **MEDIUM (security)** — After a lost race this trusts whatever sits at `targetPath`: `LoadFromAssemblyPath` then executes it. The equivalence argument in the comment holds for well-behaved writers, but on the shared temp root a local process that pre-plants a DLL at the (derivable) deterministic key wins the "race" by default — the "only ever becomes visible complete" guarantee holds against crashes, not hostile writers. Verifying the winner (length/content hash vs the source) or falling back to a private per-process copy would close it.

src/Uno.HotReload/Roslyn/CollectibleAnalyzerAssemblyLoader.cs
```diff
@@ -66,37 +74,122 @@ public Assembly LoadFromPath(string fullPath)
...
+			try
+			{
+				File.Delete(staging);
+			}
+			catch (IOException)
+			{
+				// Best effort: an orphaned staging file in the keyed directory is harmless.
+			}
+		}
+	}
```
carldebilly, line 284: **MEDIUM (resilience)** — Non-race failures leak the staging file: `File.Copy` sits outside the try, and a `Move` failure that does not match the race filter (disk full, AV lock, `UnauthorizedAccessException`) propagates leaving the partial `.staging` behind in the shared root. A best-effort delete of the staging file before rethrowing keeps the root clean.

**LOW** — The lost-race cleanup catches only `IOException`: `File.Copy` preserves the read-only attribute, so `File.Delete(staging)` can throw `UnauthorizedAccessException` and fail a load whose winning copy is valid. `catch (Exception e) when (e is IOException or UnauthorizedAccessException)` covers it.

src/Uno.HotReload/Roslyn/CollectibleAnalyzerAssemblyLoader.cs
```diff
@@ -122,30 +215,41 @@ public Assembly LoadShadowCopy(string fullPath)
...
+			// added to the application through hot reload).
+			try
 			{
-				return fromCompiler;
+				return _loader._compilerContext.LoadFromAssemblyName(assemblyName);
+			}
+			catch
+			{
+				// Not something the compiler's context can provide (same bare catch as Roslyn's
+				// loader) — resolve it as an analyzer-local dependency below.
```
carldebilly: **MEDIUM (diagnosability)** — The bare catch swallows everything (`OperationCanceledException`, `BadImageFormatException`, …) with no trace before falling through. Roslyn parity justifies the fall-through shape, but silently: a compiler-context `FileLoadException` yields a split-identity analyzer-local duplicate that later dies as a confusing `MissingMethodException` with nothing to diagnose. At minimum trace the fallback (the project's `IReporter` convention); ideally narrow to `FileNotFoundException`/`FileLoadException`/`BadImageFormatException`.

Two related notes: `LoadFromAssemblyName` also falls back to the Default ALC internally, so (a) a Default-resolvable simple name now shadows registered analyzer-local versions (Roslyn-parity behavior, but a change vs the previous probing worth stating here), and (b) the comment below ("the runtime then probes the default context") is inaccurate — that probe already happened during delegation. Also note the delegation can materialize analyzer-only dependencies into the longer-lived compiler context — an intended trade-off, relevant if per-app unload accounting is ever added.

src/Uno.HotReload/Roslyn/CollectibleAnalyzerAssemblyLoader.cs
```diff
@@ -122,30 +215,41 @@ public Assembly LoadShadowCopy(string fullPath)
...
 			}
 
 			// Then the analyzer's own directory, then the registered dependency locations.
-			// Anything else falls back to the default context (framework assemblies).
-			var candidate = Path.Combine(_directory, simpleName + ".dll");
-			if (!File.Exists(candidate) && !_loader._knownPathsByFileName.TryGetValue(simpleName + ".dll", out candidate!))
+			// Anything else returns null: the runtime then probes the default context and
+			// raises this context's Resolving event (framework assemblies land there).
+			var candidate = Path.Join(_directory, simpleName + ".dll");
```
carldebilly, line 343: **LOW (security)** — `simpleName` comes from analyzer reference metadata and is joined unvalidated. `Path.Join` (unlike `Combine`) keeps a rooted name from escaping, but `..\..\evil` still probes — and shadow-loads — outside the analyzer directory. Rejecting names containing directory separators or `..` before the join (as Roslyn's loader does) closes it.

src/Uno.HotReload.Tests/Roslyn/Given_CollectibleAnalyzerAssemblyLoader.cs
```diff
@@ -131,13 +132,105 @@ public void When_WithCollectibleAnalyzerReferences_Then_SwapsLoadersKeepingPaths
...
+
+			var resolved = analyzerContext.LoadFromAssemblyName(new AssemblyName("Dep, Version=2.0.0.0"));
+
+			Assert.AreEqual(new Version(2, 0, 0, 0), resolved.GetName().Version, "the exact requested version must win over the first registered one");
+		}
+		finally
+		{
+			Directory.Delete(root, recursive: true);
+		}
+	}
```
carldebilly, line 194: **MEDIUM (test coverage)** — The new tests pin the two headline behaviors well, but the hardening claims that motivated this PR are otherwise unexercised:
- highest-compatible-version fallback (no exact match — e.g. request `Version=3.0.0.0` with v1/v2 registered, assert v2 wins) and PKT-mismatch rejection: the untested half of `GetBestKnownPath` is exactly the branch that would silently pick a wrong winner;
- atomic publish / kill-mid-copy recovery: deterministic to simulate (pre-plant a partial `.staging` and a pdb-without-dll shadow directory, assert recovery; plus a smoke assert of no `*.staging` leftovers after a load);
- the Roslyn `LoadFromPath` same-path-same-instance expectation holds by construction but is unpinned — a one-line `Assert.AreSame(loader.LoadFromPath(p), loader.LoadFromPath(p))` would keep a future shadow-key change honest.

src/Uno.HotReload.Tests/Roslyn/Given_CollectibleAnalyzerAssemblyLoader.cs
```diff
@@ -131,13 +132,105 @@ public void When_WithCollectibleAnalyzerReferences_Then_SwapsLoadersKeepingPaths
...
+		var emit = compilation.Emit(path);
+		Assert.IsTrue(emit.Success, $"Fixture assembly emit failed: {string.Join(", ", emit.Diagnostics)}");
+		return path;
+	}
+
+	/// <summary>
+	/// A compiler-side context that — like the dev-server's per-application context — can
+	/// RESOLVE assemblies it has not loaded yet (lazy, on demand).
+	/// </summary>
+	private sealed class OnDemandContext(Dictionary<string, string> resolvable)
```
carldebilly: Nit: `OnDemandContext` only ever reads the map — `IReadOnlyDictionary<string, string>` would state that.

carldebilly (approved):
Panel review of the loader hardening. The change is sound and faithfully mirrors Roslyn's own loader shape: the compiler-context delegation (`LoadFromAssemblyName` + bare-catch fall-through) matches `AnalyzerAssemblyLoader.Core.cs`, the staging-to-rename publish is a same-volume atomic move with the destination-exists race handled, a crash between the PDB and DLL publishes self-heals on retry, and both new regression tests genuinely fail on the previous code. No public-surface change; the multi-version registry keeps the lock-free/immutable idiom.

Approving — nothing below blocks. Inline: 8 medium and a handful of low/info suggestions. The two most valuable follow-ups are per-user isolation of the shared shadow-copy root and verifying the winner after a lost publish race; the rest are identity-check uniformity on the single-candidate path, diagnosability of the silent fallbacks, staging-file cleanup on non-race failures, and test coverage for the highest-compatible-version fallback and the atomic-publish guarantees.

src/Uno.HotReload/Roslyn/CollectibleAnalyzerAssemblyLoader.cs
```diff
@@ -66,37 +74,122 @@ public Assembly LoadFromPath(string fullPath)
...
+		var staging = $"{targetPath}.{Guid.NewGuid():N}.staging";
+		File.Copy(sourcePath, staging);
+		try
+		{
+			File.Move(staging, targetPath);
+		}
+		catch (IOException) when (File.Exists(targetPath))
+		{
+			// Lost the publish race with another thread or process — the winner's copy is
+			// equivalent (the key embeds the source path and timestamp).
```
autocarl: **HIGH — migrate or validate cache entries left by the old direct-copy implementation.** This PR keeps the exact same `_shadowCopyRoot` and `(path, timestamp)` key as the already-merged code. If that code was interrupted while copying directly to the final `.dll`, the next run on this implementation reaches `File.Exists(shadowPath)`, skips `Publish` entirely, and returns the truncated final file; `LoadFromAssemblyPath` then keeps throwing until the source timestamp changes. In other words, the permanent cache-poisoning scenario described above is prevented only for newly-created keys, not repaired for users who already have the bad entry.

I reproduced this deterministically on this head: create a valid source assembly, write its first 64 bytes to the exact computed final shadow path (the state left by an interrupted old copy), then call `LoadFromPath`; it throws `BadImageFormatException`. Please version the cache schema/root (for example an `analyzers-v2` namespace) or validate an existing final entry and atomically rebuild it, and add that upgrade regression.

autocarl (changes requested):
Follow-up review on exact head `ceae5b4f517f39cc87bef13da28e8aa4e04f0130`.

The three findings carried over from #23863 are correctly addressed for a clean cache: compiler-context resolution now delegates through `LoadFromAssemblyName`, registered same-name candidates are retained and version-selected, and new shadow copies are published by staging + atomic rename.

Validation on this head:
- `Uno.HotReload` Release builds for net9.0 and net10.0: 0 warnings / 0 errors;
- `Uno.HotReload.Tests` net10.0: 120/120 passed;
- net9.0 test binary: 120/120 passed under the installed net10 roll-forward (the local host does not have the net9 runtime);
- hosted checks: 49 success, 2 neutral, no failing check;
- isolated regression probe: the legacy partial-final-cache scenario is reproducible.

One blocking upgrade-path issue remains inline: this implementation keeps the exact cache namespace/key used by the previously merged direct-copy code and trusts any existing final DLL, so the interrupted-copy artifact this change is intended to recover from remains permanently poisoned after upgrading. Requesting changes until existing entries are versioned or validated/rebuilt.

src/Uno.HotReload.Tests/Roslyn/Given_CollectibleAnalyzerAssemblyLoader.cs
```diff
@@ -131,13 +132,105 @@ public void When_WithCollectibleAnalyzerReferences_Then_SwapsLoadersKeepingPaths
...
+			var payload = CopyTo(Directory.CreateDirectory(Path.Join(root, "analyzer")).FullName, typeof(Given_CollectibleAnalyzerAssemblyLoader).Assembly.Location);
+
+			var loader = new CollectibleAnalyzerAssemblyLoader();
+			loader.AddDependencyLocation(v1); // Registered FIRST: "first file name wins" would pick this one.
+			loader.AddDependencyLocation(v2);
+			var analyzerContext = AssemblyLoadContext.GetLoadContext(loader.LoadFromPath(payload))!;
+
+			var resolved = analyzerContext.LoadFromAssemblyName(new AssemblyName("Dep, Version=2.0.0.0"));
+
+			Assert.AreEqual(new Version(2, 0, 0, 0), resolved.GetName().Version, "the exact requested version must win over the first registered one");
```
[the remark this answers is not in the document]
dr1rrb: Done — as a dedicated test rather than an extension of this one: the suggested same-context request would not exercise the scan at all, because once a context has loaded `Dep`, its by-name cache answers (or rejects) any later request for that simple name without ever calling `Load()` again. `When_RequestedVersionNotRegistered_Then_HighestCompatibleWins` uses a fresh context and requests `1.5.0.0` against registered 1.0/2.0: the exact match misses and the scan must pick the highest compatible version (2.0) — which also keeps the runtime's own version validation on the returned assembly satisfied (returning a version lower than the request would fail the bind anyway).

src/Uno.HotReload/Roslyn/CollectibleAnalyzerAssemblyLoader.cs
```diff
@@ -66,37 +74,122 @@ public Assembly LoadFromPath(string fullPath)
...
 
+	/// <summary>
+	/// Picks the registered dependency location that best satisfies <paramref name="requested"/>,
+	/// mirroring Roslyn's own loader: an exact version match wins, otherwise the highest version
+	/// whose identity is compatible — never just "whichever project registered that file name
+	/// first". Returns <c>null</c> when nothing usable was registered.
+	/// </summary>
+	private string? GetBestKnownPath(AssemblyName requested)
+	{
+		if (!_knownPathsByFileName.TryGetValue(requested.Name + ".dll", out var candidates))
```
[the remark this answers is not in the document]
dr1rrb: Done — `GetBestKnownPath(string simpleName, AssemblyName requested)`: the already-validated `simpleName` is forwarded by `Load` and used for both the dictionary lookup and the identity comparison, so the method no longer depends on a guard living in its caller.

src/Uno.HotReload/Roslyn/CollectibleAnalyzerAssemblyLoader.cs
```diff
@@ -66,37 +74,122 @@ public Assembly LoadFromPath(string fullPath)
...
+
+			if (bestVersion is null || candidateVersion > bestVersion)
+			{
+				bestPath = path;
+				bestVersion = candidateVersion;
+			}
+		}
+
+		return bestPath;
+	}
```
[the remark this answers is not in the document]
dr1rrb: Declining this one: `AssemblyName.GetAssemblyName(path)` has just probed the file, so re-checking existence before returning would not close the window — it only narrows an already-microseconds-wide race (the load can still lose to a delete after any check). Candidates that DID disappear are already skipped during the scan (the `IOException` → `continue` on the identity probe). And if the winning path vanishes between probe and load, the failure surfaces through the analyzer load-failure diagnostics rather than being silently swallowed — the same behavior as Roslyn's own loader, whose `GetBestPath` returns a path that `LoadFromAssemblyPath` can race identically. Happy to revisit if it ever shows up in practice.

src/Uno.HotReload/Roslyn/CollectibleAnalyzerAssemblyLoader.cs
dr1rrb: Done — the shadow root moved wholesale to per-user `LocalApplicationData`, so no key produced by the previous temp-rooted implementation is reachable anymore: poisoned entries are retired with the namespace rather than migrated or validated (the old temp tree becomes inert debris that OS temp cleanup owns). The relocation also closes the shared-`/tmp` first-writer-owns surface raised in the security thread, and the key now folds in the file size. Recovery from an interrupted publish of the NEW cache is pinned by `When_ShadowDirectoryHoldsPartialState_Then_LoadRecovers` (orphaned staging file + pdb-without-dll debris pre-planted at the exact computed key, load must succeed).

src/Uno.HotReload/Roslyn/CollectibleAnalyzerAssemblyLoader.cs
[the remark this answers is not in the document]
dr1rrb: Done — equivalent effect, different cut: instead of `analyzers-v2` under the shared temp root, the cache moved to per-user `LocalApplicationData`. The upgrade is clean in both directions exactly like a version bump (the old code never looks there, the new code never looks at the temp tree), and the relocation additionally closes the shared-root pre-plant/swap concerns from the security review.

dr1rrb: Done on both counts. Security: the root moved out of the shared temp tree entirely — per-user `LocalApplicationData` on every OS (created with user-only mode on Unix), so there is no world-writable parent to pre-create as permissive/symlinked and the deterministic keys are not reachable by another account; within a single user boundary, a writer that could tamper with the cache can already execute code without it. Retention: a best-effort startup sweep deletes `*.staging` files older than a day; the keyed directories themselves are kept — their liveness across concurrent dev-server processes is unknowable, and the per-key reuse is what the cache exists for.

src/Uno.HotReload/Roslyn/CollectibleAnalyzerAssemblyLoader.cs
dr1rrb: Done — the fast path is gone: every candidate, including a lone one, goes through the same name/PKT identity check, and a stale mismatched registration now falls through to the runtime's own resolution instead of being force-loaded. Pinned by `When_RequestedTokenDoesNotMatch_Then_CandidatesAreSkipped` (registered unsigned candidates, request carrying a public key token, resolution must not pick them up).

src/Uno.HotReload/Roslyn/CollectibleAnalyzerAssemblyLoader.cs
dr1rrb: Done — the comment now states the intent: ordinal order is load-bearing because `ImmutableHashSet` enumeration order is unspecified and the winner must stay stable when candidates tie.

src/Uno.HotReload/Roslyn/CollectibleAnalyzerAssemblyLoader.cs
dr1rrb: First point done — the filter now also covers `UnauthorizedAccessException` and `ArgumentException`, keeping the probe best-effort. Declining the (path, timestamp) → `AssemblyName` cache: the probe only runs when the same-directory probe already missed AND the name is registered out-of-directory, and each (context, simple name) pair resolves at most once (the ALC caches the outcome) — repeated header reads require many contexts missing on the same hot name, which is not a shape we observe. Happy to add the cache if profiling ever says otherwise.

src/Uno.HotReload/Roslyn/CollectibleAnalyzerAssemblyLoader.cs
dr1rrb: Done — the loader now takes an optional `IReporter` (threaded from the server through `WithCollectibleAnalyzerReferences(_reporter)`), and Verbose-traces the ambiguous decisions: winner + version + candidate count when several locations are registered (or the none-matched outcome), and every compiler-context fallback with the exception type. Single-candidate and same-directory resolutions stay silent by design.

src/Uno.HotReload/Roslyn/CollectibleAnalyzerAssemblyLoader.cs
dr1rrb: Done — the file size is folded into the key (path | mtime | size), which catches the realistic timestamp-preserving rewrites, and the comment now names the residual explicitly: a same-time same-size content swap still reuses the stale copy — hashing every analyzer at every load would cost what the cache exists to save.

src/Uno.HotReload/Roslyn/CollectibleAnalyzerAssemblyLoader.cs
dr1rrb: Done — `var pdb = Path.ChangeExtension(fullPath, ".pdb")` + `File.Exists(pdb)`.

src/Uno.HotReload/Roslyn/CollectibleAnalyzerAssemblyLoader.cs
dr1rrb: Addressed via the root relocation rather than content verification: with the cache under the user's profile (user-only mode on Unix), whoever can plant a file at the deterministic key already runs as the user and does not need this cache to execute code — within that boundary the equivalence argument holds, and rename-publish guards what it is meant to guard (crashes and partial states). Verifying the winner against the source on every lost race would re-read the source to defend a boundary that no longer exists, so I left it out.

src/Uno.HotReload/Roslyn/CollectibleAnalyzerAssemblyLoader.cs
dr1rrb: Done on both counts — `File.Copy` and `File.Move` now share the try, and every `IOException`/`UnauthorizedAccessException` path best-effort-deletes the staging file (with a `FileAttributes.Normal` reset first, since Copy preserves read-only) BEFORE deciding what it was: target present → lost race, carry on with the winner's copy; target absent → genuine failure, rethrow after the cleanup. The age-gated startup sweep catches whatever still slips through a hard kill.

src/Uno.HotReload/Roslyn/CollectibleAnalyzerAssemblyLoader.cs
dr1rrb: Done — the catch now excludes `OperationCanceledException` and Verbose-traces every fallback with the requested name and exception type; it deliberately stays broad otherwise (like Roslyn's loader): the throw may come from a custom `Resolving` handler on that context, and a handler bug must not take analyzer resolution down with it. Both side notes are now stated in the comment (the delegation probes Default too, so a Default-resolvable name intentionally shadows registered analyzer-local copies; dependencies the compiler's context can provide materialize into that longer-lived context), and the trailing comment is fixed — the runtime re-probes Default redundantly after a null return, then raises this context's `Resolving`.

src/Uno.HotReload/Roslyn/CollectibleAnalyzerAssemblyLoader.cs
dr1rrb: Done — simple names containing any invalid file-name character (separators included) are rejected up front (`SearchValues` over `Path.GetInvalidFileNameChars()`), before the delegation and both probes, so a hostile reference can neither probe nor shadow-load outside the analyzer directory.

src/Uno.HotReload.Tests/Roslyn/Given_CollectibleAnalyzerAssemblyLoader.cs
dr1rrb: Done, with one note. The highest-compatible fallback test was already in the head your review raced with (`When_RequestedVersionNotRegistered_Then_HighestCompatibleWins` — fresh context, request 1.5 against registered 1.0/2.0). New in this round: `When_RequestedTokenDoesNotMatch_Then_CandidatesAreSkipped` (PKT-mismatch rejection) and `When_ShadowDirectoryHoldsPartialState_Then_LoadRecovers` (orphaned staging + pdb-without-dll pre-planted at the exact computed key — the derivation is exposed internal for the test — load must recover), plus the same-path-same-instance assert. I did not add a global zero-`*.staging` assert: the sweep is age-gated (>1 day), so fresh debris from a concurrent run legitimately survives, and the recovery test pins the property that actually matters — debris never blocks a load.

src/Uno.HotReload.Tests/Roslyn/Given_CollectibleAnalyzerAssemblyLoader.cs
dr1rrb: Done — `IReadOnlyDictionary<string, string>`.

src/Uno.HotReload/Roslyn/CollectibleAnalyzerAssemblyLoader.cs
```diff
@@ -66,37 +74,122 @@ public Assembly LoadFromPath(string fullPath)
...
-				// Keep the symbols next to the shadow copy so analyzer stack traces stay resolvable.
-				if (Path.ChangeExtension(fullPath, ".pdb") is { } pdb && File.Exists(pdb))
-				{
-					File.Copy(pdb, Path.ChangeExtension(shadowPath, ".pdb"), overwrite: false);
-				}
-			}
-			catch (IOException) when (File.Exists(shadowPath))
+			// Keep the symbols next to the shadow copy so analyzer stack traces stay resolvable.
+			// Published before the assembly: once the .dll is visible the bundle is complete.
+			if (Path.ChangeExtension(fullPath, ".pdb") is { } pdb && File.Exists(pdb))
```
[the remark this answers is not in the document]
dr1rrb: Done — same fix as the original thread: `var pdb = ...` + `File.Exists(pdb)`.

kazo0:

PR #23707, opened 2026-07-09
fix(alc): sweep collectible-ALC static-cache pins (WASM batch 1)

nickrandolph:
## Summary

Batch 1 of managed sweeps that release **collectible-`AssemblyLoadContext`** pins for a downstream host that loads previewed apps into their own collectible AssemblyLoadContexts and unloads them on reload. On WebAssembly `AssemblyLoadContext.Unloading` is never raised, so process-lifetime static caches that key on (or reference) an app's `Assembly`/`Type`/delegate keep the context alive after `Unload()`. Each cache is swept from the existing ALC cleanup hook (`Application.CleanupNonDefaultAlcCaches`, invoked from `Window.CloseAlcWindow`) or exposes an ALC-scoped removal called from a reachable teardown path — matching the established `Style.ClearCachesForNonDefaultAlc` / `RemoveNonDefaultAlcEntries` pattern.

Closes #23706

## Sweeps landed

1. **`ResourceLoader`** — `ClearNonDefaultAlcAssemblies()` drops previewed-app lookup assemblies (and their parsed-resource markers) from the process-lifetime `_lookupAssemblies` list; wired into the cleanup hook.
2. **`UIElementNativeRegistrar._classNames`** (WASM) — `ClearNonDefaultAlcEntries()` drops the collectible-ALC `Type` keys of the element-type→registration-id map (invisible to a desktop-only guard). The JS-side registration map holds only strings (no ALC pin) and is intentionally left intact — its ids come from a shared counter, so deleting entries there could collide with live registrations.
3. **WindowId maps** — `AppWindow` / `ApplicationView` / `CoreDragDropManager` each gain `DestroyForWindowId`, mirroring `DisplayInformation.DestroyForWindowId`, called from `Window.CloseAlcWindow`. These retained the closed window's instance and its event subscribers.
4. **`CompositionTarget.Rendering`** (WASM) — `ClearNonDefaultAlcHandlers()` drops non-default-ALC subscribers from the static handler list; the per-frame dispatch now reuses a snapshot buffer instead of allocating a fresh `List` every frame.
6. **Hot-reload client status history** — a terminal `HotReloadClientOperation` releases its raw `Type[]` (keeping the curated string list), and the local-operation history is capped to a ~100-entry ring buffer, so retained operations no longer pin every hot-reloaded collectible type.
[...]

PR #23960, opened 2026-08-04
fix(resources): make AddLookupAssembly transactional on parse failure

### LINKED ISSUE #23959: ResourceLoader.AddLookupAssembly rollback can leave partially parsed .upri values observable
## Problem

Follow-up to [the post-merge review comment on PR #23707](https://github.com/unoplatform/uno/pull/23707#issuecomment-5180932885), itself following #23706.

`AddLookupAssembly` now removes a failed assembly registration and its parsed-resource markers when parsing throws. However, `ProcessResourceFile` mutates the live loader incrementally:

1. it records the parsed marker;
2. it obtains the live `_resources` dictionary;
3. it writes each parsed pair with `resources[key] = value`.

If a valid `.upri` declares multiple entries, successfully parses and writes the first entry, then is truncated before a later key/value, `BinaryReader.ReadString` throws `EndOfStreamException` (`IOException`). The `AddLookupAssembly` catch removes the assembly from `_lookupAssemblies` and removes its markers, but does not restore the already-mutated live resource dictionaries.

The result can be an unregistered assembly whose partially parsed resource value remains observable.

The regression added in #23707 uses an invalid magic and therefore throws before the marker or live resource dictionary is mutated; it does not cover this partial-parse case.

## Expected behavior

A failed `AddLookupAssembly` call should be transactional: after it throws, no registration, parsed marker, or resource value contributed by that attempt should remain observable.

## Regression scenario

Emit a `.upri` with:

- a valid header, loader and culture;
- `resourceCount = 2`;
- one complete key/value pair;
- a truncated stream before the second pair.

Then assert:

1. `AddLookupAssembly` throws the expected data/I/O exception;
2. `ContainsLookupAssembly(assembly)` is false;
3. the first value is not visible from the loader.

## Suggested direction

Parse the assembly into temporary dictionaries and apply them only after every `.upri` succeeds, or rebuild live resource state from surviving registrations when registration fails.

Observed by source inspection on merged PR head `1e6b388c77ed51be12b53dc16920b8a8d9ea4249` / merge commit `2fc0049a2568a3e749fe0efb4259da64c9124c91`.

nickrandolph:
**GitHub Issue:** closes #23959

## PR Type:

🐞 Bugfix

## What changed? 🚀

Follow-up to #23707. `AddLookupAssembly` rolled back the `_lookupAssemblies` registration and parsed-resource markers when parsing threw, but `ProcessResourceFile` merges each key/value pair straight into the live loader dictionaries as it reads. A `.upri` with a valid header that declares two pairs, carries one complete pair, then truncates fails on the second pair — after the rollback, the unregistered assembly's first pair remained observable through `GetString` (permanently, when the loader context was already established, since no later lookup reloads).

Every path that mutates the live loader state now goes through a single transactional parse.

- **`TransactionalParse`** (new private nested type) owns the temp dictionaries, the temp marker set, the loader resolver, and the two apply phases (`MergeIntoLiveLoaders`, `ReplaceLiveLoaders`). It replaces the three hand-rolled copies of the temp-then-apply pattern the file had grown.
- **`AddLookupAssembly` — single-assembly branch** (changed loader context): `ProcessAssemblyTransactionally` parses into temporaries with the marker set seeded from the assembly's already-parsed markers, so skip semantics are unchanged, and merges only after every file parsed successfully.
- **`AddLookupAssembly` — full-reload branch**: `ReloadResources` is itself transactional now, so the failure never touches the live state. That also makes the **culture-change path** (`EnsureLoadersCultures` → `ReloadResources`) transactional, which the earlier revision did not cover.
- Consequently the `liveLoadersTouched` control flag, the `RebuildLoaderResourcesFromSurvivors()` recovery call in the catch, and the live-marker rollback (`_parsedResources.RemoveWhere`) are all gone: removing the list entry is the whole rollback. `RebuildLoaderResourcesFromSurvivors` keeps its per-assembly guard and is still used by the ALC sweep. `ClearResources()` and the two-argument `ProcessAssembly` overload became unused and were removed.
- **Rollback logging**: the catch now names the exception and logs before rethrowing (gated on `_log.IsEnabled(LogLevel.Error)`), reporting the assembly whose registration was rolled back and whether the whole cache was being reloaded — in that branch the faulting `.upri` can belong to a pre-existing registration rather than to the assembly being rolled back.
- **Exception contract**: a `.upri` truncated in the middle of its declared pairs surfaced `BinaryReader`'s context-free `EndOfStreamException`. `ProcessResourceFile` now restates it as `InvalidDataException($"Truncated resource file {fileName}: …", error)`, matching what it already does (and documents) for every other malformed-`.upri` case. `InvalidDataException` is already in the rebuild guard's filter, so recovery semantics are unchanged.
- **Allocation**: the merge apply hands the parsed dictionary over on a `TryGetValue` miss instead of allocating an empty dictionary and copying every pair — the zero-copy ownership transfer `RebuildLoaderResourcesFromSurvivors` already used.
- **Documented narrowing**: `RebuildLoaderResourcesFromSurvivors` re-derives values only for the last established context's language preferences, so with `PreserveParsedResources` enabled it drops values preserved for other cultures (they are re-parsed the next time those languages become active). That is now stated in its XML docs.

The `HasContextChangedSignificantly` branch selection in `AddLookupAssembly` is unchanged; the two comments describing those branches were swapped and are now accurate.

### Red/green regression tests

Two new tests in `Given_ResourceLoader_Alc` emit (in memory, via `ManagedPEBuilder`) an assembly embedding a `.upri` with valid magic/version/name/culture, `resourceCount = 2`, one complete pair, truncated before the second pair. Both carry `[Description]` and `[GitHubWorkItem]`:

- `When_AddLookupAssembly_Truncated_Upri_Then_Throws_And_Partial_Value_Not_Observable` — established loader context (full-reload branch): asserts `AddLookupAssembly` throws `InvalidDataException`, `ContainsLookupAssembly` is false, the truncated `.upri`'s loader instance was never created, the first pair's value is **not** returned by `GetString`, and surviving registrations still resolve.
- `When_AddLookupAssembly_Truncated_Upri_With_Changed_Context_Then_Live_Loaders_Untouched` — changed loader context (single-assembly branch): asserts the throw, the rollback, that the loader instance was never created (absence, not emptiness, so a regression to the full-reload branch cannot keep the test green), and that no value reached the live loaders.

Live state is observed through two new `internal` test seams next to `ContainsLookupAssembly` — `TryGetMergedResourceForTests` and `ContainsNamedLoader` — replacing the private-field reflection helper the earlier revision used. `GetString` cannot be used for these assertions: it re-establishes the loader context and would reload (and wipe) the loaders, masking the leak.

Both tests fail on the pre-fix code with the leaked `TruncatedValue` observable in both branches, and on the loader-instance assertions, and pass with the fix. Full `Given_ResourceLoader_Alc` (8 tests) and `Given_AlcCacheSweep` (4 tests) pass; the full `Uno.UI.UnitTests` run is 4074 total / 12 failed, and those 12 `When_Gregorian_FixedDate*` failures reproduce identically on the unmodified merge commit `2fc0049a256` (4072 total / same 12) — environment-dependent, unrelated.

## PR Checklist ✅

### DIFF
--- a/src/Uno.UI.UnitTests/ResourceLoader/Given_ResourceLoader_Alc.cs
+++ b/src/Uno.UI.UnitTests/ResourceLoader/Given_ResourceLoader_Alc.cs
@@ -9,6 +9,7 @@ using System.Reflection.Metadata;
 using System.Reflection.Metadata.Ecma335;
 using System.Reflection.PortableExecutable;
 using System.Runtime.Loader;
+using System.Text;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Windows.Globalization;
 using _ResourceLoader = Windows.ApplicationModel.Resources.ResourceLoader;
@@ -319,6 +320,176 @@ public class Given_ResourceLoader_Alc
 		}
 	}
 
+	[TestMethod]
+	public void When_AddLookupAssembly_Truncated_Upri_Then_Throws_And_Partial_Value_Not_Observable()
+	{
+		// Issue scenario: a .upri with a VALID header that declares two key/value pairs, carries one
+		// complete pair, then truncates before the second. ProcessResourceFile used to merge each
+		// pair directly into the live loader as it read, so BinaryReader.ReadString throwing
+		// EndOfStreamException on the second pair left the FIRST pair's value observable even after
+		// the AddLookupAssembly rollback removed the registration and its parsed markers — an
+		// unregistered assembly's partially parsed resource value survived. A failed
+		// AddLookupAssembly must be fully transactional: no registration, marker, or value remains.
+		const string defaultLanguage = "en";
+		const string uiTestResources = "Uno.UI.UnitTests/Resources";
+		const string truncatedLoaderName = "Uno.UI.Tests.TruncatedUpri/Resources";
+		const string hostValue = "App70-en";
+
+		var previousCulture = CultureInfo.CurrentUICulture;
+		var previousPlo = ApplicationLanguages.PrimaryLanguageOverride;
+		var previousDefault = _ResourceLoader.DefaultLanguage;
+
+		var defaultAlcAssembly = typeof(Given_ResourceLoader_Alc).Assembly;
+		var collectibleAlc = new AssemblyLoadContext("Given_ResourceLoader_Alc.truncated", isCollectible: true);
+		try
+		{
+			// Establish the loader context BEFORE the failing registration: with an unchanged
+			// context no later GetString call reloads (and incidentally wipes) the loaders, so a
+			// leaked partial value would remain observable for the rest of the process lifetime.
+			CultureInfo.CurrentUICulture = new CultureInfo("en-US");
+			ApplicationLanguages.PrimaryLanguageOverride = defaultLanguage;
+			_ResourceLoader.DefaultLanguage = defaultLanguage;
+
+			_ResourceLoader.AddLookupAssembly(defaultAlcAssembly);
+			Assert.AreEqual(
+				hostValue,
+				_ResourceLoader.GetForCurrentView(uiTestResources).GetString("ApplicationName"),
+				"Pre-condition: the host resource resolves before the failing registration.");
+
+			var truncated = LoadAssemblyWithUpri(
+				collectibleAlc,
+				"Uno.UI.Tests.TruncatedUpri",
+				BuildTruncatedUpriPayload(truncatedLoaderName, defaultLanguage));
+
+			Assert.ThrowsExactly<EndOfStreamException>(
+				() => _ResourceLoader.AddLookupAssembly(truncated),
+				"A .upri truncated in the middle of its declared pairs must surface BinaryReader's EndOfStreamException.");
+
+			Assert.IsFalse(
+				_ResourceLoader.ContainsLookupAssembly(truncated),
+				"A failed registration must be rolled back.");
+			Assert.AreEqual(
+				string.Empty,
+				_ResourceLoader.GetForCurrentView(truncatedLoaderName).GetString("TruncatedKey"),
+				"The first pair of a truncated .upri must NOT remain observable after the failed registration — the rollback must cover partially merged values, not just the registration and markers.");
+			Assert.AreEqual(
+				hostValue,
+				_ResourceLoader.GetForCurrentView(uiTestResources).GetString("ApplicationName"),
+				"Surviving registrations must still resolve after the failed registration is rolled back.");
+		}
+		finally
+		{
+			_ResourceLoader.ClearNonDefaultAlcAssemblies();
+			collectibleAlc.Unload();
+			CultureInfo.CurrentUICulture = previousCulture;
+			ApplicationLanguages.PrimaryLanguageOverride = previousPlo;
+			_ResourceLoader.DefaultLanguage = previousDefault;
+		}
+	}
+
+	[TestMethod]
+	public void When_AddLookupAssembly_Truncated_Upri_With_Changed_Context_Then_Live_Loaders_Untouched()
+	{
+		// Same truncated .upri as above, but through AddLookupAssembly's OTHER branch: when the
+		// loader context changed since it was last established, the new assembly is parsed on its
+		// own (no full reload). That parse used to write each pair straight into the live loader
+		// too; it must instead parse into temporaries so a failure leaves the live loaders
+		// completely untouched. Observed at the merged-dictionary seam (no GetString) because a
+		// later GetString reloads for the new context and would mask the leak.
+		const string truncatedLoaderName = "Uno.UI.Tests.TruncatedUpriChangedContext/Resources";
+
+		var previousCulture = CultureInfo.CurrentUICulture;
+		var previousPlo = ApplicationLanguages.PrimaryLanguageOverride;
+		var previousDefault = _ResourceLoader.DefaultLanguage;
+
+		var collectibleAlc = new AssemblyLoadContext("Given_ResourceLoader_Alc.truncatedChangedContext", isCollectible: true);
+		try
+		{
+			// Establish a French loader context...
+			CultureInfo.CurrentUICulture = new CultureInfo("fr-FR");
+			ApplicationLanguages.PrimaryLanguageOverride = "fr";
+			_ResourceLoader.DefaultLanguage = "fr";
+
+			// ...then flip the environment to English WITHOUT the ResourceLoader observing it
+			// (neither setter below re-enters the loader), so the next AddLookupAssembly sees a
+			// significantly changed context and takes the single-assembly parse branch.
+			CultureInfo.CurrentUICulture = new CultureInfo("en-US");
+			ApplicationLanguages.PrimaryLanguageOverride = "en";
+
+			var truncated = LoadAssemblyWithUpri(
+				collectibleAlc,
+				"Uno.UI.Tests.TruncatedUpriChangedContext",
+				BuildTruncatedUpriPayload(truncatedLoaderName, "en"));
+
+			Assert.ThrowsExactly<EndOfStreamException>(
+				() => _ResourceLoader.AddLookupAssembly(truncated),
+				"A .upri truncated in the middle of its declared pairs must surface BinaryReader's EndOfStreamException.");
+
+			Assert.IsFalse(
+				_ResourceLoader.ContainsLookupAssembly(truncated),
+				"A failed registration must be rolled back.");
+			Assert.IsFalse(
+				TryGetMergedResource(truncatedLoaderName, "en", "TruncatedKey", out var leaked),
+				$"The failed single-assembly parse must not leave any value in the live loaders (found '{leaked}').");
+		}
+		finally
+		{
+			_ResourceLoader.ClearNonDefaultAlcAssemblies();
+			collectibleAlc.Unload();
+			CultureInfo.CurrentUICulture = previousCulture;
+			ApplicationLanguages.PrimaryLanguageOverride = previousPlo;
+			_ResourceLoader.DefaultLanguage = previousDefault;
+		}
+	}
+
+	/// <summary>
+	/// Builds a <c>.upri</c> payload with a valid header (magic, version 3, loader name, culture)
+	/// that declares TWO key/value pairs, carries one complete pair, then truncates — matching
+	/// <c>ProcessResourceFile</c>'s binary layout so the parser fails on the second
+	/// <see cref="BinaryReader.ReadString"/> AFTER the first pair parsed successfully.
+	/// </summary>
+	private static byte[] BuildTruncatedUpriPayload(string loaderName, string culture)
+	{
+		using var payload = new MemoryStream();
+		using (var writer = new BinaryWriter(payload, Encoding.UTF8, leaveOpen: true))
+		{
+			writer.Write("uno"u8); // magic
+			writer.Write(3); // version
+			writer.Write(loaderName);
+			writer.Write(culture);
+			writer.Write(2); // resourceCount: declares two pairs...
+			writer.Write("TruncatedKey");
+			writer.Write("TruncatedValue");
+			// ...but the stream ends before the second pair.
+		}
+
+		return payload.ToArray();
+	}
+
+	/// <summary>
+	/// Reads a merged resource value at the private per-loader dictionary seam
+	/// (<c>ResourceLoader._resources[culture][key]</c>) WITHOUT going through
+	/// <c>GetString</c> — which re-establishes the loader context (reloading, and thereby wiping,
+	/// the loaders when the culture changed) and would mask a leaked value.
+	/// </summary>
+	private static bool TryGetMergedResource(string loaderName, string culture, string key, out string? value)
+	{
+		value = null;
+
+		var loadersField = typeof(_ResourceLoader).GetField("_loaders", BindingFlags.Static | BindingFlags.NonPublic)
+			?? throw new InvalidOperationException("ResourceLoader._loaders field was not found.");
+		var loaders = (Dictionary<string, _ResourceLoader>)loadersField.GetValue(null)!;
+		if (!loaders.TryGetValue(loaderName, out var loader))
+		{
+			return false;
+		}
+
+		var resourcesField = typeof(_ResourceLoader).GetField("_resources", BindingFlags.Instance | BindingFlags.NonPublic)
+			?? throw new InvalidOperationException("ResourceLoader._resources field was not found.");
+		var resources = (Dictionary<string, Dictionary<string, string>>)resourcesField.GetValue(loader)!;
+		return resources.TryGetValue(culture, out var map) && map.TryGetValue(key, out value);
+	}
+
 	/// <summary>
 	/// Emits (in memory, via <see cref="ManagedPEBuilder"/>) a minimal assembly whose single
 	/// manifest resource is a malformed <c>.upri</c> (payload does not start with the expected
@@ -326,6 +497,15 @@ public class Given_ResourceLoader_Alc
 	/// exercised with a REAL <see cref="Assembly"/> — no fake seam in the product code.
 	/// </summary>
 	private static Assembly LoadAssemblyWithMalformedUpri(AssemblyLoadContext alc, string assemblyName)
+		=> LoadAssemblyWithUpri(alc, assemblyName, "BAD-upri-payload"u8.ToArray()); // does not start with the "uno" magic
+
+	/// <summary>
+	/// Emits (in memory, via <see cref="ManagedPEBuilder"/>) a minimal assembly whose single
+	/// manifest resource is a <c>.upri</c> with the given <paramref name="upriPayload"/>, and
+	/// loads it into <paramref name="alc"/>. Lets the malformed/truncated-resource paths be
+	/// exercised with a REAL <see cref="Assembly"/> — no fake seam in the product code.
+	/// </summary>
+	private static Assembly LoadAssemblyWithUpri(AssemblyLoadContext alc, string assemblyName, byte[] upriPayload)
 	{
 		var metadata = new MetadataBuilder();
 		metadata.AddAssembly(
@@ -343,14 +523,13 @@ public class Given_ResourceLoader_Alc
 			encBaseId: default);
 
 		// ECMA-335 II.24.2.4: each manifest resource is a 4-byte length prefix followed by the data.
-		var payload = "BAD-upri-payload"u8.ToArray(); // does not start with the "uno" magic
 		var resources = new BlobBuilder();
-		resources.WriteInt32(payload.Length);
-		resources.WriteBytes(payload);
+		resources.WriteInt32(upriPayload.Length);
+		resources.WriteBytes(upriPayload);
 
 		metadata.AddManifestResource(
 			ManifestResourceAttributes.Public,
-			metadata.GetOrAddString("Malformed.upri"),
+			metadata.GetOrAddString("Embedded.upri"),
 			implementation: default,
 			offset: 0);
 
--- a/src/Uno.UI.UnitTests/ResourceLoader/Given_ResourceLoader_Alc.cs
+++ b/src/Uno.UI.UnitTests/ResourceLoader/Given_ResourceLoader_Alc.cs
@@ -457,6 +457,83 @@ public class Given_ResourceLoader_Alc
 		}
 	}
 
+	[TestMethod]
+	[Description(
+		"A .upri truncated BEFORE its declared resource count must surface the same documented " +
+		"InvalidDataException naming the file as one truncated mid pair — the count read must not sit " +
+		"outside the normalizing catch, or one truncation point leaks BinaryReader's context-free " +
+		"EndOfStreamException while every other one is named.")]
+	[GitHubWorkItem("https://github.com/unoplatform/uno/issues/23959")]
+	public void When_AddLookupAssembly_Upri_Truncated_Before_ResourceCount_Then_Throws_InvalidDataException()
+	{
+		const string defaultLanguage = "en";
+		const string truncatedLoaderName = "Uno.UI.Tests.CountTruncatedUpri/Resources";
+
+		var previousCulture = CultureInfo.CurrentUICulture;
+		var previousPlo = ApplicationLanguages.PrimaryLanguageOverride;
+		var previousDefault = _ResourceLoader.DefaultLanguage;
+
+		var collectibleAlc = new AssemblyLoadContext("Given_ResourceLoader_Alc.countTruncated", isCollectible: true);
+		try
+		{
+			CultureInfo.CurrentUICulture = new CultureInfo("en-US");
+			ApplicationLanguages.PrimaryLanguageOverride = defaultLanguage;
+			_ResourceLoader.DefaultLanguage = defaultLanguage;
+
+			var truncated = LoadAssemblyWithUpri(
+				collectibleAlc,
+				"Uno.UI.Tests.CountTruncatedUpri",
+				BuildCountTruncatedUpriPayload(truncatedLoaderName, defaultLanguage));
+
+			var error = Assert.ThrowsExactly<InvalidDataException>(
+				() => _ResourceLoader.AddLookupAssembly(truncated),
+				"A .upri whose stream ends before its resource count is just as malformed as one ending mid pair, so it must surface as InvalidDataException too — not as the raw EndOfStreamException.");
+
+			Assert.IsTrue(
+				error.Message.Contains(truncatedLoaderName.Split('/')[0], StringComparison.Ordinal)
+					|| error.Message.Contains("resource count", StringComparison.Ordinal),
+				$"The message must identify the truncated file rather than being context-free (got '{error.Message}').");
+			Assert.IsInstanceOfType<EndOfStreamException>(
+				error.InnerException,
+				"The original EndOfStreamException must be preserved as the inner exception.");
+
+			Assert.IsFalse(
+				_ResourceLoader.ContainsLookupAssembly(truncated),
+				"A failed registration must be rolled back.");
+			Assert.IsFalse(
+				_ResourceLoader.ContainsNamedLoader(truncatedLoaderName),
+				"The failed parse must not create the truncated .upri's loader.");
+		}
+		finally
+		{
+			_ResourceLoader.ClearNonDefaultAlcAssemblies();
+			collectibleAlc.Unload();
+			CultureInfo.CurrentUICulture = previousCulture;
+			ApplicationLanguages.PrimaryLanguageOverride = previousPlo;
+			_ResourceLoader.DefaultLanguage = previousDefault;
+		}
+	}
+
+	/// <summary>
+	/// Builds a <c>.upri</c> payload with a valid header that truncates BEFORE its resource count,
+	/// so the failure lands on the <see cref="BinaryReader.ReadInt32"/> that reads the count rather
+	/// than on a pair.
+	/// </summary>
+	private static byte[] BuildCountTruncatedUpriPayload(string loaderName, string culture)
+	{
+		using var payload = new MemoryStream();
+		using (var writer = new BinaryWriter(payload, Encoding.UTF8, leaveOpen: true))
+		{
+			writer.Write("uno"u8); // magic
+			writer.Write(3); // version
+			writer.Write(loaderName);
+			writer.Write(culture);
+			// ...and the stream ends before the resource count.
+		}
+
+		return payload.ToArray();
+	}
+
 	/// <summary>
 	/// Builds a <c>.upri</c> payload with a valid header (magic, version 3, loader name, culture)
 	/// that declares TWO key/value pairs, carries one complete pair, then truncates — matching
--- a/src/Uno.UI.UnitTests/ResourceLoader/Given_ResourceLoader_Alc.cs
+++ b/src/Uno.UI.UnitTests/ResourceLoader/Given_ResourceLoader_Alc.cs
@@ -321,15 +321,19 @@ public class Given_ResourceLoader_Alc
 	}
 
 	[TestMethod]
+	[Description(
+		"A .upri truncated mid key/value pair, registered while the loader context is unchanged (the " +
+		"full-reload branch), must leave no registration, no parsed marker and no partially merged " +
+		"value observable — not even the loader instance the failed parse would have created.")]
+	[GitHubWorkItem("https://github.com/unoplatform/uno/issues/23959")]
 	public void When_AddLookupAssembly_Truncated_Upri_Then_Throws_And_Partial_Value_Not_Observable()
 	{
 		// Issue scenario: a .upri with a VALID header that declares two key/value pairs, carries one
 		// complete pair, then truncates before the second. ProcessResourceFile used to merge each
-		// pair directly into the live loader as it read, so BinaryReader.ReadString throwing
-		// EndOfStreamException on the second pair left the FIRST pair's value observable even after
-		// the AddLookupAssembly rollback removed the registration and its parsed markers — an
-		// unregistered assembly's partially parsed resource value survived. A failed
-		// AddLookupAssembly must be fully transactional: no registration, marker, or value remains.
+		// pair directly into the live loader as it read, so failing on the second pair left the FIRST
+		// pair's value observable even after the AddLookupAssembly rollback removed the registration
+		// and its parsed markers — an unregistered assembly's partially parsed resource value
+		// survived. A failed AddLookupAssembly must be fully transactional.
 		const string defaultLanguage = "en";
 		const string uiTestResources = "Uno.UI.UnitTests/Resources";
 		const string truncatedLoaderName = "Uno.UI.Tests.TruncatedUpri/Resources";
@@ -361,13 +365,16 @@ public class Given_ResourceLoader_Alc
 				"Uno.UI.Tests.TruncatedUpri",
 				BuildTruncatedUpriPayload(truncatedLoaderName, defaultLanguage));
 
-			Assert.ThrowsExactly<EndOfStreamException>(
+			Assert.ThrowsExactly<InvalidDataException>(
 				() => _ResourceLoader.AddLookupAssembly(truncated),
-				"A .upri truncated in the middle of its declared pairs must surface BinaryReader's EndOfStreamException.");
+				"A .upri truncated in the middle of its declared pairs must surface as InvalidDataException naming the file — the contract ProcessResourceFile documents for malformed content.");
 
 			Assert.IsFalse(
 				_ResourceLoader.ContainsLookupAssembly(truncated),
 				"A failed registration must be rolled back.");
+			Assert.IsFalse(
+				_ResourceLoader.ContainsNamedLoader(truncatedLoaderName),
+				"The failed parse must not even create the truncated .upri's loader — a created loader proves the parse wrote into the live state. Checked before GetForCurrentView below, which creates it on purpose.");
 			Assert.AreEqual(
 				string.Empty,
 				_ResourceLoader.GetForCurrentView(truncatedLoaderName).GetString("TruncatedKey"),
@@ -388,6 +395,11 @@ public class Given_ResourceLoader_Alc
 	}
 
 	[TestMethod]
+	[Description(
+		"The same truncated .upri through AddLookupAssembly's OTHER branch — a changed loader context, " +
+		"where only the new assembly is parsed — must leave the live loaders completely untouched: no " +
+		"loader instance created and no value merged, pinning the single-assembly transactional parse.")]
+	[GitHubWorkItem("https://github.com/unoplatform/uno/issues/23959")]
 	public void When_AddLookupAssembly_Truncated_Upri_With_Changed_Context_Then_Live_Loaders_Untouched()
 	{
 		// Same truncated .upri as above, but through AddLookupAssembly's OTHER branch: when the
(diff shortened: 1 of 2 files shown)


src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.static.cs
```diff
@@ -74,34 +74,48 @@ public static void AddLookupAssembly(Assembly assembly)
...
+			// re-hit the same parse failure forever. Roll back the just-added entry before
+			// rethrowing to the caller.
 			var lastIndex = _lookupAssemblies.LastIndexOf(assembly);
 			if (lastIndex >= 0)
 			{
 				_lookupAssemblies.RemoveAt(lastIndex);
 			}
 
-			if (!_lookupAssemblies.Contains(assembly))
+			if (liveLoadersTouched)
```
carldebilly, line 111: **HIGH** — The rollback path runs with zero logging: the failure surfaces to the caller as a bare rethrown exception, with no trace that a rollback/rebuild happened nor which assembly's `.upri` actually failed. In the reload branch the faulting file can belong to a *pre-existing* registration while the innocent just-added assembly is the one rolled back — a developer will chase the wrong assembly. Suggested fix: name the exception in the catch and, before `throw`, log with the pattern already used at the ALC-sweep recovery (`if (_log.IsEnabled(LogLevel.Error)) { _log.LogError($"AddLookupAssembly failed for '{assembly.FullName}' (live loaders touched: {liveLoadersTouched}); registration rolled back.", error); }`).

src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.static.cs
```diff
@@ -74,34 +74,48 @@ public static void AddLookupAssembly(Assembly assembly)
...
 			{
-				_parsedResources.RemoveWhere(marker => marker.Assembly == assembly);
+				// The reload merged each assembly's pairs (and parsed markers) straight into the
+				// live state as it read, so the failed assembly may have left partially parsed
+				// values behind. Re-derive the merged values and the markers from the surviving
+				// registrations — the same recovery the ALC sweep uses — so nothing contributed by
+				// the failed attempt remains observable.
+				RebuildLoaderResourcesFromSurvivors();
```
carldebilly, line 118: **MEDIUM** — Two issues with this recovery call. (1) It is unguarded: if `RebuildLoaderResourcesFromSurvivors()` itself throws (a survivor failing with an exception outside its per-assembly filter, or a failure before its apply phase), that exception replaces the original parse failure and the partially merged live state — the exact corruption this PR fixes — stays observable. Wrap it in try/catch, log the rebuild failure, and still rethrow the original exception. (2) It ignores `PreserveParsedResources = true`: the rebuild re-derives values only for the current language preferences and replaces `_parsedResources` wholesale, silently discarding preserved other-culture values that the old `RemoveWhere` rollback kept — the recovery is more destructive than the failure it recovers from. Honor the flag, or explicitly document/test this narrowing.

src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.static.cs
```diff
@@ -74,34 +74,48 @@ public static void AddLookupAssembly(Assembly assembly)
...
 		try
 		{
 			if (HasContextChangedSignificantly(out var context))
 			{
-				// The cache is still valid, we only have to load resources from the given assembly
-				ProcessAssembly(assembly, context.LanguagePreferences);
+				// The cache is still valid, we only have to load resources from the given assembly.
+				// Parse into temporaries and merge only on success, so a malformed .upri (e.g. a
+				// stream truncated in the middle of a key/value pair) can never leave a partially
+				// parsed value or marker in the live state.
```
carldebilly, line 89: **MEDIUM** — Two problems in this comment block. (1) The retained first sentence ("The cache is still valid, we only have to load resources from the given assembly") contradicts the branch it sits on — this is the *context-changed* branch, as the new tests correctly describe. (2) "can never leave a partially parsed value or marker in the live state" overclaims: `EnsureLoadersCultures` → `ReloadResources` is a third live-mutation path with no rollback — an assembly carrying a valid `en` .upri plus a truncated `fr` .upri registers fine under `en` (the `fr` file is skipped at the culture check before values are read), then leaks partial state on the first culture switch to `fr`, and is not rolled back there. Scope the claim to `AddLookupAssembly` and consider a follow-up issue for the culture-change reload path.

src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.static.cs
```diff
@@ -111,6 +125,64 @@ public static void AddLookupAssembly(Assembly assembly)
...
+
+		Dictionary<string, Dictionary<string, string>> ResolveParsedLoader(string name)
+		{
+			if (!parsedResources.TryGetValue(name, out var cultures))
+			{
+				parsedResources[name] = cultures = new Dictionary<string, Dictionary<string, string>>(StringComparer.OrdinalIgnoreCase);
+			}
+
+			return cultures;
+		}
```
carldebilly, line 157: **MEDIUM** — `ResolveParsedLoader` is byte-identical to `ResolveRebuiltLoader` below, and this method is now the third hand-rolled copy of the temp-then-apply pattern in this file, coordinated by the `liveLoadersTouched` control flag whose correctness silently depends on statement ordering inside the try block. One refactor removes all of it: extract a shared temp-parse helper (temp dictionaries + marker set + resolver + apply-into-live) and route `ReloadResources` itself through it — that makes every caller transactional (including the culture-change path), deletes the flag, and deletes the rebuild-on-failure branch.

src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.static.cs
```diff
@@ -111,6 +125,64 @@ public static void AddLookupAssembly(Assembly assembly)
...
+		// Apply phase: parsing succeeded, merge into the live loaders with the same
+		// last-writer-wins semantics ProcessResourceFile uses when writing to them directly.
+		foreach (var parsed in parsedResources)
+		{
+			var loaderResources = GetOrCreateNamedLoaderResources(parsed.Key);
+			foreach (var culture in parsed.Value)
+			{
+				if (!loaderResources.TryGetValue(culture.Key, out var resources))
+				{
+					loaderResources[culture.Key] = resources = new Dictionary<string, string>();
```
carldebilly, line 170: **LOW** — When the live loader has no entry for this culture, the apply phase allocates an empty dictionary and then copies every pair into it, even though the temp dictionary is discarded right after. `RebuildLoaderResourcesFromSurvivors` already does the zero-copy ownership transfer for this case — assign `loaderResources[culture.Key] = culture.Value` on a `TryGetValue` miss and only merge pair-by-pair into a pre-existing dictionary.

src/Uno.UI.UnitTests/ResourceLoader/Given_ResourceLoader_Alc.cs
```diff
@@ -319,13 +320,192 @@ public void When_AddLookupAssembly_Malformed_Then_Throws_And_Registration_Rolled
 		}
 	}
 
+	[TestMethod]
+	public void When_AddLookupAssembly_Truncated_Upri_Then_Throws_And_Partial_Value_Not_Observable()
```
carldebilly, line 324: **MEDIUM** — Both new tests (this one and `When_AddLookupAssembly_Truncated_Upri_With_Changed_Context_Then_Live_Loaders_Untouched`) carry their intent only in `//` comments. Please add MSTest `[Description("…")]` distilling what each test verifies and why (e.g. "A truncated .upri failing mid-pair must leave no registration, marker, or partially merged value observable"), so the intent surfaces in test runners and reports, plus `[GitHubWorkItem("https://github.com/unoplatform/uno/issues/23959")]` for regression traceability.

src/Uno.UI.UnitTests/ResourceLoader/Given_ResourceLoader_Alc.cs
```diff
@@ -319,13 +320,192 @@ public void When_AddLookupAssembly_Malformed_Then_Throws_And_Registration_Rolled
...
+				"Pre-condition: the host resource resolves before the failing registration.");
+
+			var truncated = LoadAssemblyWithUpri(
+				collectibleAlc,
+				"Uno.UI.Tests.TruncatedUpri",
+				BuildTruncatedUpriPayload(truncatedLoaderName, defaultLanguage));
+
+			Assert.ThrowsExactly<EndOfStreamException>(
+				() => _ResourceLoader.AddLookupAssembly(truncated),
+				"A .upri truncated in the middle of its declared pairs must surface BinaryReader's EndOfStreamException.");
```
carldebilly, line 366: **MEDIUM** — This pins `BinaryReader`'s bare `EndOfStreamException` as the failure contract (also in the second test), while every other malformed-`.upri` case surfaces `InvalidDataException` naming the resource file — `ProcessResourceFile` even documents that contract. A context-free "Unable to read beyond the end of the stream" gives a developer no way to tell which assembly or which of its `.upri` files truncated. Suggested fix: wrap the key/value read loop in `ProcessResourceFile` to rethrow as `InvalidDataException($"Truncated resource file {fileName}…", ex)` — `InvalidDataException` is already in the rebuild guard's filter, so recovery semantics are unchanged — and assert that type here instead.

src/Uno.UI.UnitTests/ResourceLoader/Given_ResourceLoader_Alc.cs
```diff
@@ -319,13 +320,192 @@ public void When_AddLookupAssembly_Malformed_Then_Throws_And_Registration_Rolled
...
+		return payload.ToArray();
+	}
+
+	/// <summary>
+	/// Reads a merged resource value at the private per-loader dictionary seam
+	/// (<c>ResourceLoader._resources[culture][key]</c>) WITHOUT going through
+	/// <c>GetString</c> — which re-establishes the loader context (reloading, and thereby wiping,
+	/// the loaders when the culture changed) and would mask a leaked value.
+	/// </summary>
+	private static bool TryGetMergedResource(string loaderName, string culture, string key, out string? value)
```
carldebilly, line 475: **MEDIUM** — This helper reflects over the private `_loaders`/`_resources` fields, so any rename breaks it at runtime rather than compile time. The mask-by-`GetString` rationale is valid, but this file already documents the repo's answer to that exact problem: `ContainsLookupAssembly` was added as an `internal` test seam to avoid reflecting over private state. Suggested fix: add an `internal static bool TryGetMergedResourceForTests(string loaderName, string culture, string key, out string? value)` next to `ContainsLookupAssembly` and use it here.

src/Uno.UI.UnitTests/ResourceLoader/Given_ResourceLoader_Alc.cs
```diff
@@ -319,13 +320,192 @@ public void When_AddLookupAssembly_Malformed_Then_Throws_And_Registration_Rolled
...
+			Assert.ThrowsExactly<EndOfStreamException>(
+				() => _ResourceLoader.AddLookupAssembly(truncated),
+				"A .upri truncated in the middle of its declared pairs must surface BinaryReader's EndOfStreamException.");
+
+			Assert.IsFalse(
+				_ResourceLoader.ContainsLookupAssembly(truncated),
+				"A failed registration must be rolled back.");
+			Assert.IsFalse(
+				TryGetMergedResource(truncatedLoaderName, "en", "TruncatedKey", out var leaked),
+				$"The failed single-assembly parse must not leave any value in the live loaders (found '{leaked}').");
```
carldebilly, line 433: **LOW** — This assertion can't detect loss of the code path it guards: if branch selection regressed and this scenario went through the full-reload branch instead, the rebuild recovery would also make `TryGetMergedResource` return false and the test would stay green while `ProcessAssemblyTransactionally` becomes dead code. Also assert that the truncated loader instance was never *created* in `_loaders` (absent, not merely present-but-empty) to pin the transactional branch specifically.

carldebilly (commented):
Review of 8fbeb652a (head unchanged since analysis). The core fix holds up under inspection: both `AddLookupAssembly` branches are now genuinely transactional, the removed `RemoveWhere` rollback is indeed dead in both branches, the temp-marker seeding preserves skip semantics (markers are per-`(assembly, fileName)`), the temp/live dictionary comparers match, and the two regression tests deterministically pin the two distinct branches (`PrimaryLanguageOverride`'s setter never re-enters the loader). No public API change; a hostile `.upri` cannot pre-size any allocation from its declared `resourceCount`, and the PE-emission helper is test-scoped — nice exotic-code hygiene overall.

Findings below: **1 high** (diagnosability of the rollback path), **6 medium** (recovery robustness + `PreserveParsedResources` semantics, a third copy of the temp-then-apply pattern, a contradictory/overclaiming comment block, test metadata, exception contract, test-seam fidelity), **2 low**. Two notes not anchored inline: several of the new comment blocks run 4–6 lines where repo guidance is one or two — worth a trim; and setting `liveLoadersTouched = true` just before the apply loop in `ProcessAssemblyTransactionally` would extend the recovery to a (pathological) mid-apply failure for free.

Overall: fix-first — the logging gap and the unguarded recovery call are cheap to address and materially change how a production failure gets investigated.

src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.static.cs
nickrandolph: Done in full in da42cfe3e3c.

Extracted `private sealed class TransactionalParse` - temp dictionaries + temp marker set + resolver, with two apply phases (`MergeIntoLiveLoaders` / `ReplaceLiveLoaders`). All three sites now route through it: `ProcessAssemblyTransactionally`, `ReloadResources`, and `RebuildLoaderResourcesFromSurvivors`.

That deletes exactly what you pointed at:

- the `ResolveParsedLoader` / `ResolveRebuiltLoader` duplication,
- the `liveLoadersTouched` control flag whose correctness depended on statement ordering,
- the `RebuildLoaderResourcesFromSurvivors()` recovery call in the `AddLookupAssembly` catch.

`ClearResources()` and the 2-arg `ProcessAssembly` overload fell out as unused and were removed too. As you predicted, routing `ReloadResources` through it makes the `EnsureLoadersCultures` -> `ReloadResources` culture-change path transactional as well - that was the third unguarded live-mutation path you named on the `:89` thread.

Net **-194 / +235** lines across both files, and that already includes two new test seams and the new logging.

src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.static.cs
nickrandolph: Fixed in da42cfe3e3c.

The catch now names the exception and logs before rethrowing, gated on `_log.IsEnabled(LogLevel.Error)`, following the ALC-sweep recovery pattern already in this file:

> `AddLookupAssembly failed for '{assembly.FullName}' (whole cache reloaded: {reloadedEverything}); registration rolled back, live resources unchanged.`

`reloadedEverything` replaces `liveLoadersTouched` as a **diagnostics-only** local - it is read solely by the log message, never by the rollback, and is commented to that effect. I kept it specifically to preserve the signal that motivated your finding: in the reload branch the faulting `.upri` may belong to a *pre-existing* registration while the innocent just-added assembly is the one rolled back, and without that flag in the message a developer still cannot tell the two apart.

One alternative I considered and rejected: hoisting `HasContextChangedSignificantly` out of the `try` to avoid a try-scoped local entirely. That would stop covering a throw from that call, which seemed a worse trade than a documented diagnostics local. If you would still rather not see a try-scoped local at all, I will fold the caveat into a single unconditional sentence instead - happy either way.

The message is accurate post-refactor: with all three paths transactional, live resources genuinely are unchanged in both branches now.

src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.static.cs
nickrandolph: @carldebilly - taking your two points separately.

**(1) The unguarded call is moot: it is deleted.** With `ReloadResources` transactional (see the `:157` thread), a failed parse never touches live state, so there is nothing to recover and no site where a rebuild failure could replace the original parse exception. The only surviving `RebuildLoaderResourcesFromSurvivors` caller is `ClearAlcAssembliesCore` (the ALC sweep), where there is no original exception to preserve and the apply phase is pure dictionary mutation.

**(2) You were right about `PreserveParsedResources`, and I have documented rather than fixed it - tell me if you want it fixed.** Verified by reading the code: `RebuildLoaderResourcesFromSurvivors` derives values only for `_loaderContext.LanguagePreferences`, and `ProcessResourceFile` early-returns for any `.upri` whose culture is not in those preferences. So with `PreserveParsedResources = true`, values accumulated for *other* cultures across earlier language changes are dropped, and `_parsedResources` is replaced wholesale. That is a real narrowing.

What stopped me shipping a fix here: after the refactor the only caller is the ALC sweep, which came from #23707, not this PR. Honouring the flag means re-deriving for the union of previously-parsed cultures - a behaviour change to the sweep with no test coverage today. I took the "explicitly document" option your finding offered and added a `<remarks>` block stating the narrowing and that those values are re-parsed the next time those languages become active.

Say the word and I will file a follow-up issue with the union-of-cultures fix and a test, rather than leaving it as a comment.

src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.static.cs
nickrandolph: Both problems fixed in da42cfe3e3c.

**(1) The contradictory sentence.** You are right that it sits on the wrong branch - and for the record it was mis-attached in the base commit too (verified against `2fc0049a256`), so this PR inherited it rather than introduced it. Both branch comments are now rewritten to describe what the branches actually do.

**(2) The overclaim.** Fair hit, and it is now literally true rather than papered over: the `EnsureLoadersCultures` -> `ReloadResources` third path you identified is transactional as of the `:157` refactor, so every live-mutation path - single-assembly, full reload, culture-change - goes through `TransactionalParse`. I moved the claim onto `TransactionalParse` and stated it precisely there instead of sweepingly on the branch. No follow-up issue needed.

src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.static.cs
nickrandolph: Done in da42cfe3e3c. `MergeIntoLiveLoaders` now assigns `liveResources[culture.Key] = culture.Value` on a `TryGetValue` miss and only merges pair-by-pair into a pre-existing dictionary - matching `RebuildLoaderResourcesFromSurvivors`, as you noted.

Checked the prerequisite before doing it: the temp and live per-culture dictionaries are both created by `ProcessResourceFile` with the default comparer, so the ownership transfer cannot smuggle in a mismatched comparer.

src/Uno.UI.UnitTests/ResourceLoader/Given_ResourceLoader_Alc.cs
nickrandolph: Done in da42cfe3e3c - both tests now carry a multi-line `[Description(...)]` distilling what each verifies and why, plus `[GitHubWorkItem("https://github.com/unoplatform/uno/issues/23959")]`.

Checked the repo's own usage first rather than guessing: `GitHubWorkItemAttribute` comes from `MSTest.TestFramework` 4.0.1 in `Microsoft.VisualStudio.TestTools.UnitTesting` (~30 uses in `Uno.UI.RuntimeTests`), already in scope via this file's existing using.

src/Uno.UI.UnitTests/ResourceLoader/Given_ResourceLoader_Alc.cs
nickrandolph: Done in da42cfe3e3c, and the impact check came back clean.

`ProcessResourceFile` now wraps the key/value read loop and rethrows as:

> `InvalidDataException($"Truncated resource file {fileName}: it declares {resourceCount} resource(s) but the stream ended early.", error)`

Both tests assert `InvalidDataException`. Before changing the contract I checked what depended on it: `ProcessResourceFile` has exactly **one** caller (the 4-arg `ProcessAssembly`), and `EndOfStreamException` appears nowhere else in `src/` except these two tests - so nothing else was pinned to the old type.

I also re-ran the red/green proof against `2fc0049a256` afterwards to be sure the swap did not weaken the tests: both still fail on the unfixed code on the value-leak assertion (`Expected: "" But was: "TruncatedValue"`). PR body updated, since it described the old type.

src/Uno.UI.UnitTests/ResourceLoader/Given_ResourceLoader_Alc.cs
nickrandolph: Done in da42cfe3e3c. Added `internal static bool TryGetMergedResourceForTests(string loaderName, string culture, string key, out string? value)` immediately after `ContainsLookupAssembly`, following its shape and placement - you are right that this file already documented the repo's answer to this exact problem.

The reflection helper is **removed entirely**, not left dead alongside it.

src/Uno.UI.UnitTests/ResourceLoader/Given_ResourceLoader_Alc.cs
nickrandolph: Good catch - this was the one finding that could have let the PR rot silently. Fixed in da42cfe3e3c.

Added one more reflection-free seam, `internal static bool ContainsNamedLoader(string loaderName)`, and the test now asserts the truncated loader was never *created* - absence, not emptiness - so a regression to the full-reload branch fails the test instead of passing on an empty loader. The assertion message spells that reasoning out. I added the same assertion to the first test as well (before its deliberate `GetForCurrentView` call), which strengthens it identically.

It demonstrably pins the branch: on the pre-fix code the tests now fail on `ContainsNamedLoader` *first*, before ever reaching the value-leak assertion.

src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.static.cs
```diff
@@ -111,6 +125,64 @@ public static void AddLookupAssembly(Assembly assembly)
...
+		// Seed the temporary marker set with the assembly's already-parsed files so they are
+		// skipped exactly as a parse against the live markers would skip them.
+		var parsedMarkers = new HashSet<(Assembly Assembly, string ResourceName)>();
+		foreach (var marker in _parsedResources)
+		{
+			if (marker.Assembly == assembly)
+			{
+				parsedMarkers.Add(marker);
+			}
+		}
```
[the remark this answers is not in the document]
nickrandolph: Addressed in da42cfe3e3c, though not the way the suggestion phrased it - the `foreach` and its inner `if` are gone entirely rather than converted. The marker seed is now `new TransactionalParse(_parsedResources.Where(marker => marker.Assembly == assembly))` as part of the shared-helper refactor. This is a cold path (assembly registration), so LINQ is appropriate here.

nickrandolph:
@carldebilly — thanks for the review, this was a genuinely useful one. All 10 threads addressed and resolved in `da42cfe3e3c`; each thread has its own reply with specifics.

**Your `:157` refactor request was the right call and I took it in full.** Extracting `TransactionalParse` (temp dictionaries + temp marker set + resolver + apply) and routing `ReloadResources` itself through it collapsed four of your findings at once: it deleted the `ResolveParsedLoader`/`ResolveRebuiltLoader` duplication, deleted the `liveLoadersTouched` flag, deleted the rebuild-recovery call in the catch (so the unguarded-call half of `:118` is moot), and made the `EnsureLoadersCultures` → `ReloadResources` culture-change path transactional — which is what turned the `:89` "can never leave a partially parsed value" claim from an overclaim into something literally true. Net **−194 / +235** lines including two new test seams and the new logging.

**Verification** (base = `2fc0049a256`):

- `Uno.UWP` Debug + Release (warnings-as-errors): 0 warnings, 0 errors.
- `Given_ResourceLoader_Alc` + `Given_AlcCacheSweep`: **12/12 pass**.
- Full `Uno.UI.UnitTests`: 4074 total / 12 failed on the branch vs 4072 / 12 on the base — identical failing set, all `When_Gregorian_FixedDate*` (pre-existing, environment-dependent). No new failures.
- **Red/green re-verified after the exception-contract change**, since that was the one fix that could have weakened the tests: on the pre-fix code both regression tests still fail (`Expected: "" But was: "TruncatedValue"`), and with the new branch-pinning assertion in place they fail on `ContainsNamedLoader` first.

**Two things are your call, and I deliberately did not decide them unilaterally:**

1. **`PreserveParsedResources` (your `:118` point 2) — you were right, and I documented rather than fixed.** Confirmed by reading the code: `RebuildLoaderResourcesFromSurvivors` derives values only for `_loaderContext.LanguagePreferences` and `ProcessResourceFile` early-returns for cultures outside them, so values accumulated for other cultures across earlier language changes are dropped and `_parsedResources` is replaced wholesale. I stopped short of fixing it because after the refactor its only caller is the ALC sweep from #23707 rather than anything in this PR, and honouring the flag means re-deriving for the union of previously-parsed cultures — a behaviour change to the sweep with no coverage today. Added a `<remarks>` stating the narrowing. **Want a follow-up issue with the union-of-cultures fix and a test, or is the documented caveat enough?**

2. **A diagnostics-only try-scoped local survives.** `reloadedEverything` superficially resembles the `liveLoadersTouched` pattern you objected to, but it is read solely by the new error log and never by the rollback (and is commented as such). I kept it because dropping it loses the "the faulting `.upri` may belong to a pre-existing registration while the innocent just-added assembly is rolled back" signal that motivated your HIGH finding. The alternative — hoisting `HasContextChangedSignificantly` out of the `try` — would stop covering a throw from that call. **Say the word and I'll fold the caveat into one unconditional sentence instead.**

One process note in the interest of honesty about provenance: the AST index I normally use for impact analysis isn't configured for this repo in my environment, so the "single caller of `ProcessResourceFile`", "`EndOfStreamException` appears nowhere else in `src/`", and `GitHubWorkItemAttribute` provenance checks were done by grepping `src/` and reading the call sites rather than from an index. They're single-caller/single-file cases that grep resolves unambiguously, but flagging it so you can weigh the claims accordingly.

nickrandolph:
Follow-up: acting on the **suppressed** Copilot comment from the review of `da42cfe3e3c`, because it was correct and it undercut the very contract I'd just claimed on the `:366` thread.

`resourceCount` was read *outside* the `try` that restates `EndOfStreamException` as `InvalidDataException`, so a `.upri` truncated at its count still surfaced BinaryReader's context-free exception while every other truncation point was named. Fixed in `99a89d5aec2`: the count read moved inside the `try`, the declared count stays in the message when it was read, and there's an explicit message for the case where truncation preceded it.

Red/green, run both ways:

```
# pre-fix source, new test only
failed When_AddLookupAssembly_Upri_Truncated_Before_ResourceCount_Then_Throws_InvalidDataException
  Assert.ThrowsExactly failed. Expected exception type:<System.IO.InvalidDataException>.
  Actual exception type:<System.IO.EndOfStreamException>.

# with the fix
Given_ResourceLoader_Alc + Given_AlcCacheSweep: total: 13  failed: 0  succeeded: 13
```

`Uno.Skia` Release (warnings-as-errors): 0 warnings, 0 errors.

Two notes on scope, since I checked before deciding how far to take this:

- **It was a diagnostics-quality issue, not a functional one.** `EndOfStreamException` derives from `IOException`, and the per-assembly guard in `RebuildLoaderResourcesFromSurvivors` catches `IOException`, so the skip-the-offending-assembly behaviour was never at risk either way. Worth stating plainly rather than implying the fix was more load-bearing than it is.
- **I deliberately did not widen the `try` over the header reads** (`version` / `name` / `culture`). Those are also `EndOfStreamException` sources, but covering them means wrapping `resolveLoaderResources` — a caller-supplied delegate — and relabeling exceptions thrown by code that isn't reading this file. Same guard already catches them via `IOException`. Happy to normalize those too if you'd rather the method's message contract be exhaustive; it'd want the reads hoisted above the delegate call first.

src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.static.cs
```diff
@@ -237,75 +373,30 @@ private static void RebuildLoaderResourcesFromSurvivors()
...
-		// Parse only into the temporaries here — the phase that can throw (malformed .upri) must not
-		// touch the live loaders. A per-assembly guard keeps one bad file from aborting the rebuild.
+		// Started from an empty marker set: the rebuilt state replaces the live one wholesale.
+		var parse = new TransactionalParse();
 		foreach (var assembly in _lookupAssemblies)
 		{
 			try
 			{
-				ProcessAssembly(assembly, languagePreferences, ResolveRebuiltLoader, rebuiltMarkers);
+				parse.Parse(assembly, languagePreferences);
```
autocarl: **HIGH (blocking) — the per-assembly guard does not roll back that assembly’s temporary contributions.**

`parse` is created once before the survivor loop. `ProcessResourceFile` can therefore add a marker and a first key/value pair to that shared temporary state, then throw while reading a later pair. The catch continues, but `ReplaceLiveLoaders()` subsequently publishes the rejected survivor’s partial state.

I reproduced this on exact head `99a89d5aec225f128414de33b339420cb2033cd2` with a disposable test: a valid `.upri` declared two pairs, contained `TruncatedKey = TruncatedValue`, and truncated before pair two. After `ClearAlcAssemblies`, `TryGetMergedResourceForTests` returned `true` with `TruncatedValue`. The existing bad-magic survivor test still passed because it throws before anything enters the temporary maps.

Please parse each survivor into its own scratch transaction and merge that scratch into the aggregate only after success (or add an equivalent checkpoint/rollback for both resources and markers), then add this truncated-after-first-pair rebuild regression.

autocarl (commented):
## Follow-up review — fix-first

The refactor makes `AddLookupAssembly` and `ReloadResources` transactional, but the survivor-rebuild path can still commit partial state from an assembly that its catch block says it skipped.

Validated locally on the exact head `99a89d5aec225f128414de33b339420cb2033cd2`:

- the existing bad-magic survivor test passes (**1/1**), because it fails before mutating the temporary parse state;
- a disposable regression using a valid header, `resourceCount = 2`, one complete pair, then truncation before the second pair fails (**1/1**);
- after `ClearAlcAssemblies`, `TryGetMergedResourceForTests` still finds `TruncatedKey = TruncatedValue`.

**Technical verdict: HIGH / fix-first.** I am submitting a `COMMENT` review rather than `CHANGES_REQUESTED` because this account has read-only repository permission; the maintainers retain the formal Go/No-Go decision.

src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.static.cs
nickrandolph: Confirmed and fixed in `3c553094`. Your reproduction is exact — I reproduced it myself before changing anything: with only your test added to `99a89d5aec2`, `TryGetMergedResourceForTests` returns `true` with `TruncatedValue` (1 failed / 10 total in `Given_ResourceLoader_Alc`), and the bad-magic survivor test passes right alongside it, for the reason you gave — bad magic throws before anything enters the temporaries.

**Each survivor now parses into its own scratch, folded in only after `Parse` returned.**

```csharp
var aggregate = new TransactionalParse();
foreach (var assembly in _lookupAssemblies)
{
    var scratch = aggregate.CreateScratch();
    try
    {
        scratch.Parse(assembly, languagePreferences);
    }
    catch (Exception error) when (/* filter unchanged */)
    {
        // discard this survivor's scratch whole
        ...
        continue;
    }

    aggregate.AbsorbScratch(scratch);
}

aggregate.ReplaceLiveLoaders();
```

`CreateScratch()` returns `new(_markers)` — the constructor copies into a new `HashSet`, so no mutable state is shared at creation, and seeding from the aggregate's markers keeps dedup semantics identical (a `.upri` an earlier survivor already parsed is still skipped). `AbsorbScratch` folds values **and** markers, per your point that a leaked marker is as wrong as a leaked value.

To keep one merge implementation, the existing temp→live loop moved into `private void MergeInto(Func<string, Dictionary<string, Dictionary<string, string>>> resolveTargetResources)`. `MergeIntoLiveLoaders` calls `MergeInto(GetOrCreateNamedLoaderResources)`; `AbsorbScratch` calls `scratch.MergeInto(ResolveLoaderResources)`, where the method group binds to the aggregate. Same last-writer-wins-per-key semantics and the same hand-over-instead-of-copy path, so survivor ordering is unchanged. `ReloadResources` is untouched and still has no per-assembly catch — a throw there propagates and the live state is never reached, which is correct as it stands.

**The multi-`.upri`-per-assembly case**, since that is where "skip this survivor" has to mean something precise. Assembly A embeds `1.upri` (valid), `2.upri` (declares 2 pairs, carries 1, truncates), `3.upri` (valid): `1.upri`'s marker and pairs land in A's scratch, `2.upri` adds its marker, writes `TruncatedKey`, throws out of `ProcessAssembly` and out of `Parse`, `3.upri` is never reached, the catch `continue`s and the scratch dies. The aggregate gets nothing — not even `1.upri`'s good values, which is the correct reading of a per-assembly guard. `AbsorbScratch` is the only path into the aggregate and is unreachable from the catch, and a later scratch's `_loaderResources` is always freshly allocated, so it cannot write into anything the aggregate now owns.

One deliberate behaviour change worth your attention: previously a rejected survivor's partial markers survived in `_parsedResources`, which — only with `PreserveParsedResources` enabled — could suppress a later `ReloadResources` throw by skipping the failing file. That is the stale-marker hazard your comment names, and it is now gone.

**Test:** `When_ClearAlcAssemblies_With_Truncated_Survivor_Then_Partial_Value_Not_Published`, next to the bad-magic survivor test and reusing its helpers (`BuildTruncatedUpriPayload`, `LoadAssemblyWithUpri`, `InjectLookupAssembly`, `OverrideMergedResource`). It asserts at the merged-dictionary seam before any `GetString`, checks the skipped survivor does not even materialize its loader, and carries the counterpart assertion that a good survivor's value **is** republished — so a fix that discarded too much fails it too.

**Verification, all re-run by me rather than taken on trust:**

| Run | Result |
|---|---|
| New test on unmodified `99a89d5aec2` (test only) | **RED** — `found 'TruncatedValue'`, 1 failed / 10 |
| Same test with `3c553094` | **GREEN** — 0 failed / 10 |
| `Given_ResourceLoader*` | 17/17 |
| `src/Uno.UI.UnitTests` full | 4076 total, 12 failed — all 12 pre-existing |
| `dotnet build -c Release` | 0 warnings, 0 errors |

The 12 are `Windows_Globalization.When_Calendar.When_Gregorian_FixedDate*` (local ICU/timezone-sensitive date assertions); I established they are pre-existing by running the full project on the PR head with these two files stashed and diffing the failing-test lists — identical.

Noted that you filed this as `COMMENT` rather than `CHANGES_REQUESTED` because the account is read-only. @carldebilly — flagging that explicitly, since it means this PR's `reviewDecision` never showed a blocker even though this one was HIGH/fix-first.

nickrandolph:
CI is green on `3c553094` — ADO build 231549 finished `succeeded` with zero non-succeeded jobs, so the fix is now CI-verified rather than only locally verified.

One check went red mid-run and is worth explaining, in case anyone saw it: `Tests - Desktop Skia Linux Framebuffer Runtime Tests` reported CANCELLED, from a job timeout rather than a test failure —

> The job running on agent Azure Pipelines 39 ran longer than the maximum time of 60 minutes.

The log shows the last test started was `When_MsAppData()` at 14:23:49, then ~52 minutes of silence to the cutoff. It was not a fail-fast cancellation: nothing else in the build failed. The stage passed on **attempt 2**, where the same job ran 15:35:46 → 15:56:48 — 21 minutes, comfortably inside the limit. `Runtime Tests 1` in another leg also needed a second attempt.

For the record, that leg cannot reach this change: `When_MsAppData` is a `BitmapSource`/`ms-appdata:` image test, and `ClearAlcAssemblies` — the only entry point into `RebuildLoaderResourcesFromSurvivors` — is called from just two places, `Application.Alc.cs` (which needs a dying collectible ALC) and the `Uno.UI.UnitTests` ALC tests. Nothing under `Uno.UI.RuntimeTests` calls it, or `AddLookupAssembly` at all. The remainder of the diff is `MergeInto`, a pure extraction of the loop `MergeIntoLiveLoaders` already ran.

@carldebilly — green and ready whenever you have a moment.

nickrandolph:
**Retargeted to `servicing/6.8`** and rebased onto it — `09ad8076`, 4 commits, 2 files. Same branch and same PR, so the review history above (the HIGH survivor-rollback thread and its fix, and the approval-recommended pass) carries over rather than being re-opened somewhere else.

**This also explains the conflict, and corrects something I said earlier.** The PR reported `CONFLICTING`/`dirty` against `master` while `git merge-tree` and a 4-commit rebase both applied cleanly locally, and I leaned toward GitHub's cache being stale. Retargeting settles it: against `servicing/6.8` GitHub now reports `mergeable=true`. GitHub was right that the *merge into master* was dirty — a clean rebase and a clean merge are different operations and I conflated them. `master` has moved to **.NET 11**, which is also why I could not build a master-rebased branch locally at all (`NETSDK1045` against SDK 10.0.204).

`servicing/6.8` still targets net9/net10, so the verification I could not do on master is now done:

| Check | Result |
|---|---|
| Rebase of all 4 commits onto `servicing/6.8` | clean, no conflicts |
| `dotnet build src/Uno.UI.UnitTests -c Debug` | **0 warnings, 0 errors** |
| `Given_ResourceLoader*` | **45 / 45 pass** |
| Full `src/Uno.UI.UnitTests` | 4148 total, 12 failed |
| Control: same 12 on clean `servicing/6.8` with none of these commits | **12 failed — identical set** |

The 12 are `Windows_Globalization.When_Calendar.When_Gregorian_FixedDate*` — ICU/timezone-sensitive date assertions, pre-existing on the target branch and unrelated to resources.

**And the red proof now runs on the branch that will actually ship it.** Reverting only `ResourceLoader.static.cs` to the `servicing/6.8` baseline, keeping the tests, fails four:

```
When_ClearAlcAssemblies_With_Truncated_Survivor_Then_Partial_Value_Not_Published
When_AddLookupAssembly_Truncated_Upri_Then_Throws_And_Partial_Value_Not_Observable
When_AddLookupAssembly_Truncated_Upri_With_Changed_Context_Then_Live_Loaders_Untouched
When_AddLookupAssembly_Upri_Truncated_Before_ResourceCount_Then_Throws_InvalidDataException
```

so every one of the four commits is load-bearing here, not just on the branch it was written against. Restoring the file returns 10/10.

CI is running on the new base; nothing has failed yet (the rollup briefly read FAILURE while every context was still queued or in progress). Reviewers unchanged: @carldebilly, autocarl, copilot.

PR #23975, opened 2026-08-05
refactor(sdk): Remove vacuous skiarenderer feature gates

MartinZikmund:
**GitHub Issue:** Part of epic [unoplatform/uno-private#2049](https://github.com/unoplatform/uno-private/issues/2049) — _🦋 Drop native targets_. Closes unoplatform/uno-private#2062.

## PR Type:

🔄 Refactoring (no functional changes, no api changes) · 📚 Documentation content changes

## What changed? 🚀

`SkiaRenderer` became a no-op when Skia became the only renderer, but the renderer-selection branches were only *neutralized*, not removed — the Definition of Done asked for both. This removes the 20 now-vacuous `$(UnoFeatures.Contains(';skiarenderer;'))` gates.

### Why they are vacuous

`src/Uno.Sdk/targets/Uno.Features.targets` normalises `$(UnoFeatures)` to lowercase with forced leading/trailing `;` (line 5 turns an empty value into `;`), strips `;nativerenderer;`, then **unconditionally appends `skiarenderer;` when absent** (line 13). So the check is always true.

Import order is safe. `Uno.Common.targets` imports `Uno.Features.targets` at `:70`, then the per-platform commons at `:76-87`, then `Uno.Implicit.Packages.targets` at `:93` — every gate file is downstream. 19 of the 20 gates are `<ItemGroup Condition=…>`; MSBuild evaluates properties in pass 1 and items in pass 4, so item conditions always see the final property value regardless. The one `<PropertyGroup>` gate (`Uno.Common.Wasm.targets:20`) is order-sensitive but is imported 14 lines *after* the normalisation.

### Changes

- **Android / iOS / MacCatalyst** `.targets` (3 gates each) — the `Uno.WinUI.Runtime.Skia.*` and MAUI-embedding groups become `_IsExecutable`-only.
- **`Wasm.targets`** (3) — the browser-head groups become `_IsExecutable`-only; the theming group drops the trailing `AND skiarenderer` and the now-redundant parens. `SkiaSharp.Views.Uno.WinUI` on `:33` is left as-is.
- **`ProjectSystem.targets:97`** (1) — `skia OR skiarenderer OR IsDesktop` is a disjunction with an always-true term, so `Uno.WinUI.Graphics2DSK` becomes unconditional. This also covers WinAppSDK heads, which reach this file via `:114` — not a new dependency there, the `skiarenderer` disjunct already made it true.
- **`ProjectSystem.Uno.targets`** (6) — two always-true disjunctions collapse to unconditional; `svg` and `spellchecking` keep only their real conditions; `SkiaSharp.Skottie` loses its item-level condition. The stale `<!-- Exclude the package on WASM with native renderer (not needed) -->` comment goes with it.
- **`Uno.Common.Wasm.targets:20`** (1) — `WasmShellEnableIDBFS` keeps its `== ''` opt-out clause.

`UnoFeature.SkiaRenderer` **stays** — `Uno.Features.targets:13` injects the token into every build, and `ImplicitPackagesResolver` warns on unknown feature tokens, so removing the enum member would warn on every build. Its `Sdk.props.buildschema.json` description and the docs now say it is an implied no-op instead of a live selector.

## Validation

- **Code review:** `grep skiarenderer src/Uno.Sdk/` now returns only `Uno.Features.targets` (the normalisation and its comment). All 12 edited `.targets` files parse as XML; the buildschema JSON parses.
- **Compile:** no C# involved. The `SolutionTemplate` heads consume the *published* `Uno.Sdk`, so a local restore would prove nothing without packing the SDK.
- **Runtime (behavioural equivalence — this is the real evidence):** built an MSBuild harness that replays `Uno.Features.targets`' normalisation and then imports each edited targets file, and diffed the resulting `_UnoProjectSystemPackageReference` set (identity + `PrivateAssets`/`IncludeAssets`) and `WasmShellEnableIDBFS` between the pre-change and post-change file contents.

  Matrix: 6 targets files × 13 `UnoFeatures` values (empty, `skiarenderer`, `nativerenderer`, `skia`, `lottie`, `svg`, `material`, `cupertino`, `simpletheme`, `spellchecking`, `mauiembedding`, an all-features combo, and a mixed-case `SkiaRenderer;Lottie`) × 5 project shapes (exe/library × desktop/browser-wasm/other, packable and not).

  **390 cases, 0 differences, 0 errors** (325 of them produced a non-empty package set, so the comparison is not vacuously passing).

## PR Checklist ✅

- [x] 🧪 Added [Runtime tests, UI tests, or a manual test sample](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) (for bug fixes / features, if applicable) — covered by the equivalence matrix above

## Not in this PR

- The `.github/ISSUE_TEMPLATE/*.yml` renderer dropdowns still offer a renderer choice. That is a product decision about the bug-report form, not a no-op cleanup.
- `Uno.Implicit.Packages.ProjectSystem.MacCatalyst.targets:18` uses `IncludeAssets="None"` where iOS uses `"build"`. Looks like a real divergence, but it is pre-existing and out of scope here.


doc/articles/features/using-the-uno-sdk.md
```diff
@@ -79,7 +79,7 @@ Here are the supported features:
...
 | `SimpleTheme`        | Adds support for the [Simple Design Theme](xref:Uno.Themes.Simple.GetStarted) library. If the `Toolkit` feature is also used, it will add support for the Simple Design Toolkit library.                                                    |
 | `Skia`               | Adds support for [SkiaSharp](https://github.com/mono/SkiaSharp).                                                                                                                                                                           |
-| `SkiaRenderer`       | Adds support for using Skia as the graphics rendering engine. For more details, see [Skia Rendering documentation](xref:uno.features.renderer.skia).                                                                                               |
+| `SkiaRenderer`       | No-op, kept for backwards compatibility. Skia is always used as the graphics rendering engine. For more details, see [Skia Rendering documentation](xref:uno.features.renderer.skia).                                                       |
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in `f44b3a31cb` — qualified the wording: *"Skia is always used as the graphics rendering engine **on Uno Platform targets**. The WinAppSDK target is not provided by Uno Platform and always renders natively with WinUI."* Good catch.

src/Uno.Sdk/Sdk/Sdk.props.buildschema.json
```diff
@@ -556,7 +556,7 @@
 					"helpUrl": "https://aka.platform.uno/feature-serialization"
 				},
 				"SkiaRenderer": {
-					"description": "Adds support for using Skia as a UI renderer on all targets.",
+					"description": "No-op, kept for backwards compatibility. Skia is always used as the UI renderer on all targets.",
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in `f44b3a31cb` — the schema description now reads *"Skia is always used as the UI renderer **on Uno Platform targets**. The WinAppSDK target always renders natively with WinUI."*

src/Uno.Sdk/targets/Uno.Implicit.Packages.ProjectSystem.Uno.targets
```diff
@@ -4,25 +4,25 @@
...
 	-->
-	<ItemGroup Condition="$(UnoFeatures.Contains(';lottie;')) OR $(UnoFeatures.Contains(';material;')) OR $(UnoFeatures.Contains(';cupertino;')) OR $(UnoFeatures.Contains(';simpletheme;')) OR $(UnoFeatures.Contains(';skiarenderer;')) OR $(IsDesktop) == 'true'">
+	<ItemGroup>
 		<_UnoProjectSystemPackageReference Include="Uno.WinUI.Lottie" ProjectSystem="true" />
 	</ItemGroup>
 
-	<!-- SkiaSharp.Views are always included on Skia Renderer targets -->
-	<ItemGroup Condition="$(UnoFeatures.Contains(';skia;')) OR $(UnoFeatures.Contains(';lottie;')) OR $(UnoFeatures.Contains(';svg;')) OR $(UnoFeatures.Contains(';material;')) OR $(UnoFeatures.Contains(';cupertino;')) OR $(UnoFeatures.Contains(';simpletheme;')) OR $(UnoFeatures.Contains(';skiarenderer;')) OR $(IsDesktop) == 'true'">
+	<!-- Skia is the only renderer, so SkiaSharp.Views is always included -->
+	<ItemGroup>
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in `f44b3a31cb` — reworded to state the actual reason, and to make the scope explicit:

```xml
<!-- This file is only imported for Uno Platform targets (not WinAppSDK), where Skia is the
     only renderer, so SkiaSharp.Views is always included. -->
```

Worth noting for the record that the WinAppSDK concern doesn't apply here in practice: `Uno.Implicit.Packages.ProjectSystem.Uno.targets` is only imported for the android/desktop/ios/maccatalyst/macos/wasm/TFM-less platform identifiers — `windows10*` gets `Uno.Implicit.Packages.ProjectSystem.WinAppSdk.targets` instead. The comment was still ambiguous to a reader, so it's clarified.

doc/articles/features/using-skia-rendering.md
```diff
@@ -84,23 +83,10 @@ In order to accommodate the inclusion of Skia rendering for all platforms, the U
...
-
-- Given a library that uses `net10.0-ios` or `net10.0-android`, which does not have the `SkiaRenderer` set as an `UnoFeature`, but uses platform conditional code with `#if` blocks:
-
-  - This package is only usable by both native and Skia apps if it also provides a `net10.0` TFM. In this case, for a native app, nothing changes.
-  - For a `SkiaRenderer` enabled app, the `net10.0` variant of the library will be used and will not offer iOS/Android specific conditional code, but any code that uses Uno Platform provided APIs will work properly.
-
-- Given a library that uses `net10.0-ios` or `net10.0-android`, which does have the `SkiaRenderer` set as an `UnoFeature`, but uses platform conditional code with `#if` blocks:
-
-  - This package is only usable by Skia-enabled apps.
+A library that targets `net10.0-ios` or `net10.0-android` and uses platform conditional code with `#if` blocks needs to also provide a `net10.0` TFM to be consumable by an application. The `net10.0` variant is the one that gets used: it does not offer iOS/Android specific conditional code, but any code that uses Uno Platform provided APIs works properly.
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in `f44b3a31cb` — verified the claim and swept the contradicting pages:

- **`doc/articles/controls/SKCanvasElement.md`** — `Uno.WinUI.Graphics2DSK` is now injected unconditionally (ungated `ItemGroup` in `Uno.Implicit.Packages.ProjectSystem.targets`), so the "`Skia` or `SkiaRenderer` features present" framing and the `netX.0-android with SkiaRenderer` example are gone.
- **`doc/articles/features/using-skia-hosting-native-controls.md`** — dropped the "adding `SkiaRenderer` to `UnoFeatures`" intro and the `with SkiaRenderer` column qualifiers. Platform rows are unchanged: the Browser/Android/AppleUIKit/MacOS `INativeElementHostingExtension` implementations all still exist on this branch.
- **`doc/articles/uno-build-error-codes.md`** — UNOB0017 is raised by the Skia browser head package unconditionally, so "when the `SkiaRenderer` feature is enabled" was stale.
- **`doc/articles/features/using-skia-rendering.md`** — this page itself still told readers they could opt in to native rendering and "must opt in using MSBuild properties and features"; that prose is now corrected, and the upgrade tip points at the 7.0 migration guide.

`doc/articles/migrating-to-uno-6.md` is intentionally left as-is: it documents the 5.x → 6.0 migration, where `SkiaRenderer` *was* a real selectable feature, so rewriting it would make it historically inaccurate. The implied/no-op story is already documented in [`migrating-to-uno-7.md`](https://github.com/unoplatform/uno/blob/dev/mazi/2062-skiarenderer-gates/doc/articles/migrating-to-uno-7.md).

doc/articles/features/using-skia-hosting-native-controls.md
```diff
@@ -16,9 +16,9 @@ Each target platform has its own idea of a native element.
 | Skia Desktop (Win32)            | Uno.UI.NativeElementHosting.Win32NativeWindow                                      | A native Windows window with a unique `Hwnd` |
 | Skia Desktop (X11)              | Uno.UI.NativeElementHosting.X11NativeWindow                                        | A native X11 window with a unique `XID`      |
 | Skia Desktop (macOS)            | Not yet supported.                                                                 | Not yet supported as of Uno Platform 6.0     |
```
Xiaoy312: although not really in the scope, but dont we support macos already?

doc/articles/features/using-skia-rendering.md
Xiaoy312, line 92: so this is still not support...
any chance we could make it happen?

Xiaoy312: or, the very least, we should include alternatives here, like:
suggesting the user to use runtime branching such as: [OperatingSystem.IsIOS()](https://learn.microsoft.com/en-us/dotnet/api/system.operatingsystem.isios?view=net-10.0).

doc/articles/platform-specific-csharp.md
Xiaoy312, line 1: as using-skia-rendering.md‎ mentioned:
> A library that targets `net10.0-ios` or `net10.0-android` and uses platform conditional code with `#if` blocks needs to also provide a `net10.0` TFM to be consumable by an application. The `net10.0` variant is the one that gets used: it does not offer iOS/Android specific conditional code, but any code that uses Uno Platform provided APIs works properly.

if `__IOS/ANDROID__` are not going to work, we should remove them here to avoid confusion

Xiaoy312: and, almsot the entire section of this file could use a trim

Xiaoy312 (approved):

ajpinedam (approved):

doc/articles/features/using-skia-hosting-native-controls.md
MartinZikmund: Good question, and the honest answer is "yes and no", which is why that row reads oddly.

macOS is a fully supported platform, and `Uno.UI.Runtime.Skia.MacOS` really does implement `INativeElementHostingExtension` (`src/Uno.UI.Runtime.Skia.MacOS/Native/MacOSNativeElementHostingExtension.cs:27`), registered on every macOS host start (`Hosting/MacSkiaHost.cs:55`). But this table is about the *public* type an app developer hands us, and that piece is genuinely missing on macOS:

- `MacOSNativeElement` is `internal`, and `NativeHandle` has an `internal` setter (`MacOSNativeElementHostingExtension.cs:12,22`).
- The decisive gate is `public bool IsNativeElement(object content) => content is MacOSNativeElement;` (`:125`) — an app-supplied object can never be accepted, regardless of access modifiers elsewhere.
- There is no macOS counterpart to `Win32NativeWindow` / `X11NativeWindow` / `BrowserHtmlElement`; `src/Uno.UI/NativeElementHosting/` contains only those three types.
- Every native entry point is typed `NSView<UNONativeElement>*` (`UnoNativeMac/UNONative.h:12`), so a public wrapper would have to be written and conform to that protocol, not just exposed.

What the extension powers today is our own controls — `MacOSNativeWebView : MacOSNativeElement` and `MacOSMediaPlayerPresenterExtension`. Real shipping infrastructure, just not a public embedding surface.

You're right that it's outside this PR's scope, but I'll take the row anyway for one reason: this PR rewrote the intro to "Uno Platform apps always render with Skia, and you can embed native controls in them", and the macOS row now reads as contradicting it. So I've replaced the stale "Not yet supported as of Uno Platform 6.0" (we're on 7.0) with what is actually missing: "Hosting an arbitrary `NSView` is not publicly exposed yet."

Worth flagging for whoever picks it up: macOS is closer than the table suggests — opacity is already implemented there (`uno_native_set_opacity`, `:78`), unlike X11 where `ChangeNativeElementOpacity` is still entirely commented out (`X11NativeElementHostingExtension.cs:198`). It does not yet support Z-index (`INativeElementHostingExtension.SupportsZIndex()` defaults to `false` and macOS never overrides it). I've filed a follow-up to track exposing a public macOS native-element type.

doc/articles/features/using-skia-rendering.md
MartinZikmund: Correct — it is still the case today, and the mechanism is unchanged by the native-rendering drop. It lives in `RuntimeAssetsSelectorTask.HandleSkiaMobileForNonRuntimeEnabledPackages` (`src/SourceGenerators/Uno.UI.Tasks/RuntimeAssetsSelector/RuntimeAssetsSelectorTask.cs:432-543`): on an iOS/Android head, if a NuGet package ships both `lib/netX.0-ios|android` and `lib/netX.0` and its assembly references `Uno.UI`, both the compile reference and the runtime asset are rewritten to `lib/netX.0` (`:467-534`). So `#if __IOS__` inside that platform asset is not the code that ships.

**On "any chance we could make it happen" — yes, and that is where we want to land.** The target behaviour:

- a library that targets **only** plain `netX.0` keeps resolving the `netX.0` asset, as today;
- a library that is **multi-targeted** resolves its **platform-specific** asset, so `__ANDROID__` / `__IOS__` and the platform SDK APIs work as authors expect.

The reason the swap exists in the first place has actually gone away. It was there because the native renderers exposed a different `Microsoft.UI.Xaml` API surface, so a `-ios` asset compiled against native `Uno.UI` was not compatible with a Skia head. With native rendering dropped, `Uno.WinUI` ships **no platform-specific `lib/` folders at all** — only `lib/net10.0`, `lib/net11.0` and the WinAppSDK ones — so a `netX.0-ios` library already compiles against exactly the same `Uno.UI` the head runs.

Two things still need settling, which is why I'd rather not fold it into a gate-cleanup PR:

1. **Back-compat.** Packages published against Uno 6.x still carry `-ios`/`-android` assets compiled against the *native* `Uno.UI` and are on nuget.org today; the swap silently rescues them. We already stamp `AssemblyMetadata("UnoUIRuntimeIdentifier")` (`src/Uno.Sdk/targets/Uno.Features.targets:21-26`) and read it back in `RuntimeAssetsValidatorTask`, so a metadata-based gate is available — it needs designing.
2. **Sequencing after #23999.** `refactor(build)!: Fold UI-layer Reference flavor into Skia` deletes `src/AddIns/Uno.UI.MSAL/buildTransitive/Uno.WinUI.MSAL.targets` and strips the `UnoRuntimeEnabledPackage` groups from `build/nuget/uno.winui.common.targets` and the Lottie/Svg targets. That item group is exactly the exclusion set read at `RuntimeAssetsSelectorTask.cs:444` and applied at `:464`, so once it merges MSAL/Lottie/Svg fall *into* this code path for the first time. Landing both at once would conflate two behaviour changes.

Tracked as a follow-up; it should land after #23999.

One caveat that survives either way: even with the swap narrowed, a library still needs a plain `netX.0` TFM to be consumable, because an Uno head multi-targets `netX.0-ios;netX.0-android;netX.0-browserwasm;netX.0-desktop` and an `-ios`-only package cannot restore for the other legs. So the sentence stays true — I've left it, and I take the point that it reads as a renderer consequence when it sits under **Architecture / bait-and-switch**; that framing is worth revisiting when the behaviour actually changes.

---

On your second comment — agreed, and added. `OperatingSystem.IsIOS()` / `IsAndroid()` is the right primary answer and is already what `doc/articles/platform-specific-csharp.md:34` prescribes for Skia apps, so it should be surfaced here too. I've added a tip under the section, with one caveat spelled out: runtime branching lets you branch behaviour, but you can't call `UIKit`/`Android.*` types from a `netX.0` TFM. For those, the working pattern is an interface exposed from the `netX.0` library with the platform implementation injected from the head, which does target `netX.0-ios`/`netX.0-android`.

Leaving this thread open for now since the first half is a scope call rather than a docs fix.

doc/articles/platform-specific-csharp.md
MartinZikmund: I'd rather not remove them — the two statements are about different things, and I think the fix is to make the table precise.

`using-skia-rendering.md` is about **which asset an app consumes**: when a library multi-targets `netX.0-ios` alongside a plain `netX.0`, the app resolves the `netX.0` one, so the iOS-conditional code in the iOS asset isn't what ships. That's not the same as the symbol being undefined. An **application head** targets `netX.0-ios` / `netX.0-android` directly, so `#if __IOS__` there compiles and runs normally — and heads are the primary place people write `#if`. Dropping the rows would make the table wrong for that case.

(For the record, `__ANDROID__` and `__IOS__` come from the .NET Android/iOS SDKs themselves, not from anything we ship — I verified `__ANDROID__` fires on a bare `net10.0-android` project with no Uno involvement.)

So I've added a note under the table stating the per-TFM scoping explicitly, plus the class-library caveat with a link back to the section you quoted, and — per your other comment — pointed at `OperatingSystem.IsXXX` as the preferred approach in libraries. I also fixed the `__WASM__` row's "see below" link while in there; it targeted `#webassembly-considerations`, an anchor that has been dangling since `3a96e2345f` folded that section into `### OperatingSystem.IsXXX`.

---

On the trim: agreed, and it goes well beyond the table. Concretely, in follow-up scope:

- **`__APPLE_UIKIT__` row** is wrong for consumers. Its only definition site in the whole repo is `src/Uno.CrossTargetting.targets:102`, which is our internal cross-targeting file and isn't packaged. The shipped `Uno.WinUI.Runtime.Skia.AppleUIKit.props:9` defines `__UNO_SKIA__;__UNO_SKIA_APPLE_UIKIT__`, not `__APPLE_UIKIT__`. Want to confirm with an Uno.Sdk iOS head build before editing.
- **`__DESKTOP__` row** is under-qualified: it comes from the five desktop runtime packages' `buildTransitive` targets (`…Win32.targets:10` and siblings), which `Uno.Implicit.Packages.ProjectSystem.Desktop.targets:28` only injects when `_IsExecutable == 'true'`. A `netX.0-desktop` **class library** gets neither `__DESKTOP__` nor `__UNO_SKIA__`.
- **File-suffix list:** `*.reference.cs` never compiles in an Uno.Sdk project — the glob requires `UnoRuntimeIdentifier == 'Reference'` (`src/Uno.Sdk/targets/Uno.CrossTargeting.targets:24-25`) and the SDK only ever sets it to `Skia`, on Desktop (`Uno.Common.Desktop.targets:5`). `*.crossruntime.cs` is documented as "WebAssembly, Desktop, and reference implementation" but is desktop-only in practice.
- **Type-alias example** still uses `Windows.UI.Xaml.UIElement` (UWP namespace — 7.0 ships only `Uno.WinUI`) and aliases `Android.Views.View` / `UIKit.UIView`, which is a native-rendering-era pattern; with Skia everywhere, `UIElement` *is* the visual-tree node on every Uno target.
- **`NativeWrapperControl` example** teaches per-platform `CreateNativeView()`; in 7.0 that goes through native element hosting and should point at `xref:Uno.Skia.Embedding.Native`.
- **`NETFX_CORE` row** — UWP heads are gone in 7.0.
- **Section ordering** — `OperatingSystem.IsXXX` should lead for class libraries, with `#if` positioned as the tool for heads and for code that can't compile everywhere.

That's a genuine rewrite and it overlaps #23999's Reference-flavor removal, so I'd keep it out of a PR whose whole job is deleting vacuous `skiarenderer` gates. Filed as a follow-up and I'll take it.

doc/articles/features/using-skia-rendering.md
MartinZikmund: @Xiaoy312 ignore that long paragraph, but in short - created https://github.com/unoplatform/uno-private/issues/2283 to track this change

PR #23978, opened 2026-08-05
feat!: Rewrite relative XAML URIs to the MRT local-resource form

Xiaoy312:
**GitHub Issue:** Part of #8339

## PR Type:

✨ Feature

## What changed? 🚀

BC51 of the 7.0 rollup: a relative URI in XAML now compiles to the MRT local-resource form
WinUI produces (`ms-resource:///Files/…`) instead of being emitted verbatim. This fills the
`else` branch that has carried a `// Breaking change … (#8339)` placeholder in
`XamlFileGenerator.RewriteUri` since that issue was filed.

Expected values were **measured on native WinUI (WinAppSDK 1.7)** with a self-contained
unpackaged probe app, for both application and library XAML, rather than inferred:

| WinUI target | Compiled value |
|---|---|
| a value landing in a `BitmapImage` — `BitmapImage.UriSource`, or an `ImageSource`-typed property (`Image.Source`, `ImageBrush.ImageSource`, custom DP) given a non-svg path | `ms-appx:///` + **file-relative** base |
| a value landing in an `SvgImageSource` — **including** `Image.Source="x.svg"` | `ms-resource:///Files/<raw>` |
| every other `Uri`-typed property — `BitmapIcon`/`BitmapIconSource.UriSource`, `NavigateUri`, `WebView2.Source`, custom `Uri` DPs | `ms-resource:///Files/<raw>` |

Two properties of that rewrite drove the implementation:

1. **The `ms-appx` special case is not keyed on the owner type.** Uno keyed it on the owner,
   which was wrong in both directions — `ImageBrush.ImageSource` and a custom `ImageSource` DP
   fell through to the verbatim branch, while a `Uri`-typed property on an `ImageSource`-derived
   owner took the `ms-appx` branch. Uno now keys on the **property** type.
2. **The `ms-resource` rewrite is a raw prefix concat**, dropping a leading `/` and ignoring the
   folder of the XAML file — with no assembly prefix, even in a library.

`ResourceDictionary.Source` is untouched — it resolves through `RetrieveDictionaryForSource`,
not `RewriteUri`.

### The svg carve-out is a deliberate divergence

Uno keeps **every** `ImageSource`-typed property on the `ms-appx` form, svg included. WinUI does
not, as the table above shows — that row was probed for this PR and had never been measured
before.

The discriminator is **how the sink resolves the form, not what the form can express**.
`ms-appx` is resolved as an asset path, so it carries the assembly prefix a library's own assets
need; `ms-resource` is understood only by Uno's mapping, which resolves against the app root.
Following WinUI here would leave a library's svg assets unreachable, with no MRT to fall back
on. (An assembly-qualified `ms-resource` *is* emitted for `ResourceDictionary.Source`, because
dictionary resolution is a registry string-key lookup rather than asset resolution — so the form
itself is not the constraint. An earlier version of this PR justified the carve-out with exactly
that mistaken claim.)

### Keeping images loading

`ms-resource:///Files/X` is mapped to `ms-appx:///X` — the equivalent of MRT's `Files` subtree —
so relative-URI `BitmapIcon`/`BitmapIconSource` keep loading the same asset they load today.
**The value changes; the behavior does not.**

The mapping sits where a URI is *consumed*, not only in the `ImageSource` conversion:
`ImageSource.InitFromUri` (which every `UriSource` assignment reaches through
`OnUriSourceChanged`, on every target), the `BitmapImage` Skia loader (the one place a raw DP
value is read), and `LoadedImageSurface.StartLoadFromUri`. A `UriSource` assigned directly —
ctor, `x:Bind`, a `Setter`, code-behind — therefore resolves too, while still *reporting* the URI
it was given, the way WinUI reports a `Uri`-typed property.

The mapping preserves the fragment (load-bearing for `SvgImageSource`) and rejects a non-empty
authority: `ms-resource://OtherPkg/Files/x.png` names another package's resources, which
`ms-appx` cannot reach, so mapping it would silently resolve the *app's* asset of the same name.
An unmappable `ms-resource` URI now warns rather than failing silently.

### Also in this PR

- **A path may contain a backslash**, and the `ms-appx` branch emitted a regular string literal,
  so `ImageSource="Assets\x.png"` had `\x` read as an escape and dropped the generated file
  entirely. Both branches now emit verbatim literals.
- **`UriExtensions.IsLocalResource` → `IsMsAppx`** (source-breaking). The old name said "local
  resource" but the predicate tests for `ms-appx` — the opposite end of this very mapping. The
  internal helpers were renamed to match (`MsResourceFilesPrefix`, `NormalizeMsResourceFilesUri`).
- **`UriExtensions.IsAppData` guarded against a relative URI** — it read `uri.Scheme` behind only
  a null check, where the neighbouring predicate correctly guards `IsAbsoluteUri &&`.
  Defensive hardening; no known caller reaches it.
- **`Launcher` on Skia reports launch failures the way WinUI does.** Measured on WinAppSDK 1.7:
  `LaunchUriAsync` returns `true` without launching anything for `ms-resource`, `ms-appx` and
  unregistered schemes alike. Uno's Win32 extension let a `ShellExecute` failure escape into
  `HyperlinkButton.OnClick`, which is `async void`.

`build/PackageDiffIgnore.xml` **does** carry an entry for this PR — the `IsLocalResource`
removal. Nothing else publicly visible is removed or retyped; `XamlFilePathHelper` is internal.

## Validation

Per `debugging-discipline.md`, labelled by evidence type. All local evidence is **Skia desktop,
net10.0** — `net11.0` is not buildable on this SDK and is left to CI.

- **Code review:** every consumer of a raw URI value was traced before deciding where to map.
  `OnUriSourceChanged → InitFromUri` covers all platforms, which is why the conversion operator
  does not map (it would only change the value a caller reads back).
- **Compile (analyzers on,** `-p:UnoFastDevBuild=false -p:OptimizeImplicitlyTriggeredBuild=false`**):**
  `Uno.UI.Skia`, `Uno.UI.Reference`, `Uno.Skia`, `Uno.Foundation.Skia` and the netstandard2.0
  `Uno.UI.SourceGenerators` — **0 errors, 0 warnings**. The generator matters separately: it
  compiles `XamlFilePathHelper.shared.cs` too, where the span-based `string.Concat` overload does
  not exist.
- **Runtime — generator tests:** full suite → **482 total, 480 passed, 2 skipped, 0 failed**. A
  `Given_RelativeUri` golden locks the emitted shape for every property kind, including a rooted
  path, a backslash, `SvgImageSource`, and the `Setter` path (which reaches the rewrite through
  the style's target type rather than the syntactic owner).
- **Runtime — unit tests:** new `Given_XamlFilePathHelper` → **21/21**, covering scheme/folder
  casing, query and percent-escape survival, the fragment and foreign-package cases, and
  `TryGetMsAppxAssetPath`.
- **Runtime — runtime tests:** `Given_RelativeUriRewrite` → **11/11, 0 inconclusive**, plus
  `Given_BitmapSource` → 15 passed / 1 pre-existing skip. Value assertions measured against WinUI
  run on the WinUI head too; the `ms-appx`-value and mapping assertions are excluded there, since
  MRT resolves the local-resource URI natively.
- **Fails-before / passes-after**, each verified by reverting one thing at a time:
  - neutering the scheme constant the mapping compares → **8 of the 21** unit rows fail;
  - dropping *only* the new authority clause → exactly the foreign-package row flips, which is
    the only proof that half of the fix does anything;
  - reverting the mapping in `ImageSource` → 4 of the 6 tests that existed at that point fail
    (the 2 that pass are intentional regression guards);
  - reverting only the `BitmapImage` Skia loader line → `When_Local_Resource_On_UriSource_Loads`
    times out (which also proves `ActualHeight > 0` is a real load signal, not a template-size
    proxy);
  - un-fixing the verbatim literal → the generated `MainPage` file disappears from the
    compilation;
  - re-adding the mapping in the conversion operator → the new value test fails;
  - pointing `BitmapIconSource` at a missing asset → its load test fails.

**Not validated locally:** the WinUI head, WASM, native Android/iOS, and hot reload — left to CI.
Six of the eleven `IsMsAppx` call sites (mobile, WASM, macOS, AddIns) are not compiled by any
project buildable here; their completeness proof is a repo-wide zero-hit grep for the old name
across every text format, and the mobile/WASM/macOS CI legs are the real check. The
`PackageDiffIgnore` entry is likewise CI-gated.

## PR Checklist ✅

- [x] 🧪 Added [Runtime tests, UI tests, or a manual test sample](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) (for bug fixes / features, if applicable) — 11 runtime tests, 21 unit tests, 2 source-generator goldens
- [x] 📚 Docs have been added/updated following the [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features) — migration guide updated

### Breaking change impact and migration path

This changes the compiled value of a relative URI, so it is a **breaking change** — intentional,
scoped to the 7.0 major, and toward WinUI parity. Documented in
[`migrating-to-uno-7.md`](doc/articles/migrating-to-uno-7.md). Four impacts:

1. **A relative value is now absolute at the point of use.** `WebView2.Source` with a relative
   value used to throw `ArgumentException` while the page initialized; it now navigates an
   `ms-resource:` URI no web view can service, so the failure is silent instead of loud. A
   scheme-less `NavigateUri` (`www.example.com`) becomes `ms-resource:///Files/www.example.com`
   and reaches the launcher as an unregistered scheme, and a relative
   `Control.DefaultStyleResourceUri` no longer throws while resolving.
   **Migration:** give these properties absolute URIs — `https://…`, `ms-appx:///…`.
   Measured on WinAppSDK 1.7, `Launcher.LaunchUriAsync` returns `true` without launching
   anything for `ms-resource` and `ms-appx` alike, so a relative `NavigateUri` is equally inert
   on WinUI — the silent failure is parity, but it is quieter than the exception Uno used to
   raise.
2. **In a library, the two property kinds resolve differently.** On an `ImageSource`-typed
   property the value now carries the library's assembly prefix (`ms-appx:///MyLib/Assets/x.png`),
   so a library's own assets resolve — but a library asset that previously resolved from the
   *consuming app's* root no longer does. On a `Uri`-typed property the `ms-resource` form is
   mapped against the app root, so it does not. **Migration:** a library shipping assets for
   `BitmapIcon` must use an explicit `ms-appx:///MyLib/…` URI.
3. **`ImageBrush.ImageSource` and custom `ImageSource` properties** previously kept the relative
   string and now resolve against the base URI, consistent with `Image.Source`.
4. **`UriExtensions.IsLocalResource` is renamed to `IsMsAppx`** (source-breaking, no forwarding
   shim). Same behavior. **Migration:** rename the call.

Images declared in application XAML keep loading the same asset. To keep any value verbatim,
write it as an explicit `ms-appx:///…` URI in XAML.

### Known divergence left open

Uno resolves the `ms-appx` form against the **assembly root**; WinUI resolves it against the
**XAML file's folder**. `Assets/logo.png` in `Views/MainPage.xaml` is
`ms-appx:///Assets/logo.png` here and `ms-appx:///Views/Assets/logo.png` on WinUI. This predates
BC51 and is unchanged by it, but it is a real behavior gap and is now named in the migration
note; it needs its own BC entry, since fixing it changes compiled values again.


MartinZikmund:
@Xiaoy312 there is a cspell issue

MartinZikmund (approved):

MartinZikmund:
@copilot resolve the merge conflicts in this pull request

Xiaoy312:
wip: merge conflict...

Xiaoy312:
Rebased onto `feature/breakingchanges` (96 commits, including the sweeping `.skia.cs` → `.cs` fold).

Two of this PR's files were folded away on the base, so their hunks were hand-applied to the merged files:

- `BitmapImage.skia.cs` → `BitmapImage.cs`
- `LoadedImageSurface.skia.cs` → `LoadedImageSurface.cs`

The other conflicts were append-vs-append in `PackageDiffIgnore.xml` and `migrating-to-uno-7.md`. No behavioral change — the branch's net contribution is byte-identical to the approved version apart from those two path renames.

Re-verified locally after the rebase: `Given_XamlFilePathHelper` (21/21), `Given_RelativeUri` generator snapshots (2/2), `Given_RelativeUriRewrite` runtime tests on Skia Desktop (11/11), `Given_BitmapSource` (14 passed, 1 pre-existing `[Ignore]` on #9080), and a clean `SamplesApp.Skia.Generic` build.

Issue #22788, opened 2026-03-06
fix: Algolia DocSearch results are duplicated and search bar spans full viewport width


Jen-Uno:
## Current behavior

The Algolia DocSearch integration in the docs site has several issues:

1. **Duplicate search results**: The DocSearch crawler indexes breadcrumb navigation, heading-level anchors, and same-titled pages from different sections, producing duplicate or near-duplicate results.
2. **Search bar spans full viewport width**: When the search modal opens, the search input bar stretches across the entire screen instead of staying within the modal. This is caused by docfx's default CSS rule `header { position: fixed; top: 0; left: 0; }` unintentionally applying to DocSearch's `<header class="DocSearch-SearchBar">` element.
3. **Search not visible in local builds**: The custom TOC template (`toc.html.tmpl`) that renders the DocSearch button is silently ignored by newer docfx versions, which expect `toc.html.primary.tmpl`.

## Expected behavior

- Search results should be unique and deduplicated
- The search bar should stay within the modal bounds (~560px)
- Search should be visible and functional in local doc builds

## How to reproduce

1. Visit the docs site and open DocSearch (Cmd+K)
2. Search for "MVUX" or any common term
3. Observe duplicate results and the search input spanning the full viewport width

## Workaround

None

## Works on

All platforms (browser-based docs site)

## Issues found and fixed

1. **Breadcrumb duplicates** — The crawler indexed `<div id="breadcrumb">` content as separate search records with `#breadcrumb` anchors. Fixed by filtering those in `transformItems` and adding `data-docsearch-exclude` to the breadcrumb template.

2. **Same-page heading duplicates** — DocSearch creates one record per heading level, so a single page could appear multiple times. Fixed with URL-based deduplication that keeps only the first hit per base page URL.

3. **Same-titled different pages** — Pages like workshop variants shared identical titles, making results look like duplicates. Fixed by detecting same-titled groups and appending a distinguishing URL path segment to the display text (updating both `_snippetResult` and `_highlightResult` so DocSearch actually renders it).

4. **Navigation noise in index** — The crawler was indexing navbar, footer, and sidebar content. Fixed by adding `data-docsearch-exclude` to those elements via new template overrides (`affix.tmpl.partial`, `footer.tmpl.partial`, `breadcrumb.tmpl.partial`, and `_navbar.tmpl.partial`).

[...]

PR #23992, opened 2026-08-06
docs: modernize SignalR integration tutorial

### LINKED ISSUE #22607: [Docs] Feedback: SignalR Docs Tutorial/Sample
### On which page? 📄

https://platform.uno/docs/articles/signalr.html

### What's wrong? ❗

To me its unclear, how this Sample/Tutorial exactly shows to me the integration in a Uno App, because it seems to only show ASP NET Core Programm.cs code, but nothing I would know from my regular Uno App Template.

The Tutorial should be improved by adding a Uno App context, like Supported Platforms and what we can expect in the end to see. The current registration in asp net core makes me as new to SignalR-User assume, that:
- I will end up with Razor Pages in my Uno App Solution instead of getting a Uno Wasm target driven UI? (see the `MapRazorPages()` call in the second code sample block contained in this docs page!)
- The Questions:
  - "Does Uno or at least its browserwasm target behave exactly like Razor Pages?"
  - "So am I able to register and show my Uno Wasm UI on a ASP NET Core Server to be shown at runtime?"
  - "Is the `Uno.Wasm.Bootstrap.Server` NuGet of Uno possibly required for exactly this?" since I only have been able to find docs for non-Server package using samples in the Bootstrapper docs (*not* equivalent to the Uno App Templates contained Server option, I am aware of this! But this is exactly the missing point!)

The UnoChat Sample Readme in Uno.Samples Repo does compared to this docs page show to me a (Web?) UI and Console app, while I would not know of any way to have a Console App as UI in Uno (I indeed tryed this, but was not possible to establish using the current docs. nobody tells what hostbuilder setup in programm.cs I would be needing. And also the sample global.json file shows to me a 5.3.96 version, which is defintly not a Version I would rely on, when searching for a lookup. too many versions and especially breaking changes (SkiaRenderer, Desktop target...) did happen meanwhile.

additionally, it would make sense, to let the reader know of all of the code we seem to need in the ViewModel, if I would want to beleave the Sample eitherway:
https://github.com/unoplatform/Uno.Samples/blob/master/UI/ChatSignalR/src/UnoChat/Presentation/ViewModel.cs
and turns out, the answer to the filed question above with the "will I end up using Razor Pages in a Uno App as UI instead of a Uno Wasm target" is "Yes, you will rather use Razor then Web Assembly UI build with the Uno.Sdk!"

Image

### Any feedback? 💬

Just a feedback from my POV, this tutorial paints to me a really sad image and impression when it comes to the Uno Wasm Target integration.

I know that we can not have a server side rendering, but is our Wasm target really this kind of bad integrate-able with a ASP NET Core Server, so we need to make advertisement for sending Uno Users over to using Razor / Blazor instead, where *it seems to be possible to integrate and fully/detailed document this* with a Server project? 🤷 

I have been searching for a Gprc using sample, like the Maui Client of the dotnet/eShop Sample is doing, but while I am quite sure there have been one in the Uno docs in the past, all I can find now is the Migration page here: https://platform.uno/docs/articles/guides/silverlight-migration/08-data-access-overview.html#_content of silverlight, ***which projects has been removed from VS support up from the 2017 Version!!!***

https://learn.microsoft.com/en-us/visualstudio/releases/2026/port-migrate-and-upgrade-visual-studio-projects
> <html>
> <body>
> 
> Silverlight | Silverlight projects were removed in Visual Studio 2017.
> -- | --
> 
> </body>
> </html>

## Child issues
- [ ] https://github.com/unoplatform/Uno.Samples/issues/910

Jen-Uno:
## What this PR does

Modernizes the SignalR integration tutorial (`doc/articles/signalr.md`):

- Rewrites the intro to describe ASP.NET Core SignalR and its transports, and when it's a good fit for Uno apps.
- Updates prerequisites to **.NET 9 SDK**, **Visual Studio 2022 17.8+ / VS Code / Rider**, and clarifies that an **Azure account is not required** (local dev + production hosting pointers).
- Replaces the `Startup.cs` / `ConfigureServices` pattern with the current **`Program.cs`** minimal-hosting model.
- Uses the `dotnet new web` + `dotnet add package Microsoft.AspNetCore.SignalR.Client` CLI flow, structured into clear "Create the server" / "Add SignalR to the Uno client" steps.
- Notes cross-platform client support (WASM, Desktop/Skia, iOS, Android) and adds production CORS guidance.

## Why it's a separate PR

This content was originally authored on the Algolia DocSearch branch (PR #22789 / issue #22788) but is unrelated to search infrastructure. Per the review recap in `unoplatform/uno-private#2038`, unrelated content changes should be tracked and merged separately — so it's split out here.

Fixes #22607

### DIFF
--- a/doc/articles/signalr.md
+++ b/doc/articles/signalr.md
@@ -2,51 +2,275 @@
 uid: Uno.Development.SignalR
 ---
 
-# SignalR
+# How to integrate SignalR with an Uno Platform app
 
-[SignalR](https://learn.microsoft.com/aspnet/core/signalr/introduction?view=aspnetcore-7.0) is an ASP.NET Core library that allows server-side code to be instantly pushed to the client.
+[ASP.NET Core SignalR](https://learn.microsoft.com/aspnet/core/signalr/introduction) is an open-source library that simplifies adding real-time web functionality to apps. It enables server-side code to push content to connected clients instantly using WebSockets, Server-Sent Events, or Long Polling (chosen automatically based on capabilities).
+
+SignalR is a good fit for Uno Platform apps that need:
+
+- Real-time messaging or chat
+- Live dashboards and monitoring
+- Collaborative editing
+- Push notifications from a server
+
+This guide walks you through creating an ASP.NET Core SignalR server and connecting to it from an Uno Platform client app. The SignalR .NET client (`Microsoft.AspNetCore.SignalR.Client`) works on all Uno target platforms — WebAssembly, Desktop (Windows, macOS, Linux via Skia), iOS, and Android.
 
 ## Prerequisites
 
-- Visual Studio 2019 or higher
-- Azure account (to publish SignalR service)
+- [.NET 9.0 SDK](https://dotnet.microsoft.com/download) or later
+- An Uno Platform app (created via `dotnet new unoapp` or the Uno Platform Visual Studio extension)
+- Visual Studio 2022 17.8+, VS Code, or JetBrains Rider
 
-## Steps
+> [!NOTE]
+> An Azure account is **not** required. You can run the SignalR server locally during development. For production hosting options, see [Publish an ASP.NET Core SignalR app to Azure](https://learn.microsoft.com/aspnet/core/signalr/publish-to-azure-web-app) or [Azure SignalR Service](https://learn.microsoft.com/azure/azure-signalr/signalr-overview).
 
-1. Create `ASP.NET Core web application` in Visual Studio and name it `UnoChat.Service`.
+## Step 1 — Create the SignalR server
 
-    ![project-template](Assets/project-structure.JPG)
+Create a new ASP.NET Core Web API project that will host the SignalR hub:
 
-2. Add [SignalR Hub](https://learn.microsoft.com/aspnet/core/tutorials/signalr?view=aspnetcore-7.0&tabs=visual-studio#create-a-signalr-hub) to your `[YourProjectName].Service` project in a `Hubs` folder.
+```dotnetcli
+dotnet new web -n ChatServer
+cd ChatServer
+```
 
-3. In your `Startup.cs` file, add your `SignalR` service and a `CORS policy` to the `ConfigureServices` method.
+### Define a SignalR Hub
 
-    ```csharp
-    public void ConfigureServices(IServiceCollection services)
+Create a `Hubs` folder and add a `ChatHub.cs` file:
+
+```csharp
+// Hubs/ChatHub.cs
+using Microsoft.AspNetCore.SignalR;
+
+namespace ChatServer.Hubs;
+
+public class ChatHub : Hub
+{
+    public async Task SendMessage(string user, string message)
     {
-        services.AddRazorPages();
-        services.AddSignalR();
-        
-        services.AddCors(o => o.AddPolicy(
-            "CorsPolicy",
-            builder => builder
-                .AllowAnyOrigin()
-                .AllowAnyMethod()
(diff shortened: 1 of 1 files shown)


doc/articles/signalr.md
```diff
@@ -2,51 +2,275 @@
...
+        }
+    }
+
+    private async void SendButton_Click(object sender, RoutedEventArgs e)
+    {
+        if (_connection is not null &&
+            !string.IsNullOrWhiteSpace(UserTextBox.Text) &&
+            !string.IsNullOrWhiteSpace(MessageTextBox.Text))
+        {
+            await _connection.InvokeAsync(
```
morning4coffe-dev, line 179: [P2] Handle send failures when the connection is not active

`_connection` is assigned before `StartAsync` completes and remains non-null after startup failure or disconnection. The Send button stays enabled, so this branch invokes the hub while disconnected. Running this unchanged handler with the real SignalR client produced `InvalidOperationException: The 'InvokeCoreAsync' method cannot be called if the connection is not active` both before connection and after `StopAsync`; the exception escaped through the async-void synchronization context. The connected control successfully broadcast the message and cleared the input.

Gate the send UI/state on an active connection and catch/report invocation failures inside the handler as well, since the connection can drop between a state check and the invocation.

doc/articles/signalr.md
```diff
@@ -2,51 +2,275 @@
...
+
+This guide walks you through creating an ASP.NET Core SignalR server and connecting to it from an Uno Platform client app. The SignalR .NET client (`Microsoft.AspNetCore.SignalR.Client`) works on all Uno target platforms — WebAssembly, Desktop (Windows, macOS, Linux via Skia), iOS, and Android.
 
 ## Prerequisites
 
-- Visual Studio 2019 or higher
-- Azure account (to publish SignalR service)
+- [.NET 9.0 SDK](https://dotnet.microsoft.com/download) or later
+- An Uno Platform app (created via `dotnet new unoapp` or the Uno Platform Visual Studio extension)
+- Visual Studio 2022 17.8+, VS Code, or JetBrains Rider
```
morning4coffe-dev, line 22: [P2] Align the Visual Studio prerequisite with .NET 9

Visual Studio 17.8 supports targeting .NET 8, not .NET 9. The stated .NET 9 target requires Visual Studio 2022 17.12 or newer; installing the standalone SDK does not add that target support to 17.8. Please raise the IDE floor or explain the CLI-only alternative. Microsoft's [SDK/Visual Studio support table](https://learn.microsoft.com/dotnet/core/porting/versioning-sdk-msbuild-vs#targeting-and-support-rules) documents the distinction.

morning4coffe-dev (changes requested):
Reviewed the complete tutorial rewrite at `ae788d74589120b91476fe06a3e33c680e4146d7`. The extracted server/Hub compiled and a real loopback SignalR client successfully sent and received the sample broadcast. Two corrections are detailed inline: disconnected sends escape the new async-void handler, and the Visual Studio floor does not support the stated .NET 9 target.

I also reproduced the already-reported directory issue: after the documented `cd ChatServer`, `dotnet run --project ChatServer` reports that the project path does not exist. Please fix the existing thread rather than treating this as a second inline report from me. The current platform table has ordinary single leading pipes; I did not count the earlier table warning as a reproduced finding.

Evidence scope: unchanged tutorial server, Hub and ChatPage source with the real SignalR 10.0.0 client; generated UI controls/Page and exception-observing synchronization context were stand-ins. This was not an actual Uno UI, browser/mobile, TLS-trust, production-CORS or deployment run.

PR #24012, opened 2026-08-07
fix(hr): don't block hot reload on non-fatal source-generator errors

dr1rrb:
## Problem

Surfaced by unoplatform/studio.live#3317 (Uno SDK `6.7.0-dev.914` → `6.7.0-dev.938`, which moves embedded Roslyn `4.14` → `5.6`). The regression guard `ReswSupportTests.When_ReswCorruptedDuringHotReload_Then_XamlEditsStillWork` passes on `914` and fails on `938`.

**Root cause.** Roslyn 5.x's Edit-and-Continue (`UnitTestingHotReloadService.EmitSolutionUpdateAsync`) refuses to emit **any** metadata update while a source generator reports an **Error**-severity diagnostic — even when the C# compilation itself succeeds. A corrupt `.resw` makes the Uno XAML generator report `UXAML0003` (Error) on **every** generator pass while the assembly still compiles, so a single broken generator input froze **all** hot reload (including valid edits to unrelated files) until the `.resw` was fixed. Roslyn 4.x did **not** gate on generator diagnostics — hence `914` green, `938` red.

The trap: source-generator diagnostics never appear in `Compilation.GetDiagnostics()` (only via the internal `Project.GetSourceGeneratorDiagnosticsAsync()`), so the change-set-scoped blocked-compilation audit (spec 054) can't see them — the pass reports `NoChanges`, not `Failed`, and the reason is invisible.

Confirmed with a source-generated-document dump: after the XAML edit with a corrupt `.resw`, the generated `MainPage` code **does** contain the edit (the generator is innocent), yet EnC emits 0 updates while `UXAML0003` is present and emits normally once it is gone. (EnC baseline / `commitUpdates` is not the cause — hot reload recovers as soon as the `.resw` is valid again.)

## Fix

In `HotReloadManager.ProcessSolutionChanged`, read the change-set projects' source-generator **error** diagnostics (via `Project.GetSourceGeneratorDiagnosticsAsync`, resolved through the same `ResolveAuditProjects` scope as the spec-054 audit) and **suppress them for the EnC emit** (`SpecificDiagnosticOptions → Suppress`). A non-fatal generator error then degrades to *stale generated output* instead of *frozen hot reload*, restoring 4.x behavior. Generated code that genuinely does not compile still reports a C# (`CSxxxx`) error and stays blocked through the existing audit / rude-edit paths, so this never lets non-compiling code hot-reload. The bypassed errors are surfaced in the HR log so a broken `.resw` isn't silent.

The suppression is applied to the emit itself, not as a retry after a blocked emit: `EmitSolutionUpdateAsync` advances the EnC baseline as a side effect, so a throwaway first emit would consume the change and leave a second nothing to apply.

Also centralizes hot-reload diagnostic reporting — compilation errors, EnC emit diagnostics, solution-update diagnostics, and the new bypassed-generator diagnostics — into a per-cycle `DiagnosticsReporter` (severity coloring, per-category WASM formatting caps since memory.grow() is irreversible); `GetCompilationErrors` becomes a pure getter.

## Verification

- studio.live `ReswSupportTests` — `When_ReswCorruptedDuringHotReload` + all 4 sibling resw tests pass (Desktop; the Wasm rows are unrelated to this fix — they need the `wasm-tools` workload to build the sample).
- `Uno.HotReload.Tests` — 135/135 pass.

### DIFF
--- a/src/Uno.HotReload/HotReloadManager.cs
+++ b/src/Uno.HotReload/HotReloadManager.cs
@@ -2,13 +2,11 @@
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Diagnostics;
-using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
-using Microsoft.CodeAnalysis.CSharp;
 using Uno.HotReload.Diffing;
 using Uno.HotReload.Microsoft;
 using Uno.HotReload.Tracking;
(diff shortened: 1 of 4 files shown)


kazo0:

kazo0 (approved):

ajpinedam (approved):

PR #24020, opened 2026-08-09
refactor: Drop guards and references the fold made unreachable

MartinZikmund:
**GitHub Issue:** part of #8339

> Stacked on #24001 — targets `dev/mazi/skianaming`, and should be reviewed and merged after it. Retarget to `feature/breakingchanges` once #23999 and #24001 land.

## PR Type:

## What changed? 🚀

Follow-up cleanup to the Reference fold (#23999) and the `.Skia` rename (#24001). Now that the UI layer builds only for Skia, a family of preprocessor guards can no longer go either way — this removes what has become unreachable. No behaviour change: every branch removed is one the compiler was already discarding.

**Dead preprocessor branches — 264 sites, 92 files, ~1400 lines**

`__NETSTD_REFERENCE__` is never defined for these projects, while `UNO_HAS_ENHANCED_LIFECYCLE`, `UNO_HAS_BORDER_VISUAL` and `UNO_HAS_MANAGED_SCROLL_PRESENTER` are always defined. Branches that can no longer be taken are removed; guards that are now always true are unwrapped, keeping the `#else` body wherever one existed.

Five files were nothing but a dead guard and are deleted outright: `BorderLayerRenderer`, `Control.nonlifecycle`, `IScrollContentPresenter`, `NativeScrollContentPresenter.Xamarin`, `CustomEventManager`.

**Scope was the careful part.** Only directories whose *only* compiling project carries `UnoRuntimeIdentifier=Skia` are swept. Deliberately untouched, because each still has a Windows or netcoremobile head globbing the same sources, where these symbols are undefined and the branches are still live:

| Untouched | Why |
|---|---|
| `Uno.UI.Toolkit` | `Uno.UI.Toolkit.Windows.csproj` |
| `Uno.UI.MSAL` | `.Windows` + `.netcoremobile` heads |
| `Uno.WinUI.Graphics2DSK` | `.Crossruntime` + `.Windows` heads |
| `Uno.WinUI.Graphics3DGL`, `Uno.UI.Foldable`, `Uno.UI.GooglePlay` | no Skia RID |

Cross-project `<Compile Include>` links were checked the same way — 212 files are compiled from outside their own directory, and none of the ones reachable from a non-Skia project carries any of these guards. `Generated/` is excluded: those stubs come from `Uno.WinAppSDKSyncGenerator` and have to be fixed there.

Compound conditions mixing in symbols outside this set (`IS_UNIT_TESTS`, `HAS_UNO`, `IS_NATIVE_ELEMENT`, `DEBUG`) are left alone rather than guessed at — about 30 sites.

**Dangling `InternalsVisibleTo` — 20 entries, 9 files**

Names this repo used to produce and no longer does: `Uno.UI.Runtime.WebAssembly`, `Uno.UI.MediaPlayer.WebAssembly`, `SamplesApp.Wasm` and `SamplesApp.netcoremobile` went with the native drop, `Uno.UI.RuntimeTests.Wasm` with the WASM test head, `Uno.UI.Svg.Skia` with the add-in fold. Each was verified against `master` to have been an in-repo project — `Uno.UI.Maps` and `Uno.UI.HotDesign.Client` are left in place for exactly that reason, as are names that never matched a produced assembly and so predate this work.

**Two stale descriptions**

`RemoteControlClient` still described its `#else` arm as covering "the Reference build", and the `PrismJsView` sample in the WASM interop guide imported the removed `Uno.UI.Runtime.WebAssembly` namespace *and* declared `namespace PrismJsDemo;` twice — the latter has been wrong since long before this branch, so the snippet never compiled as printed.

### Validation

Compile-validated on Skia (`net10.0`): `Uno.UI`, `Uno.UI.Composition`, `Uno.UI.XamlHost`, `Uno.UI.RemoteControl`, `Uno.UI.Svg`, `Uno.UI.Lottie`, `Uno.Foundation`, `Uno.Foundation.Logging`, `Uno.WinRT`, plus the consumers most likely to notice a removed type — `Uno.UI.Toolkit.Skia`, `Uno.UI.FluentTheme.v2` and `Uno.UI.RuntimeTests.Skia`. All clean, no new warnings. `dotnet format whitespace` run over every touched file.

## PR Checklist ✅

- [x] 🧪 Added [Runtime tests, UI tests, or a manual test sample](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) (for bug fixes / features, if applicable) — n/a, no behaviour change; existing suites are the regression net
- [x] 📚 Docs have been added/updated following the [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features) — repaired the WASM interop snippet


MartinZikmund:
Added `refactor(sync-gen): Stop emitting Reference defines for Skia-only libraries` (428d6d671e).

The generator already separates the libraries that still ship more than the Skia flavor — `Uno.UWP`, `Uno.Foundation`, `Uno.UI.Dispatching` — from the Skia-only ones, and withholds the native defines from the latter. That set is *exactly* the set that kept a Reference head through the fold, so `__NETSTD_REFERENCE__` belongs on the same switch. For `Uno.UI` and `Uno.UI.Composition` the symbol can no longer be defined, so emitting it produces stubs guarded on a permanently-false condition and `[NotImplemented]` attributes naming a platform that doesn't exist.

The flag is renamed `CurrentTypeEmitsNativeDefines` → `CurrentTypeEmitsNonSkiaDefines`, because it now gates more than the native symbols and the old name is what invited this omission in the first place.

**The committed `Generated/` stubs are unchanged.** Two reasons:

- Running the sync tool needs the .NET 11 preview SDK this branch's pipeline pins (`11.0.100-preview.6`); locally it has to be coaxed with `-p:NetCurrent=net10.0`, which is exactly the kind of TFM override `LoadProject` warns breaks transitive project-reference resolution.
- Nothing in CI diffs `Generated/` — `RunAPISyncTool` exists only in `Uno.UI.Build.csproj` and is not wired into any pipeline. The only CI use of the generator is `GenerateDoc`, in the separate path-filtered docs pipeline.

So the regeneration is a separate deliberate run rather than part of this PR, and this change simply makes that run produce the right thing.

MartinZikmund:
### Sync generator: root cause found, and `Generated/` is regenerated

The `exit 127` failure had nothing to do with the project list or the Reference repoint — and it is now fixed rather than worked around.

**Root cause.** `MSBuildWorkspace` was created with `LoadMetadataForReferencedProjects = true`, so a referenced project resolved to its **compiled output whenever one happened to exist on disk**. `LoadProject` asserts its expected references are `CompilationReference`s, so the run died with

```
Uno.Foundation not found when loading '..\..\..\Uno.UWP\Uno.Reference.csproj (net10.0)'
```

The tell was that the *netcoremobile* heads loaded fine while the Reference head did not: their `bin` folders existed from restore but contained no DLLs, so Roslyn opened them as projects. Deleting the two Reference outputs moved the failure to the next project I happened to have built. So it fails on any tree that has been built, and CI never saw it because agents are clean and the sync-gen step only restores.

**Why the assertion is right, and the flag was wrong.** The generator resolves non-generated members to decide what to stub. A metadata reference exposes only the public surface, so an internal implementation reads as missing and gets a stub generated over it. Demanding compilations is correct; preferring metadata contradicted it. `LoadMetadataForReferencedProjects = false` makes the intent explicit instead of dependent on what's in `bin/`.

Verified as a controlled before/after: with outputs present — the case that previously failed — the generator now completes, and the regenerated tree is **byte-identical** to a run from a cleaned tree.

**Regeneration is included.** 1690 files: 1497 under `Uno.UI`, 193 under `Uno.UI.Composition`, all the same shape — `#if __SKIA__ || __NETSTD_REFERENCE__` → `#if __SKIA__`, with the paired `[NotImplemented]` dropping the platform that can no longer be defined. `Uno.UWP` (4516 files), `Uno.Foundation` (12) and `Uno.UI.Dispatching` (6) keep their Reference defines untouched, since those libraries still ship that flavor. `Uno.UI`, `Uno.UI.Composition` and `Uno.UI.RuntimeTests.Skia` all build clean against the regenerated stubs.

One incidental note for anyone running the tool: a failed run **deletes the `Generated/` folders before regenerating** and leaves them deleted, so `git checkout` them after a crash.

PR #24025, opened 2026-08-10
fix(hr): report suppressed source-generator errors on the operation

### LINKED ISSUE #24022: Hot reload: suppressing source-generator errors for the EnC emit also removes them from reported diagnostics (UXAML0001 replaced by ENC0020)
@dr1rrb — follow-up to #24012, which I confirmed fixes the case it targeted. This is about a side effect of the suppression.

## Summary

#24012 suppresses source-generator error diagnostics for the EnC emit so a broken input (e.g. a corrupt `.resw`) degrades to stale generated output instead of freezing hot reload. That works. But the suppression also removes those diagnostics from what the hot-reload operation **reports**, so when the file the user just edited is itself invalid, the actionable error is replaced by a misleading one.

## Observed

Editing `MainPage.xaml` into malformed XAML, then running a hot-reload cycle:

```
HR result: RudeEdit, Diagnostics: 1
DIAG: …/Uno.UI.SourceGenerators.XamlGenerator.XamlCodeGenerator/MainPage_eb076510….cs(156,4):
      error ENC0020: Renaming method 'IMainPage_Bindings.UpdateResources()' requires restarting the application.
```

The real error is still produced — it appears in the build trace:

```
…/Presentation/MainPage.xaml(6,2): error UXAML0001: Name cannot begin with the '<' character,
   hexadecimal value 0x3C. Line 6, position 2.
```

…but it never reaches the operation's diagnostics. The generator, given invalid XAML, emits degraded code that differs structurally enough to trip a rude edit, and **that** is what surfaces.

So a user who mistypes XAML is told *"Renaming method `IMainPage_Bindings.UpdateResources()` requires restarting the application"* — naming generated code they never wrote and never see — instead of *"line 6, position 2: name cannot begin with '<'"*.

The `Failed`/`RudeEdit` outcome itself is still correct; only the reported diagnostic regressed.

## Before / after

| case | broken file | in the change set? | before #24012 | after |
|---|---|---|---|---|
| corrupt `.resw`, edit an unrelated `.xaml` | `Resources.resw` | **no** | ❌ hot reload frozen | ✅ proceeds — the fix |
| edit `MainPage.xaml` invalidly | `MainPage.xaml` | **yes** | ✅ `UXAML0001` surfaced | ⚠️ replaced by `ENC0020` |

## Suggested refinement

Suppress the generator error's **blocking effect** while still **surfacing** it — or, if the blocking and reporting paths can't be separated cleanly, only suppress diagnostics whose source file is **outside** the change set. The change-set membership above is a clean discriminator between the two cases and preserves #24012's fix intact.

## Repro

1. Take a project with a `MainPage.xaml`.
2. Overwrite it with malformed XAML (e.g. an unclosed element).
3. Run a hot-reload cycle and inspect the operation's diagnostics.

Expected: a diagnostic identifying the XAML syntax error.
Actual: `ENC0020` naming a generated `IMainPage_Bindings` member.

## Impact

A downstream consumer has a runtime test asserting `UXAML0001` appears in the hot-reload operation's diagnostics; it fails deterministically on Uno SDK `6.8.0-dev.6` (3/3 CI attempts, Desktop and Wasm) and passes on `6.7.0-dev.938`. Beyond the test, the user-facing effect is that a XAML typo during hot reload no longer says what is wrong or where.

nickrandolph:
Fixes #24022

## Problem

#24012 suppresses source-generator errors for the EnC emit so a non-fatal generator error degrades to stale generated output instead of freezing hot reload — that part works. It reports them via `diagnosticsReporter.ReportAnalyzerDiagnostics(generatorErrors)`, which routes to `IReporter.Verbose`. But a few lines later:

```csharp
// Emit (EnC) diagnostics carry on every outcome of this branch; deltas are populated on Success only.
diagnostics = emitDiagnostics;
```

assigns outright, so the suppressed generator errors never reach the operation's `diagnostics` — the collection consumers actually read.

For a broken input **outside** the change set (the corrupt `.resw` case #24012 targets) that is harmless. For the file the user **just edited** it removes the only actionable message:

```
HR result: RudeEdit, Diagnostics: 1
DIAG: …/XamlCodeGenerator/MainPage_eb076510….cs(156,4):
      error ENC0020: Renaming method 'IMainPage_Bindings.UpdateResources()' requires restarting the application.
```

while the real error exists but is confined to the trace:

```
…/Presentation/MainPage.xaml(6,2): error UXAML0001: Name cannot begin with the '<' character,
   hexadecimal value 0x3C. Line 6, position 2.
```

So a user who mistypes XAML is told a generated `IMainPage_Bindings` member was renamed — code they never wrote and cannot see — instead of the line and column of their typo.

## Fix

Carry the suppressed generator errors alongside the emit diagnostics:

```csharp
diagnostics = emitDiagnostics.AddRange(generatorErrors);
```

The suppression still applies to the **emit** — hot reload keeps running, which is #24012's behaviour — but the diagnostics are no longer dropped from what is reported.

**No dedup required**, unlike the compilation-error `FIXME` a few lines below ("Needs dedup before fixing"): generator diagnostics never appear in `compilation.GetDiagnostics()`, so they cannot already be present in `emitDiagnostics`. The two sets are disjoint by construction.

**Scope is already correct**: `generatorErrors` comes from `GetGeneratorErrorsAsync(auditedProjects, …)`, already limited to the change set's projects (spec 054), so this does not surface errors from projects the pass never touched.

The `ENC` diagnostic is retained, not replaced — consumers get both the rude-edit reason and the generator error that caused it.

## Alternative considered

The issue suggested only suppressing diagnostics whose source file is **outside** the change set, which would separate the two cases at file granularity. I did not take that: it changes the *suppression* semantics #24012 deliberately chose, whereas this change only fixes the *reporting*, which is the actual defect. If you would rather have the finer-grained suppression, this PR does not conflict with it.

## Verification

- `dotnet build src/Uno.HotReload/Uno.HotReload.csproj -c Debug` — **succeeds, 0 warnings, 0 errors**.
- **Not** verified end-to-end against the failing consumer test: that needs a dev SDK carrying this change. The expected effect is that `operation.Diagnostics` contains `UXAML0001` **and** the `ENC` rude-edit diagnostic, where today it contains only the latter.

### DIFF
--- a/src/Uno.HotReload/HotReloadManager.cs
+++ b/src/Uno.HotReload/HotReloadManager.cs
@@ -256,7 +256,18 @@ public sealed class HotReloadManager : IDisposable
 			_tracker.Output($"Found {updates.Length} metadata updates after {sw.Elapsed}");
 
 			// Emit (EnC) diagnostics carry on every outcome of this branch; deltas are populated on Success only.
-			diagnostics = emitDiagnostics;
+			//
+			// The suppressed generator errors carry too. Suppressing them keeps hot reload running (that is
+			// the point), but they remain the reason the generated code is stale, and for the file the user
+			// just edited they are the ONLY actionable message: invalid XAML reports UXAML0001 here, while
+			// the emit sees a degraded generated tree and reports something like ENC0020 naming a generated
+			// member the user never wrote. Reporting them to the console alone left consumers with the
+			// misleading half.
+			//
+			// No dedup needed, unlike the compilation-error case below: generator diagnostics never appear in
+			// compilation.GetDiagnostics(), so they cannot already be present in emitDiagnostics. They are
+			// also already scoped to the change set's projects via auditedProjects.
+			diagnostics = emitDiagnostics.AddRange(generatorErrors);
 
 			switch (rudeEdits.IsEmpty, updates.IsEmpty)
 			{


src/Uno.HotReload/HotReloadManager.cs
```diff
@@ -256,7 +256,18 @@ private async ValueTask ProcessSolutionChanged(HotReloadOperation hotReload, Imm
...
+			// the point), but they remain the reason the generated code is stale, and for the file the user
+			// just edited they are the ONLY actionable message: invalid XAML reports UXAML0001 here, while
+			// the emit sees a degraded generated tree and reports something like ENC0020 naming a generated
+			// member the user never wrote. Reporting them to the console alone left consumers with the
+			// misleading half.
+			//
+			// No dedup needed, unlike the compilation-error case below: generator diagnostics never appear in
+			// compilation.GetDiagnostics(), so they cannot already be present in emitDiagnostics. They are
+			// also already scoped to the change set's projects via auditedProjects.
+			diagnostics = emitDiagnostics.AddRange(generatorErrors);
```
dr1rrb, line 270: The `generatorErrors` are already reported to the log / agent through the [`diagnosticsReporter.ReportAnalyzerDiagnostics(generatorErrors);`](https://github.com/unoplatform/uno/pull/24025/changes#diff-8a93897fb238d1609f48aa18d101e5c45ea292974a87ba091f584587c0a51701R245)

I kept the `diagnostics` to be only about the errors that prevented the emit to produce a delta. Not against the idea to include them, would juts like to know why you need to include them? (Considered also to include diagnostics from solution update and compilation, and include all diagnostics, not only errors)

dr1rrb (approved):

dr1rrb:

dr1rrb:
Approved: this is required for the UXAML error to be reported.

PR #24057, opened 2026-08-12
fix(ios): report UserCancel when web auth sheet is dismissed

### LINKED ISSUE #19666: iOS Web Authentication Broker causes InvalidOperationException if dialog is cancelled.
### Current behavior

When using OIDC to authenticate on iOS, the system will show a dialog asking if the uses want's to use X website to login.

If the user presses the cancel button then the WebAuthenticationBrokerProvider will throw a `InvalidOperationException` with the message `An attempt was made to transition a task to a final state when it had already completed.`

### Expected behavior

Cancelling this dialog should return an unauthenticated result without throwing an exception in the Main.iOS class.

- Add an OIDC configuration to the project config file
- Run the project with the iOS head
- Press the login button
- When the dialog appears, press cancel
- The InvalidOperationException will be thrown from the Main.iOS class

Minimal Repo
[BugReport.zip](https://github.com/user-attachments/files/19181349/BugReport.zip)

### Workaround

Add a try catch statement to the Main.iOS class.
This will prevent the app immediately crashing out but is a bit of a kludge fix.

### Works on UWP/WinUI

Yes

### Environment

Uno.UI.DevServer / Uno.WinUI.DevServer, Uno.WinUI / Uno.WinUI.WebAssembly / Uno.WinUI.Skia

### NuGet package version(s)

Packages installed using the SDK (version 5.6.70)

### Affected platforms

iOS, Mac Catalyst

### IDE

Rider macOS

### IDE version

2024.3.6

### Relevant plugins

_No response_

_No response_

ajpinedam:
**GitHub Issue:** closes #19666

## PR Type:

🐞 Bugfix

## What changed? 🚀

Dismissing the `ASWebAuthenticationSession` sheet on iOS crashed the app instead of returning a `UserCancel` result.

`WebAuthenticationBrokerProvider.AuthenticateAsyncCore` called `tcs.SetResult` twice in the `error != null` branch — the canceled-login case set `UserCancel` and then fell through to the `ErrorHttp` result, because there was no `return` between them. The second call threw `InvalidOperationException` ("An attempt was made to transition a task to a final state when it had already completed") inside the Objective-C completion-handler trampoline, where nothing can catch it, so it was marshalled to an `NSException` and the app died with `SIGABRT`. Every sign-in cancel was a guaranteed crash, and the caller never saw the `UserCancel` result it would have handled gracefully.

The callback now maps the native result through a single, mutually exclusive `CreateResult(callbackUrl, error)`:

- no error → `Success`
- canceled-login error (`ASWebAuthenticationSession`, or `SFAuthenticationSession` on iOS) → `UserCancel`
- any other error → `ErrorHttp`

It also completes the task with `TrySetResult` rather than `SetResult`. The session can invoke the callback more than once — a late completion can race the `Cancel()` in the `finally` — and throwing from that native trampoline is unrecoverable, so the callback must never be the thing that throws.

Validated on the iOS simulator (iPhone 16 Pro, iOS 18.4): cancelling the sheet now reports `UserCancel` and the app stays alive.

## PR Checklist ✅

Tests: `Given_WebAuthenticationBrokerProvider` pins the four result mappings (it compiles on the Apple UIKit heads, since it builds an `NSError` directly). A `Windows.Security` / `AuthenticationBroker_Cancel` manual sample drives the real sheet — the cancel path itself needs a human tap, so it can't be covered by an automated runtime test.

### DIFF
--- a/src/Uno.UWP/Security/Authentication/Web/WebAuthenticationBrokerProvider.iOS.cs
+++ b/src/Uno.UWP/Security/Authentication/Web/WebAuthenticationBrokerProvider.iOS.cs
@@ -30,36 +30,11 @@ namespace Uno.AuthenticationBroker
 		{
 			var tcs = new TaskCompletionSource<WebAuthenticationResult>();
 
+			// TrySetResult: the session can invoke the callback more than once (e.g. a late
+			// completion racing the Cancel() below), and throwing from this native
+			// trampoline is an unrecoverable NSException.
 			void AuthSessionCallback(NSUrl? callbackUrl, NSError? error)
-			{
-				if (error != null)
-				{
-					if ((error.Domain == asWebAuthenticationSessionErrorDomain &&
-						error.Code == asWebAuthenticationSessionErrorCodeCanceledLogin)
-#if __IOS__
-						|| (error.Domain == sfAuthenticationErrorDomain &&
-							error.Code == sfAuthenticationErrorCanceledLogin)
-#endif
-						)
-					{
-						tcs.SetResult(new WebAuthenticationResult(
-							null,
-							0,
-							WebAuthenticationStatus.UserCancel));
-					}
-					tcs.SetResult(new WebAuthenticationResult(
-						error.ToString(),
-						0,
-						WebAuthenticationStatus.ErrorHttp));
-				}
-				else
-				{
-					tcs.SetResult(new WebAuthenticationResult(
-						callbackUrl?.AbsoluteString,
-						0,
-						WebAuthenticationStatus.Success));
-				}
-			}
+				=> tcs.TrySetResult(CreateResult(callbackUrl, error));
 
 			IDisposable? was = default;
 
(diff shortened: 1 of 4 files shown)


src/Uno.UWP/Security/Authentication/Web/WebAuthenticationBrokerProvider.iOS.cs
```diff
@@ -30,36 +30,11 @@ protected virtual async Task<WebAuthenticationResult> AuthenticateAsyncCore(
 		{
 			var tcs = new TaskCompletionSource<WebAuthenticationResult>();
```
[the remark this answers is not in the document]
ajpinedam: Fixed in 712c18d143 — valid concern. The awaited continuation carries the `finally { Cancel(); }`, and since the callback arrives on the main thread with the same synchronization context that was captured at the await, it would have run inline on the trampoline and disposed the session from inside its own completion handler. The TCS now uses `TaskCreationOptions.RunContinuationsAsynchronously`.

src/Uno.UWP/Security/Authentication/Web/WebAuthenticationBrokerProvider.iOS.cs
```diff
@@ -168,6 +146,39 @@ void Start()
...
+			return new WebAuthenticationResult(
+				error.ToString(),
+				0,
+				WebAuthenticationStatus.ErrorHttp);
+		}
+
+		private static bool IsUserCanceledLogin(NSError error)
+			=> (error.Domain == asWebAuthenticationSessionErrorDomain &&
+				error.Code == asWebAuthenticationSessionErrorCodeCanceledLogin)
+#if __IOS__
```
nickrandolph, line 176: Why do we need this condition here? Isn't the whole file ios only?

kazo0 (approved):

src/Uno.UWP/Security/Authentication/Web/WebAuthenticationBrokerProvider.iOS.cs
ajpinedam: Indeed, that condition is not needed in this file, but all these were already there, this PR just moved a couple of them around.

We can, remove them all in a follow up PR, considering that change has little to do with the actual fix, or if you prefer I can remove them now. @nickrandolph

kazo0:

Issue #22688, opened 2026-02-19
Flaky native tests


MartinZikmund:
### What would you like to be added? 💡

-

### Renderer 🎨

- [ ] Skia
- [ ] Native

MartinZikmund:
No longer relevant

PR #24088, opened 2026-08-15
refactor!: Align the conditional compilation vocabulary with the target framework

### LINKED ISSUE #17684: `HAS_UNO_SKIA_PLATFORM` constants are misleading
### Current behavior

image

With single project, all `HAS_UNO_SKIA_` will be defined at the same time. The platforms can't actually be differentiated at compile-time.

### Expected behavior

_No response_

_No response_

### Workaround

_No response_

### Works on UWP/WinUI

None

### Environment

_No response_

### NuGet package version(s)

_No response_

### Affected platforms

_No response_

### IDE

_No response_

### IDE version

_No response_

### Relevant plugins

_No response_

_No response_

MartinZikmund:
**GitHub Issue:** closes #17684

## PR Type:

> Stacked on #24087. It targets `feature/breakingchanges` too, so the diff below includes that PR's
> commits — the ten commits from `a960ed99` onward are what is new here. Please review after #24087 merges.

## What changed? 🚀

Conditional compilation vocabulary, continuing where #24087 left the platform axis.

**XAML prefixes.** Every prefix is now named after a target framework. The ones that named a renderer, or a
distinction that no longer exists, are removed:

| Removed | Replacement |
|---|---|
| `skia:`, `netstdref:` | `not_winappsdk:` (`win:` / `not_win:` kept as synonyms of `winappsdk:` / `not_winappsdk:`) |
| `not_skia:`, `not_netstdref:` | `winappsdk:` |
| `androidskia:`, `iosskia:`, `tvosskia:`, `wasmskia:` | `android:`, `ios:`, `tvos:`, `wasm:` |
| `macos:` | `desktop:` |
| `not_mux:`, `xamarin:`, `legacy:` | dropped — none resolved to a real condition |

**Dead conditional markup is deleted rather than migrated.** Content behind `android:`, `ios:` and `wasm:` was
excluded from every project in this repository both before and after the change, since they all target a plain
`netX.0`. It targeted the native renderers removed in 7.0 and referenced types that no longer exist — 126
elements and 76 attributes across 62 files. Seven samples whose entire body was such markup are removed, along
with the `[ActivePlatforms(Platform.Android)]` UI test that drove one of them.

**File suffixes.** `*.Apple.cs` and `*.UIKit.cs` had byte-for-byte identical rules (`IsApple` and `IsAppleUIKit`
are both `IsIOS or IsTvOS`), so the 29 `*.Apple.cs` files are renamed and the suffix retires. `*.reference.cs` is
dropped from the SDK — it selects on `UnoRuntimeIdentifier`, which nothing outside `Uno.UWP` /
`Uno.Foundation` / `Uno.UI.Dispatching` sets, so for an application it compiled nowhere and warned nowhere; those
three projects keep it through the repository's own targets. `*.iOSmacOS.cs` named the native macOS target
removed in 7.0.

**Preprocessor symbols.** The eight `HAS_UNO_SKIA_<host>` symbols and their `__UNO_SKIA_<host>__` counterparts are
removed, closing #17684. They read as a compile-time host discriminator but cannot be one: the SDK references
every desktop host package together, so a `netX.0-desktop` head defined all of them at once. Nothing branched on
them. `HAS_UNO_SKIA` and `__UNO_SKIA__` are untouched — the XAML generator emits the former.

`UNO_REFERENCE_API` is now documented as what it actually is. Every `Uno.WinUI.Runtime.Skia.*` package defines it,
including the Android, Apple UIKit and WebAssembly ones, so it is not "no platform identifier" — it is the host
axis, the same condition as `HAS_UNO`. Documented as a synonym rather than given a third spelling.

**Two defects found on the way:**

- `BatchMergeXaml` stamped a fixed `mc:Ignorable="ios android wasm skia"` onto every merged theme dictionary —
  naming prefixes the merged file does not declare while omitting `not_win` and `not_winappsdk`, which it does. It
  now collects the ignorable set from the conditional namespaces actually present.
- `DatePicker` never forwarded `UseNativeMinMaxDates` to the native flyout: the assignment sat behind
  `#if SUPPORTS_NATIVE_DATEPICKER`, a symbol no target framework defines. `NativeDatePickerFlyout` is live on
  Skia-Android and reads that property when building the dialog, so it was silently ignored on the one target
  where it does anything. An internal `INativeDatePickerFlyout` lets `DatePicker` set it without naming a type
  from the Android runtime assembly.

Plus three conditionals that read as live and could never be true (`#if __ANDROID`, `#if WinUI`, `#if !DESKTOP`),
and the branches guarded by symbols that are defined nowhere — `SILVERLIGHT`, `METRO`, `WINPRT`,
`WINDOWS_PHONE`, `__XAMARIN__`, `XAMARIN_ANDROID`, `HAS_NATIVE_COMMANDBAR`, `IS_NATIVE_ELEMENT`. Two of those had
rotted: the `IS_NATIVE_ELEMENT` arm subscribes to handlers that exist nowhere in the tree, and `AppBar`'s
`_isNativeTemplate` could only ever be false, so its `MeasureOverride` early-return never ran.

## PR Checklist ✅

- [x] 🧪 Added [Runtime tests, UI tests, or a manual test sample](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) (for bug fixes / features, if applicable) — 4 tests for the `BatchMergeXaml` change; a Skia-Android runtime test for the `DatePicker` fix
- [x] 📚 Docs have been added/updated following the [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features) — `platform-specific-xaml.md`, `platform-specific-csharp.md`, `using-the-uno-sdk.md`, `migrating-to-uno-7.md`, `GLCanvasElement.md`, `progressring.md`
- [ ] 🖼️ Validated PR `Screenshots Compare Test Run` results — pending CI

## Breaking change impact and migration

Applications and libraries using the removed prefixes, suffixes or symbols must rename. All of it is mechanical
and all of it is listed in `doc/articles/migrating-to-uno-7.md`, which gains three tables: removed XAML prefixes,
removed file suffixes, and removed preprocessor symbols.

The one behavioural change beyond renaming is `DatePicker.UseNativeMinMaxDates`, which starts taking effect on
Skia-Android. An application that set it and silently got the default behaviour will now get the behaviour it
asked for.

## Validation

- `SamplesApp.Skia.Generic` (`net10.0-desktop`) builds clean, 0 errors / 0 warnings, after every commit.
- `Uno.UI.Tasks.Tests` 23/23.
- XamlStyler check passes on all 68 changed SamplesApp XAML files.
- Not validated locally: the `*.Apple.cs` → `*.UIKit.cs` rename only compiles on an Apple target framework, so
  the merged `SoftwareBitmap` partial was verified structurally (braces balanced, all 15 members preserved) but
  no compiler has seen it. The `DatePicker` runtime test needs a Skia-Android device and has not been run here.

https://claude.ai/code/session_01PTBMNbcPLwzdYrtPzY8s9Z


src/SamplesApp/SamplesApp.Samples/Windows/UI/Xaml/Controls/DatePicker/DatePickerFlyout_Unloaded.xaml
```diff
@@ -35,16 +33,7 @@
...
-                        Date="{x:Bind Date}"
-                        MinYear="{x:Bind Date}" />
-                    <android:NativeDatePickerFlyout
-                        x:Name="TestNativeDatePickerFlyout"
-                        Date="{x:Bind Date}"
-                        MinYear="{x:Bind Date}" />
-                </Button.Flyout>
+                <Button.Flyout />
             </Button>
             <TextBlock Text="{Binding Date, ElementName=TestNativeDatePickerFlyout}" />
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in 37e549be25 — you were right, and the row had a second problem the anchor didn't cover.

Besides the dangling `ElementName`, the `<Button.Flyout />` left behind was an empty property element, so `TestNativeDatePickerFlyoutButton` ("Open NativeDatePickerFlyout") opened nothing. Nothing in the row did anything any more, so the whole `StackPanel` is gone along with the matching `root.Children.Remove(TestNativeDatePickerFlyoutButton)` in the code-behind. The managed `DatePickerFlyout` row above still covers what the sample is for — a flyout being dismissed when its owner unloads.

I did not apply the suggestion verbatim: it replaces line 38 with `</StackPanel>` while line 39 already closes that panel, which would have left an unbalanced tag.

The same migration left the identical debris in two more files, fixed in the same commit:

- `DatePicker_Features.xaml` — `btnWithNativeFlyout` had the same empty `<Button.Flyout />`. Removed, and the heading above it (which named a flyout no longer in the sample) now names the `DatePicker` below, which is what the `UseNativeMinMaxDates` toggle still drives. This is the low-severity observation from the review summary.
- `ComboBox_FullScreen_Popup.xaml` — an empty `<UserControl.Resources />`.

I swept every XAML file the PR touches for the same pattern (empty property elements, and `ElementName`/`x:Bind` references to names the PR deleted); those three were the only real hits.

**Three UITests references left over, not touched here** — all in `SamplesApp.UITests`, which no CI lane runs on this branch (`TEST_KIND: ui` appears nowhere in `build/ci`, and `wasm-run-automated-uitests.sh` is not invoked), so none of them are failing anything today. Flagging rather than deleting, since removing enabled tests is a call for @MartinZikmund:

| Test | Enabled? | Missing element |
|---|---|---|
| `UnoSamples_Tests.ComboBoxTests.cs:154` `ComboBoxTests_Fullscreen_Popup_iOS` | yes, `[ActivePlatforms(Platform.iOS)]` | `Units1`, deleted with the `ios:ComboBox` |
| `TextBlockTests.cs:424,431,438` `When_Padding_Is_Changed_Then_Cache_Is_Missed` | yes, `[ActivePlatforms(Platform.Browser)]` | `misses`, deleted with the `wasm:TextBlock` (the `misses1/2/3` locals it feeds were already never asserted) |
| `UnoSamples_Tests.DatePicker.cs:239` `DatePickerFlyout_Native_Unloaded` | no, already `// [Test] -- Broken Test` | `TestNativeDatePickerFlyoutButton` |

Validation: **Compile** — `SamplesApp.Skia.Generic` (`net10.0-desktop`) builds with 0 errors. XamlStyler passes on all 1513 SamplesApp XAML files. Not run at runtime.

src/SamplesApp/SamplesApp.Samples/Windows/UI/Xaml/Controls/ComboBox/ComboBox_FullScreen_Popup.xaml
```diff
@@ -2,253 +2,16 @@
...
+    mc:Ignorable="d not_win">
 
     <StackPanel>
-        <ios:ComboBox
-            x:Name="Units1"
-            IsPopupFullscreen="true"
-            ItemsSource="{x:Bind MyItems}"
-            Style="{ThemeResource ComboStyle}" />
         <ComboBox
             x:Name="Values2"
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in a8cdaae670 — deleted the test rather than repointing it.

`Units1` was an `ios:ComboBox` carrying its own fullscreen-popup template (a `Picker` inside a bottom-anchored `PopupBorder`), and that whole `ios:` block is what this branch removed. The assertions were about that template's geometry, so there is no equivalent subject left in the sample. Repointing at `Values2` would have made it a near-duplicate of `ComboBoxTests_Fullscreen_Popup_Generic` — same sample, same element, only the dismiss-tap coordinate differs — and that test's own comment marks it flaky on iOS/Android native (#22688), so the iOS variant would have inherited the flakiness.

The same commit removes `DatePickerFlyout_Native_Unloaded`, the third stale reference from the earlier sweep: it drove `TestNativeDatePickerFlyoutButton`, removed with the native flyout row, and had already been commented out as `// [Test] -- Broken Test`.

Validation: **Compile** — `SamplesApp.UITests` builds with 0 errors. Not run at runtime; no CI lane on this branch invokes the `ui` test kind.

src/SamplesApp/SamplesApp.Samples/Windows/UI/Xaml/Controls/TextBlockControl/TextBlock_MeasureCache.xaml
```diff
@@ -5,9 +5,8 @@
     xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
     xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
     xmlns:muxc="using:Microsoft.UI.Xaml.Controls"
-    xmlns:wasm="http://uno.ui/wasm"
     Background="{ThemeResource ApplicationPageBackgroundThemeBrush}"
-    mc:Ignorable="d wasm">
+    mc:Ignorable="d">
 
     <StackPanel Spacing="6">
         <TextBlock>This test only works on WASM.</TextBlock>
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in a8cdaae670 — deleted the three `misses` reads and their locals, as suggested.

`misses1/2/3` were captured but never compared, and the surviving width assertions (`w1 > w2`, `w3 == w1`) do not depend on them, so the test keeps exactly the coverage it actually had.

Validation: **Compile** — `SamplesApp.UITests` builds with 0 errors. Not run at runtime; no CI lane on this branch invokes the `ui` test kind.

src/SourceGenerators/Uno.UI.SourceGenerators.Tests/XamlCodeGeneratorTests/Verifiers/CSGenerator.cs
```diff
@@ -130,15 +130,15 @@ protected override async Task RunImplAsync(CancellationToken cancellationToken)
...
+					excludeXamlNamespaces = "not_android,ios,tvos,desktop,wasm,winappsdk,win";
 				}
 				else if (ReferenceAssemblies.Packages.Any(p =>
 					p.Id.StartsWith("Microsoft.iOS.Ref", StringComparison.OrdinalIgnoreCase) ||
 					p.Id.StartsWith("Microsoft.tvOS.Ref", StringComparison.OrdinalIgnoreCase)))
 				{
-					includeXamlNamespaces = "ios,not_android,not_wasm,not_skia,not_netstdref";
-					excludeXamlNamespaces = "android,wasm,skia,not_ios";
+					includeXamlNamespaces = "not_android,ios,not_tvos,not_desktop,not_wasm,not_winappsdk,not_win";
+					excludeXamlNamespaces = "android,not_ios,tvos,desktop,wasm,winappsdk,win";
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in 417d2c3eac — split the branch so tvOS gets its own lists, and dropped `Microsoft.tvOS.Ref` from the iOS condition.

The lists mirror what `Uno.UI.SourceGenerators.props` derives from the target framework, where `_IsIOS` and `_IsTvOS` are independent `<Choose>` branches — so on tvOS `tvos` and `not_ios` are included while `ios` and `not_tvos` are excluded:

```csharp
includeXamlNamespaces = "not_android,not_ios,tvos,not_desktop,not_wasm,not_winappsdk,not_win";
excludeXamlNamespaces = "android,ios,not_tvos,desktop,wasm,winappsdk,win";
```

The iOS lists are unchanged, so no existing expectation moves.

Validation: **Compile** — `Uno.UI.SourceGenerators.Tests` builds with 0 errors, passing `-p:NetCurrent=net10.0` (this branch sets `NetCurrent` to `net11.0` and no .NET 11 SDK is installed on my machine). The suite itself needs a prior `Uno.UI` reference build I could not produce locally for the same reason, so the generator tests were **not** run here — CI covers them.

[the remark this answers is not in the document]
MartinZikmund: Upgrading the validation note above — I got the suite running locally after all.

The two prerequisites it needs are reference assemblies the verifier loads from disk: `Uno.UI.dll` and `Uno.UI.Extras.dll`. Building both with `-p:UnoTargetFrameworkOverride=net10.0` first, then running the suite with `-p:NetCurrent=net10.0`:

**Runtime — `Uno.UI.SourceGenerators.Tests`: 232 total, 0 failed, 0 skipped.**

So the branch split is confirmed green against the existing expectations, not just compiling. The tvOS branch itself stays unexercised until a test opts into `Microsoft.tvOS.Ref`, as you noted.

src/Uno.UI.FluentTheme/Resources/PriorityDefault/CalendarDatePicker_themeresources.xaml
```diff
@@ -1,6 +1,6 @@
 ﻿<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
 <!-- MUX Reference controls\dev\CommonStyles\CalendarDatePicker_themeresources.xaml, tag winui3/release/1.5.2 -->
-<ResourceDictionary xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation" xmlns:not_win="http://uno.ui/not_win" xmlns:win="http://schemas.microsoft.com/winfx/2006/xaml/presentation" xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" xmlns:wasm="http://uno.ui/wasm" xmlns:uc="using:Uno.UI.Controls" mc:ignorable="not_win wasm uc">
+<ResourceDictionary xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation" xmlns:not_win="http://uno.ui/not_win" xmlns:win="http://schemas.microsoft.com/winfx/2006/xaml/presentation" xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" xmlns:uc="using:Uno.UI.Controls" mc:ignorable="not_win wasm uc">
```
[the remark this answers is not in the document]
MartinZikmund: Good catch — confirmed as a regression from this branch, fixed in 6e9643c001.

`ec35f8239d` removed both the `xmlns:wasm` declaration and the `<wasm:Setter Property="CalendarViewStyle" …>` it guarded, but left `wasm` in the ignorable list. That is the same defect the `BatchMergeXaml` change in this PR fixes for merged output — the compiler being told to ignore a prefix that is not declared — so it is fitting that it also existed in a source file.

Two further leftovers from the same removal, cleaned up in the same commit:

- The `<!-- TODO Uno: Custom style for WASM to avoid broken animation -->` comment above the deleted setter was dangling; it now described an unrelated `Foreground` setter. The `CalendarViewCrippledAnimationStyle` it referred to no longer exists anywhere in the repo.
- `AppBar.xaml` carried the same mismatch — `mc:Ignorable="d netstdref not_netstdref"` with neither prefix declared. That one **predates this branch** (it is already so on `feature/breakingchanges`); `ec35f8239d` only removed the unrelated `not_mux` declaration from that file. Fixed here since it is the same defect class and this PR is about exactly this vocabulary hygiene — happy to split it out if you would rather keep the diff tight.

I swept every committed `.xaml` in the repo for the general pattern (ignorable naming a prefix the root element does not declare). Those two files were the only hits, and the sweep is clean now.

Validation: **Compile** — `Uno.UI.FluentTheme` and `Uno.UI` both build with 0 errors (`-p:UnoTargetFrameworkOverride=net10.0`).

src/Uno.UI/UI/Xaml/Controls/PullToRefresh/RefreshVisualizer/RefreshVisualizer_themeresources.xaml
```diff
@@ -2,21 +2,17 @@
...
-    xmlns:android="http://uno.ui/android"
     xmlns:not_android="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
-    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
-    mc:Ignorable="android">
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006">
 
     <ResourceDictionary.ThemeDictionaries>
         <ResourceDictionary x:Key="Default">
             <SolidColorBrush x:Key="RefreshVisualizerForeground" Color="White"/>
             <not_android:SolidColorBrush x:Key="RefreshVisualizerBackground" Color="Transparent"/>
```
MartinZikmund: not_android: is not relevant here anymore -> can be just <SolidColorBrush -> maybe more such instances in repo

MartinZikmund: Fixed in a25c3a8840 (framework) and 1610ff7b11 (samples) — swept the whole repo, and it turned up two distinct problems.

**`Uno.UI` / `Uno.UI.FluentTheme` — dead prefixes.** Both build for a plain `netX.0`, so every negated prefix there resolves to "always included". `not_android:` and `not_wasm:` say nothing, and reading them suggests a per-platform variant the single `Uno.UI` build cannot have. Dropped from 14 elements — this file ×2, `RefreshContainer.xaml`, `Generic.xaml` (`not_android:FontIcon` + `not_wasm:ObjectAnimationUsingKeyFrames` ×4) and `TabView.xaml` — plus 4 `xmlns` declarations nothing used (`ContentDialog.xaml` and `SystemResources.xaml` never had a usage at all), with `mc:Ignorable` kept in step. No behavior change.

**`SamplesApp` — the same prefixes are now wrong rather than merely dead.** The samples compile into the `-android`, `-ios` and `-browserwasm` heads, so these now genuinely exclude, where in 6.x they were suppressed everywhere and behaved as "always". Every site was written against a native control that no longer exists, so what they hide is markup the Skia rendering needs:

- `ComboBox_Picker`, `ComboBox_Popover` — `DropDownGlyph` gone on Android
- `ComboBox_Popover` — the entire `Popup`, and the visual state driving `PopupBorder`, gone on iOS
- `ComboBox_Disabled`, `ComboBox_ToggleDisabled` — placeholder and drop-down animations gone on WebAssembly
- `ComboBox_DropDownPlacement`, `Playground` — the drop-down padding and the XAML editor pane

Plus 17 declarations naming a prefix nothing used, and their `mc:Ignorable` entries.

Left alone deliberately: `win:` / `not_win:` (still meaningful — `SamplesApp.Windows`), the `System.Xaml.Tests` fixtures, and the docs/spec, which define the vocabulary.

Validation: `SamplesApp.Skia.Generic` builds clean (0 errors, no XAML diagnostics), XamlStyler run on the 23 samples. Desktop and WinAppSDK output is unchanged by construction — these prefixes already resolved to "included" there — so the UITest snapshot baselines should not move.

One thing I did *not* touch, for scope: `ContentDialog.xaml` also declares `xmlns:ios`, `xmlns:android` and `xmlns:wasm` that nothing uses. They're harmless (declaration and `mc:Ignorable` agree), but they're the same dangling-declaration pattern in positive form if you want them swept too.

doc/articles/platform-specific-xaml.md
```diff
@@ -101,67 +101,60 @@ Consider the following XAML which is using the Windows Community Toolkit's [Blur
...
+| `desktop` | `netX.0-desktop` | `http://uno.ui/desktop` | yes |
+| `wasm` | `netX.0-browserwasm` | `http://uno.ui/wasm` | yes |
+| `winappsdk` | `netX.0-windows10.x` (WinAppSDK) | `http://schemas.microsoft.com/winfx/2006/xaml/presentation` | no |
+
+Each has a negated counterpart — `not_android`, `not_ios`, `not_tvos`, `not_desktop`, `not_wasm`, `not_winappsdk` —
+which applies to every target framework the positive one does not. The negated prefixes use the default
+presentation namespace and do not need to be listed in `mc:Ignorable`, with the exception of `not_winappsdk`, which
+the WinAppSDK XAML compiler cannot resolve.
+
+`win` and `not_win` are the historical spelling of `winappsdk` and `not_winappsdk` and behave identically.
```
Xiaoy312: ~~curious, whats the reasoning for the rename & for not wanting to keep an alias?~~
edit: the wordings and the table below make it seems to suggested we retired `(not_)?win`, but they are both still being used
can we make that clear?

doc/articles/platform-specific-xaml.md
```diff
@@ -101,67 +101,60 @@ Consider the following XAML which is using the Windows Community Toolkit's [Blur
...
-| `netstdref`     | `http://uno.ui/netstdref`                                   | yes                    |
-| `not_netstdref` | `http://uno.ui/not_netstdref`                               | yes                    |
+> [!NOTE]
+> Before Uno Platform 7.0 the platform prefixes did not resolve this way: `android:`, `ios:` and `wasm:` were
+> suppressed on every target that rendered with Skia, which by 6.0 was all of them, and a library's
+> platform-specific assembly was replaced by its `netX.0` one when consumed by a mobile head. Markup written
+> against that behaviour — in particular `not_android:` or `not_ios:` used to mean "everywhere" — now resolves
+> differently.
 
 ### Specifying namespaces
```
Xiaoy312, line 164: do we support stuffs like?: // in case of collision
```
xmlns:not_winappsdkNamespace1=...
xmlns:not_winappsdkNamespace2=...
```

doc/articles/platform-specific-xaml.md
Xiaoy312: i also see a lot of sample app pages, are using a mixture of `win` + `not_winappsdk` which feels wrong when used together in the same xaml file
^can ignore this comment

doc/articles/platform-specific-csharp.md
Xiaoy312, line 55: `XAMARIN` is not documented here

Xiaoy312: also, my understanding was XAMARIN=ios|droid|tv-etc...
why is it defined in: src/Uno.WinRT/Uno.WinRT.{Skia,Wasm}.csproj

src/Uno.Foundation/Uno.Core.Extensions/Uno.Core.Extensions.Compatibility/Reflection/TypeExtensions.cs
```diff
@@ -34,7 +34,7 @@ public static bool Is(this Type instance, Type type)
 			return type.IsAssignableFrom(instance);
 		}
 
-#if WINAPPSDK || WINPRT || XAMARIN
+#if WINAPPSDK || XAMARIN
```
Xiaoy312: do we even need this guard?

Xiaoy312 (approved):

doc/articles/platform-specific-xaml.md
MartinZikmund: You are right that it read as a retirement notice, and it should not have. Clarified in b2264125c0.

`win` and `not_win` are live, first-class and unchanged by this PR — `Uno.UI.SourceGenerators.props` emits `winappsdk;win` and `not_winappsdk;not_win` together, so the two spellings resolve identically. Nothing here deprecates them and no migration applies. Both spellings are now in the prefix table rather than only the long form, and the trailing sentence is replaced by a note saying explicitly that they are not deprecated and that `win`/`not_win` remain the more common form in our own markup.

On the follow-up you said to ignore: I did not change any markup, but the point seemed worth capturing where someone would hit it, so the note also suggests picking one spelling per file — a file that uses `win:` alongside `not_winappsdk:` reads as though the two name different conditions when they are exact opposites. Happy to drop that sentence if you would rather the doc stay silent on it.

doc/articles/platform-specific-xaml.md
MartinZikmund: No — that form is not supported, and the doc did not say so. Documented in b2264125c0.

The prefix is matched by exact comparison. `XamlFileParser.IsIncluded` receives the xmlns prefix (`XamlXmlReader` calls it with `r.Prefix`) and tests `_includeXamlNamespaces.Contains(localName)` / `_excludeXamlNamespaces.Contains(localName)` against the list the SDK emits. `not_winappsdkNamespace1` is not in that list, so it falls through to the conditional-method parse, is not a valid conditional either, and comes back `Default` — i.e. treated as an ordinary namespace and always included. It would fail silently rather than error, which is the worst version of this.

The supported way to cover several CLR namespaces is one `#using:` segment on the namespace URI:

```xml
xmlns:not_winappsdk="http://uno.ui/not_winappsdk#using:My.Custom.Namespace1;My.Custom.Namespace2"
```

so one conditional prefix per platform per file, with the namespaces listed after `#using:`. The "Specifying namespaces" section now states the exact-match rule and calls out the decorated-prefix variant as not recognised.

doc/articles/platform-specific-csharp.md
MartinZikmund: Both halves of this were right, and the second one is the more interesting. Retired the symbol entirely in b8799a63ba.

**Why it was not in that table:** `XAMARIN` is repo-internal. `Uno.Sdk` never emits it, so no consumer app has ever seen it, and `platform-specific-csharp.md` documents the consumer-facing contract. It genuinely did not belong there — but that is an argument for the symbol not existing in this shape, not for leaving it undocumented.

**Why it was in the Skia and Wasm projects:** because it had stopped meaning a platform. `Uno.CrossTargetting` defined it for iOS, tvOS and Android as the name implies, and then eleven project files defined it again for Skia, WebAssembly, Reference and netcoremobile. In practice it had drifted to mean roughly "an assembly Uno builds", which is what `HAS_UNO` already says.

The decisive part: **none of `Uno.WinRT`, `Uno.UI.Dispatching`, `Uno.UI.Composition` or `Uno.UI.XamlHost` contains a single `#if XAMARIN`.** Every one of those definitions was inert — it only made the symbol read as though it meant something it did not. (Spec 058 had already noticed the `Uno.UI.Composition` one in isolation.)

Each real use now states its own condition:

| Site | Was | Now |
|---|---|---|
| `FuncMemoizeExtensions` ×2 | `XAMARIN` | `__ANDROID__ \|\| __APPLE_UIKIT__` — identical set |
| `EnumerableExtensions` private `SkipLast` | `!XAMARIN` | `NETSTANDARD2_0` — the actual reason; netstandard2.0 lacks `Enumerable.SkipLast`, and its one caller is `XamlFileParser` |
| `CommandBarExtensions` ×2, `AttachedProperty_CollectionProperty` | `XAMARIN` | branch deleted — no compilation could reach it |
| `GlobalAssemblyInfo` | `!XAMARIN` | block deleted — CAS attributes .NET ignores, in a file no project compiles |
| SamplesApp ×14 | `XAMARIN` | `__ANDROID__ \|\| __APPLE_UIKIT__` |

Two conditions change behaviour rather than preserve it, called out so they are not accidental:

- `Given_ItemsPresenter.Epsilon` documents an Android pixel-conversion tolerance that an undefined symbol had switched off everywhere, Skia-Android included. Collapsing it to `0.0f` would have cemented that, so it is `OperatingSystem.IsAndroid() ? 0.5f : 0.0f` now. Same shape as the `SUPPORTS_NATIVE_DATEPICKER` correction in spec 057.
- `SamplesApp.csproj` defined the symbol for its Android head alone while `SamplesApp.Skia.netcoremobile` defined it for all three mobile targets. Its iOS and tvOS heads now match Android.

Validation: **Compile** — `Uno.Foundation.Skia`, `Uno.WinRT.Skia`, `Uno.UI.Dispatching.Skia`, `Uno.UI.Composition`, `Uno.UI.Extras.Skia`, `Uno.UI.RuntimeTests.Skia`, `Uno.UI.SourceGenerators` and the `SamplesApp` desktop head all build with 0 errors. `XamlGenerationTests` fails locally, but identically with this change reverted (6 pre-existing `UXAML0001` in `GlobalStaticResources02.xaml` / `VSM.Test.xaml`), so it is unrelated.

src/Uno.Foundation/Uno.Core.Extensions/Uno.Core.Extensions.Compatibility/Reflection/TypeExtensions.cs
MartinZikmund: No — and it turned out the method did not need to exist either. Removed in b8799a63ba.

Tracing what the guard actually selected:

- `WINAPPSDK` is defined only by `SamplesApp.Windows`, `SamplesApp`, `Uno.UI.RuntimeTests.Windows` and `Uno.WinUI.Graphics3DGL` — none of which compile this file.
- `TypeExtensions.cs` compiles into `Uno.Foundation.{Reference,Skia,Wasm,netcoremobile}` and, through `SourceGeneratorHelpers.props`, into the generator projects at netstandard2.0. `XAMARIN` was defined in exactly one of those five: `Uno.Foundation.netcoremobile`.

So the guard reduced to "mobile only", and there it wrapped a method with **no callers anywhere in the repo** — the four `GetConstructors()` call sites all take a `Type` and bind to the BCL instance method, not this `TypeInfo` extension. Dropping just the guard would have exposed an unused extension on five compilations instead of one, so the method and the guard are both gone.

Worth flagging since it is shared code: this lives under `Uno.Core.Extensions.Compatibility`, a vendored copy of uno.core, so this is a small divergence from upstream. Say the word if you would rather carry it and I will restore it.

src/Uno.UI/UI/Xaml/FrameworkElement.EffectiveViewport.cs
```diff
@@ -54,13 +54,9 @@ public event TypedEventHandler<_This, EffectiveViewportChangedEventArgs> Effecti
...
 		void IFrameworkElement_EffectiveViewport.InitializeEffectiveViewport()
 		{
-#if IS_NATIVE_ELEMENT
-			Loaded += ReconfigureViewportPropagationOnLoad;
-			Unloaded += ReconfigureViewportPropagationOnUnload;
-#endif
 		}
 
-#if !IS_NATIVE_ELEMENT && !UNO_HAS_ENHANCED_LIFECYCLE && !__NETSTD_REFERENCE__ // We rely on Enter/Leave with enhanced lifecycle instead of Loaded/Unloaded.
+#if !UNO_HAS_ENHANCED_LIFECYCLE && !__NETSTD_REFERENCE__ // We rely on Enter/Leave with enhanced lifecycle instead of Loaded/Unloaded.
```
MartinZikmund: Is !UNO_HAS_ENHANCED_LIFECYCLE still a thing?

src/Uno.UI.Extras/CommandBarExtensions.cs
```diff
@@ -38,11 +38,7 @@ public static string GetSubtitle(this CommandBar commandBar)
 				"NavigationCommand",
 				typeof(AppBarButton),
 				typeof(CommandBarExtensions),
-#if XAMARIN
-				new FrameworkPropertyMetadata(null, FrameworkPropertyMetadataOptions.ValueInheritsDataContext)
```
MartinZikmund, line 42: I think non-WinUI targets should still use the FrameworkPropertyMetadata

src/Uno.Sdk/targets/Uno.SingleProject.WinAppSdk.targets
```diff
@@ -32,7 +32,6 @@
 	</PropertyGroup>
 
 	<PropertyGroup Condition="'$(WindowsPackageType)' == 'MSIX'">
-		<DefineConstants>$(DefineConstants);WINAPPSDK_PACKAGED</DefineConstants>
```
MartinZikmund, line 35: Why this change?

src/Uno.UI/UI/Xaml/FrameworkElement.EffectiveViewport.cs
MartinZikmund: No — not in `Uno.UI`. And this particular one was worse than merely dead, so it is gone in 6d9541da04.

`Uno.UI.csproj` pins `<UnoRuntimeIdentifier>Skia</UnoRuntimeIdentifier>`, and `Uno.CrossTargetting.targets` defines `UNO_HAS_ENHANCED_LIFECYCLE` for both Skia and WebAssembly. So inside `Uno.UI` the symbol is always defined and `!UNO_HAS_ENHANCED_LIFECYCLE` is always false — the §4 "feature flag that is now unconditional" category from spec 058.

What made this site worth deleting rather than just noting: `ReconfigureViewportPropagationPartial` appears **exactly once in the whole repository** — that implementation. There is no declaring `partial` and no caller. A partial implementation without a declaration does not compile, so had the branch ever been taken it would have failed the build. The declaration and call site were removed at some point and this was left behind; it is debris, not a fallback.

**The other four `!UNO_HAS_ENHANCED_LIFECYCLE` branches in `Uno.UI` I deliberately left alone**, because they are the opposite case — each is a coherent Loaded/Unloaded fallback that names itself in its own comment:

| Site | What it is |
|---|---|
| `MenuFlyoutSubItem.mux.cs:26` | "Uno specific fallback for platforms without enhanced lifecycle (Android/iOS native)" |
| `SplitMenuFlyoutItem.mux.cs:26` | same |
| `ViewManager.cs:822` | "Native Android/iOS only: ElementPrepared fires after OnApplyTemplate there" |
| `CalendarViewBaseItem.h.cs:32` | `EnterImpl()` from a `Loaded` handler — same fallback shape |

Those are native-renderer code, so they belong with the 7.0 native removal rather than with a vocabulary sweep — and spec 058 §3 is explicit that this category must not be swept mechanically. `CalendarViewBaseItem` also overlaps the CalendarView work in flight, so it is better not touched from here.

The `UNO_HAS_ENHANCED_LIFECYCLE` uses in `Uno.UI.RuntimeTests` are genuinely live either way, since `RuntimeTests.Windows` builds WinAppSDK where the symbol is absent.

Validation: **Compile** — `Uno.UI` builds with 0 errors.

src/SamplesApp/SamplesApp.Samples/Uno/Web/Http/CookieManagerTests.xaml
```diff
@@ -1,47 +0,0 @@
-﻿<Page
```
MartinZikmund, line 1: This sample should stay in existence, as the APIs are still relevant on Skia WASM

MartinZikmund: Maybe others such samples were also removed and should not have been! https://github.com/unoplatform/uno/pull/24046 will allow the "single unified" samples app to actually use the "wasm" etc XAML prefixes again for the Skia WASM etc targets, so this will no longer be a problem

src/Uno.UI.Extras/CommandBarExtensions.cs
MartinZikmund: Agreed, and that is a miss on my side — fixed in 27eb380ae6.

I had collapsed this to plain `PropertyMetadata` because the `FrameworkPropertyMetadata` branch sat behind `#if XAMARIN`, and `XAMARIN` is defined in neither `Uno.UI.Extras.Skia` nor `Uno.UI.Extras.Windows`. That reading was correct about the compiler and wrong about the intent: the branch had been dead since long before this PR, so what I preserved was the broken state rather than what the code was for.

`ValueInheritsDataContext` is exactly what these two attached properties want on Uno targets, so the metadata is now selected on `HAS_UNO` — the discriminator the rest of `Uno.UI.Extras` already uses (`ColorExtensions`, `Finger`, `AutomationPropertiesExtensions`), and one that is genuinely absent on the WinAppSDK build, which has no `FrameworkPropertyMetadata`:

```csharp
#if HAS_UNO
	new FrameworkPropertyMetadata(null, FrameworkPropertyMetadataOptions.ValueInheritsDataContext)
#else
	new PropertyMetadata(null)
#endif
```

Applied to both `NavigationCommandProperty` and `BackButtonForegroundProperty`. Note this is a behaviour restoration on Skia rather than a no-op — those properties will now inherit the data context where they did not before.

The three other `RegisterAttached` calls in the file were never inside the `XAMARIN` branch, so I left them on plain `PropertyMetadata`; say the word if they should follow too.

Validation: **Compile** — `Uno.UI.Extras.Skia` builds with 0 errors.

src/Uno.Sdk/targets/Uno.SingleProject.WinAppSdk.targets
MartinZikmund: Good catch — the change was wrong, and it is reverted in 27eb380ae6.

`a0a6a87978` dropped `WINAPPSDK_PACKAGED` on the strength of spec 057 listing it under "Uno-authored, never consumed". That verdict came from a repository-wide search, and for this symbol the search is the wrong instrument: it is defined in `src/Uno.Sdk/targets/Uno.SingleProject.WinAppSdk.targets`, so it ships to applications, and nothing in this repository consuming it says nothing about whether anything consumes it.

Separating MSIX from unpackaged has no other compile-time expression, so it is a plausible thing for an app to branch on — and removing the symbol would have sent those `#if` blocks to `#else` with no diagnostic at all. Silent, which is the failure mode worth avoiding.

It is also the only one of that group with this exposure: `IS_CI_OR_DEBUG`, `IS_UNO_UI_XamlHost_PROJECT` and `UNO_MIXIN_GENERATION` are all repository-internal, where the search does settle it. Those stay removed.

The define is back (with a comment marking it as public), the empty `PropertyGroup` the removal left behind is gone, and spec 057 carries a Correction noting that for anything `Uno.Sdk` emits the test is whether the symbol is public surface, not whether the repository greps clean.

src/SamplesApp/SamplesApp.Samples/Uno/Web/Http/CookieManagerTests.xaml
MartinZikmund: Restored in 961b4692af, and I swept the rest of the removals as you suggested — one more thing to restore, none.

`CookieManagerTests` went with the native-era markup because its entire body is a `wasm:StackPanel`, so the page read as empty. That was the wrong read: `Uno.Web.Http.CookieManager` is a browser capability, not native-renderer API, so it stays relevant wherever the app runs in a browser. Restored unchanged — its `[Sample]` registration is already behind `__WASM__`, so it only appears on the head that can exercise it, and once #24046 lands the `wasm:` prefix resolves for Skia WASM too.

**The sweep.** Seven samples were removed on this branch. Triaged by what the prefixed markup actually holds:

| Sample | Content | Verdict |
|---|---|---|
| `CookieManagerTests` | whole body is `wasm:`, browser API | **restored** |
| `ComboBox_NativePopup` | `android:` ×7, native popup | stays removed |
| `Picker_Resizable` | `ios:StackPanel` around `Picker`; its own description says "normally internal to iOS ComboBox" | stays removed |
| `CommandBar_Uppercase_Resource` | everything inside `android:Grid`, on the old `http://nventive.com/android` namespace | stays removed |
| `HorizontalListViewImage` / `HorizontalListViewTest` | `android:HorizontalListView` from `Uno.UI.Controls.Legacy`; the `ios:`/`win:`/`wasm:` siblings are placeholders reading "HorizontalListView is only used on Android" | stays removed |
| `UIElement_Native_Child` | `android:Border` wrapping a `NativeView` | stays removed |

So the distinction that mattered was `wasm:` marking a *browser* capability versus `android:`/`ios:` marking *native-renderer* content — only the first kind survives the renderer removal. The six above hold nothing else.

**One unrelated thing worth knowing:** `feature/breakingchanges` has moved since I rebased (`2b3cef7968` → `06190d3c30`, picking up the `avp-composition` merge among others). It briefly made `AnimatedVisualPlayerPlayground` look deleted by this branch in a base-to-head diff, when in fact the base gained it afterwards. Nothing to fix here, but the branch is a rebase behind.

Validation: **Compile** — `SamplesApp` desktop head builds with 0 errors. The `browserwasm` head I could not build locally (`wasm-tools` workload not installed), so CI is the check for the head this sample actually targets.

src/Uno.UI.Extras/CommandBarExtensions.cs
```diff
@@ -38,11 +38,7 @@ public static string GetSubtitle(this CommandBar commandBar)
...
-#else
 				new PropertyMetadata(null)
```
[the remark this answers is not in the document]
MartinZikmund: Agreed on the substance — this is the same issue @MartinZikmund raised on line 42, already fixed in 27eb380ae6. Both properties now carry `FrameworkPropertyMetadata` with `ValueInheritsDataContext` again, and your reading of the intent matches the fix.

One deliberate difference: the guard is `#if HAS_UNO`, not `#if !WINUI`.

The two behave identically for these two projects — `Uno.UI.Extras.Windows` defines `WINUI` and, being a `-windows10.0.19041.0` target, does not get `HAS_UNO`; `Uno.UI.Extras.Skia` is the reverse. But `WINUI` is not a reliable spelling of "not WinAppSDK": it is also defined by `MyCrossRuntimeLib.Skia.csproj` and `MyCrossRuntimeLib.Wasm.csproj` in the solution templates, where it means "WinUI API surface rather than UWP" — a different question. Selecting on it would mean this file says "not WinUI-the-API-flavour" while meaning "drawn by Uno".

`HAS_UNO` is defined as exactly that distinction, and it is what the rest of this assembly already uses — `ColorExtensions.cs`, `Finger.cs`, `IInjectedPointer.cs`, `AutomationPropertiesExtensions.cs`. Given this PR exists to stop conditional symbols meaning something other than what they say, swapping `XAMARIN` for a second ambiguous symbol would have repeated the original mistake.

Validation: **Compile** — `Uno.UI.Extras.Skia` builds with 0 errors.

PR #24091, opened 2026-08-15
fix(theming): focused NumberBox/ComboBox text is black in dark theme

### LINKED ISSUE #24021: [Skia] the input foreground of  NumberBox or ComboBox is black at dark theme
### Current behavior 🐛

Image

Image

Image

### Expected behavior 🎯

_No response_

```xml
<StackPanel HorizontalAlignment="Center"
            VerticalAlignment="Center"
            Spacing="12">
    <TextBlock HorizontalAlignment="Center"
               Text="Hello Uno Platform!"
               AutomationProperties.AutomationId="HelloTextBlock" />
    <TextBox Width="100"
             Header="TextBox" />
    <NumberBox Width="100"
               Header="NumberBox" />
    <ComboBox Width="100"
              IsEditable="True"
              Header="ComboBox" />
</StackPanel>
```

### Workaround 🛠️

_No response_

### Renderer 🎨

- [x] Skia
- [ ] Native

### Affected platforms 📱💻🖥️

All platforms 🌍

6.6.42

### IDE version 🧑‍💻

_No response_

_No response_

ajpinedam:
**GitHub Issue:** closes #24021

## PR Type:

🐞 Bugfix

## What changed? 🚀

On Skia, text typed into a focused `NumberBox` or editable `ComboBox` rendered black while the app was in dark theme, making it nearly unreadable.

Both templates carry (verbatim from current WinUI) a `Focused` visual state that sets `ContentElement.RequestedTheme="Light"` on the inner `TextBox`. In WinUI that is harmless; in Uno it exposed two distinct defects — though **not** one per control: the foreground-freeze defect affects **both** controls, while the setter-resolution defect affects **only the editable `ComboBox`**. `NumberBox` applies its focused foreground through `ObjectAnimationUsingKeyFrames`, whose keyframe value resolves under its own templated-parent theme and was already correct — so the freeze fix alone makes `NumberBox` pass, while `ComboBox` needs both fixes (confirmed by ablation of the hunks; thanks @MartinZikmund for measuring this).

### 1. Foreground freeze overrode an explicitly-set `Foreground` (both controls)

`FrameworkElement.NotifyThemeChangedForInheritedProperties(freeze: true)` froze the Light-theme `DefaultTextForegroundThemeBrush` (black) onto the boundary element and its children, even when that element already had a `Foreground` set locally, by style, or by animation.

WinUI (`framework.cpp`, `NotifyThemeChangedForInheritedProperties`) skips the *entire* freeze in that case — *"there is nothing to do, because that value will be used"*. Uno only skipped the `SetValue` while still recording `_themeForeground` / `_isForegroundFrozen`, which then blocked the explicit value from cascading to children through DP inheritance. On Skia the typed text is the `TextBoxView` `DisplayBlock`, which inherits `Foreground` from `ContentElement` — so it picked up the frozen black brush. This corrupts the child display block regardless of how the focused foreground was applied, which is why it hits both controls.

Fix: early-return when the foreground precedence is above `Inheritance`. The bail clears only the stored `_themeForeground` brush an earlier walk may have left (Uno keeps it in a separate field, while WinUI shares the local `Foreground` slot, so only Uno can leak it to children) and leaves `_isForegroundFrozen` untouched — WinUI's bail is a pure cleanup jump that writes no freeze state, so the boundary keeps blocking cross-boundary `Foreground` inheritance.

### 2. VisualState setter `{ThemeResource}` re-resolved under its target's own theme (editable ComboBox only)

A `{ThemeResource}` inside a `VisualState` `Setter` is registered on the **target** element in Uno. When the same state also applied `RequestedTheme="Light"` to that target, the resulting theme walk re-resolved `TextControlForegroundFocused` under Light and produced a black brush.

In WinUI the resource reference lives on the `Setter` DO, which sits in the template's `VisualStateGroups` — outside the target's subtree — so it never resolves under the target's `RequestedTheme`.

Fix: `DependencyObjectStore.GetVisualStateSetterResolutionTheme()` returns the ambient theme surrounding the target, applied at the three resolution sites (`UpdateThemeReference`, `InnerUpdateResourceBindingsUnsafe`, `ResourceResolver.ApplyVisualStateSetter`). It is deliberately scoped to boundaries applied by the state itself (`RequestedTheme` at `Animations` precedence); a XAML-authored `RequestedTheme` keeps Uno's existing element-level theming semantics for setter targets.

Mid-walk, an ancestor's persisted theme is stale (it is persisted only after its subtree completes), so the ambient is read from the parent's in-flight walk theme when the parent is still processing a theme walk.

When the parent has **no established per-object theme** — the state of every element in an app that never switches theme at runtime, since only a theme walk persists per-object themes — the ambient now falls back to the application base theme (`FrameworkTheming.GetBaseTheme()`). `ResolveOwnerTheme`'s owner-less fallback instead reads the requested-theme-for-subtree slot, which mid-boundary already carries the very Light theme being escaped, silently turning the escape into a no-op — so the fix did nothing in exactly the cold-start-dark scenario #24021 reports. Caught in review by @MartinZikmund. The fallback is `UNO_HAS_ENHANCED_LIFECYCLE`-only; native targets keep their existing behavior.

## Behavior change

Theme resolution changes in two observable ways, both toward WinUI parity:

- An element with an explicitly-set `Foreground` (local / style / animation) no longer has the theme's default foreground frozen over it — and no longer propagates that frozen default to its children — when it acts as a theme boundary.
- A `{ThemeResource}` in a `VisualState` `Setter` now resolves under the ambient theme rather than the target's own `RequestedTheme`, **only** when that `RequestedTheme` was applied by a visual state.

Existing app code that (intentionally or not) relied on the frozen-default behaviour could see a different foreground on such elements. No public API changed.

## Validation

- **Build:** `SamplesApp.Skia.Generic` (Release, net10.0) — 0 warnings, 0 errors.
- **Runtime tests (Skia Desktop):** `Given_ElementTheme` — 115 tests, 114 passed. The regression tests fail before their respective fixes and pass after:
  - `When_NumberBox_Focused_In_Dark_Theme_Text_Is_Light` (freeze fix)
  - `When_Editable_ComboBox_Focused_In_Dark_Theme_Text_Is_Light` (freeze + setter-resolution fixes)
  - `When_Editable_ComboBox_Focused_In_Dark_Theme_Without_Theme_Walk_Text_Is_Light` (base-theme fallback) — stages the never-theme-switched cold start (the exact #24021 scenario) by clearing the per-object themes the harness's theme switch persists; on a build without the fallback it fails with the focused text at `#E4000000` (black).
  - `When_Editable_ComboBox_Focused_In_Dark_Theme_ContentElement_Foreground_Is_Light` — platform-agnostic variant asserting the setter's direct target (`ContentElement.Foreground`, no Skia display-block dependency); it is also enabled for **native WASM** to cover the unguarded `ApplyVisualStateSetter` path there. Native runtime tests are excluded from PR CI by design, so it first executes there on the post-merge `master` branch build; `Uno.UI.RuntimeTests.Wasm` (and the changed `Uno.UI` sources under `__WASM__`) were compile-validated locally.
  - Every test carries a positive control asserting `BorderElement.Background` is the dark `TextControlBackgroundFocused` value (`#B31E1E1E`), proving the Focused state actually ran and resolved Dark — the Dark dictionary aliases all text-control foregrounds to the same light brush, so the foreground assert alone can't distinguish "resolved correctly" from "state never applied".
  - The single failure, `When_Flyout_Closed_Target_Does_Not_Hold_Flyout`, is a pre-existing GC/leak-sensitive test that fails identically on the unmodified branch — unrelated to this change.

### Not yet measured

- Native Android/iOS were not exercised beyond compilation (the class-level `PlatformCondition` already excludes them). Native WASM has direct coverage through the `ContentElement` test above, but native runtime tests only run on branch builds — first execution is master's post-merge build; `Screenshots Compare Test Run` results remain the broader check.
- No performance measurement was taken. The added work is an extra precedence check on the setter path and an early return on the freeze path, so no regression is expected, but it has not been benchmarked.

## PR Checklist ✅

- [ ] 📚 Docs have been added/updated following the [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features) — no documented behaviour changes; the fix aligns with existing WinUI-documented semantics
- [ ] 🖼️ Validated PR `Screenshots Compare Test Run` results. — to be validated once the CI run completes
- [ ] ❗ Contains **NO** breaking changes — see the *Behavior change* section above; theme resolution is observably different in two cases, both toward WinUI parity

### DIFF
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_ElementTheme.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_ElementTheme.cs
@@ -4589,6 +4589,76 @@ public class Given_ElementTheme
 #endif
 
 	#endregion
+
+	#region Focused text input in dark theme (issue #24021)
+
+#if HAS_UNO
+	[TestMethod]
+	[RequiresFullWindow]
+	[PlatformCondition(ConditionMode.Include, RuntimeTestPlatforms.Skia)]
+	[GitHubWorkItem("https://github.com/unoplatform/uno/issues/24021")]
+	public async Task When_NumberBox_Focused_In_Dark_Theme_Text_Is_Light()
+	{
+		// The NumberBox template's Focused state sets ContentElement.RequestedTheme = Light
+		// (as WinUI does). The theme boundary must not override the explicitly set focused
+		// Foreground that the text display block inherits.
+		using var _ = ThemeHelper.UseApplicationDarkTheme();
+
+		var numberBox = new NumberBox { Width = 150, Value = 888 };
+		WindowHelper.WindowContent = numberBox;
+		await WindowHelper.WaitForLoaded(numberBox);
+		await WindowHelper.WaitForIdle();
+
+		var inputBox = numberBox.FindFirstDescendant<TextBox>("InputBox");
+		Assert.IsNotNull(inputBox, "InputBox should exist in the NumberBox template");
+
+		inputBox.Focus(FocusState.Programmatic);
+		await WindowHelper.WaitForIdle();
+
+		var displayBlock = inputBox.FindFirstDescendant<TextBlock>(tb => tb.Text == "888");
+		Assert.IsNotNull(displayBlock, "The text display block should exist in the TextBox visual tree");
+
+		var foreground = displayBlock.Foreground as SolidColorBrush;
+		Assert.IsNotNull(foreground, "The display block Foreground should be a SolidColorBrush");
+		Assert.IsTrue((foreground.Color.R + foreground.Color.G + foreground.Color.B) / 3 > 200,
+			$"Focused NumberBox text should be light in dark theme, but was {foreground.Color}");
+	}
+
+	[TestMethod]
+	[RequiresFullWindow]
+	[PlatformCondition(ConditionMode.Include, RuntimeTestPlatforms.Skia)]
+	[GitHubWorkItem("https://github.com/unoplatform/uno/issues/24021")]
+	public async Task When_Editable_ComboBox_Focused_In_Dark_Theme_Text_Is_Light()
+	{
+		// Same as the NumberBox scenario: the editable ComboBox's TextBox style sets
+		// ContentElement.RequestedTheme = Light in its Focused state.
+		using var _ = ThemeHelper.UseApplicationDarkTheme();
+
+		var comboBox = new ComboBox { Width = 150, IsEditable = true };
+		WindowHelper.WindowContent = comboBox;
+		await WindowHelper.WaitForLoaded(comboBox);
+		await WindowHelper.WaitForIdle();
+
+		var editableText = comboBox.FindFirstDescendant<TextBox>("EditableText");
+		Assert.IsNotNull(editableText, "EditableText should exist in the ComboBox template");
+
+		editableText.Focus(FocusState.Programmatic);
+		await WindowHelper.WaitForIdle();
+
+		editableText.Text = "888";
+		await WindowHelper.WaitForIdle();
+
+		var displayBlock = editableText.FindFirstDescendant<TextBlock>(tb => tb.Text == "888");
+		Assert.IsNotNull(displayBlock, "The text display block should exist in the TextBox visual tree");
+
+		var foreground = displayBlock.Foreground as SolidColorBrush;
+		Assert.IsNotNull(foreground, "The display block Foreground should be a SolidColorBrush");
+		Assert.IsTrue((foreground.Color.R + foreground.Color.G + foreground.Color.B) / 3 > 200,
+			$"Focused editable ComboBox text should be light in dark theme, but was {foreground.Color}");
+	}
+#endif
+
+	#endregion
 }
 
 public partial class Issue475ThemedDO : DependencyObject
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_ElementTheme.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_ElementTheme.cs
@@ -4615,6 +4615,8 @@ public class Given_ElementTheme
 		inputBox.Focus(FocusState.Programmatic);
 		await WindowHelper.WaitForIdle();
 
+		AssertFocusedStateResolvedDark(inputBox);
+
 		var displayBlock = inputBox.FindFirstDescendant<TextBlock>(tb => tb.Text == "888");
 		Assert.IsNotNull(displayBlock, "The text display block should exist in the TextBox visual tree");
 
@@ -4647,6 +4649,8 @@ public class Given_ElementTheme
 
 		editableText.Text = "888";
 		await WindowHelper.WaitForIdle();
+
+		AssertFocusedStateResolvedDark(editableText);
 
 		var displayBlock = editableText.FindFirstDescendant<TextBlock>(tb => tb.Text == "888");
 		Assert.IsNotNull(displayBlock, "The text display block should exist in the TextBox visual tree");
@@ -4685,6 +4689,8 @@ public class Given_ElementTheme
 		editableText.Text = "888";
 		await WindowHelper.WaitForIdle();
 
+		AssertFocusedStateResolvedDark(editableText);
+
 		var displayBlock = editableText.FindFirstDescendant<TextBlock>(tb => tb.Text == "888");
 		Assert.IsNotNull(displayBlock, "The text display block should exist in the TextBox visual tree");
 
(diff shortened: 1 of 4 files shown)


MartinZikmund:
Reviewed this against the WinUI C++ sources and against a native WinUI oracle app, and ran an ablation of the two hunks on Skia Desktop. The diagnosis is right and the fix works on the tested path — but I think there's one defect worth fixing before merge, plus a couple of smaller notes.

## The core premise checks out

I built a small WinUI 3 app (WinAppSDK 2.1.3) to measure what WinUI actually resolves, since the claim ("the boundary applies but the `{ThemeResource}` still resolves Dark") is surprising enough to be worth measuring rather than reasoning about. Dark page, control focused:

| Control | `ContentElement.ActualTheme` | `ContentElement.Foreground` | Focused border bg |
|---|---|---|---|
| TextBox | Dark | `#FFFFFFFF` | `#B31E1E1E` |
| NumberBox | **Light** | **`#FFFFFFFF`** | `#B31E1E1E` |
| ComboBox (editable) | **Light** | **`#FFFFFFFF`** | `#B31E1E1E` |

So WinUI does apply the `RequestedTheme="Light"` boundary and still resolves the setter's `{ThemeResource}` to the Dark value. With this PR applied, Uno Platform reproduces all nine of those values exactly.

## Blocking: the ComboBox half is a no-op when the app never switches theme

`GetVisualStateSetterResolutionTheme()` ends in `ThemeResolution.ResolveOwnerTheme(Parent)`. When the parent's stored theme is `Theme.None`, `ThemeResolution.cs` falls back to `ResourceDictionary.GetActiveBaseTheme()`, which reads the core requested-theme-for-subtree slot — the same slot the boundary walk just loaded with Light. It returns the theme it's trying to escape. Silently: no log, no assert.

And `Theme.None` is the default for the whole tree in any app that never performs a runtime theme switch. I probed a freshly loaded tree with no theme switch:

```
inner=None mid=None root=None | ContentPresenter=None | Grid=None | WindowChrome=None | XamlIslandRoot=None
```

`_theme` is assigned in exactly one place (`DependencyObjectStore.Theming.cs`, inside the theme walk); `EstablishThemeOnEnterCore` only propagates a parent theme that is *already* non-`None`.

So on an app that simply starts in dark — which is what #24021 reports — the ComboBox half of this fix does nothing. The tests here can't see it because `ThemeHelper.UseApplicationDarkTheme` performs a real application theme switch, whose walk seeds `_theme` tree-wide; they only ever exercise the safe non-`None` path.

Suggested fix: when the parent's stored theme is `None`, fall back to the application base theme (`CoreServices.Instance.Theming.GetBaseTheme()`) rather than the ambient slot, and add a regression test that never switches the application theme.

Caveat on my own evidence: I verified each link of that chain individually (the probe above, plus reading the fallback). I did not stage a full cold-start dark app end-to-end — the runtime-test harness pins its frame to Light, which is exactly what the new test has to work around.

## The two defects aren't one-per-control

Ablation on Skia Desktop, each configuration built and run:

| Configuration | NumberBox | ComboBox |
|---|---|---|
| baseline | FAIL | FAIL |
| + freeze hunk only | **PASS** | FAIL |
| + setter hunks only | FAIL | FAIL |
| + both | PASS | PASS |

The freeze defect hits **both** controls (it corrupts the child display block, not `ContentElement` itself). The setter-resolution defect hits **only** ComboBox — NumberBox uses `ObjectAnimationUsingKeyFrames`, whose keyframe object resolves under its own templated-parent theme and is already correct. That's why ComboBox needs both hunks and the setter hunk alone fixes nothing. Worth correcting in the description, since it changes what each hunk is load-bearing for.

## Smaller notes

**The assertions can pass for the wrong reason.** `(R+G+B)/3 > 200` can't distinguish "focused state resolved correctly" from "focused state never applied at all", because in the Dark dictionary `TextControlForeground`, `...PointerOver` and `...Focused` all alias `TextFillColorPrimaryBrush`. A positive control fixes it — asserting `BorderElement.Background` is `#B31E1E1E` proves the Focused state actually ran before you assert anything about the foreground.

**`_isForegroundFrozen = false` on the bail.** WinUI's bail is a pure `goto Cleanup` that writes no freeze state at all. Clearing `_themeForeground` is a justified divergence (WinUI's freeze brush and the local `Foreground` share one storage slot, so a stale freeze can't leak there, whereas Uno Platform's separate field can). Clearing the flag as well also drops the theme-boundary inheritance block, which WinUI never does.

**Platform condition.** `RuntimeTestPlatforms.Skia` already includes `SkiaWasm`, so that's covered — but `ResourceResolver.ApplyVisualStateSetter` isn't preprocessor-guarded, so the changed code also runs on the native WASM target, untested.

## On the precedence gate

`GetCurrentHighestValuePrecedence(RequestedThemeProperty) == Animations` has no WinUI counterpart — there's no `BaseValueSource` check anywhere in WinUI's `RequestedTheme` path. WinUI gets the right answer purely on the setter side: a VSM setter's `{ThemeResource}` becomes a live binding on the `CSetter` itself, resolved under the setter's own theme, and the target receives an already-resolved value. The gate is a proxy that happens to coincide for these two controls.

That's a follow-up, not a merge blocker — this PR strictly narrows a pre-existing divergence rather than widening it. FWIW I've since pushed a setter-side port on `dev/mazi/notifyanims` (#23491) that matches the oracle without any precedence heuristic, so if this lands first the gate can just be deleted when that merges. Happy either way — landing this now fixes users sooner.

Xiaoy312 (approved):

ajpinedam:
Thanks @MartinZikmund — great catch on the blocker, and the ablation was spot on. All addressed:

- **Blocker** (44f0bf3702): with no established parent theme, `GetVisualStateSetterResolutionTheme` now falls back to `FrameworkTheming.GetBaseTheme()` instead of the ambient slot. 28323378b4 adds the regression test staging the never-theme-switched cold start; without the fix it fails with the focused text at `#E4000000`.
- **Smaller notes** (80add0a113, a3c256b1ee): the bail now clears only `_themeForeground` and keeps the freeze flag; every test carries the `#B31E1E1E` BorderElement positive control; and a platform-agnostic `ContentElement.Foreground` variant is enabled for native WASM (native runtime tests only run on branch builds, so it first executes post-merge — the Wasm head compiles clean locally).
- PR description corrected: freeze defect = both controls, setter defect = editable ComboBox only.

Agreed on the precedence gate — happy to delete it once #23491 lands.

morning4coffe-dev (approved):
The foreground freeze now follows WinUI's [explicit/local/style/animated-value bailout](https://github.com/microsoft/microsoft-ui-xaml/blob/d0c6705ec46dbddb853cdfdb2eb3b42849cf304e/dxaml/xcp/core/core/elements/framework.cpp#L3526-L3537), instead of retaining a competing theme-default brush for descendants. Resolving the state setter outside the RequestedTheme boundary introduced by that same state is scoped to animation-precedence boundaries, preserving authored element-theme behavior. The focused-state background assertion is a useful positive control beyond merely observing light text. Source approval only; the older DependencyObjectStore-based head still needs its base conflicts integrated.

morning4coffe-dev (approved):
Rechecked the conflict-resolution follow-up against the previously approved patch. The substantive adaptation is to master's DependencyObjectStore-to-DependencyObject consolidation: the parent walk-theme and established-theme checks now use the parent DependencyObject directly, and ResourceResolver calls the helper on the target directly. ActualInstance is now the object itself, so these preserve the prior ambient-theme fallback and foreground-precedence behavior. The focused NumberBox/ComboBox regression cases are unchanged. Approval renewed for this head.

PR #24092, opened 2026-08-16
fix(resources): resolve localized strings by culture script

### LINKED ISSUE #24024: [Language] Got a `zh-TW` string when set language to `zh-CN`
### Current behavior 🐛

```csharp
Windows.Globalization.ApplicationLanguages.PrimaryLanguageOverride = "zh-CN";
```
will see the tips on `TabView` or `NavigationView`:

Image

Image

### Expected behavior 🎯

_No response_

_No response_

### Workaround 🛠️

_No response_

### Renderer 🎨

- [x] Skia
- [x] Native

### Affected platforms 📱💻🖥️

All platforms 🌍

6.5.31

### IDE version 🧑‍💻

_No response_

_No response_

ajpinedam:
**GitHub Issue:** closes #24024

## PR Type:

🐞 Bugfix

## What changed? 🚀

Setting `ApplicationLanguages.PrimaryLanguageOverride = "zh-CN"` produced **Traditional** Chinese strings in built-in controls (`TabView`, `NavigationView`, ...).

### Current behavior

`ResourceLoader.FindForCulture` resolved in two steps: exact culture match, then — on a miss — sibling cultures sharing the base language, ordered **reverse-alphabetically**. Script (`Hans`/`Hant`) was never part of the decision, so among `zh-*` siblings the reverse-alphabetical winner was always the Traditional one.

On the version in the report (Uno.Sdk 6.5.31 → Uno.WinUI 6.5.153) the control strings ship under `zh-Hans` / `zh-Hant` folders. A request for `zh-CN` therefore had no exact match and fell through to `zh-Hant`:

```
cultureKeys = [en, en-GB, en-us, zh-Hans, zh-Hant]     <- no zh-CN
TabViewCloseButtonTooltipWithKA = '關閉索引標籤 (Ctrl+F4)'
SettingsButtonName              = '設定'
```

`1ff113c647` later renamed those folders to `zh-CN` / `zh-TW` (present in `release/stable/6.6`, `6.7` and `master`, not in `6.5`). That makes the reported `zh-CN` case work by accident while moving the same defect onto `zh-Hans`, `zh-Hans-CN`, `zh`, and `zh-SG` — so the rename is not a fix, and the fallback itself needs correcting. This is also why the report does not reproduce on `master` as written.

### New behavior

`FindForCulture` now resolves in two steps:

1. **Exact culture** — unchanged.
2. **One ranked candidate list.** The candidates are the culture's ancestors (`zh-CN` → `zh-Hans` → `zh`) **union** its siblings under the same base language — the script subtag only ever appears on the ancestor chain, which is what lets a region-named culture reach script-named resources and vice versa. They are ordered:

   > **same script → same region → closest ancestor → base culture → ordinal**

   Script first means `zh-Hant` can never answer a Simplified Chinese request, whether it is reached as a *sibling* or as a bare `zh` folder holding Traditional content. Region next follows the Windows matching order (language+script+region → language+region → language+script → language). Below that, closest-ancestor-then-base-culture-then-reverse-alphabetical preserves the previous behavior, so script-less languages are unaffected.

   A wrong-script candidate is **demoted, never dropped**: with `zh` as the only resource culture, a `zh-TW` request still resolves to `zh`.

Supporting changes: script detection from the four-letter subtag, resolving a script-less neutral culture (`zh`) through `CultureInfo.CreateSpecificCulture` rather than a hardcoded language table, cached in a `ConcurrentDictionary`; the final sibling tiebreak now uses `StringComparer.Ordinal` (it previously used the culture-sensitive default comparer, so fallback order shifted with `CurrentCulture`); unresolvable culture names are guarded and degrade to the previous behavior.

### Impact

Old loader vs new loader swept over the same resources, 111 culture tags × 4 resource keys = 444 lookups:

| | |
|---|---|
| identical | 425 |
| changed | 19 |
| resolved value → empty | **0** |

Every change is wrong-script → right-script:

- `zh`, `zh-Hans`, `zh-Hans-CN`, `zh-SG` — Traditional → Simplified
- `sr-Cyrl` — Latin → Cyrillic (same defect, surfaced by the sweep: `Zatvori karticu` → `Затвори картицу`)

The candidate set is unchanged — it is exactly the union of the two tiers the previous code walked, so the ranking only decides which of them answers first. The change can therefore only affect *which* sibling wins; it cannot make a resolvable string unresolvable, and apps shipping only one Chinese script are unaffected (verified: every `zh-TW` / `zh-Hant` / `zh-HK` / `zh-MO` lookup is unchanged).

Two notes for reviewers:

- A nearer ancestor now beats the bare base culture. An app shipping both `zh` and `zh-Hans` of its own will see `zh-CN` resolve to `zh-Hans` where it previously resolved to `zh` (more specific wins, matching .NET/MRT).
- This also changes *which* sibling wins when an app ships both region- and script-named Traditional folders. Ordinal ordering alone lets every `zh-Hant*` key outrank `zh-HK`, which loses the exact region match for a `zh-Hant-HK` request; the region tiebreak restores it. `zh-MO` against a `zh-HK` + `zh-Hant-TW` set still resolves to `zh-Hant-TW` by the ordinal tiebreak — both are Traditional, and which fallback Macau should prefer is unaudited, so it is left as-is. Uno ships no `zh-HK` folder, so none of this touches the built-in resources.
- Globalization-invariant mode is unaffected: it already fails earlier in `ApplicationLanguages.CreateCulture` when a non-invariant `PrimaryLanguageOverride` is set, identically before and after this change; the new code guards every `CultureInfo` call and is not reached.

### Validation

- **Unit tests** — 28 new `[DataRow]` cases in `Given_ResourceLoader`. 20 cover both folder conventions (`zh-Hans`/`zh-Hant` and `zh-CN`/`zh-TW`) so a future rename cannot silently reintroduce this — against the un-fixed build 8 fail, including `zh-CN` on script-named folders (the reported case). The other 8 are `When_Chinese_And_MixedFolders`, pinning the ranking against a mixed set (a bare `zh` holding Simplified, next to `zh-HK` and `zh-Hant-TW`); 4 of them fail before the ranking change. All 28 pass after.
- **Fallback sweep** — old vs new resolution over all 889 installed culture names against Uno's own 86 resource folders: **0 diffs**. The ranking only bites on layouts Uno itself doesn't ship (a bare `zh` folder, or `zh-HK` alongside `zh-Hant-*`).
- **Full `Uno.UI.UnitTests` suite** — 4082 passing. The 13 failures are timezone-sensitive `Given_Calendar` / `DateTimeOffset` tests, identical on a clean baseline run of the same commit range.
- **Ordering equivalence** — the ranking is applied by an explicit sort rather than a chain of `OrderBy`/`ThenBy` (see Allocations below). Verified by comparing the *whole ordered candidate list*, not just the winner, across 13 resource-culture layouts × 893 requested cultures = 11,609 lists: 0 differences.
- **Allocations** — bytes per `GetString`, measured with `GC.GetAllocatedBytesForCurrentThread()` over 200k calls:

  | path | master | this PR |
  |---|---|---|
  | exact hit (`zh-Hans` → `zh-Hans`) | 480 | 456 |
  | script miss (`zh-CN` → `zh-Hans`) | 1328 | 1040 |
  | mixed set (`zh-Hant-HK` → `zh-HK`) | 1366 | 1077 |
  | typical app (`en-US` → `en`) | 1272 | 936 |

  The fallback path is the default path for an app — the templates ship `Strings/en/Resources.resw`, so an `en-US` run misses the exact key on every `x:Uid` lookup. Subtags are cached per resource culture (build-time-fixed keys) and memoized for the requested culture under a 32-entry cap, safe to cache because a name's script and region are pure functions of the name.
- **Analyzers** — `Uno.UWP/Uno.Skia.csproj` rebuilt with `UnoFastDevBuild=false`: 0 warnings, 0 errors.
- **End-to-end against the reported version** — the fixed loader driven against the actual shipped `Uno.WinUI 6.5.153` `Uno.UI.dll` returns `关闭选项卡(Ctrl+F4)` and `设置` for `zh-CN`.

This should be backported to `release/stable/6.5`, `6.6` and `6.7` — 6.5 is where the issue was reported, and on 6.6/6.7 the folder rename hides that case while the same defect still affects `zh-Hans`.

Separate pre-existing issue, not addressed here: strings are resolved once when a control materializes and are never re-resolved if `PrimaryLanguageOverride` changes afterwards.

## PR Checklist ✅

### DIFF
--- a/src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.instanced.cs
+++ b/src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.instanced.cs
@@ -26,9 +26,11 @@ partial class ResourceLoader
 {
 	private readonly Dictionary<string, Dictionary<string, string>> _resources = new(StringComparer.OrdinalIgnoreCase); // _resources[CULTURE][RES_KEY] => RES_VALUE
 
-	// Concurrent because, unlike the per-loader state, this is shared process-wide and GetString
-	// carries no thread affinity.
-	private static readonly ConcurrentDictionary<string, string?> _scripts = new(StringComparer.OrdinalIgnoreCase); // _scripts[CULTURE] => ISO 15924 script
+	// Keyed only by resource cultures, which are fixed at build time — never by the requested
+	// culture, which comes from PrimaryLanguageOverride and is caller-controlled. Concurrent
+	// because, unlike the per-loader state, this is shared process-wide and GetString carries no
+	// thread affinity.
+	private static readonly ConcurrentDictionary<string, string?> _resourceScripts = new(StringComparer.OrdinalIgnoreCase); // _resourceScripts[CULTURE] => ISO 15924 script
 
 	internal string LoaderName { get; }
 
@@ -126,10 +128,12 @@ partial class ResourceLoader
 		// for a Simplified Chinese culture just because both are written "zh-".
 		if (culture.Split('-', 2)[0] is { Length: > 0 } baseCulture)
 		{
-			var script = GetScript(culture);
+			// Resolved rather than cached: the requested culture is caller-controlled, so caching it
+			// would let a process-wide dictionary grow without bound.
+			var script = ResolveScript(culture);
 			var relatedCultures = _resources.Keys
 				.Where(x => x.StartsWith(baseCulture, StringComparison.OrdinalIgnoreCase))
-				.OrderByDescending(x => script is not null && GetScript(x) == script) // same script first
+				.OrderByDescending(x => script is not null && GetResourceScript(x) == script) // same script first
 				.ThenByDescending(x => string.Equals(x, baseCulture, StringComparison.OrdinalIgnoreCase)) // then base culture
 				.ThenByDescending(x => x, StringComparer.Ordinal); // and then, sibling cultures in reverse order (from ex-ZZ to ex-AA)
 			foreach (var related in relatedCultures)
@@ -181,11 +185,12 @@ partial class ResourceLoader
 	}
 
 	/// <summary>
-	/// Gets the ISO 15924 script of a culture (Hans, Hant, Latn, Cyrl, ...), or null when the
-	/// platform doesn't associate one with it.
+	/// Gets the ISO 15924 script of a culture holding resources (Hans, Hant, Latn, Cyrl, ...), or
+	/// null when the platform doesn't associate one with it. Cached: the resource cultures are
+	/// fixed at build time, and each is scored repeatedly while ordering the fallback candidates.
 	/// </summary>
-	private static string? GetScript(string culture)
-		=> _scripts.GetOrAdd(culture, static x => ResolveScript(x));
+	private static string? GetResourceScript(string culture)
+		=> _resourceScripts.GetOrAdd(culture, static x => ResolveScript(x));
 
 	private static string? ResolveScript(string culture)
 	{
--- a/src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.instanced.cs
+++ b/src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.instanced.cs
@@ -26,11 +26,17 @@ partial class ResourceLoader
 {
 	private readonly Dictionary<string, Dictionary<string, string>> _resources = new(StringComparer.OrdinalIgnoreCase); // _resources[CULTURE][RES_KEY] => RES_VALUE
 
-	// Keyed only by resource cultures, which are fixed at build time — never by the requested
-	// culture, which comes from PrimaryLanguageOverride and is caller-controlled. Concurrent
-	// because, unlike the per-loader state, this is shared process-wide and GetString carries no
-	// thread affinity.
-	private static readonly ConcurrentDictionary<string, string?> _resourceScripts = new(StringComparer.OrdinalIgnoreCase); // _resourceScripts[CULTURE] => ISO 15924 script
+	// Keyed only by resource cultures, which are fixed at build time. Concurrent because, unlike the
+	// per-loader state, this is shared process-wide and GetString carries no thread affinity.
+	private static readonly ConcurrentDictionary<string, (string? Script, string? Region)> _resourceSubtags = new(StringComparer.OrdinalIgnoreCase);
+
+	// The requested culture comes from PrimaryLanguageOverride and is caller-controlled, so it gets a
+	// separate memo that is dropped wholesale instead of growing. Entries can never go stale: a name's
+	// script and region are pure functions of the name.
+	private static readonly ConcurrentDictionary<string, (string? Script, string? Region)> _requestedSubtags = new(StringComparer.OrdinalIgnoreCase);
+
+	// Comfortably above the handful of language preferences a lookup cycles through.
+	private const int MaxRequestedSubtags = 32;
 
 	internal string LoaderName { get; }
 
@@ -132,28 +138,79 @@ partial class ResourceLoader
 	/// the closest match down. Script comes first, so that zh-Hant never answers a Simplified Chinese
 	/// request, whether it is reached as a sibling or as a bare zh folder holding Traditional content.
 	/// </summary>
-	private IEnumerable<string> GetFallbackCultures(string culture)
+	private List<string> GetFallbackCultures(string culture)
 	{
 		var baseCulture = culture.Split('-', 2)[0];
+		var ancestors = GetAncestors(culture).ToArray();
+		var (script, region) = GetRequestedSubtags(culture);
 
-		var ancestors = new Dictionary<string, int>(StringComparer.OrdinalIgnoreCase);
-		foreach (var ancestor in GetAncestors(culture))
+		var candidates = new List<string>();
+		foreach (var resourceCulture in _resources.Keys)
 		{
-			ancestors.TryAdd(ancestor, ancestors.Count);
+			if (AncestorDepth(resourceCulture) >= 0 ||
+				(baseCulture.Length > 0 && resourceCulture.StartsWith(baseCulture, StringComparison.OrdinalIgnoreCase)))
+			{
+				candidates.Add(resourceCulture);
+			}
 		}
 
-		// Resolved rather than cached: the requested culture is caller-controlled, so caching it
-		// would let a process-wide dictionary grow without bound.
-		var script = ResolveScript(culture);
-		var region = GetRegionSubtag(culture);
+		candidates.Sort(CompareCandidates);
 
-		return _resources.Keys
-			.Where(x => ancestors.ContainsKey(x) || (baseCulture.Length > 0 && x.StartsWith(baseCulture, StringComparison.OrdinalIgnoreCase)))
-			.OrderByDescending(x => script is not null && string.Equals(GetResourceScript(x), script, StringComparison.OrdinalIgnoreCase)) // same script first
-			.ThenByDescending(x => region is not null && string.Equals(GetRegionSubtag(x), region, StringComparison.OrdinalIgnoreCase)) // then same region (Windows matches language+region ahead of language+script)
-			.ThenBy(x => ancestors.TryGetValue(x, out var depth) ? depth : int.MaxValue) // then the closest ancestor
-			.ThenByDescending(x => string.Equals(x, baseCulture, StringComparison.OrdinalIgnoreCase)) // then the base culture, the only ancestor an unresolvable culture still reaches
-			.ThenByDescending(x => x, StringComparer.Ordinal); // and finally the remaining siblings, from ex-ZZ to ex-AA
+		return candidates;
+
+		int CompareCandidates(string left, string right)
+		{
+			var byScript = SameScript(right).CompareTo(SameScript(left)); // same script first
+			if (byScript != 0)
+			{
+				return byScript;
+			}
+
+			var byRegion = SameRegion(right).CompareTo(SameRegion(left)); // then same region
+			if (byRegion != 0)
+			{
+				return byRegion;
+			}
+
+			var byAncestor = Proximity(left).CompareTo(Proximity(right)); // then the closest ancestor
+			if (byAncestor != 0)
+			{
+				return byAncestor;
+			}
+
+			var byBaseCulture = IsBaseCulture(right).CompareTo(IsBaseCulture(left)); // then the base culture
+			if (byBaseCulture != 0)
+			{
+				return byBaseCulture;
+			}
+
+			return StringComparer.Ordinal.Compare(right, left); // and finally from ex-ZZ to ex-AA
+		}
+
+		bool SameScript(string candidate)
+			=> script is not null && string.Equals(GetResourceSubtags(candidate).Script, script, StringComparison.OrdinalIgnoreCase);
+
+		bool SameRegion(string candidate)
+			=> region is not null && string.Equals(GetResourceSubtags(candidate).Region, region, StringComparison.OrdinalIgnoreCase);
+
+		int Proximity(string candidate)
+			=> AncestorDepth(candidate) is var depth && depth >= 0 ? depth : int.MaxValue;
+
+		int AncestorDepth(string candidate)
+		{
+			for (var i = 0; i < ancestors.Length; i++)
+			{
+				if (string.Equals(ancestors[i], candidate, StringComparison.OrdinalIgnoreCase))
+				{
+					return i;
+				}
+			}
+
+			return -1;
+		}
+
+		bool IsBaseCulture(string candidate)
+			=> string.Equals(candidate, baseCulture, StringComparison.OrdinalIgnoreCase);
 	}
 
 	private bool TryGetForCulture(string culture, string resource, out string? resourceValue)
@@ -194,12 +251,35 @@ partial class ResourceLoader
 	}
 
 	/// <summary>
-	/// Gets the ISO 15924 script of a culture holding resources (Hans, Hant, Latn, Cyrl, ...), or
-	/// null when the platform doesn't associate one with it. Cached: the resource cultures are
+	/// Gets the ISO 15924 script (Hans, Hant, Latn, Cyrl, ...) and the region of a culture holding
+	/// resources, either being null when the culture carries none. Cached: the resource cultures are
 	/// fixed at build time, and each is scored repeatedly while ordering the fallback candidates.
 	/// </summary>
-	private static string? GetResourceScript(string culture)
-		=> _resourceScripts.GetOrAdd(culture, static x => ResolveScript(x));
+	private static (string? Script, string? Region) GetResourceSubtags(string culture)
+		=> _resourceSubtags.GetOrAdd(culture, static x => (ResolveScript(x), GetRegionSubtag(x)));
+
+	/// <summary>
+	/// Same for the requested culture, memoized under <see cref="MaxRequestedSubtags"/> entries
+	/// because its name is caller-controlled.
+	/// </summary>
+	private static (string? Script, string? Region) GetRequestedSubtags(string culture)
+	{
+		if (_requestedSubtags.TryGetValue(culture, out var subtags))
+		{
+			return subtags;
+		}
+
+		subtags = (ResolveScript(culture), GetRegionSubtag(culture));
+
+		if (_requestedSubtags.Count >= MaxRequestedSubtags)
+		{
+			_requestedSubtags.Clear();
+		}
+
+		_requestedSubtags[culture] = subtags;
+
+		return subtags;
+	}
 
 	private static string? ResolveScript(string culture)
 	{
--- a/src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.instanced.cs
+++ b/src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.instanced.cs
@@ -114,39 +114,46 @@ partial class ResourceLoader
 			return true;
 		}
 
-		// Then the culture's ancestors (zh-CN -> zh-Hans -> zh). The script tag only ever appears
-		// there, so a region-named culture can reach its script-named resources and vice versa.
-		foreach (var ancestor in GetAncestors(culture))
+		foreach (var candidate in GetFallbackCultures(culture))
 		{
-			if (TryGetForCulture(ancestor, resource, out resourceValue))
+			if (TryGetForCulture(candidate, resource, out resourceValue))
 			{
 				return true;
 			}
 		}
 
-		// Finally sibling cultures, same-script ones first: zh-Hant must never answer a request
-		// for a Simplified Chinese culture just because both are written "zh-".
-		if (culture.Split('-', 2)[0] is { Length: > 0 } baseCulture)
-		{
-			// Resolved rather than cached: the requested culture is caller-controlled, so caching it
-			// would let a process-wide dictionary grow without bound.
-			var script = ResolveScript(culture);
-			var relatedCultures = _resources.Keys
-				.Where(x => x.StartsWith(baseCulture, StringComparison.OrdinalIgnoreCase))
-				.OrderByDescending(x => script is not null && GetResourceScript(x) == script) // same script first
-				.ThenByDescending(x => string.Equals(x, baseCulture, StringComparison.OrdinalIgnoreCase)) // then base culture
-				.ThenByDescending(x => x, StringComparer.Ordinal); // and then, sibling cultures in reverse order (from ex-ZZ to ex-AA)
-			foreach (var related in relatedCultures)
-			{
-				if (TryGetForCulture(related, resource, out resourceValue))
-				{
-					return true;
-				}
-			}
-		}
-
 		resourceValue = null;
 		return false;
+	}
+
+	/// <summary>
+	/// Orders the resource cultures that may stand in for <paramref name="culture"/> — its ancestors
+	/// (zh-CN -> zh-Hans -> zh) and the siblings sharing its language (zh-Hant-TW for zh-TW) — from
+	/// the closest match down. Script comes first, so that zh-Hant never answers a Simplified Chinese
+	/// request, whether it is reached as a sibling or as a bare zh folder holding Traditional content.
+	/// </summary>
+	private IEnumerable<string> GetFallbackCultures(string culture)
+	{
+		var baseCulture = culture.Split('-', 2)[0];
+
+		var ancestors = new Dictionary<string, int>(StringComparer.OrdinalIgnoreCase);
+		foreach (var ancestor in GetAncestors(culture))
+		{
+			ancestors.TryAdd(ancestor, ancestors.Count);
+		}
+
+		// Resolved rather than cached: the requested culture is caller-controlled, so caching it
+		// would let a process-wide dictionary grow without bound.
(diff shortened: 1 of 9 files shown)


ajpinedam:
Good catch, and it is reachable — but the suggested remedy wouldn't have bounded it, so I fixed it from the other side.

I measured it first. Setting 500 distinct junk values on `PrimaryLanguageOverride` and doing a lookup after each:

```
GetCultureInfo("zz-Bogus")    -> CultureNotFoundException
GetCultureInfo("qqq-Zzzz-QQ") -> resolved as 'qqq-Zzzz-QQ'
GetCultureInfo("not a culture") -> CultureNotFoundException
GetCultureInfo("xx")          -> resolved as 'xx'
500 distinct junk PLO values, rejected by the setter: 0
_scripts.Count = 500
```

So the growth is real. But skipping the ones `CultureInfo` can't resolve would have caught almost none of them: under ICU, most syntactically-valid junk (`qqq-Zzzz-QQ`, `xx`, `zh-x0001`) resolves fine as a custom culture and would still have been cached. None of the 500 was rejected.

The bound has to come from the key space instead. The cache now only ever holds **resource** cultures — the `_resources` keys, fixed at build time — which is the side that actually benefits from caching, since each key is scored repeatedly while ordering the fallback candidates. The requested culture is resolved directly and never cached, because that is the caller-controlled one:

```csharp
// Resolved rather than cached: the requested culture is caller-controlled, so caching it
// would let a process-wide dictionary grow without bound.
var script = ResolveScript(culture);
```

Renamed to `_resourceScripts` / `GetResourceScript` so the invariant is visible at the use site.

Verification after the change:
- same 500-junk-tag run: `_resourceScripts.Count = 0`
- behaviour unchanged: the 111-culture × 4-key sweep is byte-identical to the previous commit (0 diff lines of 444)
- `Uno.UI.UnitTests`: 4074 passing, unchanged
- analyzer-enabled `Uno.Skia.csproj` rebuild: 0 warnings, 0 errors

Pushed as b16fd532fd.

src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.instanced.cs
```diff
@@ -128,4 +144,151 @@ private bool FindForCulture(string culture, string resource, out string? resourc
...
+			}
+		}
+	}
+
+	/// <summary>
+	/// Gets the ISO 15924 script of a culture (Hans, Hant, Latn, Cyrl, ...), or null when the
+	/// platform doesn't associate one with it.
+	/// </summary>
+	private static string? GetScript(string culture)
+		=> _scripts.GetOrAdd(culture, static x => ResolveScript(x));
```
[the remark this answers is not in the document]
ajpinedam: fixed in https://github.com/unoplatform/uno/pull/24092/commits/b16fd532fddf44f64e351f7e67663913c86d9988

src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.instanced.cs
```diff
@@ -128,4 +148,152 @@ private bool FindForCulture(string culture, string resource, out string? resourc
...
+		}
+
+		return null;
+	}
+
+	/// <summary>
+	/// Extracts the script subtag of a culture name, i.e. a four-letter subtag that is not the
+	/// language itself (zh-Hant-TW -> Hant, ca-ES-VALENCIA -> null).
+	/// </summary>
+	private static string? GetScriptSubtag(string cultureName)
```
Xiaoy312, line 313: much easier to read as a linq oneliner:
```suggestion
	private static string? GetScriptSubtag(string cultureName) =>
		cultureName.Split('-')
			.Skip(1) // index-0 is never a script -- a 4ALPHA there is a (reserved) language subtag
			.FirstOrDefault(shard => shard.Length is 4 && shard.All(char.IsLetter));
```

src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.instanced.cs
```diff
@@ -128,4 +148,152 @@ private bool FindForCulture(string culture, string resource, out string? resourc
...
+			if (GetScriptSubtag(name) is { } script)
+			{
+				return script;
+			}
+
+			if (string.Equals(current.Parent.Name, name, StringComparison.Ordinal))
+			{
+				break;
+			}
+		}
```
Xiaoy312: this is largely the GetAncestors construct from above, might want to refactor that in

src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.instanced.cs
```diff
@@ -102,23 +109,36 @@ private bool FindForCulture(string culture, string resource, out string? resourc
...
-		// if we cant find using the specific culture, fallback on base culture, and then sibling cultures
+		// Then the culture's ancestors (zh-CN -> zh-Hans -> zh). The script tag only ever appears
+		// there, so a region-named culture can reach its script-named resources and vice versa.
+		foreach (var ancestor in GetAncestors(culture))
+		{
+			if (TryGetForCulture(ancestor, resource, out resourceValue))
+			{
+				return true;
+			}
+		}
```
Xiaoy312: This tier is unconditionally preferred over the same-script sibling tier below, and the bare base language is always an ancestor — so a `Strings/zh/` or `Strings/sr/` folder re-opens the exact defect this PR fixes. `ResolveScript` already knows the answer here and the walk ignores it: tier 3 refuses a wrong-script *sibling*, while tier 2 accepts a wrong-script *ancestor*.

Measured by transcribing these helpers verbatim into a standalone harness and diffing old vs. new resolution (.NET 10, ICU):

| resource keys | request | resolves to | script(request) | script(result) |
|---|---|---|---|---|
| `sr`, `sr-Cyrl-BA` | `sr-Cyrl-RS` | **`sr`** | Cyrl | **Latn** ❌ |
| `sr`, `sr-Cyrl-BA`, `sr-Latn-RS` | `sr-Cyrl-RS` | **`sr`** | Cyrl | **Latn** ❌ |
| `zh`, `zh-Hant-TW` | `zh-TW` | **`zh`** | Hant | **Hans** ❌ |
| `zh`, `zh-Hant-TW` | `zh-HK` | **`zh`** | Hant | **Hans** ❌ |
| `az`, `az-Cyrl-AZ` | `az-Cyrl` | **`az`** | Cyrl | **Latn** ❌ |

The first row is the `sr-Cyrl` case from the PR description (`Zatvori karticu` → `Затвори картицу`) — it passes today only because Uno's own resources happen to carry no bare `sr` folder. The `zh` rows are the same story: the *key* `zh-Hant` can no longer answer a Simplified request, but a bare `zh` folder holding Traditional content still can, so the outcome is unchanged for an app laid out that way.

Smallest fix that closes it — hoist `ResolveScript(culture)` above this loop and skip an ancestor whose script is known and differs:

```csharp
		// Then the culture's ancestors (zh-CN -> zh-Hans -> zh). The script tag only ever appears
		// there, so a region-named culture can reach its script-named resources and vice versa.
		// Resolved rather than cached: the requested culture is caller-controlled, so caching it
		// would let a process-wide dictionary grow without bound.
		var script = ResolveScript(culture);
		foreach (var ancestor in GetAncestors(culture))
		{
			if (script is not null && GetResourceScript(ancestor) is { } ancestorScript && ancestorScript != script)
			{
				// A nearer ancestor in the wrong writing system is worse than a same-script
				// sibling: let the tier below answer instead.
				continue;
			}

			if (TryGetForCulture(ancestor, resource, out resourceValue))
			{
				return true;
			}
		}
```

(The `var script = ResolveScript(culture);` further down then goes away, along with its comment, which moves up here.)

I ran that against all 20 of the new `[DataRow]` cases, the five rows above, and `en` / `fi` / `ko` / `fr` / `pt` fallback guards: every new test still passes, every row above is fixed, and no guard moves.

src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.instanced.cs
```diff
@@ -102,23 +109,36 @@ private bool FindForCulture(string culture, string resource, out string? resourc
...
+			// Resolved rather than cached: the requested culture is caller-controlled, so caching it
+			// would let a process-wide dictionary grow without bound.
+			var script = ResolveScript(culture);
 			var relatedCultures = _resources.Keys
 				.Where(x => x.StartsWith(baseCulture, StringComparison.OrdinalIgnoreCase))
-				.OrderByDescending(x => x == baseCulture) // base culture first
-				.ThenByDescending(x => x); // and then, sibling cultures in reverse order (from ex-ZZ to ex-AA)
+				.OrderByDescending(x => script is not null && GetResourceScript(x) == script) // same script first
+				.ThenByDescending(x => string.Equals(x, baseCulture, StringComparison.OrdinalIgnoreCase)) // then base culture
+				.ThenByDescending(x => x, StringComparer.Ordinal); // and then, sibling cultures in reverse order (from ex-ZZ to ex-AA)
```
Xiaoy312: Worth calling out in the description that this isn't only a determinism fix — it also changes *which* sibling wins. Ordinal-descending compares index 4, `'K'` (0x4B) vs `'a'` (0x61), so every `zh-Hant*` key now outranks `zh-HK`:

| resource keys | request | before | after |
|---|---|---|---|
| `zh-HK`, `zh-Hant-TW` | `zh-Hant-HK` | `zh-HK` | **`zh-Hant-TW`** ← exact region match lost |
| `zh-HK`, `zh-Hant-TW` | `zh-MO` | `zh-HK` | `zh-Hant-TW` |
| `zh-HK`, `zh-Hant-TW` | `zh-TW` | `zh-HK` | `zh-Hant-TW` (better) |

Row 1 is the load-bearing one: `zh-HK` is an exact region match for `zh-Hant-HK` and it loses to Taiwan. Row 2 I'd call an unaudited change rather than a regression — both results are Traditional, and I haven't verified which fallback is preferred for Macau. The 444-lookup sweep in the description can't see any of this, since Uno ships no `zh-HK` folder.

The underlying gap is that this tier has no notion of region proximity at all (pre-existing), so the winner is an accident of string sort either way — the switch to `Ordinal` just changes which accident.

If you'd rather close this and the ancestor issue in one move, merging tiers 2 and 3 into a single ranked candidate list works: candidates = matching ancestors ∪ prefix-siblings, ordered by same-script → same-region → nearest-ancestor → base-culture → ordinal. Verified against the same harness — passes all 20 new `[DataRow]`s, fixes every ancestor row from the other comment, and restores `zh-Hant-HK` → `zh-HK`.

At minimum, a test pinning the intended order for a mixed `zh-HK` + `zh-Hant-*` resource set would keep this from drifting again.

Xiaoy312 (approved):
Approving — the diagnosis is right and the three-tier shape is the correct fix. Two findings from transcribing the new helpers into a standalone harness and diffing old vs. new resolution on .NET 10 / ICU, left inline:

1. **The ancestor tier takes a known-wrong-script ancestor**, and the bare base language is always an ancestor — so a `Strings/zh/` or `Strings/sr/` folder re-opens the defect, including the `sr-Cyrl` case from the description. Small fix, verified non-breaking against all 20 new `[DataRow]`s.
2. **The `StringComparer.Ordinal` switch changes outcomes, not just determinism** — `zh-Hant-HK` now loses its exact-region `zh-HK` match to `zh-Hant-TW`. Worth a note in the description and a test either way.

Neither undoes the improvement this delivers over today's reverse-alphabetical behavior, but (1) is worth landing before merge since it's the same class of bug in a common resource layout.

Also verified as **not** problems, in case they come up in review: script-string casing (`CultureInfo.Name` normalizes `zh-hant` → `zh-Hant`, so the case-sensitive `==` is safe); the private-use false positive in `GetScriptSubtag` (ICU strips `-x-…` before it is ever called, `de-DE-x-hant` → `de-DE`); malformed `PrimaryLanguageOverride` (`"-"`, `"zh-"`, `"zh Hant"`, `"zh/../.."`, 200 chars all throw `CultureNotFoundException`, so the catch type is right, and `""` yields the invariant culture which the `{ Length: > 0 }` patterns already handle); and the loose `StartsWith(baseCulture)` prefix match pulling in cross-language candidates (`fi`→`fil-PH`, `ko`→`kok-IN`, `ar`→`arn-CL` — all `script == null`, so the new key is inert and nothing regresses).

Two smaller things, not worth blocking on:

- **`mn` gets no script discrimination at all.** `ResolveScript("mn")` and `ResolveScript("mn-MN")` both return `null` — .NET's chain is `mn-MN → mn → invariant` with no script anywhere — while `mn-Cyrl` → Cyrl and `mn-Mong-CN` → Mong. So keys `{mn-Cyrl, mn-Mong-CN}` with a request for `mn` still resolve to `mn-Mong-CN`, unchanged before and after. `sr`/`az`/`uz`/`bs`/`ha`/`iu`/`tg` all resolve fine; `mn` is the one hole in deriving script from `CultureInfo.Parent` rather than CLDR likely-subtags. Probably just worth stating as a known limit.
- **The `TryGetCulture` guards do less than the description implies.** Under ICU `CultureInfo.GetCultureInfo("qq-QQ")` succeeds with `Name = "qq-QQ"`, `Parent = "qq"`, so unknown-but-well-formed names get fabricated cultures rather than being guarded — harmless, since fabricated ancestors match no resource key. It does mean every distinct requested name is interned in `CultureInfo`'s own unbounded name cache (verified same-instance on repeat; `new CultureInfo(...)` does not intern), which is the growth `fix(resources): bound the script cache to resource cultures` set out to avoid, one level down. `GetCultureInfo(name, predefinedOnly: true)` makes the guard real and stops the interning.

Nice write-up on the PR, by the way — the old-vs-new sweep is what made it possible to check the deltas rather than guess at them.

src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.instanced.cs
ajpinedam: Applied.

I checked the rewrite is behaviour-identical before swapping it in — both versions over every installed culture name plus a few hand-picked shapes (`zh--Hant`, `ca-ES-VALENCIA`, `de-DE-u-co-phonebk`, `abcd-Hant`, bare `zh`): 898 names, 0 diffs.

There's now a `GetRegionSubtag` in the same shape next to it, feeding the new region tiebreak (see the sibling-ordering thread).

src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.instanced.cs
ajpinedam: Done. The walk now lives in `GetSelfAndAncestors(CultureInfo)`, and `GetAncestors(string)` is that with the head dropped:

```csharp
private static IEnumerable<string> GetAncestors(string culture)
	=> TryGetCulture(culture) is { } info ? GetSelfAndAncestors(info).Skip(1) : [];
```

`ResolveScript` iterates it directly, so the self-parent guard exists in one place instead of two.

Small bonus: for a self-parented culture the old `GetAncestors` yielded the culture itself, despite its doc comment saying it excludes it. `Skip(1)` makes the code match the contract (harmless either way — the exact culture is tried before the walk starts).

src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.instanced.cs
ajpinedam: Confirmed and fixed — thanks for the harness rows, they made this immediate to reproduce.

I took the merged option you floated in the other thread rather than the local skip, since it closes this and the sibling ordering in one move. Tiers 2 and 3 are now a single ranked candidate list (`GetFallbackCultures`):

```
candidates = ancestors ∪ prefix-siblings, ordered by
same script → same region → closest ancestor → base culture → ordinal
```

Two deliberate details:

- **The union**, not just the prefix-siblings — an ancestor whose name doesn't start with the base language would otherwise drop out of the candidate set. Across both old tiers the set was already this union, so nothing new becomes eligible; only the order changes.
- A wrong-script candidate is **demoted, not skipped**. Same outcome as your patch — your tier 3 re-admitted the skipped ancestor at the end anyway — but it makes explicit that this can never turn a resolvable string into an empty one: with `zh` as the only resource culture, a `zh-TW` request still resolves to `zh`.

All five of your rows now land on the same-script candidate:

| resource keys | request | before | after |
|---|---|---|---|
| `sr`, `sr-Cyrl-BA` | `sr-Cyrl-RS` | `sr` | **`sr-Cyrl-BA`** ✅ |
| `sr`, `sr-Cyrl-BA`, `sr-Latn-RS` | `sr-Cyrl-RS` | `sr` | **`sr-Cyrl-BA`** ✅ |
| `zh`, `zh-Hant-TW` | `zh-TW` | `zh` | **`zh-Hant-TW`** ✅ |
| `zh`, `zh-Hant-TW` | `zh-HK` | `zh` | **`zh-Hant-TW`** ✅ |
| `az`, `az-Cyrl-AZ` | `az-Cyrl` | `az` | **`az-Cyrl-AZ`** ✅ |

Verification: the 20 existing `[DataRow]`s pass, the `en` / `fi` / `ko` / `fr` / `pt` guards don't move, and an old-vs-new sweep over all 889 installed cultures against Uno's own 86 resource folders is **0 diffs** — the ranking only bites on layouts Uno itself doesn't ship. A new `When_Chinese_And_MixedFolders` test pins this alongside the region case; details in the sibling-ordering thread.

src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.instanced.cs
ajpinedam: Row 1 was a regression and it's fixed.

The merged candidate list now carries a region tiebreak between script and the ordinal fallback, so an exact region match wins among same-script candidates. That matches the Windows matching order (language+script+region → language+region → language+script → language):

```
same script → same region → closest ancestor → base culture → ordinal
```

| resource keys | request | PR head | now |
|---|---|---|---|
| `zh-HK`, `zh-Hant-TW` | `zh-Hant-HK` | `zh-Hant-TW` | **`zh-HK`** — exact region match restored |
| `zh-HK`, `zh-Hant-TW` | `zh-MO` | `zh-Hant-TW` | `zh-Hant-TW` |
| `zh-HK`, `zh-Hant-TW` | `zh-TW` | `zh-Hant-TW` | `zh-Hant-TW` |

`zh-MO` still lands on Taiwan through the ordinal tiebreak. As you say, both candidates are Traditional and which fallback Macau should prefer is unaudited, so I've left it as the accident it is rather than guessing at region distance.

Pinned with a new `When_Chinese_And_MixedFolders`, whose resource set is a bare `zh` holding **Simplified** next to `zh-HK` and `zh-Hant-TW` — that one set discriminates this and the ancestor issue at once:

| PLO | expected | what it pins |
|---|---|---|
| `zh-CN`, `zh-Hans`, `zh` | Simplified | same script |
| `zh-TW`, `zh-Hant`, `zh-MO` | Taiwan | `zh` is the nearer ancestor but the wrong script |
| `zh-Hant-HK`, `zh-HK` | Hong Kong | same script **and** matching region |

4 of those 8 rows fail against the PR head, all 8 pass after. The description now calls out the ordering change too.

src/Uno.UWP/ApplicationModel/Resources/ResourceLoader.instanced.cs
```diff
@@ -102,30 +109,164 @@ private bool FindForCulture(string culture, string resource, out string? resourc
...
+		// Resolved rather than cached: the requested culture is caller-controlled, so caching it
+		// would let a process-wide dictionary grow without bound.
+		var script = ResolveScript(culture);
+		var region = GetRegionSubtag(culture);
+
+		return _resources.Keys
+			.Where(x => ancestors.ContainsKey(x) || (baseCulture.Length > 0 && x.StartsWith(baseCulture, StringComparison.OrdinalIgnoreCase)))
+			.OrderByDescending(x => script is not null && string.Equals(GetResourceScript(x), script, StringComparison.OrdinalIgnoreCase)) // same script first
+			.ThenByDescending(x => region is not null && string.Equals(GetRegionSubtag(x), region, StringComparison.OrdinalIgnoreCase)) // then same region (Windows matches language+region ahead of language+script)
+			.ThenBy(x => ancestors.TryGetValue(x, out var depth) ? depth : int.MaxValue) // then the closest ancestor
```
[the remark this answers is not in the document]
ajpinedam: Valid, and fixed — though the region parse turned out to be the smallest of the four contributors, so the fix is broader than caching it.

Measured with `GC.GetAllocatedBytesForCurrentThread()` over 200k `GetString` calls (Release), same probe against each revision:

| path | master | as reviewed | now |
|---|---|---|---|
| exact hit (`zh-Hans` → `zh-Hans`) | 480 | 456 | 456 |
| script miss (`zh-CN` → `zh-Hans`) | 1328 | 3430 | **1040** |
| mixed set (`zh-Hant-HK` → `zh-HK`) | 1366 | 3408 | **1077** |
| typical app (`en-US` → `en`) | 1272 | 2952 | **936** |

*(bytes allocated per `GetString`)*

You were also right that it matters: the fallback path is the **default** path for an app. The Uno templates ship `Strings/en/Resources.resw`, so running under `en-US` misses the exact key and ranks candidates on every `x:Uid` lookup. (Uno's own control strings dodge it — there's literally an `en-us` folder.) Conversely the candidate set is always small, because `ProcessResourceFile` only loads cultures whose base language matches a language preference — so this was fixed per-lookup waste recomputing an identical answer, not N-scaling.

Isolating each piece against the reviewed revision (typical path, measured independently so they overlap):

| removing… | B/call | cost |
|---|---|---|
| the five-key `OrderBy`/`ThenBy` chain | 1656 | **~1300** |
| `GetRegionSubtag` | 2472 | ~480 |
| `ResolveScript` of the requested culture | 2624 | ~330 |
| LINQ `GetScriptSubtag` | 2720 | ~230 |

So four changes rather than one:

1. `_resourceScripts` → `_resourceSubtags`, holding script **and** region per resource culture. Same bound as before: keys are build-time-fixed resource cultures.
2. The five chained ordered-enumerables → one `List<string>` and one `Sort` with an explicit tie-break ladder. Biggest single win, and the ranking now reads in the same order as the doc comment.
3. `ResolveScript` of the requested culture gets a memo capped at 32 entries, cleared wholesale when it fills. Safe to cache at all because a name's script and region are pure functions of the name — an entry can't go stale — and bounded despite `PrimaryLanguageOverride` being caller-controlled, which was the concern in your earlier comment.
4. The ancestors `Dictionary` → a small array; chains are 0–3 deep.

The LINQ one-liners stay as @Xiaoy312 asked, since they now run once per culture behind a cache instead of once per lookup.

Behaviour is unchanged, and I checked the *whole ordered candidate list* rather than just the winner: 13 resource-culture layouts × 893 requested cultures = 11,609 ordered lists, **0** differences between the LINQ chain and the sort. All 28 `[DataRow]`s pass, full suite 4082 passing (13 pre-existing timezone `Given_Calendar` failures), analyzer build clean.

One note on the numbers, since the middle column looks like a regression against something better: the pre-fix revision was cheaper on those three rows because it *short-circuited on the first matching ancestor without checking its script* — the defect @Xiaoy312 found. master, which ran a full sibling ordering on every miss, is the honest baseline, and this is now ~26% under it while additionally ranking by script, region and ancestor proximity.

PR #24096, opened 2026-08-17
fix(databinding): reject bindable metadata resolved across ALC boundaries

### LINKED ISSUE #24095: [Skia] InvalidCastException in bindings when the same assembly is loaded in a second AssemblyLoadContext
### Current behavior

In a host application that loads a secondary app into a collectible `AssemblyLoadContext` (sharing `Uno.UI` with the host, per the `Uno.UI.RuntimeTests` hosting pattern), data binding fails with an `InvalidCastException` for any type whose full name exists in both load contexts:

```
System.InvalidCastException: [A]Uno.Extensions.Reactive.UI.FeedViewState cannot be cast to
[B]Uno.Extensions.Reactive.UI.FeedViewState.
Type A originates from 'Uno.Extensions.Reactive.UI' in the context 'AppALC-…' …
Type B originates from 'Uno.Extensions.Reactive.UI' in the context 'Default' …
   at MetadataBuilder_NNN.GetError(Object instance, …)  // generated BindableMetadata
   at Uno.UI.DataBinding.BindingPath.BindingItem.GetSourceValue(…)
```

The visible symptom: any `FeedView` (or other control binding such a type as its template DataContext) in the secondary app renders blank, with a build that is completely clean.

### Root cause

`BindingPropertyHelper.BindableMetadataProvider` is a process-wide slot, and generated providers match types **by full name** (`GetBindableTypeByType(type)` → `GetBindableTypeByFullName(type.FullName)`). When the host registers its own generated provider and the secondary ALC contains an identically-named type, the provider returns the *host's* `IBindableType` — whose typed getters cast to the host context's type and throw.

The generated `App.InitializeComponent` does contain a secondary-ALC branch that nulls the provider, but it does not run under this hosting model (the host constructs the app directly), and nulling the shared slot would also destroy the host's own metadata.

### Expected behavior

A metadata lookup for a type from a different load context should miss (falling back to the reflection path) rather than return metadata for an identically-named foreign type.

### How to fix

`IBindableType` already carries its `Type`, so the lookup result can be validated in one comparison: reject a provider hit whose `IBindableType.Type != originalType`. PR to follow.

### Environment

Uno.WinUI 6.7.16, Skia desktop; reproduced with a collectible-ALC-hosted app whose `Uno.Extensions.Reactive.UI` is loaded app-locally while the host also uses MVUX.

Kunal22shah:
**GitHub Issue:** closes #24095 and  https://github.com/unoplatform/studio.live/issues/3640

## PR Type:

🐞 Bugfix

## What changed? 🚀

Generated bindable metadata providers match types **by full name** (`GetBindableTypeByType(type)` delegates to `GetBindableTypeByFullName(type.FullName)`). When the same assembly is loaded in a second `AssemblyLoadContext` (a host application previewing a secondary app, per the `Uno.UI.RuntimeTests` hosting pattern), a lookup for the secondary context's type returns the *host* context's `IBindableType` — whose typed getters then throw `InvalidCastException`, and controls binding such types (e.g. `FeedView`'s `FeedViewState` as a template DataContext) silently render blank.

`IBindableType` already carries its `Type`, so the binder now validates the lookup result against the requested type identity (`BindingPropertyHelper.GetValidatedBindableType`): a mismatch is treated as a miss and falls back to the reflection path — which is where non-colliding secondary-context types already resolved. A rejected mismatch is logged at Debug with both load-context names.

Covered call sites: `BindablePropertyDescriptor.GetPropertByBindableMetadataProvider` (named + attached properties), both indexer lookups in `BindingPropertyHelper`, and `TypeActivator.CreateInstance` (where a cross-context match would silently instantiate the other context's identically-named type). Known remaining raw provider sites, deliberately out of scope for a minimal fix: `DirectUI/PropertyAccess.cs` (`HAS_UNO` branch) and the name-based lookup in `DependencyPropertyDescriptor` (no `Type` input exists to validate against).

Behavior notes:
- Single-ALC apps are unaffected: generated providers construct every entry with `typeof(X)`, so the identity check never fires for them; hot-reload replacement types carry `#N`-suffixed names and never matched the name-keyed tables in the first place.
- A custom `IBindableMetadataProvider` that deliberately returns metadata whose `Type` differs from the queried type would now be treated as a miss; generated providers never do this.

Tests: `Given_Binder.MetadataProviderIdentity` — matching-type pass-through, mismatched-type miss, and an end-to-end binding assertion that a poisoned cross-context getter is ignored in favor of the reflected value (fails with `InvalidCastException` on the previous code).

## Verification (secondary-ALC host, FeedView repro)

Three variants were run against the same repro: a host app loading a prebuilt secondary app (using MVUX `FeedView`) into a collectible ALC.

| Variant | `InvalidCastException` | Page creation / rendering | Verdict |
|---|---|---|---|
| **A - baseline** (unpatched 6.7.16) | Thrown in `FeedViewState` typed getters via the host's bindable metadata provider | Pages create, but every `FeedView` renders blank | Bug reproduced |
| **B - host-side workaround**: load `Uno.Extensions.Reactive(.UI/.Messaging)` from the host ALC (shared identity) | Gone | Page creation fails outright - navigation errors and XAML instantiation returning null (Reactive.UI registers its XAML resources per load context) | Rejected - strictly worse |
| **C - this fix**: identity-validated metadata lookup + reflection fallback | Gone | All pages render fully, 0 navigation failures, transient converter noise gone | Shipped |

B also only works at all when host and secondary app carry byte-identical `Uno.Extensions` versions, which a host cannot guarantee; C fixes the lookup for any same-named type across contexts with no host cooperation.

## PR Checklist ✅

- [ ] 📚 Docs have been added/updated following the [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features) — N/A, internal fix

### DIFF
--- a/src/Uno.UI.UnitTests/BinderTests/Given_Binder.MetadataProviderIdentity.cs
+++ b/src/Uno.UI.UnitTests/BinderTests/Given_Binder.MetadataProviderIdentity.cs
@@ -14,7 +14,10 @@ namespace Uno.UI.Tests.BinderTests
 	/// named type from the other context. Its typed getters then throw InvalidCastException. The
 	/// binder must treat such a type-identity mismatch as a miss and fall back to reflection.
 	/// </summary>
+	// RunWithProvider swaps the static BindingPropertyHelper.BindableMetadataProvider and clears the
+	// binder caches; running concurrently with any other test that touches that shared state would flake.
 	[TestClass]
+	[DoNotParallelize]
 	public partial class Given_Binder_MetadataProviderIdentity
 	{
 		[TestMethod]
--- a/src/Uno.UI.UnitTests/BinderTests/Given_Binder.MetadataProviderIdentity.cs
+++ b/src/Uno.UI.UnitTests/BinderTests/Given_Binder.MetadataProviderIdentity.cs
@@ -1,0 +1,126 @@
+#nullable enable
+
+using System;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Microsoft.UI.Xaml;
+using Microsoft.UI.Xaml.Data;
+using Uno.UI.DataBinding;
+
+namespace Uno.UI.Tests.BinderTests
+{
+	/// <summary>
+	/// Bindable metadata providers match types by full name, so when the same assembly is loaded
+	/// in a second AssemblyLoadContext a lookup can return metadata generated for the identically
+	/// named type from the other context. Its typed getters then throw InvalidCastException. The
+	/// binder must treat such a type-identity mismatch as a miss and fall back to reflection.
+	/// </summary>
+	[TestClass]
+	public partial class Given_Binder_MetadataProviderIdentity
+	{
+		[TestMethod]
+		public void When_Provider_Returns_Matching_Type_Then_Metadata_Is_Used()
+		{
+			var matching = new BindableType(0, typeof(MetadataIdentitySource));
+
+			IBindableType? result = null;
+			RunWithProvider(
+				new StubMetadataProvider(matching),
+				() => result = BindingPropertyHelper.GetValidatedBindableType(typeof(MetadataIdentitySource)));
+
+			Assert.AreSame(matching, result);
+		}
+
+		[TestMethod]
+		public void When_Provider_Returns_Mismatched_Type_Then_Lookup_Is_A_Miss()
+		{
+			// Metadata whose Type is NOT the requested type — the shape a cross-ALC full-name match produces.
+			var mismatched = new BindableType(0, typeof(MetadataIdentityOther));
+
+			IBindableType? result = null;
+			RunWithProvider(
+				new StubMetadataProvider(mismatched),
+				() => result = BindingPropertyHelper.GetValidatedBindableType(typeof(MetadataIdentitySource)));
+
+			Assert.IsNull(result);
+		}
+
+		[TestMethod]
+		public void When_Provider_Returns_Mismatched_Type_Then_Binding_Falls_Back_To_Reflection()
+		{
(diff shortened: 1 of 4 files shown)


dr1rrb (approved):
Approving — the fix is correct, minimal and well-scoped. The identity check (`bindableType.Type != type`, reference equality) is exactly the right semantics: within one ALC a type has a single `RuntimeType` instance, across ALCs they differ, so it can neither false-positive nor false-negative. Generated providers key by exact `FullName` and build each entry with `typeof(X)`, so the guard never fires in single-ALC apps. The end-to-end test (poisoned cross-context getter ignored in favor of reflection) genuinely fails on the previous code. 👍

Two non-blocking notes — feel free to address the first inline and file a follow-up for the second:

**1. Test isolation (address now if applicable).** `RunWithProvider` mutates the static `BindingPropertyHelper.BindableMetadataProvider` and calls `ClearCaches()`. The restore-in-`finally` is clean, but it relies on this class not running concurrently with other binder tests that touch the same static state / caches. If `Uno.UI.UnitTests` parallelizes at the class/method level, this can flake. Historically these binder tests aren't parallelized so it's probably fine — but worth confirming, or adding `[DoNotParallelize]` on the class for safety.

**2. Out-of-scope provider sites (follow-up issue).** The two sites documented as deliberately out of scope — `DirectUI/PropertyAccess.cs` (`HAS_UNO` branch) and the name-based lookup in `DependencyPropertyDescriptor` — can reproduce the same class of bug through a different path. Leaving them out of a minimal fix is the right call, but let's open a tracking issue so it doesn't get rediscovered without context later.

kazo0 (approved):

PR #24132, opened 2026-08-19
fix: Cull text grippers instead of locking scroll

Xiaoy312:
**GitHub Issue:** Tracked internally — no public issue for this one.

## PR Type:

🐞 Bugfix

## What changed? 🚀

**Current behavior.** While a TextBox held a live touch caret — the Android insertion handle, or a full touch selection — an enclosing `ScrollViewer` refused to scroll away from it. `ScrollViewer.ClampOffsetsToFocusedTextBox` rewrote the offset back onto the TextBox after every scroll, and `ScrollContentPresenter` discarded touch inertia outright to make the clamp land. The user-visible result was a form that could not be scrolled past the field they had just tapped.

That clamp existed to hide a different defect: the selection grippers are drawn in a popup that no `ScrollViewer` clips, so once the TextBox scrolled out of view they kept painting at their old screen position — over the app's status/nav bar, or over whatever the scroll had moved under them. Pinning the scroll meant the TextBox could never leave the viewport, so the grippers never got the chance to strand.

**This PR** removes the scroll lock and fixes the gripper problem where it belongs:

- `TextSelectionGripperPresenter.Update` now culls each gripper against the host's **ancestor-clipped** bounds (`GetGlobalBoundsWithOptions(ignoreClipping: false, …)`) rather than its raw bounds. A TextBox scrolled out of an enclosing `ScrollViewer` still has valid raw bounds, which is why the old test passed while the grippers stranded.
- Culling probes **the point the thumb hangs from** — the bottom-center of the caret line — not the caret line as a whole. A line only fractionally visible at the clip edge would otherwise still paint a full thumb past it. This matches what native Android checks (`Editor.HandleView.isPositionVisible`). The probe is a 1px rect rather than a bare point, because both sides of the test come from float matrices and a caret line ending flush with the clip (a selectable `TextBlock`'s last line) must not flicker on rounding.
- A culled gripper's **popup is closed**, and the next `ShowAt` reopens it when the anchor scrolls back into view. That is only safe because the per-frame reposition is driven by a single presenter-level `CompositionTarget.FrameRendered` subscription keyed on the gripper mode — not by the popups, and not once per showing gripper — so the loop keeps ticking with every popup closed. Leaving them open would report a popup through the public `VisualTreeHelper.GetOpenPopupsForXamlRoot` that WinUI never exposes.
- **A gripper holding the pointer capture is never culled.** Closing the popup unloads its child, and unloading an element releases its pointer captures (`UIElement.Pointers.ClearPointersStateOnUnload`), so culling mid-drag ends the gesture: the remaining moves are dropped and the release never reaches the gripper, leaving the selection toolbar closed too. The accepted cost is a dragged gripper painting outside the clip until the finger lifts; native avoids even that by auto-scrolling the container when a handle reaches the edge, which we don't do.
- A throw escaping `Update` is contained and reported once — the subscription outlives gripper visibility, so it would otherwise take the window's render loop with it. The cull *transition* logs the anchor and the clip at `Trace`, so an empty clip (`GripperClipBounds` computing nothing for an out-of-tree or zero-sized host) is distinguishable from the anchor genuinely having scrolled away.
- `ScrollViewer.ShouldSnapToTouchTextBox` / `ClampOffsetsToFocusedTextBox` and the matching inertia special-case in `ScrollContentPresenter.Managed` are gone. The thumb side length is now a named constant (`CaretWithStemAndThumb.ThumbSize`) shared by the gripper layout and the culling math, instead of a `16` repeated in both.

⚠️ **Behavior change worth a reviewer's eye** (no API surface changes — everything removed was `internal`/`private`): a focused TextBox with a touch caret no longer pins its `ScrollViewer`. That is the point of the fix, but it does reverse a previously-asserted behavior. `When_Touch_Focused_Then_Scrolled_Away` asserted the TextBox stayed pinned inside its `ScrollViewer`; it is renamed to `When_Touch_Focused_Then_Scrolled_Away_The_Scroll_Sticks` and now asserts the opposite — the user's scroll sticks and the box is allowed to leave the viewport. The gripper stranding that the old assertion was really protecting against is covered directly instead (see below).

**Intended behavior, for the record:** a focused element is scrolled *into view on focus* — it is not permanently locked there. A user scroll expresses intent and must always override it. Focus-driven bring-into-view is untouched by this PR and still covered by `When_GotFocus_BringIntoView` and `When_OuterScrollViewer_BringIntoView_Scrolls_To_Caret`.

### Test coverage

Eight runtime tests in `Given_TextBox.skia.cs` (Skia Desktop + Skia Android, for the mobile touch conventions) and one in `Given_TextBlock.cs`:

| Test | Guards |
|---|---|
| `When_Touch_Tap_Does_Not_Lock_ScrollViewer_Android` | Android insertion handle doesn't pin the scroll |
| `When_Touch_Tap_Empty_Does_Not_Lock_ScrollViewer_Android` | …same for an empty field, which now places the handle like a filled one |
| `When_Touch_Selection_Does_Not_Lock_ScrollViewer_Android` / `_iOS` | Two-thumb selection, both conventions |
| `When_Touch_Selection_Flick_Does_Not_Lock_ScrollViewer_Android` | The `ScrollContentPresenter` inertia half — the timer tests alone would still pass if inertia stayed hijacked |
| `When_Scrolled_Out_Of_View_Grippers_Are_Hidden` | Handle hides when the box scrolls out, **and comes back** when it scrolls in |
| `When_Caret_Line_Straddles_Viewport_Edge_Grippers_Are_Hidden` | The anchor-point culling rule a few px either side of the edge, plus the **thumb ink** in the band below the viewport — present while the anchor is inside (the accepted sub-thumb-height overhang), gone once culled |
| `When_Gripper_Dragged_Past_Viewport_Edge_Drag_Keeps_Tracking` | A gripper drag survives its anchor crossing the clip edge |
| `Given_TextBlock.When_IsTextSelectionEnabled_Scrolled_Out_Of_View_Then_Hides_Grippers` | The selectable-`TextBlock` host, which takes the same clip change |

The five `*_Does_Not_Lock_*` tests assert the box ended up **outside the viewport** rather than that the offset cleared a threshold: the old clamp's target was the box's own offset inside the content, which clears any threshold low enough to be safe. Verified by injecting the clamp's exact result (`ChangeView(null, 120, …)`) immediately before the assert — all five fail, and pass again once it is removed.

**Two notes for reviewers on test timing.** Several tests wait on the scroll-end snap timer (`FeatureConfiguration.ScrollViewer.SnapDelay`) plus a margin, since that timer is what the old lock fired from — if any prove flaky on slower CI agents, the waits are the place to look. And `WindowHelper.WaitForIdle` pumps the dispatcher without necessarily producing a **rendered frame**, while culling and repositioning only happen in the per-frame `Update`: tests that depend on either wait on the gripper's own reported position, or put a real delay between injected pointer moves. Back-to-back injected moves never yield to the renderer at all, which silently passed two of these tests before it was caught.

### Known gaps, tracked separately

Accepted in this PR, so a reviewer doesn't have to re-derive them:

- A thumb whose anchor sits just inside the clip edge still overhangs it by up to `ThumbSize` (16px). That is the deliberate rule — native Android probes the same anchor point — and `When_Caret_Line_Straddles_Viewport_Edge_Grippers_Are_Hidden` asserts that ink is there.
- A gripper drag is not auto-scrolled when the handle reaches the container edge, so a drag past the clip keeps painting outside it until the finger lifts. Native scrolls the container instead; we don't, and cutting the drag short is the worse of the two.
- The ancestor clip walk behind `GetGlobalBoundsWithOptions` builds an `SKPath` only to read its `Bounds`, which costs native Skia ops per clipping ancestor on a per-frame path and loses rotation. It makes this PR cull *less*, never more, so it is a follow-up rather than a prerequisite — #24163.
- The `SelectionFlyout` gets none of this treatment: it is positioned once at show time, so a scroll that doesn't press on the text leaves it parked over unrelated content — #24164. Pre-existing, and out of scope here.

## PR Checklist ✅

- [ ] 📚 Docs have been added/updated following the [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features) — n/a, no public API or documented behavior surface
- [ ] 🖼️ Validated PR `Screenshots Compare Test Run` results. — pending first CI run
- [x] ❗ Contains **NO** breaking changes — no API changes; see the behavior-change note above

### DIFF
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.skia.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.skia.cs
@@ -5762,6 +5762,92 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 			await WindowHelper.WaitFor(() => !IsGripperShowing(SUT), timeoutMS: 5000, message: "the handle must be hidden once the point it hangs from leaves the viewport");
 		}
 
+		// The other half of the culling contract: a gripper the finger is holding must NOT be culled when its
+		// anchor crosses the clip edge. Cull() collapses the gripper, and collapsing an element releases its
+		// pointer captures, so culling mid-drag used to abort the gesture - the caret froze on the last line that
+		// was still inside the viewport and the remaining finger movement went nowhere.
+		[TestMethod]
+		[PlatformCondition(ConditionMode.Include, RuntimeTestPlatforms.SkiaDesktop | RuntimeTestPlatforms.SkiaAndroid)]
+		public async Task When_Gripper_Dragged_Past_Viewport_Edge_Drag_Keeps_Tracking()
+		{
+			using var _ = new TextBoxFeatureConfigDisposable();
+			using var __ = new DisposableAction(() =>
+				(VisualTreeHelper.GetOpenPopupsForXamlRoot(WindowHelper.XamlRoot)).ForEach((_, p) => p.IsOpen = false));
+
+			// A multi-line box taller than its own text - so its inner ScrollViewer never scrolls and nothing
+			// chases the caret - inside a ScrollViewer far too short to show all of it: the lower lines are
+			// clipped away, and dragging the handle onto one of them crosses the clip edge.
+			var SUT = new TextBox
+			{
+				Width = 280,
+				Height = 400,
+				AcceptsReturn = true,
+				Text = string.Join("\r", Enumerable.Range(1, 10).Select(i => $"Line {i}")),
+				TouchSelectionConvention = TextBox.TouchTextSelectionConvention.Android,
+			};
+
+			var scrollViewer = new ScrollViewer
+			{
+				Width = 320,
+				Height = 150,
+				Content = new StackPanel { Children = { SUT } }
+			};
+			await UITestHelper.Load(scrollViewer);
+
+			var injector = InputInjector.TryCreate() ?? throw new InvalidOperationException("Failed to init the InputInjector");
+			using var finger = injector.GetFinger();
+
+			// Tap the first line to place the Android insertion handle on it.
+			var boxBounds = SUT.GetAbsoluteBoundsRect();
+			finger.Press(new Point(boxBounds.Left + 20, boxBounds.Top + 8));
+			finger.Release();
+			await WindowHelper.WaitFor(() => IsGripperShowing(SUT), message: "the tap should show the insertion handle");
+
+			// The handle's popup is placed on a later frame, so wait until it is actually over the box before
+			// pressing it - otherwise the press misses and the drag is a no-op.
+			await WindowHelper.WaitFor(
+				() => SUT.VisibleGrippersForTesting!.Value.end.GetAbsoluteBoundsRect() is { Width: > 0 } g
+					&& g.Top < scrollViewer.GetAbsoluteBoundsRect().Bottom,
+				timeoutMS: 3000,
+				message: "the insertion handle should be positioned before dragging it");
+
+			// Past the multi-tap window, so the drag that follows is its own gesture.
+			await Task.Delay(600);
+
+			// Small steps on purpose: one of them lands the caret on the first line below the viewport, which is the
+			// step that used to cull the gripper and drop its capture. A single big move would sample the last line
+			// in one go (sampleY is clamped to the text's span) and pass either way.
+			// The real delay matters as much as the step size - culling only happens in the per-frame Update, and
+			// WaitForIdle pumps the dispatcher without necessarily producing a rendered frame, so back-to-back
+			// injected moves never give culling a chance to run at all.
+			finger.Press(SUT.VisibleGrippersForTesting!.Value.end.GetAbsoluteBoundsRect().GetCenter());
+			for (var i = 0; i < 8; i++)
+			{
+				finger.MoveBy(0, 30, stepOffsetInMilliseconds: 10);
+				await Task.Delay(60);
+				await WindowHelper.WaitForIdle();
+			}
+
+			Assert.IsTrue(IsGripperShowing(SUT), "the handle must stay up while the finger is still holding it");
+
+			var caretAtEndOfDrag = SUT.SelectionStart;
+			finger.Release();
+			await WindowHelper.WaitForIdle();
+
+			// The finger ended below every line, so the caret must have tracked all the way to the last one.
+			var lastLineStart = SUT.Text.LastIndexOf('\r') + 1;
+			Assert.IsGreaterThan(1, lastLineStart, "premise: the TextBox must hold several lines");
+			Assert.IsTrue(
+				caretAtEndOfDrag >= lastLineStart,
+				$"the drag must keep tracking the finger after the handle's anchor leaves the viewport (the caret ended at {caretAtEndOfDrag}, the last line starts at {lastLineStart})");
+
+			// The premise: nothing scrolled the form to follow the caret, so the anchor really did cross the clip.
+			Assert.AreEqual(0d, scrollViewer.VerticalOffset, "the form must not have scrolled to chase the caret");
+
+			// And once the finger is off, the handle is culled again - it now points below the viewport.
+			await WindowHelper.WaitFor(() => !IsGripperShowing(SUT), timeoutMS: 5000, message: "the handle must be culled again once the drag ends below the viewport");
+		}
+
 		#endregion
 
 		[TestMethod]
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.skia.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.skia.cs
@@ -6900,12 +6900,17 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 		// When_Scrolled_Out_Of_View_Grippers_Are_Hidden.
 		[TestMethod]
 		[GitHubWorkItem("https://github.com/unoplatform/uno-private/issues/753")]
+		[PlatformCondition(ConditionMode.Include, RuntimeTestPlatforms.SkiaDesktop | RuntimeTestPlatforms.SkiaAndroid)] // mobile conventions: run on Desktop (dev) + real Android only
 		public async Task When_Touch_Focused_Then_Scrolled_Away_The_Scroll_Sticks()
 		{
 			var SUT = new TextBox
 			{
 				Width = 400,
-				Text = "Some Text"
+				Text = "Some Text",
+				// Pinned instead of left to the platform default, so every target reaches the same state: on the
+				// Android convention a single tap leaves the insertion handle up, where iOS leaves a thumbless caret
+				// and Desktop selects the tapped word. The handle is the state the removed scroll lock keyed on.
+				TouchSelectionConvention = TextBox.TouchTextSelectionConvention.Android
 			};
 
 			var sv = new ScrollViewer()
@@ -6937,25 +6942,41 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 			var injector = InputInjector.TryCreate() ?? throw new InvalidOperationException("Failed to init the InputInjector");
 			using var finger = injector.GetFinger();
 
-			// Park the box in the viewport with an explicit offset rather than StartBringIntoView: on some targets
-			// that leaves the offset at 0, and the downward drag below then has nowhere left to scroll.
+			// Park the box in the viewport with an explicit offset instead of a bring-into-view, and bound it on both
+			// sides: past 500 the box would be pushed off the top of the viewport and the geometry below stops holding.
 			sv.ChangeView(null, 470, null, disableAnimation: true);
 			await WindowHelper.WaitFor(
-				() => sv.VerticalOffset > 400,
-				message: "the TextBox should be scrolled into view before it is tapped");
+				() => sv.VerticalOffset is > 400 and < 490,
+				message: "the TextBox should be parked inside the viewport before it is tapped");
 			await UITestHelper.WaitForIdle(true);
 
-			// make the textbox touch knob appear
+			// One tap, not two. A second tap selects the word and pops the selection flyout, and then the scroll
+			// gesture below has nowhere to land: the flyout's light-dismiss overlay fills the window, so a press on
+			// it is consumed dismissing the flyout instead of scrolling, and the toolbar itself is placed above the
+			// selection, overlapping the viewport - a press there goes to its buttons. Both are correct behaviour;
+			// they just cost the gesture this test needs.
 			finger.Press(SUT.GetAbsoluteBoundsRect().GetCenter());
 			finger.Release();
-			await UITestHelper.WaitForIdle(true);
-			finger.Press(SUT.GetAbsoluteBoundsRect().GetCenter());
-			finger.Release();
+			await WindowHelper.WaitFor(
+				() => SUT.CaretMode == TextBox.CaretDisplayMode.CaretWithThumbsOnlyEndShowing,
+				message: "the tap should leave the insertion handle up");
 			await UITestHelper.WaitForIdle(true);
 
-			// scroll
+			// Premises spelled out rather than left to a mute "the offset did not move": the touch conventions can
+			// only be observed on CI, so each half of the scenario has to say when it is the one that broke.
+			Assert.AreNotEqual(FocusState.Unfocused, SUT.FocusState, "the tap should have focused the TextBox");
+			Assert.AreEqual(0, SUT.SelectionLength, "the tap should leave a caret, not a selection");
+			Assert.IsFalse(
+				SUT.SelectionFlyout?.IsOpen is true,
+				"the selection flyout must not be up - the scroll gesture would go to it instead of the ScrollViewer");
+
+			// Drag the filler above the box rather than the box itself, so the gesture is unambiguously a scroll
+			// and not a text gesture on any target.
+			var viewport = sv.GetAbsoluteBoundsRect();
+			var dragStart = new Point(viewport.GetCenter().X, (viewport.Top + SUT.GetAbsoluteBoundsRect().Top) / 2);
+
 			var offsetBeforeScroll = sv.VerticalOffset;
-			finger.Press(sv.GetAbsoluteBoundsRect().GetCenter());
+			finger.Press(dragStart);
 			await UITestHelper.WaitForIdle(true);
 			finger.MoveBy(0, 300, stepOffsetInMilliseconds: 20);
 			await UITestHelper.WaitForIdle(true);
@@ -6965,12 +6986,16 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 			// Long enough to cover the scroll-end snap timer the old lock used to yank the offset back with.
 			await Task.Delay(TimeSpan.FromSeconds(2));
 
-			Assert.IsGreaterThan(200d, offsetBeforeScroll - sv.VerticalOffset, "the scroll the user performed must stick");
+			Assert.IsGreaterThan(
+				200d,
+				offsetBeforeScroll - sv.VerticalOffset,
+				$"the scroll the user performed must stick (offset {offsetBeforeScroll} -> {sv.VerticalOffset}, "
+				+ $"caret {SUT.CaretMode}, selection {SUT.SelectionStart}/{SUT.SelectionLength}, "
+				+ $"flyout {SUT.SelectionFlyout?.IsOpen}, dragged from {dragStart} in {viewport})");
 
 			// Spelled out rather than an IsGreaterThan whose argument order reads backwards: the box must have left
 			// the viewport entirely, which is exactly what the old pin prevented.
 			var boxBounds = SUT.GetAbsoluteBoundsRect();
-			var viewport = sv.GetAbsoluteBoundsRect();
 			Assert.IsTrue(
 				boxBounds.Top >= viewport.Bottom,
 				$"the focused TextBox must be allowed to leave the viewport (box {boxBounds}, viewport {viewport})");
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.skia.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.skia.cs
@@ -6937,7 +6937,12 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml_Controls
 			var injector = InputInjector.TryCreate() ?? throw new InvalidOperationException("Failed to init the InputInjector");
 			using var finger = injector.GetFinger();
 
-			SUT.StartBringIntoView();
+			// Park the box in the viewport with an explicit offset rather than StartBringIntoView: on some targets
+			// that leaves the offset at 0, and the downward drag below then has nowhere left to scroll.
+			sv.ChangeView(null, 470, null, disableAnimation: true);
+			await WindowHelper.WaitFor(
+				() => sv.VerticalOffset > 400,
+				message: "the TextBox should be scrolled into view before it is tapped");
 			await UITestHelper.WaitForIdle(true);
 
 			// make the textbox touch knob appear
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.skia.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.skia.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Drawing;
 using System.Linq;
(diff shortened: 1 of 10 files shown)


src/Uno.UI/UI/Xaml/Controls/Primitives/CaretWithStemAndThumb.skia.cs
```diff
@@ -85,6 +91,10 @@ public CaretWithStemAndThumb(Action repositionCallback)
 		Children.Add(thumbRing);
 	}
 
+	// Test hook: whether the gripper is actually painted. A culled gripper keeps its popup open (see Cull),
+	// and its measured size survives either way, so neither alone is a reliable signal.
+	internal bool IsShowing => _popup?.IsOpen == true && Visibility == Visibility.Visible;
```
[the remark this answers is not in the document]
ajpinedam: +1

src/Uno.UI/UI/Xaml/Controls/Primitives/CaretWithStemAndThumb.skia.cs
```diff
@@ -122,6 +133,13 @@ private void OnPopupClosed(object sender, object e)
 
 	private void OnFrameRendered() => _repositionCallback?.Invoke();
 
+	/// <summary>
+	/// The character this gripper points at has scrolled out of the visible region: stop painting (and
+	/// hit-testing) it, but keep the popup open. The popup is what drives the per-frame reposition, so
+	/// closing it here would leave nothing to notice the anchor scrolling back into view.
+	/// </summary>
+	public void Cull() => Visibility = Visibility.Collapsed;
```
ajpinedam: `Cull()` sets `Visibility = Collapsed`, and the Visibility change runs `ClearPointersStateIfNeeded` (`UIElement.Pointers.cs`), which releases the element's pointer captures → `PointerCaptureLost` → `ClearGripperPointerState` ends the drag. Since `Update()` runs every frame during a drag and the clip is now ancestor-clipped, dragging the end gripper onto a line past the outer viewport (multi-line TextBox taller than the viewport, or a selectable TextBlock where `ScrollForGripper` is a no-op) collapses the gripper mid-drag with no way to resume. Pre-change the clip was the raw TextBox bounds, so in-box drags never culled; native Android hides the handle but keeps the drag alive.

Suggestion: skip (or defer) culling while the gripper has pointer capture, plus a runtime test dragging across the clip edge.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.skia.cs
```diff
@@ -5515,6 +5515,255 @@ private static async Task AssertTouchTapOnEmptyBoxPlacesCaret(TextBox.TouchTextS
...
+			await WindowHelper.WaitForIdle();
+
+			Assert.IsGreaterThan(100d, scrollViewer.VerticalOffset, "the drag should have scrolled the form down");
+
+			// The lock fired from a timer armed on scroll-end (FeatureConfiguration.ScrollViewer.SnapDelay, 250ms by
+			// default), so give it more than that to yank the offset back before asserting it stayed put.
+			await Task.Delay(FeatureConfiguration.ScrollViewer.SnapDelay + TimeSpan.FromMilliseconds(750));
+			await WindowHelper.WaitForIdle();
+
+			Assert.IsGreaterThan(100d, scrollViewer.VerticalOffset, "the ScrollViewer must not scroll back to the focused TextBox");
```
ajpinedam: These thresholds don't discriminate: the old clamp snapped the offset back to the box top = 120 (`fillerAbove`), and 120 > 100 passes. By my read, four of the five `*_Does_Not_Lock_*` tests pass with the lock restored — only `_The_Scroll_Sticks` (clamp target 432, delta 0 < 200) and the two gripper tests actually fail-before. Suggest asserting the offset is unchanged from the post-drag/post-flick reading (or a threshold above the clamp target), and one run against a lock-restored build to confirm. (Complements Copilot's note on the `sv.Bottom > SUT.Top` assert.)

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.skia.cs
```diff
@@ -5515,6 +5515,255 @@ private static async Task AssertTouchTapOnEmptyBoxPlacesCaret(TextBox.TouchTextS
...
+						new Border { Height = 1200, Background = new SolidColorBrush(Colors.Transparent) },
+					}
+				}
+			};
+
+			return (scrollViewer, textBox);
+		}
+
+		private static bool IsGripperShowing(TextBox textBox)
+			=> textBox.VisibleGrippersForTesting is { } grippers && grippers.end.IsShowing;
```
ajpinedam, line 5556: These culling tests assert `IsShowing` — the exact flag `Cull()` sets — so they'd still pass if a collapsed popup somehow kept painting, which is the regression class this PR targets. One `UITestHelper.ScreenShot` assertion (no thumb ink outside the viewport) would make them discriminating. Related: the straddle test's "+6 still showing" assert passes trivially if no frame rendered between `ChangeView` and the assert — waiting on the gripper's reported position would pin it.

src/Uno.UI/UI/Xaml/Controls/Primitives/CaretWithStemAndThumb.skia.cs
```diff
@@ -122,6 +133,13 @@ private void OnPopupClosed(object sender, object e)
 
 	private void OnFrameRendered() => _repositionCallback?.Invoke();
 
+	/// <summary>
+	/// The character this gripper points at has scrolled out of the visible region: stop painting (and
+	/// hit-testing) it, but keep the popup open. The popup is what drives the per-frame reposition, so
```
ajpinedam: Keeping the popup open as the frame driver has four costs worth weighing: (a) the `FrameRendered` subscription persists indefinitely while a selection sits off-screen — every rendered frame pays `Update()` including the clip-path walk, for grippers no one can see; (b) in `Both` mode each gripper subscribes separately, so the full geometry (2× `TransformToVisual` + `GetTotalClipPath`) recomputes twice per frame; (c) `VisualTreeHelper.GetOpenPopupsForXamlRoot` — public API — now reports an open popup WinUI never exposes, and back-button/light-dismiss consumers may target it; (d) any future "close empty popups" cleanup silently breaks gripper reappearance.

A single presenter-level `FrameRendered` subscription, independent of popup state, with a cheap early-exit while culled and nothing moved, would address all four.

ajpinedam (commented):
The direction here is right — culling at the `GripperClipBounds` seam is where this belongs, and removing the clamp has nice side benefits: the unchecked `(TextBox)FocusManager.GetFocusedElement(XamlRoot!)!` cast is gone, and managed scrolling drops a focused-element lookup + ancestor walk from every scroll-end and inertia start.

Two things look merge-gating (details inline):

1. **Culling can kill a live gripper drag** — collapsing the gripper releases its pointer capture mid-drag.
2. **Four of the five `*_Does_Not_Lock_*` tests pass even with the old lock restored** — the assert thresholds sit below the clamp target.

Also, heads-up: this overlaps with #24099 (mine), which removes the same clamp with a content-root-driven culling design. We should pick one mechanism rather than merge both — happy to discuss which parts of each to keep.

The remaining inline notes are hardening/test-fidelity: consequences of keeping the culled popup open as the frame driver, silent-cull diagnosability, and a selectable-TextBlock coverage gap. Nits, no inline needed: a UTF-8 BOM was added to three previously BOM-less files; and since the clip used is `clipPath.Bounds` (axis-aligned bbox), rotated ancestors over-approximate and never cull — fine, but worth a code comment.

One more: the PR body already flags the behavior reversal — worth confirming on the internal item that introduced the pin that keep-in-view wasn't itself the requested UX, so we're not reversing a product decision.

src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.skia.cs
```diff
@@ -720,7 +720,11 @@ private void TouchTap(Point point)
 		#region ITextSelectionGripperHost
 		TextBlock ITextSelectionGripperHost.GripperTextSurface => this;
 
-		Rect ITextSelectionGripperHost.GripperClipBounds => this.GetAbsoluteBoundsRect();
+		// Ancestor-clipped, not the raw bounds: see the matching comment on TextBox.
+		Rect ITextSelectionGripperHost.GripperClipBounds => this.GetGlobalBoundsWithOptions(
```
ajpinedam, line 724: Selectable TextBlock gets the identical clip change — and the presenter's 1px-probe comment cites the selectable-TextBlock flush-last-line case as its reason — but no test exercises this host. One selectable-TextBlock-in-ScrollViewer culling test mirroring `When_Scrolled_Out_Of_View_Grippers_Are_Hidden` would cover it.

src/Uno.UI/UI/Xaml/Controls/Primitives/TextSelectionGripperPresenter.skia.cs
```diff
@@ -186,7 +193,7 @@ public void Update()
 			}
 			else
 			{
-				gripper.Hide();
+				gripper.Cull();
```
ajpinedam: `GetGlobalBoundsWithOptions` returns `default` for a not-in-live-tree or zero-sized host, so a wrongly-empty clip culls grippers forever with nothing separating "correctly offscreen" from "clip computed empty". A guarded `Trace`-level log on the cull-state *transition* (not per frame) would make that diagnosable. And since the handler now stays subscribed while off-screen, a throw inside `Update()` (e.g. the hard `(MatrixTransform)` cast) unwinds the render loop unlogged — a try/catch with a guarded error log around the `FrameRendered` body would contain it. Minor related: subscribe uses `Visual.CompositionTarget!` while unsubscribe null-guards a fresh read — storing the subscribed instance in a field closes the leak window.

src/Uno.UI/UI/Xaml/Controls/Primitives/CaretWithStemAndThumb.skia.cs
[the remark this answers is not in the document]
Xiaoy312: pedantic

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.skia.cs
```diff
@@ -6573,16 +6827,19 @@ public async Task When_Touch_Focused_Then_Scrolled_Away()
...
 			await UITestHelper.WaitForIdle(true);
 			finger.Release();
 			await UITestHelper.WaitForIdle(true);
 
+			// Long enough to cover the scroll-end snap timer the old lock used to yank the offset back with.
 			await Task.Delay(TimeSpan.FromSeconds(2));
 
-			SUT.GetAbsoluteBoundsRect().Bottom.Should().BeApproximately(sv.GetAbsoluteBoundsRect().Bottom, 5);
+			Assert.IsGreaterThan(200d, offsetBeforeScroll - sv.VerticalOffset, "the scroll the user performed must stick");
+			Assert.IsGreaterThan(sv.GetAbsoluteBoundsRect().Bottom, SUT.GetAbsoluteBoundsRect().Top, "the focused TextBox must be allowed to leave the viewport");
```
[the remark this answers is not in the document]
Xiaoy312: The assertion reads the other way round: `Assert.IsGreaterThan(bound, value)` asserts `value > bound`, so `IsGreaterThan(sv.Bottom, SUT.Top)` is `SUT.Top > sv.Bottom` — the TextBox ended up entirely below the viewport, which is exactly the pinning this test guards against. Same argument order as the `IsGreaterThan(100d, VerticalOffset)` asserts nearby, which pass after a 200px scroll and would read `100 > 200` if it were reversed.

Rewritten anyway in 210c358ee5 as an explicit non-intersection check that prints both rects on failure, since the argument order does read backwards at a glance.

src/Uno.UI/UI/Xaml/Controls/Primitives/CaretWithStemAndThumb.skia.cs
Xiaoy312: Fixed in 25434345f4 — `Update` now skips culling any gripper that holds the pointer capture.

Reproduced first, on a multi-line TextBox in a 150px-tall ScrollViewer with the insertion handle dragged down in 30px steps: at the step where the caret lands on the first line below the clip the gripper goes `showing=False cap=False`, the remaining moves are dropped, and the caret freezes at index 57 while the last line starts at 63. New test `When_Gripper_Dragged_Past_Viewport_Edge_Drag_Keeps_Tracking` fails on exactly that without the guard and passes with it.

Two things that fell out of building it. `IsHitTestVisible = false` is wired to the same `ClearPointersStateIfNeeded` callback, so it drops the capture identically — the visibility guard is the only clean spot. And the test needs a real delay between injected moves, not just `WaitForIdle`: the latter pumps the dispatcher without necessarily producing a rendered frame, so back-to-back moves never give the per-frame `Update` a chance to cull at all. My first two red runs passed for that reason.

The accepted cost is in the code comment — a dragged gripper can paint outside the clip until the finger lifts. No auto-scroll-at-the-edge yet.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.skia.cs
Xiaoy312: Confirmed and fixed in 210c358ee5. Your read was right: the clamp's target was the box's own offset inside the content (`CreateScrollableForm`'s `fillerAbove` = 120), and `120 > 100` cleared the threshold, so all five passed with the lock in place.

They now assert the box ended up fully outside the viewport instead of checking an offset threshold. I couldn't run against a lock-restored build — the clamp is already removed on this branch — so instead I injected the clamp's exact result, `ChangeView(null, 120, …)`, immediately before the assert: all five fail, and pass again once it's removed.

I skipped the "unchanged from the post-drag reading" variant: the clamp fires from the scroll-end snap timer (250ms), so a baseline taken after the scroll settles races that timer either way.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.skia.cs
Xiaoy312: Both fixed in 210c358ee5.

The straddle test now waits on the gripper's own reported Y changing before asserting "still showing". You were right that `WaitForIdle` doesn't guarantee a rendered frame — I hit the same thing independently while building the drag test, where two red runs passed because no frame ran between injected pointer moves.

For the ink: the straddle scenario is the one where an un-culled thumb lands somewhere observable — the band immediately below the viewport, which is the accepted sub-thumb-height overhang. The test now screenshots the root visual and asserts the thumb colour *is* in that band while the anchor is inside, then *is not* once culled. The positive assertion doubles as the control: it passes, so `RenderTargetBitmap` on the RootVisual does capture the gripper popup and the negative assertion isn't vacuous. `CaretWithStemAndThumb.ThumbFillColor` is `internal` now so the expected colour can't drift from the implementation.

src/Uno.UI/UI/Xaml/Controls/Primitives/CaretWithStemAndThumb.skia.cs
Xiaoy312: Partly addressed in 0c37ee29ea, which predates this thread — leaving it open for the rest.

(b) and (d) are gone by construction: there is now a single presenter-level `FrameRendered` subscription keyed on `GripperMode` rather than on popup state, so `Update` runs once per rendered frame however many grippers are up, and a future "close empty popups" cleanup can no longer break reappearance.

(a) and (c) I've left, and I'd rather settle them with you than guess:

- (c) closes by making `Cull()` call `Hide()` — cheap now that the loop doesn't depend on the popup — but it reopens the scroll-back-in path and I'd want to re-validate reappearance timing across every culling test.
- (a) is inherent to needing frames to notice the anchor scrolling back into view; the cheap early-exit you suggest is the mitigation. For scale: `SystemFocusVisual` already runs the same `GetTotalClipPath` walk on every rendered frame for whatever has focus, so this isn't a new class of cost, and the systemic fix is the pre-existing `TODO: Do not use SKPath here, bad for perf` on `ContainerVisual.GetPrePaintingClipping`.

Happy to take either or both in this PR if you'd prefer.

src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.skia.cs
Xiaoy312: Fixed in 210c358ee5 — `Given_TextBlock.When_IsTextSelectionEnabled_Scrolled_Out_Of_View_Then_Hides_Grippers`: a selectable TextBlock in a 150px ScrollViewer, tap to select a word, scroll it out, both grippers cull, scroll back, both return.

src/Uno.UI/UI/Xaml/Controls/Primitives/TextSelectionGripperPresenter.skia.cs
Xiaoy312: All three addressed.

- The subscribed instance is stored in a field (`_frameLoop`) and both subscribe and unsubscribe go through it — that landed in 0c37ee29ea, which also moved the subscription off the popup entirely.
- `Update` is now invoked through an `OnFrameRendered` wrapper with a try/catch and a guarded error log, reported on the first failure only since it would otherwise repeat every rendered frame (210c358ee5).
- Trace log on the cull *transition*, naming the anchor and the clip — so an empty clip, meaning `GripperClipBounds` computed nothing for an out-of-tree or zero-sized host, is distinguishable from the anchor genuinely having scrolled away.

src/Uno.UI/UI/Xaml/Controls/Primitives/CaretWithStemAndThumb.skia.cs
Xiaoy312: Update on all four, now that we're keeping this mechanism rather than #24099's.

**(b) and (d) — fixed in 0c37ee29ea.** There is a single presenter-level `FrameRendered` subscription keyed on `GripperMode` instead of one per gripper keyed on popup state, so `Update` runs once per rendered frame however many grippers are up, and nothing about reappearance depends on a popup staying open.

**(c) — fixed in e482e77c9e.** `Cull()` is gone; a culled gripper's popup is closed and `ShowAt` reopens it on the way back in. You were right that this is the part worth closing: no first-party consumer acts on it today (the Android back path and `RootViewController` both go through `CloseLightDismissPopups`, which filters `IsLightDismissEnabled` — false here — and `CloseAllFlyouts` filters flyout popups), but "nothing acts on it yet" is weaker than not exposing a popup WinUI never exposes. It only became safe once the loop stopped depending on the popup. All four culling tests pass unchanged, including the two that assert reappearance.

**(a) — I'd like to decline this one, and I'd rather say why than paper over it.** The cheap early-exit can't be made sound from anything we can read cheaply. A guard built on the surface's transform-to-root plus the selection indices misses the case that matters: resize the enclosing ScrollViewer's viewport and the ancestor clip grows while neither input moves, so a gripper culled just before that resize stays culled with its anchor plainly visible. That silent stranding is a worse failure than the cost it avoids, and it's the exact bug class this PR exists to fix.

Two things that bound the real cost. `SystemFocusVisual` already runs the identical `GetTotalClipPath` walk on every rendered frame for whatever has focus, so this is one more instance of an accepted pattern rather than a new class of cost, and it only runs while a touch selection is live. And the fix that actually works is the pre-existing `TODO: Do not use SKPath here, bad for perf` on `ContainerVisual.GetPrePaintingClipping`: `GetGlobalBoundsWithOptions` builds full clip paths and then discards everything but `.Bounds`, so a rect-only ancestor walk gives the same answer far more cheaply — for this caller and for the focus visual. I'll file that as a follow-up rather than smuggle a half-measure in here.

Leaving this thread open for you, since (a) is a decline rather than a fix — happy to be argued out of it.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls/Given_TextBox.skia.cs
```diff
@@ -5515,6 +5515,377 @@ private static async Task AssertTouchTapOnEmptyBoxPlacesCaret(TextBox.TouchTextS
...
+			// Scrolling up by this much would put the anchor exactly on the viewport's bottom edge.
+			var offsetAtEdge = scrollViewer.VerticalOffset - (viewportBottom - thumbAnchorY);
+
+			// A few px short of the edge: the anchor is still inside, so the handle stays up. Waiting on the gripper's
+			// own reported position, not just on idle: WaitForIdle pumps the dispatcher without necessarily producing a
+			// rendered frame, and both the reposition and the culling only happen in the per-frame Update.
+			var gripperTopBefore = SUT.VisibleGrippersForTesting!.Value.end.GetAbsoluteBoundsRect().Top;
+			scrollViewer.ChangeView(null, offsetAtEdge + 6, null, disableAnimation: true);
+			await WindowHelper.WaitFor(
+				() => SUT.VisibleGrippersForTesting!.Value.end.GetAbsoluteBoundsRect().Top != gripperTopBefore,
```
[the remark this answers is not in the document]
Xiaoy312: The `!=` is deliberate and an epsilon comparison would weaken it.

The predicate is a *change* detector, not an equality test: both sides are the same expression (`…end.GetAbsoluteBoundsRect().Top`), read before and after a `ChangeView`, so what it waits for is "the value the layout produced is no longer the value it produced before". Exact inequality is the strongest form of that. An epsilon band would make the wait return late — or time out — for a small but real reposition, which is exactly the case this test exercises: a 6px scroll that lands the gripper's anchor just inside the viewport edge. There is no arithmetic applied to either value, so there is no rounding drift to absorb.

src/Uno.UI/UI/Xaml/Controls/Primitives/TextSelectionGripperPresenter.skia.cs
```diff
@@ -127,10 +133,53 @@ public TextSelectionGripperPresenter(ITextSelectionGripperHost host)
...
+			// First failure only: it would otherwise repeat every rendered frame.
+			if (!_repositionFailureLogged)
+			{
+				_repositionFailureLogged = true;
+				if (this.Log().IsEnabled(LogLevel.Error))
+				{
+					this.Log().Error("Failed to reposition the text selection grippers.", e);
+				}
+			}
+		}
```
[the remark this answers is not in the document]
Xiaoy312: The broad catch is the point of this handler — narrowing it would restore the failure mode it was added for.

`OnFrameRendered` is subscribed to `CompositionTarget.FrameRendered` and *stays* subscribed while the grippers are culled (a culled gripper still needs frames to notice its anchor scrolling back in), so anything escaping `Update()` takes the window's render loop with it. That is the concern raised in the review thread on `TextSelectionGripperPresenter` and the reason the wrapper exists.

A list of "known non-fatal exceptions" can't be enumerated for a call that walks the visual tree and the text layout, and any type left off the list is a dead window rather than a logged glitch. The handler already does the two things that make a broad catch defensible here: it logs at `Error`, and it logs only the first failure so a persistent one doesn't spam every rendered frame.

src/Uno.UI/UI/Xaml/Controls/Primitives/CaretWithStemAndThumb.skia.cs
Xiaoy312: Follow-up filed for the part of (a) that isn't fixable here: #24163 — give the ancestor clip walk a rect-returning form instead of building an `SKPath` only to read its `Bounds`. That's the real cost behind the per-frame `Update`, and it also carries the precision loss (a rotated clipping ancestor over-approximates once reduced to a bbox) and the pre-existing `ContainerVisual.GetPrePaintingClipping` TODO, which names `IDirectManipulationHandler.IsInBoundsForResume` as a caller that can't use the current shape at all.

Leaving this thread open for you to call on (a) itself.

ajpinedam (approved):

ajpinedam:

PR #24137, opened 2026-08-19
docs: Document the remaining Phase-5 breaking changes

Xiaoy312:
**GitHub Issue:** part of #8339

> **Stacked on #24136** — this PR targets `dev/xygu/20260819/bc75-followups`, because it edits the
> relocation section that PR adds. Merge #24136 first, then retarget this one to
> `feature/breakingchanges`.

## PR Type:

## What changed? 🚀

Nine Phase-5 breaking changes had already landed on `feature/breakingchanges` with **no entry in the 7.0 migration guide at all**, so the guide's silence read as "nothing changed here". This adds them, widens the assembly-relocation note to every type that actually moved, and retargets one now-stale sentence in the rollup spec.

**Public API removed** gains the sharpest one:

- **BC65** — `FrameworkElement`/`ContentControl` no longer implement `IEnumerable`, and `Border`/`Panel`/`ScrollViewer` lose their Uno-only `Add(UIElement)` helpers. Together those were what made C# collection-initializer markup work, so **`new StackPanel { child1, child2 }` no longer compiles** — a silent-until-you-build break for anyone writing code-behind markup. The note shows the `Children = { … }` / `Child` / `Content` replacement.

**New section — *Visibility and signature changes (WinUI parity)*** for the five that are neither removals nor reparentings:

- **BC17** `XamlCompositionBrushBase.CompositionBrush` is `protected` (was `public`; the backing DP was already `internal`, so nothing changed there).
- **BC36** `ContentPresenter.ContentTemplateRoot` is `internal` (was `public`).
- **BC19** `FlyoutBase.Close()` removed. Scoped correctly: it was `protected internal virtual`, so this hits flyouts that *derive* from `FlyoutBase`, not general callers — and since `Hide()` is non-virtual with `OnClosing`/`OnClosed` `private protected`, the entry sends them to the public `Closing`/`Closed` events rather than to an override that does not exist.
- **BC34** `ContentPresenter.OnVerticalContentAlignmentChanged` **and** `OnHorizontalContentAlignmentChanged` removed, along with the `TextBox` override of the vertical one — all no-ops; delete the overrides, nothing replaces them.
- **BC13** `WindowActivatedEventArgs.WindowActivationState` retyped from `Windows.UI.Core.CoreWindowActivationState` to `Microsoft.UI.Xaml.WindowActivationState`. Value names are identical, so only code spelling out the old enum type needs an edit.

**Type-hierarchy changes** gains three, alongside the reparentings already documented there:

- **BC52** `RadioMenuFlyoutItem` → `MenuFlyoutItem`. `IsChecked`/`GroupName` are re-declared on the type, so instance and XAML use written against the radio item are unaffected — but a `Style` with `TargetType="ToggleMenuFlyoutItem"` no longer applies, and `IsCheckedProperty` is now a **different dependency property**: reaching the value through `ToggleMenuFlyoutItem.IsCheckedProperty` still compiles, throws under a debugger, and silently reads the wrong property in Release.
- **BC73** `TimePickerFlyoutPresenter` → `Control`, dropping the `ContentControl` surface it inherited from `FlyoutPresenter`. `IsDefaultShadowEnabled` survives — the presenter declares its own — which makes it a different DP from `FlyoutPresenter.IsDefaultShadowEnabledProperty`, so the entry says so.
- **BC27** `DoubleCollection` no longer derives from `List<double>` — `AddRange`/`Sort`/`ForEach`/casts to `List<double>` are gone, and so are the interfaces `List<T>` supplied (`IReadOnlyList<double>`, `IReadOnlyCollection<double>`, non-generic `IList`/`ICollection`), which is the half that breaks call sites rather than call syntax. (`PointCollection` already worked by composition, so it is called out as unchanged rather than lumped in.)

**Relocation note widened.** The section #24136 added covered MRT Core only, but five groups of types made the identical `Uno.UI`/`Uno.UI.Composition` → `Uno` move: five `Microsoft.UI.Input` pointer types (BC41), `Microsoft.UI.System`, `Microsoft.UI.IClosableNotifier`, `Microsoft.Graphics.DirectX`/`.Display`, and MRT Core. `PointerPoint` is far more widely consumed than any MRT Core type, so a reader following the old MRT-only checklist item would have fixed the rarest of them and still hit `TypeLoadException`. The section is now a per-group table, with an explicit warning that `Microsoft.UI.Input` is **split** across the two assemblies — only those five types moved.

The checklist gains two items — replace collection-initializer markup and enumerate `Children` instead of the element (12), and fix the visibility/retype/hierarchy compile breaks (13) — and item 11 is trimmed back to one line now that the section carries the list.

**Spec follow-through.** #24136 recorded the sibling relocations as "undocumented and unguarded". This PR documents them, so that note is retargeted in `specs/050-breaking-changes-rollup/spec.md`: what remains outstanding is the *guard* — only MRT Core has an assembly-identity test — plus the fact that `Microsoft.UI.Input` ended up split rather than moved.

### Validation

- **Docs build** — DocFX 2.73.2 (the pinned CI version): the new sections, tables and code blocks render; warning/error counts unchanged from a pre-change build of the same tree (the only two warnings on this page are the pre-existing `Toolkit.GettingStarted` / `Uno.Wasm.Bootstrap.JSInterop` xrefs, which resolve only after `import_external_docs.ps1`).
- **Content verified against source, not the commit messages** — each claim was checked in the diff against `master`, which corrected three things I would otherwise have written wrong: `Close()` was `protected internal`, not public; `CompositionBrushProperty` was already `internal` before BC17; and `PointCollection` never derived from `List<T>`, so BC27 is `DoubleCollection`-only.
- **Reviewed, then corrected.** A three-lens review pass (contract / skeptic / quality) caught five substantive errors in the first draft, all now fixed and each verified in source:
  - `Microsoft.UI.Input` is **split**, not moved — only the five pointer types went to `Uno`; ~60 other types (`InputCursor`, `PointerEventArgs`, `VirtualKeyStates`, …) stay in `Uno.UI`. The old wildcard checklist item would have sent readers to `Type.GetType("…InputCursor, Uno")`, which returns `null` silently.
  - BC19's advice pointed at `override Hide()`, which does not compile: `Hide()` is non-virtual and `OnClosing`/`OnClosed` are `private protected`. The entry now points at the public `Closing`/`Closed` events.
  - BC34 removed **both** alignment virtuals (`OnVerticalContentAlignmentChanged` *and* `OnHorizontalContentAlignmentChanged`); only one was documented.
  - BC52's DP-identity trap has a specific failure mode worth stating: `ValidatePropertyOwnerOnReadWrite` is on in Debug or under a debugger (throws) and off otherwise (silently reads the wrong property).
  - BC27 also lost the interfaces `List<T>` brought — `IReadOnlyList<double>`, `IReadOnlyCollection<double>`, non-generic `IList`/`ICollection` — not just the `List<T>` methods.

- **markdownlint could not run locally** — CI uses `markdownlint-cli@0.38.0`, which needs Node ≥16; this machine has Node 14. Checked mechanically instead against the enabled rules in `build/.markdownlint.json`: fence balance, table column counts, no hard tabs, no trailing whitespace, no consecutive blanks, heading/table/fence blank-line separation, ordered-list numbering.

## PR Checklist ✅

- [ ] 🧪 Added [Runtime tests, UI tests, or a manual test sample](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) (for bug fixes / features, if applicable) — n/a, documentation and spec text only
- [ ] 🖼️ Validated PR `Screenshots Compare Test Run` results. — n/a, no code change
- [x] ❗ Contains **NO** breaking changes — documents breaking changes that already landed


MartinZikmund (approved):

doc/articles/migrating-to-uno-7.md
```diff
@@ -199,6 +199,65 @@ package together, so all of them were defined at once in a `netX.0-desktop` head
...
+configuration:
+
+```diff
+- Type.GetType("Microsoft.UI.Input.PointerPoint, Uno.UI")
++ Type.GetType("Microsoft.UI.Input.PointerPoint, Uno")
+```
+
+That rewrite applies to the Uno-rendered heads. On a **Windows** head these types come from the
+Windows App SDK rather than from Uno Platform, so an assembly-qualified string there names
+`Microsoft.WinUI` and is unaffected either way.
```
morning4coffe-dev, line 255: [P2] Do not replace the assembly name with Microsoft.WinUI for these Windows types. Actual WinAppSDK 2.3.1 metadata puts PointerPoint and the other listed Input/System/Graphics examples in `Microsoft.InteractiveExperiences.Projection`; MRT ResourceLoader is in `Microsoft.Windows.ApplicationModel.Resources.Projection`. They are neither defined nor forwarded by Microsoft.WinUI in that reference set. This guidance would produce unresolved assembly-qualified names for the preceding PointerPoint example. Use the real per-group assembly names or advise deriving the identity from the referenced type.

doc/articles/migrating-to-uno-7.md
```diff
@@ -304,6 +363,26 @@ target framework that runs on all three. `HAS_UNO_SKIA` and `__UNO_SKIA__` are u
...
+  + var panel = new StackPanel { Children = { new TextBlock(), new Button() } };
+  ```
+
+  A `Children = { … }` initializer works because `UIElementCollection` has its own `Add`; only the
+  element-level shortcut is gone.
+
+  Dropping `IEnumerable` also removes the **read** side, which is easy to miss: an element can no
+  longer be enumerated or passed where an `IEnumerable` is expected. `foreach (var child in panel)`,
+  LINQ over an element (`panel.OfType<Button>()`, `panel.Any()`), `new List<object>(panel)`, and
+  `string.Join(",", panel)` all stop compiling. Enumerate the real collection instead —
```
morning4coffe-dev, line 383: [P2] string.Join is not a compile-time migration failure. Compiling `string.Join(",", panel)` against real native WinUI StackPanel references succeeds; .NET 10 selects the params ReadOnlySpan<object?> overload and treats panel as an object, not a child sequence. Remove it from 'all stop compiling' and distinguish this non-enumerating behavior from APIs that require IEnumerable. Use `panel.Children` for the intended child join.

doc/articles/migrating-to-uno-7.md
```diff
@@ -366,6 +445,55 @@ CultureInfo.DefaultThreadCurrentCulture = CultureInfo.DefaultThreadCurrentUICult
...
+
+These members are no longer reachable, no longer have the same signature, or — where they were never
+part of the public surface — are gone entirely. Each change moves Uno to the WinUI shape, so code
+written against WinUI is unaffected.
+
+- **`XamlCompositionBrushBase.CompositionBrush` is `protected`** (it was `public`). A derived brush
+  reads and sets it exactly as before; external code that read the composition brush off a
+  `XamlCompositionBrushBase` instance no longer compiles. WinUI never exposed it publicly.
+- **`ContentPresenter.ContentTemplateRoot` is `internal`** (it was `public`), and so is
+  `UpdateContentTemplateRoot()`. To reach the materialized root from outside, walk the tree — but only
```
morning4coffe-dev, line 458: [P2] UpdateContentTemplateRoot is still public in the submitted source. `src/Uno.UI/UI/Xaml/Controls/ContentPresenter/ContentPresenter.cs:954` declares `public void UpdateContentTemplateRoot()`. ContentTemplateRoot itself is internal, but this additional method-visibility claim should be removed or deferred until the corresponding code change actually lands.

morning4coffe-dev (changes requested):
The broader migration coverage is useful, but three statements at `294c6f77cdb5` do not match the actual source/compiler contract:

1. The Windows assembly-qualified-name advice is incorrect. In the PR's own WinAppSDK 2.3.1 dependency set, PointerPoint/PointerPointProperties, ThemeSettings, IClosableNotifier, DirectXPixelFormat and DisplayInformation are defined in `Microsoft.InteractiveExperiences.Projection`, not `Microsoft.WinUI`; MRT ResourceLoader is in `Microsoft.Windows.ApplicationModel.Resources.Projection`. I checked the actual projection metadata, including exports/forwarders, and confirmed the same split with 2.4.0. Avoid a blanket Windows assembly name or provide the accurate per-group names. The preceding example itself uses PointerPoint, so this is not merely an unrelated namespace edge case.
2. `string.Join(",", panel)` does not stop compiling when the element is not enumerable. I compiled that expression against actual native WinUI StackPanel references; on .NET 10 it binds to `string.Join(string?, params ReadOnlySpan<object?>)`. The compiler will not identify this use for the migration. Describe its non-enumerating behavior separately and show `panel.Children` where child enumeration is intended.
3. Only ContentTemplateRoot is internal in this head. `ContentPresenter.cs:954` still declares `public void UpdateContentTemplateRoot()`, so the claimed visibility reduction for that method has not shipped here.

I checked the related visibility/hierarchy source as well: CompositionBrush is protected, Hide is non-virtual with private-protected closing hooks, RadioMenuFlyoutItem and TimePickerFlyoutPresenter have the documented bases, and DoubleCollection exposes the generic list interfaces rather than List<T>'s additional interfaces. Those corrections should be preserved.

Evidence here is exact-head source inspection, official WinAppSDK projection metadata and actual C# compilation, not a native app/window execution or a full DocFX build. Existing review/stack state was left unchanged; no retarget, source push or merge was performed.

PR #24142, opened 2026-08-20
test: Make the leak-counter failures diagnosable

### LINKED ISSUE #24141: Flaky test: Given_FrameworkElement_And_Leak.When_Add_Remove is undiagnosable on WASM
### Current behavior

`Given_FrameworkElement_And_Leak.When_Add_Remove` fails on Skia-WASM for `Microsoft.UI.Xaml.Controls.NavigationView` (9 of the last 10 `master` builds) and for `CommandBarFlyout_Leak` (7 of 10), e.g.

```
Expected value to be less than or equal to 67 because Microsoft.UI.Xaml.Controls.NavigationView;Grid/RootGrid;...
```

Independently of whether the underlying retention is real, three defects in the test instrument make the failure impossible to diagnose from CI output:

1. The settle loop runs until **nothing** is alive rather than until the count meets the assertion, so it spends its whole budget even once the test would pass.
2. That loop polls through `Any()` on `RemoveDeadRefsAndGetAliveRefs()`, which is a lazy iterator whose pruning `foreach` only runs on **full** enumeration — `Any()` short-circuits, so it never actually prunes.
3. The `retainedMessage` is gated to platforms that do not include Linux Skia, so failures there assert with an empty `because`; and it is built from an **earlier** walk than the one asserted on, so the reported survivors can disagree with the failure.

### Expected behavior

A failure names the surviving instances, the elapsed settle time and the platform, so the retention can be triaged without a local repro.

### How to reproduce it

Observed on `Uno.UI - CI`, `WebAssembly Skia Runtime Tests 0`. See the rates above.

### Workaround

None. Retrying the stage usually goes green, which is what has kept this invisible.

### Affected platforms

WebAssembly (Skia), Linux (Skia).

Part of #9080. Note that two siblings in the same file (`Expander`, `NavigationViewItem`) are already excluded on `SkiaWasm` citing #9080, which suggests the platform rather than the instrument is at fault — so the rows should stay enabled and the thresholds should not be raised until the instrument can actually report what survived.

MartinZikmund:
**GitHub Issue:** closes #24141

Part of #9080 ([Epic] Flaky tests roundup).

Split out of #24127 at review request — the CI-stabilization PR should not carry this.

## PR Type:

🐞 Bugfix

## What changed? 🚀

`Given_FrameworkElement_And_Leak.When_Add_Remove` fails on Skia-WASM for `Microsoft.UI.Xaml.Controls.NavigationView` in **9 of the last 10 `master` builds**, and for `CommandBarFlyout_Leak` in **7 of 10**:

```
Expected value to be less than or equal to 67 because
Microsoft.UI.Xaml.Controls.NavigationView;Grid/RootGrid;Microsoft.UI.Xaml.Controls.Grid;…
```

Whether or not the underlying retention is real, three defects in the test *instrument* make that failure impossible to triage from CI output. This PR fixes the instrument only.

### 1. The settle loop ran past the pass bar

It looped until **nothing** was alive, rather than until the count met the assertion. `expected` is now hoisted above the loop and the condition is `.Count() > expected`, which is safe because the measurement is monotone.

### 2. …and never actually pruned

That loop polled through `.Any()` on `RemoveDeadRefsAndGetAliveRefs()` — a **lazy iterator** whose pruning `foreach` only runs on *full* enumeration. `Any()` short-circuits on the first element, so the pruning pass never executed. Switching to `.Count() > expected` fixes this as a side effect.

### 3. The failure message was gated off Linux, and disagreed with the assertion

`retainedMessage` was built under a platform condition that omits Linux Skia, so failures there asserted with an **empty** `because` string. It was also derived from an *earlier* visual-tree walk than the one the assertion samples, so the survivors it named could differ from the ones that actually failed the check.

It is now derived from a single post-settle snapshot — refcount, `aliveEarlier`, `aliveLast` and the type histogram all come from the same walk — and additionally carries `sw.Elapsed` and the platform. The elapsed time answers, for free on the next run, whether the settle loop really costs the ~30 s per style that has been assumed.

### What this deliberately does *not* do

- **No row is excluded** and **no threshold is raised.** Two siblings in this same file (`Expander`, `NavigationViewItem`) are already excluded on `SkiaWasm` citing #9080 — two *more* failing on the same platform under the same tracked issue points at the platform, not at the instrument. Masking them would destroy the only signal we have.
- **No assertion that `aliveEarlier == 0`.** Under conservative stack scanning an earlier instance can be pinned as easily as the last one, so that would be a new flake, not a tightening.

## Validation

**Compile** — `Uno.UI.RuntimeTests.Skia` builds with **0 errors, 0 warnings** on `net10.0`.

**Runtime** — not validated locally, and deliberately so: these are WASM flakes at 7/10 and 9/10, so a single green local run proves nothing either way. The value of this change is that the *next* CI failure is diagnosable; that has to be measured on CI over several runs.

## PR Checklist ✅

- [x] 🧪 Added [Runtime tests, UI tests, or a manual test sample](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) (for bug fixes / features, if applicable) — test-instrument change; no new test
- [x] 📚 Docs have been added/updated following the [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features) — n/a, test-only change

https://claude.ai/code/session_01PfEBvJHHSSfsuqwsvtPPGe

### DIFF
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_FrameworkElement_And_Leak.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_FrameworkElement_And_Leak.cs
@@ -224,22 +224,23 @@ namespace Uno.UI.RuntimeTests.Tests.Windows_UI_Xaml
 		{
 			var controlType = GetControlType(controlTypeRaw);
 
-			var weakRefs = new HashSet<WeakReference<DependencyObject>>();
+			// The ordinal records which control instance a reference came from, so surviving objects can be
+			// attributed to the last instance (tolerated) or to an earlier one.
+			var weakRefs = new HashSet<(int Ordinal, WeakReference<DependencyObject> Reference)>();
 			var totalRefCount = 0;
 			void TrackDependencyObject(DependencyObject target)
 			{
-				totalRefCount++;
-				weakRefs.Add(new WeakReference<DependencyObject>(target));
+				weakRefs.Add((totalRefCount++, new WeakReference<DependencyObject>(target)));
 			}
 
-			IEnumerable<DependencyObject> RemoveDeadRefsAndGetAliveRefs()
+			IEnumerable<(int Ordinal, DependencyObject Target)> RemoveDeadRefsAndGetAliveRefs()
 			{
-				var toRemove = new List<WeakReference<DependencyObject>>();
+				var toRemove = new List<(int Ordinal, WeakReference<DependencyObject> Reference)>();
 				foreach (var weakRef in weakRefs)
 				{
-					if (weakRef.TryGetTarget(out var target))
+					if (weakRef.Reference.TryGetTarget(out var target))
 					{
-						yield return target;
+						yield return (weakRef.Ordinal, target);
 					}
 					else
 					{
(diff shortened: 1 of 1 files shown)


morning4coffe-dev (approved):
The final survivor-count allowance is unchanged. Stopping collection when that same allowance is reached avoids an unnecessary wait, and taking one survivor snapshot keeps the reported count and assertion consistent. The new ordinals distinguish earlier/last-instance survivors diagnostically; they do not introduce a stronger assertion that every earlier instance was collected. This is test diagnostics rather than a change to WinUI lifetime behavior.

PR #23619, opened 2026-07-02
Dev/bc public api phase4

ramezgerges:
**GitHub Issue:** closes https://github.com/unoplatform/uno-private/issues/2126, closes https://github.com/unoplatform/uno/issues/13709

## PR Checklist ✅

- [ ] 🧪 Added [Runtime tests, UI tests, or a manual test sample](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) (for bug fixes / features, if applicable)
- [ ] 📚 Docs have been added/updated following the [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)
- [ ] 🖼️ Validated PR `Screenshots Compare Test Run` results.
- [ ] ❗ Contains **NO** breaking changes
- [ ] 👀 Reviewed 2 other [open pull requests](https://github.com/unoplatform/uno/pulls) (optional but appreciated!)

PR #23526, opened 2026-06-23
perf: Damage Regions

ramezgerges:
**GitHub Issue:** closes https://github.com/unoplatform/uno-private/issues/2193

## PR Checklist ✅

- [ ] 🧪 Added [Runtime tests, UI tests, or a manual test sample](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) (for bug fixes / features, if applicable)
- [ ] 📚 Docs have been added/updated following the [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)
- [ ] 🖼️ Validated PR `Screenshots Compare Test Run` results.
- [ ] ❗ Contains **NO** breaking changes
- [ ] 👀 Reviewed 2 other [open pull requests](https://github.com/unoplatform/uno/pulls) (optional but appreciated!)

PR #24153, opened 2026-08-21
feat!: pluggable drawing backend abstraction (WebGPU renderer + managed engines)

ramezgerges:
**GitHub Issue:** closes unoplatform/RnD#41

## PR Type:

✨ Feature

## What changed? 🚀

Introduces a pluggable drawing-backend abstraction for the Skia rendering targets, so the renderer and the content engines behind it are selectable seams instead of a hard SkiaSharp dependency. Design doc: [`doc/uno-drawing-backend-abstraction.md`](doc/uno-drawing-backend-abstraction.md).

### Architecture

- **Drawing SPI** (`Uno.UI.Composition.Drawing`): backend-neutral contracts — `IGraphicsProvider`/`IDrawingFactory` (renderer negotiation over host-provided swapchain contexts), `IGeometry`/`IGeometryFactory`, `IFontProvider`, `IImageEncoderDecoder`, `ISvgRenderer`, retained command lists (`IRenderRecord`) and damage-region plumbing.
- **Skia backend** (`Uno.UI.Composition.Skia`): the existing SkiaSharp renderer, repackaged as one registerable backend and still the default when referenced.
- **WebGPU backend** (`Uno.UI.Composition.WebGpu` / `.Init`): a new renderer over wgpu-native (Dawn on the browser) — analytic rounded-rect/ring clips evaluated in-shader, stencil-mask path clips, frame-solid coalescing, damage-clipped presents. Opt-in via `UNO_WEBGPU=1`, backend/present-mode knobs via `UNO_WEBGPU_BACKENDS` / `UNO_WEBGPU_PRESENT`.
- **Managed (SkiaSharp-free) engines** (`Uno.UI.Composition.Managed`): geometry (flatten/combine/stroke), font loading/shaping, image decode, and SVG — enabling fully SkiaSharp-free heads (`UnoDrawingBackendSkia=false`).
- **Host-builder seams**: `UnoPlatformHostBuilder.GraphicsBackend(...)/.GeometryFactory(...)/.FontProvider(...)/.ImageEncoderDecoder(...)/.SvgRenderer(...)`, with Skia light-up preserved when the Skia backend is referenced.

### Measured (Intel UHD 620, maximized 1920×1009, damage regions disabled so every frame pays full render cost)

- SamplesApp `Performance_1000ButtonsContinuousRendering` n=1000, uncapped: WebGPU 57.8 avg / 50.2 min fps vs Skia 51.7 / 5.3.
- SamplesApp all-samples sweep (WebGPU, 699 samples measured): median 60.0 fps, 94% ≥ 55 fps.
- Uno.Gallery all-samples sweep (theme stack rebuilt against this branch): WebGPU median 60.1 fps across 91 samples, 81 ≥ 57 fps; remaining deficits vs Skia are shadow/blur-heavy pages (ShadowContainer/FAB) and a few clipboard/picker samples, tracked for follow-up.
- WebGPU lane runs with no `SkiaSharp.dll`/`libSkiaSharp` loaded (managed fonts/images/SVG/geometry).

## PR Checklist ✅

### Breaking changes

This branch intentionally carries coordinated breaking changes; libraries compiled against earlier packages need a rebuild:

- `DependencyObject` becomes a **class** on Skia targets (was an interface). Prebuilt libraries with types directly implementing the old interface fail with `TypeLoadException` until recompiled (validated: Uno.Toolkit, Uno.Themes/Material/Cupertino rebuild with minor source fixes — generic constraints `where T : class, DependencyObject` → `where T : DependencyObject`).
- The Lottie player is folded into `Uno.UI` (`Uno.UI.Lottie` assembly removed); a type-forwarding facade restores binary compatibility for existing consumers and can be shipped if needed.
- `Style.RegisterDefaultStyleForType(Type, IXamlResourceDictionaryProvider, bool isNative)` was removed; this PR restores it as a binary-compat shim (native registrations are ignored).
- Skia-target TFMs move to net10.0/net11.0 (net9.0 dropped).

## Validation

- **Runtime**: full runtime-test suite on Skia Desktop under the CI recipe; per-fix baseline comparisons against `eb88258652`; `Given_Border` re-run after the shared clip changes (27 tests, 0 failures).
- **Runtime (hardware)**: benchmark matrix and all-samples sweeps above, on a physical Intel UHD 620 machine (Vulkan for WebGPU, OpenGL for Skia).
- Known open items on the WebGPU lane are tracked for follow-up: flyout-over-scene full re-encode (damage-AABB op culling), shadow/blur pass cost, a sample-navigation memory leak under continuous full repaint, and layout-exception robustness of the render loop.


ramezgerges:
## Perf session results (samples-list scroll + WASM investigation)

**Fixes landed** (each correctness-gated on the clip-heavy runtime batch, unchanged at its 61/72 baseline — all 10 failures pre-existing and bisect-confirmed):

| Commit | Change |
|---|---|
| `0af4b939a5` | Session-clipped recordings stamp instead of full-rebuilding (rounded-container list items) |
| `62f2bb2708` | Restamps rewrite clip uniforms in place — no per-frame bind-group churn |
| `440f1c0477` | DamageRegion accumulates bounded rects, not per-contribution geometry booleans |
| `ad5e6f53bc` | ManagedContour segments as concrete arrays (Mono AOT can't compile value-type array-interface wrappers) |
| `06c8477351` | Plain-rect clips ride a dedicated ClipU slot; per-op scissors collapse 1333→1/frame |
| `39a28a2f23`+`b252582c78` | Chunked render bundles (32-op, replay-run coalescing): ~4000→~100 encoder calls/frame |
| `0485299ed1` | Per-replay ops-span reuse across frames (guards: entry identity, stamp freshness, dup exclusion, bg-cache LRU touch) |

**Measured on UHD 620 (heat-soaked laptop; cold numbers expected higher):**

| Scenario | Before | After |
|---|---|---|
| Desktop samples-list scroll | 3–7 fps | **58–64 fps** |
| Desktop buttons n=500, full repaint | — | **60 fps** (vsync-locked, MSAA 4x) |
| WASM samples-list scroll | 0.3 fps | **~42 fps** (now composition-walk-bound) |
| WASM buttons n=500, full repaint | 5.8 fps | **52–57 fps** (now rAF/present-bound) |

Root causes, in order of discovery: rounded session clips excluded recordings from the fast stamp paths; `DamageRegion.Union`'s geometry booleans were 60–80% of WASM scroll CPU (found via CDP sampling profile, symbolized with `WasmEmitSymbolMap`); per-op scissors and per-frame re-encoding dominated the WASM emit (~4000 JS-boundary calls/frame).

Also documented for follow-up: pre-existing WebGPU failures (ellipse-clip+translate ×6, stroke-thickness 1px ×2), mono 10.0.11's AOT profiler crashing at runtime init (profile-guided AOT blocked), and the WASM composition-walk cost as the next lane.

ramezgerges:
**Full-suite stability note**: running the *unfiltered* runtime-test suite against WebGPU crashes with a wgpu `Buffer does not exist` panic partway through (state-dependent; reproduced at `When_Canvas_Larger_Than_Parent_Should_Not_Clip` after the flyout/transition clusters). Bisected via kill-switches (`UNO_WEBGPU_NO_REUSE` / `UNO_WEBGPU_NO_BUNDLES`, both still crash) and confirmed on a build with all of this session's rendering commits reverted — **pre-existing** on the branch, not introduced by the perf work. Likely in the deferred-release/layer-pool lifetime handling under heavy flyout churn; filed for a separate investigation. The clip-focused batch (72 tests) remains stable at 61/72 across all configurations.

src/Uno.UI/UI/Xaml/Shapes/Ellipse.skia.cs
```diff
@@ -0,0 +1,35 @@
+using Windows.Foundation;
```
MartinZikmund, line 1: It seems a lot of these `.skia.cs` files were added, undoing a move of the code to the plain `.cs` counterparts - there is no need for `.skia.cs` in Uno.UI anymore - all code here is compiled only once, so there is no reference build. I think this might have just been incorrect merge that has re-split these files

MartinZikmund:
Heads-up before I merge #23731 (`RichTextBlock`) — it touches the text stack, so I trial-merged your branch against mine to see exactly what you'd inherit. Short version: **6 files conflict, 14 hunks, and 5 of the 6 are mechanical.** Full recipe below so none of it is a surprise.

Measured with `git merge-tree --write-tree` at merge-base `0a4053a`, against `feature/drawing-backend-abstraction@1905314` and `dev/mazi/richtextblock-winui@56454a0`. Baselined against `master × feature/drawing-backend-abstraction`, which already conflicts on three files unrelated to me (`doc/articles/feature-flags.md`, `Uno.UI.Runtime.Skia.WebAssembly.Browser.csproj`, `tsconfig.json`) — everything below is purely the delta my PR adds.

For scale: my PR is 214 files, yours is 597, and the intersection is 8 — of which `TextBlock.cs` and `Given_TextBlock.cs` auto-merge clean. No renames or deletions of my files on your side, so there's no modify/delete tangle.

---

### 1. `Segment.cs` — one `using`

Conflict is a single line. Take both sides:

```csharp
using Microsoft.UI.Xaml.Documents.RichTextServices;   // mine, needed for ObjectRun/ObjectRunMetrics
```

and drop `using SkiaSharp;` as you already do. My other ~80 lines in that file (inline-object segments, cluster/glyph index helpers) auto-merge and don't touch Skia.

### 2. `Run.cs` — take yours, carry one line

Looks like a 356-line conflict, is basically free. You moved the shaping guts to `Run.skia.cs`; I only changed one thing inside that block. Take your side wholesale, then in `Run.skia.cs`:

```csharp
private void OnFlowDirectionChanged() => InvalidateInlinesForFormatChange();
```

(was `InvalidateInlines(false)`.) The other nine `InvalidateInlines(false)` → `InvalidateInlinesForFormatChange()` renames in `Run.cs` auto-merged already.

### 3. `Inline.cs` — take yours, then split my additions across your two partials

Your split lands well. Two conflict hunks, both resolve to **take your side**, then re-apply mine in the right halves:

**Into `Inline.cs`** (all backend-neutral, no Skia):
- the `InvalidateInlines(bool updateText, bool inherited)` overload, with the existing `InvalidateInlines(bool)` delegating to it — plus its new `case Block block:` arm, which is what makes inline invalidation reach a `RichTextBlock` owner
- `InvalidateInlinesForFormatChange()`
- the `OnPropertyChanged2` override and `IsContentFormattingProperty`

**Into `Inline.skia.cs`**: the same one-word rename as above, `InvalidateInlines(false)` → `InvalidateInlinesForFormatChange()`, in the six handlers (`OnFontFamilyChanged`, `OnFontStyleChanged`, `OnFontStretchChanged`, `OnFontWeightChanged`, `OnFontSizeChanged`, `OnIsTextScaleFactorEnabledChanged`).

Nothing else — `AboveBaselineHeight`/`BelowBaselineHeight` are master's, not mine, and you've already converted them to `FontHandle`.

### 4. `FontDetails.cs` — needs one addition to `IFont`

The only place my PR asks something of your abstraction. I added `GetTextLineBoundsMetrics(TextLineBounds)`, which implements `CCompositeFontFamily::GetTextLineBoundsMetrics` (WinUI's `TextLineBounds` property). It needs **cap height**, and `IFont` doesn't expose it today.

Suggest adding alongside the other metrics:

```csharp
/// <summary>Distance from the baseline to the top of a capital letter, positive (above the baseline).</summary>
float CapHeight { get; }
```

Two implementors:
- `SkiaFont` → `_metrics.CapHeight`
- `ManagedFont` → OS/2 `sCapHeight` (`S16(data, os2 + 88)`, version 2+; falls back to the ascent, or ~0.7 em, when absent)

Then my method ports straight over:

```csharp
internal (float baseline, float lineSpacing) GetTextLineBoundsMetrics(TextLineBounds textLineBounds)
{
    var baseline = -FontHandle.Ascent;
    var lineSpacing = LineHeight;
    var capHeight = FontHandle.CapHeight;

    return textLineBounds switch
    {
        TextLineBounds.TrimToCapHeight => (capHeight, lineSpacing - baseline + capHeight),
        TextLineBounds.TrimToBaseline => (baseline, baseline),
        TextLineBounds.Tight => (capHeight, capHeight),
        _ => (baseline, lineSpacing),
    };
}
```

### 5. `UnicodeText.cs` — take mine, all five hunks

Once #4 is done this one is free, which isn't obvious from the diff. Your four body hunks swap `SKFontMetrics.Ascent/Descent` for `FontHandle.Ascent/Descent` inside `GetLineHeightAndBaselineOffset`. My version replaces those same expressions with locals:

```csharp
var (fontBaseline, fontLineSpacing) = fontDetailsWithMaxHeightInLine.GetTextLineBoundsMetrics(textLineBounds);
var fontDescent = fontLineSpacing - fontBaseline;
```

so taking my side subsumes your rename — the `FontHandle` access moved into `GetTextLineBoundsMetrics`. Hunk 1 is the matching call-site change (the method gained a leading `textLineBounds` parameter and now returns `(height, baseline)` instead of just a height); take mine there too.

The decoration-drawing code further down that file is master's, not mine, so your conversion of it stands untouched.

### 6. `ParsedText.cs` — the only one with real work

Four hunks. Two are one-liners, two are drawing code of mine that needs re-expressing against `IDrawingSession`.

**Hunk 1 (usings):** take yours, plus keep `using Uno.UI;` (I need `ResourceResolver` for the default highlighter brush). Drop my `WinUIColor` alias — your `using Windows.UI;` covers it.

**Hunk 2:** take one line from each side — your `var drawingSession = session.Session;` and my `var alignment = ResolvedTextAlignment;` (`ResolvedTextAlignment` is new in my PR; it resolves alignment through the `RichTextBlock`/`RichTextBlockOverflow` owner chain rather than reading the local field).

**Hunks 3 and 4** are `RenderHighlightedText`, `DrawCollapsingSymbol` and my `DrawText`, all written against `SKCanvas`/`SKPaint`/`SKTextBlobBuilder`. The translation is the same one you've already applied everywhere else in the file:

| mine | yours |
|---|---|
| `SKCanvas canvas` | `IDrawingSession drawingSession` |
| `Span<SKPoint> positions` | `Span<Vector2> positions` |
| `SKPaint paint` (+ `paint.Color` reads/writes) | `Color color` passed by value |
| `SKColor? highContrastSelectionColor` | `Color?` |
| `new SKColor(255, 255, 255, 255)` | `Color.FromArgb(255, 255, 255, 255)` |
| `baseColor.Alpha / 255f` | `baseColor.A / 255f` |
| `ToSkColor(color, opacity)` | your existing `WithOpacity(color, opacity)` — delete mine |

My `DrawText` (the `_textBlobBuilder.AddPositionedRun` + `UnoSkiaApi.sk_textblob_builder_make` path) collapses entirely into your local:

```csharp
GlyphRunRenderer.Draw(drawingSession, fontInfo.FontHandle, glyphs, positions, y, color);
```

`DrawCollapsingSymbol` becomes the same call with a `Vector2[]` pen walk. `RenderHighlightedText` currently mutates `paint.Color` per region and restores it — with colour-as-value that becomes a local per region, which reads better anyway.

---

### Won't conflict, won't compile

This is the part `git merge-tree` won't warn you about — four spots of mine that merge textually clean and then break the build under the abstraction:

- **`RichTextBlock.mux.cs`** and **`RichTextBlockOverflow.mux.cs`** — each has a `TextVisual`-style visual overriding `internal override SKPath? Paint(in PaintingSession session)`. Same fix as your `TextVisual`: change to `void` and delete `return null;`. Each also calls `canvas.ClipRect(new SKRect(0, 0, w, h))`, which maps 1:1 to `IDrawingSession.ClipRect(in Rect)`.
- **`BlockLayoutHelpers.MeasureShapedAdvance`** and **`SkiaTextLine.ShapeSymbol`** — ~15 lines each driving `HarfBuzzSharp.Buffer` directly and scaling by `fontDetails.TextScale.textScaleX`. Both become `IFont.Shape(...)` and get shorter, since your advances are already in pixels.
- **`Given_SkiaTextFormatter.skia.cs`** and **`Given_BlockLayoutEngine.skia.cs`** — `defaultFont.SKFontSize` → `FontSize`. Five occurrences, pure rename.

For what it's worth, that's the complete Skia surface my PR introduces: **37 lines across 10 files**, out of 214. The `BlockLayout` and `RichTextServices` layers the port adds are backend-neutral by construction, so they should drop onto your abstraction without much argument.

---

Happy to do this resolution myself on a branch off yours rather than leaving it to you — say the word and I'll push it. And if you'd rather I hold #23731 until this one lands, that's fine too; just tell me which order you prefer.

MartinZikmund (approved):

src/Uno.UI/UI/Xaml/Application.cs
```diff
@@ -152,8 +143,8 @@ public static class TraceProvider
 				unchecked((int)0xdee07725), 0x1cbf, 0x4bf6, new byte[] { 0xac, 0x8a, 0x96, 0x3, 0x60, 0xcb, 0x35, 0x12 }
 			);
 
-			public const int LaunchedStart = 1;
-			public const int LaunchedStop = 2;
+			public const int LauchedStart = 1;
+			public const int LauchedStop = 2;
```
MartinZikmund: @ramezgerges I am afraid this PR has undone quite a few things that other PRs did - e.g. this fix was done in https://github.com/unoplatform/uno/pull/23619 - and should not have been reverted.  Many files were re-split into .cs / .skia.cs - reverting changes in https://github.com/unoplatform/uno/pull/24001. Undoing these would be probably relatively easy, but I am feeling uneasy how many other PR does this silently revert

MartinZikmund: Doing a quick check now

MartinZikmund (commented):
@ramezgerges I pushed 29 commits (`49b7efa19a9..d935b97fc1d`) that restore master changes the merges into this branch had silently reverted, e.g. the #24001 `.skia.cs` fold, the #23999/#24020 dead guards, #23619, #23526/#23986 damage bits, #23667, #23898, #23736/#23844, #23941 and #23927. Each commit names the PR it restores (`fix(merge)` / `test(merge)` / `docs(merge)`). Please pull before pushing again.

While doing that I found five regressions against master that aren't merge leftovers, so they're yours to decide on. Four are inline. The fifth has no single line to point at:

**Idle acrylic renders continuously.** `Given_AcrylicBrush.When_Idle` (a `MenuFlyoutPresenter` with acrylic plus a `ScrollBar`, expecting fewer than 100 `FrameRendered` in 5 s) passes on master `5c05f74b746` but fails on this branch, both before and after the restore commits: 364-560 frames, i.e. the render loop runs at display rate while nothing changes. So any acrylic surface on screen (flyouts, menus, NavigationView panes) keeps the GPU busy on an idle app. Reproduced with Skia desktop on Win32. I haven't isolated the cause; my first suspect is the backdrop/acrylic visual requesting a frame every tick under the neutral pipeline.

src/Uno.UI/UI/Xaml/Media/CompositionTarget.Rendering.cs
```diff
@@ -292,63 +407,86 @@ private SKPath Draw(SKCanvas? canvas, Func<Size, SKCanvas> resizeFunc)
...
-
-			if (useDamageRegion && !overlayEnabled)
+			using var fpsHelperDisposable = _fpsHelper.BeginFrame();
+			using (var present = BeginPresent(Renderer, target))
 			{
-				canvas.ClipPath(OutsetDamageForPresent(damage, rasterizationScale), antialias: false);
+				// Partial repaint: when unresized and the host preserves the swapchain's pixels, clip the clear+replay
+				// to the damage region so only the changed area is repainted; otherwise repaint the whole frame.
+				var hasDamage = !resized && lastRenderedFrame.damage is { } dmg && !dmg.IsEmpty;
+				var damageEligible = hasDamage && swapChain.PreservesContents;
```
MartinZikmund: **Damage regions are a no-op on every GL host.** Partial repaint only happens when `swapChain.PreservesContents` is true, and all GL contexts return false: `Win32OpenGLGraphicsContext`, `X11OpenGLGraphicsContext` (GLX), `X11EGLGraphicsContext`, `AndroidGLGraphicsContext` and `WasmGraphicsContext` (WebGL). Those hosts clear and replay the whole frame every time, however small the change.

On master every one of them rendered through `RetainedLayer` (`Win32WindowWrapper.Rendering.OpenGl`, `X11Renderer.UsesRetainedLayer`, `UnoSKCanvasView`, `WebGlBrowserRenderer`), so #23526's partial repaint worked on GL too. WebAssembly is hit hardest, since WebGL is its default path. It's perf rather than correctness, but it undoes what #23526/#23986 bought on those hosts. Fixing it needs a host-retained offscreen per GL context (the equivalent of `RetainedLayer` behind the neutral render target), then `PreservesContents => true`.

src/Uno.UI.Composition/Composition/DamageRegion.skia.cs
```diff
@@ -1,156 +1,206 @@
...
-		// claiming rects it no longer holds. Reset unconditionally to keep the next frame sane.
-		try
+		if (IsEmpty)
 		{
-			if (_rectCount > 0)
-			{
-				_rects.AddRect(_open);
-				using var rects = _rects.Detach();
-				rects.Transform(SKMatrix.Identity, destination);
+			return null;
```
MartinZikmund, line 150: **A frame with no damage gets a full repaint instead of none.** `Detach` returns `null` for an empty region, and `CompositionTarget.Draw` reads `null` as "no damage information" (`hasDamage = !resized && damage is { } dmg && !dmg.IsEmpty`), so `useDamage` is false and the clear + replay runs over the whole target. On master, `Draw` clips to the damage path whenever the canvas wasn't recreated, so an empty frame draws nothing.

It hits the hosts that do preserve contents (Vulkan, Metal, software, framebuffer): every frame recorded while nothing visible changed (a `CompositionTarget.Rendering` subscriber, an animation that paints nothing) is repainted in full. Keeping "empty damage" distinct from "no damage tracking" (an empty geometry instead of `null`, or a flag) should bring back master's behavior. Found by code reading, not reproduced at runtime.

src/Uno.UI.Composition/Composition/ShapeVisual.skia.cs
```diff
@@ -185,49 +182,56 @@ internal override bool TryGetLocalContentBounds(out SKRect localBounds)
...
+			if (hasFill && hasStroke && fillAlpha != strokeAlpha)
+			{
+				return false;
+			}
+			var alpha = hasFill ? fillAlpha : strokeAlpha;
+			if (alpha <= 0)
+			{
+				continue;
+			}
+			if (sprite.BuildRenderGeometry() is not { } geometry)
```
MartinZikmund, line 219: **The shadow silhouette ignores each shape's `Offset` and transform.** `BuildRenderGeometry()` doesn't apply `CompositionSpriteShape.GetRenderTransform()` (`CombinedTransformMatrix` + `Offset`), which `TryGetRenderBounds` applies and `CompositionShape.Render` applies before painting. So under a `ThemeShadow` / elevation, a `ShapeVisual` whose sprite shapes carry `Offset`, `Scale`, `RotationAngle` or `TransformMatrix` (typical LottieGen / AnimatedVisualPlayer output) casts its shadow from the untransformed geometry, in the wrong place.

On master `ShapeVisual` has no `TryAddShadowPaths` override and falls back to the picture-based shadow, which is correct, so this came in with e37f3bfce79. Applying each shape's render transform here, before the ViewBox transform, should fix it. XAML `Shape`s are fine: their Stretch goes through the geometry transform, which the silhouette does include. Found by code reading, not reproduced at runtime.

src/Uno.UI/UI/Xaml/Documents/Run.cs
```diff
@@ -268,128 +275,65 @@ private SegmentInfo GetSegmentStartingFrom(int i, ReadOnlySpan<char> text)
...
-					if (buffer.Direction == Direction.RightToLeft)
-					{
-						buffer.ReverseClusters();
-					}
-
-					var glyphs = GetGlyphs(buffer, i, textSizeX, textSizeY);
+					// Legacy non-bidi path (superseded by UnicodeText): shape each segment in the run's own
+					// FlowDirection rather than resolving bidi. Ligatures are disabled because a TextBox needs each
+					// source char to stay separately addressable (uno#15528, uno#16788).
+					var direction = this.FlowDirection;
```
MartinZikmund: **RTL-script runs are now shaped left-to-right in `RichTextBlock`.** Master shaped each segment after HarfBuzz `GuessSegmentProperties()`, so a Hebrew/Arabic segment got `RightToLeft` (plus the "spaces-only segment after RTL stays RTL" rule) even with the run's default `LeftToRight` `FlowDirection`. Since 95b40e9d064 dropped the direction-guessing `IFont.Shape` overload, this uses the run's `FlowDirection` instead.

That commit treats this path as legacy, superseded by `UnicodeText`. That holds for TextBlock, but `RichTextBlock` / `RichTextBlockOverflow` still lay out through `SkiaTextFormatter` → `ParsedText` → `Run.Segments`. So Hebrew/Arabic in a default-direction `RichTextBlock` should come out in reversed glyph order. It needs a per-segment script-based direction again (as master did) until RichTextBlock moves off `ParsedText`. Found by code reading, not reproduced at runtime.

src/Uno.UI/UI/Xaml/Media/CompositionTarget.Rendering.cs
```diff
@@ -453,134 +513,61 @@ private void ReturnFrame((FramePicture picture, SKPath path, SKPath damage) fram
...
-			lock (_gate)
-			{
-				_releasedByPipeline = true;
-				dispose = ShouldDispose();
-				_disposed |= dispose;
-			}
-
-			if (dispose)
+			var t0 = _logFramePhases ? Stopwatch.GetTimestamp() : 0;
+			_rendering?.Invoke(null, new RenderingEventArgs(Stopwatch.GetElapsedTime(_start)));
```
MartinZikmund: **`RenderingEventArgs.FrameData` is always `null` on this branch**, which reverts #23762 / #23815.

On master each recorded frame is kept per `CompositionTarget` (`_latestFrames`), and `Rendering` is raised once per recorded batch on the UI thread with `FrameData` set to the latest frame of each window. Afterwards the frame is released, unless a subscriber read `FrameData` (`FrameDataAccessed`), in which case its lifetime is left to the GC. A target drops its entry when it's unregistered, and the recording is bounded to the frame so the picture's `CullRect` reports the frame size. Here `Rendering` fires from `Draw` after each present (the pre-#23762 timing), and this constructor never receives frame data, so anything that captures frames through `FrameData` gets nothing.

It isn't a mechanical restore. Master hands out an `SKPicture`, while here the frame is an opaque `IRenderRecord`, and `SkiaRenderRecord` holds a ref-counted native pointer with no finalizer and no accessor. So this needs a decision on what `FrameData` exposes per backend (e.g. an accessor to the backend's native record, an `SKPicture` on Skia) and a finalizer-safe lifetime for frames a subscriber keeps. Porting it naively would either leak a native picture every frame or break the existing contract. There are consumers relying on this API, so it should come back before merge.

src/Uno.UI/UI/Xaml/Shapes/Ellipse.skia.cs
ramezgerges: The re-split is gone: `Ellipse.skia.cs` and the other re-split partials no longer exist, and `src/Uno.UI` now carries exactly the same three `.skia.cs` files as master (`RichEditBox.HighContrast`, `TextBox.HighContrast`, `IHighContrastAdjustmentAware`). The two this PR had added on top were new files rather than re-splits, but they carried the suffix for no reason, so 70ebacc515 renames them to plain `.cs` (`FrameRenderHelper.cs`, and `LottieVisualSourceBase.Animation.cs`, named for the type it declares).

src/Uno.UI/UI/Xaml/Application.cs
ramezgerges: Master has been merged in several times since, so the re-splits and the reverted fixes are back to master's state - `src/Uno.UI` now has the same three `.skia.cs` files master does, and `Application.skia.cs` is gone. What remains in `Application.cs` against master is deliberate and specific to this PR: the `SKCanvasVisualFactory` registration moves into the Skia backend (it lives beside `SkiaDrawingSession` now), and the `REPORT_FPS` logger goes with the render loop rework.

src/Uno.UI/UI/Xaml/Media/CompositionTarget.Rendering.cs
ramezgerges: Fixed. The GL hosts no longer report that they discard their contents: each one composes into a retained FBO it owns and blits it onto the default framebuffer at present, through one shared helper (`GLRetainedFramebuffer`), so the damage region survives the swap. Win32 GL, X11 GLX, X11 EGL and Android GL all report `_contentsPreserved` now. Validated on real hardware: damage-driven frames are pixel-identical to a forced full repaint, and it took the progress-ring page from ~30 to ~60 fps on an Intel UHD 620.

src/Uno.UI.Composition/Composition/DamageRegion.skia.cs
ramezgerges: Fixed. `CompositionTarget.Draw` now distinguishes "no damage information" from "nothing was damaged": on an unresized frame whose target preserves its contents, a null damage region means nothing changed, so the clear + replay is skipped entirely rather than repainting the whole target.

src/Uno.UI.Composition/Composition/ShapeVisual.skia.cs
ramezgerges: Fixed. The silhouette now applies `sprite.GetRenderTransform()` before the ViewBox transform, so a shape carrying `Offset`, `Scale`, `RotationAngle` or `TransformMatrix` casts its shadow from the transformed geometry.

src/Uno.UI/UI/Xaml/Documents/Run.cs
ramezgerges: Fixed. The direction the shaper infers from the run's script is honoured again rather than being forced from the run's `FlowDirection`, so an RTL segment in `RichTextBlock` shapes right-to-left.

src/Uno.UI/UI/Xaml/Media/CompositionTarget.Rendering.cs
ramezgerges: Fixed. `RenderingEventArgs.FrameData` carries the recorded frame again (the Skia backend hands out its `SKPicture`; other backends hand out null). Capture is opt-in via `RenderRecordingOptions.CaptureFrameData`, set only while a `Rendering` subscriber is attached, and a `FrameHold` reproduces master's lifetime semantics - the frame is released unless a subscriber read it. Covered by `Given_CompositionTarget.When_Rendering_Carries_FrameData`.

PR #24168, opened 2026-08-22
fix(skia-x11): Fix X11 clipboard segfault on TIMESTAMP selection requests

shrimqy:
**GitHub Issue:** n/a

## PR Type:

🐞 Bugfix

## What changed? 🚀

Fixed X11 clipboard TIMESTAMP selection replies: ownership timestamp is now passed via IntPtr[] instead of a bare IntPtr, so XChangeProperty receives a pointer to the value rather than treating the timestamp as a memory address. Fixes intermittent segfaults when paste is triggered repeatedly while the app owns the clipboard.

## PR Checklist ✅


morning4coffe-dev (approved):
The fix selects the correct XChangeProperty buffer overload: the format-32 TIMESTAMP value must be supplied through an IntPtr[] buffer with one element, not interpreted as an IntPtr address. The existing raw-pointer overload explains the invalid-memory access, and the property type/element count are retained. This is Xlib protocol/marshaling correctness; there is no corresponding microsoft-ui-xaml implementation to translate.

PR #24171, opened 2026-08-22
fix: Add missing ToggleButton theme resources

### LINKED ISSUE #24159: [Skia macOS] SettingsExpander header stays gray while expanded and not hovered
## Current behavior

On Uno Desktop for macOS (Skia), a
`CommunityToolkit.WinUI.Controls.SettingsExpander` header renders with a gray
background whenever `IsExpanded` is `true`, even when the pointer never enters
the header and the header is not pressed.

The minimal repro toggles `IsExpanded` from a separate fixed button. Keeping
the pointer on that button makes the state unambiguous: the adjacent reference
`SettingsCard` retains its normal background, while the expanded
`SettingsExpander` header turns gray. Collapsing it restores the normal
background.

The same SettingsExpander usage in the originating native WinUI 3 application
retains the expected normal background.

The Toolkit template's `Checked` state does not target
`ToggleButtonGrid.Background`; on Uno, the checked state appears to expose the
underlying `ExpanderContentBackground` (`CardBackgroundFillColorSecondaryBrush`)
instead of retaining the normal `SettingsCardBackground`.

## Expected behavior

An expanded, non-hovered, non-pressed SettingsExpander header should use the
same normal background as a SettingsCard. The gray interaction backgrounds
should be limited to pointer-over and pressed states, matching native WinUI 3.

## How to reproduce

Minimal repro:
https://github.com/morning4coffe-dev/uno-settings-expander-macos-repro

```bash
git clone https://github.com/morning4coffe-dev/uno-settings-expander-macos-repro.git
cd uno-settings-expander-macos-repro
dotnet run \
  --project UnoSettingsExpanderRepro/UnoSettingsExpanderRepro.csproj \
  -f net10.0-desktop
```

1. Keep the pointer in the fixed command area at the top.
2. Click **Toggle from code**.
3. Compare the expanded **Backup destination** header with the adjacent
   **Reference SettingsCard**.
4. Click **Toggle from code** again and observe that the collapsed header
   returns to the normal background.

## Workaround

No application-level workaround has been validated. Copying the Toolkit
template and explicitly setting `ToggleButtonGrid.Background` to
`SettingsCardBackground` in the `Checked` state should avoid the visual
difference, but duplicates the control template.

## Renderer

- [x] Skia
- [ ] Native

## Affected platforms

- Desktop (macOS)
- Windows App SDK does not reproduce in the originating application.

## Versions

- Uno.Sdk: `6.8.0-dev.21`
- Uno.WinUI: `6.8.0-dev.46`
- Uno package commit:
  `e915a60e74130469db6e92128f3d7b2f88bf8d81` (2026-08-19)
- CommunityToolkit.WinUI.Controls.SettingsControls: `8.2.251219`
- .NET SDK: `10.0.302`
- macOS: `26.5.2` on Apple silicon

## IDE

GitHub Copilot CLI / `dotnet` CLI.

## Additional information

I found no matching issue in either `unoplatform/uno` or
`CommunityToolkit/Windows`. The packaged Uno commit is recent, and no relevant
Expander, ToggleButton, theme, or VisualState fix appears in the commits
between it and current `master`.

The originating application also appeared to reposition its parent
`ScrollViewer` during interaction. The repro includes offset instrumentation,
but the isolated result held `180.0 -> 180.0`, so scrolling is intentionally
not asserted as part of this report.

morning4coffe-dev:
**GitHub Issue:** closes #24159

## PR Type:

🐞 Bugfix

## What changed? 🚀

A `ToggleButton` carrying an **explicit** `Style` that doesn't set `Background` rendered with
Uno's legacy `#33000000` instead of the Fluent `ToggleButtonBackground` (`#B3FFFFFF` in light
theme). The Community Toolkit `SettingsExpander` hits this exactly: its `Expander` template
declares `<ToggleButton x:Name="ExpanderHeader" Style="{StaticResource ExpanderHeaderDownStyle}" />`
without a local `Background`, and neither `ExpanderHeaderDownStyle` nor its visual states touch
`ToggleButtonGrid.Background`, so the header background comes straight from the built-in style.

### Root cause: a WinUI theme-resource parity gap, not default-style registration

WinUI resolves the built-in style from the framework's own `generic.xaml` and never from
`Application.Resources` (`CControl::ApplyBuiltInStyle` → `FxCallbacks::Control_GetBuiltInStyle`
→ `DefaultStyles::GetDefaultStyleByTypeInfo` → `StyleCache::GetFrameworkStyles`). Uno's
"legacy built-in style + Fluent implicit style" split is therefore the correct shape.

What makes WinUI still show Fluent values for an explicitly-styled control is the *indirection*
in `generic.xaml` plus the app-level override rule:

```xml
<!-- dxaml/xcp/dxaml/themes/generic.xaml -->
<Style TargetType="ToggleButton">
  <Setter Property="Background" Value="{ThemeResource ToggleButtonBackground}" />
...
<StaticResource x:Key="ToggleButtonBackground" ResourceKey="SystemControlBackgroundBaseLowBrush" />
```

```cpp
// dxaml/xcp/core/dll/xcpcore.cpp - CCoreServices::LookupThemeResource
IFC(m_pThemeResources->GetKeyNoRef(strKey, &pValue));
// If this is a theme resource key, allow Application.Resources to override it.
```

`XamlControlsResources` redefines `ToggleButtonBackground`, so the built-in style picks up the
Fluent brush. Uno already implements the same override
(`ThemeResourceReference.RefreshValue` → `ResourceResolver.TryApplicationResourceOverride`) and
already uses those indirection keys for `Button`, `CheckBox`, `ComboBox`, `Slider`, `TextBox`, …
— **`ToggleButton` was the outlier**: `SystemResources.xaml` only carried the `Checked*` subset of
the `ToggleButton*` block, and `Generic.xaml` hard-coded the raw `SystemControl*` brushes, which
`XamlControlsResources` does not (and must not) redefine.

### This change

- `SystemResources.xaml`: adds the 28 missing `ToggleButton*` theme resources (Background /
  Foreground / BorderBrush × Normal, PointerOver, Pressed, Disabled, Indeterminate*) to all three
  theme dictionaries, with the exact values from WinUI's `generic.xaml`.
- `Generic.xaml`: points `XamlDefaultToggleButton` and its visual states at those keys instead of
  the raw `SystemControl*` brushes. Non-Fluent appearance is unchanged — the new keys are
  `StaticResource` aliases of the very same brushes.
- Reverts the earlier `XamlFileGenerator` change that let Fluent theme assemblies register
  built-in default styles for `Uno.UI` control types. That diverged from WinUI (see above) and
  regressed `Given_TextBox.When_Size_Zero_Default`,
  `Given_TextBox.When_Size_Zero_Fluent_ComboBoxTextBoxStyle`, `Given_ToolTip.When_Switch_Theme_Uwp`,
  `Given_ListViewBase.When_ThemeChange`,
  `Given_Button.When_AccentButtonStyle_Has_Same_Height_As_Default` and
  `Given_ComboBox.When_Customized_Popup_Placement` in CI.
- Keeps the `Style.RegisterDefaultStyleForType` cache invalidation (a materialized default style
  survived provider replacement, e.g. across a hot-reload `GlobalStaticResources` refresh) and its
  unit test.
- Runtime tests: `Given_Style.When_Explicit_Style_Uses_Fluent_Default_Value` (Fluent value wins)
  and `Given_Style.When_Explicit_Style_Uses_Uwp_Default_Value` (legacy value still wins without
  `XamlControlsResources`).

### Validation

Skia Desktop (Win32), `net10.0`, Release — all green after the change, and the five reproduced
CI regressions fail before it:

| Test | Before | After |
| --- | --- | --- |
| `Given_Button.When_AccentButtonStyle_Has_Same_Height_As_Default` | Failed | Passed |
| `Given_ListViewBase.When_ThemeChange` | Failed | Passed |
| `Given_TextBox.When_Size_Zero_Default` | Failed | Passed |
| `Given_TextBox.When_Size_Zero_Fluent_ComboBoxTextBoxStyle` | Failed | Passed |
| `Given_ToolTip.When_Switch_Theme_Uwp` | Failed | Passed |
| `Given_ToolTip.When_Switch_Theme_Fluent` | Passed | Passed |
| `Given_ComboBox.When_Customized_Popup_Placement` (×4) | Passed | Passed |
| `Given_Style.When_Explicit_Style_Uses_Fluent_Default_Value` | Failed (`#33000000` vs `#B3FFFFFF`) | Passed |
| `Given_Style.When_Explicit_Style_Uses_Uwp_Default_Value` | Passed | Passed |

Analyzer-enabled build (`-p:UnoFastDevBuild=false`) of `Uno.UI.UnitTests` and its whole
`Uno.UI`/`Uno.UI.FluentTheme.*` graph: 0 warnings, 0 errors.

The 'before' column is a measured run: reverting only the two theme files on top of this branch
rebuilds a ToggleButton whose explicit-style background is `#33000000` again.

## PR Checklist ✅

- [ ] 📚 Docs have been added/updated following the [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features) — not applicable, WinUI parity fix


MartinZikmund:
I think this is may actually be the correct behavior for most built-in controls. Only the ones that were WinUI 2+ have the fluent as default correctly. Can you verify if WinUI behaves the same way?

morning4coffe-dev:
You were right — thanks for pushing back. I verified against the WinUI 3 source
(`microsoft-ui-xaml` @ `439bda90582f7892ca595a193f55dbfc6ae2010c`) and reworked the PR.

**WinUI never resolves the built-in style from `Application.Resources`.**
`CControl::ApplyBuiltInStyle` → `FxCallbacks::Control_GetBuiltInStyle`
(`dxaml/xcp/dxaml/lib/FxCallbacks.cpp:854`) → `DefaultStyles::GetDefaultStyleByTypeInfo` →
`m_cache.GetFrameworkStyles(...)` (`dxaml/xcp/dxaml/lib/DefaultStyles.cpp`). It always comes from
the framework's `generic.xaml`, which still carries the pre-Fluent
`SystemControl*`-based styles (e.g. `<Style TargetType="ToggleButton">` at
`dxaml/xcp/dxaml/themes/generic.xaml:6171`). So Uno's split — legacy built-in style at
`ImplicitStyle` precedence + Fluent implicit style at `ExplicitStyle` precedence — is the correct
WinUI shape, and my previous approach (registering Fluent styles as built-in defaults) diverged
from it. It also broke `Given_TextBox.When_Size_Zero_Default`,
`Given_TextBox.When_Size_Zero_Fluent_ComboBoxTextBoxStyle`, `Given_ToolTip.When_Switch_Theme_Uwp`,
`Given_ListViewBase.When_ThemeChange`,
`Given_Button.When_AccentButtonStyle_Has_Same_Height_As_Default` and
`Given_ComboBox.When_Customized_Popup_Placement` — all reproduced locally on Skia Desktop and all
green again after the revert.

**The real gap is theme-resource parity.** WinUI's built-in style doesn't reference the raw brush,
it references an indirection key that `XamlControlsResources` can override:

```xml
<!-- generic.xaml -->
<Style TargetType="ToggleButton">
  <Setter Property="Background" Value="{ThemeResource ToggleButtonBackground}" />
...
<!-- Resources for Windows.UI.Xaml.Controls.Primitives.ToggleButton -->
<StaticResource x:Key="ToggleButtonBackground" ResourceKey="SystemControlBackgroundBaseLowBrush" />
```

and `CCoreServices::LookupThemeResource` (`dxaml/xcp/core/dll/xcpcore.cpp:2344`) explicitly allows
`Application.Resources` to override anything found in the global theme dictionaries. Uno already
implements exactly that (`ThemeResourceReference.RefreshValue` →
`ResourceResolver.TryApplicationResourceOverride`), and `Generic.xaml` already uses those keys for
`Button` (`ButtonBackground`), `CheckBox` (`CheckBoxBackgroundUnchecked`), `ComboBox`, `Slider`,
`TextBox`, … — `ToggleButton` was simply the outlier: `SystemResources.xaml` only had the
`ToggleButton*Checked*` subset, and the built-in style hard-coded
`{ThemeResource SystemControlBackgroundBaseLowBrush}`, a key `XamlControlsResources` does not
redefine.

The PR is now just that parity fix: add the 28 missing `ToggleButton*` keys to the three theme
dictionaries with WinUI's exact values, and point `XamlDefaultToggleButton` at them. Non-Fluent
appearance is unchanged (the new keys are `StaticResource` aliases of the same brushes), which
`Given_Style.When_Explicit_Style_Uses_Uwp_Default_Value` now asserts alongside the Fluent case.

For the record on the same audit, this indirection gap is *not* unique to `ToggleButton` — I
diffed every implicit `<Style TargetType="X">` in `generic.xaml` against Uno''s `Generic.xaml`
and these still hard-code `SystemControl*` brushes or literals where WinUI uses an
app-overridable key:

| Uno built-in style | Setter | WinUI key | Uno value |
| --- | --- | --- | --- |
| `HyperlinkButton` | `Foreground` | `HyperlinkButtonForeground` | `SystemControlHyperlinkTextBrush` |
| `RadioButton` | `Background` / `Foreground` | `RadioButtonBackground` / `RadioButtonForeground` | `Transparent` / `SystemControlForegroundBaseHighBrush` |
| `Slider` | `Background` / `Foreground` | `SliderTrackFill` / `SliderTrackValueFill` | `SystemControlForegroundBaseMediumLowBrush` / `SystemControlHighlightAccentBrush` |
| `PasswordBox` | `Background` / `Foreground` | `TextControlBackground` / `TextControlForeground` | `SystemControlBackgroundAltHighBrush` / `SystemControlForegroundBaseHighBrush` |
| `ListViewHeaderItem`, `GridViewHeaderItem` | `Background` | `ListViewHeaderItemBackground` / `GridViewHeaderItemBackground` | `Transparent` |
| `ToggleSwitch` | `Foreground` | `ToggleSwitchContentForeground` | `SystemControlForegroundBaseHighBrush` |

None of those are reachable from #24159 and each carries its own visual risk, so I kept them out
of this PR rather than bundling a broad theme rewrite into a bugfix — happy to file a follow-up
issue if you''d like them tracked.

src/Uno.UI/UI/Xaml/Style/Style.cs
```diff
@@ -289,6 +289,7 @@ private static void EnumerateSetters(Style style, Dictionary<object, SetterBase>
 		public static void RegisterDefaultStyleForType(Type type, IXamlResourceDictionaryProvider dictionaryProvider)
 		{
 			_lookup[type] = ProvideStyle;
+			_defaultStyleCache.Remove(type);
```
MartinZikmund, line 293: Is this specific line needed? If yes, why?

morning4coffe-dev: Yes. `GetDefaultStyleForType` reads `_defaultStyleCache` before invoking the provider in `_lookup`. Replacing `_lookup[type]` alone therefore leaves an already materialized Style in use and the replacement provider is never consulted. Removing this entry makes the next lookup materialize the replacement style.

`When_Default_Style_Provider_Is_Replaced` exercises that exact sequence: materialize the original style, register a replacement provider, then assert the replacement is returned. The focused explicit-style suite passes. I added a short explanation next to the invalidation as well. This is Uno resource-provider/cache integration, not an additional ToggleButton state change.

morning4coffe-dev:
The tvOS failure was traced to the two new tests comparing an application-theme resource with a control using a different host theme. The device logs consistently showed dark expected brushes (`#0FFFFFFF` / `#33FFFFFF`) versus light-theme actual brushes (`#B3FFFFFF` / `#33000000`).

The tests now explicitly give the control the application theme used for the resource lookup, while loading it under an opposite-theme host to exercise that boundary. No production style behavior was changed for this CI fix. The two explicit-style cases and the existing ToggleButton state-resource case all pass in desktop runtime QA. Fresh CI is running for the new commit; the reviewer-owned cache-invalidation discussion remains open for the reviewer.

PR #24199, opened 2026-08-25
feat(toolkit): Add opt-in touch scrollbar thumb drag

ajpinedam:
**GitHub Issue:** Tracked internally (customer-reported). No public issue.

## PR Type:

✨ Feature

## What changed? 🚀

A customer reported that on iOS/iPadOS WebAssembly, touching a scrollbar reveals it but dragging the thumb does not scroll — in a plain `ScrollViewer` and in CommunityToolkit `DataGrid`.

**This turned out to be intended WinUI behaviour, not a bug.** `dxaml\xcp\dxaml\lib\ScrollBar_Partial.cpp` (`AttachEvents`, tag `winui3/release/1.6.5`) calls `put_IgnoreTouchInput(TRUE)` on both thumbs and all eight repeat buttons — ten parts, which Uno ports one-for-one — and `Thumb::put_IgnoreTouchInput` exists solely for that purpose, defaulting to `FALSE` with `ScrollBar` as its only caller. WinUI also drives `TouchIndicator` deliberately. So a touch drag on a scrollbar thumb is not possible on WinUI by design.

The gap is nonetheless real on touch-only targets: a browser on a tablet has no mouse, so the scrollbar is the only affordance and WinUI's assumption does not hold. That makes this an **enhancement**, which must not change default behaviour on any target.

It ships as an opt-in `Uno.UI.Toolkit` attached property:

```xml
<Style TargetType="ScrollBar">
  <Setter Property="toolkit:ScrollBarExtensions.IsTouchThumbDragEnabled" Value="True" />
</Style>
```

Applied through an implicit `Style` as above it also covers scrollbars an app cannot reach directly, including `DataGrid`'s internal bars.

### This does change core code, behind the opt-in

An earlier revision tried to do this entirely from the Toolkit, by walking the template to clear `IgnoreTouchInput` and by pinning `IndicatorMode`. iPad validation showed that approach is not sound, for two reasons:

1. `ScrollBar.AttachEvents` re-applies `IgnoreTouchInput = true` to its parts every time it attaches them — on template application and on `Loaded` — so anything set from outside is undone. The `ScrollViewer` template also defers its bars (`x:Load="False"`), so with `Auto` visibility a bar is realized only once an axis overflows, after any external code has run.
2. `IndicatorMode` is driven by the `ScrollViewer`'s visual states, i.e. at **animation precedence**, so a local value set from the Toolkit loses. After any finger pan the bar returned to `TouchIndicator`, where the template collapses the interactive root the thumb lives in.

So the opt-in is now a per-instance internal flag on `ScrollBar`, honoured in the two places that decide this, both gated:

- `AttachEvents` applies `IgnoreTouchInput = !IsTouchThumbDragEnabled` to the ten parts instead of a hard-coded `true` — the exact line WinUI hard-codes.
- `IsConscious()` becomes per-instance (`ShouldUseDynamicScrollbars() && !IsTouchThumbDragEnabled`), so an opted-in bar takes the same path WinUI takes when the OS "automatically hide scroll bars" setting is off: it keeps the interactive indicator instead of the thin touch one.

Default behaviour is unchanged on every target, which is pinned by a test (below). The Toolkit extension is now ~45 lines that set that flag, with no visual-tree walking and no timing assumptions.

### Known limitation

While enabled the bar stays visible and interactive, so a pan started **on** the scrollbar drags the thumb rather than panning the content — the same as any pointer-driven scrollbar. Documented in `doc/articles/Uno.UI.Toolkit.md`.

### Validation

- **Runtime** — Skia Desktop (`SamplesApp.Skia.Generic`), `Given_ScrollBar`: **10 passed / 0 skipped**.
  - Touch thumb drag for **both orientations**, asserting `Value` plus the `ThumbTrack` and `EndScroll` sequence.
  - Both a **standalone** `ScrollBar` (the `DataGrid` shape) and a `ScrollViewer`-hosted one, the latter across **`Visible` and `Auto`** bar visibility — `Auto` matters because the template defers those bars.
  - `When_Touch_Panned_Then_Thumb_Is_Still_Draggable` pans with a finger first, asserts the bar really reached `TouchIndicator`, and only then drags the thumb, so it cannot pass without entering the state that broke on device.
  - `When_Reloaded_Then_Touch_Thumb_Drag_Survives` covers the re-attach path from (1) above.
  - `When_Touch_Drags_Thumb_Without_OptIn_Then_Value_Is_Unchanged` pins the **default** behaviour: without opting in, a touch drag must not move the bar.
  - Fails-before/passes-after: with the core change neutered, 8 of those 10 fail — while the default-behaviour test still passes, which is the evidence the change is genuinely gated.
- **Device** — validated on a physical iPad (iPadOS Safari, Skia WebAssembly) with the added manual sample `ScrollViewer_TouchScrollBarDrag` (category *Scrolling*), which has an in-page toggle for the opt-in so both states can be compared without a rebuild.
- **Compile** — 0 errors.

## PR Checklist ✅

### DIFF
--- a/doc/articles/Uno.UI.Toolkit.md
+++ b/doc/articles/Uno.UI.Toolkit.md
@@ -120,3 +120,30 @@ You can also apply `FromJson` directly inside `Page.DataContext`, either by refe
 ```
 
 If the JSON string is empty or invalid, the extension throws a `XamlParseException`, allowing issues to surface early during page initialization.
+
+## ScrollBar - dragging the thumb with a finger
+
+WinUI has the `ScrollBar` parts ignore touch on purpose: touch scrolling goes through direct manipulation, and the bar is only an indicator, so a finger on the thumb does nothing. A pointer reveals the interactive bar on hover, and touch has no hover. That is the default in Uno Platform as well.
+
+On a touch-only target - a browser on a tablet most notably - there is no pointer to reveal that bar with, so the user has nothing to drag. Opt into an interactive bar with `ScrollBarExtensions.IsTouchThumbDragEnabled`:
+
+```xml
+xmlns:toolkit="using:Uno.UI.Toolkit"
+```
+
+```xml
+<ScrollBar Orientation="Vertical"
+           toolkit:ScrollBarExtensions.IsTouchThumbDragEnabled="True" />
+```
+
+While enabled, the bar keeps its interactive (mouse) indicator instead of the thin touch indicator, which both keeps it visible and lets a finger drag the thumb. Panning the content with a finger is unaffected. The property has no effect on Windows.
+
+The bar of a `ScrollViewer` is a template part, so it cannot be addressed directly. Reach it with an implicit `Style` in your application - or page - resources, which is also how the bars inside a third-party control such as the CommunityToolkit `DataGrid` are reached:
+
+```xml
+<Style TargetType="ScrollBar">
+    <Setter Property="toolkit:ScrollBarExtensions.IsTouchThumbDragEnabled" Value="True" />
+</Style>
+```
+
+A known limitation: a finger pan which *starts* on a visible bar does not scroll the content, because hit-testing resolves into the `ScrollBar` and the content presenter is a sibling of it. Start the pan over the content instead.
--- a/doc/articles/Uno.UI.Toolkit.md
+++ b/doc/articles/Uno.UI.Toolkit.md
@@ -136,7 +136,9 @@ xmlns:toolkit="using:Uno.UI.Toolkit"
            toolkit:ScrollBarExtensions.IsTouchThumbDragEnabled="True" />
 ```
 
-While enabled, the bar keeps its interactive (mouse) indicator instead of the thin touch indicator, which both keeps it visible and lets a finger drag the thumb. Panning the content with a finger is unaffected. The property has no effect on Windows.
+While enabled, the bar keeps its interactive (mouse) indicator instead of the thin touch indicator, which both keeps it visible and lets a finger drag the thumb. This also shows on a device with both a mouse and a touch screen: after a finger pan the bar stays expanded instead of switching to the thin indicator. Panning the content with a finger is unaffected.
+
+The property is honoured wherever scroll bar input is handled by Uno Platform's own `ScrollBar` - that is, on every Skia target: Desktop (Windows, macOS, Linux), WebAssembly, Android, and iOS. It has no effect on Windows (WinAppSDK), where the WinUI `ScrollBar` handles the input. The native Android, iOS, and WebAssembly UI backends are not supported targets for this opt-in - touch scrolling there is performed by the OS - so leave the property unset on them.
 
 The bar of a `ScrollViewer` is a template part, so it cannot be addressed directly. Reach it with an implicit `Style` in your application - or page - resources, which is also how the bars inside a third-party control such as the CommunityToolkit `DataGrid` are reached:
 
--- a/src/SamplesApp/SamplesApp.Samples/Windows_UI_Xaml_Controls/ScrollViewerTests/ScrollViewer_TouchScrollBarDrag.xaml.cs
+++ b/src/SamplesApp/SamplesApp.Samples/Windows_UI_Xaml_Controls/ScrollViewerTests/ScrollViewer_TouchScrollBarDrag.xaml.cs
@@ -3,7 +3,7 @@ using System.Linq;
 using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Controls;
 using Microsoft.UI.Xaml.Controls.Primitives;
-using Microsoft.UI.Xaml.Media;
+using Uno.UI.Extensions;
 using Uno.UI.Samples.Controls;
 using Uno.UI.Toolkit;
 
@@ -45,13 +45,13 @@ namespace UITests.Windows_UI_Xaml_Controls.ScrollViewerTests
 
 		private void TryResolveScrollBars()
 		{
-			_verticalScrollBar ??= FindTemplateChild<ScrollBarControl>(Viewport, "VerticalScrollBar");
-			_horizontalScrollBar ??= FindTemplateChild<ScrollBarControl>(Viewport, "HorizontalScrollBar");
+			_verticalScrollBar ??= Viewport.FindFirstDescendant<ScrollBarControl>("VerticalScrollBar");
+			_horizontalScrollBar ??= Viewport.FindFirstDescendant<ScrollBarControl>("HorizontalScrollBar");
 
 			if (_verticalScrollBar is { } verticalScrollBar)
 			{
-				_verticalInteractiveRoot ??= FindTemplateChild<FrameworkElement>(verticalScrollBar, "VerticalRoot");
-				_verticalThumb ??= FindTemplateChild<FrameworkElement>(verticalScrollBar, "VerticalThumb");
+				_verticalInteractiveRoot ??= verticalScrollBar.FindFirstDescendant<FrameworkElement>("VerticalRoot");
+				_verticalThumb ??= verticalScrollBar.FindFirstDescendant<FrameworkElement>("VerticalThumb");
 			}
 
 			foreach (var scrollBar in GetScrollBars())
@@ -116,27 +116,6 @@ namespace UITests.Windows_UI_Xaml_Controls.ScrollViewerTests
 				+ $" — bars shown: h={Viewport.ComputedHorizontalScrollBarVisibility} v={Viewport.ComputedVerticalScrollBarVisibility}"
 				+ $" — {bars} — last scroll event: {_lastScroll}";
 			StandaloneValue.Text = $"{StandaloneScrollBar.Value:F0}";
-		}
-
-		private static T FindTemplateChild<T>(DependencyObject root, string name)
-			where T : FrameworkElement
-		{
-			var count = VisualTreeHelper.GetChildrenCount(root);
-			for (var i = 0; i < count; i++)
-			{
-				var child = VisualTreeHelper.GetChild(root, i);
-				if (child is T match && match.Name == name)
-				{
-					return match;
-				}
-
-				if (FindTemplateChild<T>(child, name) is { } descendant)
-				{
-					return descendant;
-				}
-			}
-
-			return null;
 		}
 	}
 }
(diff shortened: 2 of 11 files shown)


MartinZikmund:
@ajpinedam is it possible to touch drag the bar on WinUI? I think that it may actually just be intended this way. I think it may be possible to confirm via WinUI C++ sources, but if it would be Uno specific, it would be better as some extension in Uno Toolkit or such

ajpinedam:
@MartinZikmund 

You're right, and it's confirmed in the WinUI sources.

`ScrollBar_Partial.cpp` (`AttachEvents`, tag `winui3/release/1.6.5`) calls `put_IgnoreTouchInput(TRUE)` on both thumbs *and* all eight repeat buttons — ten parts, which Uno ported one-for-one. And `Thumb::put_IgnoreTouchInput` exists purely for that purpose (defaults to `FALSE`, `ScrollBar` is its only caller). So a touch drag on the thumb is intended WinUI behavior, not a bug. Same for `TouchIndicator`: WinUI drives it deliberately, so my second change diverged too.

The gap is still real on touch-only targets — a browser on a tablet has no mouse, so the scrollbar is the only affordance, and there's no WinUI equivalent — but that makes this an enhancement, not a fix, and it shouldn't change default behavior on every target.

So I'll take your suggestion: moving it out of core into an opt-in `Uno.UI.Toolkit` extension (an attached property on the ScrollViewer/ScrollBar). `Uno.UI` already has `InternalsVisibleTo("Uno.UI.Toolkit")`, so the Toolkit can clear `IgnoreTouchInput` on the thumbs, and `ScrollBar.IndicatorMode` is public — no core changes needed at all. I'll reshape this PR accordingly.

src/Uno.UI/UI/Xaml/Controls/ScrollBar/ScrollBar.uno.cs
```diff
@@ -7,6 +7,36 @@ namespace Microsoft.UI.Xaml.Controls.Primitives;
...
+
+	/// <summary>
+	/// Keeps the bar in its interactive state, so that its thumb can be dragged with a finger.
+	/// </summary>
+	/// <remarks>
+	/// Opted into through <see cref="Uno.UI.Toolkit.ScrollBarExtensions.IsTouchThumbDragEnabledProperty"/>.
+	/// WinUI has the bar parts ignore touch and shows the non-interactive touch indicator instead, which
+	/// stays the default here. A bar which opted in is treated as not conscious, which is the same path a
+	/// system with auto-hiding scroll bars turned off takes.
+	/// </remarks>
```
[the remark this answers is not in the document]
ajpinedam: Good catch — confirmed `Uno.UI` has no ProjectReference to `Uno.UI.Toolkit` (only `InternalsVisibleTo`, which does not help a cref), so this would indeed be CS1574. Changed to `<c>Uno.UI.Toolkit.ScrollBarExtensions.IsTouchThumbDragEnabled</c>` in dc66ff81fd.

src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml_Controls_Primitives/Given_ScrollBar.cs
```diff
@@ -29,5 +39,394 @@ public void When_Value_Changed()
...
+			finally
+			{
+				finger.Release();
+				TestServices.WindowHelper.WindowContent = null;
+			}
+		}
+
+		[TestMethod]
+		[Ignore("While the indicator is shown the interactive root takes the touch and the track ignores it, so a pan started on the strip does nothing.")]
+		public async Task When_Touch_Presses_Indicator_Strip_Then_Content_Still_Pans()
```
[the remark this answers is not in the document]
ajpinedam: Agreed — took option (a) in c2b33472aa: the test is removed, and the limitation it encoded is documented in `doc/articles/Uno.UI.Toolkit.md` (a finger pan which *starts* on a visible bar does not scroll the content, because hit-testing resolves into the ScrollBar and the content presenter is a sibling of it).

I considered keeping it as an active test asserting the current limited behaviour, but that would enshrine the limitation as expected and push whoever fixes the hit-testing toward "fixing" the test instead. The remaining Ignores in this class are all platform gates behind `#if !HAS_INPUT_INJECTOR` / `!UNO_HAS_MANAGED_SCROLL_PRESENTER`; the class now runs 10/10 with no skips on Skia Desktop.

src/Uno.UI/UI/Xaml/Controls/ScrollBar/ScrollBar.uno.cs
```diff
@@ -7,6 +7,36 @@ namespace Microsoft.UI.Xaml.Controls.Primitives;
...
+
+	/// <summary>
+	/// Keeps the bar in its interactive state, so that its thumb can be dragged with a finger.
+	/// </summary>
+	/// <remarks>
+	/// Opted into through <c>Uno.UI.Toolkit.ScrollBarExtensions.IsTouchThumbDragEnabled</c>.
+	/// WinUI has the bar parts ignore touch and shows the non-interactive touch indicator instead, which
+	/// stays the default here. A bar which opted in is treated as not conscious, which is the same path a
+	/// system with auto-hiding scroll bars turned off takes.
+	/// </remarks>
```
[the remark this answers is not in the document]
ajpinedam: You're right, and the description was staler than that — it still described the tree-walk mechanism, an `[Ignore]`d test that no longer exists, and draft status. Rewritten with an explicit "This does change core code, behind the opt-in" section.

For the record on why core is now touched: iPad validation showed the Toolkit-only approach could not work. `ScrollBar.AttachEvents` re-applies `IgnoreTouchInput = true` to its parts on every attach (template application and `Loaded`), and the `ScrollViewer` template defers its bars (`x:Load="False"`), so with `Auto` visibility a bar is realized after any external code has run. Separately, `IndicatorMode` is driven by the ScrollViewer's visual states at animation precedence, so a local value set from the Toolkit lost after every finger pan.

So the opt-in is a per-instance internal flag honoured in `AttachEvents` (the exact line WinUI hard-codes `TRUE`) and in `IsConscious()`. Default behaviour is unchanged on every target, pinned by `When_Touch_Drags_Thumb_Without_OptIn_Then_Value_Is_Unchanged` — with the core change neutered, 8 of 10 tests fail while that one still passes.

src/SamplesApp/SamplesApp.Samples/Windows_UI_Xaml_Controls/ScrollViewerTests/ScrollViewer_TouchScrollBarDrag.xaml.cs
```diff
@@ -0,0 +1,142 @@
...
+					+ $" draggable={_verticalInteractiveRoot?.IsHitTestVisible.ToString() ?? "?"}"
+					+ $" thumbOpacity={_verticalThumb?.Opacity.ToString("F1") ?? "?"}";
+
+			Status.Text = $"opt-in: {OptIn.IsChecked is true} — offset: {Viewport.HorizontalOffset:F0},{Viewport.VerticalOffset:F0}"
+				+ $" — bars shown: h={Viewport.ComputedHorizontalScrollBarVisibility} v={Viewport.ComputedVerticalScrollBarVisibility}"
+				+ $" — {bars} — last scroll event: {_lastScroll}";
+			StandaloneValue.Text = $"{StandaloneScrollBar.Value:F0}";
+		}
+
+		private static T FindTemplateChild<T>(DependencyObject root, string name)
```
Xiaoy312: just use FindFirstDescendant (see: src\Uno.UI\Extensions\ViewExtensions.visual-tree.cs), dont re-invent the wheel

doc/articles/Uno.UI.Toolkit.md
```diff
@@ -120,3 +120,30 @@ You can also apply `FromJson` directly inside `Page.DataContext`, either by refe
 ```
 
 If the JSON string is empty or invalid, the extension throws a `XamlParseException`, allowing issues to surface early during page initialization.
+
+## ScrollBar - dragging the thumb with a finger
+
+WinUI has the `ScrollBar` parts ignore touch on purpose: touch scrolling goes through direct manipulation, and the bar is only an indicator, so a finger on the thumb does nothing. A pointer reveals the interactive bar on hover, and touch has no hover. That is the default in Uno Platform as well.
+
+On a touch-only target - a browser on a tablet most notably - there is no pointer to reveal that bar with, so the user has nothing to drag. Opt into an interactive bar with `ScrollBarExtensions.IsTouchThumbDragEnabled`:
```
Xiaoy312, line 128: could we clarify here, how would this affect other platforms? weirdly or no-effect?
if this value is only supposed to used on specific target, we should show the user how to? (or straight up, gate it on the dp setter)

Xiaoy312 (approved):

src/SamplesApp/SamplesApp.Samples/Windows_UI_Xaml_Controls/ScrollViewerTests/ScrollViewer_TouchScrollBarDrag.xaml.cs
ajpinedam: Thanks

ajpinedam: Done in 53a6941095 — the local walker is gone, the sample now uses `Uno.UI.Extensions.ViewExtensions.FindFirstDescendant<T>(name)`.

doc/articles/Uno.UI.Toolkit.md
ajpinedam: Clarified in 18fd4d4ff2 — the doc now states the reach explicitly instead of only the Windows carve-out.

The short version: the opt-in is honoured wherever scroll bar input is handled by Uno's own managed `ScrollBar` — that's every Skia target (Desktop, WASM, Android, iOS), since the mechanism is entirely managed (`Thumb`/`RepeatButton.IgnoreTouchInput` filters `PointerDeviceType.Touch` in the shared mux code). On Windows it's inert: the Toolkit callback body is `#if HAS_UNO`, so the attached property just stores a value. On the native Android/iOS/WASM backends nothing gates it in core, but touch scrolling there is OS-driven direct manipulation and the combination was never validated — so the doc now calls those backends out as unsupported for this opt-in rather than promising behaviour.

On gating the DP setter: I'd rather not. The supported set is "all Skia targets", not one platform, so a gate would exist only to hard-deny the maintenance-only native backends — where default behaviour is already unchanged unless someone explicitly opts in (pinned by `When_Touch_Drags_Thumb_Without_OptIn_Then_Value_Is_Unchanged`). A silently no-op setter also makes "I set it and nothing happened" harder to debug, and documenting platform reach instead of gating setters follows the existing pattern (e.g. `ScrollViewer.ShouldFallBackToNativeScrollBars`). Happy to add the gate if you feel strongly, though.

ajpinedam:

Issue #24207, opened 2026-08-26
DevServer: share/dedup Hot Reload workspace per (solution + baseline + target framework)


dr1rrb:
_Follow-up from the investigation on #24151._

Today every app connection builds its own workspace + EnC baseline even for the same solution, so two connections = two full workspaces (~2x ~500 MB), no sharing.

A naive "one workspace per solution" is **not** currently possible:
- a late-joining app has no stored deltas to catch up with the current generation;
- the baseline is not necessarily the same — in particular an app for a **different target framework** has a different baseline, and EnC deltas must be **accumulated per target**.

**Proposal (long-term)**: key Hot Reload workspaces on **(solution + baseline identity + target framework)** and share when they match; accumulate the EnC delta stream **per target** so that additional or reconnecting apps of the same target attach to the existing workspace (replaying stored deltas to the current generation) instead of spawning a new one. This bounds memory for the common multi-app / reconnect case at the root, without breaking multi-targeting.

Issue #24206, opened 2026-08-26
DevServer: server-side WebSocket keepalive/timeout on /rc to reap half-open connections


dr1rrb:
_Follow-up from the investigation on #24151._

The server connection loop (`RemoteControlServer.HandleConnectionAsync`) only exits when `ReceiveAsync` returns null/throws. There is **no server-side WebSocket keepalive** (`app.UseWebSockets()` is configured without `KeepAliveInterval`). Keep-alive/ping messages exist in the protocol, but the server side does not use them to detect a dead peer.

**Consequence**: if an app dies without a clean TCP close (hard crash, network drop, or a client reconnect that abandons the old socket), the old `/rc` connection stays "open" server-side → its workspace + `DebuggingSession` (~500 MB) linger indefinitely, coexisting with the new connection's workspace. Measured: two concurrent connections = two workspaces (~1.35 GB), zero `Disposing connection`.

**Proposal**: enable server-side WebSocket keepalive with a timeout (or an application-level ping deadline using the existing KeepAlive messages), and dispose the connection scope (→ `EndSession` + workspace dispose) when a peer stops responding. This reaps orphaned workspaces with no active kill and no risk of a reconnect storm.

khaskoul:
Verified on `Uno.WinUI.DevServer` 6.8.0-dev.159 (Uno.Sdk 6.8.0-dev.39), Windows, `net10.0-desktop`. **Partially
  fixed** — the keepalive lands, the workspace reclaim does not.

  **Fixed:** `RemoteControlServer.HandleConnectionAsync` now bounds every receive with
  `receiveCts.CancelAfter(keepAliveTimeout)`, where `ResolveKeepAliveTimeout()` = `uno-devserver:keep-alive-timeout-ms`
  if set, else `KeepAliveMessage.Interval (30s) × 2.1` = 63 s.

  Tested against a real half-open socket (`NtSuspendProcess` on a connected app — TCP stays established, client stops
  pinging):

  - default: connection closed 63 s after the last ping (lifetime 94.6 s), then `Disposing connection` + `Unloading
  assembly context UnoApp1`
  - `--uno-devserver:keep-alive-timeout-ms=15000`: lifetime drops to 17–20 s

  Deadline is causal, config knob works, and the "zero `Disposing connection`" symptom is gone.

  **Not fixed:** the ticket also expects the scope disposal to give "`EndSession` + workspace dispose" and to "reap
[...]

Issue #24205, opened 2026-08-26
DevServer: cap concurrent app Hot Reload workspaces; disable HR on oldest beyond N without closing the socket


dr1rrb:
_Follow-up from the investigation on #24151._

Each `/rc` app connection gets its own DI scope → its own `RemoteControlServer` + MSBuild workspace + EnC `DebuggingSession` ≈ **+330–900 MB each**. Measured: host 108 MB → 439 MB after 1 app → 1348 MB after 2 apps, with nothing disposed. There is no cap.

If an agent restarts an app in a loop, or a crash/half-open socket leaves a previous connection alive, workspaces stack and memory explodes.

**Proposal**: cap the number of active **app** Hot Reload workspaces (e.g. ~5). Beyond the cap, disable HR for the oldest app(s) — i.e. dispose their workspace / `DebuggingSession` to reclaim the ~500 MB — **without closing the `/rc` socket** (closing it would trigger the client's auto-retry → reconnect storm). Re-arm the workspace if that app becomes active again. Workspace lifetime is independent from the socket, so this is feasible.

Related: #24151 companion issues on server-side keepalive (reaping half-open sockets) and per-solution/baseline/target workspace sharing.

khaskoul:
Verified fixed on `Uno.WinUI.DevServer` 6.8.0-dev.159 (Uno.Sdk 6.8.0-dev.39), Windows, desktop head.

  Launched up to 10 concurrent app connections against a standalone DevServer host:

  | apps | host WS | event |
  |---|---|---|
  | 1 | 565 MB | |
  | 5 | 2043 MB | ~370 MB/app |
  | 6 | 2402 MB | cap fired |
  | 8 | 2784 MB | cap fired |
  | 10 | 3462 MB | cap fired |

  `HotReloadWorkspaceRegistry` caps at 5 (`DefaultMaxConcurrentWorkspaces`), evicts the oldest, and logs:

  > Hot reload disabled for this connection: …reached its concurrent hot-reload workspace limit (5). The connection is
  kept alive; restart the app to get a fresh hot-reload session.

  All evicted apps stayed connected and running — the `/rc` socket is kept alive as proposed, no reconnect storm.

  Reclaim confirmed by `gcroot` on a process dump at 8 apps / 4 evictions: **5 rooted `DebuggingSession` (= the cap), 3
  unrooted** — exactly the apps evicted while connected. The workspace and EnC session are genuinely unrooted, nothing
  leaked.

Issue #24204, opened 2026-08-26
Uno source generators: migrate heavy legacy ISourceGenerators to cache-friendly incremental pipelines


dr1rrb:
_Follow-up from the investigation on #24151._

Uno's heavy generators (`XamlCodeGenerator`, `DependencyObjectGenerator`, `BindableTypeProviders*`, `RemoteControlGenerator`, `NativeCtorsGenerator`, …) are legacy `ISourceGenerator`, which re-run **fully** on every compilation. Under Hot Reload every edit triggers a full re-generation + recompile → ~70 MB transient allocation per edit on a real app for a ~150 KB delta (measured).

**Mini-repro** (edit of a file unrelated to the generator, per edit):

| Generator | re-runs / edit | alloc / edit |
|---|---|---|
| legacy `ISourceGenerator` | 1 (full) | ~0.6 MB |
| cache-friendly `IIncrementalGenerator` | 0 | ~0.03 MB (~20x less) |
| `IIncrementalGenerator` with `.Combine(CompilationProvider)` | 1 (full) | ~0.6 MB (== legacy) |

**Important caveat (perf != mechanical conversion).** The win only materializes if the pipeline is *cache-friendly*: equatable value models, no `Compilation`/`ISymbol` captured in the pipeline hot path. Roslyn docs "Authoring a cache-friendly generator" and dotnet/roslyn Discussion #67469 confirm `.Combine(CompilationProvider)` defeats caching (the Compilation changes on every keystroke → full re-run). Note `XamlCodeBehindGenerator` already uses `.Combine(context.CompilationProvider)` — converting it naively would gain nothing.

**Proposal**: rewrite the hot-path generators as incremental with equatable extracted models; audit existing incremental generators for the `CompilationProvider` trap.

PR #24209, opened 2026-08-26
fix(devserver): memory-conscious GC profile for the Hot Reload host

### LINKED ISSUE #24203: DevServer Hot Reload host: reconsider System.GC.Server (+ optional GCHeapHardLimitPercent)
_Follow-up from the investigation on #24151._

The Hot Reload host (`Uno.UI.RemoteControl.Host`) ships with `"System.GC.Server": true` in its runtimeconfig. For a long-lived background broker this hides collectable garbage until the OS commit limit is reached.

**Measured** (heavy Uno app, real EnC path): each hot-reload emit allocates ~70 MB of transient garbage for a ~150 KB metadata delta. With Server GC, gen2 stayed effectively frozen (0–1 collections) while the managed heap climbed monotonically by hundreds of MB. A forced gen2 GC reclaims the bulk instantly (the original reporter's gcdump reclaimed ~1 GB in one collection). So most of the growth is collectable garbage the GC feels no pressure to collect, and `DOTNET_gcServer=0` cannot override runtimeconfig — users currently have no opt-out.

**Proposal**
- Switch the host to Workstation GC (or at least make it configurable). Workstation GC applies pressure far sooner for this allocation pattern and bounds RSS.
- Optionally set a heap limit. An absolute `GCHeapHardLimit` is hard to choose (client-machine dependent, low value = risk); prefer **`GCHeapHardLimitPercent`** with a generous bound so the GC compacts under pressure instead of letting the OS terminate unrelated apps.

Fast mitigation, not a root-cause fix — see companion issues on generators and workspace/connection lifetime.

dr1rrb:
**GitHub Issue:** closes #24203

## PR Type:

🐞 Bugfix

## What changed? 🚀

The Hot Reload dev-server host (`Uno.UI.RemoteControl.Host`) is a long-running background daemon on the developer's machine, but it inherited the `Microsoft.NET.Sdk.Web` default GC profile (**Server GC**), which is throughput-oriented and holds onto committed memory. Under Hot Reload each emit allocates a large amount of *transient* garbage (tens of MB on a real app, for a ~150 KB delta); with Server GC the runtime feels no pressure to collect it, so the host's working set climbs monotonically into the multi-GB range over a working session (see #24203 / #24151).

This aligns the host's GC profile with a **memory-conscious background daemon**:

- `ServerGarbageCollection=false` + `ConcurrentGarbageCollection=true` → Workstation background GC: pressure is applied early, so the transient garbage is actually collected and the working set stays bounded.
- New `runtimeconfig.template.json`:
  - `System.GC.ConserveMemory=5` → keeps the heap compact and **decommits freed memory back to the OS** (so RSS actually goes down, not just the managed heap).
  - `System.GC.HeapHardLimitPercent=66` → relative last-resort ceiling (no absolute, machine-dependent value to pick).
  - `System.GC.HighMemoryPercent=70` → the GC becomes aggressive earlier when the **machine** is under memory pressure — a good citizen on a shared dev box.

Measured (synthetic LOH churn, 40 iterations, ~27 MB live heap in all three cases) — peak / final process working set:

| GC profile | RSS peak | RSS final |
|---|---|---|
| Server GC (previous default) | 670 MB | 534 MB |
| Workstation GC | 387 MB | 387 MB |
| Workstation + ConserveMemory | 317 MB | 317 MB (flat) |

The emitted `runtimeconfig.json` was verified to contain the five expected `System.GC.*` values.

This is the fast, low-risk mitigation track. The root-cause tracks (non-incremental generators, connection/workspace lifetime, per-solution/baseline/target workspace sharing) are tracked separately in #24204, #24205, #24206 and #24207.

## PR Checklist ✅

- [ ] 🧪 Added [Runtime tests, UI tests, or a manual test sample] — N/A: runtime GC configuration only; validated by measured working-set behavior and the emitted `runtimeconfig.json`.
- [ ] 📚 Docs have been added/updated following the documentation template — N/A.
- [ ] 🖼️ Validated PR `Screenshots Compare Test Run` results — N/A (no UI change).

### DIFF
--- a/src/Uno.UI.RemoteControl.Host/Uno.UI.RemoteControl.Host.csproj
+++ b/src/Uno.UI.RemoteControl.Host/Uno.UI.RemoteControl.Host.csproj
@@ -8,6 +8,12 @@
 		<Nullable>enable</Nullable>
 		<NoWarn>$(NoWarn);VSTHRD200</NoWarn>
 		<DefineConstants>$(DefineConstants);IS_DEVSERVER</DefineConstants>
+
+		<!-- The dev-server is a long-running background daemon on the developer's machine.
+		     The default Web SDK profile (Server GC) is throughput-oriented and retains committed
+		     memory; switch to a memory-conscious Workstation profile instead. See #24203. -->
+		<ServerGarbageCollection>false</ServerGarbageCollection>
+		<ConcurrentGarbageCollection>true</ConcurrentGarbageCollection>
 	</PropertyGroup>
 
 	<Import Project="../targetframework-override-noplatform.props" />
@@ -98,5 +104,13 @@
 
 	<Target Name="GetTargetPath" />
 	<Target Name="GetCopyToPublishDirectoryItems" />
+
+
+	<ItemGroup>
+		<!-- runtimeconfig.template.json is a build-time input merged into the generated
+		     Uno.UI.RemoteControl.Host.runtimeconfig.json; it must not be copied to output
+		     nor packed. See #24203. -->
+		<Content Remove="runtimeconfig.template.json" />
+	</ItemGroup>
 
 </Project>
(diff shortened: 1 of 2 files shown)


kazo0 (approved):

dr1rrb:

PR #24216, opened 2026-08-26
fix(wasm): Redispatch resize once viewport settles after rotation

ajpinedam:
**GitHub Issue:** n/a — customer-reported, tracked internally.

## PR Type:

🐞 Bugfix

## What changed? 🚀

**Current behavior:** in a WebAssembly app using the Skia renderer, rotating the device between portrait and landscape can leave the app rendered at the wrong dimensions (controls squished / at the wrong scale). Some mobile browsers — observed with Chrome on iPad, while Safari is fine — fire `resize` / `orientationchange` **before** the layout viewport (`document.documentElement`) has actually reflowed to the new size, and never fire a corrective `resize` afterward. Since `WebAssemblyWindowWrapper` sizes the window (and thus the Skia surface) from `documentElement.getBoundingClientRect()` inside that early `resize` handler, it keeps the stale pre-rotation dimensions until some other event forces a re-measure. Because rendering and input mapping both use the stale size, taps also land offset from the visuals.

**This PR** installs a `ResizeObserver` on `document.documentElement` — the exact element `resize()` measures from — in the Skia WASM host. `ResizeObserver` fires after layout, i.e. precisely when the element has actually settled to its post-rotation size. When the observed size differs from the last size the host measured, it re-dispatches a synthetic `window.resize`, restoring the corrective event the browser should have fired; the existing resize path then re-measures against the settled dimensions. The host records the last measured size on every `resize()`, so when a native resize already handled the change the observer callback no-ops — no redundant dispatches, no loops (the observer only fires on actual size changes). Browsers without `ResizeObserver` keep the previous behavior.

The fix is host-level and benefits app code too: application-level `window.resize` listeners were equally stale, and now receive the corrective event. Dispatching the global event (rather than re-measuring internally) is deliberate — `BrowserRenderer` sizes the canvas from its own `window.resize` listener, so an internal-only re-measure would fix the window bounds but leave the canvas stale.

Design notes / scope:
- The `window.resize` listener stays the primary size source; the observer is only the corrective for the mis-ordered case. The native listener uniquely covers DPR-only resizes (browser zoom fires `resize` without changing `documentElement`'s CSS size, which a `ResizeObserver` cannot see).
- No feedback-loop risk: `uno.css` pins `html`/`body` and the canvas to `position: fixed; width/height: 100%; overflow: hidden`, so the viewport drives `documentElement` — never the canvas — and `resize()` records the last measured size, making the observer a no-op whenever a native resize already handled the change.
- The native (non-Skia) WebAssembly target has the same measure-on-resize pattern in `WindowManager.ts` and may exhibit the same staleness, but per the Skia-first policy it is maintenance-only and intentionally out of scope here.
- DPR-only staleness (`RasterizationScale` on zoom/monitor moves) is a pre-existing, unrelated behavior and is unchanged.

## PR Checklist ✅

  - The race requires physically rotating a device and a browser that mis-orders the events, so it is not runtime-testable. Added the manual sample `Wasm_ViewportOrientationTelemetry` (Wasm category), which live-compares `XamlRoot.Size` against `documentElement.getBoundingClientRect()` with a STALE/OK banner and raw `resize`/`orientationchange` counters, so the symptom and the fix are directly observable on a device.
  - Manual validation on a physical iPad in Chrome (the affected combination): without the fix the stale size reproduces after rotation and sticks; with the fix (validated both as the patched host in the published `SamplesApp.Skia.WebAssembly.Browser` served over LAN, and as the equivalent logic A/B-toggled in a telemetry repro app) every rotation converges to the settled viewport size. Safari behavior is unchanged.

Credit: the root-cause analysis and the proposed `ResizeObserver` approach come from the reporting customer's investigation.

### DIFF
--- a/src/SamplesApp/SamplesApp.Samples/Wasm/Wasm_ViewportOrientationTelemetry.xaml.cs
+++ b/src/SamplesApp/SamplesApp.Samples/Wasm/Wasm_ViewportOrientationTelemetry.xaml.cs
@@ -108,8 +108,7 @@ public sealed partial class Wasm_ViewportOrientationTelemetry : Page
 					EventCountsText.Text = $"{parts[2]} resize / {parts[3]} orientationchange";
 
 					var stale = Math.Abs(unoSize.Width - docWidth) > 2 || Math.Abs(unoSize.Height - docHeight) > 2;
-					StatusText.Text = stale ? "STALE" : "OK";
-					StatusBorder.Background = new SolidColorBrush(stale ? Microsoft.UI.Colors.IndianRed : Microsoft.UI.Colors.MediumSeaGreen);
+					SetStatus(stale ? "STALE" : "OK", stale ? Microsoft.UI.Colors.IndianRed : Microsoft.UI.Colors.MediumSeaGreen);
 					return;
 				}
 
@@ -125,8 +124,7 @@ public sealed partial class Wasm_ViewportOrientationTelemetry : Page
 		{
 			DocRectText.Text = $"Telemetry failed: {_telemetryError}";
 			EventCountsText.Text = "-";
-			StatusText.Text = "ERROR";
-			StatusBorder.Background = new SolidColorBrush(Microsoft.UI.Colors.IndianRed);
+			SetStatus("ERROR", Microsoft.UI.Colors.IndianRed);
 			return;
 		}
 #endif
@@ -144,6 +142,16 @@ public sealed partial class Wasm_ViewportOrientationTelemetry : Page
 		_telemetryError = error;
 		_jsGateway?.Dispose();
 		_jsGateway = null;
+	}
+
+	// Allocate a brush only when the status actually transitions (the timer ticks at 10 Hz).
+	private void SetStatus(string status, Windows.UI.Color color)
+	{
+		if (StatusText.Text != status)
+		{
+			StatusText.Text = status;
+			StatusBorder.Background = new SolidColorBrush(color);
+		}
 	}
 #endif
 }
--- a/src/SamplesApp/SamplesApp.Samples/Wasm/Wasm_ViewportOrientationTelemetry.xaml.cs
+++ b/src/SamplesApp/SamplesApp.Samples/Wasm/Wasm_ViewportOrientationTelemetry.xaml.cs
@@ -133,7 +133,7 @@ public sealed partial class Wasm_ViewportOrientationTelemetry : Page
 
 		DocRectText.Text = "DOM telemetry is available only when running in a browser";
 		EventCountsText.Text = "-";
-		StatusText.Text = "N/A";
+		SetStatus("N/A");
 	}
 
 #if __CROSSRUNTIME__
--- a/src/SamplesApp/SamplesApp.Samples/Wasm/Wasm_ViewportOrientationTelemetry.xaml.cs
+++ b/src/SamplesApp/SamplesApp.Samples/Wasm/Wasm_ViewportOrientationTelemetry.xaml.cs
@@ -102,12 +102,12 @@ public sealed partial class Wasm_ViewportOrientationTelemetry : Page
 			try
 			{
 				var parts = _jsGateway.ExecuteJavascript(PollScript).Split(',');
-				if (parts.Length >= 4
-					&& double.TryParse(parts[0], NumberStyles.Float, CultureInfo.InvariantCulture, out var docWidth)
-					&& double.TryParse(parts[1], NumberStyles.Float, CultureInfo.InvariantCulture, out var docHeight))
+				if (parts is [var width, var height, var resize, var orientation]
+					&& double.TryParse(width, NumberStyles.Float, CultureInfo.InvariantCulture, out var docWidth)
+					&& double.TryParse(height, NumberStyles.Float, CultureInfo.InvariantCulture, out var docHeight))
 				{
 					DocRectText.Text = $"{docWidth:0} x {docHeight:0}";
-					EventCountsText.Text = $"{parts[2]} resize / {parts[3]} orientationchange";
+					EventCountsText.Text = $"{resize} resize / {orientation} orientationchange";
 
 					var stale = Math.Abs(unoSize.Width - docWidth) > 2 || Math.Abs(unoSize.Height - docHeight) > 2;
 					SetStatus(stale ? "STALE" : "OK", stale ? Microsoft.UI.Colors.IndianRed : Microsoft.UI.Colors.MediumSeaGreen);
--- a/src/SamplesApp/SamplesApp.Samples/Wasm/Wasm_ViewportOrientationTelemetry.xaml.cs
+++ b/src/SamplesApp/SamplesApp.Samples/Wasm/Wasm_ViewportOrientationTelemetry.xaml.cs
@@ -1,0 +1,110 @@
+#nullable enable
+
+using System;
+using System.Globalization;
+using Microsoft.UI.Xaml;
+using Microsoft.UI.Xaml.Controls;
+using Microsoft.UI.Xaml.Media;
+using Uno.UI.Samples.Controls;
+#if __CROSSRUNTIME__
+using Uno.UI.NativeElementHosting;
+#endif
+
+namespace UITests.Shared.Wasm;
+
+[Sample("Wasm",
+	Name = nameof(Wasm_ViewportOrientationTelemetry),
+	Description = "Compares the Uno window size against the browser's documentElement size after device rotation. The banner turns STALE if the Uno-measured size lags the settled viewport size.",
+	IsManualTest = true,
+	IgnoreInSnapshotTests = true)]
+public sealed partial class Wasm_ViewportOrientationTelemetry : Page
+{
+	private const string InstallCountersScript = @"
+(function() {
+	let t = globalThis.__unoViewportTelemetry;
+	if (!t) {
+		t = { resize: 0, orientation: 0 };
+		globalThis.__unoViewportTelemetry = t;
+		window.addEventListener(""resize"", function() { t.resize++; });
+		window.addEventListener(""orientationchange"", function() { t.orientation++; });
+	}
+	return ""ok"";
+})();";
+
+	private const string PollScript = @"
+(function() {
+	const r = document.documentElement.getBoundingClientRect();
+	const t = globalThis.__unoViewportTelemetry || { resize: 0, orientation: 0 };
+	return Math.round(r.width) + "","" + Math.round(r.height) + "","" + t.resize + "","" + t.orientation;
+})();";
+
+	private readonly DispatcherTimer _pollTimer;
+
+#if __CROSSRUNTIME__
+	private BrowserHtmlElement? _jsGateway;
+#endif
+
+	public Wasm_ViewportOrientationTelemetry()
+	{
+		this.InitializeComponent();
+
+		// Poll on an interval rather than only on SizeChanged so a stuck stale
+		// size is visibly stuck instead of sampled once right after rotation.
+		_pollTimer = new DispatcherTimer { Interval = TimeSpan.FromMilliseconds(100) };
+		_pollTimer.Tick += (_, _) => UpdateTelemetry();
+
+		Loaded += OnPageLoaded;
+		Unloaded += OnPageUnloaded;
+	}
+
+	private void OnPageLoaded(object sender, RoutedEventArgs e)
+	{
+#if __CROSSRUNTIME__
+		if (OperatingSystem.IsBrowser())
+		{
+			_jsGateway = BrowserHtmlElement.CreateHtmlElement("div");
+			_jsGateway.ExecuteJavascript(InstallCountersScript);
+		}
+#endif
+
+		_pollTimer.Start();
+		UpdateTelemetry();
+	}
+
+	private void OnPageUnloaded(object sender, RoutedEventArgs e)
+	{
+		_pollTimer.Stop();
+
+#if __CROSSRUNTIME__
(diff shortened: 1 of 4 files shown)


src/Uno.UI.Runtime.Skia.WebAssembly.Browser/UI/Xaml/Window/WebAssemblyWindowWrapper.cs
```diff
@@ -86,14 +86,22 @@ internal void RaiseNativeSizeChanged(Size newWindowSize)
...
+				Console.WriteLine($"RaiseNativeSizeChanged target for {instance} does not exist");
+			}
 		}
-		else
+		catch (Exception e)
 		{
-			Console.WriteLine($"RaiseNativeSizeChanged target for {instance} does not exist");
+			// A managed exception must not cross back into the JS DOM-event callback.
+			Application.Current.RaiseRecoverableUnhandledException(e);
 		}
```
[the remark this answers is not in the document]
ajpinedam: Valid catch — confirmed: `WebAssemblyWindowWrapper.Initialize()` is awaited in `WebAssemblyBrowserHost.InitializeAsync()` and the JS side runs an initial `resize()` there, all before `Application.Start` in `RunLoop()`, so `OnResize` always fires at least once pre-startup. Fixed in 3519548930 by switching to the null-safe `RaiseRecoverableUnhandledExceptionOrLog` helper (same pattern as `MacOSWindowHost`), and applied the same fix to the pre-existing `OnViewportOcclusionChanged` catch which had the identical latent issue.

src/SamplesApp/SamplesApp.Samples/Wasm/Wasm_ViewportOrientationTelemetry.xaml.cs
```diff
@@ -0,0 +1,149 @@
...
+					&& double.TryParse(parts[0], NumberStyles.Float, CultureInfo.InvariantCulture, out var docWidth)
+					&& double.TryParse(parts[1], NumberStyles.Float, CultureInfo.InvariantCulture, out var docHeight))
+				{
+					DocRectText.Text = $"{docWidth:0} x {docHeight:0}";
+					EventCountsText.Text = $"{parts[2]} resize / {parts[3]} orientationchange";
+
+					var stale = Math.Abs(unoSize.Width - docWidth) > 2 || Math.Abs(unoSize.Height - docHeight) > 2;
+					StatusText.Text = stale ? "STALE" : "OK";
+					StatusBorder.Background = new SolidColorBrush(stale ? Microsoft.UI.Colors.IndianRed : Microsoft.UI.Colors.MediumSeaGreen);
+					return;
```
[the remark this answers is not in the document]
ajpinedam: Fixed in ba93a8a95b: status text/brush updates now go through a `SetStatus` helper that only assigns (and only allocates the brush) when the status actually transitions, so the 10 Hz tick no longer allocates in steady state.

src/SamplesApp/SamplesApp.Samples/Wasm/Wasm_ViewportOrientationTelemetry.xaml.cs
```diff
@@ -0,0 +1,157 @@
...
+namespace UITests.Shared.Wasm;
+
+[Sample("Wasm",
+	Name = nameof(Wasm_ViewportOrientationTelemetry),
+	Description = "Compares the Uno Platform window size against the browser's documentElement size after device rotation. The banner turns STALE if the app-measured size lags the settled viewport size.",
+	IsManualTest = true,
+	IgnoreInSnapshotTests = true)]
+public sealed partial class Wasm_ViewportOrientationTelemetry : Page
+{
+	private const string InstallCountersScript = @"
```
Xiaoy312: use the """ raw string

src/SamplesApp/SamplesApp.Samples/Wasm/Wasm_ViewportOrientationTelemetry.xaml.cs
```diff
@@ -0,0 +1,157 @@
...
+		var unoSize = XamlRoot?.Size ?? default;
+		UnoBoundsText.Text = $"{unoSize.Width:0} x {unoSize.Height:0}";
+
+#if __CROSSRUNTIME__
+		if (_jsGateway is not null)
+		{
+			try
+			{
+				var parts = _jsGateway.ExecuteJavascript(PollScript).Split(',');
+				if (parts.Length >= 4
```
Xiaoy312: why greater-than-or-equal, it should be just is-4?
```suggestion
				if (parts.Length is 4
```

Xiaoy312: can also give name of the args at same time too:
```suggestion
				if (parts is [var width, var height, var resize, var orientation]
```

src/SamplesApp/SamplesApp.Samples/Wasm/Wasm_ViewportOrientationTelemetry.xaml.cs
```diff
@@ -0,0 +1,157 @@
...
+				{
+					DocRectText.Text = $"{docWidth:0} x {docHeight:0}";
+					EventCountsText.Text = $"{parts[2]} resize / {parts[3]} orientationchange";
+
+					var stale = Math.Abs(unoSize.Width - docWidth) > 2 || Math.Abs(unoSize.Height - docHeight) > 2;
+					SetStatus(stale ? "STALE" : "OK", stale ? Microsoft.UI.Colors.IndianRed : Microsoft.UI.Colors.MediumSeaGreen);
+					return;
+				}
+
+				DropGateway("unexpected telemetry format");
```
Xiaoy312:
```suggestion
				DropGateway($"unexpected telemetry format: {string.Join(',', parts)}");
```
better if we keep the result without splitting in an intermediary var

src/SamplesApp/SamplesApp.Samples/Wasm/Wasm_ViewportOrientationTelemetry.xaml.cs
```diff
@@ -0,0 +1,157 @@
...
+	// diagnostic must diagnose itself.
+	private void DropGateway(string error)
+	{
+		_telemetryError = error;
+		_jsGateway?.Dispose();
+		_jsGateway = null;
+	}
+
+	// Allocate a brush only when the status actually transitions (the timer ticks at 10 Hz).
+	private void SetStatus(string status, Windows.UI.Color color)
```
Xiaoy312:
```suggestion
	private void SetStatus(string status, bool isError = false)
```

src/SamplesApp/SamplesApp.Samples/Wasm/Wasm_ViewportOrientationTelemetry.xaml.cs
```diff
@@ -0,0 +1,157 @@
...
+			DocRectText.Text = $"Telemetry failed: {_telemetryError}";
+			EventCountsText.Text = "-";
+			SetStatus("ERROR", Microsoft.UI.Colors.IndianRed);
+			return;
+		}
+#endif
+
+		DocRectText.Text = "DOM telemetry is available only when running in a browser";
+		EventCountsText.Text = "-";
+		StatusText.Text = "N/A";
```
Xiaoy312: if you have a helper method for that, always use it:
```suggestion
		SetStatus("N/A", isError: true/false); // depends how you feel
```

src/SamplesApp/SamplesApp.Samples/Wasm/Wasm_ViewportOrientationTelemetry.xaml.cs
Xiaoy312: you can move this out of #if too

src/SamplesApp/SamplesApp.Samples/Wasm/Wasm_ViewportOrientationTelemetry.xaml.cs
```diff
@@ -0,0 +1,157 @@
...
+#endif
+	}
+
+	private void UpdateTelemetry()
+	{
+		var unoSize = XamlRoot?.Size ?? default;
+		UnoBoundsText.Text = $"{unoSize.Width:0} x {unoSize.Height:0}";
+
+#if __CROSSRUNTIME__
+		if (_jsGateway is not null)
```
Xiaoy312, line 104: the flow could be improved..:
```
if (a)
	try
		if (b) { ... return; }
		/*else fall-through aka c=true*/
	catch /*fall-through aka c=true*/

if(c) { ... return; }

...
```

Xiaoy312 (approved):

src/SamplesApp/SamplesApp.Samples/Wasm/Wasm_ViewportOrientationTelemetry.xaml.cs
ajpinedam: Fixed in c6ecb4b732 — both scripts converted to `"""` raw string literals.

src/SamplesApp/SamplesApp.Samples/Wasm/Wasm_ViewportOrientationTelemetry.xaml.cs
ajpinedam: Fixed in b009a278ec — went straight to the list pattern with named vars: `parts is [var width, var height, var resize, var orientation]`, and the `parts[n]` indexers now use the names.

src/SamplesApp/SamplesApp.Samples/Wasm/Wasm_ViewportOrientationTelemetry.xaml.cs
ajpinedam: Fixed in 425d7676c0 — the unsplit result stays in a `result` var (split moved into the pattern match) and the error now reports the raw payload.

src/SamplesApp/SamplesApp.Samples/Wasm/Wasm_ViewportOrientationTelemetry.xaml.cs
ajpinedam: Fixed in 35d56d484e — signature is now `SetStatus(string status, bool isError = false)` with the brush color computed inside, and the method moved out of `#if __CROSSRUNTIME__` (only `DropGateway` stays gated, since it touches the gated fields).

src/SamplesApp/SamplesApp.Samples/Wasm/Wasm_ViewportOrientationTelemetry.xaml.cs
ajpinedam: Fixed in 6d8daa5dd8 — now `SetStatus("N/A")`. Went with the default `isError: false`: N/A on a non-browser head is expected, not a failure, so it keeps the banner's default green.

src/SamplesApp/SamplesApp.Samples/Wasm/Wasm_ViewportOrientationTelemetry.xaml.cs
ajpinedam: After the other threads landed, the method now matches this shape exactly: `if (_jsGateway is not null)` → `try { if (parse-ok) { …; return; } DropGateway(…); } catch { DropGateway(…); }` → `if (_telemetryError is not null) { …; return; }` → N/A fallback. The `DropGateway` calls are the "fall-through aka c=true" steps — they set `_telemetryError` so the same tick falls into the error block.

src/SamplesApp/SamplesApp.Samples/Wasm/Wasm_ViewportOrientationTelemetry.xaml.cs
```diff
@@ -0,0 +1,159 @@
...
+#endif
+
+namespace UITests.Shared.Wasm;
+
+[Sample("Wasm",
+	Name = nameof(Wasm_ViewportOrientationTelemetry),
+	Description = "Compares the Uno Platform window size against the browser's documentElement size after device rotation. The banner turns STALE if the app-measured size lags the settled viewport size.",
+	IsManualTest = true,
+	IgnoreInSnapshotTests = true)]
+public sealed partial class Wasm_ViewportOrientationTelemetry : Page
```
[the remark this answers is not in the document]
ajpinedam: Fixed in dbdbdde8c8 — gated with `#if __CROSSRUNTIME__` rather than `#if __WASM__`: this sample targets the Skia browser head (`SamplesApp.Skia.WebAssembly.Browser`), which compiles samples with `__SKIA__`/`__CROSSRUNTIME__` only — `__WASM__` is defined solely on the native-DOM head, so that gate would hide the sample exactly where this PR needs it. `Wasm_CustomEvent` uses `__WASM__` because it is a native-DOM sample (`HtmlId`). With `__CROSSRUNTIME__` the sample stays registered on both WASM heads and disappears from the Windows/Android/iOS heads; Skia desktop still lists it with the N/A message since browser-vs-desktop there is runtime-only (`OperatingSystem.IsBrowser()`).

ajpinedam:

src/SamplesApp/SamplesApp.Samples/Wasm/Wasm_ViewportOrientationTelemetry.xaml.cs
Xiaoy312: perhaps i wasnt clear... that was the messy current form
didnt think of a better one yet, but it could have less nested and jumps/fall-through
oh well~

PR #24228, opened 2026-08-27
perf(textblock): Skip the arrange-time text re-shape

### LINKED ISSUE #24219: [Skia] TextBlock shapes its text twice per layout pass

`TextBlock.ArrangeOverride` re-parses and re-shapes its entire text whenever the arrange constraint differs from the measure constraint by **full `Size` equality**:

```csharp
if (_lastParsedTextCreationValues.availableSize != availableSizeWithoutPadding || ...alignment...)
```

A panel virtually always measures a child with an unconstrained (or generous) height and then arranges it at its desired height, so the `Height` component differs on essentially every pass — even though the *width*, which is what drives line breaking, is identical. Every `TextBlock` therefore runs ICU boundary analysis, HarfBuzz shaping and line breaking **twice** per layout pass.

An allocation profile of a layout pass shows two call sites of almost equal weight: `MeasureOverride → ParseText` (29.0 MB in the top stack) and `ArrangeOverride → ParseText` (28.2 MB).

### Why is this needed? 🤔

Height is not free to ignore — `UnicodeText` reads `availableSize.Height` to decide `isEarlyLastLine`, which drops the lines that do not fit. But it only matters when it actually truncated something, and **WinUI makes exactly that distinction**. `BlockNode::CanBypassMeasure` (`dxaml/xcp/core/text/BlockLayout/BlockNode.cpp`) requires the width to match, and then:

```cpp
if (m_pBreak != NULL)
{
    // If there is a break for this paragraph, then height constraint needs to be
    // the same to break at the same place.
    bypass = IsCloseReal(availableSize.height, m_prevAvailableSize.height);
}
else
{
    // If there was no break, we can bypass as long as the desired height will fit
    // in the available space.
    bypass = (m_desiredSize.height <= availableSize.height);
}
```

Uno's unconditional full-`Size` comparison is strictly more conservative than WinUI's, at the cost of a full re-shape.

### Measured impact 📊

|---|---|---:|---:|---:|
| `layout.full-pass` | time / pass | 32.41 ms | 19.03 ms | **−41.3 %** |
| `layout.full-pass` | alloc / pass | 8,194,446 B | 4,143,044 B | **−49.4 %** |
| `layout.full-pass` | alloc / element | 2,635.7 B | 1,332.6 B | **−49.4 %** |
| `layout.resize-pass` | time / pass | 32.93 ms | 18.57 ms | **−43.6 %** |
| `layout.resize-pass` | alloc / pass | 8,191,210 B | 4,137,379 B | **−49.5 %** |
| *control* `tree.build` | time | 33.00 ms | 33.32 ms | +1.0 % (noise) |
| *control* `dp.set-get` | time | 0.2413 ms | 0.2434 ms | +0.9 % (noise) |

A 3,109-element tree (nested `Grid`/`StackPanel`, `Border` at every level, `TextBlock` leaves), every element dirtied per pass. 3 runs each, median. Raw `layout.full-pass` time runs: 32.41 / 33.45 / 31.45 → 19.42 / 19.03 / 18.86 ms. Measured against the branch base, with no other fix applied.

### Renderer 🎨

- [x] Skia

### For which platforms 📱💻🖥️

All platforms 🌍

MartinZikmund:
**GitHub Issue:** closes #24219

## PR Type:

## What changed? 🚀

`TextBlock.ArrangeOverride` re-parsed and re-shaped its whole text whenever the arrange constraint differed from the measure constraint by **full `Size` equality**. A panel measures a child with an unconstrained height and arranges it at its desired height, so the height differs on nearly every pass — and the text was shaped twice per layout pass, through ICU boundary analysis, HarfBuzz shaping and line breaking.

Height only matters when it actually dropped a line, so `UnicodeText` now records that (`IsHeightTruncated`) and `TextBlock` applies WinUI's rule from `BlockNode::CanBypassMeasure`: the width and adjusted alignment must match, but the height only has to match when the previous layout was truncated by it — otherwise it is enough that the text still fits.

The bypass condition is exact rather than approximate: for every line *i*, `totalHeight_i + nextLineHeight_i` is the prefix sum through *i+1*, which is `<= outSize.Height`, so `arrangeHeight >= outSize.Height` implies no line can be dropped that was not dropped before. The boundary case — a panel arranging at exactly the desired height — lands on the `<=` side.

### Measured impact 📊

|---|---|---:|---:|---:|
| `layout.full-pass` | time / pass | 32.41 ms | 19.03 ms | **−41.3 %** |
| `layout.full-pass` | alloc / pass | 8,194,446 B | 4,143,044 B | **−49.4 %** |
| `layout.full-pass` | alloc / element | 2,635.7 B | 1,332.6 B | **−49.4 %** |
| `layout.resize-pass` | time / pass | 32.93 ms | 18.57 ms | **−43.6 %** |
| `layout.resize-pass` | alloc / pass | 8,191,210 B | 4,137,379 B | **−49.5 %** |
| *control* `tree.build` | time | 33.00 ms | 33.32 ms | +1.0 % (noise) |
| *control* `dp.set-get` | time | 0.2413 ms | 0.2434 ms | +0.9 % (noise) |

A 3,109-element tree (nested `Grid`/`StackPanel`, `Border` at every level, `TextBlock` leaves), every element dirtied per pass. 3 runs each, median. Raw `layout.full-pass` time runs: 32.41 / 33.45 / 31.45 → 19.42 / 19.03 / 18.86 ms. Measured against the branch base, with no other fix applied.

### Validation ✅

Runtime tests, Skia Desktop, `Given_TextBlock*` + `Given_TextBox*`: **363 run, 358 passed, 7 skipped, 5 failed**. All 5 failures reproduce **unchanged on the baseline** with the change reverted and the app rebuilt, so none is caused by this PR (`When_Inlines_Transitively_Change`, `When_IsTextSelectionEnabled_CRLF`, `When_Caret_Line_Straddles_Viewport_Edge_Grippers_Are_Hidden`, `When_Copy_Paste` — clipboard, flaky in both directions — and `When_Multiline_Pointer_TripleTap_With_Wrapping`).

## PR Checklist ✅

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/TextBlock/TextBlock.cs
@@ -1373,7 +1373,7 @@ namespace Microsoft.UI.Xaml.Controls
 		private bool _grippersShown;
 		private Microsoft.UI.Input.PointerPoint? _lastPointerDownPoint;
 
-		private (Size availableSize, Size outSize, TextAlignment? alignment) _lastParsedTextCreationValues = (Size.Empty, Size.Empty, TextAlignment.Left);
+		private (Size availableSize, Size outSize, TextAlignment? alignment, bool heightTruncated) _lastParsedTextCreationValues = (Size.Empty, Size.Empty, TextAlignment.Left, false);
 		internal IParsedText ParsedText { get; private set; } = Microsoft.UI.Xaml.Documents.ParsedText.Empty;
 
 		internal event EventHandler? DrawingFinished;
@@ -1494,7 +1494,7 @@ namespace Microsoft.UI.Xaml.Controls
 				size.Width += CaretThickness;
 			}
 
-			_lastParsedTextCreationValues = (availableSizeWithoutPadding, size, adjustedTextAlignment);
+			_lastParsedTextCreationValues = (availableSizeWithoutPadding, size, adjustedTextAlignment, ret.IsHeightTruncated);
 			return ret;
 		}
 
@@ -1561,12 +1561,13 @@ namespace Microsoft.UI.Xaml.Controls
 			var padding = Padding;
 			var availableSizeWithoutPadding = finalSize.Subtract(padding);
 
-			// There's no reason to re-parse the text if the available size hasn't changed since the last measure/arrange.
-			// Note that MeasureOverride doesn't have these checks. If something in the text block has changed that would
-			// require a re-parse, the ParseText call during the measure pass will catch it. There are no changes that
-			// would require a re-parse that would invalidate arrange but not measure, except TextAlignment, which we explicitly check.
+			// There's no reason to re-parse the text if the constraint that shaped it hasn't changed since the last
+			// measure/arrange. Note that MeasureOverride doesn't have these checks. If something in the text block has
+			// changed that would require a re-parse, the ParseText call during the measure pass will catch it. There are
+			// no changes that would require a re-parse that would invalidate arrange but not measure, except
+			// TextAlignment, which we explicitly check.
 			var arrangedSize = _lastParsedTextCreationValues.outSize;
-			if (_lastParsedTextCreationValues.availableSize != availableSizeWithoutPadding || _lastParsedTextCreationValues.alignment != GetAdjustedTextAlignment())
+			if (NeedsReparseForArrange(availableSizeWithoutPadding))
 			{
 				ParsedText = ParseText(availableSizeWithoutPadding, out arrangedSize);
 			}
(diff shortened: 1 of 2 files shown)


morning4coffe-dev (approved):
Code review assessment: the height-reuse rule matches the relevant WinUI BlockNode contract: a broken/truncated layout requires the same height; otherwise reuse is allowed while desired height fits. Width/alignment still invalidate the arrange reuse, and the measure path continues to refresh other shaping inputs. I also checked that UnicodeText's retained available size is consumed for width, not stale height. Reference: https://github.com/microsoft/microsoft-ui-xaml/blob/439bda90582f7892ca595a193f55dbfc6ae2010c/dxaml/xcp/core/text/BlockLayout/BlockNode.cpp#L391-L427 . No blocking source issue found; automated-check status is separate.

PR #24234, opened 2026-08-27
perf(uno-ui): Walk the resource scope without an iterator

### LINKED ISSUE #24225: DependencyObject.GetResourceDictionaries allocates an iterator on every call

`GetResourceDictionaries` is a `yield return` iterator, so **every call allocates a state machine**. It is called once per theme reference per element entering the tree, on top of every implicit-style lookup (`GetImplicitStyle`) and every `{StaticResource}` visual-tree retrieval (`ResourceResolver.TryVisualTreeRetrieval`).

While a virtualised `ListView` materialises containers during a scroll, `<GetResourceDictionaries>d__219` is the **single largest allocator** in the profile at 6.3 %, reached through `FrameworkElement.OnFwEltLoading → UpdateResourceBindings → UpdateAllThemeReferences → UpdateThemeReference`.

### Why is this needed? 🤔

Returning a `readonly struct` enumerable with a struct enumerator removes the allocation while leaving the walk **unchanged** — still lazy, still yielding the containing dictionary first, then non-empty ancestor `Resources` nearest-first, then `Application.Resources`, and still short-circuiting at the caller's first match. It still implements `IEnumerable<ResourceDictionary>` so the two cold `.ToArray()` callers keep working, while `foreach` binds to the struct enumerator and allocates nothing.

WinUI's equivalent walk (`ScopedResources::TraverseVisualTreeResources`) is likewise a plain loop that allocates nothing, so this moves Uno *towards* WinUI rather than away.

### Measured impact 📊

|---|---|---:|---:|---:|
| `list.scroll-step` | alloc / step | 154,522 B | 133,264 B | **−13.8 %** |
| `list.scroll-step` | alloc / realised container | 6,718.4 B | 5,794.1 B | **−13.8 %** |
| `list.scroll-step` | time / step | 1.6573 ms | 1.6444 ms | −0.8 % (noise) |
| *control* `layout.full-pass` | alloc | 4,096,383 B | 4,095,809 B | ±0 |
| *control* `grid.full-pass` | alloc | 51,833 B | 51,833 B | ±0 |
| *control* `tree.build` | alloc | 21,929,730 B | 21,976,875 B | ±0 |
| *control* `dp.set-get` | alloc | 48,001 B | 48,001 B | ±0 |

A `ListView` over 2,000 items with a two-`TextBlock` `DataTemplate` and 23 realised containers, scrolled 137 px per step (deliberately not a multiple of the item height, so every step straddles items and forces recycling), driven through the dispatcher tick. 3 runs of the full workload set in identical order, median.

The layout and grid workloads are unmoved because their trees are dirtied in place rather than re-entered — this cost is paid on **Enter**, which is what list virtualisation does constantly.

### Renderer 🎨

- [x] Skia

### For which platforms 📱💻🖥️

All platforms 🌍

MartinZikmund:
**GitHub Issue:** closes #24225

## PR Type:

## What changed? 🚀

`GetResourceDictionaries` was a `yield return` iterator, so every call allocated a state machine — and it is called once per theme reference per element entering the tree, on top of every implicit-style lookup and every `{StaticResource}` visual-tree retrieval. It was the largest single allocator while a virtualised list materialises containers during a scroll.

Return a `readonly struct` enumerable with a struct enumerator instead. The walk is unchanged: still lazy, still containing-dictionary first, then non-empty ancestor `Resources` nearest-first, then `Application.Resources`, and still short-circuiting at the caller's first match — so consumers see exactly the sequence the iterator produced. It still implements `IEnumerable<ResourceDictionary>` for the two cold `.ToArray()` callers, while `foreach` binds to the struct enumerator and allocates nothing. WinUI's equivalent walk (`ScopedResources::TraverseVisualTreeResources`) is likewise an allocation-free loop.

`ResourceResolver.TryVisualTreeRetrieval` is rewritten from a `?.` + null-check to `if (scope?.Target is DependencyObject owner)`, so the struct is enumerated directly rather than through a nullable.

### Measured impact 📊

|---|---|---:|---:|---:|
| `list.scroll-step` | alloc / step | 154,522 B | 133,264 B | **−13.8 %** |
| `list.scroll-step` | alloc / realised container | 6,718.4 B | 5,794.1 B | **−13.8 %** |
| `list.scroll-step` | time / step | 1.6573 ms | 1.6444 ms | −0.8 % (noise) |
| *control* `layout.full-pass` | alloc | 4,096,383 B | 4,095,809 B | ±0 |
| *control* `grid.full-pass` | alloc | 51,833 B | 51,833 B | ±0 |
| *control* `tree.build` | alloc | 21,929,730 B | 21,976,875 B | ±0 |
| *control* `dp.set-get` | alloc | 48,001 B | 48,001 B | ±0 |

A `ListView` over 2,000 items with a two-`TextBlock` `DataTemplate` and 23 realised containers, scrolled 137 px per step (deliberately not a multiple of the item height, so every step straddles items and forces recycling), driven through the dispatcher tick. 3 runs of the full workload set in identical order, median.

The layout and grid workloads are unmoved because their trees are dirtied in place rather than re-entered — this cost is paid on **Enter**, which is what list virtualisation does constantly.

### Validation ✅

Two runtime-test suites on Skia Desktop, each set-compared against a baseline build: the **327-test** resources/theming/list suite (`Given_ThemeResource`, `Given_ElementTheme`, `Given_ResourceDictionary`, `Given_ResourceResolver`, `Given_Style`, `Given_ListViewBase`, …) and the **937-test** layout/rendering/framework suite. **1,264 tests, 0 new failures, 0 newly passing** in both.

## PR Checklist ✅

### DIFF
--- a/src/Uno.UI/UI/Xaml/DependencyObject.Store.cs
+++ b/src/Uno.UI/UI/Xaml/DependencyObject.Store.cs
@@ -1581,39 +1581,147 @@ namespace Microsoft.UI.Xaml
 		/// <summary>
 		/// Returns all ResourceDictionaries in scope using the visual tree, from nearest to furthest.
 		/// </summary>
-		internal IEnumerable<ResourceDictionary> GetResourceDictionaries(
+		/// <remarks>
+		/// Returns a struct enumerable rather than a <c>yield return</c> iterator: this is walked once per
+		/// theme reference per element entering the tree, so the state-machine allocation showed up as the
+		/// single largest allocator while a virtualised list materialises containers during a scroll. The
+		/// walk itself is unchanged — still lazy and still short-circuiting at the first match, so callers
+		/// see exactly the sequence the iterator produced. WinUI's equivalent
+		/// (ScopedResources::TraverseVisualTreeResources) is likewise a plain loop that allocates nothing.
+		/// </remarks>
+		internal ResourceDictionaryScope GetResourceDictionaries(
 			bool includeAppResources,
 			FrameworkElement? resourceContextProvider = null,
 			ResourceDictionary? containingDictionary = null)
+			=> new(this, includeAppResources, resourceContextProvider, containingDictionary);
+
+		/// <summary>
+		/// The lazily-walked sequence of <see cref="ResourceDictionary"/> in scope for a
+		/// <see cref="DependencyObject"/>, nearest first. Implements <see cref="IEnumerable{T}"/> so LINQ
+		/// still works on the cold paths, but <c>foreach</c> binds to the struct enumerator and allocates
+		/// nothing.
+		/// </summary>
+		internal readonly struct ResourceDictionaryScope : IEnumerable<ResourceDictionary>
 		{
-			if (containingDictionary is not null)
+			private readonly DependencyObject _owner;
+			private readonly FrameworkElement? _resourceContextProvider;
+			private readonly ResourceDictionary? _containingDictionary;
+			private readonly bool _includeAppResources;
+
+			internal ResourceDictionaryScope(
+				DependencyObject owner,
+				bool includeAppResources,
+				FrameworkElement? resourceContextProvider,
+				ResourceDictionary? containingDictionary)
 			{
-				yield return containingDictionary;
+				_owner = owner;
+				_includeAppResources = includeAppResources;
+				_resourceContextProvider = resourceContextProvider;
+				_containingDictionary = containingDictionary;
 			}
 
-			// for non-FE, favor context provider over actual-instance
-			var candidate = ActualInstance as FrameworkElement ?? resourceContextProvider ?? ActualInstance;
-			while (candidate is not null)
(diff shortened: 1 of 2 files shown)


morning4coffe-dev (commented):
The static-tree order and allocation-free foreach shape are sound, but the claim that the lazy walk is completely unchanged needs qualification. The new enumerator initializes the candidate before yielding containingDictionary and reads fe.Parent before returning the current dictionary. The old iterator resumed those operations after the consumer returned. WinUI also advances to the parent after its resource visitor: https://github.com/microsoft/microsoft-ui-xaml/blob/439bda90582f7892ca595a193f55dbfc6ae2010c/dxaml/xcp/components/resources/ScopedResources.cpp#L139-L174 . A reparent during resource resolution can therefore observe a different ancestry. Please defer advancement until resumption, or explicitly cover/document the intended reentrant behavior; the static corpus does not establish equivalence for that callout case.

PR #24235, opened 2026-08-27
perf(vsm): Stop materialising empty state-trigger collections

### LINKED ISSUE #24226: Every VisualState materialises an empty trigger collection when read

`VisualState.StateTriggers` is a lazily-materialising getter: when the DP holds no list it creates a `DependencyObjectCollection<StateTriggerBase>`, subscribes to its `VectorChanged`, and writes it back through `SetValue`. The profile shows the tell-tale shape — a *getter* ending in a `SetValue`:

```
ArrayPool<short>.Rent
DependencyPropertyDetailsCollection.TryGetPropertyDetails
DependencyObject.SetValue
VisualState.set_StateTriggers      <-- a setter, reached from...
VisualState.get_StateTriggers      <-- ...its own getter
VisualStateGroup.ExecuteOnTriggers
VisualStateGroup.OnOwnerElementChanged
```

That is right for a caller intending to **add** a trigger, but all four consumers in `VisualStateGroup` only ever **read**: `ExecuteOnTriggers` (reached from three call sites — owner changed / loaded / unloaded), `HasStateTriggers` (just a `Count > 0` check) and `GetActiveTrigger` (twice).

So every visual state of every templated control materialises a collection, a `VectorChanged` subscription, a `DependencyPropertyDetails` entry and a pooled-array rent — to represent **no triggers at all**, which is what almost every state has; only adaptive/custom-trigger states ever hold any.

### Why is this needed? 🤔

A non-materialising accessor that reads the DP directly fixes it, with identical behaviour: when no collection exists there are no triggers, so each loop body would have executed zero times anyway and `HasStateTriggers` would have returned false. A caller that genuinely adds a trigger still goes through `StateTriggers` and gets the collection created lazily exactly as before.

### Measured impact 📊

|---|---|---:|---:|---:|
| `list.scroll-step` | alloc / step | 133,277 B | 121,814 B | **−8.6 %** |
| `list.scroll-step` | time / step | 1.6278 ms | 1.5791 ms | **−3.0 %** |
| *control* `layout.full-pass` | alloc | 4,096,259 B | 4,096,127 B | ±0 |
| *control* `grid.full-pass` | alloc | 51,833 B | 51,833 B | ±0 |
| *control* `tree.build` | alloc | 21,946,936 B | 21,942,484 B | ±0 |

Same 2,000-item virtualised `ListView` workload, 3 runs of the full set in identical order, median. Measured on a tree that already has the resource-scope enumerator fix applied.

### Renderer 🎨

- [x] Skia

### For which platforms 📱💻🖥️

All platforms 🌍

MartinZikmund:
**GitHub Issue:** closes #24226

## PR Type:

## What changed? 🚀

`VisualState.StateTriggers` is a lazily-materialising getter: with no list on the DP it creates a `DependencyObjectCollection`, subscribes to its `VectorChanged` and writes it back through `SetValue`. That is right for a caller intending to *add* a trigger, but all four consumers in `VisualStateGroup` only read — `ExecuteOnTriggers` (three call sites), `HasStateTriggers`, and `GetActiveTrigger` twice. So every visual state of every templated control materialised a collection, a subscription, a `DependencyPropertyDetails` entry and a pooled-array rent to represent **no triggers at all**, which is what almost every state has.

Add `StateTriggersOrDefault`, which reads the DP without materialising, and use it at the four read sites. Behaviour is identical: with no collection there are no triggers, so each loop body ran zero times anyway and `HasStateTriggers` returned false. A caller that actually adds a trigger still goes through `StateTriggers` and creates it lazily as before.

### Measured impact 📊

|---|---|---:|---:|---:|
| `list.scroll-step` | alloc / step | 133,277 B | 121,814 B | **−8.6 %** |
| `list.scroll-step` | time / step | 1.6278 ms | 1.5791 ms | **−3.0 %** |
| *control* `layout.full-pass` | alloc | 4,096,259 B | 4,096,127 B | ±0 |
| *control* `grid.full-pass` | alloc | 51,833 B | 51,833 B | ±0 |
| *control* `tree.build` | alloc | 21,946,936 B | 21,942,484 B | ±0 |

Same 2,000-item virtualised `ListView` workload, 3 runs of the full set in identical order, median. Measured on a tree that already has the resource-scope enumerator fix applied.

### Validation ✅

Runtime tests, Skia Desktop — `Given_VisualStateManager`, `Given_EventTrigger`, `Given_Button`, `Given_CheckBox`, `Given_RadioButton`, `Given_ToggleSwitch`, `Given_HyperlinkButton`, `Given_AppBarButton`, `Given_Style`, `Given_ListViewBase`: **221 run, 214 passed, 11 skipped, 7 failed**. Every visual-state and control test passes; the only failures are the `Given_ListViewBase` set already established as pre-existing against baseline builds in the other PRs of this series.

## PR Checklist ✅

### DIFF
--- a/src/Uno.UI/UI/Xaml/VisualState.cs
+++ b/src/Uno.UI/UI/Xaml/VisualState.cs
@@ -119,6 +119,18 @@ namespace Microsoft.UI.Xaml
 			internal set => SetValue(StateTriggersProperty, value);
 		}
 
+		/// <summary>
+		/// The trigger collection if one has already been created, without materialising an empty one.
+		/// </summary>
+		/// <remarks>
+		/// Reading <see cref="StateTriggers"/> lazily creates the collection, subscribes to it and writes it
+		/// back through <c>SetValue</c>. That is pure overhead for the common case of a state with no
+		/// triggers at all, and it is paid for every state of every templated control as it is loaded.
+		/// Callers that only read should use this instead.
+		/// </remarks>
+		internal IList<StateTriggerBase> StateTriggersOrDefault
+			=> GetValue(StateTriggersProperty) as IList<StateTriggerBase>;
+
 		internal static DependencyProperty StateTriggersProperty { get; } =
 			DependencyProperty.Register(
 				name: "StateTriggers",
--- a/src/Uno.UI/UI/Xaml/VisualStateGroup.cs
+++ b/src/Uno.UI/UI/Xaml/VisualStateGroup.cs
@@ -181,9 +181,14 @@ namespace Microsoft.UI.Xaml
 			for (var stateIndex = 0; stateIndex < States.Count; stateIndex++)
 			{
 				var state = States[stateIndex];
-				for (var triggerIndex = 0; triggerIndex < state.StateTriggers.Count; triggerIndex++)
+				if (state.StateTriggersOrDefault is not { } triggers)
 				{
-					var trigger = state.StateTriggers[triggerIndex];
+					continue;
+				}
+
+				for (var triggerIndex = 0; triggerIndex < triggers.Count; triggerIndex++)
+				{
+					var trigger = triggers[triggerIndex];
 
 					action(trigger);
 				}
@@ -564,7 +569,7 @@ namespace Microsoft.UI.Xaml
 			for (var stateIndex = 0; stateIndex < States.Count; stateIndex++)
 			{
 				var state = States[stateIndex];
-				if (state.StateTriggers.Count > 0)
+				if (state.StateTriggersOrDefault is { Count: > 0 })
 				{
 					return true;
 				}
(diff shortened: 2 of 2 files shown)


morning4coffe-dev (approved):
Code review assessment: reading the nullable existing trigger collection is aligned with WinUI's standard VSM path, which returns when m_pStateTriggerCollection is null rather than creating an empty collection: https://github.com/microsoft/microsoft-ui-xaml/blob/439bda90582f7892ca595a193f55dbfc6ae2010c/dxaml/xcp/components/vsm/StandardVisualStateManagerDataSource.cpp#L388-L400 . The normal StateTriggers getter still creates/subscribes the collection when needed. No blocking behavior issue found. The nullable return annotation can be clarified without replacing it with a materialized DP value.

PR #24236, opened 2026-08-27
fix(binding): Drop dead INotifyPropertyChanged subscriptions

### LINKED ISSUE #24214: Binding source accumulates PropertyChanged subscriptions for every element ever bound to it
### Current behavior 🐛

A long-lived `INotifyPropertyChanged` binding source accumulates `PropertyChanged` subscriptions **forever** — one for every element that was ever bound to it, never removed.

Measured by reading the source's own invocation-list length across create → bind → add to tree → remove → collect cycles:

```
batch 1/5: subscribers=35    targets alive=1/35
batch 2/5: subscribers=60    targets alive=1/60
batch 3/5: subscribers=85    targets alive=1/85
batch 4/5: subscribers=110   targets alive=1/110
batch 5/5: subscribers=135   targets alive=1/135
```

`targets alive=1/135` is the important half: the **bound elements are collected** (the 1 is the most recent, still held by a local), so this is not element retention. What leaks is the subscription itself — the delegate plus the weak-reference wrapper it captures — held by the source for its entire lifetime.

Two consequences:

1. **Unbounded memory growth** proportional to the number of elements ever bound to that source.
2. **Every `PropertyChanged` raise gets slower**, because the raise walks the whole invocation list including the dead entries. Measured at 135 accumulated subscriptions, a single raise costs **2.873 µs** against **0.455 µs** with only the live one — and that gap grows without bound.

This bites hardest in the shapes real apps use: a singleton or long-lived view model bound from item templates in a virtualised list, or from a page that is navigated to repeatedly.

### Expected behavior 🎯

Subscriptions belonging to collected targets should be dropped, so the handler count stays proportional to the number of *live* bindings.

1. Create a view model implementing `INotifyPropertyChanged` and keep it alive for the whole run.
2. Repeatedly: create a `TextBlock`, set its `DataContext` to that view model, `SetBinding(TextBlock.TextProperty, new Binding { Path = new PropertyPath("Name") })`, add it to the visual tree, then remove it.
3. `GC.Collect()` between cycles.
4. Read the view model's handler count, e.g. via a `public int SubscriberCount => PropertyChanged?.GetInvocationList().Length ?? 0;` property on the view model.

The count grows by exactly one per cycle and never decreases.

### Workaround 🛠️

Avoid binding many transient elements to one long-lived source, or ensure the binding is explicitly cleared (`ClearValue` on the bound property, or dropping the `DataContext`) before the element is released.

### Renderer 🎨

- [x] Skia

### Affected platforms 📱💻🖥️

All platforms 🌍

Reproduced on the current development branch; the code path is present in shipping code.

**Cause.** `BindingPath.SubscribeToNotifyPropertyChanged` attaches a closure to the source that holds only a *weak* reference to the value handler — which is exactly why the target stays collectable. But the subscription is removed only by the `IDisposable` that method returns, and nothing disposes it when the target is simply garbage-collected. The handler therefore stays attached permanently, with a dead weak reference inside it.

**Fix.** Make it a self-purging weak event: when the handler is raised and finds its value handler has gone, detach itself from the source — the same thing WPF's `PropertyChangedEventManager` does when it finds a dead listener. The dead-check has to sit above the property-name filter so that a raise for *any* property purges, and the detach should go through the same weak data-context reference the disposable uses, so a source that raises with a different `sender` still works. Unsubscribing during a raise is safe, since delegate invocation lists are immutable.

Purging then happens on the source's next raise, which is the standard limitation of self-purging weak events (WPF shares it): a source that never raises keeps its subscriptions, but it also never pays the raise cost.

I have a fix with this evidence and will open a PR.

MartinZikmund:
**GitHub Issue:** closes #24214

## PR Type:

🐞 Bugfix

## What changed? 🚀

A long-lived binding source accumulated `PropertyChanged` subscriptions forever: one per element ever bound to it, never removed. The bound elements themselves **are** collected — what leaked is the subscription, the delegate plus its captured weak-reference wrapper.

`SubscribeToNotifyPropertyChanged` attaches a closure holding only a weak reference to the value handler, which is why the target stays collectable, but the subscription is removed only by the returned `IDisposable` — and nothing disposes it when the target is simply garbage-collected.

Make it self-purging: when raised and the value handler has gone, detach from the source, the way WPF's `PropertyChangedEventManager` does with a dead listener. The dead-check sits above the property-name filter so a raise for *any* property purges, and detaching goes through the same weak data-context reference the disposable uses, so a source raising with a different `sender` still works. Unsubscribing during a raise is safe: invocation lists are immutable, so the in-flight raise completes over the list it started with.

Purging happens on the source's next raise, which is the standard limitation of self-purging weak events (WPF shares it): a source that never raises keeps its subscriptions, but it also never pays the raise cost.

### Measured impact 📊

| Metric | Before | After |
|---|---:|---:|
| subscriptions held after 135 bound-then-removed elements | **135** (unbounded) | **1** (bounded) |
| subscriptions between raises | 35 → 60 → 85 → 110 → 135 | 26, flat |
| subscriptions after each raise | unchanged | **1** |
| cost per `PropertyChanged` raise | 2.873 µs | **0.455 µs** (**−84 %**) |

The raise cost is the second-order effect: every raise walks the whole invocation list, dead entries included. −84 % is the figure *after only 135 cycles* — the gap grows without bound, because the before-case list never stops growing while the after-case stays at the number of live bindings.

Measured on the source's own invocation-list length, so the evidence is **deterministic** and does not depend on the GC actually collecting anything.

### Validation ✅

**A regression test is included**, and was verified red before the fix and green after:

- without the fix: **30 subscriptions held, 0 targets alive** → fails
- with the fix: **0 held** → passes

It asserts the precise invariant — the source must never hold more subscriptions than there are live targets — rather than a fixed number, so it stays meaningful even if the GC declines to collect some targets on a given run. `Given_BindingMemoryLeak`: **7 run, 7 passed**.

Wider: `Given_BindingExpression`, `Given_xBind`, `Given_XBind_NavigatedTo`, `Given_BindableFoundationStructs`, `Given_BindableNullableValueType`, `Given_NonFE_DataContextBinding`, `Given_Frame_DataContext`, `Given_ResourceDictionary_DataContext`, `Given_FrameworkElement_And_Leak`, `Given_ListViewBase`: **278 run, 270 passed, 11 skipped, 7 failed**, set-compared against a baseline build — **0 new failures**.

## PR Checklist ✅

- [x] 🧪 Added [Runtime tests, UI tests, or a manual test sample](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) (for bug fixes / features, if applicable) — `Given_BindingMemoryLeak.When_Source_Outlives_Targets_Then_Subscriptions_Are_Dropped`, verified red→green

### DIFF
--- a/src/Uno.UI/DataBinding/BindingPath.cs
+++ b/src/Uno.UI/DataBinding/BindingPath.cs
@@ -495,8 +495,27 @@ namespace Uno.UI.DataBinding
 
 				var newValueActionWeak = Uno.UI.DataBinding.WeakReferencePool.RentWeakReference(null, propertyChangedValueHandler);
 
-				System.ComponentModel.PropertyChangedEventHandler handler = (s, args) =>
+				System.ComponentModel.PropertyChangedEventHandler? handler = null;
+				handler = (s, args) =>
 				{
+					// The subscription is only detached by the disposable returned below, and nothing disposes it
+					// when the binding target is simply collected — so without this the source would hold a dead
+					// handler for the rest of its life, and every raise would walk all of them. Detaching here on
+					// the first raise after the value handler has gone is the self-purging weak-event pattern
+					// (the same thing WPF's PropertyChangedEventManager does when it finds a dead listener).
+					if (newValueActionWeak.IsDisposed
+						|| newValueActionWeak.Target is not IPropertyChangedValueHandler valueHandler)
+					{
+						// Detached through the same weak reference the disposable uses, so a source that raises
+						// with a different sender is still handled correctly.
+						if (dataContextReference.Target is System.ComponentModel.INotifyPropertyChanged deadSource)
+						{
+							deadSource.PropertyChanged -= handler;
+						}
+
+						return;
+					}
+
 					if (args.PropertyName == propertyName || string.IsNullOrEmpty(args.PropertyName))
 					{
 						if (typeof(BindingPath).Log().IsEnabled(Uno.Foundation.Logging.LogLevel.Debug))
@@ -504,11 +523,7 @@ namespace Uno.UI.DataBinding
 							typeof(BindingPath).Log().Debug($"Property changed for {propertyName} on [{dataContextReference.Target?.GetType()}]");
 						}
 
-						if (!newValueActionWeak.IsDisposed
-						&& newValueActionWeak.Target is IPropertyChangedValueHandler handler)
-						{
-							handler.NewValue();
-						}
+						valueHandler.NewValue();
 					}
 				};
 
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_BindingMemoryLeak.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_Xaml/Given_BindingMemoryLeak.cs
@@ -1,7 +1,9 @@
 #nullable enable
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
 using System.Runtime.CompilerServices;
 using System.Threading.Tasks;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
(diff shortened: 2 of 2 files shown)


ajpinedam (approved):

src/Uno.UI/DataBinding/BindingPath.cs
```diff
@@ -495,20 +495,35 @@ private static void TryPrependItem(
...
+					// when the binding target is simply collected — so without this the source would hold a dead
+					// handler for the rest of its life, and every raise would walk all of them. Detaching here on
+					// the first raise after the value handler has gone is the self-purging weak-event pattern
+					// (the same thing WPF's PropertyChangedEventManager does when it finds a dead listener).
+					if (newValueActionWeak.IsDisposed
+						|| newValueActionWeak.Target is not IPropertyChangedValueHandler valueHandler)
+					{
+						// Detached through the same weak reference the disposable uses, so a source that raises
+						// with a different sender is still handled correctly.
+						if (dataContextReference.Target is System.ComponentModel.INotifyPropertyChanged deadSource)
```
morning4coffe-dev, line 511: [P2] Use disposal-safe access when purging an in-flight handler

An earlier handler in the same PropertyChanged invocation can rebind the target. BindingItem.SetWeakDataContext then returns/disposes the old data-context reference, while the source's captured multicast invocation can still invoke this old handler. The new branch is entered when newValueActionWeak is disposed, but dataContextReference.Target throws ObjectDisposedException when its reference has also been disposed. Use the existing disposal-safe TryGetTarget API (and retain correct handling for sources that use a different sender), with a reentrant-rebinding regression. This should detach or harmlessly ignore an obsolete callback, not throw out of the source's event.

morning4coffe-dev (changes requested):
The self-purging weak-listener goal is sound, but the new dead-listener path can throw during a legitimate reentrant PropertyChanged delivery. I verified the disposed-reference concern in the existing thread against ManagedWeakReference.Target and BindingItem.SetWeakDataContext; please fix that lifetime edge before approval. This is an implementation regression, not a claim that WinUI requires retaining dead subscribers.

PR #24242, opened 2026-08-27
perf(uno-ui): Use the cached bool boxes in generated DP setters

### LINKED ISSUE #24241: Generated bool DependencyProperty setters box on every set

`DependencyObject.SetValue` takes an `object`, so a `bool` is boxed at the call site — **24 bytes on every set**.

Uno.UI already keeps the two boxed instances for exactly this reason (`Uno.UI.Helpers.Boxes.BooleanBoxes`, with a `Boxes.Box(bool)` helper) and uses them by hand in a handful of places. But both DependencyProperty generators emit the plain form:

```csharp
private void SetIsEnabledValue(bool value) => SetValue(IsEnabledProperty, value);
//                                                                       ^^^^^ boxed, every time
```

so every generated `bool` setter allocates.

Both generators should emit `SetValue(XProperty, global::Uno.UI.Helpers.Boxes.Box(value))` when the property type is `bool`, passing `value` through unchanged for every other type — which also makes it automatic for any `bool` DP added later.

- `src/SourceGenerators/Uno.UI.SourceGenerators.Internal/DependencyObject/DependencyPropertyGenerator.cs`
- `src/SourceGenerators/Uno.UI.SourceGenerators.Internal/Mixins/DependencyPropertyMixinGenerator.cs` (its data model declares 13 `bool` properties)

### Why is this needed? 🤔

`bool` dependency properties are set constantly — visual-state changes, `IsEnabled` propagation down a subtree, focus and tab-stop flags — and each set allocates a box that is immediately discarded.

### Measured impact 📊

1,000 × toggling `Control.IsEnabled` (a `[GeneratedDependencyProperty]` bool) on a `Button`, 3 runs, median:

| Metric | Before | After | Δ |
|---|---:|---:|---:|
| alloc / set | 72.0 B | 48.0 B | **−33.3 %** |

Byte-identical across every run in both configurations — exactly **24 B per set**, one box. The remaining 48 B is the change-notification path and is unrelated.

### Why it cannot change behaviour ✅

`DependencyObject.AreDifferent` (`DependencyObject.Store.cs:2313`) takes the `newValue is ValueType` branch for a boxed `bool` and compares with `object.Equals` — a **value** comparison, not a reference comparison:

```csharp
if (newValue is ValueType || newValue is string)
{
    return !object.Equals(previousValue, newValue);
}
```

So change detection never depended on box identity, and sharing the two boxed instances is invisible to it.

### Renderer 🎨

- [x] Skia

### For which platforms 📱💻🖥️

All platforms 🌍

**Follow-up, deliberately not included:** **151 hand-written `bool` DP setters** in `Uno.UI` still call `SetValue(XProperty, value)` directly and pay the same 24 B, including hot ones on the base types — `UIElement.IsHitTestVisible` (72 B/set measured) and `UIElement.IsTabStop` (24 B/set measured) are both hand-written, not generated. That is a 151-site mechanical sweep and is worth doing separately.

Found by allocation profiling and measured before/after with a headless harness that project-references the real framework. Full methodology and raw runs: `specs/059-performance-improvements/evidence.md`.

MartinZikmund:
**GitHub Issue:** closes #24241

## PR Type:

## What changed? 🚀

`DependencyObject.SetValue` takes an `object`, so a `bool` is boxed at the call site — 24 bytes on **every set**. Uno.UI already keeps the two boxed instances in `Uno.UI.Helpers.Boxes` and uses them by hand in a few places, but both DependencyProperty generators emitted the plain `SetValue(prop, value)`, so every generated `bool` setter allocated.

Both generators now emit `SetValue(XProperty, global::Uno.UI.Helpers.Boxes.Box(value))` when the property type is `bool`, and pass `value` through unchanged for every other type — so any `bool` DP added later gets this automatically.

- `DependencyPropertyGenerator` — both emission sites (attached and instance setters)
- `DependencyPropertyMixinGenerator` — its data model declares 13 `bool` properties

`Uno.UI.Helpers.Boxes` is internal to Uno.UI, which is the only assembly using `[GeneratedDependencyProperty]`; any other consumer would get a compile error here rather than a silent behaviour change.

### Measured impact 📊

1,000 × toggling `Control.IsEnabled` (a `[GeneratedDependencyProperty]` bool) on a `Button`, 3 runs, median:

| Metric | Before | After | Δ |
|---|---:|---:|---:|
| alloc / set | 72.0 B | 48.0 B | **−33.3 %** |

Byte-identical across every run in both configurations — exactly **24 B per set**, one box. The remaining 48 B is the change-notification path and is unrelated to this change.

Measured with a headless harness that project-references the real framework, rebuilt from the working tree; full methodology and raw runs in `specs/059-performance-improvements/evidence.md` (PR #24238).

### Why it cannot change behaviour ✅

`DependencyObject.AreDifferent` (`DependencyObject.Store.cs:2313`) takes the `newValue is ValueType` branch for a boxed `bool` and compares with `object.Equals` — a **value** comparison:

```csharp
if (newValue is ValueType || newValue is string)
{
    return !object.Equals(previousValue, newValue);
}
```

Change detection therefore never depended on box identity, so sharing the two boxed instances is invisible to it.

### Validation ✅

- **`DependencyPropertyGenerator` golden tests: 6 run, 6 passed.**
- **Runtime tests, Skia Desktop** — `Given_Control`, `Given_Button`, `Given_CheckBox`, `Given_ToggleSwitch`, `Given_RadioButton`, `Given_VisualStateManager`, `Given_UIElement`, `Given_FrameworkElement`: **684 run, 658 passed, 2 skipped, 26 failed**, compared as *sets of failing test names* against a baseline build — **0 new failures, 0 newly passing.**

> ⚠️ Note for reviewers: the wider `Uno.UI.SourceGenerators.Tests` suite is **not** a usable signal locally — it fails 340/482 on the **unmodified** tree and its pass count varies run to run (110, 128, 129 on identical inputs), because those tests load `src/*/bin/**` directly rather than through project references. The 6 DependencyProperty golden tests above were therefore checked individually. CI is the real check.

### Follow-up (not in this PR)

**151 hand-written `bool` DP setters** in `Uno.UI` still call `SetValue(XProperty, value)` directly and pay the same 24 B — including hot ones on the base types: `UIElement.IsHitTestVisible` (measured 72 B/set) and `UIElement.IsTabStop` (24 B/set) are hand-written, not generated. That is a 151-site mechanical sweep and is better done separately.

## PR Checklist ✅

- [ ] 🧪 Added [Runtime tests, UI tests, or a manual test sample](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) (for bug fixes / features, if applicable) — n/a, allocation-only change with no behaviour difference; covered by the existing generator goldens and control suites above

### DIFF
--- a/src/SourceGenerators/Uno.UI.SourceGenerators.Internal/DependencyObject/DependencyPropertyGenerator.cs
+++ b/src/SourceGenerators/Uno.UI.SourceGenerators.Internal/DependencyObject/DependencyPropertyGenerator.cs
@@ -490,7 +490,7 @@ namespace Uno.UI.SourceGenerators.DependencyObject
 				}
 			}
 
-			builder.AppendLineIndented($"private static void Set{propertyName}Value({propertyTargetName} instance, {propertyTypeName} value) => instance.SetValue({propertyOwnerTypeName}.{propertyName}Property, value);");
+			builder.AppendLineIndented($"private static void Set{propertyName}Value({propertyTargetName} instance, {propertyTypeName} value) => instance.SetValue({propertyOwnerTypeName}.{propertyName}Property, {GetBoxedValueExpression(propertyTypeName)});");
 
 			GeneratePropertyStorage(builder, propertyName);
 
@@ -613,7 +613,7 @@ namespace Uno.UI.SourceGenerators.DependencyObject
 
 			if (propertyData.PropertyHasSetter)
 			{
-				builder.AppendLineIndented($"private void Set{propertyName}Value({propertyTypeName} value) => SetValue({propertyName}Property, value);");
+				builder.AppendLineIndented($"private void Set{propertyName}Value({propertyTypeName} value) => SetValue({propertyName}Property, {GetBoxedValueExpression(propertyTypeName)});");
 			}
 
 			if (localCache)
@@ -716,6 +716,22 @@ namespace Uno.UI.SourceGenerators.DependencyObject
 				}
 			}
 		}
+
+		/// <summary>
+		/// The expression a generated setter passes to <c>SetValue</c>. <c>SetValue</c> takes an <c>object</c>,
+		/// so a <c>bool</c> would be boxed on every set — 24 bytes each time, for properties that are set
+		/// constantly (visual states, IsEnabled propagation, focus flags). Uno.UI already keeps the two
+		/// boxed instances, so use them.
+		/// </summary>
+		/// <remarks>
+		/// <c>Uno.UI.Helpers.Boxes</c> is internal to Uno.UI, which is the only assembly using
+		/// <c>[GeneratedDependencyProperty]</c>. Any other consumer would get a compile error here rather
+		/// than a silent behaviour change.
+		/// </remarks>
+		private static string GetBoxedValueExpression(string propertyTypeName)
+			=> propertyTypeName == "bool"
+				? "global::Uno.UI.Helpers.Boxes.Box(value)"
+				: "value";
 
 		private static void GeneratePropertyStorage(IndentedStringBuilder builder, string propertyName)
 		{
--- a/src/SourceGenerators/Uno.UI.SourceGenerators.Internal/Mixins/DependencyPropertyMixinGenerator.cs
+++ b/src/SourceGenerators/Uno.UI.SourceGenerators.Internal/Mixins/DependencyPropertyMixinGenerator.cs
@@ -54,7 +54,11 @@ public sealed class DependencyPropertyMixinGenerator : IIncrementalGenerator
 					sb.AppendLine($"\t\t{dp.Modifier}public {dp.PropertyType} {dp.Name}");
 					sb.AppendLine("\t\t{");
 					sb.AppendLine($"\t\t\tget {{ return ({dp.PropertyType})this.GetValue({dp.Name}Property); }}");
-					sb.AppendLine($"\t\t\tset {{ this.SetValue({dp.Name}Property, value); }}");
+					// SetValue takes an object, so a bool would be boxed on every set. Uno.UI already keeps the
+					// two boxed instances, and these properties (IsTabStop, focus flags and the like) are set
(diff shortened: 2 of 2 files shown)


Youssef1313:
@MartinZikmund FYI, https://github.com/unoplatform/uno/pull/12677

Feel free to continue with your PR if it's more ready. You can take also a look at mine to see if there are interesting differences and try to take the best of the two worlds.

morning4coffe-dev (approved):
Code review assessment: the emitted setters still call SetValue with the same Boolean value, now using the existing true/false boxes; non-Boolean types retain their previous expression. I checked helper accessibility for the actual consumers, including the friend unit-test assembly. This is a managed allocation optimization, not a change to WinUI Boolean/DP semantics. No blocking source issue found. Add golden assertions for both instance/attached and mixin Boolean setters to protect the new code-generation branch. Any failed automated-review check remains a separate merge gate.

MartinZikmund:
Superseded by #12677, which has been brought up to current `master`.

That PR does the same thing generalised: the generated setter boxing now covers `int` and `double` as well as `bool`, the `DefaultValue`/`FrameworkPropertyMetadata` defaults are boxed too, and an analyzer (`UnoInternal0002`) enforces it across the whole of `Uno.UI` rather than just the generator output.

The `DependencyPropertyMixinGenerator` hunk from this PR — which #12677 predated — is carried over there.

One correction worth recording: this PR's comment asserted that `Uno.UI` is the only assembly using `[GeneratedDependencyProperty]`, so a consumer without access to `Boxes` "would get a compile error rather than a silent behaviour change". That isn't right — `Uno.UI.UnitTests` uses the attribute too, and a compile error is not an acceptable outcome for an external consumer either way. #12677 instead has the generator check that `Uno.UI.Helpers.Boxes` is actually accessible from the compilation and emit the plain value when it isn't.

PR #24255, opened 2026-08-27
feat(codegen): Stamp OriginalSourceLocation on ResourceDictionary

### LINKED ISSUE #24254: Generate OriginalSourceLocation for all ResourceDictionary declarations

When Hot Reload code generation is enabled, the XAML generator stamps the declaration site of the objects it creates so a live object can be mapped back to its markup — `FrameworkElementHelper.SetBaseUri(...)` for `FrameworkElement`s, and the weak `"OriginalSourceLocation"` property (`MarkupHelper.SetElementProperty`) for everything else.

The `"OriginalSourceLocation"` channel is opt-in per type, and currently covers `VisualState` / `AdaptiveTrigger` / `StateTrigger`, `Style`, the templates, and a top-level `<ResourceDictionary x:Class="…">` (#15846 covered exactly that last case).

`ResourceDictionary` instances are otherwise never stamped. None of these carry a source location today:

- a dictionary file **without** `x:Class` (the instance `GlobalStaticResources` exposes for the file);
- an inline `<ResourceDictionary>` under `ResourceDictionary.MergedDictionaries`;
- a `<ResourceDictionary x:Key="…">` under `ResourceDictionary.ThemeDictionaries`;
- the explicit `<ResourceDictionary>` of a `FrameworkElement.Resources` / `Application.Resources`.

Request: every `ResourceDictionary` materialized from a `<ResourceDictionary>` element of the file being generated should carry `"OriginalSourceLocation"`, using the existing helper (no new runtime API).

Dictionaries pulled in through `Source=` should be excluded — that instance belongs to the referenced file, which stamps it itself — as should typed dictionary subclasses, whose own generated `InitializeComponent` already stamps `this`.

### Why is this needed? 🤔

A designer surface (Hot Design's resource editing) that holds a `ResourceDictionary` instance cannot tell which file and line declared it, so it cannot navigate to or edit the declaration. Every other object a user can select in markup already exposes that information through one of the two channels.

### For which platforms 📱💻🖥️

All platforms 🌍 (code generation)

Prior art: #15846, which added the stamp for the `x:Class`'d top-level dictionary only.

DerekRaven:
**GitHub Issue:** closes #24254

## PR Type:

✨ Feature

## What changed? 🚀

When Hot Reload code generation is enabled, the XAML generator stamps the declaration site of the objects it creates: `FrameworkElementHelper.SetBaseUri(...)` for `FrameworkElement`s, and the weak `"OriginalSourceLocation"` property (`MarkupHelper.SetElementProperty`) for everything else. That second channel is opt-in per type and covered `VisualState` / `AdaptiveTrigger` / `StateTrigger`, `Style`, the templates, and a top-level `<ResourceDictionary x:Class="…">` — so **`ResourceDictionary` instances were otherwise never stamped**, and a designer holding one could not tell which file/line declared it.

This stamps every `ResourceDictionary` a file's generated code materializes, reusing the existing `TrySetOriginalSourceLocation` helper and the existing emission patterns (no new runtime API, no new metadata channel):

| Where | What |
|---|---|
| `InitializeAndBuildResourceDictionary` | the one place emitting `new ResourceDictionary { … }`, stamped through `CreateApplyBlock` (as `Style` and the templates do, since the dictionary is in expression position). Covers dictionary files without `x:Class`, inline merged dictionaries, theme dictionaries, and dictionaries built through `BuildChild` |
| `BuildTopLevelResourceDictionary` | the `CreateWithCapacity` fast path is a statement context, so it is stamped directly on the field |
| `BuildExtendedProperties` | `<applied>.Resources`, next to the existing `SetBaseUri`, for the explicit `<X.Resources><ResourceDictionary>` of the top-level control and of every child |
| `BuildApplicationInitializerBody` | same, for `Application` — App.xaml never goes through `BuildExtendedProperties` |
| `BuildTypedResourceDictionary` | a typed subclass (`<local:MyDictionary/>`), whether merged into another dictionary or used as a whole `Resources` value, stamped **set-if-absent** |
| `MarkupHelper` | the key is declared once as `OriginalSourceLocationPropertyName`, with the `file:///<path>#L<line>:<position>` grammar documented for the out-of-repo consumers that parse it |

### Typed subclasses are stamped set-if-absent, not skipped

A subclass generated from an `x:Class` dictionary file stamps its own declaration site in the `InitializeComponent` its constructor runs — before the use-site apply block — so a plain use-site stamp would overwrite a declaration site with a reference site. A subclass **defined in code** (a library theme dictionary, e.g. Uno.Themes' `BaseTheme` family, whose constructor points `Source` at the library's own style bundle) has no `InitializeComponent` at all, and the declaring markup is the only source location it can ever get.

Rather than test the type, the use-site stamp is guarded:

```csharp
if (global::Uno.UI.Helpers.MarkupHelper.GetElementProperty<string>(__p1, "OriginalSourceLocation") is null)
{
    global::Uno.UI.Helpers.MarkupHelper.SetElementProperty(__p1, "OriginalSourceLocation", "file:///…#L8:10");
}
```

which keeps the declaration site winning in both cases. (A compile-time "does it have a generated backing class" test would not work for a dictionary of the *same* compilation: its generated `InitializeComponent` doesn't exist in the symbol table yet at generation time.)

Deliberately **not** stamped:

- dictionaries pulled in with `Source=` — that instance belongs to the referenced file (which stamps it itself) and is shared, so stamping here would attribute another file's dictionary to this one and overwrite a correct value;
- a `Resources` member with no explicit `<ResourceDictionary>` element (`<Page.Resources><Style/></Page.Resources>`) — there is no dictionary element in the markup to point at, and the owner already carries its own location;
- anything that already carries a location — a declaration site is never replaced by a reference site.

Every emission is inside the existing `_isHotReloadEnabled` guard, so generated code is unchanged when Hot Reload codegen is off. The design, the non-goals and the edge cases are written up in `specs/057-resourcedictionary-source-location/spec.md`.

### Confirmed against the real consumer

Hot Design was pointed at this branch's generator (overriding the `uno.winui` 6.7.16 analyzer) and driven by hand on the `MvvmMaterialFrame` sample, whose App.xaml merges `<utum:MaterialToolkitTheme …/>` — a **code-defined** `Uno.Themes.BaseTheme` subclass that carries no location without R5:

```
stampedLocation='file:////…/MvvmMaterialFrame/App.xaml#L14:6'   ← the <utum:MaterialToolkitTheme> element
identifier: valid=True editable=True line=14:6
readOnly=False subject=MaterialToolkitTheme
```

The design system's property grid went from uniformly read-only to editable. That closes the review question of whether a dictionary-level location is what the consumer actually needs.

### Validation

- **Generator tests** — five new snapshot tests in `Given_HotReloadEnabledInBuild` (the only suite that enables `UnoForceHotReloadCodeGen`): a dictionary file (fast path + inline merged + themed + a `Source=` merged dictionary that must stay unstamped); a page with an explicit `<Page.Resources><ResourceDictionary>` plus a `Grid` with its own, covering both the `useGenericApply` and object-initializer paths; and a page using a code-defined `ResourceDictionary` subclass both merged and as a whole `Grid.Resources`. `ResourceDictionaryCodeBehind`'s expected output was regenerated: the file-level instance of an `x:Class`'d dictionary is now stamped too. Full suite green (the 14 skips are pre-existing `Assert.Inconclusive` cases).
- **Compile** — `Uno.UI.UnitTests` built with `-p:UnoForceHotReloadCodeGen=true`: 0 warnings / 0 errors, and the emitted `App_*.cs` carries `SetElementProperty(Resources, "OriginalSourceLocation", "…/App/App.xaml#L12:4")`. This is how the `Application` path is validated — the XAML-generator test harness cannot process an `Application` root at all (a test with an App.xaml root fails on `master` with `UXAML0001 … BuildApplicationInitializerBody@466`, with or without this change).
- **Runtime** — `Uno.UI.UnitTests` run with the same flag (which materializes the app's dictionaries, merged and themed, through the new emissions): the failure set is identical to the baseline captured with this change stashed (Linux-only path/UNC/case-sensitivity expectations).
- **Runtime, the typed-subclass rule** — a throwaway page (deleted after the run) merging a code-defined subclass and the project's `x:Class` `Subclassed_Dictionary`, reading the property back off each dictionary:

  | Dictionary | Location read back |
  |---|---|
  | the page's `<Page.Resources><ResourceDictionary>` | `Temp_SourceLocation.xaml#L6:4` |
  | merged **code-defined** subclass | `Temp_SourceLocation.xaml#L8:6` — the declaring line ✅ |
  | merged **`x:Class`** subclass | `Subclassed_Dictionary.xaml#L1:2` — its own file, not the use site ✅ |
  | `Test_Page_Other`'s `Resources = new Subclassed_Dictionary()` | `Subclassed_Dictionary.xaml#L1:2` |
  | `Application.Current.Resources` | `App/App.xaml#L12:4` |

## PR Checklist ✅

- [x] 🧪 Added [Runtime tests, UI tests, or a manual test sample](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) (for bug fixes / features, if applicable) — source-generator snapshot tests, see above
- [ ] 📚 Docs have been added/updated following the [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features) — n/a, internal Hot Reload metadata with no public API surface
- [ ] 🖼️ Validated PR `Screenshots Compare Test Run` results. — pending CI


DerekRaven:
The "typed dictionary subclasses" exclusion looks slightly too broad. Its rationale — "their `InitializeComponent` already stamps `this`" — only holds for subclasses that *have* a generated backing class (`BuildResourceDictionaryBackingClass`, i.e. `x:Class` XAML dictionaries). A **code-defined** `ResourceDictionary` subclass has no `InitializeComponent`, so with the use site excluded it ends up stamped by nobody.

The concrete case is Uno.Themes' `BaseTheme` family. An app declares:

```xml
<Application.Resources>
    <ResourceDictionary>
        <ResourceDictionary.MergedDictionaries>
            <utum:MaterialToolkitTheme ColorOverrideSource="ms-appx:///Styles/ColorPaletteOverride.xaml" />
        </ResourceDictionary.MergedDictionaries>
    </ResourceDictionary>
</Application.Resources>
```

`MaterialToolkitTheme` is defined in C# (its constructor points `Source` at the library's own style bundle), so tooling holding the live instance has no way back to the App.xaml (or merged file) line that declared it — the same "a designer holding one could not tell which file/line declared it" problem this PR fixes for plain `<ResourceDictionary>` elements, and arguably its highest-value instance: the theme element is where an application's design tokens live (`DefaultCornerRadius`, seed colours, font overrides), so it is exactly the dictionary a design or diagnostic tool wants to trace to source. `Source` is no substitute here — on these types it names the library's styles file, not the declaring document.

Suggested amendment: make the use-site stamp **set-if-absent** rather than excluding subclasses by type. An `x:Class` subclass stamps `this` during construction (`InitializeComponent`), which runs before the use-site apply block — so set-if-absent preserves the "don't overwrite the declaration site with the reference site" invariant without any type test, while a code-defined subclass (which reaches the apply block unstamped) gains its declaration site. If a runtime presence check is unattractive, narrowing the exclusion to `IsResourceDictionarySubclass` types that actually have a generated backing class achieves the same.

Happy to validate against the Uno.Themes case once amended.

DerekRaven:
Good catch — the rationale only held for subclasses that *have* a generated backing class, and a code-defined one was left stamped by nobody. Amended in e9334497c9 with the set-if-absent variant.

`BuildTypedResourceDictionary` now appends an apply block like the other dictionary paths, and `TrySetOriginalSourceLocation` grew a `preserveExisting` option that wraps the stamp in a presence check, so the emission is:

```csharp
new /* typed resource dictionary */ global::TestRepro.MyCodeDictionary()
.GenericApply(__that, __nameScope, ApplyTo_Pag_ResΞ0_ResDic_MerDicΞ0_MyCodDic)

private void ApplyTo_Pag_ResΞ0_ResDic_MerDicΞ0_MyCodDic(global::TestRepro.MyCodeDictionary __p1, …)
{
    if (global::Uno.UI.Helpers.MarkupHelper.GetElementProperty<string>(__p1, "OriginalSourceLocation") is null)
    {
        global::Uno.UI.Helpers.MarkupHelper.SetElementProperty(__p1, "OriginalSourceLocation", "file:///…/MainPage.xaml#L8:10");
    }
}
```

No new runtime API — `MarkupHelper.GetElementProperty` has been there since 2022 (and is in the 5.0.118 package the snapshot tests compile against). It covers both use sites, since `BuildTypedResourceDictionary` is also what `RegisterAndBuildResources` calls for `<Grid.Resources><local:MyDictionary/></Grid.Resources>`.

I went with set-if-absent over narrowing the exclusion for a reason worth recording: whether a subclass has a generated backing class isn't reliably knowable from the generator for a dictionary of the **same** compilation — its generated `InitializeComponent` doesn't exist in the symbol table yet at generation time, so the test would only work for cross-assembly dictionaries.

### Validated against the real shape

A throwaway page in `Uno.UI.UnitTests` (deleted after the run) merging a code-defined subclass and the project's `x:Class` `Subclassed_Dictionary`, dumping `GetElementProperty<string>(d, "OriginalSourceLocation")` for each, built with `-p:UnoForceHotReloadCodeGen=true`:

| Dictionary | Location read back |
|---|---|
| the page's `<Page.Resources><ResourceDictionary>` | `Temp_SourceLocation.xaml#L6:4` |
| merged **code-defined** subclass (the Uno.Themes shape) | `Temp_SourceLocation.xaml#L8:6` — the declaring line ✅ |
| merged **`x:Class`** subclass | `Subclassed_Dictionary.xaml#L1:2` — **its own** file, not the use site (`#L9:6`) ✅ |
| `Test_Page_Other`'s `Resources = new Subclassed_Dictionary()` | `Subclassed_Dictionary.xaml#L1:2` |
| `Application.Current.Resources` | `App/App.xaml#L12:4` |

So the ordering argument holds in practice: the generated code-behind constructor calls `InitializeComponent()` (which stamps `this`) before the apply block runs, and the guard preserves it.

Committed coverage: `SetOriginalSourceLocationInOutputForTypedResourceDictionaries` pins the guarded emission for both use sites. Full generator suite is 471 passed / 0 failed, and `Uno.UI.UnitTests` under `UnoForceHotReloadCodeGen=true` still matches the `master` baseline exactly (3949 passed; same 137 Linux-only path/UNC failures).

The spec now carries this as R5, and the "typed dictionary subclasses" non-goal is replaced by the narrower invariant — a declaration site is never overwritten by a reference site.

src/SourceGenerators/Uno.UI.SourceGenerators.Tests/XamlCodeGeneratorTests/Out/Given_HotReloadEnabledInBuild/SeOrSoLoInOuFoReDiFi/XamlCodeGenerator_OtherDictionary_218afd2e503a1577f4ce46cd7f3fc921.cs
```diff
@@ -0,0 +1,111 @@
...
+		private const string __baseUri_prefix_OtherDictionary_218afd2e503a1577f4ce46cd7f3fc921 = "ms-appx:///TestProject/";
+		[global::System.ComponentModel.EditorBrowsable(global::System.ComponentModel.EditorBrowsableState.Never)]
+		private const string __baseUri_OtherDictionary_218afd2e503a1577f4ce46cd7f3fc921 = "ms-appx:///TestProject/0/OtherDictionary.xaml";
+		// This non-static inner class is a means of reducing size of AOT compilations by avoiding many accesses to static members from a static callsite, which adds costly class initializer checks each time.
+		[global::System.ComponentModel.EditorBrowsable(global::System.ComponentModel.EditorBrowsableState.Never)]
+		internal string __OtherDictionary_218afd2e503a1577f4ce46cd7f3fc921_checksum() => "87413f85bfe7702d0b326695f480cd59dd74d40a";
+		[global::System.Runtime.CompilerServices.CreateNewOnMetadataUpdate]
+		internal sealed class ResourceDictionarySingleton__OtherDictionary_218afd2e503a1577f4ce46cd7f3fc921 : global::Uno.UI.IXamlResourceDictionaryProvider
+		{
+			private static global::Microsoft.UI.Xaml.NameScope __nameScope = new global::Microsoft.UI.Xaml.NameScope();
```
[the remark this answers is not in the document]
DerekRaven: These files under `XamlCodeGeneratorTests/Out/**` are committed snapshots of **generator output**, compared byte-for-byte by the tests — they are not hand-written source, and editing one would just make the test fail against what the generator actually emits.

The `private static global::Microsoft.UI.Xaml.NameScope __nameScope = …` field is emitted by `BuildTopLevelResourceDictionary` for every XAML dictionary in the repo and predates this PR (it is in the pre-existing `ReDiCoBe` snapshot too). Making it `readonly` would be a generator-wide output change, unrelated to stamping `OriginalSourceLocation`, so I am leaving it out of scope here.

src/SourceGenerators/Uno.UI.SourceGenerators.Tests/XamlCodeGeneratorTests/Out/Given_HotReloadEnabledInBuild/SeOrSoLoInOuFoExReDi/XamlCodeGenerator_MainPage_0e3f323f9a22a3699cbcd4f0217eee4a.cs
```diff
@@ -0,0 +1,211 @@
...
+
+			Bindings = new MainPage_Bindings(this);
+			((global::Microsoft.UI.Xaml.FrameworkElement)this).Loading += __UpdateBindingsAndResources;
+			((global::Microsoft.UI.Xaml.FrameworkElement)this).Unloaded += __StopTracking;
+		}
+		partial void OnInitializeCompleted();
+
+		[global::System.Diagnostics.CodeAnalysis.UnconditionalSuppressMessage("Trimming", "IL2026", Justification = "Generated code")]
+		[global::System.Diagnostics.CodeAnalysis.UnconditionalSuppressMessage("Trimming", "IL2111", Justification = "Generated code")]
+		private void ApplyTo_Pag_ResΞ0_ResDic_TheDicΞ0_ResDic(global::Microsoft.UI.Xaml.ResourceDictionary __p1, MainPage __that, global::Microsoft.UI.Xaml.NameScope __nameScope)
```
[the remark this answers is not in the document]
DerekRaven: Same as above: this is a committed snapshot of **generator output**, not hand-written source.

The `__nameScope` parameter of the `ApplyTo_*` methods is the shape `XamlLazyApplyBlockIIndentedStringBuilder` emits for every apply block in every XAML file (`.GenericApply(__that, __nameScope, ApplyTo_…)`), and shadowing the class field is intentional there — the apply method is deliberately non-capturing so Hot Reload can replace it. Renaming it would be a generator-wide output change, out of scope for this PR.

DerekRaven:
Ran an eight-lens review panel over the branch. Four findings landed in 63434c8ffb; the rest are recorded below with a disposition.

### 1. The "which dictionary" rule was derived twice, and the copies disagreed

`FindResourcesDictionaryDeclaration` re-derived what `RegisterAndBuildResources` already decides. The counterexample:

```xml
<Page.Resources>
  <ResourceDictionary Source="ms-appx:///…/Other.xaml" />
  <ResourceDictionary><x:Double x:Key="A">1</x:Double></ResourceDictionary>
</Page.Resources>
```

`RegisterAndBuildResources` takes `Objects.First()`, so it emits `Resources = RetrieveDictionaryForSource(…)` — the page's dictionary *is* the referenced file's shared instance — while the old predicate found the second, `Source`-less declaration and stamped that shared instance with this file's location. Exactly the mis-attribution the `Source=` non-goal exists to prevent.

Fixed by reading the member once: a `ResourcesMember` record produced by `FindResourcesMember`, which `RegisterAndBuildResources` now emits from and whose `OwnDictionaryDeclaration` the stamps use — non-null exactly when the owner keeps the dictionary it creates itself. Two consequences worth noting: an entirely empty `<ResourceDictionary/>` is no longer stamped (so R3 no longer forces a dictionary to materialize — the edge case in the spec is gone rather than accepted), and `<ResourceDictionary Source="…">` that *also* carries `MergedDictionaries` is now stamped, because that shape does populate the owner's own dictionary. New snapshot test `SetOriginalSourceLocationNotSetForResourcesFromSource` pins the counterexample: the generated file contains no `OriginalSourceLocation` at all.

The name-based `Type.Name == "ResourceDictionary"` match (rather than symbol resolution) is now deliberate and commented — it is the same test the emission uses, so the two cannot disagree about an aliased or same-named foreign type.

### 2. Unescaped path interpolated into a C# literal

`_fileDefinition.FilePath.Replace("\\", "/")` was the only sanitization. A quote or newline is legal in a file name on unix-like systems, so `a".xaml` breaks the build and a crafted name injects C# into generated source. The URIs now go through `SymbolDisplay.FormatLiteral(…, quote: true)` (`FileUriLiteral` / `GetSourceLocationLiteral`), and `SetBaseUri`'s URI argument — same pre-existing hazard — goes through the same helper.

Not changed: a `#` in a project path still makes the `#L<line>:<pos>` fragment ambiguous to a parser. Escaping it would change the value format every existing consumer already reads, so it wants a separate decision.

### 3. Nothing committed asserted the behaviour

The panel was right that golden snapshots freeze text and can't tell a correct location from a confidently wrong one, and that deleting the semantic probe was the wrong call. `Uno.UI.UnitTests` already sets `UnoForceHotReloadCodeGen`, so the probe is now a committed test — `Given_ResourceDictionarySourceLocation` over `App/Xaml/ResourceDictionarySourceLocation.xaml`:

| Case | Asserted |
|---|---|
| the page's `<Page.Resources><ResourceDictionary>` | `…#L6:4` |
| the `Grid`'s own dictionary (initializer path) | `…#L18:5` |
| merged **code-defined** subclass | `…#L8:6` — the declaring markup |
| merged **`x:Class`** subclass | `Subclassed_Dictionary.xaml`, not this file |
| merged dictionary from a `Source` | not attributed to this file |
| `Application.Current.Resources` | `App.xaml` |

That also closes the `Application` coverage gap the spec had flagged as untestable — the generator harness still can't host an `Application` root, but this asserts the emission where it actually runs. Exact line:column values are intentional: they're what catches a stamp landing on the neighbouring element.

### 4. `<applied>.Resources` was stamped for any object with a member named `Resources`

`Resources` is declared by `FrameworkElement` and `Application` only, so the stamp is now gated on `isFrameworkElement` (a local `BuildExtendedProperties` already computes). After fix 1 the stamp also only fires where the generator itself populates that dictionary, so the null-getter path is doubly closed.

### Not changed, with reasons

- **`preserveExisting` → `MarkupHelper.SetElementPropertyIfUnset`** — worth doing (it would also make the check atomic and avoid the read attaching a null entry), but it adds public runtime surface for an out-of-repo consumer; better as its own PR than smuggled into this one. Same for naming `"OriginalSourceLocation"` in a shared constant and documenting the `file:///…#L:` grammar on that member.
- **App.xaml stamp sits outside the `SetResolutionContext(__currentAlc)` block** — real, and I'd rather move it in a change that can be exercised against a secondary-ALC app than blind.
- **Generator-time LINQ in `BuildExtendedProperties`** — `FindResourcesMember` is short-circuited by `_isHotReloadEnabled &&`, so it costs nothing in Release codegen; de-LINQ-ing it is a good follow-up under the source-generator rules.
- **Feature-scoped kill switch** — reasonable, but the exposure is bounded: `UnoForceHotReloadCodeGen` appears in no shipping project, so `Uno.UI`'s 261 dictionaries and the Fluent bundles emit nothing in packed Release builds.
- **An `x:Class` dictionary file yields two stamped instances with the same location** (the `x:Class` type and the file-level one in `GlobalStaticResources`) — pre-existing duality, now visible through this property. Worth documenting for consumers rather than disambiguating unilaterally.

### After the fixes

Generator suite 486 total / 472 passed / 0 failed — regenerating the snapshots changed no pre-existing golden. `Uno.UI.UnitTests` 4114 total / 3954 passed, with the 137 failures byte-identical to the `master` baseline (Linux-only path/UNC expectations).

DerekRaven:
Answers to the three questions the review panel left open, with the changes they produced.

### The consumer requirement is now confirmed, not inferred

The panel's top objection was that no code in this repo reads the stamp, so "a designer holding a dictionary can't tell which file declared it" was an assumption. It is now measured. Hot Design's `SelectDesignSystem` was changed to derive its identifier the ordinary way (which reads `OriginalSourceLocation`), the generator from this branch was applied over the `uno.winui` 6.7.16 analyzer in a real app, and the designer was driven by hand in the `MvvmMaterialFrame` sample — whose App.xaml merges `<utum:MaterialToolkitTheme …/>`, a **code-defined** `Uno.Themes.BaseTheme` subclass with no `InitializeComponent` of its own:

```
theme=Uno.Toolkit.UI.Material.MaterialToolkitTheme
stampedLocation='file:////…/MvvmMaterialFrame/App.xaml#L14:6'
identifier: valid=True editable=True file='…/App.xaml' line=14:6
readOnly=False subject=MaterialToolkitTheme valid=True editable=True
```

`#L14:6` is the `<utum:MaterialToolkitTheme>` element. The property grid went from uniformly read-only to editable — the exact case raised in review. Without R5 that instance carries no location at all.

### The format now has an owner (`a1feaf6062`)

The grammar was authored in *two* repos: produced here as `file:///<path>#L<line>:<position>`, and re-created on the consumer side without the `file:///` prefix (`XamlFileLocationHelper.FormatFileLocation`), which also strips the prefix on parse. Nothing pinned either half, and the string is not merely a label — the consumer matches markup elements by it.

It is now declared once, as `MarkupHelper.OriginalSourceLocationPropertyName`, whose remarks document the grammar and the two properties a consumer depends on. The generator keeps a private copy of the name (it cannot reference `Uno.UI`) pointing at the runtime constant. Closes the "undocumented magic string" and "no shared constant" findings too.

### `#` in a path: nothing to escape here — but read the fragment from the end

I probed the real consumer parser with seven inputs:

| Input | Result |
|---|---|
| `file:////repo/app/App.xaml#L14:6` | OK |
| `file:///C:/repo/app/App.xaml#L14:6` | OK (Windows drive) |
| `file:////repo/C#App/App.xaml#L14:6` | **OK** — a bare `#` is harmless |
| `file:////repo/C#Lib/App.xaml#L14:6` | **throws `IndexOutOfRangeException`** |
| `file:////repo/my#IDcards/App.xaml#L14:6` | OK for this function |
| `file:////repo/a b/Ünïcode/App.xaml#L14:6` | OK — spaces and non-ASCII survive |
| `file:////repo/app/App.xaml#L14:6#L20:2` | OK — takes the *first* fragment |

So percent-encoding is the wrong fix: spaces, non-ASCII, `#` and drive letters all round-trip today precisely because nobody treats the value as a URI, and encoding would break every consumer, none of which decodes. The real defect is a consumer splitting on **all** occurrences of `#L` and indexing `[1]`; a path segment containing `#L` (a `C#Lib` folder) then throws. That is fixed by parsing right-anchored, and it belongs in the consumer — it needs no wire change. The producer's contract now says so explicitly in the constant's remarks, and a snapshot test pins that the path reaches the literal verbatim (a quote can't be covered in this harness: Roslyn rejects it in the `hintName` first, and only a directory could carry one).

### Also settled: can the `Resources` stamp have a second writer? (`39997f0108`)

R3 writes unconditionally while R5 is set-if-absent, which looked asymmetric. It isn't: the two `Resources`-targeting emissions are mutually exclusive by construction (`BuildApplicationInitializerBody` only for an `Application` root, the `BuildExtendedProperties` one gated on `isFrameworkElement`, and `Application` is not one), the dictionary is owner-created, and the stamp runs during construction — before any code could alias one element's dictionary onto another. The framework treats a dictionary as singly-owned anyway: `FrameworkElement.Resources`' setter calls `SetResourceOwner(this)`. The only repeat writer is the *same* file re-stamping after a Hot Reload edit, with the declaration's current line — which is why the write must stay unconditional: a guard there would pin a stale location. Recorded in the spec.

### Still open, and needing a human

- **Why was #15846 closed with the `x:Class`-only fix?** If that was a deliberate scope call, this PR reopens it on inference.
- **Should the declaration site win for a dictionary the user cannot edit?** Set-if-absent points the designer at markup inside a package for a library's `x:Class` dictionary, where the user's own reference site was the actionable location. Today's case was the code-defined subclass, where the declaring line *is* in the user's file — so the question stands for the library variant only.

Deferred by choice, all recorded above: `SetElementPropertyIfUnset` in `MarkupHelper`, the ALC scope of the App.xaml stamp, de-LINQ-ing `FindResourcesMember`, and a feature-scoped kill switch.

src/Uno.UI/Helpers/MarkupHelper.cs
```diff
@@ -97,6 +97,29 @@ public static IXamlServiceProvider CreateParserContext(object? target, Type prop
...
+		/// sequence "#L" (a "C#Lib" folder, say) — so a parser that splits on the first occurrence
+		/// mis-reads such a path, and one that splits on every occurrence fails outright.
+		/// </para>
+		/// <para>
+		/// The value is not URI-escaped beyond the separator replacement: spaces and non-ASCII
+		/// characters appear verbatim, so it must not be handed to a URI parser as-is.
+		/// </para>
+		/// </remarks>
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		public const string OriginalSourceLocationPropertyName = "OriginalSourceLocation";
```
carldebilly: **HIGH** — This new member is `public const`, and neither half of that is what you want here.

On visibility: it has zero in-repo consumers (the generator deliberately keeps its own copy, and the new test hardcodes the literal), and `src/Uno.UI/AssemblyInfo.cs:35` already grants `InternalsVisibleTo("Uno.UI.HotDesign.Client")` — so the out-of-repo consumer this exists for needs no public surface. The repo has no `PublicAPI.*.txt` and no ApiCompat gate, so nothing catches this mechanically, which also makes the checklist's "no public API surface" (and #24254's "no new runtime API") inaccurate. `[EditorBrowsable(Never)]` hides it from IntelliSense but it still ships in the reference assembly and binds forever.

On `const`: the value is baked into each consumer at *its own* compile time, so the "declared once" guarantee does not hold across the very assembly boundary the member exists to serve — a consumer built against today's Uno.UI keeps the old literal if the key ever changes while the generator emits the new one. That is the silent drift the doc block sets out to prevent.

Suggested direction: `internal static readonly string` behind the existing IVT. `Uno.UI.UnitTests` is IVT'd too (`AssemblyInfo.cs:7`), so the test below can still consume it. Knock-on edits: tick the docs checkbox or restate the n/a basis; if a parse helper is ever added next to it (see the general comment), keep it `internal` behind the same IVT rather than widening the surface again.

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.cs
```diff
@@ -1089,14 +1097,57 @@ private void BuildChildSubclasses(IIndentedStringBuilder writer, bool isTopLevel
...
+				}
 			}
 		}
 
+		/// <summary>
+		/// Name of the weak property carrying an object's XAML declaration site. Must match
+		/// <c>MarkupHelper.OriginalSourceLocationPropertyName</c>, which documents the value's format
+		/// for its consumers — the generator cannot reference it, as it does not reference Uno.UI.
+		/// </summary>
+		private const string OriginalSourceLocationPropertyName = "OriginalSourceLocation";
```
carldebilly: **MEDIUM** — The comment here (and `spec.md` §192) says the generator cannot reference the runtime constant, so it keeps a private copy. But that mechanism already exists and is in use: `Uno.UI.SourceGenerators.csproj:62` links `..\..\Uno.UI\UI\Xaml\XamlFilePathHelper.shared.cs` into the generator for exactly this purpose, and line 1634 of this file consumes `XamlFilePathHelper.LocalResourcePrefix`.

So the string now has three independent declarations — generator, `MarkupHelper`, and the new unit test — across the boundary the PR set out to unify.

Suggested direction: put the name in a linked `.shared.cs` (netstandard2.0-clean, no `Microsoft.UI.Xaml` types) and have `MarkupHelper.OriginalSourceLocationPropertyName` forward to it. Knock-on edits: this comment and `spec.md` §"format is owned by the producer" both need rewriting; no snapshots change, since the emitted text is identical. This also removes the need for a test asserting the two copies agree.

src/Uno.UI.UnitTests/Windows_UI_Xaml/Given_ResourceDictionarySourceLocation.cs
```diff
@@ -0,0 +1,71 @@
...
+	/// <remarks>
+	/// This project sets UnoForceHotReloadCodeGen, so the locations are always generated here.
+	/// </remarks>
+	[TestClass]
+	public class Given_ResourceDictionarySourceLocation
+	{
+		private const string DeclaringFile = "ResourceDictionarySourceLocation.xaml";
+
+		private static string GetSourceLocation(ResourceDictionary dictionary)
+			=> MarkupHelper.GetElementProperty<string>(dictionary, "OriginalSourceLocation") ?? "";
```
carldebilly: **MEDIUM** — Still open from the previous round: this passes the literal `"OriginalSourceLocation"` although the last commit added `MarkupHelper.OriginalSourceLocationPropertyName` precisely so the key is declared once, and this file already has `using Uno.UI.Helpers;`.

What is new since that comment: the constant has zero consumers repo-wide, `spec.md:158` states that consumers read the key *through the constant*, and the spec's own prose is inconsistent the same way (§Overview and §Test plan quote the literal, §Non-goals quotes the constant). The one place that could demonstrate the single-declaration claim is the place that contradicts it — and with a third literal here, nothing in the repo fails if the key is ever renamed.

Suggested direction: use `MarkupHelper.OriginalSourceLocationPropertyName`, and align the spec's prose quotes with it. No other edits needed. This does not argue for making the generator reference `Uno.UI` — see the separate comment on its private copy.

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.cs
```diff
@@ -2714,70 +2784,101 @@ private string GetFullGenericTypeName(INamedTypeSymbol? propertyType)
...
+				return null;
+			}
+
+			// To be able to have MergedDictionaries, the first node of the Resource node
+			// must be an explicit resource dictionary.
+			// Note: matching the type by name, rather than resolving it, is what makes an aliased or
+			// same-named foreign type behave identically here and in the emission below.
+			var explicitDictionary = resourcesMember.Objects.Any(o => o.Type.Name == "ResourceDictionary")
+				? resourcesMember.Objects.First()
+				: null;
```
carldebilly: **MEDIUM** — `ExplicitDictionary` takes `Objects.First()` unconditionally, while `Source` two lines down reads `FirstOrDefault(o => o.Type.Name == "ResourceDictionary")`. Two different selection rules in one construction, and the first one now feeds a stamped location.

Failure shape: for `<Page.Resources><Style/><ResourceDictionary>…</ResourceDictionary></Page.Resources>`, the R3 stamp points at the `<Style>` element's line — a dictionary attributed to an element that is not one. More broadly, this method exists so the decision cannot drift, and it currently derives "the explicit dictionary" three ways in ten lines: `Any(pred) ? First() : null`, `FirstOrDefault(pred)`, and `IsResourceDictionarySubclass`.

Suggested direction, kept narrow so the emitted text stays byte-identical: leave `Root`/`Merged`/`Theme` on `Objects.First()`, and have `ExplicitDictionary` — used solely for the stamp — read the first actual `ResourceDictionary`, reusing one local for both it and `Source`. Knock-on edits: no snapshots move (behaviour is unchanged on valid markup); keep the name-vs-symbol comment accurate. While here, hoisting the `ResourcesMember` next to the existing `RegisterAndBuildResources` call would remove the second `FindResourcesMember` walk per element that `BuildExtendedProperties` now performs, and the construction reads better with named arguments.

src/Uno.UI.UnitTests/Windows_UI_Xaml/Given_ResourceDictionarySourceLocation.cs
```diff
@@ -0,0 +1,71 @@
...
+			// The instance belongs to the referenced file and is shared, so this file must not claim it.
+			Assert.IsFalse(location.Contains(DeclaringFile), $"Expected no location of this file, got '{location}'");
+		}
+
+		[TestMethod]
+		public void When_Application_Resources_Then_Location_Is_The_Dictionary_Declaration()
+		{
+			var app = UnitTestsApp.App.EnsureApplication();
+
+			StringAssert.Contains(GetSourceLocation(app.Resources), "App.xaml#L");
```
carldebilly: **MEDIUM** — This is the only coverage of the `Application` path — the one requirement case with no snapshot, since the generator harness cannot host an App.xaml root — and it is also the weakest assertion of the five: `Contains("App.xaml#L")` only.

Concretely: passing `topLevelControl` instead of `resourcesDeclaration` at line 567 would emit `App.xaml#L1:2`, and this test would still pass. The path with no golden is therefore the path with no real guard.

Suggested direction: assert the exact location, `EndsWith("App.xaml#L12:4")` (line 12, column 4 — two tabs), matching the exactness of the other four cases. Knock-on edit: that pins App.xaml's `Application.Resources` line, so it is worth a note in the fixture that the line must not move.

src/Uno.UI.UnitTests/Windows_UI_Xaml/Given_ResourceDictionarySourceLocation.cs
```diff
@@ -0,0 +1,71 @@
...
+
+		[TestMethod]
+		public void When_Merged_Dictionary_From_Source_Then_Location_Is_Not_The_Declaring_Markup()
+		{
+			var sut = new ResourceDictionarySourceLocation();
+			var location = GetSourceLocation(sut.Resources.MergedDictionaries[2]);
+
+			// The instance belongs to the referenced file and is shared, so this file must not claim it.
+			Assert.IsFalse(location.Contains(DeclaringFile), $"Expected no location of this file, got '{location}'");
+		}
```
carldebilly, line 66: **MEDIUM** — The `Source=` exclusion test is purely negative, and the helper coalesces a missing property to `""` — so these assertions pass when nothing at all is stamped. Point `Source=` at a nonexistent file, or drop the referenced file's own R2 stamp, and the test stays green: it cannot distinguish correct attribution from no location whatsoever, which is exactly the regression it is meant to guard.

Suggested direction: add the positive leg — `StringAssert.Contains(location, "Test_Dictionary.xaml#L")` — so the referenced file's own declaration is pinned alongside the exclusion. Alternatively assert absence explicitly through the raw `GetElementProperty` returning null. No knock-on edits.

specs/057-resourcedictionary-source-location/spec.md
```diff
@@ -0,0 +1,272 @@
...
+|---|---|
+| `InitializeAndBuildResourceDictionary`: apply block + `TrySetOriginalSourceLocation` after the initializer | R1 |
+| `BuildTopLevelResourceDictionary`: `TrySetOriginalSourceLocation` on the `CreateWithCapacity` field | R2 |
+| `BuildExtendedProperties`: stamp `<applied>.Resources`, gated on `isFrameworkElement`, from `FindResourcesMember` | R3 |
+| `BuildApplicationInitializerBody`: same, on `Resources` | R3 |
+| `BuildTypedResourceDictionary`: apply block + set-if-absent stamp | R5 |
+| `TrySetOriginalSourceLocation`: `preserveExisting` option, emitting the `GetElementProperty … is null` guard | R5 |
+| `MarkupHelper.OriginalSourceLocationPropertyName`: the key and the documented grammar, plus the generator's private copy of the name | R1–R5 |
+| new `ResourcesMember` record + `FindResourcesMember`, which `RegisterAndBuildResources` now emits from and the R3 stamps read | R3 |
+| `FileUri` / `FileUriLiteral` / `GetSourceLocationLiteral`: the emitted URIs go through `SymbolDisplay.FormatLiteral` | R1–R5, `SetBaseUri` |
```
carldebilly, line 279: **MEDIUM** — The "Validation performed" section pins one local run's counts (`486 total, 472 passed, 14 skipped`; `4114 total, 3954 passed, 23 skipped`), and the Implementation section restates the diff as a table of method names. Nothing keeps either in sync with the code, and the counts already disagree with the PR body (471 and 3949) — the drift has arrived before the merge.

Suggested direction: keep the requirements and the test plan, which are the durable part, and replace the run log with a pointer to the `Given_HotReloadEnabledInBuild` suite. Prose-only change, no knock-on edits.

specs/057-resourcedictionary-source-location/spec.md
```diff
@@ -0,0 +1,272 @@
+# Feature Specification: `OriginalSourceLocation` for `ResourceDictionary`
+
+**Repo**: `uno` (Uno.UI.SourceGenerators — XAML generator)
+**Created**: 2026-08-27
+**Status**: Implemented (working tree, not committed)
```
carldebilly: **LOW** — Still open from the previous round: `Status: Implemented (working tree, not committed)`, in the commit that commits the spec. A tracked source-of-truth document that denies its own existence misleads the next reader about whether the behaviour is live.

Suggested direction: set the status to `Implemented` and drop the parenthetical. No knock-on edits.

carldebilly (approved):
Reviewed across architecture, contract, correctness, quality, operability, security and style lenses. **Approving** — nothing here endangers the mechanism, and most of what follows is a one-line edit. Counts: 1 high, 7 medium, 2 low, plus 9 lower-confidence signals and 4 nits.

**What holds up.** The emission is in the right place and correctly gated: every new site is inside the pre-existing `_isHotReloadEnabled` guard, and `BuildXamlApplyBlocks` only emits handler types when HR is off, so "byte-identical when off" survives — exactly one pre-existing golden moved. `ResourcesMember` reproduces master's branch order and predicates. Line/position values check out character-by-character against the snapshots (`#L8:10`, `#L6:6`, `#L17:8`) and the tab-indented runtime fixture. The snapshots are a real assertion, not decoration: `CSGenerator` compares embedded `Out/**` resources byte-for-byte and throws on an unexpected or missing file, and `WRITE_EXPECTED` is `#error`-guarded, so a regressed or moved stamp fails the suite. Neither `Application.Resources` nor `FrameworkElement.Resources` can return null, so the new dereferences cannot throw. And `SymbolDisplay.FormatLiteral` closes a real pre-existing injection sink: master interpolated the raw path into the emitted literal at two sites, and no new emission site bypasses the escaping.

**The embedded XAML sources channel is intact.** `XamlCodeGeneration.XamlSources.cs` is untouched, the new `XamlCodeGenerator_EmbeddedXamlSources.cs` files are per-test snapshots structurally identical to the existing ones, and nothing changes the file set, the list hash or `NormalizePath`. The two channels stay mutually resolvable — provider key `//Project/0/Ma in Pägé.xaml` vs stamp `file://///Project/0/Ma in Pägé.xaml#L1:2`: stripping the literal `file:///` prefix and the trailing fragment lands exactly on the key, non-ASCII and spaces included. Two gaps around that, both cheap: nothing *asserts* the relationship (the snapshots assert each channel in isolation, though both are now snapshotted side by side), and the new doc block does not state the join a consumer needs — strip `file:///`, compare `OrdinalIgnoreCase`.

**Main theme.** The one finding worth resolving before merge is the shape of the new key member — see the inline comment on `MarkupHelper.cs`. The rest is prose that has drifted between commits, plus two items from the previous round that are still open (`spec.md:5` and the hardcoded key in the new test).

**Two findings with no single line to anchor:**

- *Validation counts in the PR body disagree with the spec and with the tree.* The body says "three new snapshot tests"; there are five (`Given_HotReloadEnabledInBuild.cs` 387/427/487/548/600, and five new snapshot folders). It says 471 passed where the spec says 472, and 3949 where the spec says 3954. The body was written at the first commits and not refreshed as later ones added tests. Restating them once from the tree — or dropping them and pointing at the suite — would make the validation section verifiable again. The Hot Design run is worth keeping as a manual observation, labelled as such, since the throwaway page leaves no artefact to re-run.
- *Issue #24254 still specifies excluding typed subclasses,* while this PR deliberately stamps them set-if-absent. The divergence and its reason (a code-defined library theme dictionary has no `InitializeComponent`, so excluding it by type leaves it stamped by nobody) are convincing, and are recorded in `spec.md` R5 and the commit body — but the requirement of record still reads the other way. A comment on the issue would close that.

**Lower-confidence signals**, listed once so they are not lost: `TrySetOriginalSourceLocation`'s `element` parameter now receives an arbitrary expression (`__p1.Resources`) while `preserveExisting` emits it twice — harmless today since no call site combines both, but worth documenting as side-effect-free or hoisting to a local in the emitted block. `FileUriLiteral` re-runs `SymbolDisplay.FormatLiteral` per `FrameworkElement` while `FileUri` right above it memoizes. The documented grammar `file:///<path>` omits that an already-rooted unix path yields `file:////`, as the quoted sample output shows. A dictionary file whose root is a typed subclass would take R5's set-if-absent path rather than R2's unconditional write, so attribution would depend on the root's shape — worth a snapshot test. The comment at `XamlFileGenerator.cs:1933` invokes "the `InitializeComponent` its constructor just ran", but the generator emits no constructor; R5's precedence rests on user-authored code-behind. And `XamlCodeGeneration.XamlSources.cs:86` interpolates `f.FilePath` into a verbatim literal without doubling quotes — the same class of hazard this PR just fixed one file over, pre-existing and out of scope here, but adjacent enough to be worth a follow-up.

**Nits:** the six-member `ResourcesMember` is constructed positionally (named arguments would make the roles readable); the two new files lack `#nullable enable` and use braced namespaces where the sibling test file is file-scoped; the runtime assertions hard-code fixture line/column literals and `MergedDictionaries` indexes, so one inserted XAML line breaks three tests with an opaque message; and the five new test methods carry no `[Description]` — though there are none in these suites today, so local practice arguably wins.

DerekRaven:
Thanks — the review pass is in `3638afd112`. Seven of the eight inline findings are fixed; one I could not reproduce and have reported below rather than claim.

### Fixed

**HIGH — `MarkupHelper.cs`: `public const` was wrong on both halves.** Both arguments land, and the `const` one especially: the member existed to prevent drift and I chose the one modifier that guarantees it across the assembly boundary it serves. Now `internal static readonly`, reached through the existing `InternalsVisibleTo`, so the checklist's "no public API surface" is accurate again.

**MEDIUM — the generator *can* reference a shared declaration.** You were right that my rationale was factually wrong: `Uno.UI.SourceGenerators.csproj` already links `XamlFilePathHelper.shared.cs` and line 1634 consumes `LocalResourcePrefix`. The name now lives there, `MarkupHelper.OriginalSourceLocationPropertyName` forwards to it, and the three declarations are one. Emitted text is unchanged — no snapshot moved.

**MEDIUM — the test hardcoded the key.** Now reads `MarkupHelper.OriginalSourceLocationPropertyName`, so the constant has a consumer and a rename fails the build.

**MEDIUM — the `Application` assertion was the weakest where it mattered most.** Exactly right that `Contains("App.xaml#L")` would survive passing `topLevelControl`. Now `EndsWith("App.xaml#L12:4")`, with a comment recording that it pins line 12 column 4 of `App/App.xaml` and why the exactness is deliberate.

**MEDIUM — the `Source=` test could pass with nothing stamped.** Added the positive leg: the merged instance must carry `Test_Dictionary.xaml#L…`, i.e. the referenced file's own declaration. It passes, which also confirms R2 stamps that file-level dictionary.

**MEDIUM — the run log in the spec.** Replaced with a pointer to the suites, keeping only the two things they cannot cover (the `Application` path's exactness, and the Hot Design observation, labelled as manual). The PR body's counts are corrected too — you were right that it said three snapshot tests where there are five, and that its numbers had drifted from the spec's.

**LOW — the spec status.** Now `Implemented`.

### Changed, but I could not reproduce the failure

**MEDIUM — `ExplicitDictionary` vs `Source` selection rules.** I made the change (the stamp now reads the first actual `ResourceDictionary`, sharing one local with `Source`), because the maintainability point stands — three derivations in ten lines in the method that exists to stop the decision drifting.

But the failure shape does not reach a stamp, and I could not make it. `PopulatesOwnDictionary` is driven by `Root`/`Merged`/`Theme`, all read from that same first node — so it is false in exactly the cases where `ExplicitDictionary` would have been wrong. I tried both content shapes for `<Page.Resources><X/><ResourceDictionary>…`:

- `<TextBlock x:Key="FirstNode" Text="…" />` → no resources emitted at all, no stamp
- `<x:String x:Key="FirstNode">…</x:String>` → same

In both, `RegisterAndBuildResources` emits nothing for that member (pre-existing behaviour I have not changed), so there is no stamp to mis-attribute. I wrote a snapshot test for it, saw it prove nothing either way, and **deleted it** rather than commit a test that cannot fail — which would have been the same defect as the two you flagged above. If you know a shape that does reach the stamp with a non-dictionary first node, I will happily pin it.

### Also taken from the review body

The four-slash reality (`file:////` for an already-rooted path) and the embedded-sources join (strip the prefix and the fragment, compare `OrdinalIgnoreCase`) are now in the doc block — that join was the useful half I had left out.

### Not taken here

Left as follow-ups, deliberately: `FileUriLiteral` re-running `FormatLiteral` per element while `FileUri` memoizes; the `:1933` comment invoking a constructor the generator does not emit; a snapshot for a dictionary file whose root is a typed subclass; and `XamlCodeGeneration.XamlSources.cs:86`'s unquoted path interpolation — pre-existing and one file over, but the same class as what this PR fixed, so worth its own issue. Named arguments on `ResourcesMember` are in; `#nullable enable` is on the new test file.

I will comment on #24254 to record that the issue text still specifies excluding typed subclasses while this PR stamps them set-if-absent, with the reason.

src/Uno.UI/Helpers/MarkupHelper.cs
DerekRaven: Fixed in 3638afd112 — now `internal static readonly`, forwarding to the linked `XamlFilePathHelper`. Both halves of your point landed; the `const` one especially, since the member existed to prevent exactly the drift a compile-time constant guarantees across the assembly boundary it serves. The checklist's "no public API surface" is accurate again.

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.cs
DerekRaven: Fixed in 3638afd112. You were right that the rationale was wrong — the csproj already links `XamlFilePathHelper.shared.cs` and line 1634 consumes `LocalResourcePrefix` from it. The name now lives there and `MarkupHelper.OriginalSourceLocationPropertyName` forwards to it, so the three declarations are one. Emitted text is unchanged, no snapshot moved. The comment here and the spec section are rewritten.

src/Uno.UI.UnitTests/Windows_UI_Xaml/Given_ResourceDictionarySourceLocation.cs
DerekRaven: Fixed in 3638afd112 — the test now reads `MarkupHelper.OriginalSourceLocationPropertyName`, so the constant has a consumer and a rename fails the build. The spec prose quotes it consistently too.

src/Uno.UI.UnitTests/Windows_UI_Xaml/Given_ResourceDictionarySourceLocation.cs
DerekRaven: Fixed in 3638afd112 — now `EndsWith("App.xaml#L12:4")`, with a comment recording that it pins line 12 column 4 of `App/App.xaml` (the two-tab `<ResourceDictionary>`) and why the exactness is deliberate. You were right that the loose form would have survived passing `topLevelControl`.

src/Uno.UI.UnitTests/Windows_UI_Xaml/Given_ResourceDictionarySourceLocation.cs
DerekRaven: Fixed in 3638afd112 — added the positive leg, `StringAssert.Contains(location, "Test_Dictionary.xaml#L")`, so the referenced file's own declaration is pinned alongside the exclusion. It passes, which also confirms R2 stamps that file-level dictionary.

specs/057-resourcedictionary-source-location/spec.md
DerekRaven: Fixed in 3638afd112 — the run log is gone, replaced by a pointer to the suites, keeping only what they cannot cover (the `Application` path's exactness, and the Hot Design run labelled as a manual observation). The PR body is corrected too: it said three snapshot tests where there are five, and its counts had drifted from the spec's.

specs/057-resourcedictionary-source-location/spec.md
DerekRaven: Fixed in 3638afd112 — `Status: Implemented`, parenthetical dropped.

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlFileGenerator.cs
DerekRaven: Changed in 3638afd112 — the stamp now reads the first actual `ResourceDictionary`, sharing one local with `Source`, because your maintainability point stands (three derivations in ten lines, in the method that exists to stop the decision drifting).

I could not reproduce the failure, though. `PopulatesOwnDictionary` is driven by `Root`/`Merged`/`Theme`, all read from that same first node, so it is false in exactly the cases where `ExplicitDictionary` would have been wrong. I tried both content shapes of `<Page.Resources><X/><ResourceDictionary>…`:

- `<TextBlock x:Key="FirstNode" Text="…" />` → no resources emitted at all, no stamp
- `<x:String x:Key="FirstNode">…</x:String>` → same

`RegisterAndBuildResources` emits nothing for that member (pre-existing, unchanged here), so there is no stamp to mis-attribute. I wrote a snapshot test for it, saw it prove nothing either way, and deleted it rather than commit a test that cannot fail — the same defect as the two test findings above. Resolving since the change is in, but please reopen if you know a shape that reaches the stamp with a non-dictionary first node and I will pin it.

src/Uno.UI.UnitTests/Windows_UI_Xaml/Given_ResourceDictionarySourceLocation.cs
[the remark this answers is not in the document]
DerekRaven: Fixed in 3638afd112 — the test now reads `MarkupHelper.OriginalSourceLocationPropertyName`. Same point carldebilly raised; the key is also declared once now, in the linked `XamlFilePathHelper`, with `MarkupHelper` forwarding to it.

src/SourceGenerators/Uno.UI.SourceGenerators.Tests/XamlCodeGeneratorTests/Out/Given_HotReloadEnabledInBuild/SeOrSoLoInOuFoTyReDi/XamlCodeGenerator_MainPage_0e3f323f9a22a3699cbcd4f0217eee4a.cs
```diff
@@ -0,0 +1,195 @@
...
+			OnInitializeCompleted();
+
+			Bindings = new MainPage_Bindings(this);
+			((global::Microsoft.UI.Xaml.FrameworkElement)this).Loading += __UpdateBindingsAndResources;
+			((global::Microsoft.UI.Xaml.FrameworkElement)this).Unloaded += __StopTracking;
+		}
+		partial void OnInitializeCompleted();
+		[global::System.Diagnostics.CodeAnalysis.UnconditionalSuppressMessage("Trimming", "IL2026", Justification = "Generated code")]
+		[global::System.Diagnostics.CodeAnalysis.UnconditionalSuppressMessage("Trimming", "IL2111", Justification = "Generated code")]
+		private void ApplyTo_Pag_ResΞ0_ResDic_MerDicΞ0_MyCodDic(global::TestRepro.MyCodeDictionary __p1, MainPage __that, global::Microsoft.UI.Xaml.NameScope __nameScope)
```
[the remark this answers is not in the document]
DerekRaven: Declined — this file is under `XamlCodeGeneratorTests/Out/**`, a committed snapshot of **generator output** that the tests compare byte-for-byte. Editing it would only make the test fail against what the generator actually emits. The `__nameScope` parameter is the shape `XamlLazyApplyBlockIIndentedStringBuilder` gives every apply block in every XAML file, and the shadowing is intentional: the apply method is deliberately non-capturing so Hot Reload can replace it. Changing it would be a generator-wide output change, unrelated to this PR.

src/SourceGenerators/Uno.UI.SourceGenerators.Tests/XamlCodeGeneratorTests/Out/Given_HotReloadEnabledInBuild/SeOrSoLoInOuFoTyReDi/XamlCodeGenerator_MainPage_0e3f323f9a22a3699cbcd4f0217eee4a.cs
```diff
@@ -0,0 +1,195 @@
...
+		{
+			if (global::Uno.UI.Helpers.MarkupHelper.GetElementProperty<string>(__p1, "OriginalSourceLocation") is null)
+			{
+				global::Uno.UI.Helpers.MarkupHelper.SetElementProperty(__p1, "OriginalSourceLocation", "file://///Project/0/MainPage.xaml#L8:10");
+			}
+		}
+
+		[global::System.Diagnostics.CodeAnalysis.UnconditionalSuppressMessage("Trimming", "IL2026", Justification = "Generated code")]
+		[global::System.Diagnostics.CodeAnalysis.UnconditionalSuppressMessage("Trimming", "IL2111", Justification = "Generated code")]
+		private void ApplyTo_PagΞ0_Gri_ResΞ0_MyCodDic(global::TestRepro.MyCodeDictionary __p1, MainPage __that, global::Microsoft.UI.Xaml.NameScope __nameScope)
```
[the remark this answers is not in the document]
DerekRaven: Declined — this file is under `XamlCodeGeneratorTests/Out/**`, a committed snapshot of **generator output** that the tests compare byte-for-byte. Editing it would only make the test fail against what the generator actually emits. The `__nameScope` parameter is the shape `XamlLazyApplyBlockIIndentedStringBuilder` gives every apply block in every XAML file, and the shadowing is intentional: the apply method is deliberately non-capturing so Hot Reload can replace it. Changing it would be a generator-wide output change, unrelated to this PR.

PR #24260, opened 2026-08-28
fix(wasm): Request a record so the splash is removed

### LINKED ISSUE #23586: [WASM Skia] Apps don't start rending app on mobile ("stays" rendering loading page)
### Current behavior 🐛

On Android, initial page render sometimes works, sometimes doesn't (works about half the time). On newer iOS (v18), initial page render doesn't seem to ever work on its own. When it gets stuck showing the loading screen on those platforms, you can tap/slide your finger around the loading screen and can get it to show the initial app page.

On older iOS (v16.x - still supported but max upgradeable version on some older devices) I can't seem to get it to work at all, just stays on the loading screen no matter how much I click or drag my finger around.

The app is actually running (I can see HTTP requests the initial page makes after it loads in the browser debugger), it's just not showing the page, rather staying showing the loading screen instead. The browser debug console does not indicate any errors or that anything is wrong.

It does not have this problem on `6.5.x` SDK versions, nor on desktop browsers, as far as I can tell.

Video showing it "stuck" on the loading screen - after I see the HTTP requests the initial page makes after load in the attached browser debug console, I can tap/move my finger around the screen until the initial page eventually appears. The loading screen stays showing indefinitely until some kind of interaction triggers it to draw the page, though I'm not sure what exactly that interaction is, all I know is that random tapping and dragging eventually makes the page appear:

https://github.com/user-attachments/assets/8f313960-9f17-48c7-8704-4e3782bdae89

### Expected behavior 🎯

_No response_

_No response_

### Workaround 🛠️

_No response_

### Renderer 🎨

- [x] Skia
- [ ] Native

### Affected platforms 📱💻🖥️

WebAssembly

`6.7.0-dev.99`, `6.7.0-dev.165`

### IDE version 🧑‍💻

_No response_

_No response_

MartinZikmund:
**GitHub Issue:** closes #23586

## PR Type:

🐞 Bugfix

## What changed? 🚀

WASM Skia apps intermittently stay on the bootstrap loading screen. The app is fully running — and, as it turns out, **fully rendered**: hiding the loader on a stuck device reveals a correctly laid out, painted app underneath. Only the splash removal never happens. Tapping the screen recovers it.

`WebAssemblyWindowWrapper.ShowCore` waits for `CompositionTarget.FrameRendered`, which is raised from exactly one place — the end of `CompositionTarget.Render()`, i.e. the **record** pass. But to trigger it, it called `((IXamlRootHost)host).InvalidateRender()`, which only asks for a **present**. That never sets `RenderRequested`, and `EnqueueRenderCallback` only reaches `Render()` through that flag:

```csharp
if (_renderedAheadOfTime)
{
    _renderedAheadOfTime = false;
    if (_renderRequestedAfterAheadOfTimePaint) { /* reschedule */ }
    else { /* nothing — no Render(), so no FrameRendered */ }
}
else if (RenderRequested)
{
    RenderRequested = false;
    Render();                 // the only path that raises FrameRendered
}
```

So when the target has already painted ahead of time — the normal state once the tree has settled, which is exactly where `ShowCore`'s continuation runs after the font preload — the requested tick consumes `_renderedAheadOfTime`, records nothing, and `FrameRendered` is never raised. The one-shot subscription waits forever and the loader is never removed. Because `PersistBootstrapperLoader()` disables the bootstrapper's own removal, there is no timeout or fallback.

This also explains the recovery asymmetry: a **tap** dirties the tree, which forces a record and fires the event, while a **resize** whose canvas dimensions are unchanged only produces a present and leaves the app stuck.

The fix routes the request through the API that actually marks the target as needing a record:

```csharp
((ICompositionTarget)compositionTarget).RequestNewFrame();
```

All three states in `RequestNewFrame` converge on a subsequent `Render()`, so `FrameRendered` is guaranteed to be raised after the handler subscribes.

Note this is a different defect from the one addressed by #23590 (which re-armed a dropped `requestAnimationFrame`). That fix was correct as far as it went, but the render pump was not the remaining problem — `requestAnimationFrame` is verifiably alive and firing in ~3 ms in the stuck state, and the frame had already been produced.

### Validation

**What is proven.** On a physical device (Samsung SM-F966B, Android 16, Edge 151) driven over USB via the
DevTools protocol, a stuck instance was captured and inspected live:

- `requestAnimationFrame` firing normally (3 ms), `document.visibilityState: visible`
- .NET runtime fully booted, heap at 386 MB; canvas correctly sized at 1080x2082
- **hiding `.uno-loader` revealed the app fully rendered underneath** — correct layout, text and list

So the frame exists and the render pump is healthy; only the splash removal never happens. That rules out
a dropped `requestAnimationFrame` as the remaining cause and points at the `FrameRendered` subscription,
which the state machine above shows cannot be guaranteed by `InvalidateRender()`. The recovery asymmetry
corroborates it: a tap dirties the tree and forces a record (recovers), while a synthetic `resize` with
unchanged canvas dimensions produces only a present (does not recover).

**What is not proven.** The hang is rare enough that I could not demonstrate the fix statistically. Using a
harness that reloads the app against a deliberately slowed origin (40 ms per asset) and checks whether
`.uno-loader` clears, the observed rate was **1 stuck in 32 runs (~3%) before the fix** and **0 in 24 runs
after**. At a 3% rate, 0-in-24 would occur ~48% of the time by chance, so the post-fix run is not evidence
of anything on its own. The case for this change rests on the captured stuck instance and the state
machine, not on that comparison.

A deterministic regression test is the right follow-up: it needs a seam to put the target into
`_renderedAheadOfTime` at the moment the handler subscribes, then assert that `FrameRendered` is still
raised. Happy to add that here or separately.

## PR Checklist ✅

  - No automated test: the failure is a startup timing race in the WASM host, reachable only from a real browser boot. Validated with the device harness described above. A deterministic test would need a seam to force the `_renderedAheadOfTime` state at the moment `ShowCore` subscribes; happy to add that as a follow-up if wanted.


morning4coffe-dev (commented):
Source review: RequestNewFrame is the appropriate dirty-record request for a completion path that waits on FrameRendered; a present-only host invalidation can leave that event unfired. The existing unsubscribe/removal callback remains in place. This browser splash lifecycle has no direct native WinUI source analogue, so I am not claiming a literal WinUI port. No blocking issue found in the changed call, but the current master conflict and failed review check must be handled before merge readiness.

morning4coffe-dev (approved):
Re-reviewed exact head `9656bf89223177e3fdc0a8c08b70f79d56102e0e`. The force-push was rebase-only and the patch remains correct: subscription precedes `RequestNewFrame`, scheduler coalescing still produces the required record, and the callback unsubscribes before removing the splash. The previous conflict and review-check blockers are resolved; no blocking code issue found.

PR #24267, opened 2026-08-28
docs: clarify x:Bind literal argument scope

### LINKED ISSUE #20928: [x:True x:False] Clarify where the x extensions for true or false are available
### What would you like clarification on? 🤔

- in which properties or elements are they usable?

https://github.com/user-attachments/assets/a729d6c9-4bf9-4aa7-8ebe-f01f24bec5ab

Image

It seems like they are not available in TargetNullValue or FallbackValue of xaml, where they migh make sense like in the example. lets say the button should only be enabled if the textbox is not empty, but for such "easy" Task you might rather not want to use just for this codebehind. that would be in my opinion a good usecase for it. but the compiler is other opinion...

### Concern 📚

Developing with Uno 🛠️

### Renderer 🎨

- [ ] Skia
- [ ] Native

### Affected platforms 📱💻🖥️

_No response_

### Any feedback? 💬

Repro so you can check out that this is not working yourself
[UnoApp-xTrue-xFalse Smallrepro.zip](https://github.com/user-attachments/files/21026228/UnoApp-xTrue-xFalse.Smallrepro.zip)

#19155 @jeromelaban @MartinZikmund 
sorry for taking so long, I just forgot about it and stamped it for myself as not usable.
Seems like its only available in the direct Bind function but not for those Values to be set. I was thinking about this as workaround for the Navigator providing the DataContext not at the point where it is bound and the SourceGen Roslyn issue. I can not change those, but I were hoping that would be usable without codebehind...

SergioAlmeida29:
**GitHub Issue:** closes #20928

## PR Type:

## What changed? 🚀

Clarifies that `x:True`, `x:False`, and `x:Null` are literal arguments supported in `{x:Bind ...}` function calls. They are not general-purpose values for `FallbackValue`, `TargetNullValue`, or arbitrary XAML properties.

## PR Checklist ✅

- [ ] 🧪 Runtime tests or a manual sample are not applicable to this documentation-only change.
- [x] 📚 Docs have been updated following the documentation template where applicable.
- [ ] 🖼️ Screenshots Compare Test Run is not applicable to documentation-only changes.

## Validation

- `git diff --check`: passed.
- Markdownlint and cSpell: passed.
- DocFX: unavailable in the local environment.

### DIFF
--- a/doc/articles/features/windows-ui-xaml-xbind.md
+++ b/doc/articles/features/windows-ui-xaml-xbind.md
@@ -57,6 +57,8 @@ Uno supports the [`x:Bind`](https://learn.microsoft.com/windows/uwp/xaml-platfor
     <TextBlock Text="{x:Bind TestString(x:Null)}" />
     ```
 
+    In this context, `x:True`, `x:False`, and `x:Null` are literal arguments scoped to `{x:Bind ...}` function calls; they are not general-purpose values that can be assigned to `FallbackValue`, `TargetNullValue`, or arbitrary XAML properties.
+
   - Quote escaping
 
     ```xml


morning4coffe-dev (changes requested):
[P2] Please distinguish x:Bind argument tokens from the general {x:Null} markup extension. The new sentence currently groups x:Null with x:True/x:False and says it cannot be assigned to arbitrary XAML properties. WinUI explicitly supports `<object property="{x:Null}" />`: https://learn.microsoft.com/en-us/windows/apps/develop/platform/xaml/x-null-markup-extension . Uno's own ContentPresenter_Content_DataContext.xaml and ContentPresenter_Template.xaml also use Content="{x:Null}". It is fine to explain that a bare function-argument token is not an attribute-value replacement, but the documentation must not imply that the markup extension itself is unavailable.

PR #24269, opened 2026-08-28
docs: clarify Uno 6 platform and renderer names

### LINKED ISSUE #21735: [Docs][SupportedFeatures-Tables][Enhancement-Request] Please update the Platform specific feature/api support tables with v6.0 target and renderer naming conventions
### What would you like clarification on? 🤔

I was reading the Image Control Page, that links to the [`StorageFile.ReadStorageFileFromApplicationUrlAsync` Docs](doc/articles/features/file-management.md) and with the not longer Multi Head Uno App Solution/Project Strucutre, it's difficult for me to really understand [this table on the top region of this page](https://platform.uno/docs/articles/features/file-management.html#supported-features) that should tell me as reader about the supported state for our target platforms in Uno.

Image
*(Reference Screenshot of this table)*

The Difficultness in this, to help you understand my point, is the naming of the column headers that should tell me the platform.
Since you added, up from the uno.sdk v6.0, the Differencing between the Renderers `SkiaRenderer` and `Native`, the naming convention of `Skia` as equivalent for `Desktop` is in my opinion (not longer) a real equivalent we should automatic imply.

I for example do target `net9.0-desktop` and develop on Windows 11 pro which would then be `Desktop (Windows)` in my understanding. Assuming you would mean in this table for example `Skia` == `Desktop` and the user does maybe make use of `SkiaRenderer` while targeting Android, which we now can, thanks to Uno's 6.0 release, we have a  duplication of targets. And for me searching for the answer for **Is this feature supported on Desktop targeting to run on Windows?** well, I still don't know the answer!

**Maybe you mean:**

`Skia (Wpf)` == `Desktop (Windows)` -> *Yes, we have support for this on your target!*

**But could also mean:**

`Skia (Wpf)` == `Uno Islands embedded in Wpf app`, which I tryed and hopelessly failed (following the Uno Islands docs tutorial and sample app) with the error of the Wpf Designer & Compiler in the end was telling me, that WinUI controls are not supported to be embeeded in Wpf apps.

I have at the moment no full list of the Desktop target supported full list, but `Skia (Wpf)` is defintly nothing that would make me imply Windows, so would be really nice if you could consider to update those in the future!

### Concern 📚

Clarification of capabilities 💡

### Renderer 🎨

- [x] Skia
- [x] Native

### Affected platforms 📱💻🖥️

Desktop (Windows), Desktop (macOS), Desktop (X11), Desktop (Linux Framebuffer), Build tasks

### Any feedback? 💬

> [!TIP]
> Maybe you could change the column header namings to the ones we have here for the `docs/Enhancement-Request` issue GitHub template, that would fit this need much better 👍 

> [!NOTE]
> While the GitHub unoplatform repo(s) issue template I am currently using is showing this maybe to be targeted Naming, this does **not apply**, until now, to all other active Uno Repository's

This caused for me already multiple times the same problem and difficultness as mentioned for the tables in the docs  when opening a Bug Report or other Report with those in the unoplatform repositories.
So, if you would consider to update those too, it would help me a lot, to provide you better and more useable Feedback/Reports in the future 👍 💯 
Requested the issue template part here (maybe wording improvable/could be more clear there, but same goal) in:
- https://github.com/unoplatform/uno/issues/21017

SergioAlmeida29:
**GitHub Issue:** closes #21735

## PR Type:

## What changed? 🚀

Updates 24 supported-feature tables and templates to use Uno 6 target and renderer names: Windows App SDK, Native/Skia mobile and WebAssembly, and the four Skia Desktop hosts (Windows, macOS, X11, and Linux Framebuffer). Existing support marks are preserved.

## PR Checklist ✅

- [ ] 🧪 Runtime tests or a manual sample are not applicable to this documentation-only change.
- [x] 📚 Docs have been updated following the documentation template where applicable.
- [ ] 🖼️ Screenshots Compare Test Run is not applicable to documentation-only changes.

## Validation

- `git diff --check`: passed.
- Markdownlint: passed for all 24 changed documentation files.
- `npm run strict`: reached DocFX, which is unavailable in the local environment.

The specialized renderer/implementation tables were intentionally left unchanged because their columns describe different dimensions and need independent support verification.

### DIFF
--- a/doc/articles/features/windows-storage-pickers.md
+++ b/doc/articles/features/windows-storage-pickers.md
@@ -19,23 +19,23 @@ Legend
 * 💬 Partially supported (see below for more details)
 * ✖ Not supported
 
-| Picker         | Windows App SDK   | WebAssembly | Android | iOS    | Desktop |
-|----------------|-------|-------------|---------|--------|-------|
-| FileOpenPicker | ✔    | ✔ (1)       | ✔      | ✔      | ✔    |
-| FileSavePicker | ✔    | ✔ (1)       | ✔      | 💬 (2) | ✔    |
-| FolderPicker   | ✔    | ✔           | ✔      | ✔      | ✔    |
+| Picker         | Windows App SDK | Android (Native/Skia) | iOS (Native/Skia) | WebAssembly (Native/Skia) | Desktop (macOS, Skia) | Desktop (X11, Skia) | Desktop (Linux Framebuffer, Skia) | Desktop (Windows, Skia) |
+|----------------|-----------------|-----------------------|-------------------|----------------------------|-----------------------|---------------------|------------------------------------|-------------------------|
+| FileOpenPicker | ✔               | ✔                     | ✔                 | ✔ (1)                      | ✔                     | ✔                   | ✔                                  | ✔                       |
+| FileSavePicker | ✔               | ✔                     | 💬 (2)            | ✔ (1)                      | ✔                     | ✔                   | ✔                                  | ✔                       |
+| FolderPicker   | ✔               | ✔                     | ✔                 | ✔                          | ✔                     | ✔                   | ✔                                  | ✔                       |
 
 *(1) - Multiple implementations supported - see WebAssembly section below*\
 *(2) - See iOS section below*
 
 On some platforms, you can further customize the file-picking experience by utilizing additional properties:
 
-| Feature                 | Windows App SDK  | WebAssembly | Android | iOS   | Desktop |
-|-------------------------|------|-------------|---------|-------|-----|
-| SuggestedFileName       | ✔   | ✔           | ✖      | ✖     | ✔ |
-| SuggestedStartLocation  | ✔   | ✔ (1)       | 💬 (4) | ✔ (3) | ✔ |
-| SettingsIdentifier      | ✔   | ✔ (1)       | ✔      | ✖     | ✔ |
-| SetMultipleFilesLimit   | ✖   | ✖           | ✖      | ✔     | ✖ |
+| Feature                | Windows App SDK | Android (Native/Skia) | iOS (Native/Skia) | WebAssembly (Native/Skia) | Desktop (macOS, Skia) | Desktop (X11, Skia) | Desktop (Linux Framebuffer, Skia) | Desktop (Windows, Skia) |
+|------------------------|-----------------|-----------------------|-------------------|----------------------------|-----------------------|---------------------|------------------------------------|-------------------------|
+| SuggestedFileName      | ✔               | ✖                     | ✖                 | ✔                          | ✔                     | ✔                   | ✔                                  | ✔                       |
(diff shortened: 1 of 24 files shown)


doc/articles/features/windows-storage-pickers.md
```diff
@@ -19,23 +19,23 @@ Legend
...
 * ✖ Not supported
 
-| Picker         | Windows App SDK   | WebAssembly | Android | iOS    | Desktop |
-|----------------|-------|-------------|---------|--------|-------|
-| FileOpenPicker | ✔    | ✔ (1)       | ✔      | ✔      | ✔    |
-| FileSavePicker | ✔    | ✔ (1)       | ✔      | 💬 (2) | ✔    |
-| FolderPicker   | ✔    | ✔           | ✔      | ✔      | ✔    |
+| Picker         | Windows App SDK | Android (Native/Skia) | iOS (Native/Skia) | WebAssembly (Native/Skia) | Desktop (macOS, Skia) | Desktop (X11, Skia) | Desktop (Linux Framebuffer, Skia) | Desktop (Windows, Skia) |
+|----------------|-----------------|-----------------------|-------------------|----------------------------|-----------------------|---------------------|------------------------------------|-------------------------|
+| FileOpenPicker | ✔               | ✔                     | ✔                 | ✔ (1)                      | ✔                     | ✔                   | ✔                                  | ✔                       |
```
morning4coffe-dev, line 24: [P2] Do not advertise an unimplemented framebuffer picker backend

The new `Desktop (Linux Framebuffer, Skia)` column marks all three pickers as supported, but this head's [FrameBufferHost registrations](https://github.com/unoplatform/uno/blob/72ec7e7d2d2776a5991bcde42bc716796548d98b/src/Uno.UI.Runtime.Skia.Linux.FrameBuffer/Hosting/FramebufferHost.cs#L155-L168) contain no `IFileOpenPickerExtension`, `IFileSavePickerExtension`, or `IFolderPickerExtension`. Those are registered by the X11 host, not the framebuffer host. The corresponding Skia `Pick*TaskAsync` methods explicitly throw `NotSupportedException` when the extension is absent.

Please mark framebuffer file/folder picking unsupported, and update its dependent entries in the optional-properties table too, unless a real framebuffer implementation is being provided. Copying the old generic Linux/Desktop support marks into this explicit backend column sends framebuffer users to examples that cannot open a picker.

morning4coffe-dev (changes requested):
Reviewed all 24 documentation changes at `72ec7e7d2d2776a5991bcde42bc716796548d98b`. Separating the desktop backends is useful, but the newly explicit framebuffer picker support does not match that head's implementations. Please correct both picker tables as described inline.

I also checked the framebuffer clipboard path: it does register `LocalClipboardExtension`, so I am not reporting an absent clipboard backend. This is a documentation/source review, not framebuffer hardware execution or a device-validated Native/Skia image-codec support matrix.

PR #24282, opened 2026-08-29
feat: Add opt-in press-to-stop-momentum for ManipulationMode panning

ajpinedam:
**GitHub Issue:** Tracked internally (customer-reported). No public issue.

## PR Type:

✨ Feature

## What changed? 🚀

A customer reported that on a tablet, tapping a CommunityToolkit `DataGrid` while it is still decelerating after a fling both stops the momentum **and** selects the row under the finger, instead of the first tap only halting the scroll.

**This turned out to be intended WinUI behaviour, not a bug.** The `DataGrid` pans its rows through the `ManipulationMode` of its `DataGridRowsPresenter` (runtime-confirmed: `TranslateX, TranslateY, TranslateRailsX, TranslateRailsY, TranslateInertia, System`), and in WinUI gesture recognition for that path is per-element and independent of an ancestor's inertia:

- interaction contexts are keyed per `CUIElement` (`CInteractionManager::m_mapInteractions`), so a descendant's tap recognition knows nothing about the panning ancestor's inertia;
- `CInteractionManager::IsManipulationInertiaProcessing()` has **zero callers**, so nothing gates press dispatch while an inertia is running;
- the tap-suppression guard that does exist — `GestureOutputMapper::OnTapped`/`OnSecondaryTapped`/`OnHolding` dropping the message when `IsInteractionStopping() && IsInteractionStopped()` — is armed only by `ElementGestureTracker::Stop()`, whose callers are DirectManipulation-steals-the-pointer, capture-changed and focus. Not "a press landed during my own inertia".

So WinUI activates the pointed item on that press, and Uno Platform now keeps doing the same by default.

Worth noting the corollary: Uno Platform's own `ScrollViewer` is **not** diverging. Its touch panning goes through the direct-manipulation pipeline, which swallows the press that stops an inertia before it reaches the visual tree — the same thing DirectManipulation does on WinUI. Only the `UIElement.ManipulationMode` path is affected, which is why the two behave differently today.

The gap is nonetheless real on touch-only targets: a browser on a tablet has no mouse, and users expect the OS convention that the first press only stops the momentum. That makes this an **enhancement**, which must not change default behaviour on any target.

So it ships as an opt-in Toolkit extension:

```xml
<Border ManipulationMode="TranslateY,TranslateInertia"
        toolkit:ManipulationExtensions.IsTapToStopInertiaEnabled="True">
```

While enabled, the press that aborts the inertia raises no `Tapped`, `DoubleTapped`, `RightTapped` or `Holding` on the element, its ancestors or its descendants; a tap at rest is unaffected. It is a no-op on Windows.

Applied through an implicit `Style` it also reaches panning elements an app cannot address directly — including `DataGrid`'s `DataGridRowsPresenter`, which is the customer's actual case:

```xml
<Style TargetType="primitives:DataGridRowsPresenter">
    <Setter Property="toolkit:ManipulationExtensions.IsTapToStopInertiaEnabled" Value="True" />
</Style>
```

### This does change core code, and the default is unchanged

The Toolkit cannot observe what it needs on its own, so `Uno.UI` gains four internals. All of them are **purely additive and behaviour-neutral** — nothing in core reads them, so no code path changes unless the opt-in is set. Core reports the fact, the opt-in decides what to do with it:

- `GestureRecognizer.Manipulation.IsCoasting`
- `GestureRecognizer.LastDownStoppedInertia` — informative only; the gestures of that press are still recognized exactly as before
- `UIElement.LastPointerDownStoppedInertia`
- `UIElement.PreventGestures(pointer, gestures)`

`GestureRecognizer.ProcessDownEvent` gains one assignment to compute the flag; every other line is a new member. The unchanged default is pinned by `When_PressStopsInertia_Without_OptIn_Then_TapIsRaised`.

### Platform reach

The opt-in is honoured on the targets which use the managed pointer pipeline — every Skia target (Desktop, and Skia on Android, iOS and WebAssembly) and the WebAssembly backend. It is inert on WinAppSDK, where the callback body is `#if HAS_UNO` so the property just stores its value and cross-platform XAML stays valid. On the native Android and iOS backends pointer events bubble natively, so the order the gestures are muted in has not been validated and those backends are documented as unsupported for this opt-in.

The property is deliberately not gated per target: the default is unchanged unless it is set, and a setter that silently refuses a value is harder to diagnose than one documented as inert.

### Implementation note

The muting needs two passes because of the order in which a press reaches the gesture recognizers: each element updates its own gestures **before** invoking its handlers, ancestors are visited from the closest to the root, and the original source updates its own gestures **last of all**. So the extension mutes the original-source→tracked-element path on the press (retroactively, those gestures already exist) and the remainder — the ancestors and the original source — on the release, which still runs before their gestures are recognized.

### Validation

- **Runtime** — Skia Desktop (`SamplesApp.Skia.Generic`), `Given_ManipulationExtensions`: 3 passed.
  - `When_PressStopsInertia_Without_OptIn_Then_TapIsRaised` pins the **default**, i.e. the WinUI behaviour, so parity cannot regress silently.
  - `When_PressStopsInertia_With_OptIn_Then_NoTapIsRaisedOnThePath` asserts no tap on the original source, an intermediate element, the coasting element and an ancestor — the four positions the two passes have to cover.
  - `When_TapAtRest_With_OptIn_Then_TapIsRaisedOnThePath` asserts the opt-in leaves a normal tap alone.
  - Both halves of the muting are load-bearing: removing the retroactive pass fails on the intermediate element, removing the forward flag fails on the original source.
- **Runtime (regression)** — 92 passed / 13 skipped across `Given_InputManager`, `Given_GestureRecognizer`, `Given_ScrollViewer`, `Given_SwipeControl`, `Given_FlipView`, `Given_RefreshContainer` and `Uno_UI_Toolkit`.
- **Unit** — `Given_GestureRecognizer`: the recognizer reports `LastDownStoppedInertia` for the aborting press **and still raises the tap**, plus a no-false-positive case.
- **Against the customer's repro** — a rebuilt copy of the reported project (real `DataGrid` 7.1.205) on Skia macOS confirmed the implicit `Style` does deliver an attached property to `DataGridRowsPresenter`, and that its `ManipulationMode` is the inertia-enabled one above.
- **Compile** — `Uno.UI` and `Uno.UI.Toolkit` for Skia, Reference, WebAssembly and iOS, and an analyzer-enabled build (no `UnoFastDevBuild`) of `Uno.UI.Toolkit` and `Uno.UI.RuntimeTests`: 0 errors, no new warnings.
- **Device** — confirmed on iPadOS against a LAN-served WebAssembly Skia SamplesApp build of this branch, through the `Manipulation_TapStopsInertia` sample (its in-app toggle flips between the default and the opt-in on the same screen).

## PR Checklist ✅

### DIFF
--- a/src/SamplesApp/SamplesApp.Samples/Windows_UI_Input/GestureRecognizerTests/Manipulation_TapStopsInertia.xaml.cs
+++ b/src/SamplesApp/SamplesApp.Samples/Windows_UI_Input/GestureRecognizerTests/Manipulation_TapStopsInertia.xaml.cs
@@ -1,0 +1,69 @@
+using System;
+using System.Linq;
+using Microsoft.UI.Xaml;
+using Microsoft.UI.Xaml.Controls;
+using Microsoft.UI.Xaml.Input;
+using Uno.UI.Samples.Controls;
+using Uno.UI.Toolkit;
+using Windows.Foundation;
+
+namespace UITests.Windows_UI_Input.GestureRecognizerTests
+{
+	[Sample(
+		"Gesture Recognizer",
+		Description = "ManipulationExtensions.IsTapToStopInertiaEnabled: the tap which stops the momentum of a ManipulationMode element does not select the pointed row.",
+		IsManualTest = true)]
+	public sealed partial class Manipulation_TapStopsInertia : Page
+	{
+		private const int RowCount = 40;
+		private const double RowHeight = 45; // Including the 1px margin between the rows.
+
+		private Border _selectedRow;
+		private int _selectionCount;
+
+		public Manipulation_TapStopsInertia()
+		{
+			this.InitializeComponent();
+
+			Rows.ItemsSource = Enumerable.Range(1, RowCount).Select(i => $"Row {i}").ToArray();
+			Viewport.SizeChanged += (_, e) => ViewportClip.Rect = new Rect(new Point(), e.NewSize);
+
+			UpdateStatus("ready");
+		}
+
+		private void OptIn_Changed(object sender, RoutedEventArgs e)
+			=> Viewport.SetIsTapToStopInertiaEnabled(OptIn.IsChecked is true);
+
+		private void Row_Tapped(object sender, TappedRoutedEventArgs e)
+		{
+			if (_selectedRow is { } previous)
+			{
+				previous.BorderThickness = default;
+			}
+
+			var row = (Border)sender;
+			row.BorderThickness = new Thickness(3);
+			_selectedRow = row;
+			_selectionCount++;
+
+			UpdateStatus($"selected {((TextBlock)row.Child).Text}");
+		}
+
+		private void Viewport_ManipulationInertiaStarting(object sender, ManipulationInertiaStartingRoutedEventArgs e)
+			=> UpdateStatus("coasting");
+
+		private void Viewport_ManipulationDelta(object sender, ManipulationDeltaRoutedEventArgs e)
+		{
+			var minOffset = Math.Min(0, Viewport.ActualHeight - RowCount * RowHeight);
+			RowsTransform.Y = Math.Clamp(RowsTransform.Y + e.Delta.Translation.Y, minOffset, 0);
+
+			if (!e.IsInertial)
+			{
+				UpdateStatus("dragging");
+			}
+		}
+
+		private void UpdateStatus(string state)
+			=> Status.Text = $"{state} — selections: {_selectionCount}";
+	}
+}
--- a/src/SamplesApp/SamplesApp.Samples/Windows_UI_Input/GestureRecognizerTests/Manipulation_TapStopsInertia.xaml.cs
+++ b/src/SamplesApp/SamplesApp.Samples/Windows_UI_Input/GestureRecognizerTests/Manipulation_TapStopsInertia.xaml.cs
@@ -52,6 +52,9 @@ namespace UITests.Windows_UI_Input.GestureRecognizerTests
 		private void Viewport_ManipulationInertiaStarting(object sender, ManipulationInertiaStartingRoutedEventArgs e)
 			=> UpdateStatus("coasting");
 
+		private void Viewport_ManipulationCompleted(object sender, ManipulationCompletedRoutedEventArgs e)
+			=> UpdateStatus("completed");
+
 		private void Viewport_ManipulationDelta(object sender, ManipulationDeltaRoutedEventArgs e)
 		{
 			var minOffset = Math.Min(0, Viewport.ActualHeight - RowCount * RowHeight);
--- a/src/Uno.UI.Toolkit/ManipulationExtensions.cs
+++ b/src/Uno.UI.Toolkit/ManipulationExtensions.cs
@@ -1,3 +1,5 @@
+#nullable enable
+
 using Microsoft.UI.Xaml;
 
 #if HAS_UNO
@@ -139,7 +141,7 @@ namespace Uno.UI.Toolkit
 				}
 			}
 
-			private static void MutePath(PointerRoutedEventArgs args, PointerIdentifier pointer, UIElement upTo)
+			private static void MutePath(PointerRoutedEventArgs args, PointerIdentifier pointer, UIElement? upTo)
 			{
 				if (args.OriginalSource is not UIElement originalSource || Mute(originalSource))
 				{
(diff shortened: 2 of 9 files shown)


src/SamplesApp/SamplesApp.Samples/Windows_UI_Input/GestureRecognizerTests/Manipulation_TapStopsInertia.xaml.cs
```diff
@@ -0,0 +1,69 @@
...
+
+			var row = (Border)sender;
+			row.BorderThickness = new Thickness(3);
+			_selectedRow = row;
+			_selectionCount++;
+
+			UpdateStatus($"selected {((TextBlock)row.Child).Text}");
+		}
+
+		private void Viewport_ManipulationInertiaStarting(object sender, ManipulationInertiaStartingRoutedEventArgs e)
```
Xiaoy312, line 52: log one for ManipulationCompleted too

src/Uno.UI.UnitTests/Windows_UI_Input/Given_GestureRecognizer.cs
```diff
@@ -1163,6 +1163,53 @@ public void Manipulation_Inertia_TranslateYOnly()
 			);
 		}
 
+		[TestMethod]
+#if RUNTIME_NATIVE_AOT
+		[Ignore(".BeEquivalentTo() unsupported under NativeAOT; see: https://github.com/AwesomeAssertions/AwesomeAssertions/issues/290")]
```
Xiaoy312: why are we relying on this? i can understand for legacy code, but newly added code just work around it
(dont change the rest yet, but start here to experiment)

Xiaoy312 (approved):

src/SamplesApp/SamplesApp.Samples/Windows_UI_Input/GestureRecognizerTests/Manipulation_TapStopsInertia.xaml.cs
ajpinedam: Added in 131769cf31e55b52ce19b5a37d2d26a7dc83bba9 — `ManipulationCompleted` now logs a `completed` status. This also makes the opt-in success case observable: the status used to stay frozen on `coasting` after the (muted) tap stopped the momentum. Note that a plain tap on a row doesn't raise `ManipulationCompleted` (the manipulation never leaves `Starting`), so the `selected Row N` status isn't overwritten.

src/Uno.UI.UnitTests/Windows_UI_Input/Given_GestureRecognizer.cs
ajpinedam: Reworked in b7c2ecdbc6f835ee5d9d43064e289aaaecfb567f: dropped the ignore and replaced `BeEquivalentTo` with direct asserts on `Position` / `TapCount` / `PointerDeviceType` — all `Equals`-based, so they run fine under NativeAOT. Left the legacy ignores untouched as suggested.

Side note from digging into this: `RUNTIME_NATIVE_AOT` is only defined when `SkiaPublishAot=true` (src/Directory.Build.props), i.e. the Android NativeAOT SamplesApp publish — and that build only compiles the unit-test files linked via `UnitTestsImport.props`, which doesn't include this file. So the ignore here was actually unreachable; I had copied it from the neighboring tests, where it's currently inert for the same reason.

src/Uno.UI.Toolkit/ManipulationExtensions.cs
```diff
@@ -0,0 +1,169 @@
...
+			private void OnPointerReleased(object sender, PointerRoutedEventArgs args)
+			{
+				var pointer = args.Pointer.UniqueId;
+				if (_pendingPointers.Remove(pointer))
+				{
+					MutePath(args, pointer, upTo: null);
+				}
+			}
+
+			private static void MutePath(PointerRoutedEventArgs args, PointerIdentifier pointer, UIElement upTo)
```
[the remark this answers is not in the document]
ajpinedam: Addressed in b844b06f608852baf73b71fe7fac770f938984bf, with one correction to the rationale: this file had no `#nullable enable` (the repo opts in per-file), so there were no nullability warnings to avoid — and annotating just the parameter would have introduced CS8632 instead. Since the file is new in this PR, the right fix was enabling nullable on it: `#nullable enable` plus `UIElement? upTo`, which does make the intended contract explicit (`null` = mute the whole path to the root). Verified the Skia, Wasm and Reference flavors compile clean with no further annotations needed.

agneszitte:

PR #24321, opened 2026-08-31
fix(vs): Marshal dispose to the UI thread

### LINKED ISSUE #24149: [VS] `InfoBarFactory.Dispose` throws `COMException`: Dispose must be called on the UI thread
## Current behavior 🐛

Closing a solution in Visual Studio (or otherwise disposing the Remote Control VS entry point off the UI thread) throws:

```text
System.Runtime.InteropServices.COMException: Dispose must be called on the UI thread.
```

The exception is thrown from `ThreadHelper.ThrowIfNotOnUIThread()` inside `InfoBarFactory.Dispose()`.

`IVsInfoBar*` COM objects are UI-thread-affine. `InfoBarFactory.CreateAsync` already marshals with `SwitchToMainThreadAsync()`, but `Dispose()` / `RemoveAllInfoBars()` / `InfoBar.Close()` only *assert* the UI thread and do not marshal to it.

`EntryPoint.Dispose()` calls `_infoBarFactory?.Dispose()` (and DTE event unsubscription) with no main-thread switch. It is a synchronous `IDisposable.Dispose()` loaded by the Uno Platform Studio VSIX via reflection (`DevServerLauncher`), so callers can (and do) invoke it from a thread-pool task.

Observed stack (VS 2026 Insiders, solution closing):

```text
Microsoft.VisualStudio.Shell.ThreadHelper.ThrowIfNotOnUIThread()
Uno.UI.RemoteControl.VS.Notifications.InfoBarFactory.Dispose()
Uno.UI.RemoteControl.VS.EntryPoint.Dispose()
Studio.Extensions.Tools.DevServerLauncher.DisposeEntryPointAsync()
Studio.Extensions.Tools.DevServerLauncher.DisposeAsync()
UnoSolutionTemplate.UnoServiceProvider.DisposeAsync()
UnoSolutionTemplate.UnoPlatformPackage.OnSolutionClosing.<AnonymousMethod>__0()
System.Threading.Tasks.Task.Execute()   // thread-pool, not UI thread
```

Throwing site:

https://github.com/unoplatform/uno/blob/master/src/Uno.UI.RemoteControl.VS/Notifications/InfoBar.cs

```csharp
internal void Dispose()
{
    ThreadHelper.ThrowIfNotOnUIThread();
    RemoveAllInfoBars();
}
```

Caller:

https://github.com/unoplatform/uno/blob/master/src/Uno.UI.RemoteControl.VS/EntryPoint.cs

```csharp
public void Dispose()
{
    // ...
    _dte.Events.BuildEvents.OnBuildBegin -= _onBuildBeginHandler;
    // ...
    _infoBarFactory?.Dispose();
}
```

`EntryPoint.cs` also has `#pragma warning disable VSTHRD010` / `VSTHRD109`, which hides the analyzer that would have flagged this.

## Expected behavior 🎯

Disposing `EntryPoint` / `InfoBarFactory` from any thread (including `OnSolutionClosing` on a thread-pool task) should marshal InfoBar / DTE teardown onto the UI thread and complete without throwing. Info bars should be closed; COM objects should not be touched off the main thread.

## How to reproduce it (as minimally and precisely as possible) 🔬

1. Install the Uno Platform Visual Studio extension (Studio VSIX) and open an Uno solution so Remote Control / DevServer is loaded (`Uno.UI.RemoteControl.VS.EntryPoint`).
2. Have the debugger attached to `devenv` (or wait for the unhandled exception).
3. Close the solution (or close VS).
4. Observe `COMException` from `InfoBarFactory.Dispose` on a worker thread.

No Uno app-head change is required; this is IDE-extension teardown.

## Workaround 🛠️

None for users. Avoid leaving InfoBars open does not help: `ThrowIfNotOnUIThread()` runs *before* `RemoveAllInfoBars()`.

## Renderer 🎨

N/A (Visual Studio extension, not an Uno app renderer).

## Affected platforms 📱💻🖥️

- Visual Studio (Windows) — Uno.UI.RemoteControl.VS / Uno Platform Studio VSIX
- Not an application runtime issue

## Uno.Sdk version (and other relevant versions) 📦

- Code: `unoplatform/uno` `src/Uno.UI.RemoteControl.VS`
- Loaded by Uno Platform Studio (`Studio.Extensions.dll` → `DevServerLauncher`)
- Reproduced while debugging `devenv.exe`

## IDE version 🧑‍💻

Visual Studio Enterprise 2026 18.10.0-insiders

## Proposed solution

The VS add-in (`EntryPoint`) is documented as being constructed/disposed by `DevServerLauncher` via `Assembly.LoadFrom` + `Activator.CreateInstance` (see `specs/001-fast-devserver-startup`). That path does **not** guarantee the UI thread, so **`Dispose` must marshal itself**.

### 1. Marshal in `InfoBarFactory.Dispose` (required)

`CreateAsync` already uses `SwitchToMainThreadAsync`. `Dispose` should do the same instead of throwing:

```csharp
internal void Dispose()
{
    if (ThreadHelper.CheckAccess())
    {
        RemoveAllInfoBars();
        return;
    }

    ThreadHelper.JoinableTaskFactory.Run(async () =>
    {
        await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
        RemoveAllInfoBars();
    });
}
```

Keep `ThrowIfNotOnUIThread()` on `RemoveAllInfoBars()` / `InfoBar.Close()` — those stay UI-only; only the public dispose entry point marshals.

### 2. Marshal in `EntryPoint.Dispose` (strongly recommended)

The same method also unhooks `_dte.Events.BuildEvents.*`, which is COM and UI-thread-affine. Even after (1), DTE unsubscription can still throw off the UI thread.

```csharp
public void Dispose()
{
    if (_isDisposed)
    {
        return;
    }
    _isDisposed = true;

    void DisposeCore()
    {
        ThreadHelper.ThrowIfNotOnUIThread();
        try
        {
            _ct.Cancel(false);
            _dte.Events.BuildEvents.OnBuildBegin -= _onBuildBeginHandler;
            _dte.Events.BuildEvents.OnBuildDone -= _onBuildDoneHandler;
            _dte.Events.BuildEvents.OnBuildProjConfigBegin -= _onBuildProjConfigBeginHandler;
            _globalJsonObserver?.Dispose();
            _debuggerObserver?.Dispose();
            _infoBarFactory?.Dispose();
            _unoMenuCommand?.Dispose();
            _appLaunchIdeBridge?.Dispose();
            _appLaunchStateConsumer?.Dispose();
        }
        catch (Exception e)
        {
            _debugAction?.Invoke($"Failed to dispose Remote Control server: {e}");
        }
    }

    if (ThreadHelper.CheckAccess())
    {
        DisposeCore();
        return;
    }

    ThreadHelper.JoinableTaskFactory.Run(async () =>
    {
        await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
        DisposeCore();
    });
}
```

Use `JoinableTaskFactory` (not `Dispatcher.Invoke` / `ThreadHelper.Generic.Invoke`) to avoid VS deadlocks.

### 3. Optional follow-up in uno.studio (caller)

`UnoPlatformPackage.OnSolutionClosing` currently fires `DisposeEntryPointAsync` as a bare `Task` (`Task.Execute` on the stack). That VSIX should also `await JoinableTaskFactory.SwitchToMainThreadAsync()` before disposing `EntryPoint`. That repo is not this one; **uno still needs (1)+(2)** because `IDisposable.Dispose` cannot assume the caller is on the UI thread.

### 4. Analyzer

Re-enable `VSTHRD010` around the dispose/DTE paths (or suppress only the true false positives) so this class of bug is caught again.

## Anything else we need to know? 💬

Related files:

- https://github.com/unoplatform/uno/blob/master/src/Uno.UI.RemoteControl.VS/Notifications/InfoBar.cs
- https://github.com/unoplatform/uno/blob/master/src/Uno.UI.RemoteControl.VS/EntryPoint.cs
- VS add-in load contract: `specs/001-fast-devserver-startup/spec-appendix-c-testing.md` (item 32, `DevServerLauncher` reflection)

`unoplatform/studio` (public) is plugins/docs only. The VSIX caller lives in the private `uno.studio` repo; the throwing code and the fix belong in **this** repository.

ajpinedam:
**GitHub Issue:** closes #24149

## PR Type:

🐞 Bugfix

## What changed? 🚀

`Uno.UI.RemoteControl.VS.EntryPoint` can be disposed from any thread — for example, the Studio VSIX (`DevServerLauncher`) disposes the reflection-loaded entry point from a thread-pool task while the solution is closing. The teardown, however, touches UI-thread-affine COM objects (DTE `BuildEvents` unsubscription, `IVsInfoBar*` cleanup, menu command disposal), which throws `COMException` when invoked off the UI thread.

Because `Dispose` wraps the teardown in a `try/catch` and `_isDisposed` is set up-front, the exception was swallowed and the teardown was silently aborted partway through: `_unoMenuCommand`, `_appLaunchIdeBridge`, and `_appLaunchStateConsumer` leaked, and open InfoBars were never removed.

This PR marshals the teardown to the UI thread:

- `EntryPoint.Dispose` now takes a `ThreadHelper.CheckAccess()` fast path when already on the UI thread, and otherwise runs the teardown via `ThreadHelper.JoinableTaskFactory.Run` + `SwitchToMainThreadAsync`. The teardown body is extracted into `DisposeCore`, which keeps `ThrowIfNotOnUIThread` as a guard. Marshaling once at the `EntryPoint` level covers all child disposables.
- `InfoBarFactory.Dispose` gets the same treatment (it is also reachable independently), replacing its former `ThrowIfNotOnUIThread` hard requirement.

## PR Checklist ✅

- [ ] 🧪 Added [Runtime tests, UI tests, or a manual test sample](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) (for bug fixes / features, if applicable) — not applicable: VS extension teardown path; validation requires a devenv + VSIX debugging session (manual verification with the reporter's scenario).
- [ ] 📚 Docs have been added/updated following the [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features) — not applicable, no user-facing behavior change.
- [ ] 🖼️ Validated PR `Screenshots Compare Test Run` results — not applicable, VS tooling only.

### DIFF
--- a/src/Uno.UI.RemoteControl.VS/EntryPoint.cs
+++ b/src/Uno.UI.RemoteControl.VS/EntryPoint.cs
@@ -890,21 +890,43 @@ public partial class EntryPoint : IDisposable
 
 		try
 		{
-			_ct.Cancel(false);
-			_dte.Events.BuildEvents.OnBuildBegin -= _onBuildBeginHandler;
-			_dte.Events.BuildEvents.OnBuildDone -= _onBuildDoneHandler;
-			_dte.Events.BuildEvents.OnBuildProjConfigBegin -= _onBuildProjConfigBeginHandler;
-			_globalJsonObserver?.Dispose();
-			_debuggerObserver?.Dispose();
-			_infoBarFactory?.Dispose();
-			_unoMenuCommand?.Dispose();
-			_appLaunchIdeBridge?.Dispose();
-			_appLaunchStateConsumer?.Dispose();
+			// DTE events, InfoBars and menu commands are UI-thread-affine COM objects, but Dispose
+			// can be invoked from any thread (e.g. the Studio VSIX disposes the entry point from a
+			// thread-pool task while the solution is closing).
+			if (ThreadHelper.CheckAccess())
+			{
+				DisposeCore();
+			}
+			else
+			{
+				ThreadHelper.JoinableTaskFactory.Run(async () =>
+				{
+					await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+
+					DisposeCore();
+				});
+			}
 		}
 		catch (Exception e)
 		{
 			_debugAction?.Invoke($"Failed to dispose Remote Control server: {e}");
 		}
+	}
+
+	private void DisposeCore()
+	{
+		ThreadHelper.ThrowIfNotOnUIThread();
+
+		_ct.Cancel(false);
+		_dte.Events.BuildEvents.OnBuildBegin -= _onBuildBeginHandler;
+		_dte.Events.BuildEvents.OnBuildDone -= _onBuildDoneHandler;
+		_dte.Events.BuildEvents.OnBuildProjConfigBegin -= _onBuildProjConfigBeginHandler;
+		_globalJsonObserver?.Dispose();
+		_debuggerObserver?.Dispose();
+		_infoBarFactory?.Dispose();
+		_unoMenuCommand?.Dispose();
+		_appLaunchIdeBridge?.Dispose();
+		_appLaunchStateConsumer?.Dispose();
 	}
 
 	protected IServiceProvider VisualStudioServiceProvider
--- a/src/Uno.UI.RemoteControl.VS/Notifications/InfoBar.cs
+++ b/src/Uno.UI.RemoteControl.VS/Notifications/InfoBar.cs
@@ -59,9 +59,19 @@ public class InfoBarFactory
 
 	internal void Dispose()
 	{
-		ThreadHelper.ThrowIfNotOnUIThread();
+		// IVsInfoBar* objects are UI-thread-affine, but Dispose can be invoked from any thread.
+		if (ThreadHelper.CheckAccess())
+		{
+			RemoveAllInfoBars();
+			return;
+		}
 
-		RemoveAllInfoBars();
+		ThreadHelper.JoinableTaskFactory.Run(async () =>
+		{
+			await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+
+			RemoveAllInfoBars();
+		});
 	}
 }
 


morning4coffe-dev (approved):
Marshaling the entire COM-affine teardown with JoinableTaskFactory, with a UI-thread fast path, addresses the actual failure rather than merely removing ThrowIfNotOnUIThread. DisposeCore retains its thread guard, and InfoBarFactory's independently reachable disposal receives the same protection. The existing non-atomic disposed flag remains a separate concurrent-disposal hardening concern, not proof that this thread-affinity fix is wrong. This is Visual Studio threading, not microsoft-ui-xaml runtime behavior.

src/Uno.UI.RemoteControl.VS/EntryPoint.cs
```diff
@@ -890,23 +890,45 @@ public void Dispose()
...
-			_dte.Events.BuildEvents.OnBuildDone -= _onBuildDoneHandler;
-			_dte.Events.BuildEvents.OnBuildProjConfigBegin -= _onBuildProjConfigBeginHandler;
-			_globalJsonObserver?.Dispose();
-			_debuggerObserver?.Dispose();
-			_infoBarFactory?.Dispose();
-			_unoMenuCommand?.Dispose();
-			_appLaunchIdeBridge?.Dispose();
-			_appLaunchStateConsumer?.Dispose();
+			// DTE events, InfoBars and menu commands are UI-thread-affine COM objects, but Dispose
+			// can be invoked from any thread (e.g. the Studio VSIX disposes the entry point from a
```
dr1rrb, line 894: IMO we should instead make sure that each UI component is wrapped into a service that implement `IAsyncDisposable` and which ensure to go back on dispatcher only if needed for the given service. We already did it for the UDEI and some components.

dr1rrb (approved):
Approving since this will not cause any issue, but not the best fix possible

src/Uno.UI.RemoteControl.VS/EntryPoint.cs
ajpinedam: follow up ticket https://github.com/unoplatform/uno/issues/24503

PR #24325, opened 2026-09-01
fix(ci): Stop iOS test stalls from eating the job

### LINKED ISSUE #24324: iOS Skia runtime test shards stall to the 90 minute job wall with no diagnostics
### Current behavior 🐛

An `iOS Skia Runtime Tests <n>` job periodically stalls and runs to the 90 minute job wall, while its **sibling shards on the same build** finish in 22-34 minutes. It leaves nothing behind to explain why.

Two live examples (both stalled concurrently, on unrelated PRs):

| Build | shard 0 | shard 1 | shard 2 | shard 3 |
|---|---|---|---|---|
| 231326 | 29.5m ✅ | 22.0m ✅ | 34.3m ✅ | **90.8m 🔴 killed** |
| 231328 | 26.2m ✅ | 22.1m ✅ | **~90m 🔴 killed** | 25.1m ✅ |

Same commit, same app bundle, comparable shard sizes — so this is a per-run stall, not a slow suite or a slow PR.

Over the last 60 builds the `Run iOS Simulator Tests` task is p50 23.5m / p90 50.1m / p99 73.8m, max *successful* 80.1m — against a 90 minute budget. There is essentially no headroom above p99.

**Three compounding causes**, all in `build/test-scripts/ios-uitest-run.sh` and `build/ci/tests/.azure-devops-tests-ios-runner.yml`:

1. **A failed simulator boot is explicitly tolerated.** When `xcrun simctl bootstatus` blows its 180s watchdog, `UNOBLD006` warns *"Continuing anyway; the app install below will surface a hard failure if it is genuinely unusable"* — but it doesn't. The install succeeds, the app even reports a PID, yet SpringBoard is down so not one test runs. This is exactly what build 231326 did (boot 3:03, watchdog killed it), then waited 74 minutes for a results file that could never appear.

2. **The script's own timeout is unreachable dead code.** `UITEST_TEST_TIMEOUT` is set to `90m`, the same as `JobTimeoutInMinutes: 90` — but the wait loop only starts 10-15 minutes into the job (artifact download, toolchain, brew/idb, boot, install, transform-tool build). Its deadline therefore lands *after* the agent has already killed the job. In build 231028 the job started 17:55:38 and the loop at 18:05:46, so its deadline was 19:35 while ADO killed the job at 19:26. The timeout branch has never run on CI — and with it, neither has the teardown that captures the device log, the screenshot and the crash reports.

3. **The wait prints nothing.** The heartbeat `echo` inside the loop is commented out, so the log goes silent for the entire wait. Build 231028 shows an **80 minute hole**:
   ```
   18:05:46 Waiting for /tmp/TestResult-20260829180530.xml to be available...
   19:25:39 ##[error]The Operation will be canceled.
   ```

Related but distinct: #24160 (iOS Skia CI app size / build timeouts) and #24032 (macOS Desktop `RunIdleAsync` hang).

### Expected behavior 🎯

A stalled shard should fail on the harness's own terms, well before the job wall, with the device log, app log, screenshot and crash reports published — and a simulator that never boots should be recycled or fail fast rather than being launched into.

Not deterministic — roughly 1-2% of iOS Skia runtime-test shards. Observed on Azure Pipelines definition 5 (`Uno.UI - CI`), builds 231028, 231326, 231328.

### Workaround 🛠️

Retry the `runtime_tests_skia_ios` stage. Note the fix for cause 1 can reuse the existing `UNO_IOS_HARNESS_CRASHED` sentinel, which already routes a pre-test harness death to the re-run step.

### Affected platforms 📱💻🖥️

iOS (Skia) — CI only.

The duration variance is **macOS-agent-shaped, not iOS-shaped**. p99/p50 across the last 60 builds:

| Stage | p50 | p99 | ratio |
|---|---|---|---|
| Tests - iOS Skia | 25.8m | 77.2m | **3.00x** |
| Tests - Desktop Skia macOS | 20.3m | 61.1m | **3.02x** |
| Tests - Desktop Skia Windows | 18.7m | 43.5m | 2.33x |
| Tests - Desktop Skia Linux | 15.5m | 23.0m | 1.48x |
| Tests - Android Skia | 10.5m | 17.5m | 1.66x |

Both macOS-hosted legs share a 3x tail; Linux does not. iOS just dies more often because its p99 sits closest to its budget.

Note also that build 231028 booted cleanly (2:12, no `UNOBLD006`) and **still** stalled for 80 minutes, so there is a second stall mode that cause 1 does not explain. Diagnosing it needs the instrumentation above.

Finally, the app already writes a `<results>.canary` file before its first test (`App.Tests.cs`), and `wasm-run-skia-runtime-tests.sh` relies on it — but the iOS runner never checks it. That is a ready-made "did the app actually start?" signal.

Only 2 of 4768 `[TestMethod]`s carry a `[Timeout]`, so `GetTestTimeout` returns null and the harness awaits each body unbounded; a single hung test loses the whole shard's ~2043 results, since the XML is written only at the end.

MartinZikmund:
**GitHub Issue:** closes #24324

## PR Type:

## What changed? 🚀

An `iOS Skia Runtime Tests <n>` shard periodically stalled and ran to the 90 minute job wall while its sibling shards on the *same* build finished in 22-34 minutes — and left nothing behind to explain why. #24324 has the full analysis; this PR addresses the three compounding causes.

**1. A failed simulator boot is no longer tolerated.** `UNOBLD006` used to warn "continuing anyway" and press on. The install then succeeds and the app even reports a PID, but SpringBoard is down, so not one test runs. The device is now shut down, erased and re-booted once (that clears a boot wedged on Data Migration / System App); a second failure fails the step with `UNOBLD009` while the existing `UNO_IOS_HARNESS_CRASHED` sentinel can still route it to the re-run step that was already built for exactly this case.

**2. The results wait is bounded by the real job budget.** `UITEST_TEST_TIMEOUT` was set to `90m`, identical to `JobTimeoutInMinutes: 90` — but the wait only starts 10-15 minutes into the job, so its deadline landed *after* the agent had already killed the job. The timeout branch was unreachable dead code, and with it the teardown that captures the device log, screenshot and crash reports. A new first step records `UNO_IOS_JOB_START_EPOCH`, and the wait is clamped to the remaining budget less a 12 minute teardown reserve. On the observed stuck-job shape that is a **68m wait instead of 90m**, ending at minute 78 of 90. It also follows automatically if `JobTimeoutInMinutes` is ever changed, so the two timeouts cannot silently drift apart again.

**3. The wait is no longer silent.** The heartbeat `echo` had been commented out, so a stall was an hour-long hole in the log (build 231028 went 80 minutes without printing a line). The app's console is now streamed to the logs artifact, and the loop beats once a minute with elapsed / remaining and what the app logged since the last beat — so a stall names the test it stalled on.

**Plus a canary gate.** The app already writes `<results>.canary` before its first test and `wasm-run-skia-runtime-tests.sh` already relies on it; the iOS runner never looked. It now waits up to 5 minutes for it and fails fast with `UNOBLD008` (dumping the app log) when the app launches but never reaches a test — the shape these stalls took.

Also `cancelTimeoutInMinutes: 0 → 10`, so if the wall is ever reached anyway the `always()` publish steps still have room to run.

### Validation

CI shell/YAML, so this was validated by extracting the **shipped** blocks and driving them locally:

- **Budget clamp**, 5 scenarios: the real stuck-job shape (90m job, 10m elapsed → clamps 90m to 68m); a short suite (no clamp); budget exhausted (→ `UNOBLD010`, exit 1); job timeout raised to 120m (follows automatically); env unset for local dev (no clamp).
- **Canary gate**, both paths: canary present → proceeds and marks tests started; canary absent → `UNOBLD008`, dumps the app log, exits 1, **and sets `UNO_IOS_HARNESS_CRASHED`** so the re-run fires.
- **Heartbeat**: run against a simulated app writing log lines, confirming it reports the elapsed/remaining figures, the per-beat line delta and the app's recent output, and exits promptly when results appear.
- `bash -n` and a YAML parse on both files; confirmed the staged blob is LF (`*.sh text eol=lf`) so it stays valid on the macOS agents.

Not validated locally: anything needing a real Mac agent, specifically the `xcrun simctl spawn … log stream` invocation. If that turns out to be wrong it degrades gracefully — stderr lands in the same file and the heartbeat shows `+0 lines` — it cannot fail a run.

### What this does *not* fix

Build 231028 booted cleanly (2:12, no `UNOBLD006`) and still stalled 80 minutes, so a second stall mode remains unidentified. The canary gate won't catch it; the heartbeat and the now-reachable teardown are what will finally name it. The underlying macOS-agent variance (p99/p50 of 3.00x on iOS Skia and 3.02x on Desktop Skia macOS, vs 1.48x on Linux) is also untouched.

## PR Checklist ✅

- [ ] 🧪 Added [Runtime tests, UI tests, or a manual test sample](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) (for bug fixes / features, if applicable) — n/a, this is CI shell/YAML with no product code; see the Validation section for how each block was exercised
- [ ] 📚 Docs have been added/updated following the [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features) — n/a, no user-facing behaviour change
- [ ] 🖼️ Validated PR `Screenshots Compare Test Run` results — to be checked once this PR's own run completes

### DIFF
--- a/build/ci/tests/.azure-devops-tests-ios-runner.yml
+++ b/build/ci/tests/.azure-devops-tests-ios-runner.yml
@@ -19,7 +19,10 @@ parameters:
 jobs:
 - job: ${{ parameters.JobName }}
   timeoutInMinutes: ${{ parameters.JobTimeoutInMinutes }}
-  cancelTimeoutInMinutes: 0
+  # The test script now ends the run before the job wall and does its own teardown, but if the wall
+  # is ever reached anyway, this is the window in which the always() publish steps below still run.
+  # At 0 a timed-out job published nothing at all.
+  cancelTimeoutInMinutes: 10
   dependsOn: ${{ parameters.dependsOn }}
   displayName: ${{ parameters.JobDisplayName }}
   pool:
@@ -43,6 +46,12 @@ jobs:
       value: $[ dependencies.${{ parameters.BuildJobName }}.outputs['set_ios_artifact.SamplesAppArtifactName'] ]
 
   steps:
+  # Recorded first so the test script can tell how much of the job budget the setup steps have
+  # already spent, and stop waiting for results while there is still room to run its teardown.
+  - bash: |
+      echo "##vso[task.setvariable variable=UNO_IOS_JOB_START_EPOCH]`date +%s`"
+    displayName: Record job start time
+
   - checkout: self
     clean: true
 
@@ -105,6 +114,8 @@ jobs:
       UITEST_RUNTIME_TEST_GROUP: "${{ parameters.UITEST_RUNTIME_TEST_GROUP }}"
       UITEST_RUNTIME_TEST_GROUP_COUNT: "${{ parameters.UITEST_RUNTIME_TEST_GROUP_COUNT }}"
       UITEST_TEST_TIMEOUT: "${{ parameters.UITEST_TEST_TIMEOUT }}"
+      UITEST_JOB_TIMEOUT_MINUTES: "${{ parameters.JobTimeoutInMinutes }}"
+      UNO_IOS_JOB_START_EPOCH: "$(UNO_IOS_JOB_START_EPOCH)"
       UNO_UITEST_IOSBUNDLE_PATH: "$(build.sourcesdirectory)/build/$(SamplesAppArtifactName)/SamplesApp.app"
       SAMPLESAPP_BUNDLE_ID: "${{ parameters.SAMPLESAPP_BUNDLE_ID }}"
 
@@ -138,6 +149,8 @@ jobs:
       UITEST_RUNTIME_TEST_GROUP: "${{ parameters.UITEST_RUNTIME_TEST_GROUP }}"
       UITEST_RUNTIME_TEST_GROUP_COUNT: "${{ parameters.UITEST_RUNTIME_TEST_GROUP_COUNT }}"
       UITEST_TEST_TIMEOUT: "${{ parameters.UITEST_TEST_TIMEOUT }}"
+      UITEST_JOB_TIMEOUT_MINUTES: "${{ parameters.JobTimeoutInMinutes }}"
+      UNO_IOS_JOB_START_EPOCH: "$(UNO_IOS_JOB_START_EPOCH)"
       UITEST_VARIANT: "${{ parameters.UITEST_VARIANT }}"
       UNO_UITEST_IOSBUNDLE_PATH: "$(build.sourcesdirectory)/build/$(SamplesAppArtifactName)/SamplesApp.app"
       SAMPLESAPP_BUNDLE_ID: "${{ parameters.SAMPLESAPP_BUNDLE_ID }}"
--- a/build/test-scripts/ios-uitest-run.sh
+++ b/build/test-scripts/ios-uitest-run.sh
@@ -47,6 +47,10 @@ UNO_IOS_TESTS_STARTED=false
 report_harness_crash() {
 	local status=$?
 
+	if [ -n "${APP_LOG_PID:-}" ]; then
+		kill -TERM "$APP_LOG_PID" 2>/dev/null || true
+	fi
+
 	if [ "$status" -ne 0 ] && [ "$UNO_IOS_TESTS_STARTED" != "true" ]; then
 		echo "##vso[task.setvariable variable=UNO_IOS_HARNESS_CRASHED]true"
 	fi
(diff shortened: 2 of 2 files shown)


morning4coffe-dev (approved):
Code review assessment: the canary matches the existing app-side producer, pre-test failures retain the harness-crashed retry signal, and the results deadline correctly uses the original job start/budget with a teardown reserve while preserving shorter suite limits. Missing results still fail through the existing transform path; diagnostics are not a success fallback. No blocking issues found. The early process-liveness check in the canary loop and stronger shutdown/erase recovery diagnostics are reasonable non-blocking follow-ups; I did not exercise the simulator-recycle path locally.

PR #24330, opened 2026-09-01
test(hotreload): record a wall-clock baseline for the reload cycle

Jen-Uno:
## What this does

Uno's hot-reload pipeline cannot measure itself end to end. The server records a duration for its own slice, the client records a WebSocket ping RTT, and nothing correlates the two — so there is no edit-to-frame figure and no way to tell whether a change made the loop faster.

This PR adds the spec describing how to close that gap (`specs/057-hotreload-end-to-end-instrumentation/`) plus its R7 benchmark as an HRApp runtime test, so the cycle cost is at least **recorded** before anyone optimises it.

Test-project only: the diff touches `src/Uno.UI.RuntimeTests/Tests/HotReload/Frame/HRApp/**` and `specs/**`. No shipping code, no public API.

## Measured baseline

Three edit tiers, 6 iterations each (1 warm-up + 5 measured), 18/18 samples recorded. Times in ms, edit-to-tree-updated. Warm-up excluded from min/median/max.

| Tier | Edit target | n | Warm-up | Min | **Median** | Max |
|---|---|---|---|---|---|---|
| `small-leaf-page-attribute` | one attribute in a leaf page | 6 | 2050.6 | 728.4 | **731.6** | 744.0 |
| `medium-shared-usercontrol` | a `UserControl` instantiated elsewhere | 6 | 730.6 | 724.9 | **729.7** | 730.8 |
| `large-app-resourcedictionary` | an app-level `ResourceDictionary` entry | 6 | 905.0 | 737.1 | **743.3** | 747.4 |

**Edit size barely moves the number.** A ~14 ms spread across three tiers built to differ by an order of magnitude in scope. The cycle is dominated by fixed constants — the 250 ms `FileSystemObserver` buffer and the 100 ms trailing delay are ~350 ms, roughly half of every median — not by the size of the change. An earlier run of the same tiers on the same machine gave 742.2 / 739.0 / 751.0 ms, so the figures reproduce within ~12 ms across runs.

Scope, so the numbers are not over-read:

- **Hardware**: one macOS development machine (Apple M5 Pro, macOS 26.5.2, .NET 10.0.101, Debug). Not a controlled or uncontended environment, and not the platform CI measures.
- **Span**: **edit-to-tree-updated on the dev-server-owned path — NOT edit-to-frame.** It ends when the local hot-reload operation completes, not when a frame containing the change is presented. R5 is what would make it edit-to-frame. The IDE-owned paths (Visual Studio, VS Code, Rider) never reach `HotReloadHelper` and are not measured at all.

## Degradation

A separate 15-iteration run of the `[Ignore]`d degradation test passed: steady-state median **734.5 ms** (n=14), least-squares trend **−0.90 ms/reload**, 95% CI **[−1.62, −0.17]**.

**A positive slope is excluded at this sample size** — the cycle does not measurably slow down over 15 consecutive reloads. The interval sitting entirely below zero nominally reads as a slight speedup; that is far more likely a warming artifact than real, and no claim is made for it.

**This says nothing about the multi-app / ALC memory accumulation** recorded in `specs/044-alc-memory-leak-fixes` (~+48 MB managed and +150 MB RSS per reload, OOM in 10–15 cycles). That is a different scenario, measured in megabytes; this test never loads a second app. A flat wall-clock curve is not evidence against it. Per-operation memory counters are R6.

Measured on the tree before the review fixes below. The only behavioural change since is the recording gate, and all 15 samples were recorded, so it does not affect these figures.

## Verification status

Labelled precisely, with nothing implied beyond each level:

- **Code review — done.** Four independent passes: correctness/test-fidelity, repository conventions, spec factual accuracy, CI cost. They found real defects; those are disclosed below rather than left to be rediscovered.
- **Compile-verified — yes, but not on this branch's base.** The code builds clean in a space-free copy. **This checkout cannot build at all**: `src/Directory.Build.targets:188` hard-errors on a space in the repository path, and this clone lives under one. A symlink does not help — MSBuild resolves back to the real path.
- **Tests executed and passing — yes, on macOS.** 3 passed / 1 skipped / 0 failed, 18/18 samples, producing the table above. The skip is the `[Ignore]`d degradation test, run separately.
- **NOT verified on this branch's base.** The verified tree is 66 commits behind `origin/master` @ `84b5571e68b7`. Exactly one of those 66 touches the hot-reload area: `6f057b77ddce` (devserver `/rc` keep-alive). Its watchdog is `KeepAliveMessage.Interval` (30 s) × `KeepAliveTimeoutFactor` (2.1) ≈ **63 s without a received frame**; the benchmark drives an edit every ~750 ms, so it stays well inside. That is reasoning from source, not a measurement.
- **NOT verified on Windows** — the only CI leg that runs these tests. No pipeline run attached yet.

## Review findings — fixed before opening

The review found a defect that would have defeated the test's own purpose, plus several smaller ones. All fixed in this branch, and the suite re-run green afterwards:

1. **A green pass was possible with zero hot reloads.** `HotReloadHelper.UpdateAsync` had no old-text guard, and a no-op returns *without* an error — the server reports `NoChanges` and the client returns early (`ClientApi.cs:303`, `:354`). Six no-op samples satisfied both assertions, so the suite could report a green "baseline" measuring nothing but a round trip. Now only a pass with `ApplicationUpdated is true` is recorded, so a no-op trips the completeness assertion instead. That all 18 samples still record is positive evidence the edits genuinely apply.
2. **The duration assertion was vacuous.** `ElapsedMs > 0` is satisfied by any round trip; it now asserts against the ~350 ms of fixed constants.
3. **Spec premise was inverted.** It claimed the 250 ms debounce lands *before* the server's `StartTime`. It does not: `HotReloadManager.cs:141-143` starts the operation before the buffer closes, by design and by comment. The debounce is *inside* `DurationMs`. R2's formula depended on this and was corrected — `StartTime - ObservedAtUtc` would read ≈0.
4. **R1 was described as "one field".** It is not: `ServerHotReloadProcessor.ProcessFrame` has no case for `HotReloadClientOperationEvent` and no `default`, so the client's outcome event is dropped server-side today. R1 also needs a frame handler.
5. Smaller: the 100 ms delay is not "unconditional" (six early returns bypass it — the word had propagated into a code comment); base64 inflation is ~1.33×, not ~4×; const naming corrected to PascalCase per `.editorconfig`; the tier `CancellationTokenSource` instances are now disposed; a displaced `<summary>` restored.

## Open — worth a reviewer's opinion

1. **The measurement runs against an inherited visual tree.** The tests never set `UnitTestsUIContentHelper.Content`, and the engine does not reset content between classes, so whatever ran previously is still mounted. The medium tier's stated purpose (partial tree reload) may therefore not be exercised, and the same test can yield different costs under `--filters` than in a full-suite run. This is a design question, not a typo, so it is left for discussion rather than patched blind.
2. **Spec number 057 is taken** by `specs/057-resourcedictionary-source-location/`. Kept as-is because the directory names are distinct and the repo already carries two `055-` specs; renumbering also means updating references elsewhere. Happy to renumber if preferred.
3. **CI cost.** Nothing gates these out: the Skia Windows runtime-test stage is unconditional, `UnoEnableHRuntimeTests` is set on that job, and no filter excludes the new class. 3 tiers × 6 iterations = 36 additional round trips, roughly **+36 s to +108 s per attempt**, and the outer test retries up to 3×.
4. **`Given_AppDictionary.When_Change_AppResource_LotOfTimes`** already performs 15 edit+revert cycles on the same file with the same strings, and carries a commented-out `//[Ignore("Failing ramdomly on CI")]`. "Why a new test rather than instrumenting that one?" is a fair question this diff does not answer. The shared strings are also a dirty-file risk if a forward edit lands and then cancels.
5. **`HRPERF|` output does not reach the Tests tab.** HRApp sets `IsConsoleOutputEnabled = false` and its results collapse into one outer test, so the summary survives only in the raw step log.
6. **`HotReloadTiming._samples` is process-lifetime** and never cleared (`Clear()` has no callers). Running a scenario twice in one process would fail the count assertion spuriously; not reachable today only because HRApp sets `Attempts = 1`.

## Known gaps this PR does not fix

- **These tests cannot run on macOS as committed.** `Uno.UI.RuntimeTests.HRApp.Skia.csproj:48` references only `Uno.icu-win` (a macOS run needs `Uno.icu-macos`), and HRApp inherits no native assets, so `libUnoNativeMac.dylib` is unresolvable at launch. The baseline above was obtained with both patched locally; the patches are deliberately not in this branch, since HRApp's cross-platform asset story is its own change.
- **HR runtime tests are enabled on one CI leg only.** `UnoEnableHRuntimeTests` occurs exactly once: `build/ci/tests/.azure-devops-tests-windows-skia.yml:69`. Every other leg reports without exercising hot reload — which is why the two gaps above went unnoticed.
- **R1–R6 are unimplemented.** Only R7 lands here, so the correlated stage record this benchmark is ultimately meant to guard does not exist yet.

## Checklist

- [x] Test-project only — no shipping code or public API touched
- [x] Tests executed and passing (macOS): 3 passed / 1 skipped / 0 failed, 18/18 samples
- [x] Conventional Commits (`docs(hotreload):`, `test(hotreload):`)
- [x] Spec added, with every `file:line` citation re-verified against master
- [x] Confirmed review defects fixed and the suite re-run green
- [ ] **Linked GitHub issue — MISSING.** The spec flags this itself (`spec.md:15-16`); an issue must be filed and linked in both the spec header and this description before merge.
- [ ] Release build with zero warnings — not verified; this checkout cannot build
- [ ] Verified on this branch's base (66 commits newer than the verified tree)
- [ ] Verified on the Windows CI leg that actually runs these tests; no pipeline run attached
- [ ] Open items above triaged (inherited visual tree; spec number; CI cost; overlap with `When_Change_AppResource_LotOfTimes`)

### DIFF
--- a/specs/057-hotreload-end-to-end-instrumentation/spec.md
+++ b/specs/057-hotreload-end-to-end-instrumentation/spec.md
@@ -1,0 +1,317 @@
+# Hot-Reload: End-to-End Instrumentation — Make the Edit-to-Frame Cycle Measurable
+
+**Repo**: `uno` (Uno.HotReload, Uno.UI.RemoteControl, Uno.UI.RemoteControl.Server.Processors)
+**Created**: 2026-08-27
+**Status**: Proposed
+**Related**: [spec 050](../050-hotreload-updatefile-workspace-gate/spec.md) (R6 — the per-request
+info file that already carries the correlation id to the app),
+[spec 051](../051-hotreload-baseline-identity-handshake/spec.md) (baseline MVID mismatch — a
+silent drop this instrumentation would make visible),
+[spec 055](../055-hotreload-noop-pass-dedup/spec.md) (the `0.007 s` no-op emit measurement),
+[spec 044](../044-alc-memory-leak-fixes/spec.md) (per-reload memory accumulation),
+[spec 001](../001-fast-devserver-startup/) (`TIMELINE|` staged startup measurement — the model
+this spec follows for the edit cycle)
+
+> **Note**: this spec has no associated GitHub issue yet. One should be filed and linked here
+> before any PR, per the repository's contribution rules.
+
+## Overview & Objectives
+
+Uno's hot-reload pipeline **cannot measure itself end to end**, and the gap is structural
+rather than a matter of missing log lines.
+
+The pipeline already emits three durations. None of them is, or can become, an edit-to-frame
+figure:
+
+- `ServerHotReloadProcessor.cs:221-222` computes
+  `DurationMs = Current.CompletionTime - Current.StartTime` — a **server-side** operation
+  window.
+- `ServerHotReloadProcessor.cs:77` records `WaitDurationMs`, the workspace-gate queue time.
+- `RemoteControlClient.Status` records `_roundTrip`, a WebSocket ping RTT.
+
+The 250 ms watcher debounce (`FileSystemObserver.cs:242`,
+`bufferTimer.Change(250, Timeout.Infinite); // Wait for 250 ms without any file change`) is
+**inside** the server's window, not before it: `HotReloadManager.ProcessFileChanges` awaits
+`StartOrContinueHotReload()` *before* awaiting the file set, and says so in a comment —
+*"Notify the start of the hot-reload processing as soon as possible, even before the buffering
+of file change is completed"* (`HotReloadManager.cs:141-143`). `DurationMs` therefore already
+contains the debounce, which is why it cannot be compared across edits of different shapes.
+
+What `DurationMs` does exclude is everything after the server completes: the 100 ms
+`await Task.Delay(100, ct); // […] this is just for safety.`
+(`ClientHotReloadProcessor.ClientApi.cs:392`, reached on the fully-successful path only — six
+early returns bypass it) and **both visual-tree walks** land **entirely after**
+`CompletionTime`. So `DurationMs` covers a window that both starts too early to be a
+per-edit cost and ends too soon to be a developer-perceived latency, and no amount of careful
+reading turns it into one.
+
+**The blocking defect is narrower than "add instrumentation", but it is not a single field.**
+One field carries the id; a server-side frame handler is also required, because
+`ServerHotReloadProcessor.ProcessFrame` has no case for `HotReloadClientOperationEvent` and no
+`default`, so the client's outcome event is sent and silently dropped today. A
+correlation id already spans the server and reaches the app:
+`UpdateFileResponse.HotReloadCorrelationId` is the `HotReloadServerOperation.Id`
+(`HotReloadInfoHelper.cs:76` documents exactly this), and the client blocks on it in
+`WaitForServerHotReloadAsync` (`ClientApi.cs:350`). But when the client reports its own
+outcome it sends:
+
+```csharp
+// ClientHotReloadProcessor.Common.Status.cs:769-777
+_ = client.SendMessage(new Messages.HotReloadClientOperationEvent
+{
+    OperationSequenceId = Id,   // <- client-LOCAL sequential id, not the server correlation id
+    StartTime = StartTime,
+    …
+});
+```
+
+`Id` is the client's own counter. **The client half of every hot-reload operation therefore
+cannot be joined to the server half.** Two accurate timelines exist and there is no key
+between them.
+
+### Key objective
+
+Make one hot-reload cycle observable as a single correlated record spanning file-change
+observation to rendered frame, so that (a) a regression is detectable, (b) the stage that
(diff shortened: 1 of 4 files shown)


Jen-Uno:
Closing: this was exploratory measurement work, not a proposed change. The wall-clock figures it reported are useful as a one-off investigation, but the branch modifies a shared hot-reload test helper (`HotReloadHelper.cs`, which every HR runtime test goes through) and that is not a change worth carrying for a benchmark. Withdrawing rather than leaving it open for review.

The findings from the investigation are recorded separately and need no code change.

Issue #2159, opened 2019-11-22
[ios] Keypad is zooming in by changing any calculator 


Hpancha:
### **Steps To Reproduce:**

1. Launch an UNO Calculator app on any ios device.
2. Now change the type of the calculator by selecting from the burger menu.
3. Observe the keypad.

### **Actual behavior:**

- The keypad is zooming in every time the user changes the calculator type.

### **Expected behavior:**

- The keypad has to be stable as appear in Android app.

### **Attachment:**

[Keypad_issue.zip](https://github.com/unoplatform/calculator/files/3880314/Keypad_issue.zip)

### **Device and Application Information:**
 - **OS Build:** ios 13
 - **Build Version:** 1.2.0-uno.572
 - **Devices:**  iPhone XR (13)

### **Note:**
- Refer the attached video to get better idea.

jeromelaban:
As weird as it is, this is not a bug. This may get changed when we update to a later version of the original source.

Issue #2157, opened 2019-11-22
[Calculator][Android] Last drop-down option isn't visible for some converters


Hpancha:
### **Steps To Reproduce:**

1. Launch an UNO Calculator app on Android device.
2. Select any converter from **Volume,** **Weight and Mass,** or **Data**
3. Put the app in the landscape mode.
4. Select the drop-down for 1st value and scroll all the way down.
5. Observe the last option.

### **Actual behavior:**

- The last option for the drop-down does not appear and it's cutting from the bottom. 
- I could reproduce this issue only in 2 devices. May be it's related to the edge screen of the phone but I'm still not sure.
- This issue doesn't occur for the drop-down of 2nd value.

### **Expected behavior:**

- User can see all the options clearly without any trimming issue.

### **Attachment:**

![trimming_issue](https://user-images.githubusercontent.com/51453677/69457495-27c9f580-0d3b-11ea-8e55-a66465e54a53.jpg)

### **Device and Application Information:**
 - **OS Build:** android
 - **Build Version:** 1.2.0-uno.558
 - **Devices:**  Samsung Note 10+ (9), Samsung S10+ (9)

### **Note:**
- Refer the attached image to get better idea.

sakshi173:
This issue is fixed. Below is the screenshot.
![calc2](https://user-images.githubusercontent.com/74563602/120375476-d7950180-c2e8-11eb-8faa-0d66622dd00c.png)

Youssef1313:
@MartinZikmund Should this get closed based on @sakshi173 comment?

sakshi173:
Verified issue is fixed, closing.

Issue #2156, opened 2019-11-22
[iOS][Android] Reveal password button not releasing


carlh98:
## Current behavior

When the password box is focused and you click on the reveal image (on the right) and you drag your finger outside of the password box area, the password is still revealed

## Expected behavior

When the password box is focused and you click on the reveal image (on the right) and you drag your finger outside of the password box area, the password should be hidden

## How to reproduce it (as minimally and precisely as possible)

- Launch App
- Enter in a new password
- While the password box is still in focus, press and hold the show password button then drag your finger down and let go
- [Android] if the previous step does not work, click on the show password
- The password will still be visible

Sample:
[RevealPassword.zip](https://github.com/unoplatform/uno/files/3881277/RevealPassword.zip)

## Environment

Nuget Package:
Uno.UI : 2.0.512-dev.4073

Package Version(s):

Affected platform(s):

- [x] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] WebAssembly renderers for Xamarin.Forms
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

Visual Studio:

- [ ] 2017 (version: )
- [ ] 2019 (version: )
- [ ] for Mac (version: )

Relevant plugins:

- [ ] Resharper (version: )

PR #24351, opened 2026-09-02
docs: Close out the 7.0 documentation and reconciliation leftovers

MartinZikmund:
**GitHub Issue:** relates to #8339

Closes the following internal 7.0 items (referenced in the commit trailers):
uno-private#2050, #2055, #2056, #2057, #2098, #2126, #2135, #2156, #2157, #2284.
Advances #2112, #2114 and #2159, which stay open on stated dependencies.

## PR Type:

(plus removal of the leftover WASM DOM-renderer surface, one build fix and one project-file fix; see below)

## What changed? 🚀

An audit of the 7.0 epic found that the release's biggest systemic gap is **documentation, not code**. For 25 of 106 sub-issues, the only outstanding work was a missing `migrating-to-uno-7.md` entry. Several source-breaking changes had shipped with no consumer-facing old → new mapping. This PR closes out those leftovers, one commit per item. It also removes the last orphaned pieces of the WASM DOM renderer.

**Migration guide entries for breaking changes that shipped undocumented**

- **`Background` moved off `FrameworkElement`** (BC38): which eight types keep it and which lose it, why `<Setter Property="Background">` stops resolving in a style whose target type doesn't declare it, and how to migrate.
- **`DataContext` restricted to `FrameworkElement`** (BC58): this changes only the API surface. `{Binding}` on a non-`FrameworkElement` object (`Brush`, `Run`, `ColumnDefinition`, …) still resolves through the inheritance context, as in WinUI. Uno's multi-parent sharing is stricter than WinUI's, though: it lacks WinUI's `ResourceDictionary` and `ContentControl` exceptions, so a brush resource shared by several elements stops resolving its bindings, and the entry says so. The entry also explains why flyout content still binds against its target.
- **`IAnimatedVisualSource`**: maps the nine-method Uno-only contract to WinUI's `TryCreateAnimatedVisual` with an old → new table, and notes that unmodified LottieGen output now compiles.
- **`DependencyObject` is a class** (BC26): `partial` is no longer required, and a type can no longer inherit another base class and pick up `DependencyObject` alongside it.
- **`UserControl : Control`** (BC14): the lost `ContentControl` surface, and `Content` retyped from `object` to `UIElement`.
- **`DependencyPropertyValuePrecedences`** (BC39): the removed and renamed enum members, which consumers can reach through the public `DependencyObjectExtensions` precedence overloads.
- **SkiaSharp 4** and **Vulkan** as the new 7.0 defaults. Both are user-visible behavior changes.
- **WASM DOM interop**: the removed surface (`HtmlElementAttribute`, `HtmlCustomEventArgs`, `UIElement.HtmlId`, …). The entry now points to `BrowserHtmlElement` as the replacement instead of listing it as removed.

**Docs that still taught a removed model**

- `api-differences.md` and `uno-internals-overview.md` still described `UIElement` inheriting a native view type and `DependencyObject` as an interface. They also linked a source-generator file deleted in `c371084bb52`.
- `migrating-guidance.md` told readers to add `partial` when inheriting `DependencyObject`, which is no longer true. Its deep link pointed at a heading this PR renames.
- `feature-flags.md` documented the three `UseVulkanOn*` flags as `Default: false`; they are `true` (`FeatureConfiguration.cs:520,527,539`).
- `using-vulkan-rendering.md` called Vulkan opt-in, but it is the default. This PR adds a **Disabling Vulkan** section and rewrites the fallback list to match the new default.
- `using-skia-desktop.md` called SkiaSharp 4 opt-in, but it is the default.
- `windows-ui-xaml-window.md` said a second window throws on both Android and iOS. iOS gained scene-backed windows in #19083, so the limitation now applies only to Android, and the doc spells out the iOS opt-in requirements.
- `platform-specific-csharp.md` corrections:
  - `__APPLE_UIKIT__` is defined only by `src/Uno.CrossTargetting.targets`, which is repo-internal and in no nuspec, so consumers never get it.
  - `__UNO_SKIA__` ships in the `Uno.WinUI.Runtime.Skia.*` packages. Android, iOS, tvOS and browser-WASM class libraries still get it through those packages' build assets, but `net10.0-desktop` and plain `net10.0` class libraries don't, so the row now says exactly that.
  - The type-alias sample aliased a UWP namespace that 7.0 doesn't ship, and the UWP/`NETFX_CORE` row is gone.
  - The `||` in the symbols table is escaped so the rows render.
- `migrating-from-previous-releases.md` stopped at 6.6, so nothing linked readers to the 7.0 guide.

**Removal of the leftover WASM DOM-renderer surface**

The DOM renderer was already gone, but its browser-side layer and a few consumers were still in the tree. None of it was compiled or shipped: on `master`, no project file references `src/Uno.UI/ts`, `tsBindings` or `WasmCSS`, except the two cross-project includes moved below.

- **Deleted, orphaned:**
  - `src/Uno.UI/ts/**`: `WindowManager.ts`, `HtmlDom.ts`, `MonoSupport.ts`, the `Interop/**` Mono shims and the `Windows/**` DOM extensions.
  - `src/Uno.UI/tsBindings/**` and `src/Uno.UI/tsconfig.json`.
  - `src/Uno.UI/WasmCSS/Uno.UI.css`.
- **Public type removed:** `Uno.Extensions.HtmlCustomEventArgs`. Its only producer, `UIElementWasmExtensions`, was removed earlier, so no app could receive one. It is covered in both PackageDiffIgnore IgnoreSets.
- **Consumers removed:**
  - The `Wasm_CustomEvent` sample and its `ItemExclusions.props` block.
  - The six UI tests driving that sample, in `Given_CustomEvents` and `WasmTests`.
  - `Given_HtmlElementAttribute.wasm.cs` and its `Compile` include.
  - The `HtmlElementAttribute` probe in `XamlFileGenerator`, which can no longer resolve.
- **Dead code:** removed the `#if __WASM__` cleanup block from `Application.Alc.cs`. `__WASM__` is never defined for `Uno.UI`, and none of the three members the block called exist anymore.
- **Moved to their real consumers:**
  - `WebView.ts` moves to `Uno.UI.Runtime.Skia.WebAssembly.Browser/ts/Runtime/`. That project was its only consumer and already `<Link>`ed it there.
  - The ambient `ts/types/**` `.d.ts` files move to `Uno.WinRT`, which the browser runtime already references.
  - The cross-project `TypeScriptCompile` and `tsconfig.json` entries are updated to match, and `specs/043-drop-native-rendering/progress.md` records the new locations.

**Build and project-file fixes**

- **`build/PackageDiffIgnore.xml`**: BC62 renamed `Application.TraceProvider.LauchedStart`/`LauchedStop` to `Launched*`, and the package diff reads that as two removed public fields. Neither old name had an ignore entry, so the Phase 4 "package-diff clean" clause was unmet. The entries are added to **both** the 6.6 and 6.7 IgnoreSets, which are maintained as identical copies (the same applies to `HtmlCustomEventArgs` above).
- **`UnoAppWinUILinuxValidation`** pinned `SkiaSharp.Skottie` at `2.88.7`, while the repo ships `4.151.1`. That folder's `Directory.Build.targets` deliberately does not chain to `src/Directory.Build.targets`, so the central `PackageReference Update` never retargets it. The stale pin was live and conflicted with the SkiaSharp-4-based `Uno.WinUI.Lottie` that the same head references.

**Spec reconciliation**: the BC26, BC39 and BC58 impact specs still described their *plans* rather than what shipped, including "open decision" sections the code had already answered. Two plan claims were false against the tree and are corrected: `DependencyObjectStore` no longer exists (it was folded into `DependencyObject`), and `DependencyObjectGenerator` was deleted outright rather than trimmed. The specs cite only public references.

## Validation

- **Docs, compile only.** When the doc commits were authored, a full local `docfx docfx.json` produced a warning/error set **byte-identical to the base commit**: 178 warnings / 16 errors, 196 normalized lines, with no new diagnostics. That build caught one regression, a renamed heading that broke an inbound `#anchor` deep link, and it is fixed. On the current head, the `Uno.UI - docs` pipeline and `Check Docs links` pass. No browser validation of the rendered pages was performed.
- **WASM DOM-renderer removal, compile only.**
  - `Uno.UI` (net10.0) builds with 0 warnings and 0 errors.
  - The SamplesApp desktop head and `SamplesApp.UITests` build clean.
  - `Uno.UI.SourceGenerators.Tests` has the same failure set with and without the `XamlFileGenerator` change (21 pre-existing failures).
  - After the TypeScript move, the browser runtime's emitted `Uno.Runtime.Wasm.js` is unchanged: 6380 lines, with the same `WebView` and `BrowserPointerInputSource` occurrence counts.
- **Runtime.** No runtime validation was performed. The deleted code was not compiled into any shipped assembly, and the only C# behavior change is the removal of an already-dead `#if __WASM__` block.
- **Other C# edits** (`TextElement.cs`, `Given_DependencyProperty.cs`) change comments only.

## PR Checklist ✅

- [x] 🧪 Added [Runtime tests, UI tests, or a manual test sample](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) (for bug fixes / features, if applicable). N/A: no new behavior. This PR removes one sample, six UI tests and one runtime test that exercised the removed DOM-interop surface.
- [x] 🖼️ Validated PR `Screenshots Compare Test Run` results. N/A: no rendering change, since the deleted CSS/TypeScript was not compiled or shipped.
- [ ] ❗ Contains **NO** breaking changes. See below.

**Breaking change:** this PR removes the public `Uno.Extensions.HtmlCustomEventArgs` type. Nothing could raise it once the DOM renderer was removed, so apps can only be affected at compile time. The 7.0 migration guide lists it with the rest of the removed WASM DOM-interop surface and points to `Uno.UI.NativeElementHosting.BrowserHtmlElement` as the replacement. Everything else here *documents* breaking changes already on `master`, and introduces none.

## Note for reviewers

This PR does **not** resolve one item flagged during the audit. `UserControl` lost the entire `ContentControl` surface and had `Content` retyped, yet `PackageDiffIgnore.xml` has no `UserControl` entries. By comparison, the analogous `MediaPlayerPresenter` reparent needed explicit entries for exactly that kind of inherited-member loss. Settling it needs a pack plus a `generatepkgdiff` run. A green CI diff would not settle it either, because `Uno.PackageDiff` has been unpublished since 2025-02-24 (`uno-private#2312`).

https://claude.ai/code/session_01GDmguoPp2ywTHVJQjCP8QD


doc/articles/migrating-to-uno-7.md
```diff
@@ -553,6 +615,59 @@ change only breaks code that used the Uno-only members leaked by the wrong base.
...
+
+  Non-`FrameworkElement` `DependencyObject`s lose the member entirely — `Brush` and its
+  subclasses, `Transform`, `GradientStop`, `Setter`, `Style`, `DependencyObjectCollection`,
+  `ElementFactory`, `FlyoutBase`, and so on. Reading or setting `.DataContext` on one, or
+  subscribing to its `DataContextChanged`, is now a compile error, as is referencing
+  `Brush.DataContextProperty` and friends.
+
+  **This one can break silently.** Uno let a non-`FrameworkElement` inherit the ambient
+  `DataContext`, so `{Binding}` on a `Brush`, `Transform` or `Setter` resolved against the
+  surrounding view model — a pattern WinUI never supported. Those bindings no longer resolve,
```
morning4coffe-dev, line 661: [P2] A FrameworkElement-only DataContext property does not mean WinUI never supports ambient bindings on non-FrameworkElements. WinUI's [BindingExpression::AttachToDataContext](https://github.com/microsoft/microsoft-ui-xaml/blob/d0c6705ec46dbddb853cdfdb2eb3b42849cf304e/dxaml/xcp/dxaml/lib/BindingExpression_Partial.cpp#L566-L601) explicitly finds a mentor when the target is not an FE, then subscribes to and reads that mentor's DataContext. CBrush derives from CMultiParentShareableDependencyObject, whose [GetMentor implementation](https://github.com/microsoft/microsoft-ui-xaml/blob/d0c6705ec46dbddb853cdfdb2eb3b42849cf304e/dxaml/xcp/components/DependencyObject/CMultiParentShareableDependencyObject.cpp#L20-L50) supports inheritance-context parents, subject to its sharing rules. Please remove the blanket 'WinUI never supported' claim and distinguish any actual Uno 7 binding limitation from API-surface parity.

specs/050-breaking-changes-rollup/bc26-dependencyobject-as-class.md
```diff
@@ -35,14 +35,20 @@ The interface only ever existed because **native mobile `UIElement` must inherit
...
 
-## Open decision (needs maintainer confirmation)
+## Decision (resolved)
 
-- **End-state design:** class on **all** remaining targets (recommended, since native is gone) vs. Skia/Wasm-only as the old issue scoped it.
-- **How much of the generated mixin folds into the base class** vs. stays emitted per-type. Settle this *before* coding — it defines the whole diff.
+- **End-state: a class on every remaining target**, not the Skia/Wasm-only scoping of the original 2024 issue. `src/Uno.UI/UI/Xaml/DependencyObject.cs` is now `public partial class DependencyObject : IDependencyObjectInternal, IWeakReferenceProvider`.
+- **The whole mixin folded into the base class.** Rather than trimming what `DependencyObjectGenerator` emitted, the generator was removed outright (deleted in `c371084bb52`); DO detection is now by base type. Nothing DO-related is emitted per-type any more.
+- Phase 2 went further than this spec anticipated: `DependencyObjectStore` was **dropped entirely** (uno-private#2210) and its storage folded onto `DependencyObject` itself — see `DependencyObject.Store.cs` / `.Binder.cs`.
```
morning4coffe-dev, line 42: [P2] Please keep these public reconciliation specs independent of private tracking artifacts, as required by AGENTS.md's Public Documentation and Spec References section. This addition and the corresponding BC58/roll-up additions refer to private tracker identifiers. Cite the public Uno PRs/source commits instead, or describe downstream tracking generically without identifiers; readers of the public source should not need private context.

morning4coffe-dev (changes requested):
Two documentation corrections are needed before approval: the migration guide conflates the FrameworkElement-only DataContext API with WinUI's separate mentor-based binding mechanism, and the public reconciliation specs introduce references to private tracking artifacts. I checked the binding claim directly against microsoft-ui-xaml source. Also, please refresh the PR description/validation scope: this head now includes DOM-runtime cleanup and removal of HtmlCustomEventArgs, not only the two comment-only C# edits described there.

doc/articles/migrating-to-uno-7.md
MartinZikmund: Agreed, the claim was wrong. WinUI resolves `{Binding}` on a non-FrameworkElement target through the mentor lookup in `AttachToDataContext`, and Uno 7 does the same: BC58 changed only the API surface. #23547 kept the inheritance context, and `Given_NonFE_DataContextBinding` covers `ColumnDefinition.Width`, `RowDefinition.Height`, `Run.Text` and `StateTrigger.IsActive`, which were confirmed against native WinUI.

I rewrote the section. It now says that bindings on these objects keep resolving, and that the only thing that is gone is giving such an object a `DataContext` of its own. It also notes that the WinUI sharing rules still apply: an object shared by several owners loses its inheritance context, and `Setter.Value` does not resolve `{Binding}` on WinUI. Checklist item 15 and the Decision/Pros/Cons in the BC58 spec are corrected to match.

specs/050-breaking-changes-rollup/bc26-dependencyobject-as-class.md
MartinZikmund: Fixed. The private tracker references in `bc26-dependencyobject-as-class.md` (two places), `bc58-datacontext-frameworkelement-only.md` and `spec.md` now cite the public PR that removed `DependencyObjectStore`, unoplatform/uno#23702. No private identifiers remain in `specs/050-breaking-changes-rollup`.

doc/articles/features/using-vulkan-rendering.md
```diff
@@ -103,9 +103,31 @@ To force software rendering on both paths, disable each one:
...
+FeatureConfiguration.Rendering.UseVulkanOnWin32 = false;
+```
+
 ## Fallback Behavior
 
-When Vulkan is requested but unavailable, the application automatically falls back to the next available backend:
+When Vulkan is unavailable, the application automatically falls back to the next available backend:
 
 1. **Vulkan** (if requested)
 2. **OpenGL / OpenGL ES** (platform default)
```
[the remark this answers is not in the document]
MartinZikmund: Fixed. The section now opens with "When Vulkan is unavailable or has been disabled". The list shows Vulkan as the default on Android, Linux (X11) and Windows (Win32), and the stale "(if requested)" and "(platform default)" qualifiers are gone, which matches the fallback order in `Win32RenderingBackend`.

src/Directory.Build.targets
```diff
@@ -266,6 +266,17 @@
...
+	<!--
+		Microsoft.TypeScript.MSBuild's TypeScriptDeleteOutputFromOtherConfigs deletes the files
+		listed by every Tsc*.out found under obj\ other than its own, assuming they belong to
+		another configuration. Under multi-targeting the sibling TFM's obj\<config>\<tfm>\ looks
+		exactly like that, so each inner build deletes the other's freshly emitted javascript\*.js
+		- which then breaks the embedded resource injection that reads it. The stale-configuration
+		cleanup buys us nothing: that JavaScript is embedded into the assembly, never read back
+		from bin\. This file is imported after the NuGet package targets, so this override wins.
+	-->
+	<Target Name="TypeScriptDeleteOutputFromOtherConfigs" />
```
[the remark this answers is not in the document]
MartinZikmund: This no longer applies. The `TypeScriptDeleteOutputFromOtherConfigs` override has been removed from `src/Directory.Build.targets` entirely: it went out on `master` along with the native WASM-DOM project heads (c3608f48c4). Since the rebase onto `master`, this PR does not touch that file.

src/Uno.UI.Runtime.Skia.WebAssembly.Browser/Uno.UI.Runtime.Skia.WebAssembly.Browser.csproj
```diff
@@ -67,11 +67,8 @@
 		<EmbeddedResource Include="WasmCSS\**\*.css" />
 		<TypeScriptCompile Include="ts\**\*.ts;ts\**\*.tsx" />
 
-		<TypeScriptCompile Include="..\Uno.UI\ts\types\**\*.ts" />
-
-		<TypeScriptCompile Include="..\Uno.UI\ts\Windows\UI\Xaml\Controls\WebView.ts">
-			<Link>ts\Runtime\WebView.ts</Link>
-		</TypeScriptCompile>
+		<!-- Ambient .d.ts for the browser APIs; owned by Uno.WinRT, which compiles them too. -->
+		<TypeScriptCompile Include="..\Uno.WinRT\ts\types\**\*.ts" />
```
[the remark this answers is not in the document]
MartinZikmund: `WebView.ts` is still compiled and embedded. It was moved, not dropped. The "Move the shared TypeScript to its real consumers" commit moves the file from `src/Uno.UI/ts/Windows/UI/Xaml/Controls/WebView.ts` into this project at `ts/Runtime/WebView.ts`, which is the same path the old `<Link>` gave it. The existing `<TypeScriptCompile Include="ts\**\*.ts;ts\**\*.tsx" />` glob picks it up, so an explicit item would duplicate it. I compared the emitted `Uno.Runtime.Wasm.js` before and after the move and it is unchanged, with the same line count and the same number of `WebView` occurrences.

src/Uno.UI.Runtime.Skia.WebAssembly.Browser/tsconfig.json
```diff
@@ -13,8 +13,7 @@
 	"lib": [ "ES2018", "DOM", "DOM.Iterable" ]
   },
 	"include": [
-		"../Uno.UI/ts/types/**/*",
-		"../Uno.UI/ts/Windows/UI/Xaml/Controls/WebView.ts",
+		"../Uno.WinRT/ts/types/**/*",
```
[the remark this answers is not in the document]
MartinZikmund: The file was not retired. It now lives in this project at `ts/Runtime/WebView.ts`, so the `"ts/**/*"` include already covers it, and a separate entry for the old `src/Uno.UI` path would point at a file that no longer exists. You are right that the progress note was out of date: it still gave the pre-move location. I updated `specs/043-drop-native-rendering/progress.md` to say where `WebView.ts` and the ambient `ts/types/**` ended up (`src/Uno.WinRT/ts/types`).

specs/050-breaking-changes-rollup/bc26-dependencyobject-as-class.md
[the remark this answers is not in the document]
MartinZikmund: Fixed. All four references (`bc26-dependencyobject-as-class.md` lines 42 and 50, `bc58-datacontext-frameworkelement-only.md` line 46, `spec.md` line 327) now cite the public unoplatform/uno#23702 instead of the private tracker.

morning4coffe-dev (approved):

src/Uno.UI.Runtime.Skia.WebAssembly.Browser/Uno.UI.Runtime.Skia.WebAssembly.Browser.csproj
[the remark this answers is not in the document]
MartinZikmund: No change needed: the glob already covers these files. MSBuild's `*.ts` wildcard matches any file name ending in `.ts`, and `.d.ts` files end in `.ts`. So `..\Uno.WinRT\ts\types\**\*.ts` picks up every file in that folder. All five are `index.d.ts` (`css-font-loading-module`, `emscripten`, `requirejs`, `webmidi`, `wicg-file-system-access`), so the comment is accurate as written.

The same folder is also listed in this project's `tsconfig.json` `include` (`../Uno.WinRT/ts/types/**/*`), and that list is what `tsc` actually compiles from. Narrowing the item to `*.d.ts` would not change the compiled set, so I left the item using the same `ts\**\*.ts` pattern as `Uno.WinRT.Wasm.csproj`.

doc/articles/migrating-to-uno-7.md
```diff
@@ -553,6 +615,64 @@ change only breaks code that used the Uno-only members leaked by the wrong base.
...
+
+  **`{Binding}` on these objects keeps working.** A non-`FrameworkElement` has no
+  `DataContext` of its own, but — as in WinUI — its bindings resolve against the
+  `DataContext` of the `FrameworkElement` it is attached to, through its inheritance
+  context. `ColumnDefinition.Width`, `RowDefinition.Height`, `Run.Text` and
+  `StateTrigger.IsActive` all still bind to the surrounding view model. The change is to the
+  API surface only: an object like this can no longer be given a `DataContext` of its own, so
+  bind on the owning element or set the value from code instead.
+
+  The inheritance context follows WinUI's rules. An object shared by several owners — a brush
```
morning4coffe-dev: This does not follow WinUI's complete multi-parent rule: WinUI retains inheritance context for a first `ResourceDictionary` parent and limited `ContentControl`/`ScrollContentControl` sharing, while Uno's `AssociateParent` explicitly omits those exceptions. Please document Uno's stricter behavior instead of claiming parity, and update the PR body's matching claim.

doc/articles/platform-specific-csharp.md
```diff
@@ -57,10 +57,10 @@ The following conditional symbols are predefined for each Uno platform:
 | Android         | `__ANDROID__`      | |
 | iOS             | `__IOS__`          | |
 | tvOS            | `__TVOS__`         | |
-| iOS or tvOS     | `__APPLE_UIKIT__`  | |
+| iOS or tvOS     | `__APPLE_UIKIT__`  | Defined inside the Uno Platform repository only — **not** in consumer projects. In your own code write `__IOS__ \|\| __TVOS__`. |
 | WebAssembly     | `__WASM__`         | Only available in the `net10.0-browserwasm` target framework |
 | Desktop         | `__DESKTOP__`      | Only available in the `net10.0-desktop` target framework. |
-| Skia            | `__UNO_SKIA__`     | |
+| Skia            | `__UNO_SKIA__`     | Defined by the `Uno.WinUI.Runtime.Skia.*` packages, which the Uno.Sdk references only for executables. A **class library** does not get it — use `__DESKTOP__`, `__WASM__` or a runtime check instead. |
```
morning4coffe-dev: Only a library targeting plain `net10.0` lacks this symbol. Platform-targeted Android/iOS/tvOS/browser-WASM libraries receive the runtime package's build assets and therefore do get `__UNO_SKIA__`. Please qualify this row accordingly and correct the PR body, which repeats that the SDK references these packages only for executables.

morning4coffe-dev (changes requested):
The earlier mentor-binding correction and private-reference cleanup are addressed, but the current 7.0 guidance still has four correctness inconsistencies. The two changed-line issues are called out inline: WinUI multi-parent parity is overstated, and platform-targeted class libraries are incorrectly excluded from `__UNO_SKIA__`.

Two directly related stale guidance surfaces are outside the current diff and therefore could not be anchored inline: `doc/articles/migrating-guidance.md:31-40` and `AGENTS.md:146-165` still describe the removed native-renderer inheritance/UWP model, while `.claude/rules/source-generators.md:12-14` describes the deleted `DependencyObjectGenerator` as if it survived in reduced form. Please update those plus the matching claims in the PR description before approval.

specs/050-breaking-changes-rollup/bc58-datacontext-frameworkelement-only.md
```diff
@@ -1,6 +1,6 @@
 # BC58 — `DataContext` on `FrameworkElement` only (root cause) · folds in BC54
 
-**Epic:** [#8339](https://github.com/unoplatform/uno/issues/8339) · **Issue:** [#13201](https://github.com/unoplatform/uno/issues/13201) (OPEN umbrella) · **Also resolves BC54** ([#12491](https://github.com/unoplatform/uno/issues/12491), closed as dup) · **Danger:** 4/5 · **Effort:** M · **Phase:** 7 (ship last, own PR)
+**Epic:** [#8339](https://github.com/unoplatform/uno/issues/8339) · **Issue:** [#13201](https://github.com/unoplatform/uno/issues/13201) (OPEN umbrella) · **Also resolves BC54** ([#12491](https://github.com/unoplatform/uno/issues/12491), closed as dup) · **Danger:** 4/5 · **Effort:** M · **Phase:** 7 (ship last, own PR) · **Status:** ✅ implemented
```
morning4coffe-dev, line 3: This spec is marked implemented, but the opening design sections still describe the deleted `DependencyObjectGenerator` and `DependencyObjectStore` as current and prescribe generator gating/internal flyout forwarding that the later “As implemented” section says did not ship. The rollup also leaves BC58/BC54 unchecked, and BC26 retains the same plan-versus-outcome contradiction. Please label the old sections as the original plan or replace them with shipped behavior, then mark the completed rollup items consistently.

morning4coffe-dev (changes requested):
The force-push is a pure rebase, so the September 15 blockers remain at `6bc95ea`: the unresolved inline claims still overstate WinUI inheritance-context parity and incorrectly exclude all class libraries from `__UNO_SKIA__`, with both repeated in the PR body. Native-renderer/UWP and deleted-source-generator guidance also remains stale. Additionally, the BC26/BC58 reconciliation still mixes implemented outcomes with pre-implementation instructions and leaves the BC58/BC54 rollup unchecked.

doc/articles/migrating-to-uno-7.md
MartinZikmund: Good catch, you're right. Fixed in c460e052ac. The paragraph now says Uno tracks the inheritance context more strictly than WinUI: a second owner permanently turns it off. It also names the two exceptions in WinUI's `CMultiParentShareableDependencyObject::GetMentor` that Uno's `AssociateParent` doesn't implement:
- a first `ResourceDictionary` parent
- a first `ContentControl`/`ScrollContentControl` parent, with at most two parents

It spells out the case people will actually hit: a brush resource shared by several elements keeps its bindings on WinUI but loses them on Uno. The PR body's matching claim is corrected too, and 3104933855 fixes the same wording in the BC58 spec's Decision section.

doc/articles/platform-specific-csharp.md
MartinZikmund: Agreed. Fixed in c460e052ac, and the PR body is corrected. I checked `Uno.Implicit.Packages.ProjectSystem.*.targets`: Android, iOS, tvOS and WASM reference the runtime package with `IncludeAssets="build"` when `_IsExecutable` is false, so those class libraries do get `__UNO_SKIA__`.

One more nuance: plain `net10.0` isn't the only exception. The Desktop targets reference `Uno.WinUI.Runtime.Skia.{Win32,X11,MacOS,Linux.FrameBuffer}` only when `_IsExecutable` is true. `Uno.Common.Desktop.targets` then adds `__DESKTOP__` but not `__UNO_SKIA__`, so a `net10.0-desktop` class library lacks it too. The row now says Android/iOS/tvOS/browser-WASM libraries get the symbol, while `net10.0-desktop` and plain `net10.0` libraries don't (use `__DESKTOP__` or a runtime check).

doc/articles/migrating-to-uno-7.md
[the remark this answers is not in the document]
MartinZikmund: Fixed in c460e052ac. This is the same point as the earlier review thread on this line, and I followed your suggested rewording: the paragraph now describes Uno's stricter behavior and names the `ResourceDictionary`/`ContentControl` exceptions Uno doesn't replicate. The BC58 spec's Decision section had the same overstated parity claim and is fixed in 3104933855.

doc/articles/platform-specific-csharp.md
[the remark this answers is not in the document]
MartinZikmund: Fixed in c460e052ac. This is the same point as the earlier review thread on this line. The remark is narrower now, but it isn't limited to plain `net10.0`: `net10.0-desktop` class libraries also lack `__UNO_SKIA__`, because the Desktop targets reference the Skia runtime packages only for executables and `Uno.Common.Desktop.targets` adds just `__DESKTOP__`. So the row says Android/iOS/tvOS/browser-WASM libraries get it, while `net10.0-desktop` and plain `net10.0` libraries don't.

specs/050-breaking-changes-rollup/bc58-datacontext-frameworkelement-only.md
MartinZikmund: Fixed in 3104933855.

**BC58 spec**
- The TL;DR is now past tense.
- "Current state" is now "Original state (before the change)", with a note that the generator and `DependencyObjectStore` no longer exist.
- "What changes" is now "Original plan (superseded)" and says steps 1-3 didn't ship as written.
- The file list and validation sections are labeled as the original plan.
- "As implemented" cites #23547 and notes that BC26 later deleted the generator.
- The Decision section no longer claims full `GetMentor` parity.

**BC26 spec:** the same treatment. "What changes" is labeled as the superseded plan (the generator was deleted rather than trimmed, and `DependencyObjectStore.cs` went in phase 2), and validation is labeled as the original plan. Both specs note that their "BC58 first" sequencing held (#23547 merged before #23537).

**Rollup:** BC58 and BC54 are checked off with PR #23547, and their notes and file lists now describe what shipped rather than the generator/internal-forwarding plan. BC14 (#23566) and BC38 (#23573) had also landed but were still unchecked, so I checked them off the same way and updated their file lists to files that still exist.

AGENTS.md
```diff
@@ -169,9 +169,11 @@ This applies to the **UI rendering layer** (`Uno.UI` native views), *not* to pla
 
 XAML files are parsed to C# via source generators (`XamlFileGenerator` in `Uno.UI.SourceGenerators`), not .xbf like WinUI. Generates `InitializeComponent()`, named fields, and x:Bind expressions.
 
-### DependencyObject on Mobile
+### DependencyObject
 
-On Android/iOS, `DependencyObject` is an **interface** (not base class) since `UIElement` must inherit from native view classes. Source generators provide the implementation via `DependencyObjectGenerator`.
+`DependencyObject` is a **class** on every target, as in WinUI — inherit from it directly.
+
+Before 7.0 it was an *interface*, because the native Android/iOS renderers forced `UIElement` to inherit a native view class, and a `DependencyObjectGenerator` supplied the implementation. Rendering is Skia everywhere now, so the interface and that generator are both gone.
```
morning4coffe-dev, line 173: [P2] This sentence correctly says rendering is Skia everywhere, but the preceding architecture section still documents native Android/iOS/DOM renderers as live maintenance targets and lists their removed inheritance hierarchies. The same obsolete model remains in `doc/articles/migrating-guidance.md`, `.claude/agents/architect.md`, and `.claude/rules/platform-targeting.md`. Please reconcile all four guidance surfaces with the Uno 7 Skia-only UI model.

.claude/agents/architect.md
```diff
@@ -43,7 +43,7 @@ This is the Uno Platform framework — a single C#/XAML codebase implementing th
...
 - **Platform-specialization mechanism (`.claude/rules/platform-targeting.md`):** Is the *narrowest fitting* mechanism chosen — file suffix (`.skia.cs`/`.Android.cs`/`.UIKit.cs`/`.wasm.cs`/`.reference.cs`/`.crossruntime.cs`) vs `#if` vs `OperatingSystem.IsX()` runtime checks vs `ApiExtensibility`? Flag `OperatingSystem.IsAndroid()` used for compile-time exclusion, `#if __ANDROID__` in a `.skia.cs` file, or logic dumped into a cross-platform file that belongs in a suffixed partial.
-- **Layering & DependencyObject:** On Android/iOS `DependencyObject` is an *interface* (UIElement must inherit native views) implemented by `DependencyObjectGenerator` — don't propose inheriting from it. Logic that belongs in the shared layer should not be duplicated per-platform, and vice-versa.
+- **Layering & DependencyObject:** `DependencyObject` is a plain class on every target (WinUI parity) — inheriting from it directly is correct; flag any advice built on the pre-7.0 interface or its removed `DependencyObjectGenerator`. Logic that belongs in the shared layer should not be duplicated per-platform, and vice-versa.
```
morning4coffe-dev, line 46: [P2] The companion `.claude/rules/source-generators.md` still contradicts this corrected statement by describing a surviving DependencyObject generator that only emits registrations and `[Bindable]`. `DependencyObjectGenerator.cs` is absent; the active `DependencyPropertyGenerator` and separate `DependencyObjectAvailabilityGenerator` have different responsibilities. Please update that rule to describe the generators that actually remain.

specs/050-breaking-changes-rollup/spec.md
```diff
@@ -316,21 +317,21 @@ _Danger 3-4. Heavier multi-file changes: remove the legacy templated-parent mech
...
+- [x] **BC54** — `FlyoutBase.DataContext` -> non-public  `d4·L` · #12491 · PR #23547
+  - Resolved by BC58. `FlyoutBase` no longer has a `DataContext`; instead of the planned internal `FlyoutBase`->`Popup` forwarding, the placement target's `DataContext` is copied onto the presenter when the flyout opens and cleared when it closes, as in WinUI. **Folded into the BC58 spec.**
+  - Files: `src/Uno.UI/UI/Xaml/Controls/Flyout/FlyoutBase.cs`
+- [x] **BC26** — `DependencyObject` becomes a class  `d4·L` · #17099 · PR #23537, #23702 · **[impact spec](bc26-dependencyobject-as-class.md)**
+  - Landed as a class on every target; `DependencyObjectGenerator` was deleted outright rather than trimmed, and phase 2 (#23702) went on to drop `DependencyObjectStore` and fold its storage onto `DependencyObject`.
+  - Files: `src/Uno.UI/UI/Xaml/DependencyObject.cs`, `src/Uno.UI/UI/Xaml/DependencyObject.Store.cs`, `src/Uno.UI/UI/Xaml/DependencyObject.Binder.cs`, `src/Uno.UI/UI/Xaml/UIElement.skia.cs`
```
morning4coffe-dev: [P2] `src/Uno.UI/UI/Xaml/UIElement.skia.cs` does not exist at this HEAD, so this implemented-file list points readers to removed source. `UIElement` now derives `DependencyObject` in `src/Uno.UI/UI/Xaml/UIElement.cs`; please replace or remove the stale path.

morning4coffe-dev (changes requested):
The inheritance-context and `__UNO_SKIA__` blockers are fixed, and BC26/BC58 now distinguish historical plans from implemented outcomes. Two source-truth blockers remain: contributor guidance still teaches the removed native Android/iOS/DOM UI renderers and UWP-era inheritance, and `.claude/rules/source-generators.md` still describes the deleted `DependencyObjectGenerator` as surviving. The rollup also references nonexistent `UIElement.skia.cs`.

AGENTS.md
MartinZikmund: Fixed in 519a81c3a5. All four guidance surfaces now describe the Uno 7 Skia-only UI model.

**`AGENTS.md`**
- "Rendering Engines" now names Skia as the only renderer and points to native-view *embedding* for hosting native content.
- The scope section is now "Development scope: Skia-only UI". There are no native UI renderers to maintain; the non-UI WinRT APIs keep their per-platform implementations.
- The native Android/iOS/DOM inheritance table is replaced by `UIElement : DependencyObject` on every target.
- A note under the file-suffix table says the platform suffixes only compile in per-platform projects (the WinRT layer and platform hosts/add-ins). `Uno.UI` builds once for Skia, so a `.Android.cs`/`.UIKit.cs`/`.wasm.cs` partial there is never compiled. The partial-class example follows suit.

**`.claude/rules/platform-targeting.md`**
- The scope section is rewritten to match.
- Rules 1-2 now scope suffixes to per-platform projects, and state that only `.skia.cs`/`.crossruntime.cs` compile under `Uno.UI`.
- The "Skia vs native Android control" trap is reframed around the WinRT layer's Skia vs Android variants.

**`.claude/agents/architect.md`:** it now flags changes that reintroduce a native UI path, instead of asking reviewers to protect native targets.

**`doc/articles/migrating-guidance.md`:** the Xamarin.iOS nested-class section and the native-property name-collision section are removed. A 7.0 note says both no longer apply and links to native-view embedding.

.claude/agents/architect.md
MartinZikmund: Fixed in 1beb2d66a3. `source-generators.md` now says there is no DependencyObject generator: nothing is emitted per `DependencyObject` type, and `DependencyObjectGenerator`/`DependencyObjectStore`/`IDependencyObjectStoreProvider` are all gone. It then lists the generators that actually remain:
- `DependencyPropertyGenerator` (`Uno.UI.SourceGenerators.Internal/DependencyObject/`, incremental) expands `[GeneratedDependencyProperty]`, including attached properties.
- `DependencyObjectAvailabilityGenerator` (`BindableTypeProviders/`) emits the `DependencyObject` availability linker hints and the substitution file used by XAML resource trimming.
- `BindableTypeProvidersSourceGenerator` emits `BindableMetadataProvider`.

It also notes that all three detect `DependencyObject` by walking the base-type chain. The project map no longer calls `Uno.UI.SourceGenerators.Internal` "utilities".

specs/050-breaking-changes-rollup/spec.md
MartinZikmund: Fixed in daed0bd596. The BC26 file list now points at `src/Uno.UI/UI/Xaml/UIElement.cs`, which declares `UIElement : DependencyObject`. The `UIElement.skia.cs` mention in the BC26 impact spec is under "Original state (before the change)", so it stays as history.

MartinZikmund:
@morning4coffe-dev Thanks for the review. All three items are addressed and pushed; details are in the resolved threads:

- **Native renderer model in contributor guidance** (519a81c3a5): `AGENTS.md`, `.claude/rules/platform-targeting.md`, `.claude/agents/architect.md` and `doc/articles/migrating-guidance.md` now describe the Skia-only UI. Platform file suffixes are scoped to per-platform projects, and the native base-class table and the Xamarin-era native-view sections are gone.
- **`.claude/rules/source-generators.md`** (1beb2d66a3): it now says there is no DependencyObject generator and lists the generators that remain (`DependencyPropertyGenerator`, `DependencyObjectAvailabilityGenerator`, `BindableTypeProvidersSourceGenerator`).
- **Stale `UIElement.skia.cs` in the BC26 rollup entry** (daed0bd596): replaced with `UIElement.cs`.

.claude/agents/architect.md
```diff
@@ -39,11 +39,11 @@ Flag tech debt being introduced. Call out bad patterns. Identify scalability con
...
-- **Layering & DependencyObject:** On Android/iOS `DependencyObject` is an *interface* (UIElement must inherit native views) implemented by `DependencyObjectGenerator` — don't propose inheriting from it. Logic that belongs in the shared layer should not be duplicated per-platform, and vice-versa.
+- **Skia-only UI (AGENTS.md → "Development scope", `.claude/rules/platform-targeting.md`):** `Uno.UI` compiles once, for Skia, and that assembly runs on every platform; the native Android View, UIKit and WASM DOM renderers were removed in 7.0. Flag changes that reintroduce a native UI path: native-view inheritance, DOM rendering, or `.Android.cs`/`.UIKit.cs`/`.wasm.cs` partials under `Uno.UI`, which are never compiled there. This applies to the UI layer only — `Uno.WinRT`/`Uno.Foundation` non-UI WinRT APIs keep per-platform implementations and are still actively enhanced.
```
morning4coffe-dev: [P2] This path/suffix rule is too broad: several `.wasm.cs` files physically under `src/Uno.UI` are live sources explicitly linked and compiled by `Uno.UI.Runtime.Skia.WebAssembly.Browser.csproj` (including the NativeWebView partials). Please say that the `Uno.UI` project excludes these suffixes by default while platform runtime projects may link them, and review the owning MSBuild project rather than physical path alone. The matching statements in `AGENTS.md` and `platform-targeting.md` need the same qualification.

.claude/rules/source-generators.md
```diff
@@ -9,9 +9,13 @@ paths:
...
+- **No DependencyObject generator.** `DependencyObject` is a concrete base **class** that carries the property store and binder directly, so nothing is emitted per `DependencyObject` type. `DependencyObjectGenerator`, `DependencyObjectStore` and `IDependencyObjectStoreProvider` are all gone. The generators that remain around dependency objects have narrower jobs:
+  - `DependencyPropertyGenerator` (`Uno.UI.SourceGenerators.Internal/DependencyObject/`, incremental) expands `[GeneratedDependencyProperty]` declarations into DP registrations and backing fields, including attached properties.
+  - `DependencyObjectAvailabilityGenerator` (`Uno.UI.SourceGenerators/BindableTypeProviders/`) walks referenced assemblies for `DependencyObject`-derived types (and `[AdditionalLinkerHint]` targets) and emits availability linker hints plus the linker substitution file used by XAML resource trimming.
+  - `BindableTypeProvidersSourceGenerator` (same folder) emits the `BindableMetadataProvider` for binding metadata.
+  - All of them detect `DependencyObject` by walking the base-type chain, not by interface.
```
morning4coffe-dev: [P2] `DependencyPropertyGenerator` does not rely exclusively on the base-type chain. It intentionally checks `AllInterfaces` as a compatibility fallback for projects referencing released Uno packages where `DependencyObject` is still an interface. Please document base-class detection as the current primary path while preserving this implemented fallback.

morning4coffe-dev (changes requested):
The three previous blockers are fixed, but the replacement guidance introduces two source-truth errors: live `.wasm.cs` files linked into the WebAssembly runtime are described as never compiled, and the generator guidance omits the intentional interface-detection fallback used for compatibility with released Uno packages. These need correction before approval.

.claude/agents/architect.md
MartinZikmund: You're right, I overstated it. `Uno.UI.Runtime.Skia.WebAssembly.Browser.csproj` links four live `.wasm.cs` files from under `src/Uno.UI`: `InputReturnTypeExtensions`, `InputScopeExtensions`, `NativeWebView` and `NativeWebView.Interop`.

Fixed in 0254a40542, with the same qualification in all three places:
- **`AGENTS.md`:** the note under the suffix table now says the `Uno.UI` project excludes `.Android.cs`/`.UIKit.cs`/`.wasm.cs` by default, that platform runtime projects can still link such files (naming the WebAssembly runtime's `NativeWebView` and input-scope partials), and that a file should be judged by the MSBuild project that compiles it, not by its path.
- **`architect.md`:** it flags partials that expect the `Uno.UI` project to compile them, and tells reviewers to check the owning project's `Compile Include`/`Link`.
- **`platform-targeting.md`:** rule 1 now says "the `Uno.UI` project itself" compiles only `.skia.cs`/`.crossruntime.cs`, with the same linking caveat. The `__ANDROID__` trap is scoped to files the `Uno.UI` project compiles.

.claude/rules/source-generators.md
MartinZikmund: Good catch, thanks. Fixed in 211101eb73. The rule now calls base-type walking the *primary* detection path, and says `DependencyPropertyGenerator` deliberately keeps an `AllInterfaces` match as a compatibility fallback for previously released Uno packages where `DependencyObject` is still an interface. It also says to keep that fallback until a stable release ships `DependencyObject` as a class. The two `BindableTypeProviders` generators really are base-type-only, so the sentence stays scoped to `DependencyPropertyGenerator`. While there, I added that `DependencyObjectAvailabilityGenerator` only runs when `UnoXamlResourcesTrimming` is on or inside the Uno solution.

MartinZikmund:
@morning4coffe-dev Both are fixed, and the threads are resolved:
- **Suffix scoping** (0254a40542): `AGENTS.md`, `architect.md` and `platform-targeting.md` now say the `Uno.UI` *project* excludes `.Android.cs`/`.UIKit.cs`/`.wasm.cs` by default. They also note that platform runtime projects can link them (e.g. the WebAssembly runtime's `NativeWebView` partials) and tell readers to judge a file by its owning MSBuild project.
- **`DependencyPropertyGenerator` detection** (211101eb73): base-type walking is documented as the primary path, and the `AllInterfaces` fallback for released packages is kept.

morning4coffe-dev (approved):
Verified the tagged follow-up at `211101eb73c8`. Platform-suffix guidance now correctly follows the owning MSBuild project and acknowledges linked `.wasm.cs` sources, while the generator documentation preserves the released-package `AllInterfaces` compatibility fallback. The previous source-truth findings are resolved; approving this head.

PR #24381, opened 2026-09-03
feat(x11): Linux AT-SPI accessibility backend for the Skia.X11 head

### LINKED ISSUE #24380: Linux / X11 AT-SPI accessibility backend for the Skia.X11 head

A **Linux / X11 AT-SPI accessibility backend** for the `Uno.UI.Runtime.Skia.X11` head.

6.6 brought Skia accessibility to Windows, macOS and WebAssembly. On Linux the Skia head renders to X11 but publishes **no** AT-SPI2 tree, so a Uno app is invisible to Orca and to AT-SPI-based automation. This adds an `Accessibility/` provider to the X11 head that publishes the automation-peer tree over the accessibility D-Bus (`org.a11y.atspi`), mirroring how the macOS/WASM heads consume the shared `SkiaAccessibilityBase` / `AccessibilityRouter` abstraction.

Scope of this work:
- Read path + registration (`org.a11y.atspi.Socket.Embed`), serving `Accessible` + `Component`.
- Full interface set: `Action`, `Value`, `Text`, `EditableText`, `Selection`.
- Live events via the base class hooks (`state-changed`, `property-change`, `selection-changed`, `children-changed`) and focus tracking.
- Write path driving the real control through its automation provider (invoke / toggle / set-value / set-text / select).

Discussed in https://github.com/unoplatform/uno/discussions/24256 (maintainers confirmed it is in scope).

### Why is this needed? 🤔

Any toolkit that paints its own pixels (Skia) publishes no control-level accessibility unless its peer tree is bridged to the platform API. Without this, Uno Skia apps on Linux are unusable with the Orca screen reader and invisible to AT-SPI-based tooling. It closes the last desktop gap after the 6.6 Windows/macOS/WASM work.

### Renderer 🎨

- [x] Skia

### For which platforms 📱💻🖥️

Skia Desktop (Linux / X11)

The implementation was verified headless in a container (Xvfb + dbus + at-spi2-core): an external AT-SPI client reads the tree and drives controls over the bus (EditableText write + read-back through Text, Value set + read-back, Action `do_action`, checkbox toggle), with live state/property signals observed on the bus. It reuses the existing `DBus/` layer's `Tmds.DBus.Protocol` dependency.

egarim:
**GitHub Issue:** closes #24380

Discussed in https://github.com/unoplatform/uno/discussions/24256 — maintainers confirmed a Linux AT-SPI backend is in scope and suggested mirroring the macOS seam.

## PR Type:

✨ Feature

## What changed? 🚀

Adds a Linux / X11 **AT-SPI2 accessibility backend** to `Uno.UI.Runtime.Skia.X11`. Previously the X11 head rendered to the screen but published no accessibility tree, so Uno apps were invisible to Orca and to AT-SPI-based automation. This brings the X11 head to behavioral parity with the macOS and WASM Skia heads.

New `src/Uno.UI.Runtime.Skia.X11/Accessibility/`:
- **`AtspiServer`** — discovers the a11y bus (`org.a11y.Bus.GetAddress`), registers via `org.a11y.atspi.Socket.Embed`, and serves each node over D-Bus (`Tmds.DBus.Protocol`, already referenced by the head's `DBus/` layer). Serves `Accessible`, `Component`, `Application` (root), `Action`, `Value`, `Text`, `EditableText`, `Selection`, plus `Properties`/`Introspectable`.
- **`X11Accessibility : SkiaAccessibilityBase`** — builds an `AtspiNode` tree from the automation peers (skipping `AccessibilityView.Raw`), enriched from `AriaMapper.GetAriaAttributes` + the automation providers; implements the base `Update*` / `SetNativeFocus` / child hooks to emit AT-SPI `state-changed` / `property-change` / `selection-changed` / `children-changed` signals; and implements the write path (`IWriteTarget`) that drives the real control through its provider on the UI thread.
- **`AtspiRoleMap`** — `AutomationControlType` → numeric `AtspiRole` id (ids match `at-spi2-core/atspi-constants.h`; unmapped types fall back to `panel`).

Host wiring mirrors the macOS head: `X11XamlRootHost` implements `IAccessibilityOwner`, `AccessibilityRouter` is initialized in `X11ApplicationHost` and activated on `FocusIn`.

**Verification.** Built from source and run headless in a container (Xvfb + dbus + at-spi2-core). An external AT-SPI client:
- reads the published tree (roles, states, bounding boxes);
- writes over the bus and reads back — `EditableText.SetTextContents` → `Text.GetText`, `Value.SetCurrentValue` → read back, checkbox toggle via `Action.DoAction` flips the `checked` state;
- observes live `state-changed` / `property-change` signals during those writes.

## PR Checklist ✅

- [x] 🧪 Added [Runtime tests, UI tests, or a manual test sample] — verified headless against at-spi2-core with an external AT-SPI client (read + write + live events); happy to add runtime tests under `Windows_UI_Xaml_Automation` if the team prefers a specific harness for the Linux head.
- [ ] 📚 Docs have been added/updated
- [x] 🖼️ Validated PR `Screenshots Compare Test Run` results — n/a (no visual change).
- [x] ❗ Contains **NO** breaking changes — additive; no other head or TFM is touched.

## Notes for reviewers

- Suggested split if you'd rather review in chunks: (1) read path + registration, (2) live events + write path. Happy to break it up.
- Coordinates are currently window-relative (`GetExtents` returns client-space); wiring the X11 window origin for true screen coordinates is a small follow-up, noted in code.
- `Tmds.DBus.Protocol` is pinned at the head's existing `0.92.0`; a subtlety worth knowing is that `MessageWriter` is a `ref struct` (writing the message body through a by-value helper silently drops it), and 0.92 tears the connection down on an unhandled incoming call — both are handled here.

### DIFF
--- a/src/Uno.UI.Runtime.Skia.X11/Accessibility/AtspiServer.cs
+++ b/src/Uno.UI.Runtime.Skia.X11/Accessibility/AtspiServer.cs
@@ -240,7 +240,7 @@ internal sealed class AtspiServer
 		}
 	}
 
-	public void EmitChildrenChanged(AtspiNode parent, bool added, int index, AtspiNode? child)
+	public void EmitChildrenChanged(AtspiNode parent, bool added, int index)
 	{
 		try
 		{
@@ -779,6 +779,8 @@ internal sealed class AtspiServer
 					writer.WriteVariantString(node.Name);
 					break;
 				case (AtspiDbus.AccessibleInterface, AtspiDbus.DescriptionProperty):
+					writer.WriteVariantString(node.Description ?? AtspiDbus.EmptyString);
+					break;
 				case (AtspiDbus.AccessibleInterface, AtspiDbus.AccessibleIdProperty):
 					writer.WriteVariantString(AtspiDbus.EmptyString);
 					break;
@@ -911,8 +913,11 @@ internal sealed class AtspiServer
 				SetState(AtspiDbus.SensitiveState);
 			}
 
-			SetState(AtspiDbus.ShowingState);
-			SetState(AtspiDbus.VisibleState);
+			if (!node.Offscreen)
+			{
+				SetState(AtspiDbus.ShowingState);
+				SetState(AtspiDbus.VisibleState);
+			}
 
 			if (node.Focusable)
 			{
@@ -952,10 +957,25 @@ internal sealed class AtspiServer
 				SetState(AtspiDbus.FocusedState);
 			}
 
+			// The AT-SPI state set is two 32-bit words; states with index >= 32 live in
+			// the second word.
+			uint state1 = 0;
+			void SetState1(int bit) => state1 |= 1u << (bit - 32);
+
+			if (node.Required)
+			{
+				SetState1(AtspiDbus.RequiredState);
+			}
+
+			if (node.ReadOnly)
+			{
+				SetState1(AtspiDbus.ReadOnlyState);
+			}
+
 			using var writer = context.CreateReplyWriter(AtspiDbus.UInt32ArraySignature);
 			var array = writer.WriteArrayStart(DBusType.UInt32);
 			writer.WriteUInt32(state0);
-			writer.WriteUInt32(0);
+			writer.WriteUInt32(state1);
 			writer.WriteArrayEnd(array);
 			context.Reply(writer.CreateMessage());
 		}
@@ -1062,14 +1082,6 @@ internal sealed class AtspiServer
 				writer.WriteString(key);
 				writer.WriteString(value);
 			}
-			writer.WriteDictionaryEnd(dictionary);
-			context.Reply(writer.CreateMessage());
-		}
-
-		private static void ReplyEmptyDictionary(MethodContext context, string signature)
-		{
-			using var writer = context.CreateReplyWriter(signature);
-			var dictionary = writer.WriteDictionaryStart();
 			writer.WriteDictionaryEnd(dictionary);
 			context.Reply(writer.CreateMessage());
 		}
@@ -1221,5 +1233,7 @@ internal sealed class AtspiServer
 		public const int FocusedState = 12;
 		public const int SelectableState = 22;
 		public const int SelectedState = 23;
+		public const int RequiredState = 33;
+		public const int ReadOnlyState = 43;
 	}
 }
--- a/src/Uno.UI.Runtime.Skia.X11/Accessibility/AtspiServer.cs
+++ b/src/Uno.UI.Runtime.Skia.X11/Accessibility/AtspiServer.cs
@@ -1,0 +1,680 @@
+#nullable enable
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Tmds.DBus.Protocol;
+using Uno.Foundation.Logging;
+
+namespace Uno.WinUI.Runtime.Skia.X11;
+
+/// <summary>
+/// Serves an AT-SPI accessibility tree over the desktop accessibility D-Bus.
+/// </summary>
+internal sealed class AtspiServer
+{
+	private readonly DBusConnection _connection;
+	private readonly string _uniqueName;
+	private AtspiReference _rootParent;
+	private IReadOnlyDictionary<string, AtspiNode> _nodesByPath = new Dictionary<string, AtspiNode>();
+
+	private AtspiServer(DBusConnection connection, string uniqueName)
+	{
+		_connection = connection;
+		_uniqueName = uniqueName;
+		_rootParent = AtspiReference.Null;
+	}
+
+	public static async Task<AtspiServer?> TryStartAsync(string applicationName)
+	{
+		DBusConnection? connection = null;
+
+		try
+		{
+			var sessionBusAddress = DBusAddress.Session;
+			if (sessionBusAddress is null)
+			{
+				if (typeof(AtspiServer).Log().IsEnabled(LogLevel.Debug))
+				{
+					typeof(AtspiServer).Log().Debug("No D-Bus session bus available; AT-SPI bridge will be disabled.");
+				}
+
+				return null;
+			}
+
+			var a11yBusAddress = await GetA11yBusAddressAsync(sessionBusAddress);
+			if (string.IsNullOrEmpty(a11yBusAddress))
+			{
+				if (typeof(AtspiServer).Log().IsEnabled(LogLevel.Debug))
+				{
+					typeof(AtspiServer).Log().Debug("No AT-SPI accessibility bus address available; AT-SPI bridge will be disabled.");
+				}
+
+				return null;
+			}
+
+			connection = new DBusConnection(a11yBusAddress);
+			await connection.ConnectAsync();
+
+			var uniqueName = connection.UniqueName ?? "";
+			var server = new AtspiServer(connection, uniqueName);
+			server.SetRoot(new AtspiNode
+			{
+				Path = AtspiDbus.RootPath,
+				Name = applicationName,
+				Role = AtspiDbus.ApplicationRole,
+				RoleName = AtspiDbus.ApplicationRoleName
+			});
+			connection.AddMethodHandler(new NodeHandler(server));
+			server._rootParent = await server.EmbedAsync();
+
+			if (server.Log().IsEnabled(LogLevel.Debug))
+			{
+				server.Log().Debug($"AT-SPI bridge started for '{applicationName}' on {uniqueName}.");
+			}
+
+			return server;
+		}
+		catch (Exception ex)
+		{
+			connection?.Dispose();
+
+			if (typeof(AtspiServer).Log().IsEnabled(LogLevel.Debug))
+			{
+				typeof(AtspiServer).Log().Debug($"Unable to start AT-SPI bridge: {ex.Message}");
+			}
+
+			return null;
+		}
+	}
+
+	public void SetRoot(AtspiNode root)
+	{
+		if (root.Path != AtspiDbus.RootPath)
+		{
+			throw new ArgumentException($"The AT-SPI root node path must be '{AtspiDbus.RootPath}'.", nameof(root));
+		}
+
+		var nodesByPath = new Dictionary<string, AtspiNode>(StringComparer.Ordinal);
+		AddNode(root, nodesByPath);
+		Volatile.Write(ref _nodesByPath, nodesByPath);
+	}
+
+	public ValueTask StopAsync()
+	{
+		_connection.Dispose();
+		return default;
+	}
+
+	private static async Task<string?> GetA11yBusAddressAsync(string sessionBusAddress)
+	{
+		using var sessionConnection = new DBusConnection(sessionBusAddress);
+		await sessionConnection.ConnectAsync();
+
+		MessageBuffer request;
+		using (var writer = sessionConnection.GetMessageWriter())
+		{
+			writer.WriteMethodCallHeader(
+				AtspiDbus.BusService,
+				AtspiDbus.BusPath,
+				AtspiDbus.BusInterface,
+				AtspiDbus.GetAddressMethod,
+				AtspiDbus.EmptySignature,
+				MessageFlags.None);
+			request = writer.CreateMessage();
+		}
+
+		return await sessionConnection.CallMethodAsync<string>(
+			request,
+			static (message, _) => message.GetBodyReader().ReadString(),
+			null);
+	}
+
+	private async Task<AtspiReference> EmbedAsync()
+	{
+		MessageBuffer request;
+		using (var writer = _connection.GetMessageWriter())
+		{
+			writer.WriteMethodCallHeader(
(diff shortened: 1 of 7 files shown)


egarim:
Thanks for the review — addressed in afe71854ac (and the coordinate follow-up f4e3700e59). Point by point:

**🟡 ComboBox item enumeration (`X11Accessibility.cs:412`)** — fixed. Items are now enumerated only while the dropdown is open; when closed the combo just exposes its selected value as text (via `SelectionBoxItem`, matching the macOS head) and stays `expandable`. This avoids materializing a large/virtualized `ItemsSource` on every rebuild. A rebuild is coalesced-triggered when the dropdown opens.

**🔵 Startup/disposal race (`X11Accessibility.cs:106`)** — fixed. `_server` is assigned on the D-Bus connect continuation (not the UI thread) while `DisposeCore` runs on the UI thread; they now share a lock, so a server that finishes starting as the window closes is stopped rather than leaked. The `IsDisposed` check and the assignment are inside the same lock that `DisposeCore` takes to read/null `_server`.

**🔵 Retained `UIElement`s on dispose (`X11Accessibility.cs:821`)** — fixed. `DisposeCore` now also resets the published `_elementsSnapshot` (the reader-thread lookup dictionary), so the visual tree isn't retained after the window closes.

**🟡 `Value` interface "advertised but not dispatched" (`AtspiServer.cs:463`)** — this one is a false positive: the AT-SPI `org.a11y.atspi.Value` interface is entirely *property*-based (`CurrentValue`, `MinimumValue`, `MaximumValue`, `MinimumIncrement`), served through `org.freedesktop.DBus.Properties` `Get`/`Set` — it has no D-Bus *methods*, so there is nothing to route in the method-dispatch switch. Verified against a real client: `Atspi.Value.set_current_value` writes and reads back correctly over the bus (a slider driven `0 → 2414`, read back `2414`).

**Tests (`X11Accessibility.cs:72`)** — fair point. The whole contract is currently verified headless (Xvfb + dbus + at-spi2-core) with an external AT-SPI client exercising read + write + live events; `Uno.UI.RuntimeTests` doesn't reference the X11 head today, so there's no obvious home for an in-harness test. Happy to (a) add a deterministic `AtspiRoleMap` unit test, and/or (b) contribute the container-based AT-SPI smoke as a CI job — whichever fits how you'd like Linux-only a11y covered. Guidance welcome.

All changes re-verified against the headless AT-SPI client after the fixes (EditableText write + read-back, Value set + read-back, checkbox toggle via Action, live state/property signals; extents shift correctly when the window moves).

src/Uno.UI.Runtime.Skia.X11/Accessibility/AtspiServer.cs
```diff
@@ -0,0 +1,1239 @@
...
+			{
+				case AtspiDbus.GetTextMethod:
+				{
+					var reader = context.Request.GetBodyReader();
+					var start = reader.ReadInt32();
+					var end = reader.ReadInt32();
+					var text = node.Text;
+					start = Math.Max(0, Math.Min(start, text.Length));
+					end = end < 0 ? text.Length : Math.Max(start, Math.Min(end, text.Length));
+					ReplyString(context, AtspiDbus.StringSignature, text.Substring(start, end - start));
```
morning4coffe-dev, line 593: [P1] Translate AT-SPI character offsets instead of using them as UTF-16 indices. In a real private Linux D-Bus run of this exact server, `A\U0001F600B` reports CharacterCount=4 and GetText(1,2) returns U+FFFD. DeleteText(1,2) then reports success while passing `0041 DE00 0042` (a lone low surrogate) to IWriteTarget.SetText, rather than `AB`. CharacterCount and the read/insert/delete paths need a consistent Unicode-character-to-UTF-16 mapping; partial operations must not split surrogate pairs.

src/Uno.UI.Runtime.Skia.X11/Accessibility/AtspiServer.cs
```diff
@@ -0,0 +1,1239 @@
...
+		public const string EmptySignature = "";
+		public const string ReferenceSignature = "(so)";
+		public const string ReferenceArraySignature = "a(so)";
+		public const string StringSignature = "s";
+		public const string UInt32Signature = "u";
+		public const string Int32Signature = "i";
+		public const string BoolSignature = "b";
+		public const string VariantSignature = "v";
+		public const string ExtentsSignature = "(iiii)";
+		public const string PositionSignature = "(ii)";
```
morning4coffe-dev, line 1208: [P2] GetPosition and GetSize must reply with `ii`, not `(ii)`. The authoritative Component.xml declares separate x/y and width/height out arguments; only GetExtents declares a struct. Actual replies from this server advertise `(ii)` for both methods, so clients checking the contract receive the wrong type. Use separate integer outputs without a struct wrapper. Also honor the coordinate-type input in position/extents/hit-testing methods: the probe's screen and window requests currently return identical absolute coordinates.

morning4coffe-dev (changes requested):
Reviewed the seven-file backend at `afe71854ac40` and exercised the unchanged AtspiServer/AtspiNode source over a real private Linux D-Bus session, against the pinned at-spi2-core interface definitions. Two protocol defects need correction:

**P1: text operations split supplementary Unicode characters.** For `A\U0001F600B`, CharacterCount returns 4 rather than 3, GetText(1, 2) returns U+FFFD rather than the supplementary character, and DeleteText(1, 2) reports success while delivering UTF-16 units `0041 DE00 0042` to the write target. That is a lone low surrogate, not `AB`. Map the protocol's character offsets to UTF-16 indices consistently for count/read/edit operations; whole-string retrieval and the ASCII control both succeeded, so this is specifically the offset handling rather than a general transport problem.

**P2: GetPosition/GetSize have the wrong D-Bus output signature.** Both real replies advertise `(ii)`, but Component.xml defines two separate integer outputs, `ii`. GetExtents is the method that returns a struct. Also, screen-relative and window-relative GetPosition requests return the same coordinates because the coordinate-type argument is ignored; the other coordinate-taking Component methods need consistent handling too.

The protocol run also confirms [the existing GetAll finding](https://github.com/unoplatform/uno/pull/24381#discussion_r3924973248): a populated Accessible node returns an empty property dictionary. I did not duplicate that inline. Conversely, [the missing Value-dispatch objection](https://github.com/unoplatform/uno/pull/24381#discussion_r3923484896) is not valid for this protocol: Value.xml defines properties, not methods, and the actual Properties.Get/Set path changed CurrentValue from 5 to 12 in the control case. Existing reviewers' thread states were preserved.

Reference: [Component.xml](https://github.com/GNOME/at-spi2-core/blob/9d2b47eae6d19c2b8b905be960567ba5f17a77ea/xml/Component.xml), [Text.xml](https://github.com/GNOME/at-spi2-core/blob/9d2b47eae6d19c2b8b905be960567ba5f17a77ea/xml/Text.xml) and [Value.xml](https://github.com/GNOME/at-spi2-core/blob/9d2b47eae6d19c2b8b905be960567ba5f17a77ea/xml/Value.xml).

Validation scope: actual Linux D-Bus transport and exact server dispatch/serialization, with a node-only synchronous write target and a logging adapter. The dispatcher was attached directly; registry discovery, actual X11 UI peer construction/provider mutation and Orca were not exercised. This is not full cross-head behavioral-parity or end-to-end screen-reader approval. The isolated container/image were removed and the initially stopped Docker engine was restored to stopped.

src/Uno.UI.Runtime.Skia.X11/Accessibility/AtspiServer.cs
egarim: Confirmed — the Text/EditableText paths (and the `CharacterCount` property) were treating AT-SPI character offsets as UTF-16 indices exactly as you measured. Fixed in 5145645277: offsets are now mapped code-point → UTF-16 (`Utf16Index`/`CharacterCount` helpers), clamped at the ends, and a surrogate pair can never be split — `InsertText`'s `length` argument is treated as characters as well.

Against your probe string `A\U0001F600B`: `CharacterCount` = 3, `GetText(1,2)` returns the emoji intact, `DeleteText(1,2)` yields `AB`, and an insert at character offset 2 lands after the pair.

Honesty note on verification: the project compiles and the mapping logic is covered by a standalone check reproducing exactly the cases above; I have not re-run a live Linux D-Bus session for this commit.

src/Uno.UI.Runtime.Skia.X11/Accessibility/AtspiServer.cs
egarim: Half confirmed, half by-design:

**Wire signature — confirmed and fixed** in 5145645277. `GetPosition`/`GetSize` now reply `ii` (two separate out arguments per Component.xml); only `GetExtents` keeps its declared `(iiii)` struct. The old `(ii)` struct reply advertised the wrong type in the signature header even where the payload bytes happened to coincide.

**Coordinate type — intentional in this PR.** All coordinates are absolute screen coordinates via the window origin (f4e3700e59), and the coordinate-type argument is deliberately not branched on yet — that matches the PR's documented coordinate model rather than being an oversight. Happy to add proper `ATSPI_COORD_TYPE_WINDOW` translation (origin subtraction) as a follow-up commit or a separate PR if you'd prefer strict conformance now — it's a small, contained change on top of the same origin data.

Verification note: compiles locally; no live Linux D-Bus re-run for this commit.

egarim:
Update on the outstanding items — three commits pushed:

**`d598435bf8` — merge conflict resolved.** Merged `master` (1547 commits) into the branch: `X11XamlRootHost.cs` keeps master's `IX11Renderer` and `BackgroundColor` alongside this branch's accessibility members, and the new `SkiaAccessibilityBase.UpdateRoleDescription` abstract member is implemented as a no-op, matching `Win32Accessibility`. The PR is mergeable again.

**`971253b7de` — Introspect no longer lies.** `Introspect` hardcoded `<node/>` for every node, so dynamic-discovery clients saw an empty interface list despite the object serving Accessible/Component/Action/Value/Text/EditableText/Selection. The interface set now comes from a single `InterfaceNames` helper shared with `GetInterfaces`, so the two can never disagree (introspection XML additionally lists Properties + Introspectable).

**`289f1c3802` — AT-SPI object paths are now stable across tree rebuilds.** Previously every rebuild reset the path counter and reassigned paths in DFS order, so a client holding a cached object reference could silently resolve to a *different control* after any tree change — including routing a write to the wrong element. Paths are now allocated once per element identity (`Visual.Handle`) in maps that persist across rebuilds; data-only combo items (no container element) key on the owning combo's handle + item index. Maps are cleared only on dispose.

**Two Copilot findings from the 2026-09-23 review are false positives** — checked against the authoritative at-spi2-core XML at the pinned commit `9d2b47eae6d19c2b8b905be960567ba5f17a77ea`:
- *"Encode GetExtents reply as four top-level integers"* — Component.xml declares GetExtents' out-arg as a **`(iiii)` struct**; only GetPosition/GetSize are flat (`ii`), which this server already does. Current code is correct.
- *"Return empty replies for void EditableText methods"* — EditableText.xml declares **boolean** out-args on SetTextContents/InsertText/CutText/DeleteText/PasteText; only CopyText is void, and the server already dispatches CopyText with an empty reply. Current code is correct.

**Deferred (acknowledged, out of scope for this round):** the `ChildrenChanged` signal variant carries a bare int instead of an `(so)` child reference (a known Tmds.DBus.Protocol 0.92 limitation, noted in the code), and one AT-SPI application root shared across windows instead of per-`X11XamlRootHost` servers (architectural; affects multi-window apps) — happy to take that as a follow-up PR.

**Validation evidence** (labels per the repo's protocol): *Compile* — `Uno.UI.Runtime.Skia.X11.csproj` and the full `Uno.UI-Skia-only.slnf` both build clean after the merge + fixes. *Runtime* — not executed here (no Linux X11/D-Bus session on this machine); to validate live: `dbus-run-session` + `at-spi2-core` and introspect any node (`busctl --user introspect <bus> /org/a11y/atspi/accessible/1`), and verify a cached accessible still resolves to the same control after opening a combo dropdown (which triggers a rebuild).

egarim:
@ramezgerges this should be ready for another look — all review points to date are addressed:

- `morning4coffe-dev`'s Sep 14 review: character-offset mapping and the `GetPosition`/`GetSize` wire signature, fixed in `5145645277` (details in the two review threads).
- Merge conflict with `master`: resolved in `d598435bf8` (incl. the new `UpdateRoleDescription` abstract member, no-op like Win32).
- `Introspect` now reports each node's real interface set, single-sourced with `GetInterfaces` (`971253b7de`).
- AT-SPI object paths are stable across tree rebuilds — cached client references can no longer resolve to a different control (`289f1c3802`).
- The two Copilot findings from Sep 23 are false positives per the pinned at-spi2-core XML — spec citations in [the summary comment](https://github.com/unoplatform/uno/pull/24381#issuecomment-5814857051).

Known deferrals (happy to take as follow-ups): `ChildrenChanged`'s `(so)` child reference (Tmds.DBus.Protocol 0.92 limitation) and a process-scoped application root for multi-window apps.

PR #24382, opened 2026-09-03
feat: Add MarkupHelper.SetParent and CreateFrameworkTemplate for markup-library codegen

MartinZikmund:
**GitHub Issue:** relates to unoplatform/uno-private#2075, unblocks [unoplatform/uno.csharpmarkup#906](https://github.com/unoplatform/uno.csharpmarkup/pull/906)

## PR Type:

✨ Feature

## What changed? 🚀

Adds two small, purely-additive public members to `MarkupHelper`, matching its existing `[EditorBrowsable(Never)]` "bridge for markup-library codegen" convention (`CreateDataTemplate`/`CreateControlTemplate`/`CreateItemsPanelTemplate`, `OnTemplateMemberCreated`):

- **`MarkupHelper.SetParent(DependencyObject target, object parent)`** — a public way to parent a `DependencyObject` created outside its owner's normal content-property assignment path (e.g. an item added by generated code to a collection property), so it inherits `DataContext` as though it had been parsed as part of the declaration. This is the capability `IDependencyObjectStoreProvider` used to provide before `d8cd44b77a6`/`b5722b7eab0` (BC26) folded `DependencyObjectStore` into `DependencyObject` and made `Parent` internal, with no public replacement.
- **`MarkupHelper.CreateFrameworkTemplate(object? owner, FrameworkTemplateBuilder? factory)`** — completes the `CreateXxxTemplate` set. `DataTemplate`/`ControlTemplate`/`ItemsPanelTemplate` all got a `MarkupHelper` factory when their constructors were narrowed to internal (`ab549229406`); the base `FrameworkTemplate` class did not.

Both were Uno-only public surface to begin with (WinUI has no builder constructor for any of these four types, and no public `DependencyObject.Parent`) — this isn't restoring WinUI parity, it's completing the existing Uno-only bridge that external codegen consumers (like C# Markup) need and generated XAML code already gets for free via same-assembly internal access.

Concretely, this unblocks the last two compile errors in [uno.csharpmarkup#906](https://github.com/unoplatform/uno.csharpmarkup/pull/906) (retargeting C# Markup to Uno 7.0) — see that PR's "Known-red" section for the full analysis.

## PR Checklist ✅

- [x] 🧪 Added Runtime tests / UI tests (`Given_MarkupHelper` unit test; `Given_FrameworkTemplate.When_CreatedThroughMarkupHelper_Then_BehavesLikeDirectConstruction` runtime test)
- [ ] 📚 Docs — n/a, both members are `[EditorBrowsable(Never)]` internal-facing codegen helpers, matching the undocumented `CreateDataTemplate`/`CreateControlTemplate`/`CreateItemsPanelTemplate`/`OnTemplateMemberCreated` siblings
- [ ] 🖼️ Screenshots Compare Test Run — n/a, no rendering/UI changes

https://claude.ai/code/session_01NieLLgaAvny2Uxdt5xxK1M

### DIFF
--- a/src/Uno.UI/Helpers/MarkupHelper.cs
+++ b/src/Uno.UI/Helpers/MarkupHelper.cs
@@ -156,6 +156,7 @@ namespace Uno.UI.Helpers
 		/// content-property assignment path -- for example, an item added by generated code to a collection
 		/// property -- so that it inherits DataContext and other ambient/inherited properties from
 		/// <paramref name="parent"/> as though it had been parsed as part of the declaration.
+		/// Passing <see langword="null"/> detaches the object, clearing the inherited values.
 		/// </summary>
 		/// <remarks>
 		/// <see cref="DependencyObject"/>'s Parent is internal -- WinUI has no equivalent public member, as its
@@ -163,7 +164,7 @@ namespace Uno.UI.Helpers
 		/// the consuming app's assembly, reaches it through this helper.
 		/// </remarks>
 		[EditorBrowsable(EditorBrowsableState.Never)]
-		public static void SetParent(DependencyObject target, object parent)
+		public static void SetParent(DependencyObject target, object? parent)
 			=> target.SetParent(parent);
 
 		/// <summary>
--- a/src/Uno.UI/Helpers/MarkupHelper.cs
+++ b/src/Uno.UI/Helpers/MarkupHelper.cs
@@ -152,6 +152,21 @@ namespace Uno.UI.Helpers
 			=> settings?.OnMemberCreated(target);
 
 		/// <summary>
+		/// Sets the parent of a <see cref="DependencyObject"/> that was created outside its owner's normal
+		/// content-property assignment path -- for example, an item added by generated code to a collection
+		/// property -- so that it inherits DataContext and other ambient/inherited properties from
+		/// <paramref name="parent"/> as though it had been parsed as part of the declaration.
+		/// </summary>
+		/// <remarks>
+		/// <see cref="DependencyObject"/>'s Parent is internal -- WinUI has no equivalent public member, as its
+		/// Parent is computed by the engine from the live visual tree -- so generated code, which compiles into
+		/// the consuming app's assembly, reaches it through this helper.
+		/// </remarks>
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		public static void SetParent(DependencyObject target, object parent)
+			=> target.SetParent(parent);
+
+		/// <summary>
 		/// Creates a <see cref="DataTemplate"/> from a factory.
 		/// </summary>
 		/// <remarks>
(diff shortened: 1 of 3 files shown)


src/Uno.UI/Helpers/MarkupHelper.cs
```diff
@@ -151,6 +151,21 @@ public static void SetElementProperty<TInstance>(object target, string propertyN
...
+		/// <paramref name="parent"/> as though it had been parsed as part of the declaration.
+		/// </summary>
+		/// <remarks>
+		/// <see cref="DependencyObject"/>'s Parent is internal -- WinUI has no equivalent public member, as its
+		/// Parent is computed by the engine from the live visual tree -- so generated code, which compiles into
+		/// the consuming app's assembly, reaches it through this helper.
+		/// </remarks>
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		public static void SetParent(DependencyObject target, object parent)
+			=> target.SetParent(parent);
```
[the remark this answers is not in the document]
MartinZikmund: Good catch, fixed in 28fe3ab2bfc — `parent` is now `object?`.

You're right that this is an established pattern rather than a hypothetical: `DependencyObject.Parent` is declared `internal object? Parent { get; set; }`, and `Given_DependencyProperty.DataContext.When_Object_ChangesParent_With_Detach` already calls `item1.SetParent(null)` and asserts the inherited `DataContext` clears. Since `MarkupHelper.cs` is `#nullable enable` (unlike `DependencyObjectExtensions.cs`, which is nullable-oblivious), the non-nullable parameter would have forced codegen consumers into a CS8625 workaround to express detach.

Also documented the null behaviour on the summary, and added `When_SetParent_Null_Then_DataContextIsCleared` to `Given_MarkupHelper` in 3c98b223d5d to lock it in. All three tests pass; no new nullable warnings, since the internal `SetParent` extension it forwards to is nullable-oblivious.

src/Uno.UI/Helpers/MarkupHelper.cs
```diff
@@ -173,6 +189,12 @@ public static void OnTemplateMemberCreated(DependencyObject target, TemplateMate
 		[EditorBrowsable(EditorBrowsableState.Never)]
 		public static ItemsPanelTemplate CreateItemsPanelTemplate(object? owner, FrameworkTemplateBuilder? factory) => new(owner, factory);
 
+		/// <summary>
+		/// Creates a <see cref="FrameworkTemplate"/> from a factory. See <see cref="CreateDataTemplate"/>.
+		/// </summary>
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		public static FrameworkTemplate CreateFrameworkTemplate(object? owner, FrameworkTemplateBuilder? factory) => new(owner, factory);
```
Xiaoy312, line 196: ??? we shouldnt need to create a FrameworkTemplate; it should've been an "`abstract`" class
if you need for testing, mark this as `internal`, or better yet use a "concrete" impl of this like, DataTemplate/ControlTemplate

Xiaoy312 (approved):

src/Uno.UI/Helpers/MarkupHelper.cs
MartinZikmund: This is something C# markup already used, so changing would require some rewrite on C# markup side...

PR #24383, opened 2026-09-03
fix(a11y): Skip hidden or detached adds in the WASM semantic tree

ajpinedam:
**GitHub Issue:** closes #

## PR Type:

🐞 Bugfix

## What changed? 🚀

On Skia for WebAssembly, the accessibility bridge emitted semantic DOM nodes for controls that should not be in the tree, and those nodes carried zero geometry:

- `UIElement.AddChild` raises the accessibility hook even while the subtree is still being constructed (generated XAML builds bottom-up, before the page is attached). `WebAssemblyAccessibility.OnChildAdded` emitted such children against the semantic root because no live ancestor could be resolved, and nothing re-parented or pruned them once the subtree attached under a `Collapsed` ancestor.
- The same happened for a child added to a live parent that already sits under a `Collapsed` ancestor: the pruning check only inspected the child's own `Visibility`.

Both produced root-level `0x0` nodes (a `16x6` `<button>` in Chromium) for controls that are not rendered, which is what an automation or assistive-technology client then reads.

`OnChildAdded` now skips adds whose parent is not in the live tree and outermost adds beneath a Collapsed ancestor; the attach of the subtree root, or the ancestor being shown (pruned-handle re-emit / Visibility re-add), replays the descendants through the existing recursion. Skipped adds are Trace-logged. This mirrors the guard the macOS host already carries and the router-side drop on Skia desktop.

Visible controls were already positioned correctly; the new geometry tests pin that as a regression guard.

### Behavior change

Controls hosted under a `Collapsed` ancestor no longer appear in the Skia-WASM accessibility tree while hidden. This is WinUI parity (Collapsed subtrees are absent from UIA) and only removes phantom zero-size nodes, but a client that relied on finding a hidden control by `aria-label` will now get "not found" instead of a `0x0` node.

### Tests

Skia-WASM, DOM-level, `Given_AccessibleGeometry`: node rect matches XAML bounds for a live add and for a detached-then-attached subtree; a detached-built control lands under its named group's node; no node under a Collapsed ancestor for both orderings, and the node appears with correct bounds once the ancestor is shown. Shared helpers added to `WasmSemanticDomHelper` (`EnsureAccessibilityEnabledAsync`, `GetSemanticElementRect`, `GetSemanticParentId`).

Runtime validation (Skia-WASM, headless Edge): unfixed build 2 pass / 2 fail on the original four tests; live-tree guard alone leaves the live-Collapsed-ancestor test failing; full fix 5/5; whole `Windows_UI_Xaml_Automation` namespace 213 pass / 0 fail / 19 platform skips.

### Not yet measured

- Skia desktop hosts are not touched by this change (Win32 goes through `AccessibilityRouter`, macOS already has the equivalent guard), so they were not re-run.
- No perf measurement; the guard is a bool read plus, for outermost adds only, an ancestor walk that replaces a JS interop call and dictionary insert per skipped element.

## PR Checklist ✅


ajpinedam:
@Mergifyio backport servicing/6.8

morning4coffe-dev (approved):
Skipping detached additions and pruning beneath a collapsed ancestor is consistent with WinUI's [ChildIsAcceptable visibility gate](https://github.com/microsoft/microsoft-ui-xaml/blob/d0c6705ec46dbddb853cdfdb2eb3b42849cf304e/dxaml/xcp/dxaml/lib/FrameworkElementAutomationPeer_partial.cpp#L810-L830) and recursive child walk. The change does not equate offscreen/zero-size/transparent content with Collapsed. Existing subtree attach and pruned-handle reveal paths supply the deferred emission, and the new cases exercise both ordering variants, geometry and semantic parenting. No blocking source-level discrepancy found.

PR #24394, opened 2026-09-04
test(alc): DO NOT MERGE — probe collectible-ALC reclamation on desktop vs WebAssembly

nickrandolph:
> **Draft, and DO NOT MERGE as-is.** Both commits are a temporary measurement instrument that asserts nothing. They exist to get the WebAssembly leg measured by CI, because that is the only environment that can answer the questions below. The probe is deleted and replaced by real tests before this is proposed for merge.

## Why

studio.live has been chasing a collectible-`AssemblyLoadContext` reclamation failure on WebAssembly across three repos ([studio.live#3359], plus a patch series against a pinned `dotnet/runtime` fork). Every diagnostic lives outside uno, so uno cannot see the regression in its own CI and each round trip costs a cross-repo negotiation.

The intent is to invert that: land a reproduction in uno's own runtime-test suite on `servicing/6.8`, so uno owns the fix and downstream consumers pick up a 6.8-dev pack when it settles.

There is a nice piece of corroboration that this belongs here: `Given_HotReloadClientOperation_Alc.cs:32` already points at `AlcUnloadMemoryRuntimeTests` for collectible-ALC collection coverage — **a test that does not exist in this repo.** It only exists downstream. That dangling reference is precisely the gap.

## What is not knowable without CI

Designing the real tests needs facts the repo does not record, and the WASM leg is the half I cannot obtain locally with confidence:

1. **Does an unloaded collectible ALC ever get reclaimed on WASM, at any collect-round count?** A first local pass had it survive 10 double-collect rounds where desktop reclaimed it in 1. But Mono's interpreter scans the stack conservatively, so a stale stack slot referencing the context looks identical to a genuine pin. This pass runs up to 200 rounds with deliberate stack churn between them to separate those two.

2. **Does either heap move once the collectible payload is substantial?** The first pass saw `HEAPU8` completely flat across 5 cycles that emitted a single empty type — unsurprising, since the downstream staircase came from loading a multi-megabyte previewed app. This pass emits 40 types × 10 methods over 25 cycles and samples both heaps.

## What the probe measures

Per platform: dynamic-code support, whether `AssemblyBuilderAccess.RunAndCollect` yields a genuinely collectible assembly, whether `Assembly.Location` is usable, the collect round at which each of three variants (empty ALC / tiny emitted assembly / payload emitted assembly) is reclaimed or that it never is, and a 25-cycle memory series.

Both heaps are sampled deliberately, and **never compared across legs**: `MemoryManager.AppMemoryUsage` is `Module.HEAPU8.length` on WASM (`MemoryManager.wasm.cs`) but `GC.GetGCMemoryInfo().MemoryLoadBytes` on desktop Skia (`MemoryManager.skia.cs`) — the desktop value is machine-wide memory load, not app-scoped. `GC.GetTotalMemory` is the managed heap on both.

Everything goes to `Console`, which reaches the job log on WASM through Chrome's `--enable-logging=stderr`. Grep the logs for `[ALC-PROBE]`.

## Results already in hand

**Desktop Skia (`SkiaWin32`, .NET 10.0.8), run locally:**

```
IsDynamicCodeSupported=True IsDynamicCodeCompiled=True
assemblyLocation='...\Uno.UI.RuntimeTests.dll' isEmpty=False
emptyAlc:            collected=True atRound=1
emitTinyAssembly:    collected=True atRound=1
emitPayloadAssembly: collected=True atRound=1
managedHeap: 31,705,416 -> 31,795,912 bytes across 25 cycles (flat)
```

Desktop reclaims every variant on the **first** collect round, including the 400-method payload, with a flat managed heap. That is the clean baseline the WASM leg gets compared against.

**WASM (`SkiaWasm`, .NET 10.0.1 browser-wasm), from a local publish + harness drive — the reason this PR exists:**

```
IsDynamicCodeSupported=True  IsDynamicCodeCompiled=False
emit=ok isCollectible=True
assemblyLocation='' isEmpty=True
emptyAlcCollected=False iterations=10
appMemoryUsage=402522112 (flat across 5 tiny cycles)
```

Two things worth flagging from that, both of which this run is meant to confirm or refute under CI conditions:

- **Emit works on the WASM interpreter.** `RunAndCollect` produces a collectible assembly, so a collectible-ALC test *is* writable on that leg — which the existing fixtures' exclusions imply is not the case. `LoadFromAssemblyPath` is confirmed unavailable (`Location` empty), matching the comment in `Given_HotReloadClientOperation_Alc`.
- **An empty collectible ALC — nothing ever loaded into it — was not reclaimed.** If that holds at 200 rounds with stack churn, it is a three-line reproduction that needs no Uno involvement at all, and the `[PlatformCondition(ConditionMode.Exclude, ... SkiaWasm | SkiaUIKit)]` exclusions on `Given_CollectibleAlcAssociations` / `Given_WindowCollectibleEventPrune` are excluding wasm for a sound reason rather than a circumstantial one.

## What happens next

Once the WASM numbers are in, the probe is deleted and replaced by the real coverage: a managed-reachability test whose assertion direction is decided by finding 1, and — only if finding 2 shows the heap actually moves — a wasm-only reclamation test. Neither assertion is written until the measurement supports it; a test that cannot fail, or that encodes GC noise as a defect, is worse than no test.

Not requesting review of the probe itself. Flagging @carldebilly for visibility on the direction, since this is a deliberate move of ownership into uno.

[studio.live#3359]: https://github.com/unoplatform/studio.live/issues/3359

### DIFF
--- a/src/Uno.UI.RuntimeTests/Tests/AssemblyLoadContext/Given_CollectibleAlc_Probe.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/AssemblyLoadContext/Given_CollectibleAlc_Probe.cs
@@ -13,23 +13,28 @@ namespace Uno.UI.RuntimeTests.Tests.AssemblyLoadContext;
 /// <summary>
 /// TEMPORARY MEASUREMENT INSTRUMENT — DELETE BEFORE MERGE.
 ///
-/// Establishes, per platform, which collectible-<see cref="System.Runtime.Loader.AssemblyLoadContext"/>
-/// shapes are actually available, because the answer decides how the real tests can be written and
-/// the repo does not record it anywhere:
+/// Establishes, per platform, how a collectible
+/// <see cref="System.Runtime.Loader.AssemblyLoadContext"/> actually behaves, because the
+/// reproduction's design depends on it and the repo does not record it.
 ///
-/// <list type="bullet">
-///   <item>the real-assembly path is Skia-desktop-only — <c>AlcApp</c> is built by shelling out to
-///   <c>dotnet build</c> at test time and loaded with <c>LoadFromAssemblyPath</c>, and
-///   <c>Given_HotReloadClientOperation_Alc</c> records that <c>Assembly.Location</c> is empty on
-///   WASM;</item>
-///   <item>the emit path (<c>AssemblyBuilderAccess.RunAndCollect</c>) is guarded on
-///   <see cref="RuntimeFeature.IsDynamicCodeSupported"/> throughout the repo, but nothing states
-///   its value on the WASM interpreter leg.</item>
+/// The first pass established: emit works on both legs (WASM reports
+/// <c>IsDynamicCodeSupported=True, IsDynamicCodeCompiled=False</c>), <c>Assembly.Location</c> is
+/// empty on WASM so <c>LoadFromAssemblyPath</c> is unavailable there, an EMPTY collectible ALC dies
+/// in 1 collect round on desktop but survived 10 on WASM, and <c>HEAPU8</c> did not move across 5
+/// tiny emit/unload cycles.
+///
+/// This pass characterises the two open questions that first pass raised:
+/// <list type="number">
+///   <item>Does the WASM ALC die at ANY round count, and is its survival merely Mono's conservative
+///   stack scan seeing a stale slot? Hence many more rounds, with deliberate stack churn between
+///   them to overwrite the staging frame's slots.</item>
+///   <item>Does the linear heap move at all once the collectible payload is substantial rather than
+///   a single empty type? Hence a many-type/many-method payload over more cycles, sampling both the
+///   linear heap and the managed heap so "managed grew" is distinguishable from "linear grew".</item>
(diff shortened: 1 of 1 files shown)


jonpryor:
> But Mono's interpreter scans the stack conservatively, so a stale stack slot referencing the context looks identical to a genuine pin.

It's not *just* Mono's interpreter; that's MonoVM's SGen GC *in general*. Android tests around Java.Lang.Object finalizer behavior ran into a similar problem of "object references are on the stack, the GC scans the stack, keeps the instance alive longer than necessary", and solved that by doing the work on a *separate Thread*:

https://github.com/dotnet/android/blob/da55b862c9b930709d018457c018554a0fad7f1f/external/Java.Interop/tests/Java.Interop-Tests/Java.Interop/JavaObjectTest.cs#L22-L29

https://github.com/dotnet/android/blob/da55b862c9b930709d018457c018554a0fad7f1f/external/Java.Interop/tests/Java.Interop-Tests/Java.Interop/FinalizerHelpers.cs#L40-L45

PR #24410, opened 2026-09-04
fix(skia): Keep keyboard occlusion pad out of the content arrange

### LINKED ISSUE #24409: Soft keyboard occlusion re-arranges ScrollViewer content that sizes itself to the viewport
### Current behavior 🐛

When the on-screen keyboard appears, Uno pads the nearest constrained `ScrollViewer` by the full keyboard overlap so that the `BringIntoView` which follows lands the focused element right above the input pane.

That padding shrinks the **layout** viewport, so content that sizes itself to the viewport is re-arranged into the smaller box. A sign-in card centered in a `Grid` inside a full-window `ScrollViewer` therefore jumps up by **half the keyboard height** as soon as the keyboard opens — even when the focused field sits well above the keyboard and nothing was ever occluded.

Measured with a runtime test on Skia Desktop (viewport 768, occlusion 307.2):

| | pad bottom | viewport | extent | card top |
|---|---|---|---|---|
| keyboard hidden | 0 | 768 | 768 | 284 |
| keyboard shown | 307.2 | 461 | 461 | **130.5** (−153.5 = half the occlusion) |

This is a regression introduced by 6f6bd62a6673fc63cd18328010e2394fbb7ffea2 (`fix(skia): Fix keyboard occlusion scroll padding`, shipped in 6.7.103), which changed the pad from "the amount the focused element is occluded by" to the full overlap. The previous formula avoided this symptom, but was itself broken: it never padded enough to make `BringIntoView` move a genuinely occluded field, which is the bug that commit fixed.

### Expected behavior 🎯

Content that is not occluded should stay exactly where it is. The keyboard overlap should become **scrollable extent** rather than being squeezed out of the content, so that fields behind the keyboard can still be brought into view.

On a Skia mobile head (Android / iOS / WebAssembly on a touch device):

```xml
<ScrollViewer HorizontalAlignment="Stretch" VerticalAlignment="Stretch">
    <Grid>
        <Border Width="300" Height="200"
                HorizontalAlignment="Center" VerticalAlignment="Center">
            <TextBox PlaceholderText="user" Height="40" />
        </Border>
    </Grid>
</ScrollViewer>
```

1. Run the app and tap the `TextBox` to open the soft keyboard.
2. The centered card jumps upward by half the keyboard height, although it was never covered by the keyboard.

Device-independent repro as a runtime test: set `InputPane.GetForCurrentView().OccludedRect` to the bottom 40% of the window while the `TextBox` is focused, and compare the card's position before and after.

### Workaround 🛠️

Pin the `ScrollViewer` content's `MinHeight` to the `ScrollViewer`'s own `ActualHeight` on `SizeChanged` (padding does not change the `ActualHeight` of the element it is set on, so that is the un-occluded height). The card then keeps centering on the full window and the keyboard overlap becomes scrollable extent, so `BringIntoView` still works for genuinely occluded fields.

Alternatively, handle `InputPane.Showing` and set `args.EnsuredFocusedElementInView = true` — but that opts out of the framework adjustment entirely, so occluded fields will no longer be scrolled into view.

### Renderer 🎨

- [x] Skia
- [ ] Native

### Affected platforms 📱💻🖥️

Android, iOS, WebAssembly

6.7.22 (Uno.WinUI 6.7.103); also reproduces on `master`.

### IDE version 🧑‍💻

_No response_

Customer-reported on 6.7.103. A fix is on the way.

ajpinedam:
**GitHub Issue:** closes #24409

## PR Type:

🐞 Bugfix

## What changed? 🚀

When the soft keyboard opens, `InputPane` pads the nearest constrained `ScrollContentPresenter` by the full keyboard overlap, so that the minimal `BringIntoView` which follows lands the focused element flush above the input pane. That padding shrinks the **layout** viewport, so content that sizes itself to the viewport gets re-arranged into the smaller box: a sign-in card centered in a `Grid` inside a full-window `ScrollViewer` jumps up by *half* the keyboard height even when the focused field was never occluded, and the area behind the keyboard is squeezed out of the content instead of becoming scrollable.

The presenter now arranges its content against the **un-occluded** height:

```csharp
childRect.Height = Math.Max(finalSize.Height + _occludedRectPadding.Bottom, desiredSize.Height);
```

The viewport still shrinks — that is what makes `BringIntoView` stop above the keyboard — but the pad turns into scrollable extent rather than moving viewport-sized content. It is a strict no-op whenever no occlusion pad is applied (`_occludedRectPadding.Bottom == 0`), i.e. whenever the keyboard is down.

This is the regression half of 6f6bd62a667 (`fix(skia): Fix keyboard occlusion scroll padding`, in 6.7.103). That commit correctly replaced a pad that was never large enough to move a genuinely occluded field; this PR keeps that behavior and removes the layout side effect it introduced.

Measured on Skia Desktop (viewport 768, occlusion 307.2):

| | pad bottom | viewport | extent | card top |
|---|---|---|---|---|
| before this PR | 307.2 | 461 | 461 | 284 → **130.5** |
| after this PR | 307.2 | 461 | **768** | 284 → **284** |

### Validation

- New runtime test `Given_InputPane.When_Focused_Element_Not_Occluded_Then_Centered_Content_Is_Not_Moved`, verified **fail-before / pass-after** on `master` (fails with "Centered card moved by 153.5px although nothing was occluded", passes with the fix). It waits for the pad to actually be applied before asserting, so it cannot pass vacuously, and it also asserts that the extent grows by the occlusion and that the viewport still shrinks.
- New runtime test `Given_InputPane.When_App_Ensured_Focused_Element_In_View_Then_No_Pad_Is_Applied`, covering the WinUI opt-out (`InputPaneVisibilityEventArgs.EnsuredFocusedElementInView`) on the same code path.
- New runtime test `Given_InputPane.When_Occluded_Field_In_Viewport_Sized_Content_Then_Scrolled_Above_Occlusion`, guarding the bring-into-view contract in the geometry this change actually alters: content that sizes itself to the viewport with the focused field behind the keyboard. It passes both before and after by design — before, the field was squeezed above the keyboard by the shrunken arrange; after, the scroll does the work — which is the point: the guarantee holds either way. The two pre-existing occlusion tests use content ~2040px tall, where this change is a mathematical no-op (`desired` already exceeds `viewport + pad`), so they could not have shown this.
- The three existing `Given_InputPane` tests (the 6f6bd62a667 coverage) still pass — the occluded-field bring-into-view behavior is unchanged.
- Regression sweep: all 55 `Given_ScrollViewer` / `Given_ScrollViewer_Zoom` runtime tests pass (4 skipped), plus 33 `ScrollViewerTests.Given_ScrollViewer_Anchoring` / `_ScrollChaining`.
- All runs on the Skia Desktop head (macOS) against `master`.

Reported by a customer on 6.7.103; a backport to `release/stable/6.7` should follow once this merges, where the same change was validated on the branch directly.

## PR Checklist ✅

### DIFF
--- a/src/Uno.UI.RuntimeTests/Tests/Windows_UI_ViewManagement/Given_InputPane.cs
+++ b/src/Uno.UI.RuntimeTests/Tests/Windows_UI_ViewManagement/Given_InputPane.cs
@@ -195,6 +195,135 @@ public class Given_InputPane
 		}
 	}
 
+	// The occlusion pad shrinks the viewport so BringIntoView can land the focused element above the
+	// keyboard. It must not re-arrange content that sizes itself to that viewport: a sign-in card
+	// centered in a Grid used to jump up by half the keyboard height even though it was never occluded.
+	[TestMethod]
+	[RunsOnUIThread]
+	[RequiresFullWindow]
+	[PlatformCondition(ConditionMode.Include, RuntimeTestPlatforms.Skia)]
+	public async Task When_Focused_Element_Not_Occluded_Then_Centered_Content_Is_Not_Moved()
+	{
+		var (scrollViewer, card, textBox) = BuildCenteredCardPage();
+
+		var inputPane = InputPane.GetForCurrentView();
+		try
+		{
+			await UITestHelper.Load(scrollViewer);
+			Assert.IsTrue(textBox.Focus(FocusState.Programmatic), "TextBox failed to take focus.");
+			await WindowHelper.WaitForIdle();
+
+			var viewportHeight = scrollViewer.ActualHeight;
+			var scrollViewerTopLeft = scrollViewer.TransformToVisual(null).TransformPoint(default);
+			var occludedTop = scrollViewerTopLeft.Y + (viewportHeight * 0.6);
+			var occlusionHeight = viewportHeight * 0.4;
+
+			var cardTopBefore = GetTop(card);
+			Assert.IsTrue(GetBottom(textBox) < occludedTop, "Test setup: the focused TextBox must start above the keyboard top.");
+
+			inputPane.OccludedRect = new Rect(scrollViewerTopLeft.X, occludedTop, scrollViewer.ActualWidth, occlusionHeight);
+
+			// Guards against a vacuous pass: wait until the occlusion path actually padded the presenter.
+			await WindowHelper.WaitFor(
+				() => scrollViewer.Padding.Bottom > 0,
+				message: "The occlusion pad was never applied.");
+			await WindowHelper.WaitForIdle();
+
+			Assert.AreEqual(cardTopBefore, GetTop(card), 0.5, $"Centered card moved by {cardTopBefore - GetTop(card):F1}px although nothing was occluded.");
+
+			// The pad becomes scrollable extent, so the area behind the keyboard stays reachable.
+			Assert.AreEqual(viewportHeight, scrollViewer.ExtentHeight, 0.5, "The occluded area was squeezed out of the content instead of becoming scrollable extent.");
+			Assert.IsTrue(scrollViewer.ViewportHeight < viewportHeight, "The viewport must still shrink so BringIntoView stops above the keyboard.");
+
+			inputPane.OccludedRect = new Rect(0, 0, 0, 0);
+
+			await WindowHelper.WaitFor(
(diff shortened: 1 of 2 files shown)


Xiaoy312 (approved):

ajpinedam:
@Mergifyio backport servicing/6.8

ajpinedam:

Issue #4444, opened 2020-11-02
Support system accent color retrieval with UIColorType.Accent


MartinZikmund:
## What would you like to be added:

`UISettings` class allows retrieval of `UIColorType.Accent` and its darker and lighter versions. This can be supported on Android where system accent color can be set.

## Why is this needed:

UWP parity.

## For which Platform:

- [ ] iOS
- [x] Android
- [ ] WebAssembly
- [ ] WebAssembly renderers for Xamarin.Forms
- [ ] macOS
- [ ] Skia
  - [ ] WPF
  - [ ] GTK (Linux)
  - [ ] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

MartinZikmund:
Gotchas:
- The accent color can potentially be anything (its "rules" are not clearly defined). We need to ensure that we don't use it as the "default accent" color, in case it would break accessibility (or adjust it by shifting the color to a safe range)
- It will require adding some color helpers for RGB -> HSL conversion, so the color brightness can be changed properly to scale to darker/lighter

davidjohnoliver:
Note that (according to [UWP docs](https://docs.microsoft.com/en-us/windows/uwp/design/style/color#accent-color) at least, I haven't tested) the base accent color can also be overridden by app code:

```xaml
<Application.Resources>
    <ResourceDictionary>
        <Color x:Key="SystemAccentColor">#107C10</Color>
    </ResourceDictionary>
</Application.Resources>
```

MartinZikmund:
@davidjohnoliver we have that tracked in #3555 , that would be even more useful to support 👍  (I think most apps override accent color anyway). However, even if overwritten in resources, `UISettings` should still return the "system" color.

jeromelaban:
Now supported in major browsers: https://twitter.com/una/status/1506383666548822020

kucint:
@MartinZikmund: Uno currently only uses the default "windows" blue, and is ignoring the windows "accent-color", see [#16608](https://github.com/unoplatform/uno/discussions/16608#discussioncomment-9368032)
Can this be fixed?

MartinZikmund:
# Repro generator

This issue was classified as not reproducible with a code sample or test.

[...]

PR #24424, opened 2026-09-06
feat(x11): Support OS accent color via the Settings portal

ajpinedam:
**GitHub Issue:** related to #4444 (Linux part; closed by the parent PR #22851)

> **Stacked on #22851** (`dev/mazi/accentcolor`). Only the commits on top of that branch belong to this PR; it should be rebased onto `master` (or the base switched) once #22851 merges.

## PR Type:

✨ Feature

## What changed? 🚀

#22851 introduces `IAccentColorExtension` for Skia hosts and implements it for Win32 (registry) and macOS (`NSColor.controlAccentColor`). On Linux the X11 host had no implementation, so `UISettings.GetColorValue(UIColorType.Accent*)` and the `SystemAccentColor*` resources stayed on the default blue palette.

This PR adds `LinuxAccentColorExtension` to `Uno.UI.Runtime.Skia.X11`, reading the accent from the XDG Settings portal (`org.freedesktop.portal.Settings`, version 2, `org.freedesktop.appearance` / `accent-color`), the same portal the existing `LinuxSystemThemeHelper` already uses for `color-scheme` and `contrast`:

- **Initial read** via `ReadOne`; the value is a `(ddd)` struct of sRGB components in `[0,1]`, converted to a `Color` and expanded with `AccentColorPalette.FromAccentColor` (Linux exposes a single accent color, like macOS).
- **Live updates** via the `SettingChanged` signal; the change is marshalled to the UI thread and raised through `AccentColorChanged`, so the shared `AccentColorHelper` → `Application.OnAccentColorChanged` path refreshes the resources and raises `UISettings.ColorValuesChanged`.
- **Graceful fallbacks** (default palette, no crash): no session bus, portal older than version 2, backend without the optional `accent-color` key (`DBusErrorReplyException`), and out-of-range components, which the portal spec defines as "no accent color set".
- Registered in `X11ApplicationHost` next to `ISystemThemeHelperExtension`. The framebuffer host is not covered (no session bus / portal in that scenario).
- `SystemAccentColorSample` description now mentions Linux (GNOME Settings → Appearance, KDE System Settings → Colors & Themes).

Supported desktops are those whose portal backend implements `accent-color`: GNOME 47+ (`xdg-desktop-portal-gnome`), KDE Plasma 6 (`xdg-desktop-portal-kde`), and others following the version 2 spec. Older backends fall back to the default palette.

## Validation

- **Compile**: `Uno.UI.Runtime.Skia.X11` and `SamplesApp.Skia.Generic` (net10.0, Release) build clean on Windows.
- **Runtime** (WSL2 Ubuntu 24.04 / WSLg, .NET 10): `SamplesApp.Skia.Generic` run against a mock `org.freedesktop.portal.Settings` v2 service on a private session bus, with a temporary runtime test (not committed) asserting `UISettings.GetColorValue(Accent)`, the resolved `SystemColorControlAccentBrush`, and `UISettings.ColorValuesChanged`:
  - initial `(1, 0, 0)` → `GetColorValue(Accent)` = `#FFFF0000`, brush red
  - live `SettingChanged` → green on both, `ColorValuesChanged` raised
  - `(-1, -1, -1)` (out of range) → default `#FF0078D7`, brush follows
  - `(0.5, 0.25, 0.75)` → `(128, 64, 191)` rounding
  - a `color-scheme` change leaves the accent untouched
  - portal without the `accent-color` key → debug log, default palette, no crash

  Not verified on a real GNOME/KDE portal backend or under Wayland: I do not have a Linux desktop at hand, so a manual pass with `SystemAccentColorSample` on GNOME 47+ / Plasma 6 would be welcome.

### Behaviour worth knowing

- The portal is read **asynchronously**: the first frames use the default palette and the OS accent arrives through `AccentColorChanged` a few dispatcher turns later, exactly like the system theme on X11 today. `{ThemeResource}` consumers update; `{StaticResource SystemAccentColor}` consumers resolved during `OnLaunched` keep the default (WinUI parity: static resources never re-resolve). A bounded synchronous first read (`Task.Run` + short `Wait`) would remove the flash at the cost of blocking startup on DBus; I kept the async design to match the theme helper and avoid a UI-thread wait. Happy to switch if reviewers prefer no flash.
- The watch is installed **before** the initial read, so a change racing the read, or a malformed initial value, can never leave the accent stale for the app's lifetime.
- Degraded environments log at **Warning** (no session bus, Settings portal older than v2) and Debug (backend without the `accent-color` key); only unexpected exceptions log at Error.

### Follow-ups (not in this PR)

- The X11 host now opens three independent portal connections (`LinuxSystemThemeHelper`, `X11TextScaleFactorExtension`, this one). A shared Settings-portal client fanning out one `SettingChanged` watch would remove the duplication and the duplicated "no session bus" logs; also align the siblings' `version != 2` gate with the forward-compatible `version < 2` used here.
- The `(ddd)` → palette conversion is pure and could be exposed `internal static` with an `InternalsVisibleTo` for a unit test; today the X11 runtime has no test project reference.
## PR Checklist ✅

- [x] 🧪 Added [Runtime tests, UI tests, or a manual test sample](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) (for bug fixes / features, if applicable) — covered by the existing manual `SystemAccentColorSample`; the portal cannot be driven from CI, see Validation above.
- [ ] 📚 Docs have been added/updated following the [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features) — to follow the parent PR's documentation.

### DIFF
--- a/src/Uno.UI.Runtime.Skia.X11/Helpers/Theming/LinuxAccentColorExtension.cs
+++ b/src/Uno.UI.Runtime.Skia.X11/Helpers/Theming/LinuxAccentColorExtension.cs
@@ -1,0 +1,180 @@
+using System;
+using System.Threading.Tasks;
+using Tmds.DBus.Protocol;
+using Uno.Foundation.Logging;
+using Uno.Helpers.Theming;
+using Uno.UI.Dispatching;
+using Uno.WinUI.Runtime.Skia.X11.DBus;
+using Windows.UI;
+
+namespace Uno.WinUI.Runtime.Skia.X11;
+
+// https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Settings.html
+
+internal class LinuxAccentColorExtension : IAccentColorExtension
+{
+	private const string Service = "org.freedesktop.portal.Desktop";
+	private const string ObjectPath = "/org/freedesktop/portal/desktop";
+	private const string AppearanceNamespace = "org.freedesktop.appearance";
+	private const string AccentColorKey = "accent-color";
+	private const string NotFoundErrorName = "org.freedesktop.portal.Error.NotFound";
+
+	public static LinuxAccentColorExtension Instance { get; } = new();
+
+	public event EventHandler? AccentColorChanged;
+
+	private AccentColorPalette? _currentPalette;
+
+	private LinuxAccentColorExtension()
+	{
+		_ = Init();
+	}
+
+	// Null (default palette) until the asynchronous portal read completes; the OS value then arrives through
+	// AccentColorChanged. Init continues off the UI thread and the setter marshals to the dispatcher, so the
+	// first write always lands after the startup frame that created this instance has subscribed.
+	public AccentColorPalette? GetAccentColorPalette() => _currentPalette;
+
+	private AccentColorPalette? CurrentPalette
+	{
+		get => _currentPalette;
+		set
+		{
+			if (NativeDispatcher.Main.HasThreadAccess)
+			{
+				// The shades are derived from the accent, so comparing Accent is enough to detect a change.
+				if (_currentPalette?.Accent != value?.Accent)
+				{
+					_currentPalette = value;
+					AccentColorChanged?.Invoke(this, EventArgs.Empty);
+				}
+			}
+			else
+			{
+				NativeDispatcher.Main.Enqueue(() => CurrentPalette = value);
+			}
+		}
+	}
+
+	private async Task Init()
+	{
+		try
+		{
+			var sessionsAddressBus = DBusAddress.Session;
+			if (sessionsAddressBus is null)
+			{
+				if (this.Log().IsEnabled(LogLevel.Warning))
+				{
+					this.Log().Warn("Unable to observe the system accent color, the default accent will be used. (Unable to determine the DBus session bus address)");
+				}
+				return;
+			}
+
+			var connection = new DBusConnection(sessionsAddressBus);
+			await connection.ConnectAsync().ConfigureAwait(false);
+
+			var desktopService = new DBusService(connection, Service);
+			var settings = desktopService.CreateSettings(ObjectPath);
+
+			// ReadOne and the accent-color key are version 2 additions; later versions are additive.
+			var version = await settings.GetVersionAsync().ConfigureAwait(false);
+			if (version < 2)
+			{
+				if (this.Log().IsEnabled(LogLevel.Warning))
+				{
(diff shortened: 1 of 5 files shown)


src/Uno.UI.Runtime.Skia.X11/Helpers/Theming/LinuxAccentColorExtension.cs
```diff
@@ -0,0 +1,180 @@
...
+					{
+						this.Log().Error("Unable to process a desktop portal accent color change.", e);
+					}
+				}
+			}).ConfigureAwait(false);
+
+			try
+			{
+				var result = await settings.ReadOneAsync(AppearanceNamespace, AccentColorKey).ConfigureAwait(false);
+				CurrentPalette = ToPalette(result);
```
morning4coffe-dev, line 122: [P2] Do not let the initial read overwrite a newer accent signal

The watch is already active while this read is pending. If it reports a new green accent and the older ReadOne request then completes with red, both updates are queued and this assignment restores red after green. The UISettings value and theme resources can therefore stay stale until the user changes the accent again, despite the watch-before-read ordering.

The exact initializer/publication logic reproduces green -> red with controlled DBus and dispatcher stand-ins; a version check at dispatcher application time keeps green. Track whether an accent signal superseded the initial snapshot (including changes while its UI callback is queued), and add an overlapping-read/signal test.

morning4coffe-dev (changes requested):
Reviewed only the Linux delta over the declared #22851 parent (`7933da57fc32`), without rebasing or retargeting the stack.

The portal namespace/signature, version >= 2 handling and finite RGB range checks match the Settings specification. The exact conversion also produced (128,64,191) for (0.5,0.25,0.75), and rejected out-of-range, NaN and infinity inputs in the controlled probe.

The startup ordering still has a stale-read race. Installing the watch before ReadOne prevents a missed-subscription window, but an older pending ReadOne response can arrive after a newer accent SettingChanged signal. The initializer publishes both unconditionally, so the older snapshot overwrites the newer live value and remains until another signal arrives.

I executed the source initializer, conversion and dispatcher-publication logic with controlled DBus/dispatcher stand-ins: a pending red read followed by a green signal and then the red reply yielded notifications green -> red and a final red palette. A diagnostic generation check when applying the initial snapshot on the dispatcher kept green.

Please prevent the initial snapshot from superseding a newer signal and add a deterministic overlapping-read/signal regression case. Evidence is from exact source with stand-ins, not a live GNOME/KDE session or a new full SamplesApp runtime run.

PR #21513, opened 2025-09-24
feat: Replace `Generic.xaml` resources with WinUI styles

MartinZikmund:
**GitHub Issue:** closes #12839

## What is the current behavior? 🤔

- `Generic.xaml` is significantly outdated

## What is the new behavior? 🚀

`Generic.xaml` is now imported from WinUI (Windows SDK 10.0.26100 / WindowsAppSDK 1.7.3 `generic.xaml`), with the Uno
adaptations kept in an explicitly marked compatibility block at the end of the file and in a much smaller
`SystemResources.xaml`.

Layering: the imported `Generic.xaml` is excluded from `mergedstyles.xaml` and stays the base dictionary, so the Uno
and MUX per-control dictionaries continue to override it. `Given_GenericResources` pins that (the implicit
`ScrollViewer` style still derives from `DefaultScrollViewerStyle`).

Notable behaviour changes:

- **SplitView** now carries the full MUX state machine — 16 visual transitions and 11 states, against 8 and 9 before —
  including all the `OpenInline*` transitions that were missing. The Uno-specific parts (`PaneRoot.Clip` driven from
  `TemplateSettings`, `not_win:DoubleAnimation` in place of keyframe animations) are unchanged from before this PR.
- **TwoPaneView** now uses a verbatim port of MUX's `TwoPaneView.xaml`; the superseded `TwoPaneView_rs1.xaml`
  (commented out in full) and `TwoPaneView_v1.xaml` are removed.
- **ListViewItem foregrounds** now resolve to the Fluent values (`#E4000000` light / `#FFFFFFFF` dark) instead of pure
  black, which is a parity fix; `Given_ListViewBase` is updated accordingly.
- Eleven controls that previously had no default style at all now inherit WinUI's — `ContentPresenter`, `GroupItem`,
  `Hub`, `HubSection`, `ListBox`, `ListBoxItem`, `ListPickerFlyoutPresenter`, `PickerFlyoutPresenter`, `RichEditBox`,
  `SemanticZoom`, `AppBarElementContainer`. This is WinUI-correct but has a wide blast radius; the
  `Screenshots Compare Test Run` results are the thing to look at here.

Uno-only pieces deliberately kept, and why:

- `WebView` / `WebView2` styles: WinUI's `generic.xaml` has none, and
  `CoreWebView2.GetNativeWebViewFromTemplate` resolves the native view through a `ContentPresenter` named
  `WebViewTemplateRoot` as the first child.
[...]

PR #24435, opened 2026-09-07
fix(dp): Synchronize the dependency property lookup cache

### LINKED ISSUE #24433: DependencyProperty lookup cache races with DependencyProperty.Register
### Current behavior 🐛

`DependencyProperty`'s name-to-property lookup cache is not safe against a concurrent registration.

- `DependencyProperty.GetProperty(Type, string)` is UI-thread-only by contract (it throws otherwise unless
  `FeatureConfiguration.DependencyProperty.DisableThreadingCheck` is set).
- `DependencyProperty.Register` is not: it runs on whatever thread first touches a type's static constructor.

Both paths mutate the same pooled lookup key, `_searchPropertyCacheEntry`, whose comment states the assumption
outright: *"This assumes that uses are non-reentrant."* A registration on a background thread can therefore rewrite
the key that a UI-thread lookup is in the middle of querying with, producing a wrong hit or a wrong miss.

Separately, `RegisterProperty` invalidates the cache **before** adding the property to the registry:

```csharp
ResetGetPropertyCache(ownerType, name);
_registry.Add(ownerType, name, newProperty);
```

so a lookup interleaved between the two lines resolves a miss and can publish that stale miss into the cache, after
which the freshly registered property is never found.

### Expected behavior 🎯

Concurrent `Register` / `GetProperty` should never return a property belonging to a different type or name, and never
cache a miss for a property that is being registered.

Timing-dependent, so not reproducible by a plain repro app. The two deterministic tests in
`Given_DependencyProperty_CacheThreading` drive the interleavings directly through test hooks and fail before the fix.

### Workaround 🛠️

Ensure every `DependencyProperty.Register` runs on the UI thread, which in practice means forcing the owning type's
static constructor from the UI thread before any background use.

### Renderer 🎨

- [x] Skia

### Affected platforms 📱💻🖥️

All platforms 🌍

Found while reviewing the `Generic.xaml` import (#12839); unrelated to that change, so it is being fixed separately.

MartinZikmund:
**GitHub Issue:** closes #24433

## PR Type:

🐞 Bugfix

## What changed? 🚀

`DependencyProperty`'s name-to-property lookup cache was not safe against a concurrent registration.

`GetProperty(Type, string)` is UI-thread-only by contract, but `DependencyProperty.Register` is not — it runs on
whatever thread first touches a type's static constructor. Both paths mutated the same pooled lookup key,
`_searchPropertyCacheEntry`, whose own comment stated the assumption: *"This assumes that uses are non-reentrant."*

- The cache and the pooled lookup key are now guarded by a dedicated gate.
- `RegisterProperty` adds the property to the registry **before** invalidating the cache. Previously it invalidated
  first, so a lookup interleaved between the two lines could publish a stale miss and the freshly registered property
  would never be found again.
- A cache version counter makes a lookup that resolved against a since-invalidated cache retry instead of publishing
  a stale entry.

The two hooks the tests drive are compiled only under `DEPENDENCY_PROPERTY_CACHE_TESTS`, which `Uno.UI.UnitTests`
sets on its own `Uno.UI` build, so the shipped assembly is unchanged.

Split out of #21513, where this was found while reviewing the `Generic.xaml` import — it is unrelated to that change.

## PR Checklist ✅

Two deterministic unit tests in `Given_DependencyProperty_CacheThreading` drive the interleavings through the test
hooks and fail before this change. No docs change: internal behavior only, no public API surface.

Note for reviewers: `Uno.UI.UnitTests` now builds `Uno.UI` with an extra define, into its own output path. That is
the part of this change most worth pushing back on if you would rather the hooks were unconditional.

https://claude.ai/code/session_01TFELmYiNR7C7xmcmCR8BDq

### DIFF
--- a/src/Uno.UI.UnitTests/DependencyProperty/Given_DependencyProperty_CacheThreading.cs
+++ b/src/Uno.UI.UnitTests/DependencyProperty/Given_DependencyProperty_CacheThreading.cs
@@ -1,0 +1,194 @@
+#nullable enable
+
+using System;
+using System.Reflection;
+using System.Threading;
+using Microsoft.UI.Xaml;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Uno.UI.Tests.DependencyPropertyTests;
+
+[TestClass]
+[DoNotParallelize]
+public class Given_DependencyProperty_CacheThreading
+{
+	[TestMethod]
+	public void When_Uncached_Properties_Are_Resolved_Concurrently()
+	{
+		var firstExpected = DependencyProperty.Register(
+			FirstPropertyName,
+			typeof(int),
+			typeof(FirstCacheOwner),
+			new PropertyMetadata(0));
+		var secondExpected = DependencyProperty.Register(
+			SecondPropertyName,
+			typeof(int),
+			typeof(SecondCacheOwner),
+			new PropertyMetadata(0));
+		using var firstKeyUpdated = new ManualResetEventSlim();
+		using var secondLookupStarted = new ManualResetEventSlim();
+		using var secondKeyUpdated = new ManualResetEventSlim();
+		using var releaseFirstLookup = new ManualResetEventSlim();
+		using var releaseSecondLookup = new ManualResetEventSlim();
+		DependencyProperty.GetPropertyCacheSearchKeyUpdatedTestHook =
+			(type, propertyName) =>
+			{
+				if (type == typeof(FirstCacheOwner) && propertyName == FirstPropertyName)
+				{
+					firstKeyUpdated.Set();
+					Assert.IsTrue(releaseFirstLookup.Wait(TimeSpan.FromSeconds(5)));
+				}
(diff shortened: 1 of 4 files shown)


morning4coffe-dev (commented):
The cache-key lock, resolving outside that lock, and generation-checked publication address the cache race coherently. However, this must not be described as full WinUI-equivalent registration thread safety: DependencyPropertyRegistry.Add still writes the shared outer/per-type HashtableEx tables without a writer gate, and HashtableEx explicitly assumes a single concurrent writer. Different type initializers can register concurrently. This registry limitation predates the diff but matters to the scenario being enabled; I agree with the existing thread asking for a registry-level synchronization strategy. WinUI protects registration with DynamicMetadataStorageInstanceWithLock: https://github.com/microsoft/microsoft-ui-xaml/blob/439bda90582f7892ca595a193f55dbfc6ae2010c/dxaml/xcp/dxaml/lib/MetadataAPI.cpp#L226-L269 . CI failures/pending work are separate from this source assessment; withholding approval.

Issue #6670, opened 2021-08-05
Cleanup CI warnings


Youssef1313:
## What would you like to be added:

Cleaning up CI warnings. There are lots of CI warnings, from the SDK, NuGet, and the compiler.

## Why is this needed:

Just a cleanup

## For which Platform:

- [ ] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] WebAssembly renderers for Xamarin.Forms
- [ ] macOS
- [ ] Skia
  - [ ] WPF
  - [ ] GTK (Linux)
  - [ ] Tizen
- [ ] Windows
- [ ] Build tasks
- [ ] Solution Templates

## Anything else we need to know?

These warnings should be promoted to error after fixing.

Youssef1313:
I fixed very few in https://github.com/unoplatform/uno/pull/6664

retupmoc258:
I assume this is in reference to the fact that when you make anything in the XAML editor, Visual Studio doesn't recognize it?  For example, I'm getting errors for every type I use in XAML, from Page and Window, to Grid and Rectangle.  Nothing seems to be recognized by the compiler.

https://snipboard.io/FyIYfZ.jpg

Youssef1313:
No, this issue is different. It's *mostly* fixed, except for few NuGet warnings:

![image](https://github.com/unoplatform/uno/assets/31348972/63628975-de60-4a36-b629-4cc77f2b1a67)

We should also figure out how to turn these NuGet warnings into errors after fixing them.

jeromelaban:
These warnings will likely be removed once we move to msbuild-based nuget packaging.

MartinZikmund:
# Repro generator

This issue was classified as not reproducible with a code sample or test.

**Reason:** CI/build enhancement to cleanup warnings.

PR #24445, opened 2026-09-07
ci: Cut the biggest CI warning sources (338 -> ~45 on the build summary)

MartinZikmund:
**GitHub Issue:** #6670

## PR Type:

## What changed? 🚀

The last master CI run (build 232343) reports **338 warnings** on its Azure DevOps summary, and
**565 warning lines** across the raw job logs. This is the first batch of cleanup, taken largest
offender first.

Baseline was measured by downloading all 2235 step logs for that build and counting against the
task-level timeline records only — the job and stage logs concatenate their children, so counting
every log double-counts.

### 282 — `UseDotNet@2` scanning the whole checkout for `global.json`

83% of the warnings on the build summary were one message repeated six times in every job that
installs the SDK:

```
The global.json at path: '.../src/SolutionTemplate/5.3/uno53AppWithLib/global.json' has the wrong
format. ... Error while trying to read: %s
```

`useGlobalJson: true` makes the task glob `**/global.json` from the sources directory and warn once
for every file that declares no `sdk.version`. The six `src/SolutionTemplate/**/global.json` files
only pin `Uno.Sdk.Private`, so each of them warns, in every job, on every agent.

`dotnet-install.yml` already stages the root `global.json` from `build/ci/<subdir>/_global.json`, so
it knows exactly which SDK is wanted. It now reads `sdk.version` out of that file and passes it to
`UseDotNet@2` by name. The task installs the same SDK it resolved before and no longer walks the
23k-file checkout.

### 50 — `WMC1506` OneWay `x:Bind` with nothing to notify

The WinUI XAML compiler raises `WMC1506` when no step in a OneWay binding path raises change
notifications. Each site here binds a source that is assigned once and never mutated — the WinRT
`ContactPhone`/`ContactAddress`/`ContactEmail`/`StorageFile` snapshots, the POCO item view models,
and the `DataTemplateSelector` built in a page constructor — so `OneTime` is what they mean.

`When_xLoad_Order` is deliberately untouched: it drives `x:Load` through an explicit
`Bindings.Update()` call, which is the behaviour under test.

### 48 + 6 — a global `NoWarn` property voiding two projects' suppression lists

`Given_HotReloadWorkspace.BuildTestApp` passed `-p:NoWarn=CS8619`. A **global** MSBuild property
replaces the value a project sets rather than adding to it, so this silently discarded the whole
`NoWarn` list that `Uno.UI.RuntimeTests.HRApp.Skia.csproj` and `Uno.UI.RuntimeTests.HRUnoLib.csproj`
declare — and `UNO0008` from `src/Directory.Build.props` with it. Every hot-reload run therefore
reported the `IDE0055`/`IDE0051`/`UNO0008` diagnostics those projects suppress on purpose.

`CS8619` moves into the two csproj files, where it composes with `$(NoWarn)`. The HRApp sources are
also formatted properly so `IDE0055` has nothing to report either way, and one genuine `CS8604` is
fixed by asserting against the `const` the other five tests in the class already use.

### 39 — `MSB4011` duplicate imports in `SamplesApp.csproj`

The head imported `Uno.UI.Tasks.targets` and `Uno.UI.SourceGenerators.props` directly, and then
imported `sourcegenerators.local.props`, which imports both of them again. Both direct conditions
are strict subsets of the `local.props` one, so the second import was always ignored — with an
`MSB4011` per TFM. The direct imports are removed and the rationale they carried is folded into the
comment on the import that actually provides them.

### 16 — `XA0101` on the Skia Android runtime package

`build/*.*` is a packaging input, but declaring it as `@(Content)` makes the Android SDK raise
`@(Content) build action is not supported` once per props/targets file per TFM. `None` with
`Pack`/`PackagePath` is the idiom the same project already uses for `uno.png`.

### 16 — `CA1014` on `Uno.HotReload`

The one project that opts into `AnalysisMode=Recommended`, so the only one this fires on. The
assembly has never been audited for CLS compliance and is consumed only from C#, so it now declares
that explicitly — the rule's documented resolution.

### Net effect

| | Before | After |
|---|---:|---:|
| Warnings on the ADO build summary | 338 | ~45 |
| Warning lines in the raw job logs | 565 | ~108 |

### Left for follow-up

Deliberately out of scope here, because each needs a decision or real work rather than a cleanup:

- `XA4211`/`XA1006` (6) — raising the samples' `targetSdkVersion` from 36 to 37 pins the app to a
  **preview** Android API level, which is a behavioural decision.
- `MT7156` (16) — `Assets/theme-dark` and `Assets/theme-light` collapse onto one iOS/tvOS
  `LogicalName`, plus Resizetizer emitting both `x.png` and `x.scale-100.png`.
- `NU5128`/`NU5129` (18) — nuspec dependency groups not matching the `lib/` folders.
- `APPX1102` (12), `PRI263` (8), `APT2000` (8).
- 3 `WMC1506` in `When_xLoad_Order`, as described above.
- Agent-side noise: low disk space on the macOS pool, empty iOS/tvOS runtime-test artifacts,
  simulator boot timeouts. The empty-artifact warning may be masking a real staging gap and is worth
  investigating rather than silencing.

## Validation

No behavioural change is intended anywhere in this PR, so each item was checked against the tree it
came from rather than only compiled.

**`MSB4011`** — `msbuild /pp` output compared before and after for `net10.0-android`. Both carry the
same single copy of `Uno.UI.Tasks.targets` and `Uno.UI.SourceGenerators.props`, the same 307
`UsingTask` elements and the same `UnoResourcesGeneration`/`UnoAssetsGeneration` targets;
`UnoUIMSBuildTasksPath`, `ShouldRunGenerator`, `_UnoUnderlyingPlatform`, `_IsAndroid`,
`AssignTargetPathsDependsOn` and `_IsXamlTrimmingAvailable` all evaluate identically. The old tree
reproduces both warnings, the new one is silent.

**`XA0101`** — `Uno.UI.Runtime.Skia.Android` goes from 2 warnings to 0, and the packed
`Uno.WinUI.Runtime.Skia.Android.nupkg` entry list is identical before and after, with
`build/Uno.WinUI.Runtime.Skia.Android.props` and `.targets` still present.

**`CA1014`** — reproduced under `AnalysisMode=All`: raised at `HEAD`, absent with the fix.

**Hot reload** — `HRApp` and `HRUnoLib` build clean; re-adding `-p:NoWarn=CS8619` brings the
warnings straight back, which is the mechanism this PR removes.

**Builds** — `SamplesApp` `net10.0-android` and `net10.0-desktop` both succeed with 0 errors and no
`MSB4011`; the retargeted samples compile. `dotnet format whitespace --verify-no-changes` and
`dotnet xstyler -d src/SamplesApp -r -p` (1521/1521) both pass.

The `dotnet-install.yml` change is the one thing that cannot be exercised locally. The YAML parses,
the resulting `UseDotNet@2` inputs are the same shape as the `haveCheckout: false` task already in
that file, and the PowerShell that extracts the version was run against both
`build/ci/net10/_global.json` and `build/ci/net11/_global.json`. It needs a CI run to confirm.

## PR Checklist ✅

- [x] 🧪 Added [Runtime tests, UI tests, or a manual test sample](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) (for bug fixes / features, if applicable) — n/a, no behaviour change; the before/after evidence above stands in
- [x] 📚 Docs have been added/updated following the [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features) — n/a, no user-facing surface changes

https://claude.ai/code/session_01TFELmYiNR7C7xmcmCR8BDq

### DIFF
--- a/build/ci/templates/dotnet-install.yml
+++ b/build/ci/templates/dotnet-install.yml
@@ -4,10 +4,20 @@ parameters:
   haveCheckout: true
 
 steps:
-  - bash: |
-      rm global.json
+  - pwsh: |
       # The file is named _global.json to avoid being picked up by the UseDotnet@2 task
-      cp build/ci/${{ parameters.globalJsonSubdir }}/_global.json global.json
+      Copy-Item build/ci/${{ parameters.globalJsonSubdir }}/_global.json global.json -Force
+
+      # UseDotNet@2's useGlobalJson mode globs **/global.json across the whole checkout and warns
+      # once per file that carries no sdk.version -- which every src/SolutionTemplate global.json
+      # does, since they only pin Uno.Sdk.Private. Read the version here instead and install it by
+      # name, which also skips the repository-wide scan.
+      $sdkVersion = (Get-Content global.json -Raw | ConvertFrom-Json).sdk.version
+      if ([string]::IsNullOrWhiteSpace($sdkVersion)) {
+        throw "build/ci/${{ parameters.globalJsonSubdir }}/_global.json does not declare an sdk.version."
+      }
+      Write-Host "##vso[task.setvariable variable=GlobalJsonSdkVersion]$sdkVersion"
+      Write-Host "Pinned .NET SDK version: $sdkVersion"
     displayName: "Using .NET global.json from build/ci/${{ parameters.globalJsonSubdir }}"
     condition: ${{ eq(parameters.haveCheckout, true) }}
 
@@ -29,7 +39,8 @@ steps:
     condition: ${{ eq(parameters.haveCheckout, true) }}
     inputs:
       packageType: sdk
-      useGlobalJson: true
+      useGlobalJson: false
+      version: $(GlobalJsonSdkVersion)
       includePreviewVersions: true
       installationPath: $(DOTNET_INSTALL_DIR)
 
--- a/src/SamplesApp/SamplesApp.Samples/Microsoft/UI/Xaml/Controls/TabViewTests/TabViewItemsSourceTests.xaml
+++ b/src/SamplesApp/SamplesApp.Samples/Microsoft/UI/Xaml/Controls/TabViewTests/TabViewItemsSourceTests.xaml
@@ -28,11 +28,11 @@
         </muxc:TabView.TabStripHeader>
         <muxc:TabView.TabItemTemplate>
             <DataTemplate x:DataType="local:TabItemViewModel">
-                <muxc:TabViewItem Header="{x:Bind Header, Mode=OneWay}" IsClosable="False">
+                <muxc:TabViewItem Header="{x:Bind Header, Mode=OneTime}" IsClosable="False">
                     <Button
                         x:Name="ContentButton"
                         Margin="5"
-                        Content="{x:Bind Header, Mode=OneWay}" />
+                        Content="{x:Bind Header, Mode=OneTime}" />
                 </muxc:TabViewItem>
             </DataTemplate>
         </muxc:TabView.TabItemTemplate>
--- a/src/SamplesApp/SamplesApp.Samples/Windows/ApplicationModel/Contacts/PickContact.xaml
+++ b/src/SamplesApp/SamplesApp.Samples/Windows/ApplicationModel/Contacts/PickContact.xaml
@@ -71,7 +71,7 @@
                     <DataTemplate x:DataType="contacts:ContactPhone">
                         <Border
                             Padding="8"
-                            x:DefaultBindMode="OneWay"
+                            x:DefaultBindMode="OneTime"
                             BorderBrush="Black"
                             BorderThickness="1">
                             <StackPanel>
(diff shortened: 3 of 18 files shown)


morning4coffe-dev (approved):
The warning cleanup is coherent at this head: SDK selection uses the intended global.json instead of scanning nested projects; the surviving sourcegenerators.local.props import still establishes UnoUIMSBuildTasksPath and imports the tasks; Pack/PackagePath metadata is retained; and the affected sample bindings represent snapshot/non-notifying properties rather than live notifications. This does not change WinUI control-engine behavior. Approving the source changes, not asserting that the remaining platform CI gates are green.

morning4coffe-dev (approved):
Rechecked this head against my saved original review basis. All PR-specific added/deleted lines are unchanged; the differences are updated master context around SamplesApp and Android packaging. I rechecked the surviving sourcegenerators.local.props import and retained Pack/PackagePath metadata in that context. The latest CI timeline (build 232888) has 50 task-level warnings, no incomplete tasks and no failed tasks, so the large warning reduction is observable rather than only projected. Approval renewed at the current head; no new WinUI control-engine behavior change.

PR #24488, opened 2026-09-10
docs: add Claude Code cloud environment setup

mtmattei:
**GitHub Issue:** closes #

## PR Type:

## What changed? 🚀

Claude Code cloud sessions run on a fresh Ubuntu 24.04 VM with common toolchains pre-installed, but the .NET SDK is not one of them ([Claude Code docs](https://code.claude.com/docs/en/cloud-environments#installed-tools)). The `uno-app` MCP therefore fails to start with `ENOENT: Executable not found in $PATH: dotnet`, and nothing in our getting-started docs covers it.

Three changes:

**1. `get-started-ai-claude.md` — new "Using Claude Code in a cloud environment" section.**

Documents the setup script that installs the .NET SDK, the `DOTNET_ROLL_FORWARD` variable needed when a project targets `net9.0`, and the verification steps. Also notes the two constraints that shape the script: setup scripts must exit zero or the session fails to start, and the filesystem is snapshotted afterward so later sessions skip the script.

**2. `common-issues-ai-agents.md` — new troubleshooting entry for setup scripts failing with exit code 100.**

The base image ships third-party PPAs served from `ppa.launchpadcontent.net`, which is not on the default network allowlist (`ppa.launchpad.net` and `archive.ubuntu.com` are). `apt-get update` exits 100 when any repository fails to refresh, which aborts session start before the .NET install runs. This fails in a confusing way, because the error names Python and PHP repositories in what looks like a .NET problem.

**3. `includes/use-uno-check-inline-linux-noheader.md` — corrected Ubuntu feed guidance.**

The include currently says "If you are using Ubuntu 24.04 or earlier, use the Microsoft feed to install .NET." Microsoft's current guidance is the opposite: for Ubuntu 24.04 and later, ".NET is available in the Ubuntu package manager feeds. The Microsoft package repository no longer contains .NET packages for Ubuntu" ([Install .NET on Ubuntu](https://learn.microsoft.com/dotnet/core/install/linux-ubuntu-install)). The Microsoft feed column reads **None** for 24.04, 25.04, 25.10, and 26.04. Following the current text leads to a missing package rather than an install.

This third change is separable from the first two if you would rather it went in its own PR. It is included here because the new cloud section installs from the Ubuntu feed, which the include currently contradicts. Note it is a shared include, so it also affects the Codex, GitHub Copilot CLI, Cursor, and Antigravity getting-started pages.

### Verification

The setup script in this PR was run in a Claude Code cloud session. `dotnet --version` returns `10.0.112`, and the Uno Platform MCPs connect afterward. The exit-code-100 failure and its error output were reproduced first, which is where the PPA workaround comes from.

## PR Checklist ✅

- [x] 🧪 Added [Runtime tests, UI tests, or a manual test sample](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) (for bug fixes / features, if applicable) — n/a, documentation only
- [x] 🖼️ Validated PR `Screenshots Compare Test Run` results. — n/a, documentation only

### DIFF
--- a/doc/articles/common-issues-ai-agents.md
+++ b/doc/articles/common-issues-ai-agents.md
@@ -42,6 +42,19 @@ The Uno Platform App MCP may fail to start in Claude/Codex/Copilot CLI when it i
 
 To fix this issue, change directories to a folder that contains the `.sln` or `.slnx` file of your project — or, for solution-less projects, the folder containing the `global.json` file that declares your `Uno.Sdk` version.
 
+## A cloud environment setup script fails with exit code 100
+
+A [setup script](https://code.claude.com/docs/en/cloud-environments#setup-scripts) that installs the .NET SDK can fail before it reaches the install:
+
+```text
+E: Failed to fetch https://ppa.launchpadcontent.net/deadsnakes/ppa/ubuntu/dists/noble/InRelease  403  Forbidden
+E: The repository 'https://ppa.launchpadcontent.net/deadsnakes/ppa/ubuntu noble InRelease' is no longer signed.
+```
+
+The base image carries third-party PPAs served from `ppa.launchpadcontent.net`, which is not on the default network allowlist. `apt-get update` exits 100 when any repository fails to refresh, and a setup script that exits non-zero stops the session from starting, so the failure happens before the .NET SDK install runs.
+
+Disable those repository files before updating, and make the update non-fatal. See [Using Claude Code in a cloud environment](xref:Uno.GetStarted.AI.Claude) for the full script.
+
 ## Diagnosing Dev Server issues
 
 > [!TIP]
--- a/doc/articles/get-started-ai-claude.md
+++ b/doc/articles/get-started-ai-claude.md
@@ -68,7 +68,7 @@ For the full skill catalog, update instructions, and other installation options,
     grep -rlE 'launchpadcontent\.net|ppa\.launchpad' /etc/apt/sources.list.d/ 2>/dev/null | xargs -r -I{} mv {} {}.disabled
 
     apt-get update -qq || true
-    apt-get install -y -qq dotnet-sdk-10.0
+    apt-get install -y dotnet-sdk-10.0
     dotnet --version
     ```
 
--- a/doc/articles/get-started-ai-claude.md
+++ b/doc/articles/get-started-ai-claude.md
@@ -48,9 +48,15 @@ For the full skill catalog, update instructions, and other installation options,
 
 ## Using Claude Code in a cloud environment
 
-[Claude Code on the web](https://code.claude.com/docs/en/claude-code-on-the-web) runs each session on a fresh Ubuntu 24.04 virtual machine. The .NET SDK is not pre-installed there, so the `uno-app` MCP fails to start with `ENOENT: Executable not found in $PATH: dotnet`. Install the SDK with a [setup script](https://code.claude.com/docs/en/cloud-environments#setup-scripts), which runs once before Claude Code launches.
+[Claude Code on the web](https://code.claude.com/docs/en/claude-code-on-the-web) runs each session on a fresh Ubuntu 24.04 virtual machine. The .NET SDK is not pre-installed there, so the `uno-app` MCP fails to start with `ENOENT: Executable not found in $PATH: dotnet`. Install the SDK with a [setup script](https://code.claude.com/docs/en/cloud-environments#setup-scripts) on the cloud environment, which runs once before Claude Code launches.
 
-1. Open [claude.ai/code](https://claude.ai/code), then open the environment selector and edit the environment your sessions use.
+1. Open [claude.ai/code](https://claude.ai/code).
+
+1. In the row above the message box, select the cloud icon showing the current environment's name, for example **Default**. This opens the environment selector. There is no settings page or direct URL for it.
+
+1. Under **Cloud**, hover over the environment your sessions use, then select the settings icon that appears on its right. To set up a separate environment for Uno Platform work instead, select **Add cloud environment**.
+
+1. The environment dialog opens with fields for **Name**, **Network access**, **Environment variables**, and **Setup script**. Leave **Network access** on **Trusted**, which reaches the Ubuntu package feed that provides the SDK.
 
 1. In the **Setup script** field, enter:
 
@@ -66,23 +72,34 @@ For the full skill catalog, update instructions, and other installation options,
     dotnet --version
     ```
 
-1. If any project in your repository targets `net9.0`, add the following to the **Environment variables** field. Only the .NET 10 runtime is available from the Ubuntu 24.04 feed, so `net9.0` assemblies need to roll forward:
+1. If any project in your repository targets `net9.0`, add the following line to the **Environment variables** field, which takes `.env` format. Only the .NET 10 runtime is available from the Ubuntu 24.04 feed, so `net9.0` assemblies need to roll forward:
 
     ```text
     DOTNET_ROLL_FORWARD=LatestMajor
     ```
 
-1. Save the environment, then start a **new** session. Sessions that are already running keep the configuration they started with.
+1. Select **Create environment**, or **Save** when you are editing an existing environment.
 
-1. In the session, run `dotnet --version` to confirm the SDK is present, then run `/mcp` to confirm the Uno Platform MCPs are connected.
+1. Start a **new** session. Sessions that are already running keep the configuration they started with, so an existing session will not pick up the change.
 
-The script writes to disk what later sessions need. Setup scripts must exit zero or the session fails to start, which is why `apt-get update` ends in `|| true` and the script ends in `dotnet --version` so a failed install surfaces immediately. After the script completes, the filesystem is snapshotted and reused, so later sessions start with the SDK already installed and skip the script.
+1. The first session in the environment runs the setup script before Claude Code launches, which adds roughly 35 seconds. When it finishes, confirm the install:
+
+    ```bash
+    dotnet --version
+    ```
+
+    This should report a `10.x` version.
+
+1. Run `/mcp` to confirm the Uno Platform MCPs are connected. The `uno-app` entry should now start instead of reporting `ENOENT`.
+
+> [!IMPORTANT]
+> A setup script must exit zero or the session fails to start. That is why `apt-get update` ends in `|| true`, and why the script ends in `dotnet --version` so that a failed install surfaces as a failed session start rather than as a missing binary later. If your session does not start, see [A cloud environment setup script fails with exit code 100](xref:Uno.UI.CommonIssues.AIAgents).
 
 > [!NOTE]
-> .NET 10 is published in the Ubuntu 24.04 package feed, which the default **Trusted** network access level reaches through `archive.ubuntu.com`. .NET 9 is published only in the `ppa:dotnet/backports` repository, which is not on the default allowlist. Target `net10.0`, or set `DOTNET_ROLL_FORWARD` as shown above.
+> The setup script runs on the first session in the environment. Anthropic then snapshots the filesystem and reuses it, so later sessions start with the SDK already installed and skip the script. The script runs again when you change it, when you change the environment's allowed network hosts, and when the snapshot expires after roughly seven days.
 
 > [!TIP]
-> Installing a workload, for example `dotnet workload install android`, adds several minutes to the script. Setup scripts should finish within roughly five minutes, so add workloads on their own line ending in `|| true` and confirm the base install works first.
+> .NET 10 is published in the Ubuntu 24.04 package feed, which the **Trusted** network access level reaches through `archive.ubuntu.com`. .NET 9 is published only in the `ppa:dotnet/backports` repository, which is not on the default allowlist, so target `net10.0` or set `DOTNET_ROLL_FORWARD` as shown above. Installing a workload, for example `dotnet workload install android`, adds several minutes to a script that should finish within roughly five minutes, so add workloads on their own line ending in `|| true` and confirm the base install works first.
 
 ## Next Steps
 
--- a/doc/articles/get-started-ai-claude.md
+++ b/doc/articles/get-started-ai-claude.md
@@ -46,6 +46,44 @@ Uno Platform ships a catalog of agent skills — covering areas such as MVUX, na
 
 For the full skill catalog, update instructions, and other installation options, see [Skills & Plugins](xref:Uno.PlatformStudio.Skills).
 
+## Using Claude Code in a cloud environment
+
+[Claude Code on the web](https://code.claude.com/docs/en/claude-code-on-the-web) runs each session on a fresh Ubuntu 24.04 virtual machine. The .NET SDK is not pre-installed there, so the `uno-app` MCP fails to start with `ENOENT: Executable not found in $PATH: dotnet`. Install the SDK with a [setup script](https://code.claude.com/docs/en/cloud-environments#setup-scripts), which runs once before Claude Code launches.
+
+1. Open [claude.ai/code](https://claude.ai/code), then open the environment selector and edit the environment your sessions use.
+
+1. In the **Setup script** field, enter:
+
+    ```bash
+    #!/bin/bash
+    export DEBIAN_FRONTEND=noninteractive DOTNET_CLI_TELEMETRY_OPTOUT=1
+
+    # Disable third-party PPAs that are not on the default network allowlist
+    grep -rlE 'launchpadcontent\.net|ppa\.launchpad' /etc/apt/sources.list.d/ 2>/dev/null | xargs -r -I{} mv {} {}.disabled
+
+    apt-get update -qq || true
+    apt-get install -y -qq dotnet-sdk-10.0
+    dotnet --version
+    ```
+
+1. If any project in your repository targets `net9.0`, add the following to the **Environment variables** field. Only the .NET 10 runtime is available from the Ubuntu 24.04 feed, so `net9.0` assemblies need to roll forward:
+
+    ```text
+    DOTNET_ROLL_FORWARD=LatestMajor
+    ```
+
+1. Save the environment, then start a **new** session. Sessions that are already running keep the configuration they started with.
+
+1. In the session, run `dotnet --version` to confirm the SDK is present, then run `/mcp` to confirm the Uno Platform MCPs are connected.
+
+The script writes to disk what later sessions need. Setup scripts must exit zero or the session fails to start, which is why `apt-get update` ends in `|| true` and the script ends in `dotnet --version` so a failed install surfaces immediately. After the script completes, the filesystem is snapshotted and reused, so later sessions start with the SDK already installed and skip the script.
+
+> [!NOTE]
+> .NET 10 is published in the Ubuntu 24.04 package feed, which the default **Trusted** network access level reaches through `archive.ubuntu.com`. .NET 9 is published only in the `ppa:dotnet/backports` repository, which is not on the default allowlist. Target `net10.0`, or set `DOTNET_ROLL_FORWARD` as shown above.
+
+> [!TIP]
+> Installing a workload, for example `dotnet workload install android`, adds several minutes to the script. Setup scripts should finish within roughly five minutes, so add workloads on their own line ending in `|| true` and confirm the base install works first.
+
 ## Next Steps
 
 Now that you are set up, let's [create your first app](xref:Uno.GettingStarted.CreateAnApp.AI.Claude).
--- a/doc/articles/includes/use-uno-check-inline-linux-noheader.md
+++ b/doc/articles/includes/use-uno-check-inline-linux-noheader.md
@@ -2,7 +2,7 @@
 1. If `dotnet --version` returns `command not found`:
     - Follow the [official directions](https://learn.microsoft.com/dotnet/core/install/linux?WT.mc_id=dotnet-35129-website#packages) for installing .NET.
       > [!IMPORTANT]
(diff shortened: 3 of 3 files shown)


doc/articles/get-started-ai-claude.md
```diff
@@ -46,6 +46,61 @@ Uno Platform ships a catalog of agent skills — covering areas such as MVUX, na
...
+
+    # Disable third-party PPAs that are not on the default network allowlist
+    grep -rlE 'launchpadcontent\.net|ppa\.launchpad' /etc/apt/sources.list.d/ 2>/dev/null | xargs -r -I{} mv {} {}.disabled
+
+    apt-get update -qq || true
+    apt-get install -y dotnet-sdk-10.0
+    dotnet --version
+    ```
+
+1. If any project in your repository targets `net9.0`, add the following line to the **Environment variables** field, which takes `.env` format. Only the .NET 10 runtime is available from the Ubuntu 24.04 feed, so `net9.0` assemblies need to roll forward:
```
carldebilly, line 75: **MEDIUM** — "Only the .NET 10 runtime is available from the Ubuntu 24.04 feed" is not accurate. Microsoft's *Install .NET on Ubuntu* table lists the built-in 24.04 feed as carrying **10.0 and 8.0**, with 9.0, 7.0 and 6.0 in `ppa:dotnet/backports`. The conclusion for `net9.0` still holds, but the stated reason does not — and a reader with `net8.0` projects is pushed into a two-major roll-forward when `dotnet-runtime-8.0` is installable from the same feed.

Suggested: "the Ubuntu 24.04 feed offers .NET 10 and .NET 8; .NET 9 is only in `ppa:dotnet/backports`".

doc/articles/get-started-ai-claude.md
```diff
@@ -46,6 +46,61 @@ Uno Platform ships a catalog of agent skills — covering areas such as MVUX, na
...
+
+> [!NOTE]
+> The setup script runs on the first session in the environment. Anthropic then snapshots the filesystem and reuses it, so later sessions start with the SDK already installed and skip the script. The script runs again when you change it, when you change the environment's allowed network hosts, and when the snapshot expires after roughly seven days.
+
+> [!TIP]
+> .NET 10 is published in the Ubuntu 24.04 package feed, which the **Trusted** network access level reaches through `archive.ubuntu.com`. .NET 9 is published only in the `ppa:dotnet/backports` repository, which is not on the default allowlist, so target `net10.0` or set `DOTNET_ROLL_FORWARD` as shown above. Installing a workload, for example `dotnet workload install android`, adds several minutes to a script that should finish within roughly five minutes, so add workloads on their own line ending in `|| true` and confirm the base install works first.
```
carldebilly, line 102: **MEDIUM** — This advice undoes the failure detection the `[!IMPORTANT]` note depends on. Followed literally, `dotnet workload install android || true` becomes the script's last command, so the script always exits 0. A failed SDK install then starts the session cleanly, and the reader is back at the `ENOENT: dotnet` symptom this section exists to fix.

Suggested: tell readers to keep `dotnet --version` as the final line and to place workload lines above it.

doc/articles/get-started-ai-claude.md
```diff
@@ -46,6 +46,61 @@ Uno Platform ships a catalog of agent skills — covering areas such as MVUX, na
...
+1. The environment dialog opens with fields for **Name**, **Network access**, **Environment variables**, and **Setup script**. Leave **Network access** on **Trusted**, which reaches the Ubuntu package feed that provides the SDK.
+
+1. In the **Setup script** field, enter:
+
+    ```bash
+    #!/bin/bash
+    export DEBIAN_FRONTEND=noninteractive DOTNET_CLI_TELEMETRY_OPTOUT=1
+
+    # Disable third-party PPAs that are not on the default network allowlist
+    grep -rlE 'launchpadcontent\.net|ppa\.launchpad' /etc/apt/sources.list.d/ 2>/dev/null | xargs -r -I{} mv {} {}.disabled
```
carldebilly, line 68: **MEDIUM** — `grep -rl` matches file *contents*, so any file under `sources.list.d/` that merely mentions the string is renamed whole — including one where the line is commented out, or a deb822 `.sources` file carrying several stanzas alongside the offending one. Combined with the snapshot, those repositories stay disabled for the snapshot's lifetime, and the reader is never told which files were touched or that the rename is reversible.

A narrower form, which also makes reruns idempotent (the `--include` filter cannot re-match `*.disabled`) and prints an inventory into the setup log:

```bash
grep -rlE '^[^#]*launchpadcontent\.net' /etc/apt/sources.list.d/ \
  --include='*.list' --include='*.sources' 2>/dev/null \
  | while IFS= read -r f; do echo "Disabling $f"; mv -n -- "$f" "$f.disabled"; done
```

Worth one sentence saying the files are only renamed and that `mv f.disabled f` restores them. The troubleshooting entry would want the same note.

doc/articles/get-started-ai-claude.md
```diff
@@ -46,6 +46,61 @@ Uno Platform ships a catalog of agent skills — covering areas such as MVUX, na
...
+
+    apt-get update -qq || true
```
carldebilly, line 70: **MEDIUM** — `|| true` suppresses the exit code, which the script needs, but `-qq` additionally suppresses the output. A GPG failure, an expired `Release` file, or a substituted mirror then leaves no evidence anywhere, and the install proceeds against whatever index is on disk. `-qq` was already dropped from `apt-get install` in 2fa36a4; dropping it here too keeps the diagnostic that made this section possible in the first place.

Suggested: `apt-get update || echo "WARNING: apt-get update reported errors; the package index may be stale."`

doc/articles/includes/use-uno-check-inline-linux-noheader.md
```diff
@@ -2,7 +2,7 @@
 1. If `dotnet --version` returns `command not found`:
     - Follow the [official directions](https://learn.microsoft.com/dotnet/core/install/linux?WT.mc_id=dotnet-35129-website#packages) for installing .NET.
       > [!IMPORTANT]
-      > If you are using Ubuntu 24.04 or earlier, use the Microsoft feed to install .NET. On later versions, you may install .NET directly from Ubuntu's official feed. In any case, you may refer to the directions from Microsoft on [installing .NET on Ubuntu](https://learn.microsoft.com/dotnet/core/install/linux-ubuntu#register-the-microsoft-package-repository) and make sure to select your Ubuntu version.
+      > On Ubuntu 24.04 and later, install .NET from the Ubuntu package feed. The Microsoft package repository no longer publishes .NET packages for Ubuntu. On Ubuntu 22.04, .NET is available from both feeds and you should use only one of them. On Ubuntu earlier than 22.04, use the Microsoft package repository. Refer to the directions from Microsoft on [installing .NET on Ubuntu](https://learn.microsoft.com/dotnet/core/install/linux-ubuntu-install) and make sure to select your Ubuntu version.
```
carldebilly, line 5: **MEDIUM** — The 22.04 and pre-22.04 paths do not lead anywhere usable. Microsoft's table shows 22.04 carrying 8.0/7.0/6.0 in the built-in feed and 8.0/7.0/6.0/3.1 in the Microsoft repository — .NET 9 and .NET 10 are in neither, both are in `ppa:dotnet/backports` only. So "available from both feeds… use only one of them" hands a 22.04 reader .NET 8 at best, and `uno.check` then fails on a current Uno project. For "earlier than 22.04", the linked page has no pivot below 22.04, so "select your Ubuntu version" cannot be followed.

This also covers the 22.10/23.x gap raised earlier. Suggested: name `ppa:dotnet/backports` for 22.04 through 23.x, and state that releases earlier than 22.04 are out of .NET support.

carldebilly, line 5: **MEDIUM** — This include reaches more articles than the description states. Beyond the five AI getting-started pages, the transitive set adds `get-started-rider.md`, `get-started-vscode.md`, and — through `includes/use-uno-check-inline.md` — `get-started-dotnet-new.md`: eight in total. The new text is vendor-neutral and reads correctly on all of them, so this is a disclosure gap rather than a defect, but the owners of the IDE Linux setup paths are not currently being asked to review a change to their own instructions.

Either split this file into its own PR as you offered, or list all eight consumers in the description.

doc/articles/get-started-ai-claude.md
carldebilly, line 102: **MEDIUM** — Three alert boxes closing the procedure is unusual for this docs set: no sibling `get-started-ai-*` article stacks them, and `uno-development/docfx.md` is the only other article under `doc/articles` that does. The `[!IMPORTANT]` explains the script from about 30 lines below it, and this `[!TIP]` is a single ~110-word paragraph holding three unrelated points — its roll-forward clause restates step 6 in full, and its one genuinely new fact (workloads need their own line) comes last. Someone asking "how do I add the Android workload" will not look under a tip about feed availability.

Suggested: move the `[!IMPORTANT]`'s rationale next to the script step — its content is correct, only its distance from what it explains is the problem — drop the restated roll-forward clause, and promote the workload guidance to its own step.

doc/articles/get-started-ai-claude.md
```diff
@@ -46,6 +46,61 @@ Uno Platform ships a catalog of agent skills — covering areas such as MVUX, na
...
+    ```bash
+    dotnet --version
+    ```
+
+    This should report a `10.x` version.
+
+1. Run `/mcp` to confirm the Uno Platform MCPs are connected. The `uno-app` entry should now start instead of reporting `ENOENT`.
+
+> [!IMPORTANT]
+> A setup script must exit zero or the session fails to start. That is why `apt-get update` ends in `|| true`, and why the script ends in `dotnet --version` so that a failed install surfaces as a failed session start rather than as a missing binary later. If your session does not start, see [A cloud environment setup script fails with exit code 100](xref:Uno.UI.CommonIssues.AIAgents).
```
carldebilly, line 96: **LOW** — The link text names a specific section, but the `xref` is bare, so the reader lands at the top of the troubleshooting article. This docs set deep-links with `xref:UID#anchor`; `explore-your-app-ai-agents.md:63` does exactly that against this same uid.

Suggested: `xref:Uno.UI.CommonIssues.AIAgents#a-cloud-environment-setup-script-fails-with-exit-code-100`. The reciprocal link in `common-issues-ai-agents.md` needs the same treatment.

doc/articles/common-issues-ai-agents.md
```diff
@@ -42,6 +42,19 @@ The Uno Platform App MCP may fail to start in Claude/Codex/Copilot CLI when it i
...
+A [setup script](https://code.claude.com/docs/en/cloud-environments#setup-scripts) that installs the .NET SDK can fail before it reaches the install:
+
+```text
+E: Failed to fetch https://ppa.launchpadcontent.net/deadsnakes/ppa/ubuntu/dists/noble/InRelease  403  Forbidden
+E: The repository 'https://ppa.launchpadcontent.net/deadsnakes/ppa/ubuntu noble InRelease' is no longer signed.
+```
+
+The base image carries third-party PPAs served from `ppa.launchpadcontent.net`, which is not on the default network allowlist. `apt-get update` exits 100 when any repository fails to refresh, and a setup script that exits non-zero stops the session from starting, so the failure happens before the .NET SDK install runs.
+
+Disable those repository files before updating, and make the update non-fatal. See [Using Claude Code in a cloud environment](xref:Uno.GetStarted.AI.Claude) for the full script.
```
carldebilly, line 56: **LOW** — Same as the reciprocal link: the text names a section but the `xref` is bare, landing the reader on the top of the getting-started article rather than the cloud section holding the script.

Suggested: `xref:Uno.GetStarted.AI.Claude#using-claude-code-in-a-cloud-environment`.

doc/articles/get-started-ai-claude.md
```diff
@@ -46,6 +46,61 @@ Uno Platform ships a catalog of agent skills — covering areas such as MVUX, na
 
 For the full skill catalog, update instructions, and other installation options, see [Skills & Plugins](xref:Uno.PlatformStudio.Skills).
 
+## Using Claude Code in a cloud environment
+
+[Claude Code on the web](https://code.claude.com/docs/en/claude-code-on-the-web) runs each session on a fresh Ubuntu 24.04 virtual machine. The .NET SDK is not pre-installed there, so the `uno-app` MCP fails to start with `ENOENT: Executable not found in $PATH: dotnet`. Install the SDK with a [setup script](https://code.claude.com/docs/en/cloud-environments#setup-scripts) on the cloud environment, which runs once before Claude Code launches.
```
carldebilly, line 51: **LOW** — "a fresh Ubuntu 24.04 virtual machine" plus "runs once" reads as though the script executes on every session, which the `[!NOTE]` further down contradicts: the filesystem is snapshotted and reused. That note is also the reader's only route to "how do I force a re-run", and it is left to inference.

Suggested: say "runs on Ubuntu 24.04" here, let the caching note own the snapshot behaviour, and state explicitly there that editing the script forces a re-run.

doc/articles/common-issues-ai-agents.md
```diff
@@ -42,6 +42,19 @@ The Uno Platform App MCP may fail to start in Claude/Codex/Copilot CLI when it i
...
+## A cloud environment setup script fails with exit code 100
+
+A [setup script](https://code.claude.com/docs/en/cloud-environments#setup-scripts) that installs the .NET SDK can fail before it reaches the install:
+
+```text
+E: Failed to fetch https://ppa.launchpadcontent.net/deadsnakes/ppa/ubuntu/dists/noble/InRelease  403  Forbidden
+E: The repository 'https://ppa.launchpadcontent.net/deadsnakes/ppa/ubuntu noble InRelease' is no longer signed.
+```
+
+The base image carries third-party PPAs served from `ppa.launchpadcontent.net`, which is not on the default network allowlist. `apt-get update` exits 100 when any repository fails to refresh, and a setup script that exits non-zero stops the session from starting, so the failure happens before the .NET SDK install runs.
```
carldebilly, line 54: **LOW** — "`apt-get update` exits 100 when any repository fails to refresh" is broader than apt's actual default. The noble `apt-get(8)` manpage documents `--error-on=any` — "fail the update command if any error occurred, even a transient one" — precisely because the default does not behave that way. The 403 / "no longer signed" case shown above does exit 100; a transient failure does not.

Suggested: scope the sentence to the signature and 403 case actually observed.

doc/articles/get-started-ai-claude.md
```diff
@@ -46,6 +46,61 @@ Uno Platform ships a catalog of agent skills — covering areas such as MVUX, na
...
+
+    ```text
+    DOTNET_ROLL_FORWARD=LatestMajor
+    ```
+
+1. Select **Create environment**, or **Save** when you are editing an existing environment.
+
+1. Start a **new** session. Sessions that are already running keep the configuration they started with, so an existing session will not pick up the change.
+
+1. The first session in the environment runs the setup script before Claude Code launches, which adds roughly 35 seconds. When it finishes, confirm the install:
```
carldebilly, line 85: **LOW** — "roughly 35 seconds" is a single measurement presented as a general figure. Install time varies with archive latency and package size, and an unhedged number turns into a support question when someone waits noticeably longer.

Suggested: attribute it ("in our test, about 35 seconds") or make it qualitative.

doc/articles/get-started-ai-claude.md
```diff
@@ -46,6 +46,61 @@ Uno Platform ships a catalog of agent skills — covering areas such as MVUX, na
...
+
+    apt-get update -qq || true
+    apt-get install -y dotnet-sdk-10.0
+    dotnet --version
+    ```
+
+1. If any project in your repository targets `net9.0`, add the following line to the **Environment variables** field, which takes `.env` format. Only the .NET 10 runtime is available from the Ubuntu 24.04 feed, so `net9.0` assemblies need to roll forward:
+
+    ```text
+    DOTNET_ROLL_FORWARD=LatestMajor
```
carldebilly, line 78: **LOW** — Worth noting that `DOTNET_ROLL_FORWARD` is a host-level setting: per Microsoft's precedence rules it overrides each application's own `runtimeconfig.json` `rollForward` value, for every dotnet process in the session — Dev Server and MSBuild nodes included. This is also the one path the verification did not exercise: `dotnet --version` returning 10.0.112 proves the SDK installed, not that a `net9.0` Uno project builds and runs under the roll-forward.

Suggested: note the variable's scope, and confirm one `net9.0` build in a cloud session.

carldebilly (approved):
Documentation review of the Claude Code cloud environment guidance. I read the three changed files against the surrounding docs conventions and checked the external claims against Microsoft's *Install .NET on Ubuntu* tables, the noble `apt-get(8)` manpage, and Anthropic's cloud-environment documentation.

The change is worth having: the `ENOENT: dotnet` failure is real and undocumented, and the troubleshooting entry cross-links the how-to instead of duplicating the script, which is the right pattern here. The notes below are about the accuracy of a few claims and the shape of the closing block, not about the approach.

**7 medium, 7 low.** Three themes:

1. **Two .NET availability claims do not match Microsoft's current tables.** The built-in Ubuntu 24.04 feed carries .NET 10 *and* .NET 8; and on 22.04 neither feed carries .NET 9 or 10 — both are in `ppa:dotnet/backports` only. The second one leaves a 22.04 reader unable to reach a version current Uno supports.
2. **The script and the prose around it disagree.** The `[!TIP]` tells readers to end workload lines in `|| true`, which makes the script's last command always exit 0 and silently undoes the failure detection the `[!IMPORTANT]` note relies on.
3. **The repository-disabling step needs a narrower match and a way back.** `grep -rl` matches file contents, so a commented-out line or a multi-stanza deb822 `.sources` file is renamed whole, and nothing tells the reader which files were touched or that `.disabled` is reversible.

**On the earlier automated review:** the objection that a failing `apt-get install` aborts the script before `dotnet --version` does not hold. The script has no `set -e`, so execution continues, `dotnet --version` exits 127, and the session fails to start exactly as the note describes. The note is correct as written — what actually breaks that argument is the `|| true` workload advice in the `[!TIP]`. Separately, the `ppa\.launchpad` alternation is defensible, since Launchpad serves PPA content from `ppa.launchpadcontent.net` and a `ppa.launchpad.net` source can still fail; it is the inline comment's stated reason ("not on the default network allowlist") that is wrong for that host, so the comment is what needs fixing rather than the pattern.

**Not anchored to a line:** the description's issue reference reads `**GitHub Issue:** closes #` with no number.

**Lower-confidence observations**, noted without individual comments: the section asserts the `uno-app` MCP will connect but does not say how the MCP registration reaches a fresh VM (the article's own registration step uses `--scope user`, which writes local-machine config); the seven-day snapshot expiry and five-minute script budget are not observable from a single session; the categorical "there is no settings page or direct URL" claim about a third-party UI will rot with nothing to trigger an update; the troubleshooting entry's "disable those repository files" gives the instruction without the match criterion; and the "Trusted" tier's reach is asserted without citing where that allowlist is documented.

One CI note: `Uno.UI - docs (Setup Validations)` is red on the conventional-commit gate — `2fa36a421c1` and `aeafa75c26a` carry no type prefix. A squash-merge with a conformant subject clears it.

morning4coffe-dev (changes requested):
Reviewed all three changed documents and the existing discussion at `2fa36a421c1e`. The cloud setup section is useful, but the existing correctness requests still need attention:

- [The optional workload advice](https://github.com/unoplatform/uno/pull/24488#discussion_r3993090588) can hide a missing SDK. In actual Bash with failing apt/dotnet command stand-ins, the documented base script exits 127; appending the advised `dotnet workload install android || true` changes that to 0. Placing the optional step before the final SDK verification restores exit 127. Keep the critical verification last. Conversely, the earlier automated claim that a nonzero apt-get install inherently stops a script without errexit is incorrect; the base script's final check works as written.
- [The repository-file rewrite](https://github.com/unoplatform/uno/pull/24488#discussion_r3993090597) needs narrower, idempotent handling. Running the actual grep/xargs/mv line on owned fixtures disabled a file containing only a commented PPA reference, disabled a mixed Ubuntu/PPA sources file wholesale, and renamed already-disabled files again on each run. Preserve unrelated active sources and make the operation auditable/reversible.
- [The Ubuntu feed guidance](https://github.com/unoplatform/uno/pull/24488#discussion_r3993090620) needs version-specific instructions. Microsoft's current [tables](https://learn.microsoft.com/dotnet/core/install/linux-ubuntu-install) put .NET 10 and 8 in the Ubuntu 24.04 built-in feed, while .NET 9/10 on 22.04 require the backports feed rather than either built-in/Microsoft feed named in the include.

I also confirmed the registration gap noted in the existing human review: Anthropic's [cloud-environment documentation](https://code.claude.com/docs/en/cloud-environments#what-carries-over-from-your-setup) explicitly says local/user-scoped MCP registrations do not carry over. The earlier commands in this article use `--scope user`; installing dotnet alone does not make those registrations appear in a fresh cloud clone. Document the project-scoped `.mcp.json` or cloud-registration prerequisite before promising connected MCPs.

No duplicate inline comments were added and existing conversations were left unchanged. Validation here is source/authoritative-document review and bounded Bash diagnostics, not an Ubuntu package installation, Claude cloud session, MCP connection or net9 application build. CI 233024 also remains blocked on the two category-missing commit subjects; its failure is a commit-convention gate, not evidence that the new docs built successfully.

morning4coffe-dev (changes requested):
The previous cloud-setup blockers remain: user-scoped MCP registration does not carry into cloud sessions; the apt-source command can disable unrelated or allowed repositories; optional workload advice can hide a failed SDK installation; Ubuntu feed guidance remains inaccurate; and the required documentation setup validation is red. Please address the existing unresolved threads and restore green docs checks.

PR #24494, opened 2026-09-10
fix(android): Keep the compiled java in the .aar

### LINKED ISSUE #24493: Uno.WinUI.Runtime.Skia.Android ships a net10.0-android .aar with no jar, breaking every Android head
### Current behavior 🐛

Every Android head built against Uno 7.0 fails at `javac` while compiling the Java Callable Wrappers:

```
obj\Release\net10.0-android\android\src\crc645c0cc48a56ad0c27\UnoWebViewHandler.java:5: error: package Uno.UI does not exist
        extends Uno.UI.UnoWebViewHandlerJavascriptInterface
                      ^
obj\Release\net10.0-android\android\src\crc645c0cc48a56ad0c27\UnoWebViewHandler.java:13: error: cannot find symbol
                if (getClass () == UnoWebViewHandler.class) {
```

`Uno/UI/UnoWebViewHandlerJavascriptInterface.class` never reaches javac's classpath, because in the `Uno.WinUI.Runtime.Skia.Android` package `lib/net10.0-android36.0/Uno.UI.Runtime.Skia.Android.aar` contains **only** `proguard.txt` — there is no `libs/*.jar`. The `lib/net11.0-android37.0` aar in the *same* package does carry it, so the defect is per-TFM.

Root cause: `Uno.UI.Runtime.Skia.Android.csproj` compiles the java through `@(AndroidJavaSource)`. The .NET Android SDK compiles it in `_CompileBindingJava`, which adds the resulting jar to `@(EmbeddedJar)` **inside the target**, and that target sits only on the build path (`_AndroidBuildDependsOn`).

The project is built twice in one invocation. The second instance is the NuGet pack pass (`BuildProjectReferences=false`), which reaches `_CreateAar` through `ResolveReferences → UpdateAndroidResources → _UpdateAndroidResourcesDependsOn` and never runs `_CompileBindingJava`. `@(EmbeddedJar)` is empty there, `_CreateAarCache` hashes a different input set, and `_CreateAar` reruns and overwrites the good aar with a jar-less one:

```
_CreateAar  Input file "obj\...\Uno.UI.Runtime.Skia.Android.aar.cache" is newer than
            output file "bin\Release\net10.0-android\Uno.UI.Runtime.Skia.Android.aar"
```

### Expected behavior 🎯

`lib/net10.0-android36.0/Uno.UI.Runtime.Skia.Android.aar` carries `libs/<hash>.jar` containing `Uno/UI/UnoWebViewHandlerJavascriptInterface.class`, as the `net11.0-android37.0` aar already does, and Android heads build.

Straight from this repo, no app needed:

```bash
dotnet build src/Uno.UI.Runtime.Skia.Android/Uno.UI.Runtime.Skia.Android.csproj -c Release -p:UnoTargetFrameworkOverride=net10.0-android
```

Then inspect `src/Uno.UI.Runtime.Skia.Android/bin/Release/net10.0-android/Uno.UI.Runtime.Skia.Android.aar` — it holds `proguard.txt` and nothing else. Reproduced on master @ `52379e2f3e6`.

Control: a bare `net10.0-android` class library with a single `@(AndroidJavaSource)` item and no `GeneratePackageOnBuild` produces an aar that **does** contain the jar, so the Android SDK is not at fault on its own — the pack pass is the trigger.

Consumer-side repro: any Uno 7.0 app head on `net10.0-android`, e.g. the Uno.Themes sample apps in unoplatform/Uno.Themes#1722.

### Workaround 🛠️

Re-declare the jar so every pass sees it, in `Uno.UI.Runtime.Skia.Android.csproj`:

```xml
<Target Name="_UnoKeepBindingJarInAar" BeforeTargets="_CreateAarCache" Condition="Exists('$(_AndroidIntermediateBindingClassesZip)')">
	<ItemGroup>
		<EmbeddedJar Include="$(_AndroidIntermediateBindingClassesZip)" Condition="'@(EmbeddedJar)' == ''" />
	</ItemGroup>
</Target>
```

### Renderer 🎨

- [x] Skia
- [ ] Native

### Affected platforms 📱💻🖥️

Android

`Uno.Sdk.Private` 7.0.0-dev.645, 7.0.0-dev.663 and 7.0.0-dev.697 — all affected. .NET SDK 10.0.102 with `Microsoft.Android.Sdk` 36.1.x, TFM `net10.0-android36.0`.

### IDE version 🧑‍💻

N/A — reproduced from the CLI (`dotnet build`).

This blocks the Uno.Themes 7.0 retarget (unoplatform/Uno.Themes#1722 — all three Android sample jobs), and the uno.toolkit.ui retarget queued behind it.

MartinZikmund:
**GitHub Issue:** closes #24493

## PR Type:

🐞 Bugfix

## What changed? 🚀

`lib/net10.0-android36.0/Uno.UI.Runtime.Skia.Android.aar` in the `Uno.WinUI.Runtime.Skia.Android` package contained **only** `proguard.txt` — no `libs/*.jar`. `Uno/UI/UnoWebViewHandlerJavascriptInterface.class` therefore never reached javac's classpath in consuming apps, and every Android head died compiling the Java Callable Wrapper for `UnoWebViewHandler`:

```
UnoWebViewHandler.java:5: error: package Uno.UI does not exist
        extends Uno.UI.UnoWebViewHandlerJavascriptInterface
```

`_CompileBindingJava` is what adds the compiled java to `@(EmbeddedJar)`, and it does so **inside the target**, which sits only on the build path (`_AndroidBuildDependsOn`). The project is built twice in one invocation: the second instance is the NuGet pack pass (`BuildProjectReferences=false`), which reaches `_CreateAar` through `ResolveReferences → UpdateAndroidResources → _UpdateAndroidResourcesDependsOn` and never runs `_CompileBindingJava`. `@(EmbeddedJar)` is empty there, `_CreateAarCache` hashes a different input set, and `_CreateAar` reruns and overwrites the good aar with a jar-less one.

This re-declares the jar in a target that every pass reaches, so the aar is identical whichever path builds it.

### Not a 7.0 regression

The aar has been jar-less since c0a9b6ae60b4a842205adf93c85403823085e071 (2025-08-06) moved the java to `@(AndroidJavaSource)` — `uno.winui.runtime.skia.android` **6.7.65**'s net10.0-android36.0 aar is `['proguard.txt']` too. It was masked because `uno.winui` 6.x still shipped Android TFMs carrying `Uno.UI.BindingHelper.Android.aar`, whose jar holds the *same* class. Uno 7.0 dropped the Android TFMs from `uno.winui`, so nothing supplies it any more.

`Uno.UI.BindingHelper.Android` survives the identical SDK behaviour because it is packed by the hand-authored `build/nuget/Uno.WinUI.nuspec`, which copies the aar out of `bin/Release/<tfm>/`. No `GeneratePackageOnBuild` → no pack pass → `_CreateAar` is never re-entered on it.

### Minimal trigger

Reproduced standalone in a ~20-line project (`AndroidJavaSource` + a `net10.0-android` class library):

| ingredients | compiled java in the aar |
|---|---|
| `AndroidJavaSource` alone | ✅ present |
| + `GeneratePackageOnBuild` | ✅ present |
| + one AndroidX `PackageReference` | ❌ **dropped** |

`GeneratePackageOnBuild` alone is not enough — something else must feed `_CreateAarInputs`. The AndroidX package contributes a `proguard.txt`; without it the item list is empty and MSBuild skips `_CreateAar` in the second pass ("skipped because it has no inputs"). That is why `Uno.Foundation` and `Uno.UI.Dispatching` escape in the same build, while `Uno.UI.Runtime.Skia.Android` (which references `Xamarin.AndroidX.AppCompat` and `Xamarin.AndroidX.DocumentFile`) does not.

### Why `net11.0-android37.0` was unaffected

The .NET 11 Android SDK already fixes this, by moving `_CreateAar` off the reference-resolution path and onto the pack path:

```diff
--- Microsoft.Android.Sdk.BuildOrder.targets   36.1.69                  (.NET 10)
+++ Microsoft.Android.Sdk.BuildOrder.targets   37.0.0-preview.7.2131    (.NET 11)
     <_UpdateAndroidResourcesDependsOn>
       ...
       _UpdateAndroidResgen;
-      _CreateAar;
     </_UpdateAndroidResourcesDependsOn>

     <_IncludeAarInNuGetPackageDependsOn>
       _BuildAndroidGradleProjects;
       _CategorizeAndroidLibraries;
+      _CreateAar;
     </_IncludeAarInNuGetPackageDependsOn>
```

`Xamarin.Android.Javac.targets` is otherwise unchanged between the two (one unrelated `AdditionalStubSourceDirectories` parameter), so `_CompileBindingJava` still contributes `@(EmbeddedJar)` from inside the target — 37.x simply no longer re-enters `_CreateAar` where that item list is empty.

**Implication for this change:** once the repo's Android SDK reaches 37.x, the target below becomes a no-op — its `Condition` still holds but `@(EmbeddedJar)` is already populated, so it adds nothing. It is safe to keep and safe to delete at that point.

## Verification

`dotnet build src/Uno.UI.Runtime.Skia.Android/Uno.UI.Runtime.Skia.Android.csproj -c Release -p:UnoTargetFrameworkOverride=net10.0-android`, then inspecting the outputs:

| | `bin/Release/net10.0-android/*.aar` | `lib/net10.0-android36.0/*.aar` in the nupkg |
|---|---|---|
| before | `['proguard.txt']` | `['proguard.txt']` |
| after | `['libs/<hash>.jar', 'proguard.txt']` | `['libs/<hash>.jar', 'proguard.txt']` |

with `libs/<hash>.jar` containing `Uno/UI/UnoWebViewHandlerJavascriptInterface.class`.

Note that CI cannot catch a regression here: the sample heads in this repo consume `Uno.UI.Runtime.Skia.Android` through a `ProjectReference`, not through the produced package, which is exactly why this shipped unnoticed. The check above is on the packaged artifact.

Not verified: no Android app was launched from the fixed package — the failure was a compile-time classpath one.

## PR Checklist ✅

  > Packaging-only fix, and not observable through a `ProjectReference` — it would need a check on the produced nupkg.

https://claude.ai/code/session_01MbHmxnFUWzjgeBQgdZGC1T

### DIFF
--- a/src/Uno.UI.Runtime.Skia.Android/Uno.UI.Runtime.Skia.Android.csproj
+++ b/src/Uno.UI.Runtime.Skia.Android/Uno.UI.Runtime.Skia.Android.csproj
@@ -44,6 +44,17 @@
 		<AndroidJavaSource Include="..\Uno.UI.BindingHelper.Android\Uno\UI\UnoWebViewHandlerJavascriptInterface.java" />
 	</ItemGroup>
 
+	<!--
+	  _CompileBindingJava is what puts the compiled java in @(EmbeddedJar), and it only runs on the
+	  build path. The pack pass reaches _CreateAar through ResolveReferences instead, so it rewrites
+	  the .aar with no jar in it. Re-declare the jar so every pass produces the same .aar.
+	-->
+	<Target Name="_UnoKeepBindingJarInAar" BeforeTargets="_CreateAarCache" Condition="Exists('$(_AndroidIntermediateBindingClassesZip)')">
+		<ItemGroup>
+			<EmbeddedJar Include="$(_AndroidIntermediateBindingClassesZip)" Condition="'@(EmbeddedJar)' == ''" />
+		</ItemGroup>
+	</Target>
+
 	<ItemGroup>
 		<PackageReference Include="SkiaSharp" />
 		<PackageReference Include="HarfBuzzSharp" />


morning4coffe-dev (approved):
Independently reproduced the packaging failure with the real .NET 10 Android SDK (36.1.69), GeneratePackageOnBuild and a proguard input in an isolated class library. Both builds succeeded, but the baseline nupkg's nested AAR contained only proguard.txt. With this exact target, the same package contained libs/<hash>.jar and Uno/UI/UnoWebViewHandlerJavascriptInterface.class. I also checked the SDK ordering: _CompileBindingJava creates the EmbeddedJar item inside its target, while _CreateAarCache hashes the current item list before _CreateAar packages it, so restoring the existing jar before that cache step addresses the pack-pass overwrite. Approving the scoped artifact fix; the unrelated-platform CI failure still prevents merge.

PR #24502, opened 2026-09-11
fix(wasm): Keep the TextBox in sync with Android soft keyboard input

### LINKED ISSUE #22230: [WASM Skia] Backspace key does not work (TextBox, PasswordBox)
### Current behavior 🐛

Backspace soft keyboard key on Android never works on `PasswordBox` and sometimes does not work on `TextBox`. The easiest way to reproduce textbox failing is to enter text with spaces in it into a textbox and then try to backspace it:

https://github.com/user-attachments/assets/8fc25c90-d717-4327-af96-24f3468788ba

This was tested with Material theme but probably applies to all themes.

### Expected behavior 🎯

Backspace button should work.

_No response_

### Workaround 🛠️

_No response_

### Renderer 🎨

- [x] Skia
- [ ] Native

### Affected platforms 📱💻🖥️

WebAssembly

6.4.53, 6.5.0-dev.100

### IDE version 🧑‍💻

_No response_

_No response_

mikernet:
**GitHub Issue:** closes https://github.com/unoplatform/uno/issues/22230

## PR Type:

🐞 Bugfix

## What changed? 🚀

On Skia WebAssembly, soft keyboards on Android browsers type through IME composition on the hidden native input and report key events without a code. Several things went wrong with that:

1. The composition path inserted the preedit itself and skipped the next input event after compositionend, assuming the browser fires the committed preedit's input event after compositionend (Safari's order). Chrome fires it before, so the skipped event was the space or punctuation that committed the word. The TextBox and the hidden input drifted apart: spaces disappeared, backspace looked dead, and when Gboard re-opened the composition on the whole token the preedit was spliced at the stale caret, duplicating text.

2. Keys were mapped from `event.code` only, and soft keyboards send an empty code, so Backspace mapped to `VirtualKey.None` after the keydown had already been prevented. Backspace never worked in a PasswordBox (no composition there) and stopped working in a TextBox outside a composition. Such keys are now left to the browser and synced from the input event instead.

3. Managed text and caret changes were ignored while a composition was open, and the keyboard keeps one open for the whole word being typed: tapping elsewhere did not move the caret the keyboard inserts at, and clearing the text brought it back on the next keystroke. Replacing the input's value also restored the caret on a later selectionchange, which could land under the keyboard's next composition.

This PR makes the hidden input's value the source of truth for the text: every input event syncs it to the TextBox (through the `TextAlreadyApplied` mode the Android extension already uses), and composition events only report the preedit and where it sits in the value. `ImeCompositionEventArgs` gains a `StartIndex` so platforms can report where a composition begins when it covers existing text. Soft keyboard keys are left to the browser and synced from the input event. Taps, the delete button and `Text` set from code reach the hidden input even during a composition, ending it as a native app would, and the caret is restored synchronously when the value is replaced.

WebKit applies a commit through input events of its own, with no `compositionupdate` for the committed text, and delivers `compositionend` either after them (iOS, verified on a device) or before them. The completion now follows the input event that puts the committed text in place, so `TextCompositionChanged` and the text change precede `TextCompositionEnded` and its handlers see the committed text, as they do on Chrome and as WinUI documents the order. A commit the browser never applies ends the composition with the text as-is at the next keystroke. `TextCompositionEnded` reports its range clamped to the text as it is when a composition ends without its commit, so a coercion or `Text` set from a handler cannot leave it out of bounds.

Validated with new Skia WASM runtime tests that replay Chrome's and WebKit's composition orderings (a commit equal to the preedit, a candidate differing from it and a cancel, with `compositionend` delivered before or after the input events that apply them), a composition re-opened on existing text, the caret moved and the text cleared while composing, and Backspace without a code in a TextBox and a PasswordBox. The existing `Given_TextBox` IME tests pass on WASM and Skia Desktop, and the fix was verified on an Android phone with Gboard.

I did extensive manual testing (iOS Safari and Android Chrome) with all sorts of input sequences, clears, cuts/pastes, bindings, BeforeTextChanging, and multiline inputs, and it appears to be working correctly as far as I can tell.

## PR Checklist ✅


src/Uno.UI.Runtime.Skia.WebAssembly.Browser/ts/Runtime/BrowserInvisibleTextBoxViewExtension.ts
```diff
@@ -435,27 +551,23 @@
...
-					BrowserInvisibleTextBoxViewExtension.waitingAsyncOnSelectionChange = true;
-					BrowserInvisibleTextBoxViewExtension.nextSelectionStart = input.selectionStart;
-					BrowserInvisibleTextBoxViewExtension.nextSelectionEnd = input.selectionEnd;
-					BrowserInvisibleTextBoxViewExtension.nextSelectionDirection = input.selectionDirection;
-					input.value = text;
+				// The value being reported echoes back here unchanged and is a no-op. Any other text
+				// (the TextBox rejecting or coercing the input, the delete button, Text set from code)
+				// does have to reach the input, and ends the composition like it would in a native app.
+				const inputText = BrowserInvisibleTextBoxViewExtension.toInputText(text);
+				if (input.value != inputText) {
```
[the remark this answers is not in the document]
mikernet: The `#if !__SKIA__` only removes the `if (!_isInputModifyingText)` condition, so on Skia `SetTextNative` runs unconditionally for every text change; `When_SoftKeyboard_Text_Cleared_During_Composition` and `When_SoftKeyboard_Text_Set_By_TextCompositionEnded_Handler` cover `Text = ""` and an assignment from a `TextCompositionEnded` handler reaching the hidden input.

mikernet:
@MartinZikmund This PR is ready for review.

mikernet:
@ramezgerges Not sure if you want to review this as well, IIRC you did a bunch of IME related work.

morning4coffe-dev (changes requested):
Requesting changes for the remaining WebKit completion-order issue already acknowledged in the description and raised in [the earlier discussion](https://github.com/unoplatform/uno/pull/24502#discussion_r3988577017). I am not adding a duplicate inline comment or changing that thread's resolution.

I replayed the exact transformed TypeScript bridge with minimal DOM/export stand-ins. Chrome's input-before-compositionend ordering and the existing value-before-end test ordering both expose the committed value at completion. But when WebKit delivers compositionend while the input still contains the preedit, the current trace is `Text=preedit -> Ended(commit=candidate, Text=preedit) -> Text=candidate`. Cancellation likewise reports completion before the restored value arrives.

The managed side makes this observable: WasmImeTextBoxExtension reports TextAlreadyApplied=true and TextBoxCore.OnImeCompositionCompleted therefore skips replacing the committed text before raising TextCompositionEnded. Updating Text in the next input event is too late for consumers reacting to that completion. This is the event-contract regression relative to the previous managed-commit path, even though the eventual text is correct.

[WinUI's documented composition order](https://learn.microsoft.com/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.controls.textbox.textcompositionended) places text-changing/changed and composition-changed notifications before composition-ended. Please defer completion until the committed/restored value has been synchronized, or otherwise preserve that contract, and cover candidate text differing from preedit plus cancellation. The current Safari-order test updates the DOM value before compositionend and does not exercise the acknowledged problematic ordering.

Evidence limits: exact JavaScript-handler execution with replayed browser orderings and managed/native source inspection; not a new physical Safari/WebKit or full managed TextBox runtime run.

mikernet:
@morning4coffe-dev - Addressed. The completion now follows the input event that puts the committed text in place, whichever side of it WebKit delivers `compositionend` on, so `TextCompositionChanged` and the text change precede `TextCompositionEnded` and its handlers see the committed text. Verified on an iPhone with the Japanese keyboard: iOS applies a candidate through `deleteCompositionText`/`insertFromComposition` before `compositionend`, and the sample log now reads Started, Changed あ, Changed 亜, Ended 亜. Covered by tests replaying that order plus `compositionend`-first for an equal commit, a differing candidate and a cancel, with the preedit still in the input or already removed, and a commit the browser never applies.

src/Uno.UI/UI/Xaml/Controls/TextBoxCore/TextBoxCore.IME.cs
```diff
@@ -182,14 +201,24 @@ private void OnImeCompositionEnded()
...
+		RaiseTextCompositionEnded(startIndex, length);
 		InvalidateTextBoxRender();
 	}
 
+	// A text change from outside the composition (a coercion, Text set from code) may have left the
+	// text shorter than the range; handlers get the range as it fits the text now.
+	private void RaiseTextCompositionEnded(int startIndex, int length)
+	{
+		var start = Math.Min(startIndex, Text.Length);
+		_host.RaiseTextCompositionEnded(new TextCompositionEndedEventArgs(start, Math.Min(length, Text.Length - start)));
```
morning4coffe-dev, line 213: [P2] Do not clamp Android's committed range against the previous preedit

This helper now also runs from OnImeCompositionCompleted. AndroidImeTextBoxExtension invokes CompositionCompleted with TextAlreadyApplied=true from the editing-state notification, but TextInputConnection.EndBatchEdit calls `_editable.EndBatchEdit()` before `ActiveTextBox.ProcessTextInput(...)`. Consequently `Text` here is still the old preedit, not the committed value.

Executing those exact callbacks with preedit `ab x` and commit `ab longer` gives StartIndex=3, Length=1 on this head, versus Length=6 before this change; the subsequent text sync applies `ab longer` without another completion. The existing stale-text timing is not introduced by this PR, but silently shortening the committed range is. Please defer this completion until the platform's text sync, or restrict the new clamp to external-change/coercion completion, with an Android regression case whose candidate is longer than its preedit.

morning4coffe-dev (changes requested):
Re-reviewed `a80e267d507b` against my previous reviewed tree `b818b9c5ae31`. The commit series was replaced, but both trees have the same master base; the follow-up delta is four files, not a base merge.

The earlier WebKit completion-order blocker is addressed. I executed both exact TypeScript versions against real headless Edge DOM inputs with synthetic composition/input sequences and managed-export stand-ins: all 14 current-head cases passed, while the previous version failed six. Differing-candidate and cancellation completion now wait for the synchronized value; the cases also covered both positions of WebKit's separate removal event, the author's reported iOS insertion-before-end sequence, repeated preedit, a never-applied prefix commit, external replacement and detach. This confirms the earlier fix without claiming that Edge is Safari or that I reran the author's physical iPhone test.

One new shared-code regression still needs fixing before approval: the new unconditional completion-range clamp also runs on Android's text-already-applied path, whose completion callback actually precedes the managed text sync. I independently reproduced the concern noted in the latest automated review using the exact Android composition callback and EndBatchEdit methods plus the old/new TextBoxCore completion methods. With preedit `ab x` and committed value `ab longer`, the previous code reports start 3 / length 6; this head reports start 3 / length 1, then updates Text to `ab longer` without another completion. Equal-length candidates do not expose it.

The stale Text observed during Android completion already existed; the new regression is truncating the committed range against that stale value. Please preserve the range until the Android text sync has completed, or limit this defensive clamping to the external-change/coercion paths it is intended to protect, and cover a candidate longer than its preedit.

Validation scope: real browser DOM replay of exact JS with managed stand-ins, selected exact C# method execution with native editing-state/host stand-ins, and source comparison with WinUI's text-composition event contract. Not a full Uno WASM app, Android device run, or new native WinUI session. No source changes or other reviewers' thread states were made.

morning4coffe-dev (approved):
Re-reviewed `b59a87e56994` against `a80e267d507b`, with the same master base. The Android committed-range regression is fixed: normal completion now publishes committedText.Length directly, while the bounded helper remains for completion without a commit. ResetComposition clears the session state before invoking the event.

I reran the exact Android callback/EndBatchEdit and old/new shared completion methods with the same editing-state and host stand-ins. The longer-candidate case changes from the previous incorrect range 3+1 to 3+6, and the equal-length control remains 3+3. The cancellation/coercion helper still reports 0+0 when the text has been cleared. This restores the committed range without pretending the pre-existing Android callback-before-text-sync timing was also changed.

The TypeScript refactoring was also executed in a real headless Edge DOM using the same fourteen synthetic composition/input sequences and managed-export stand-ins: 14/14 passed for both this head and the previously reviewed WebKit fix. The differing-candidate/cancellation ordering, separate WebKit removal events, repeated preedit, never-applied commit, external replacement and detach behavior remain covered. I also read the test-helper consolidation and the added per-keystroke selection assertion.

Both findings I raised in the earlier reviews are now addressed. Validation scope is the exact C# callback comparison and real browser DOM replay, not a new physical Android/iPhone test or full Uno/WinUI runtime suite.

src/Uno.UI.Runtime.Skia.WebAssembly.Browser/ts/Runtime/BrowserInvisibleTextBoxViewExtension.ts
```diff
@@ -193,36 +218,209 @@
...
 
 			input.addEventListener("compositionstart", () => {
+				if (!BrowserInvisibleTextBoxViewExtension.isCurrentInput(input)) {
+					return;
+				}
+				// A composition still open here never got its compositionend (a commit the browser never
+				// applied); it is over now. A handler of its end may have moved focus, replacing the input.
+				if (BrowserInvisibleTextBoxViewExtension.isComposing) {
+					BrowserInvisibleTextBoxViewExtension.endComposition();
+					if (!BrowserInvisibleTextBoxViewExtension.isCurrentInput(input)) {
```
morning4coffe-dev: [P2] Check the input's owner when focus can reuse the same element

After endComposition clears isComposing, a TextCompositionEnded handler can focus another ordinary single-line TextBox. focus()/canRetarget() then reuse this very input, replace its value and change currentHandle. This identity check still passes, so the old control's compositionstart continues and calls OnCompositionStarted for the newly focused control.

The exact bridge in a real DOM produces currentHandle=2, sameElement=true, a spurious Started(second), and isComposing=true. The added test only covers a textarea replacement (AcceptsReturn=true), which does not exercise this path. Capture and re-check the owner/session generation across the synchronous callback, not only the element reference, and include a same-kind focus-handoff case.

morning4coffe-dev (changes requested):
Follow-up on `e3bc40b775fa`, which arrived after my approval of `b59a87e56994`. The earlier WebKit completion-order and Android committed-range findings remain fixed. The new input-identity guard also fixes the reentrant focus handoff when the second control requires a different DOM element, as the added multiline test demonstrates.

A same-kind handoff still escapes that guard. I executed the exact previous/current TypeScript in a real headless Edge DOM, with a synchronous completion-export handler calling the actual bridge focus method for a second control. The previous revision incorrectly started a composition on the second control in both cases. This revision passes the replacement-input case, but two ordinary single-line TextBoxes reuse the same input: currentHandle changes from 1 to 2 while isCurrentInput(input) stays true, and the stale callback raises CompositionStarted for the second control and leaves isComposing=true.

The managed callback is synchronously reentrant: TextBox's ITextBoxHost.RaiseTextCompositionEnded directly invokes TextCompositionEnded. The input object is therefore not a sufficient identity for the owning TextBox. Please guard the owner/session generation across endComposition as well, and exercise the new test with a same-kind second control. A reviewer-only currentHandle guard made both real-DOM cases pass; no product code was modified.

Validation scope: exact JavaScript in real browser DOM with synthetic composition events and managed-export stand-ins, plus managed dispatch-source inspection. This is not a physical IME or complete Uno app run. My earlier approval applies to the earlier inspected commit; this latest-head review records the newly verified remaining case.

morning4coffe-dev (approved):
Re-reviewed `4244ded69456` against `e3bc40b775fa`. The focus-owner finding is addressed: the bridge captures currentHandle before synchronous callbacks and checks it as well as the DOM input identity afterward. This covers the same-element retarget path, not only input-to-textarea replacement, and the owner check is carried through compositionend and the input-sync completion path as well.

I reran the exact old/new TypeScript in a real headless Edge DOM. Both owner-handoff cases pass at this head, including the previously failing two-single-line-TextBox case; the previous head passes only replacement and still produces the spurious second-control composition when the element is reused. The original fourteen composition-order/lifecycle cases also remain 14/14. The added runtime tests now parameterize the second control's AcceptsReturn value and assert that its text/composition state stays untouched.

The shared TextBoxCore IME source is byte-identical to the revision where the Android committed-range fix was verified, so that correction is retained. The earlier WebKit ordering cases remain covered by the browser replay. The reported focus-owner thread was already resolved when I checked it; I did not change another reviewer's thread state.

All blocking findings I raised are addressed at this inspected commit. Evidence scope: real browser DOM with synthetic events and managed-export stand-ins, exact-source comparison and the previously recorded callback checks, not a new physical IME/device or full Uno/WinUI runtime-suite run.

mikernet:
@morning4coffe-dev Any chance this will/can get a backport to a `6.7.x` release?

morning4coffe-dev:
@mikernet A `6.7.x` backport needs a separate compatibility check. `release/stable/6.7` has the browser bridge, but not the newer `TextBoxCore` files this patch changes. I haven't validated the equivalent changes/tests on that branch. The current PR has my approval but is still unmerged, so I can't confirm release inclusion yet.

morning4coffe-dev:
@mikernet I think that should be possible, cc @agneszitte / @ajpinedam for more details.

PR #24509, opened 2026-09-12
fix: Refresh NavigationView ancestor selection

### LINKED ISSUE #24508: NavigationView retains selected ancestors after child reparenting
### Current behavior 🐛

Moving the selected data item between hierarchical NavigationView groups can leave `IsChildSelected` set on its former ancestors. Collapsing an unrelated former parent then moves the selection indicator to that group although `SelectedItem` still identifies the original leaf. Selecting a root item can also set its own `IsChildSelected` flag.

### Expected behavior 🎯

Only the current ancestors of the selected item report `IsChildSelected`. Removing selection or recycling an ancestor clears the flag. Collapsing an unrelated group does not take the selected leaf's indicator.

1. Bind a left NavigationView to two groups with `ObservableCollection` child sources and expanded NavigationViewItem containers.
2. Select a leaf under the first group.
3. Remove that same data object from the first collection, add it to the second, and set `SelectedItem` to that object.
4. Collapse the first group.
5. Observe the indicator on the unrelated group while `SelectedItem` remains the leaf in the second group.

The accompanying PR contains runtime tests with the complete minimal setup, plus cases for insertion before a selected ancestor, three-level hierarchy, pane-mode rebinding, clearing selection and a collapsed destination.

### Workaround 🛠️

No application-side mutation of `IsSelected` or `IsChildSelected` is used.

### Renderer 🎨

- [x] Skia
- [ ] Native

### Affected platforms 📱💻🖥️

Desktop (X11), reproduced with the real NavigationView tree. Other platforms have not been runtime-validated for this report.

Uno.Sdk 6.7.22 / Uno.WinUI 6.7.103 (source `cd452e3ee242c9a2a8dee5e65ad8088b6f5dbf83`). The stale-index cleanup is also present on current master `0a4053a40b7bc7fe5f7a0ef5aaa890c5cd5e7a6d` and current `release/stable/6.7` by source inspection.

### IDE version 🧑‍💻

.NET CLI 10.0.400 on Linux x64.

A development build of the equivalent fix at the 6.7.103 source passed 27 collection-migration steps and 96 realized-container samples, followed by native pointer collapse, keyboard activation and focus-release migration. This is a source-built development comparison, not a published package validation. Full upstream runtime CI remains required.

YoungSx:
**GitHub Issue:** closes #24508

## PR Type:

🐞 Bugfix

## What changed? 🚀

Moving a selected child between ObservableCollection groups can leave its former ancestors marked IsChildSelected. Collapsing an unrelated former parent then takes the selection indicator even though SelectedItem still identifies the moved leaf.

This Uno-specific correction tracks actual ancestor containers instead of resolving a stale previous index path after collection changes. It clears state on recycling/deselection, projects it for prepared containers, preserves ancestor identity on index-only changes, and excludes the selected leaf itself. Ancestor entries are removed before invoking dependency-property callbacks so synchronous changes cannot invalidate enumeration.

Follow-up 415d7c743ccc999cc62cd2cfc9b8eab090340d9d addresses the requested changes: after refreshing ancestors, the outer selection callback stops if a synchronous property callback superseded its captured selection. Real runtime coverage includes both a root fallback and a nested fallback chosen by that callback.

## Validation

**Code review:** reviewed ownership, reentrancy, API contracts, security, performance, operability and test quality through the repository's eight-lens panel. The current guard uses final selected-item identity, matching the helper's contract; this is not a claim of generation-level intent tracking.

**Compile:** full source SamplesApp Release/net10.0-desktop built on Linux ARM64 with .NET SDK 10.0.400. Local analyzers were disabled using UnoFastDevBuild. The final build succeeded; existing MSB4011 duplicate-import warnings appear during initial project evaluation even though the build summary reports zero warnings. The previous nullable automation-provider conversion in this PR's runtime test was fixed with a provider-presence assertion.

**Runtime:** four methods / six parameterized executions from NavigationViewTests.AncestorSelection.cs passed twice on the real Skia X11 control tree, with zero failures, skips or retries. Coverage: data-bound reparenting; a collapsed destination; three-level index shifts with/without a pane-mode transition; clearing selection; synchronous ancestor-callback reselection to root and nested fallback items.

The compact-mode fixture respects native leaf auto-close: it reopens the pane before reselecting an explicit nested item and before checking its realized ancestor containers. Selected identity and the top-level ancestor are checked before that second reopening. All original ancestor assertions remain. This is not coverage of selecting a detached descendant with the pane kept closed.

**Negative control:** removing the five-line outer callback guard makes both new fallback rows fail at the retained-selection assertion. Restoring it makes both pass. Earlier development comparisons at Uno 6.7.103 also reproduced the original application indicator defect with the published assembly and passed with the equivalent source fix; those comparisons are not published-package validation.

Reproduction:

```bash
dotnet build src/SamplesApp/SamplesApp/SamplesApp.csproj -c Release -f net10.0-desktop -p:UnoFastDevBuild=true -p:UnoTargetFrameworkOverride=net10.0-desktop -p:UseSharedCompilation=false -m:1 -nr:false
# Base64-encode the fully-qualified method filter as required by the runtime-tests skill.
export UITEST_RUNTIME_TESTS_FILTER=$(printf %s Microsoft.UI.Xaml.Tests.MUXControls.ApiTests.NavigationViewTests.When_AncestorDeselectionCallbackSelectsFallback_NewSelectionIsPreserved | base64 -w0)
cd src/SamplesApp/SamplesApp/bin/Release/net10.0-desktop
xvfb-run -a dotnet SamplesApp.dll --runtime-tests=results.xml
```

Evidence on the validation host: /tmp/pr24509-final3/results.xml, /tmp/pr24509-repeat/results.xml, /tmp/pr24509-negative/results.xml, /tmp/pr24509-final3-build.log. The build process alone used a 3 GiB GC heap cap, workstation GC and GCConserveMemory=9 to limit host pressure; these settings did not change source or the test app's runtime configuration.

**Pending:** maintainer re-review, official upstream runtime/strict CI, and native WinUI behavioral parity. The MUX tests are excluded by the Windows runtime project, so its ordinary job must not be counted as execution of these cases. The Microsoft implementation still uses the old index-path algorithm: this change is intentionally Uno-specific, not a literal port.

Consumers must wait for a published package containing the fix. The downstream application PR salmonloop/salmon-egg#214 remains Draft and must not substitute a development DLL or disable its failing GUI gate to appear green.

## PR Checklist ✅

- [x] 🧪 Added Runtime tests for the bug and adjacent lifecycle cases.
- [ ] 📚 Docs have been added/updated following the documentation template (no public API changes).
- [ ] 🖼️ Validated PR Screenshots Compare Test Run results.
- [ ] 👀 Reviewed 2 other open pull requests (optional).

### DIFF
--- a/src/Uno.UI/UI/Xaml/Controls/NavigationView/NavigationView.cs
+++ b/src/Uno.UI/UI/Xaml/Controls/NavigationView/NavigationView.cs
@@ -276,6 +276,11 @@ public partial class NavigationView : ContentControl
 	private void OnSelectionModelSelectionChanged(SelectionModel selectionModel, SelectionModelSelectionChangedEventArgs e)
 	{
 		var selectedItem = selectionModel.SelectedItem;
+		if (m_appliedTemplate && (selectedItem is null || selectedItem != SelectedItem))
+		{
+			// An index-only change preserves ancestor identities and may precede repeater index updates.
+			UpdateIsChildSelected(selectionModel.SelectedIndex);
+		}
 
 		// Ignore this callback if:
 		// 1. the SelectedItem property of NavigationView is already set to the item
@@ -1032,7 +1037,7 @@ public partial class NavigationView : ContentControl
 					}
 					else if (!IsPaneOpen && indexPath.GetSize() == 0)
 					{
-						UpdateIsChildSelected(indexPathFromModel, null); // Update IsChildSelected Property of Parent Chain
+						UpdateIsChildSelected(null); // Update IsChildSelected Property of Parent Chain
 						if (m_prevIndicator == null && m_nextIndicator == null && m_activeIndicator != null)
 						{
 							// Remove selection indication if we are not in the middle of an ongoing animation
@@ -1353,6 +1358,7 @@ public partial class NavigationView : ContentControl
 				SetNavigationViewItemRevokers(nvi);
 
 				var item = MenuItemFromContainer(nvi);
+				UpdateIsChildSelectedForContainer(nvi);
 
 				if (SelectedItem == item && IsVisible(nvi))
 				{
@@ -1399,6 +1405,11 @@ public partial class NavigationView : ContentControl
 	{
 		if (args.Element is NavigationViewItemBase nvib)
 		{
+			if (nvib is NavigationViewItem item)
+			{
+				SetIsChildSelected(item, false);
+			}
+
 			var nvibImpl = nvib;
 			nvibImpl.Depth = 0;
 			nvibImpl.IsTopLevelItem = false;
@@ -2654,37 +2665,53 @@ public partial class NavigationView : ContentControl
 
 	private void UpdateSelectionModelSelection(IndexPath ip)
 	{
-		var prevIndexPath = m_selectionModel.SelectedIndex;
 		m_selectionModel.SelectAt(ip);
-		UpdateIsChildSelected(prevIndexPath, ip);
+		UpdateIsChildSelected(m_selectionModel.SelectedIndex);
 	}
 
-	private void UpdateIsChildSelected(IndexPath prevIP, IndexPath nextIP)
+	private void UpdateIsChildSelected(IndexPath nextIP)
 	{
-		if (prevIP != null && prevIP.GetSize() > 0)
+		var selectedItem = m_selectionModel.SelectedItem;
+		// The old path may already refer to different rows after a collection mutation.
+		while (m_selectedAncestors.Count > 0)
 		{
-			UpdateIsChildSelectedForIndexPath(prevIP, false /*isChildSelected*/);
+			var index = m_selectedAncestors.Count - 1;
+			var ancestor = m_selectedAncestors[index];
+			m_selectedAncestors.RemoveAt(index);
+			if (ancestor.TryGetTarget(out var container))
+			{
+				container.IsChildSelected = false;
+			}
+			if (m_selectionModel.SelectedItem != selectedItem)
+			{
+				return;
+			}
 		}
 
-		if (nextIP != null && nextIP.GetSize() > 0)
+		if (nextIP != null && nextIP.GetSize() > 2)
 		{
-			UpdateIsChildSelectedForIndexPath(nextIP, true /*isChildSelected*/);
+			UpdateIsChildSelectedForIndexPath(nextIP);
 		}
 	}
 
-	private void UpdateIsChildSelectedForIndexPath(IndexPath ip, bool isChildSelected)
+	private void UpdateIsChildSelectedForIndexPath(IndexPath ip)
 	{
+		var selectedItem = m_selectionModel.SelectedItem;
 		// Update the isChildSelected property for every container on the IndexPath (with the exception of the actual container pointed to by the indexpath)
 		var container = GetContainerForIndex(ip.GetAt(1), ip.GetAt(0) == c_footerMenuBlockIndex /*inFooter*/);
 		// first index is fo mainmenu or footer
 		// second is index of item in mainmenu or footer
 		// next in menuitem children
 		var index = 2;
-		while (container != null)
+		while (container != null && index < ip.GetSize())
 		{
 			if (container is NavigationViewItem nvi)
 			{
-				nvi.IsChildSelected = isChildSelected;
+				SetIsChildSelected(nvi, true);
+				if (m_selectionModel.SelectedItem != selectedItem)
+				{
+					return;
+				}
 				var nextIR = nvi.GetRepeater();
 				if (nextIR != null)
 				{
@@ -2698,6 +2725,34 @@ public partial class NavigationView : ContentControl
 			}
 			container = null;
 		}
+	}
+
+	private void UpdateIsChildSelectedForContainer(NavigationViewItem container)
+	{
+		var selectedPath = m_selectionModel.SelectedIndex;
+		var containerPath = GetIndexPathForContainer(container);
+		var isAncestor = selectedPath != null && containerPath.GetSize() > 1 && containerPath.GetSize() < selectedPath.GetSize();
+		for (var index = 0; isAncestor && index < containerPath.GetSize(); index++)
+		{
+			isAncestor = containerPath.GetAt(index) == selectedPath.GetAt(index);
+		}
+		SetIsChildSelected(container, isAncestor);
+	}
+
+	private void SetIsChildSelected(NavigationViewItem container, bool isChildSelected)
+	{
+		for (var index = m_selectedAncestors.Count - 1; index >= 0; index--)
+		{
+			if (!m_selectedAncestors[index].TryGetTarget(out var ancestor) || ancestor == container)
+			{
+				m_selectedAncestors.RemoveAt(index);
+			}
+		}
+		if (isChildSelected)
+		{
+			m_selectedAncestors.Add(new global::System.WeakReference<NavigationViewItem>(container));
+		}
+		container.IsChildSelected = isChildSelected;
 	}
 
 	private void RaiseItemInvoked(
--- a/src/Uno.UI.RuntimeTests/MUX/Microsoft_UI_Xaml_Controls/NavigationView/NavigationViewTests.AncestorSelection.cs
+++ b/src/Uno.UI.RuntimeTests/MUX/Microsoft_UI_Xaml_Controls/NavigationView/NavigationViewTests.AncestorSelection.cs
@@ -8,7 +8,9 @@ using Microsoft.UI.Xaml.Automation.Provider;
 using Microsoft.UI.Xaml.Controls;
 using Microsoft.UI.Xaml.Markup;
 using Private.Infrastructure;
+using Uno.UI.RuntimeTests;
 using Uno.UI.RuntimeTests.Helpers;
+using static Private.Infrastructure.TestServices;
 
 namespace Microsoft.UI.Xaml.Tests.MUXControls.ApiTests;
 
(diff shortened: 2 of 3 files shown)


src/Uno.UI/UI/Xaml/Controls/NavigationView/NavigationView.cs
```diff
@@ -276,6 +276,11 @@ private void OnOverflowItemsSourceCollectionChanged(object sender, object args)
 	private void OnSelectionModelSelectionChanged(SelectionModel selectionModel, SelectionModelSelectionChangedEventArgs e)
 	{
 		var selectedItem = selectionModel.SelectedItem;
+		if (m_appliedTemplate && (selectedItem is null || selectedItem != SelectedItem))
+		{
+			// An index-only change preserves ancestor identities and may precede repeater index updates.
+			UpdateIsChildSelected(selectionModel.SelectedIndex);
```
morning4coffe-dev, line 282: [P2] Preserve selection changes made by ancestor-property callbacks

`selectedItem` is captured before this call, but clearing an old ancestor invokes synchronous `IsChildSelected` dependency-property callbacks. If a callback selects a fallback item while the model is clearing a removed leaf, `UpdateIsChildSelected` correctly notices the new selection and returns. Its caller does not: it continues to `SetSelectedItemAndExpectItemInvokeWhenSelectionChangedIfNotInvokedFromAPI(selectedItem)` using the captured `null`, undoing the callback's selection.

Executing the exact methods with minimal model/container stand-ins produces `Fallback -> null`; a post-refresh selection-change guard preserves `Fallback`. Guard or revalidate this outer callback after the refresh, not only the helper, and add a real runtime test that selects a fallback from the old ancestor's property-change callback.

morning4coffe-dev (changes requested):
The ancestor-identity approach addresses the stale-index-path problem, but the new early ancestor refresh creates a reentrancy boundary that the outer selection callback does not protect.

I exercised the exact selected methods with minimal container/selection-model stand-ins. When clearing the model selection resets an old ancestor's IsChildSelected, a synchronous property callback can select a fallback item. UpdateIsChildSelected notices the model change and returns, but OnSelectionModelSelectionChanged then continues with its previously captured null selectedItem and clears that fallback. The trace is `ancestor callback selects Fallback -> SelectedItem=Fallback -> SelectedItem=null`. A diagnostic post-refresh selection guard preserves the fallback.

The concrete SelectionModel raises SelectionChanged synchronously, and IsChildSelected is a public dependency property, so this is a reachable callback boundary rather than only a list-enumerator concern. Please protect the caller against the superseding selection and add a runtime case covering that fallback/reselection path.

WinUI source note: the pinned Microsoft implementation (`d0c6705ec46d`, controls/dev/NavigationView/NavigationView.cpp) still uses the previous-index-path approach and has no new ancestor refresh before this callback's ignore/selection checks. This is an intentional Uno-specific algorithm change, not a literal port; native behavioral parity should remain explicitly pending as the PR already states.

Evidence scope: exact-method execution with stand-ins and source tracing, not a full XAML/native WinUI runtime reproduction.

src/Uno.UI/UI/Xaml/Controls/NavigationView/NavigationView.cs
YoungSx: Addressed in 415d7c743ccc999cc62cd2cfc9b8eab090340d9d.

The outer `OnSelectionModelSelectionChanged` callback now returns immediately after ancestor refresh if the SelectionModel's selected item differs from its captured value. The check follows the existing helper's final-selection identity contract; the caller reads the current index afterwards, so no separate selection-generation owner is introduced.

Added a real loaded NavigationView runtime test with two rows: root fallback and fallback under another parent. A synchronous old-ancestor property callback selects the fallback after removing the selected leaf. Both rows assert the callback ran, the fallback remains selected, the old ancestor is cleared, and the nested replacement ancestor is selected where applicable.

Validation on Linux ARM64, .NET SDK 10.0.400, using the full source-built SamplesApp Release/net10.0-desktop:
- All six ancestor runtime cases passed twice, with no retries or skips.
- Removing the production guard makes both new fallback rows fail at the retained-selection assertion; restoring it makes both pass.
- Full build passed. Initial project evaluation still emits existing MSB4011 duplicate-import warnings; local analyzers were disabled with UnoFastDevBuild, so this is not a strict CI result.

The full build also exposed the prior test's nullable automation-provider cast; it now asserts the provider exists before collapsing. The compact-pane fixture now explicitly reopens the native pane around reselection before inspecting nested containers. It still checks selected identity and the top-level ancestor while the pane is closed, and retains all insertion/deselection/clearing assertions. This test does not claim coverage of selecting a detached descendant while the pane remains closed.

Native WinUI parity and the upstream official runtime/strict CI remain pending. Please re-review this change and enable the required upstream validation. Thank you for identifying the outer-callback boundary.

morning4coffe-dev (approved):
Re-reviewed `415d7c743ccc999cc62cd2cfc9b8eab090340d9d` against my previous `e8f70b7` review and the complete three-file feature delta. Revalidating the captured item after the synchronous ancestor callbacks fixes the reported ownership boundary: the outer callback no longer applies its stale selection over the replacement, and the selected index is read after that revalidation.

I independently built the source SamplesApp in Release/net11.0-desktop and ran all six ancestor-selection cases twice on the real Skia Win32 control tree; all passed without skips or retries. A reviewer-only copy removing just the five-line outer guard made both root/nested fallback rows fail at the retained-selection assertion. Recompiling the original reviewed source restored all six passes. The provider-presence assertion and explicit compact-pane reopening were also reviewed; this does not claim coverage of detached-child selection while the pane remains closed.

Local builds used UnoFastDevBuild, not strict analyzer coverage; initial project evaluation still emitted the existing MSB4011 duplicate-import warnings. Native WinUI behavioral parity and the full upstream matrix remain pending. I requested the missing ordinary CI, and build 233542 is now running with this exact head verified as a merge parent. This is not a claim that CI has passed.

My reported reentrancy issue is fixed in this head. Approving the revision; normal independent-review and CI merge gates still apply.

morning4coffe-dev:
CI follow-up for `415d7c743ccc999cc62cd2cfc9b8eab090340d9d`:

The initial iOS shard's only listed failure was `Given_InputPane.When_Focused_Element_Not_Occluded_Then_Centered_Content_Is_Not_Moved`. The exact base `06690db8` reproduces the same failure in build 233431's first attempt (test run 3942304); its targeted **one-test** retry passed (run 3942331). This is baseline-reproduced intermittent behavior, not new evidence against the ancestor-selection fix. I am not calling the whole base suite green.

The screenshot notice also needs its comparison scope kept clear: the comparer log used builds **211834 (May 10)** and **212586 (May 14)** as baselines, not the PR's exact base. Its 2,310 Windows differences therefore are not an isolated PR-versus-base regression count; this does not establish pixel parity either.

Update: build 233542 has now completed successfully for this exact head, and the required checks are passing. Keeping the source approval unchanged; independent approval is still required before merging. No check or queue bypass has been applied.

MartinZikmund:
@YoungSx  This control is fully ported from WinUI, so Uno-specific changes should only be done in case there is an upstream bug in the WinUI control itself (and it should then also be reported on WinUI repository). Can you please check whether this issue is also happening on WinAppSDK? If not, then we must have some other issue in our ported sources.

morning4coffe-dev:
@MartinZikmund @YoungSx I checked this against **native Windows App SDK 2.4.0/.NET 10**, with no Uno control implementation loaded. The original stale-ancestor problem does reproduce upstream.

I compiled the unchanged six-case fixture from `415d7c74` into a standalone native app, supplying only native host/load/idle helpers and the test metadata marker. In two fresh runs, with 250 ms and then 750 ms idle barriers:

- Moving the selected leaf between expanded parents, or into a collapsed destination, reaches the retained-selection assertion successfully but fails because the former source container still has `IsChildSelected=true`.
- Inserting a root before the ancestor and then selecting a top-level item leaves the original parent marked child-selected, in both normal and compact-pane variants.
- The two fallback-callback rows stop earlier: native never clears the old ancestor, so the callback count is **0**, not 1. Those rows do **not** establish native behavior for the later reentrancy boundary fixed in the PR.

That is four repeatable original-bug scenarios after their initial selection/ancestor positive controls. The separately inspected [upstream source at 439bda9](https://github.com/microsoft/microsoft-ui-xaml/blob/439bda90582f7892ca595a193f55dbfc6ae2010c/controls/dev/NavigationView/NavigationView.cpp#L2494-L2511) still resolves cleanup through the previous index path; I am not claiming that checkout is the exact source commit of the packaged runtime.

This provides native repro evidence for an upstream report. Please file/link that report as requested before settling the Uno-specific implementation. It is not a claim of complete algorithm parity or mobile coverage, and I have not opened an upstream issue or bypassed the remaining merge gates.

YoungSx:
@MartinZikmund @morning4coffe-dev Thanks for investigating and verifying this against native Windows App SDK!

As requested, I have filed the upstream issue in the WinUI repository with the repro details:
https://github.com/microsoft/microsoft-ui-xaml/issues/11927

Given that this is confirmed to be an upstream WinUI defect and the Uno-specific fix plus regression tests in this PR have been verified and approved, could you please advise if any further steps are needed, or proceed with the independent review for merging? Thank you!

YoungSx:
@MartinZikmund @morning4coffe-dev Friendly follow-up. Since we have now confirmed the stale-ancestor defect reproduces on native Windows App SDK 2.4.0 and filed the upstream WinUI issue (microsoft/microsoft-ui-xaml#11927), would you be open to merging this Uno-side fix as an interim workaround, gated so it can be cleanly reverted once WinUI ships the upstream fix?

The change is scoped to refreshing NavigationView ancestor selection and is covered by the new root-level and nested fallback regression tests (the two added cases fail when the fix is removed). It has been APPROVED and the six-case Skia/X11 fixture passed two consecutive rounds with no retries or skips.

Happy to add a code comment linking #11927 with a removal condition, or adjust the gating if that helps. Thanks for your time!

PR #24514, opened 2026-09-15
chore(sdk): Pin Uno.Sdk.Extras to stable 6.3.1

agneszitte:
**GitHub Issue:** relates to unoplatform/private#1137 (6.7 SR2)

## PR Type:

🧹 Chore

## What changed? 🚀

`src/Uno.Sdk/packages.json` (the `Uno.Sdk.Private` manifest) pinned `Uno.Sdk.Extras` to **`6.4.0-dev.13`** — an unbranched `main` dev build — inside a stable SDK. This repoints it to the current stable **`6.3.1`**.

**This is not a downgrade.** The entire delta between `release/stable/6.3` and the `main` build that produced `6.4.0-dev.13` is:

```
.vsts-ci.yml                              |  5 ++
build/ci/templates/host-cleanup-linux.yml | 41 ++   (new file)
version.json                              |  2 +-
```

13 commits, **zero shipped-code change** — CI plumbing and a version bump. `6.3.1` and `6.4.0-dev.13` produce functionally identical packages, so this removes a prerelease pin from a stable SDK at no behavioural cost to anyone already on 6.7.

## Why now

The 6.6 line scrubbed its dev pins before shipping (uno.templates `a4d00c58`, *"fix: replace dev versions with stable 6.6/6.5 releases"*), which replaced this exact literal — `6.4.0-dev.13` → `6.3.1`. The 6.7 line skipped that step, so the pin has shipped in `Uno.Sdk` 6.7.18 and 6.7.22.

The matching change to the public manifest is in unoplatform/uno.templates#2247.

## PR Checklist ✅

- [ ] 🧪 Automated test — N/A, a version pin in the SDK manifest.
- [ ] 📚 Docs — N/A.


morning4coffe-dev (approved):
Verified the pin rather than relying on the PR description. The base/head manifests differ semantically only in `sdkextras.version`, changing `6.4.0-dev.13` to `6.3.1`; all other groups and overrides are unchanged.

Both cached package extractions match their original nupkg entries. The nuspecs have the same dependency surface, and the six shipped props/targets files match after normalizing the generated commit-qualified task identities. A read-only metadata/IL comparison found no differences across 830 records, including 505 methods, after that same identity normalization. The exact repository commits recorded in the packages also differ only in CI/version files. The packages are not byte-identical: version, generated identities and signing/build metadata differ. No package task code was executed by this comparison, and I did not run a complete packaging matrix.

Code approval, not a green-CI or merge-readiness claim. CI 233666 checks out `cd4f990c99485926cd973bd1ecadf95fde1f336f`, whose parents include this head. Android bucket 0 fails in SDK setup at `sdkmanager --install tools` (`Failed to find package 'tools'`), before its runtime tests; the aggregate was still in progress when inspected.

PR #24519, opened 2026-09-16
perf(uno-ui): Index DO collections in tree walks

MartinZikmund:
**GitHub Issue:** closes #24517

## PR Type:

🔄 Refactoring (no functional changes, no api changes)

## What changed? 🚀

The Enter, Leave and resource-binding walks visited the collections a property holds through `IEnumerable`, so each visit boxed a `List<T>.Enumerator` (40 B). Every `VisualStateGroup`, `VisualState`, `Setter`, `StateTrigger` and `VisualTransition` collection of every templated control paid it on each Enter, each Leave and each theme update — in a virtualized list, on every recycled container. `VisualStateManager.VisualStateGroups`, which XAML codegen stores as a `VisualStateGroup[]`, boxed an array enumerator on the same walks, and `TextBlock.UpdateThemeBindings` boxed one per `InlineCollection`.

`DependencyObjectCollectionBase` now carries the backing list as an indexable `IList`, and the six walk sites enumerate through one `DependencyObjectItems` ref struct that indexes framework DO collections and arrays, and falls back to the value's own enumerator for anything else. `TextBlock` uses the struct enumerator `InlineCollection` already had.

Two details worth calling out:

- **Mutation during a walk keeps throwing.** Indexing alone would silently skip or revisit an item if a callback modified the collection mid-walk, where the old enumerator threw. The collection now bumps a version on each mutation and the walker checks it, so the behaviour is the same as `List<T>.Enumerator` — including the Enter/Leave `catch` that swallows it, and the theming walk where it propagates. WinUI instead `Lock()`s the collection during the walk (`CDOCollection::EnterImpl`); adopting that would change which code throws, so it belongs in its own change.
- **Not a full parity port.** In WinUI the collection's own `EnterImpl`/`LeaveImpl` walks its items; in Uno the owner still drives the walk (the existing `TODO Uno` in `DependencyObject.PropertySystem.mux.cs`). This change only removes the boxing from the owner-driven walk; the seam it adds is the data source a future port would reuse.

### Measured impact 📊

Managed bytes allocated per iteration on the UI thread (`GC.GetAllocatedBytesForCurrentThread`, median of 7 rounds, two runs per side; reproduces within ~1 %):

| Workload | Before | After | Δ |
|---|---:|---:|---:|
| Enter + Leave `Button` | 18,592 B | 17,376 B | −6.5 % |
| Enter + Leave `CheckBox` | 24,744 B | 21,848 B | −11.7 % |
[...]

PR #24523, opened 2026-09-16
perf(theming): Walk resource dictionaries without an iterator

### LINKED ISSUE #24522: Resource dictionary lookups allocate an iterator on every walk

`DependencyObject.GetResourceDictionaries` returns the ResourceDictionaries in scope for an object, nearest first. It is a C# iterator (`yield return`), so **every call allocates a compiler-generated state machine** before producing a single dictionary — including the very common case where no ancestor carries resources and the walk yields nothing at all.

It runs on the hottest resource paths:

- **once per theme reference**, from Phase A of `UpdateThemeReference` — an object with N `{ThemeResource}` references performs N separate ancestor walks and allocates N iterators (the code's own comment cites `ListViewItemPresenter` carrying 41+ theme refs)
- **twice per element Enter** for the implicit-style lookup, since `ApplyStyles` runs from both `EnterImpl` and the Loading pass
- **on every visual-tree resource retrieval**, from `ResourceResolver.TryVisualTreeRetrieval`

An allocation profile of a scrolling `ListView` puts the generated iterator type at the top of the table, ahead of every text-shaping type; it ranks first in template build as well.

### Why is this needed? 🤔

A struct enumerator that visits the same dictionaries in the same order removes the allocation without touching resolution semantics. The ancestor order is load-bearing (nearest scope wins, then the app resources), so it is reproduced exactly rather than re-derived; callers that genuinely want an array keep using the iterator.

### Measured impact 📊

Managed bytes allocated per iteration on the UI thread (`GC.GetAllocatedBytesForCurrentThread`, median of 7 rounds, two runs per side; figures reproduce within ~1%):

| Workload | Before | After | Δ |
|---|---:|---:|---:|
| Enter + Leave `TextBox` | 30,590 B | 22,152 B | −27.6 % |
| Enter + Leave `Button` | 18,592 B | 14,520 B | −21.9 % |
| Enter + Leave a 40-control page | 796,857 B | 650,481 B | −18.4 % |
| Enter + Leave `CheckBox` | 24,744 B | 20,488 B | −17.2 % |
| Enter + Leave `ListViewItem` | 37,828 B | 32,024 B | −15.3 % |
| Enter + Leave `ToggleSwitch` | 48,139 B | 41,292 B | −14.2 % |
| `ListView` (2,000 items) scroll step | 178,516 B | 157,212 B | −11.9 % |
| Theme toggle on that page | 540,152 B | 509,512 B | −5.7 % |

Over a longer run of the same scroll workload — 1,000 steps, so the collection count is a resolvable integer rather than the 0–2 a short run produces:

| Metric (1,000 scroll steps) | Before | After | Δ |
|---|---:|---:|---:|
| Allocated per step | 189,498 B | 168,583 B | −11.0 % |
| Gen0 collections, total | 15 | 13 | −2 |

Both sides reproduced their gen0 count exactly across two runs.

### Renderer 🎨

- [x] Skia

### For which platforms 📱💻🖥️

All platforms 🌍

Measured with a headless Skia app (`Uno.UI.Runtime.Skia.Headless`) that project-references the framework, with Fluent resources merged. Wall-clock is not reported: it varied by more than the effect between identical runs on the measuring machine. Gen0 collection counts were captured but at 0–2 per round are too coarse to resolve a difference.

A separate, larger opportunity remains on the same path and is deliberately **not** covered here: `UpdateAllThemeReferences` performs one full ancestor walk *per theme reference*, so an object with N references walks the tree N times. Collapsing that into a single walk needs care, because the surrounding code notes that updating one reference can cascade and modify the theme-reference map, and a newly resolved `Style` can change the ancestor scope mid-loop.

MartinZikmund:
**GitHub Issue:** closes #24522

## PR Type:

## What changed? 🚀

`DependencyObject.GetResourceDictionaries` returns the ResourceDictionaries in scope, nearest first. It is a C# iterator, so **every call allocates a state machine** — including the very common case where no ancestor carries resources and the walk yields nothing. It runs once per `{ThemeResource}` reference (an object with N references performs N separate ancestor walks), twice per element Enter for the implicit-style lookup, and on every visual-tree resource retrieval.

`EnumerateResourceDictionaries` now returns a struct enumerator that performs that walk without allocating, and the three hot lookups use it: Phase A of `UpdateThemeReference`, `GetImplicitStyle`, and `ResourceResolver.TryVisualTreeRetrieval`. **The iterator remains, but delegates to the walker**, so the visit order — containing dictionary, then each ancestor's non-empty `Resources`, then the application resources — exists in exactly one place and the two `.ToArray()` callers are untouched.

Two details worth calling out:

- **The parent link is read after the consumer's body**, exactly as the iterator does. This matters: `ResourceDictionary.TryGetValue` can hit `TryMaterializeLazy`, which runs a generated `ResourceInitializer` under a pushed resolver scope — arbitrary XAML construction — from inside the loop body. Reading `fe.Parent` before yielding would have let a re-parent during materialization send the walk up a stale chain, resolving from the wrong ancestor scope. The failure mode there is a subtly wrong brush, not an exception.
- **Order is now pinned by a test.** `Given_Style.When_ImplicitStyle_Nearest_Scope_Wins` defines the same implicit style at application level and on a nearer ancestor and asserts the nearest wins. The pre-existing implicit-style tests define each key at a single level, so they would pass even if the walk were reordered or the application-resources tail dropped.

### Measured impact 📊

Managed bytes allocated per iteration on the UI thread (`GC.GetAllocatedBytesForCurrentThread`, median of 7 rounds, two runs per side; reproduces within ~1 %):

| Workload | Before | After | Δ |
|---|---:|---:|---:|
| Enter + Leave `TextBox` | 30,590 B | 22,152 B | −27.6 % |
| Enter + Leave `Button` | 18,592 B | 14,520 B | −21.9 % |
| Enter + Leave a 40-control page | 796,857 B | 650,481 B | −18.4 % |
| Enter + Leave `CheckBox` | 24,744 B | 20,488 B | −17.2 % |
| Enter + Leave `ListViewItem` | 37,828 B | 32,024 B | −15.3 % |
| Enter + Leave `ToggleSwitch` | 48,139 B | 41,292 B | −14.2 % |
| `ListView` (2,000 items) scroll step | 178,516 B | 157,212 B | −11.9 % |
| Theme toggle on that page | 540,152 B | 509,512 B | −5.7 % |

Over 1,000 scroll steps, so that the collection count resolves:

| Metric (1,000 scroll steps) | Before | After | Δ |
|---|---:|---:|---:|
| Allocated per step | 189,498 B | 168,583 B | −11.0 % |
| Gen0 collections, total | 14–15 | 13 | −1 to −2 |

Each build reproduced its own gen0 count across runs, but a later master build measured 14 rather than 15, so the gen0 delta sits at the edge of this instrument's resolution; the byte figures are the reliable ones. **The claim here is GC pressure, not frame time**: wall-clock on the measuring machine varied by more than the effect between identical runs, so it is not reported.

Measured with a headless Skia app (`Uno.UI.Runtime.Skia.Headless`) project-referencing the framework, with Fluent resources merged.

### Validation ✅

Runtime tests, Skia Desktop — `Given_Style`, `Given_ThemeResource`, `Given_ElementTheme`, `Given_ElementTheme_Resolution_Regression`, `Given_MergedAppResources_ThemeResource`, `Given_Theme_Materialization`, `Given_FrameworkElement_ThemeResources`, `Given_ResourceDictionary`, `Given_ResourceResolver`, `Given_VisualStateManager`, `Given_Storyboard`, `Given_MenuFlyout`, `Given_TextBlock`, `Given_Run`, `Given_Hyperlink`, `Given_Button`, `Given_CheckBox`, `Given_ToggleSwitch`, `Given_ListViewBase`, `Given_DependencyObjectCollection`: **517 run, 509 passed, 15 skipped, 1 inconclusive, 7 failed**.

The 7 failures are the `Given_ListViewBase` layout-rounding/timeout set plus `Given_TextBlock.When_Inlines_Transitively_Change`, all of which fail identically on a build of the merge-base without this change.

The new test was checked against two deliberately broken walkers. With the walk skipping ancestor dictionaries entirely it fails, along with the two pre-existing implicit-style tests. With the walk visiting the **application resources before the ancestors** — an order inversion, which is the regression class this rewrite could actually introduce — it is the **only** test in `Given_Style` that fails; `When_ImplicitStyle`, `When_Style_Flows_To_Popup` and `When_StyleFailsToApply` all still pass. It also passes on a build of the merge-base, so it asserts behaviour this PR preserves rather than introduces.

### Perf axis: work removed 📊

This change trades an allocation for a struct walk, so the table above **is** the performance evidence: the CPU benefit is the GC pressure removed, not fewer instructions along the walk. Stated exactly, and deterministically:

| Metric | Before | After |
|---|---:|---:|
| Heap allocations per `GetResourceDictionaries` / `EnumerateResourceDictionaries` call | 1 iterator state machine | **0** |
| Ancestor-walk semantics (visit order, parent-read timing) | unchanged | unchanged |

**CPU time is deliberately not reported, and that is demonstrable rather than asserted.** Re-measuring on this machine with a thread-cycle counter (`QueryThreadCycleTime` sampled on the UI thread, which is where the work runs) showed that **reversing the order of the two measurement phases flips the sign of the apparent result**, and that the spread *within* a single phase reaches 74–97 % on the same build and workload. Any timing column produced here would report machine state rather than the change. Wall-clock is worse and is not reported either. The allocation figures above are unaffected: bytes are deterministic and reproduce to ~10 B on the low-variance workloads.

**Provenance.** Measured with a headless Skia app (`Uno.UI.Runtime.Skia.Headless`) project-referencing the framework, with Fluent resources merged; bytes via `GC.GetAllocatedBytesForCurrentThread` on the dispatcher thread, median of 7 rounds, two runs per side. **Not BenchmarkDotNet** — this repository has no BenchmarkDotNet harness, and the workload runs on the dispatcher thread, which is where the allocation counter has to be read.
## PR Checklist ✅

### Follow-ups this PR deliberately leaves open

- `UpdateAllThemeReferences` still performs one full ancestor walk **per theme reference**; collapsing N walks into one is the larger win, but the surrounding code notes that updating one reference can cascade and modify the theme-reference map, and a newly resolved `Style` can change the ancestor scope mid-loop.
- The two `GetResourceDictionaries(...).ToArray()` callers in `UpdateResourceBindings` still allocate an array per theme update; they could rent from `ArrayPool`, as `UpdateAllThemeReferences` already does.
- `ResourceResolver.TryVisualTreeRetrieval` still allocates a LINQ enumerator from `CurrentScope.Sources.FirstOrDefault()` on this same path.
- `ApplyStyles` re-walks for an implicit style it usually already resolved; caching would remove the walks rather than cheapen them.
- An unresolved theme reference returns with no log at any level — the "my ThemeResource didn't apply" support case.

https://claude.ai/code/session_012QzKQawubFcPPX9JSoZLkX

---

**Internal tracking:** unoplatform/uno-private#2349 (finding 6), under unoplatform/uno-private#2346.


src/Uno.UI/UI/Xaml/DependencyObject.Store.cs
```diff
@@ -1566,43 +1566,112 @@ internal IEnumerable<ResourceDictionary> GetResourceDictionaries(
...
-			{
-				// In the case of StaticResource resolution we skip Application.Resources because we assume these were already checked at initialize-time.
-				yield return Application.Current.Resources;
+				_current = null;
+				return false;
 			}
+
+			public readonly ResourceDictionary Current => _current!;
+
+			public readonly ResourceDictionaryWalker GetEnumerator() => this;
```
morning4coffe-dev, line 1666: [P1] The self-enumerating reference-bearing struct is not performance-neutral on Android. Exact-head Android runtime shard 3 timed out twice, with several normally sub-second tests taking 60–87 seconds versus a same-base build completing the shard in 11.9 minutes. Avoid the `GetEnumerator() => this`/`foreach` shape at early-exit call sites; use an explicit walker with guaranteed reference cleanup or retain the iterator path on Android, then rerun the shard.

MartinZikmund: Thanks — this was a real finding, and I could corroborate it independently before fixing it. The ADO timeline for build 233998 records `Tests - Android Skia` at **attempt 3**, so that job failed twice and passed on the third try. The green check currently on the PR is that third-attempt pass, not evidence of health.

**Mechanism, confirmed in the code.** The walker carries four reference fields (`_containingDictionary`, `_candidate`, `_advanceFrom`, `_current`). On a *complete* walk they all end up null naturally — `_candidate` falls out of the loop, `_advanceFrom` is consumed, and `_current` is nulled before `MoveNext` returns false. But all three call sites stop at the first match (`GetImplicitStyle` returns from inside the loop body; the theme Phase A and `TryVisualTreeRetrieval` loops break on a hit), and because the struct did not implement `IDisposable`, `foreach` emitted **no cleanup call at all**. So on the common path the enumerator's stack slot kept the ancestor and dictionary it stopped on rooted past the loop — exactly the retention you describe, and exactly the kind of thing conservative stack scanning punishes during deep recursive theme propagation.

**Fix (ec53d2f1b0):** `ResourceDictionaryWalker` now implements `IDisposable` and nulls all four fields. For a struct enumerator implementing `IDisposable`, `foreach` emits a constrained call from a `finally`, so cleanup is guaranteed on every exit path and nothing is boxed.

I verified the emission rather than assuming it, using a throwaway that mirrors the exact shape (`GetEnumerator() => this`, reference-bearing struct):

```
--- early exit (break on first item) ---     DisposeCalls=1  heldRefAtDispose=True
--- early exit via return from a method ---  DisposeCalls=1  heldRefAtDispose=True
--- full enumeration ---                     DisposeCalls=1
```

`heldRefAtDispose=True` is the important column: the enumerator was **still holding a live reference at the moment of disposal**, so without `Dispose` those references genuinely outlived the loop. The `return`-from-body case is the one that matters most, since that is `GetImplicitStyle`.

**One place I followed the substance of your request but not its letter, so flagging it rather than leaving you to notice.** You suggested avoiding the `GetEnumerator() => this`/`foreach` shape at early-exit sites. I kept the shape and added guaranteed cleanup instead, because the proof above shows disposal is genuinely guaranteed on every exit path and it keeps the call sites readable. If you would rather the shape went too, I am happy to convert the three sites to an explicit `while (walker.MoveNext())` inside `try`/`finally` — say the word.

Two notes on the design space:

- It **cannot** become a `ref struct` (which would allow a plain `Dispose()` without the interface): `GetResourceDictionaries` still walks it with `foreach` from inside an iterator method, where ref structs are not allowed.
- The temp returned by `EnumerateResourceDictionaries(...)` also occupies a slot, but it only ever holds the *starting* element, which the caller already roots. The copy `foreach` iterates is the one that walks upward and accumulates ancestor references, and that is the copy being disposed.

**What I cannot claim:** that this resolves the Android slowdown. It is an intermittent GC-pressure symptom on a platform I cannot run here (Skia Desktop only), so I have not reproduced the 60–87 s tests locally and will not assert a fix I have not observed. Could you rerun the Android shard against this head when you get a chance? If it still times out, the fallback you suggested — retaining the iterator path on Android — is the next thing to try, and I would rather do that than keep guessing at the mechanism.

MartinZikmund:
Addressed the review feedback in `ec53d2f1b0` and `089481fa9d`. Summary of each item:

### Copilot review items

**1. Missing `GitHubWorkItem` metadata — fixed.** `When_ImplicitStyle_Nearest_Scope_Wins` now carries `[GitHubWorkItem("https://github.com/unoplatform/uno/issues/24522")]`, matching the `When_ImplicitStyle` test directly above it.

**2. `#if` around the platform skip — fixed.** Replaced with `[PlatformCondition(ConditionMode.Exclude, RuntimeTestPlatforms.NativeAndroid | RuntimeTestPlatforms.SkiaAndroid)]`. `RuntimeTestPlatforms` has no combined `Android` alias, so both flags together are what reproduces the previous `__ANDROID__` guard regardless of which renderer the leg uses.

Changing a platform gate can silently stop a test running rather than skip it visibly, so I checked that specifically rather than trusting a green count: on Skia Desktop the class reports 4/4 with `When_ImplicitStyle_Nearest_Scope_Wins` → **Passed** (not Skipped, not absent), both before and after the change.

**3. No Android coverage for the nearest-scope ordering — not addressed in code, deliberately.** The skip mirrors the pre-existing one on `When_ImplicitStyle` immediately above it: `ContentControl` implicit styles do not pass on Android CI, which predates this PR. Removing the skip would add a known-failing test rather than coverage. If Android coverage for the ordering is wanted, the right shape is a separate test using a target type whose implicit style does pass there — I would rather not guess which one blind, since I cannot run the Android leg locally. Happy to add it if someone can point at a type that works.

### Screenshot differences

The `skia-windows-screenshots: 2310` figure is not from this change. The same three counts appear on every PR from that window:

| PR | build | skia-linux | skia-windows | wasm |
|---|---|---:|---:|---:|
| #24523 (this one) | 233998 | 57 | 2310 | 11 |
| #24525 | 233999 | 57 | 2310 | 11 |
| #24529 | 234033 | 57 | 2310 | 11 |
| #24519 (merged) | 233976 | 57 | 2310 | 11 |

Identical on four independent PRs including one already merged, so it is a baseline artifact on the Windows leg rather than anything a PR did. The asymmetry points the same way: a resource-ordering regression would move both Skia legs alike, not 2310 of 2408 on Windows while Linux moves 57 of 2348.

### Android runtime shard

Covered in the inline thread, but noting it here since it is the one open item: the ADO timeline shows `Tests - Android Skia` completed on **attempt 3** for this build, so the currently green check is a third-attempt pass. `ec53d2f1b0` addresses the retention mechanism, but I have not been able to reproduce the Android symptom locally and am not claiming it fixed — a rerun of that shard against this head is the thing that would tell us.

src/Uno.UI/UI/Xaml/DependencyObject.Store.cs
morning4coffe-dev: This still reproduces at `089481fa9d9f`. Adding `IDisposable` did not restore Android runtime performance: exact-head shards 0, 2, and 3 reached the 45-minute timeout, while shards 1 and 4 completed in roughly 10 minutes; the prior successful shard-3 baseline completed in 11m52s. Please keep this thread open and use the proposed Android iterator fallback or an explicit non-`foreach` walker whose reference cleanup is effective under Android code generation, then rerun all Android shards.

PR #24541, opened 2026-09-17
refactor(composition): Align the render-scheduling gate with its sibling locks

### LINKED ISSUE #24540: perf(composition): Render-scheduling gate still uses a plain object monitor on the per-property-set path

`CompositionTarget`'s render-scheduling gate is still a plain `object` monitor, while the two sibling gates in the *same partial class* were already migrated to `System.Threading.Lock`:

- `src/Uno.UI/UI/Xaml/Media/CompositionTarget.RenderScheduling.cs:69` — `private readonly object _renderingStateGate = new();`
- `src/Uno.UI/UI/Xaml/Media/CompositionTarget.Rendering.cs:55-56` — `private readonly Lock _frameGate = new();` / `private readonly Lock _xamlRootBoundsGate = new();`

`using System.Threading;` is already present in the file, so this one simply appears to have been missed by that migration.

It matters because of how hot the gate is: `lock (_renderingStateGate)` is taken by `ICompositionTarget.RequestNewFrame()` (`:89`), which is called from `Visual.OnPropertyChangedCore` and `Compositor.InvalidateRenderPartial` on **every composition property set** — Offset, Size, Opacity, TransformMatrix, Clip, AnchorPoint, CenterPoint, Scale, Rotation — i.e. many times per Arrange pass, on every frame that touches layout.

There is a second, smaller issue in the same file: `EnqueueRenderCallback` takes the gate at `:127`, and then takes it **again** at `:147` inside the `else if (RenderRequested)` branch, purely to wrap a single `bool` assignment. That inner block is strictly nested inside the outer lock, so it is a same-thread recursive acquire that adds nothing.

### Why is this needed? 🤔

A CPU profile of the layout/render path attributes roughly **14.7% of managed samples to `Monitor.Enter_Slowpath`**, and about **21% of those trace through `RequestNewFrame`** — on the order of **3% of total managed CPU** spent on this one field's lock type. `System.Threading.Lock` exists precisely to avoid that path for uncontended, non-condition-variable gates.

This is a lock-type substitution with no semantic change, so the gain is CPU only — there is no allocation difference to measure.

### Renderer 🎨

- [x] Skia

### For which platforms 📱💻🖥️

All platforms — `CompositionTarget` is shared `Uno.UI` code, reached on every Skia head (Win32, macOS, X11, WASM, Skia-on-Android/iOS).

Two things that make the substitution safe here, both verified in source:

1. **Nothing the plain `object` gate supported is used.** `_renderingStateGate` is private and referenced only in `CompositionTarget.RenderScheduling.cs` — the declaration, three comments, and five `lock (…)` statements (`:89`, `:127`, `:185`, `:210`, `:221`). There is no `Monitor.Wait`/`Pulse`/`PulseAll` anywhere under `src/Uno.UI/UI/Xaml/Media`, and the field is never passed as an argument, converted to `object`, or used as a dictionary/`ConditionalWeakTable` key. (If condition-variable semantics are ever needed on this gate, it would have to go back to a plain object — `Lock` has no equivalent.)

2. **The gate is genuinely re-entered on a live path, and `Lock` supports that.** `EnqueueRenderCallback` calls `Render()` at `:149` while holding the gate; `Render()` (`CompositionTarget.Rendering.cs:114`) calls `RequestNewFrame()` at `:170` when `_isRenderingActive`, which re-enters `lock (_renderingStateGate)` at `:89` on the same thread. That recursion predates this change and is unaffected by it — but it is the reason the field must be a reentrant lock rather than something like a `SpinLock`.

Removing the redundant nested `lock` at `:147` is separately safe: the assignment stays inside the outer lock taken at `:127`, so `RenderRequested` remains protected and the locking scope around `Render()` is unchanged.

MartinZikmund:
**GitHub Issue:** closes #24540

## PR Type:

> **Re-scoped after measuring.** This was opened as a perf change on the strength of a profile attributing ~3% of managed CPU to `Monitor.Enter_Slowpath` via `RequestNewFrame`. I measured it, and **the evidence does not support a perf claim** — see below. It stands as a consistency/cleanup change, and reviewers should judge it on that basis.

## What changed? 🚀

`CompositionTarget`'s render-scheduling gate was the only one of three in the class still using a plain `object` monitor:

| field | file | type |
|---|---|---|
| `_renderingStateGate` | `CompositionTarget.RenderScheduling.cs:69` | `object` → **`Lock`** |
| `_frameGate` | `CompositionTarget.Rendering.cs:55` | already `Lock` |
| `_xamlRootBoundsGate` | `CompositionTarget.Rendering.cs:56` | already `Lock` |

`using System.Threading;` was already present, so this looks like an oversight in that migration rather than a deliberate exception.

It also removes a redundant nested acquire: `EnqueueRenderCallback` takes the gate, then took it **again** a few lines later to wrap a single `bool` assignment, inside the block that already holds it.

### 1. Exposure — how hot is this gate really?

Instrumented `RequestNewFrame` with an acquisition counter and drove repeated measure/arrange passes over a 340-element tree in a headless Skia harness (`RenderTargetBitmap` never reaches this path, so an arrange-driven workload is required):

| | |
|---|---:|
| acquisitions per arrange pass | **1,022** |
| cost of one arrange pass | **30.0 ms** (84.3M thread cycles) |

### 2. The primitive — what actually changed, benchmarked directly

`lock(object)` vs `lock(Lock)` on .NET 10.0.11, 50M iterations, best-of-7, with an empty-loop baseline (0.56 ns) subtracted. **Three runs:**

| case | run 1† | run 2 | run 3 |
|---|---:|---:|---:|
| `lock(object)` net | 28.49 | 21.66 | 17.61 |
| `lock(Lock)` net | 28.88 | 17.42 | 18.32 |
| **delta** | **−0.39** | **+4.24** | **−0.70** |

† run 1 had no baseline control and fewer reps; included for completeness.

**The same primitive measured 28.5, 21.7 and 17.6 ns across runs — a ~60% swing.** The uncontended difference between the two lock types is below this machine's noise floor. I cannot claim it.

Two effects *did* reproduce consistently across all three runs:

- **Recursive re-acquire is genuinely cheaper on `Lock`**: `object` 15.9–27.4 ns vs `Lock` 6.1–7.5 ns.
- **`Lock` was consistently *slower* under contention**: 2 threads hammering one gate, `Lock` worse by 4.9–11.7 ns/acquire. This is a tight-loop worst case and does not model this gate (UI-thread dominant), but it is a real signal and I am not going to omit it because it is inconvenient.

### 3. Putting them together

Taking the single most favourable delta measured (+4.24 ns) and the measured exposure:

**1,022 acquisitions × 4.24 ns = 4.3 µs per arrange pass = 0.014% of a 30 ms pass.**

And that is the optimistic bound; two of three runs put the delta at or below zero. The nested-lock removal saves one recursive acquire (~16–27 ns) per `EnqueueRenderCallback`, i.e. roughly once per frame — also negligible.

**Conclusion: this change has no measurable performance effect.** I could not reproduce the profile's ~3% attribution, and I am not restating it as a justification. One caveat in the other direction, for completeness: an `object` monitor that has ever been contended inflates permanently and stays on the slow path, while `Lock` has no such state — that may be what the profile's `Enter_Slowpath` samples represent, and a single-threaded headless harness cannot reproduce it.

## What the change is still worth

- **Consistency**: the three gates in this class now use one primitive, and the odd one out was on the hottest path.
- **A redundant recursive lock is gone** — real, if small, and it removes a confusing "why is this locked twice?" for the next reader.
- `Lock`'s cheaper recursive acquire is relevant here specifically because **this gate is genuinely re-entered** (below).

## Why it is safe

1. **Nothing the plain `object` gate supported is used.** The field is private and referenced only in `CompositionTarget.RenderScheduling.cs`. There is no `Monitor.Wait`/`Pulse`/`PulseAll` anywhere under `src/Uno.UI/UI/Xaml/Media`, and it is never passed as an argument, converted to `object`, or used as a dictionary/`ConditionalWeakTable` key. (If condition-variable semantics are ever needed here, it must revert to a plain object — `Lock` has no equivalent.)
2. **The gate is genuinely re-entered on a live path, and `Lock` supports that.** `EnqueueRenderCallback` calls `Render()` while holding it; `Render()` calls `RequestNewFrame()` when `_isRenderingActive`, which re-enters the same lock on the same thread. That recursion predates this PR. Removing the nested acquire does not affect it — the recursion comes from the `Render()` callout, not the inner block.

## Validation

- **Runtime tests** (Skia Desktop, `Uno.UI.RuntimeTests.Tests.Windows_UI_Composition`): **67/67 passed**, 0 failed, 0 skipped.
- **Strict build**: `Uno.UI` in Release *without* `UnoFastDevBuild` — **0 errors, 0 warnings**.
- All instrumentation used for the measurements above was reverted; the commit contains only the two changes described.

**Reviewer's call:** given there is no demonstrable win and a synthetic contended regression, it is entirely reasonable to reject this or to take only the nested-lock removal. I would rather surface that than have it found in review.

## PR Checklist ✅

- [ ] 🧪 Added [Runtime tests, UI tests, or a manual test sample](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) (for bug fixes / features, if applicable) — **no new tests**: behaviour-preserving lock-type substitution with nothing new to assert. Guarded by the existing composition suite above.
- [ ] 📚 Docs have been added/updated following the [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features) — not applicable.
- [ ] 🖼️ Validated PR `Screenshots Compare Test Run` results. — to be checked once CI publishes them.

https://claude.ai/code/session_012QzKQawubFcPPX9JSoZLkX


ajpinedam (approved):

morning4coffe-dev (approved):
LGTM! Switching to System.Threading.Lock aligns the render-scheduling gate with sibling locks and eliminates the redundant recursive lock in EnqueueRenderCallback.

PR #24551, opened 2026-09-18
fix(ItemsRepeater): Capture the items source in the OnLoaded revoker

### LINKED ISSUE #24550: ItemsRepeater: setting ItemsSource after an unload/reload throws NullReferenceException (OnLoaded revoker reads the field)
### Current behavior 🐛

After an `ItemsRepeater` has been unloaded and loaded again, setting `ItemsSource = null` (or to any other value) throws `NullReferenceException` from inside the setter.

Cause: the Uno-specific re-subscription in `ItemsRepeater.OnLoaded` stores a revoker lambda that reads the **field** `m_itemsSourceView` when it is disposed:

```csharp
if (_dataSourceSubscriptionsRevoker.Disposable is null && m_itemsSourceView is not null)
{
    m_itemsSourceView.CollectionChanged += OnItemsSourceViewChanged;
    _dataSourceSubscriptionsRevoker.Disposable = Disposable.Create(() =>
    {
        m_itemsSourceView.CollectionChanged -= OnItemsSourceViewChanged; // reads the field at dispose time
    });
}
```

`OnDataSourcePropertyChanged` assigns `m_itemsSourceView = newValue` **first** and only then disposes the previous revoker, so on an `ItemsSource = null` the lambda runs against a null field and throws. In the compiled 6.6.176 Skia runtime this is `ItemsRepeater.<OnLoaded>b__NN_0`, whose IL is `ldarg.0; ldfld m_itemsSourceView; ...; callvirt ItemsSourceView::remove_CollectionChanged` (same source at tag 6.6.184 and on `master`).

Because the throw happens before `-=` runs, the previous `ItemsSourceView` also stays subscribed to `OnItemsSourceViewChanged`.

The revoker that `OnDataSourcePropertyChanged` itself installs captures the local `newValue` instead, and does not have this problem.

**Visible symptom:** `NavigationView.UpdatePaneDisplayMode` re-hosts the pane (unloading/reloading the footer `ItemsRepeater`, which arms the lambda) and then calls `UpdateFooterRepeaterItemsSource`, whose first step clears the footer repeater's `ItemsSource`. The exception aborts that method before the footer is repopulated, so the NavigationView footer (e.g. Settings and any `FooterMenuItems`) renders empty after a pane display-mode change, e.g. when the window is resized across an adaptive breakpoint.

### Expected behavior 🎯

Changing `ItemsSource` after an unload/reload cycle works as it does before one, and the previous items source is unsubscribed.

```csharp
var repeater = new ItemsRepeater { ItemsSource = new ObservableCollection<string> { "a", "b" } };
var root = new Border { Child = repeater };
// put root in the window and wait for it to load
root.Child = new TextBlock();   // unload the repeater
root.Child = repeater;          // load it again
repeater.ItemsSource = null;    // NullReferenceException
```

### Workaround 🛠️

After a `NavigationView` pane display-mode change, set `FooterMenuItemsSource` to the control's own `FooterMenuItems` and back to `null`; that routes to `UpdateFooterRepeaterItemsSource` a second time without re-hosting the pane, and the second pass completes because the repeater's `ItemsSource` is already null.

### Renderer 🎨

- [X] Skia
- [ ] Native

### Affected platforms 📱💻🖥️

All platforms 🌍 (the code is under `#if HAS_UNO`; observed on Desktop (Windows) Skia)

Uno.Sdk 6.6.33 / Uno.WinUI 6.6.176; unchanged on `master`.

A PR with the fix and a runtime test follows.

SolidRockProgrammer:
**GitHub Issue:** closes #24550

## PR Type:

🐞 Bugfix

## What changed? 🚀

**Current behavior.** After an `ItemsRepeater` is unloaded and loaded again, `OnLoaded` re-subscribes to the items source and stores a revoker lambda that reads the **field** `m_itemsSourceView` when it is disposed. `OnDataSourcePropertyChanged` assigns `m_itemsSourceView = newValue` before it disposes the previous revoker, so:

- setting `ItemsSource = null` after an unload/reload throws `NullReferenceException` from the setter;
- setting it to another source unsubscribes the *new* view instead of the old one, leaving the old view subscribed.

In the compiled 6.6.176 Skia runtime this is `ItemsRepeater.<OnLoaded>b__NN_0`, whose IL is `ldarg.0; ldfld m_itemsSourceView; ldarg.0; ldftn OnItemsSourceViewChanged; newobj NotifyCollectionChangedEventHandler; callvirt ItemsSourceView::remove_CollectionChanged`. The source is unchanged at 6.6.184 and on `master`.

The revoker installed by `OnDataSourcePropertyChanged` itself captures the local `newValue` and does not have the problem. This PR makes `OnLoaded` do the same: capture the view in a local and use it in both the subscribe and the revoker.

**Where it shows up.** `NavigationView.UpdatePaneDisplayMode` re-hosts the pane (unloading and reloading the footer `ItemsRepeater`) and then calls `UpdateFooterRepeaterItemsSource`, whose first step clears the footer repeater's `ItemsSource`. The exception aborts that method before the footer is repopulated, so after a pane display-mode change (for example resizing across an adaptive breakpoint) the NavigationView footer — Settings and any `FooterMenuItems` — renders empty.

**Test.** Added `Given_ItemsRepeater.When_UnloadAndReload_Then_CanClearItemsSource`: load, unload, reload, set `ItemsSource = null` (throws before this change), then mutate the previous source.

**What I did not do:** I have not built Uno or run the runtime tests locally; I am relying on CI for both. The defect and the fix were established by reading the shipped runtime assembly and the `master` source.

## PR Checklist ✅

- [ ] 📚 Docs have been added/updated following the [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features) — not applicable


SolidRockProgrammer:
Hi Uno team — a gentle nudge on this one when a reviewer has a moment. It's a small fix (the `OnLoaded` revoker captures the local items source rather than reading the `m_itemsSourceView` field it is about to dispose), linked to #24550. Without it, a `NavigationView` whose pane display mode changes after an unload/reload throws a `NullReferenceException` out of `UpdateFooterRepeaterItemsSource` and leaves the footer items empty, and in our app it also occasionally precedes a fatal render-loop NRE. We're carrying an app-side workaround until a pinned Uno release includes the fix. Happy to adjust anything. Thanks!

PR #24554, opened 2026-09-18
fix!: Close out 7.0 constant and symbol leftovers

MartinZikmund:
**GitHub Issue:** closes unoplatform/uno-private#2327; part of unoplatform/uno-private#2199

## PR Type:

🐞 Bugfix

## What changed? 🚀

This PR fixes the 25 issues a verified review of 7.0's preprocessor constants and naming leftovers turned up. Most of them come from one pattern: repo-internal symbols (`__SKIA__`, `UNO_REFERENCE_API`, `__APPLE_UIKIT__`, `WINAPPSDK`) leaking into consumer-facing places, while consumer symbols varied with project shape. There is **one commit per issue**, with the fix and its test in separate commits. Each commit body explains that issue's root cause.

### Consumer symbols (#2327)

- `feat(sdk)!` **Consumer symbols now come from two places only.** `uno.winui.common.targets` defines the host symbols (`HAS_UNO`, `__UNO__` and the legacy synonyms `UNO_REFERENCE_API`, `HAS_UNO_SKIA`, `__UNO_SKIA__`) uniformly, plus `__APPLE_UIKIT__` on ios/tvos. The Uno.Sdk defines the TFM symbols. The 8 runtime packages define none.
  - Before this, a `net10.0-desktop` class library and a non-Sdk mobile library lacked `UNO_REFERENCE_API`, so `#if UNO_REFERENCE_API … #else` took the WinAppSDK branch there.
  - `__APPLE_UIKIT__` ships from `uno.winui.common.targets`, not the AppleUIKit runtime's `build/*.props` as #2327 suggested. That file isn't transitive and is absent without Uno.Sdk, so it would miss plain Microsoft.NET.Sdk libraries.
  - The consumer symbol table now lists only symbols every project shape gets, plus a "legacy synonyms" table. A new contributor page, `uno-development/preprocessor-symbols.md`, covers repo-internal symbols.
- `fix(analyzers)` + `fix` + `test(analyzers)` **UNO0001 works again.** Since the 2026-08 sync, all 3,395 Uno.UI/Composition stubs carry `[NotImplemented("__SKIA__")]`. That is a repo-internal symbol no consumer defines, so the analyzer never fired for any Uno.UI API. It also gave false positives on mobile.
  - The analyzer now reads the tokens per assembly. For single-build Uno.UI-family assemblies, `__SKIA__` means "not implemented on any Uno target". For the per-flavor WinRT assemblies, the consumer's TFM is mapped to a flavor.
  - Hand-written lists are normalized: stale attributes removed, `IS_UNIT_TESTS` tokens stripped.
  - The tests use the real 7.0 consumer symbol sets.
- `fix(winrt)` `ICorePointerInputSource` no longer carries a bare `[NotImplemented]` in the Reference and mobile WinRT builds.

### XAML generator

- `feat(sdk)` + `test` **UXAML0007 / UNOB0027.** Removed XAML prefixes (`skia:`, `netstdref:`, `not_mux:`, …) and removed file suffixes (`*.Apple.cs`, `*.reference.cs`, `*.iOSmacOS.cs`) used to change behaviour silently. Content was dropped on every target, applied on every target, or compiled for every TFM. Both now warn. `migrating-to-uno-7.md` describes the actual behaviour, and its `legacy:` row now says the `using:` alias is live.
- `fix(xaml)!` Drop the `legacy:` type-resolution fallback and the 4 SamplesApp samples that depended on it. The `Microsoft.X → Windows.X` retry is kept: an instrumented build showed it still rescues UWP→WinUI-migrated names like `Microsoft.UI.Text.FontWeight`.
- `refactor(xaml)` Remove the unsettable `UnoXamlLazyVisualStateManagerEnabled` read. Behaviour is unchanged.
- `refactor(wasm)!` Remove `TSBindingsGenerator`, along with the public `TSInteropMessageAttribute`/`CodeGeneration`, which has no remaining input.
- `fix(generator)` + `test` NativeCtorsGenerator emits `#if __IOS__ || __TVOS__`, skips non-partial declarations (which previously caused CS0260), and drops the Xamarin wording.
- `chore(generator)` Prune 6 dead `CompilerVisibleProperty` entries and the dead `EffectiveViewportTemplate` input. `UnoRemoteControlConfigCookie` stays, since it is a rebuild trigger.
- `fix(telemetry)` + `test` XAML telemetry classified every desktop/Android/iOS build as "Reference" because it matched `__SKIA__`/`__TIZEN__`. It now classifies from the target platform. The Tizen orphans are deleted.
- `fix(bindable)` + `test` Defining `UNO_DISABLE_KNOWN_MISSING_TYPES` in a Debug build with ≥1000 bindable types failed with CS0103 in generated code. The same block printed a literal `{type.FullName}`. Both are fixed.
- `chore(ios)!` Remove the Hot Restart helper generator. Hot Restart isn't supported in Visual Studio 2026 ([docs](https://learn.microsoft.com/dotnet/maui/ios/hot-restart)), and 7.0's net10+ requires VS 2026. The generated delegate also bypassed the Skia AppleUIKit host. Uno0004/Uno0005 and `UnoDisableHotRestartHelperGeneration` are retired.

### Uno.Sdk and build

- `fix(sdk)` The `Extensions` UnoFeature never took effect: its check was case-sensitive and ran after the SDK lowercases `UnoFeatures`. This had been the case since 6.7.
- `fix(sdk)!` Remove the dead `IsApple` properties (both the SDK one and the repo one). Migration note for the dropped `IsMacCatalyst`/`IsIOSOrCatalyst`: a bare-boolean condition on either is now MSB4100.
- `fix(sdk)` Docs now name the switches the SDK actually reads. The unreachable UNOB0011/UNOB0015 are retired.
- `fix(sdk)!` Remove the dead `UnoUniversalImageLoaderVersion`, wire `VlcNativeWindowsAssetsVersion`, and document the undocumented `*Version` knobs.
- `fix(ui)!` Trimming feature switches:
  - `UnoDragDropExternalSupport` is removed; it was a no-op.
  - The dynamic-template switch now gets its Release substitution on Skia, which the docs already promised.
  - `UnoEnableAlcAppSupport` is now emitted as a feature switch.
- `fix(desktop)` `UNO_WEBVIEW2_BACKEND` now only warns (one backend is left). WebView.md is fixed.
- `fix(analyzers)` + `test(analyzers)` UNO0007 no longer says ProgressRing needs Lottie. UNOX0001 is retired.
- `fix(build)` The NuGet-cache override (`UnoNugetOverrideVersion`) wrote to `uno.ui.devserver` and to `android30.0` for net11, so it silently did nothing there.

### Public API no-ops

- `refactor(api)!` Remove 6 public FeatureConfiguration/WinRTFeatureConfiguration flags and `PagePool.IsPoolingEnabled`, all silent no-ops. The ProgressRing asset URIs became no-ops on the 7.0 line itself; use `DeterminateSource`/`IndeterminateSource` instead.
- `fix(api)!` `PrettyPrint`/`ViewExtensions` were accidentally public in the WinAppSDK `Uno.UI.Extras.dll` (their guard was always false there). They are now internal on WinAppSDK.
- `refactor(msal)` Delete the dead `#if __WASM__` arms. MSAL.md and the migration guide now say `WithUnoHelpers()` does nothing on WebAssembly.

### DevServer and docs

- `fix(devserver)` + `test(devserver)` The DevServer client's Android-emulator / iOS-simulator / WASM host candidates were compile-time branches that can never be defined in this plain-net10.0 project. They are now runtime checks.
- `docs` ×3: snippets guarded with symbols no build defines (`__WINDOWS__`, `WINUI`, `HAS_UNO_SKIA`, `WINDOWS_UWP`, `NETFX_CORE`); `bundle://` and `DefaultImageLoader`, which were removed; `__MACCATALYST__ → __DESKTOP__` advice missing an `OperatingSystem.IsMacOS()` check.

### Runtime-test coverage on the WinUI leg

- `test` 33 test methods were skipped on native WinUI by Uno-only symbols (`!__SKIA__`, `__CROSSRUNTIME__`-derived locals, `UNO_HAS_MANAGED_SCROLL_PRESENTER`). They now run there and pass. 24 that genuinely differ on WinUI get `PlatformCondition(Exclude, NativeWinUI)`, each with the difference observed.
- `test` Port the MUX `DateTimePickerHelper`/`LoopingSelectorHelper` pieces and re-enable the parked DatePicker and ProgressRing tests. One test, `CanOpenAndCloseUsingKeyboard`, stays ignored with a current reason (a real Skia bug, recorded).

### Please review these decisions

1. `HAS_UNO_SKIA`/`__UNO_SKIA__` are kept as deprecated synonyms of `HAS_UNO`, planned for removal in 8.0 alongside the drawing-backend vocabulary. They are not removed now.
2. With UNO0001 working again, consumers see it for Uno.UI stubs. That affects `TreatWarningsAsErrors` builds (see the migration note). A few Generated stubs are really implemented, e.g. `MicaBackdrop.Kind`. Fixing those needs the sync generator (tracked separately).
3. A *trimmed* host app that loads secondary apps into ALCs must now set `UnoEnableAlcAppSupport=true`. This matches the ALC spec's documented default of `false`, which is trimmed.
4. `PrettyPrint`/`ViewExtensions` become internal on WinAppSDK. An earlier attempt to narrow them, in #16161, was reverted without a stated reason.
5. The Hot Restart generator is removed rather than rewritten.

## PR Checklist ✅

### Breaking changes

This is 7.0. Every removal has a row in `doc/articles/migrating-to-uno-7.md` and, where it is public API, a `PackageDiffIgnore` entry. The consumer-visible changes:

- `UNO_REFERENCE_API`/`HAS_UNO_SKIA`/`__UNO_SKIA__` are now defined for every non-WinAppSDK project shape.
- `__APPLE_UIKIT__` is now defined on ios/tvos.
- A plain-`net10.0` Headless project no longer gets `__DESKTOP__`.
- UNO0001 fires again.
- UXAML0007 and UNOB0027 are new warnings.
- Removed public API: the 6 FeatureConfiguration flags, `PagePool.IsPoolingEnabled` and `TSInteropMessageAttribute`.
- Removed properties: `UnoDragDropExternalSupport`, `UnoUniversalImageLoaderVersion`, `UnoUseWebView2WPF`, `IsApple` and `UnoDisableHotRestartHelperGeneration`.
- `PrettyPrint`/`ViewExtensions` are no longer public on WinAppSDK.

### Validation

- **Compile (strict, no `UnoFastDevBuild`, net10.0):**
  - 0 errors: Uno.UI, Composition, Uno.WinRT (Reference/Skia/Wasm/netcoremobile-android), Extras (Skia + Windows), MSAL, the Win32 runtime, RemoteControl, Uno.Sdk, `Uno.UI.RuntimeTests.Skia`, `Uno.UI.RuntimeTests.Windows` and SamplesApp `net10.0-desktop`.
  - All 104 changed `.cs` files pass `dotnet format whitespace --verify-no-changes`.
- **Unit tests:** Uno.Analyzers.Tests 118/118. Uno.UI.SourceGenerators.Tests 529 passed, 14 skipped, 0 failed (baseline 493; 50 new). DevServer client tests 40/40.
- **Runtime:**
  - Skia desktop: 116 passed, 0 failed, 7 skipped (ignored tests) across every touched runtime-test class.
  - Native WinUI (WinAppSDK MSIX): all re-enabled tests pass or are excluded with the WinUI difference observed. One late exclusion (`Given_Image`, a 1px difference at 250% scale) is compile-checked only.
- **Evaluation:** consumer define sets were checked with MSBuild evaluation of the real package build files across 12 project shapes: non-Sdk android/ios/tvos libraries, desktop library and exe, wasm, plain net10.0 (with and without Headless), and WinAppSDK.
  - tvOS was simulated, since there is no tvOS workload on the validating machine.
  - No packed-NuGet end-to-end run.
- **Not run:** a trimmed publish (for the linker substitutions) and a DocFX build.

### Found along the way (not fixed here)

These were recorded for follow-up:

- XAML generator telemetry has never been sent since 93e435fb58: the client is only created when the user opts *out*.
- `uno-internals-wasm.md` still documents the removed WASM DOM renderer.
- The `MicaBackdrop.Kind` Generated stub is actually implemented.
- 7 Uno-vs-WinUI differences surfaced by the re-enabled tests: ScrollViewer `ZoomMode=Disabled` resets `ZoomFactor`; composition `Translation` presence; `PropertySet.TryGet*` returns animated values; `StartAnimation` accepts calls WinUI rejects; LoopingSelector touch pan is ignored after the flyout reopens; the MUX input helpers are broken on WinAppSDK; `Path` measures non-zero outside the live tree.

https://claude.ai/code/session_01XKeq31czCtZUwXSBUq43ji


ajpinedam (approved):

src/Uno.UI.RuntimeTests/IntegrationTests/common/TestServices.InputHelper.cs
```diff
@@ -119,6 +119,26 @@ public static void Tap(Point point)
...
+			{
+				EnsureInputInjectorSupported();
+				MUXControlsTestApp.Utilities.RunOnUIThread.Execute(() =>
+				{
+					using var finger = InputInjector.TryCreate()?.GetFinger() ?? throw new InvalidOperationException("Failed to create finger");
+					var topLeft = element.TransformToVisual(WindowHelper.XamlRoot.Content).TransformPoint(new Point(0, 0));
+					var center = new Point(topLeft.X + element.RenderSize.Width / 2, topLeft.Y + element.RenderSize.Height / 2);
+
+					const uint steps = 10;
+					var durationInMilliseconds = Math.Sqrt(relX * relX + relY * relY) / velocityFactor;
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in 46d2cdb4c62 — cast both operands to `double` before squaring.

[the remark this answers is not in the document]
MartinZikmund: Fixed in 46d2cdb4c62 — cast both operands to `double` before squaring (same fix as the duplicate thread on this line).

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlCodeGeneration.Telemetry.cs
```diff
@@ -170,27 +172,59 @@ private string BuildIDEName()
...
+		/// Classifies the build by target platform, from the target framework or, when it names none, from the
+		/// platform symbols. A plain netX.0 project is reported as "Plain".
+		/// </summary>
+		internal static string GetUnoRuntime(string targetFramework, string defineConstants)
 		{
-			var constants = _generatorContext.GetMSBuildPropertyValue("DefineConstantsProperty");
+			var separatorIndex = targetFramework?.IndexOf('-') ?? -1;
+			if (separatorIndex >= 0)
+			{
+				var platform = targetFramework.Substring(separatorIndex + 1).TrimEnd('.', '0', '1', '2', '3', '4', '5', '6', '7', '8', '9');
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in 46d2cdb4c62 — added the explicit `targetFramework is not null` check alongside `separatorIndex >= 0` so the dereference is provably safe to analyzers (behavior was already correct at runtime, since the null-conditional/coalescing combo makes separatorIndex -1 when targetFramework is null).

src/SourceGenerators/Uno.UI.SourceGenerators.Tests/BindableTypeProvidersTests/Given_BindableTypeProvidersSourceGenerator.cs
```diff
@@ -0,0 +1,98 @@
...
+		var builder = new StringBuilder("namespace MyProject;\r\n");
+		for (var i = 0; i < count; i++)
+		{
+			builder.Append($"[Microsoft.UI.Xaml.Data.Bindable] public class BindableType{i:0000} {{ public int Value {{ get; set; }} }}\r\n");
+		}
+
+		return builder.ToString();
+	}
+
+	private sealed class Test(string[] preprocessorSymbols, bool isDictionaryMode) : CSharpSourceGeneratorVerifier<BindableTypeProvidersSourceGenerator>.Test
```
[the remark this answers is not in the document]
MartinZikmund: No action — `Test` here derives from `CSharpSourceGeneratorVerifier<T>.Test`, matching the naming convention the testing library and its other usages in this test project already use (see `Verifiers/CSGenerator.cs`). Renaming would break consistency with the sibling verifier classes.

src/Uno.UI.RemoteControl/RemoteControlClient.cs
```diff
@@ -1293,43 +1294,82 @@ internal static bool IsPreferredEndpointMatch(string? preferred, string candidat
...
+				}
+			}
+		}
+		catch (IOException)
+		{
+			// Unreadable is treated as "not an emulator" — this is a best-effort heuristic, not a hard requirement.
+		}
+		catch (UnauthorizedAccessException)
+		{
+		}
```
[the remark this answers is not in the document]
MartinZikmund: Fixed in 46d2cdb4c62 — merged into a single `catch (Exception e) when (e is IOException or UnauthorizedAccessException)`, keeping the existing explanatory comment.

doc/articles/uno-development/preprocessor-symbols.md
```diff
@@ -0,0 +1,73 @@
...
+
+| Symbol | Defined for | Consumer equivalent |
+| ------ | ----------- | ------------------- |
+| `__SKIA__` | `UnoRuntimeIdentifier=Skia`: `Uno.UI` and everything built on it, and the desktop flavor of the WinRT layer | In `Uno.UI` code it is always true, so it is `HAS_UNO`. In the WinRT layer it names the desktop flavor: `__DESKTOP__` |
+| `__WASM__` | The WebAssembly flavor of the WinRT layer | `__WASM__` |
+| `__NETSTD_REFERENCE__` | The Reference flavor of the WinRT layer, packed as the `lib/netX.0` assembly that plain `netX.0` libraries compile against | None |
+| `__CROSSRUNTIME__` | The Skia, WebAssembly and Reference flavors, not the Android, iOS and tvOS variants of the WinRT layer | `HAS_UNO` |
+| `UNO_REFERENCE_API` | Same as `__CROSSRUNTIME__` | `HAS_UNO`. Consumers get `UNO_REFERENCE_API` as a legacy synonym of `HAS_UNO`, which is a different condition |
+| `__ANDROID__`, `__IOS__`, `__TVOS__` | The Android, iOS and tvOS variants | Same symbols |
+| `__APPLE_UIKIT__` | The iOS and tvOS variants | `__APPLE_UIKIT__` |
```
Xiaoy312: for the WinRT project(s) only

Xiaoy312: we can also chain a `*` or dagger-symbol to refer to the "OperatingSystem.IsXyz" line below

src/SourceGenerators/SourceGeneratorHelpers/Helpers/PlatformHelper.cs
```diff
@@ -22,10 +22,6 @@ public static bool IsValidPlatform(GeneratorExecutionContext context)
 		public static bool IsAndroid(GeneratorExecutionContext context)
 			=> context.GetMSBuildPropertyValue("AndroidApplication")?.Equals("true", StringComparison.OrdinalIgnoreCase) ?? false;
 
-		public static bool IsIOS(GeneratorExecutionContext context)
```
Xiaoy312, line 25: this seems like an odd one to remove? especially the sibling IsAndroid is still there

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlCodeGeneration.Telemetry.cs
```diff
@@ -170,27 +172,59 @@ private string BuildIDEName()
...
+		/// </summary>
+		internal static string GetUnoRuntime(string targetFramework, string defineConstants)
 		{
-			var constants = _generatorContext.GetMSBuildPropertyValue("DefineConstantsProperty");
+			var separatorIndex = targetFramework?.IndexOf('-') ?? -1;
+			if (separatorIndex >= 0 && targetFramework is not null)
+			{
+				var platform = targetFramework.Substring(separatorIndex + 1).TrimEnd('.', '0', '1', '2', '3', '4', '5', '6', '7', '8', '9');
+
+				return platform.ToLowerInvariant() switch
```
Xiaoy312:
```suggestion
			if (targetFramework?.Split('-', 2) is [_, var platform])
			{
				// trim trailing version, eg: android12.3
				var identifier = platform.TrimEnd('.', '0', '1', '2', '3', '4', '5', '6', '7', '8', '9');

				return identifier.ToLowerInvariant() switch
```

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlNamespaceValidation.cs
```diff
@@ -7,19 +7,76 @@
...
+		["not_netstdref"] = UseInstead("winappsdk"),
+		["androidskia"] = UseInstead("android"),
+		["not_androidskia"] = UseInstead("not_android"),
+		["iosskia"] = UseInstead("ios"),
+		["not_iosskia"] = UseInstead("not_ios"),
+		["tvosskia"] = UseInstead("tvos"),
+		["not_tvosskia"] = UseInstead("not_tvos"),
+		["wasmskia"] = UseInstead("wasm"),
+		["not_wasmskia"] = UseInstead("not_wasm"),
+		["macos"] = UseInstead("desktop"),
```
Xiaoy312: this is nto exactly a 1:1 replacement

src/Uno.Sdk/targets/Uno.CrossTargeting.targets
Xiaoy312, line 1: instead of "Removed" we should refer them as "Retired"
removed has other confusing meanings

Xiaoy312 (approved):

doc/articles/uno-development/preprocessor-symbols.md
MartinZikmund: Fixed — clarified the row to say "of the WinRT layer", matching the sibling __WASM__/__NETSTD_REFERENCE__ rows. In fe9a8b384c8.

src/SourceGenerators/SourceGeneratorHelpers/Helpers/PlatformHelper.cs
MartinZikmund: Confirmed dead code — grepped src/SourceGenerators for PlatformHelper.IsIOS and there are zero callers left (the sibling IsAndroid is kept because it does have two live call sites, in AndroidResourcesGenerator and XamlCodeGeneration). No behavior change.

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlCodeGeneration.Telemetry.cs
MartinZikmund: Applied — though the literal suggestion didn't compile as-is: this project targets netstandard2.0, which has neither the single-char Split(char, int) overload nor System.Index (needed for the [_, var platform] list pattern) without a polyfill. Adapted to the same Split-based decomposition using the char[] overload and a Length check instead of the list pattern. In fe9a8b384c8.

src/SourceGenerators/Uno.UI.SourceGenerators/XamlGenerator/XamlNamespaceValidation.cs
MartinZikmund: Good catch — you're right, and it's worse than non-1:1: 'macos'/'not_macos' named the native Xamarin.Mac head, whose MSBuild plumbing was deleted in 2023 (172aa347aa1). Since then 'macos:' has been dead on every target, and 'not_macos:' has applied on every target — exactly like not_mux/xamarin/legacy, not like a renamed symbol. Pointing 'macos' at 'desktop' would resurrect markup that's compiled nowhere for two years, and pointing 'not_macos' at 'not_desktop' would newly exclude markup from desktop where it always applied. Changed both to 'remove/drop' advice (matching not_mux's treatment) instead of a false desktop/not_desktop equivalence, and updated the two doc tables and the test file's expected messages to match. In fe9a8b384c8.

src/Uno.Sdk/targets/Uno.CrossTargeting.targets
MartinZikmund: Renamed the target and its item to _UnoWarnRetiredFileSuffixes / _UnoRetiredSuffixFile — both are private (leading underscore), self-contained to this file, with no external references, so it was a safe rename. In fe9a8b384c8.

doc/articles/migrating-to-uno-7.md
```diff
@@ -101,8 +106,11 @@ longer tested by Uno Platform.
...
 | `Uno.UI.FluentTheme.v2` assembly merged into `Uno.UI.FluentTheme` | With V1 gone there is a single set of Fluent styles, so the two assemblies were collapsed into one. All the styles ship in `Uno.UI.FluentTheme`; both assemblies come from the `Uno.WinUI` package, so no reference changes. Only a direct reference to the `Uno.UI.FluentTheme.v2` assembly, or to `XamlControlsResourcesV2`, needs updating. |
-| `Uno.UniversalImageLoader` no longer injected (Android) | Skia handles image loading internally. If you initialized it manually, remove the `ConfigureUniversalImageLoader();` call. |
+| `Uno.UniversalImageLoader` no longer injected (Android) | Skia handles image loading internally. If you initialized it manually, remove the `ConfigureUniversalImageLoader();` call. The `UnoUniversalImageLoaderVersion` property is removed and is silently ignored if still set. |
```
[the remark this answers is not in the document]
MartinZikmund: Added — folded into the existing WebView2 backend row rather than a new one, since it's the same topic. Points at UnoDisableWebView2Workarounds (with the value spelled out) and notes it also turns off the package-alias workaround, not just the WPF reference. In 5d237f0a684.

build/nuget/uno.winui.common.targets
```diff
@@ -10,19 +10,18 @@
 
 	<!-- Feature configuration -->
 	<PropertyGroup>
-		<UnoDragDropExternalSupport Condition="'$(UnoDragDropExternalSupport)'==''">true</UnoDragDropExternalSupport>
 		<UnoEnableDynamicDataTemplateUpdate Condition="'$(UnoEnableDynamicDataTemplateUpdate)'=='' AND '$(Optimize)' == 'true'">false</UnoEnableDynamicDataTemplateUpdate>
 		<UnoEnableDynamicDataTemplateUpdate Condition="'$(UnoEnableDynamicDataTemplateUpdate)'=='' AND '$(Optimize)' != 'true'">true</UnoEnableDynamicDataTemplateUpdate>
+		<UnoEnableAlcAppSupport Condition="'$(UnoEnableAlcAppSupport)'==''">false</UnoEnableAlcAppSupport>
```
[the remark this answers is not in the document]
MartinZikmund: Added a row — but caught (via review) that my first draft mischaracterized the effect: it's not about ALC loading or collectibility, it's that trimmed builds stub Application.HasSecondaryApps to false (LinkerSubstitution.Skia.xml), which drops ALC-aware resource resolution for secondary Application instances (Application.cs, ResourceDictionary.cs, DependencyObject.alc.cs). Also corrected "now defaults to false" - it's a brand-new MSBuild property in this PR, not a flipped default (the pre-existing CompilerVisibleProperty the XAML generator reads already treated unset as false). In 5d237f0a684.

doc/articles/interop/MSAL.md
```diff
@@ -61,6 +61,28 @@ Follow [Considerations for using Xamarin iOS with MSAL.NET](https://learn.micros
...
+    .Build();
+
+var authResult = await _app.AcquireTokenInteractive(scopes)
+    .WithCustomWebUi(myCustomWebUi) // required on WebAssembly
+    .ExecuteAsync();
+```
+
+`myCustomWebUi` implements `Microsoft.Identity.Client.Extensibility.ICustomWebUi` and drives the
+authentication flow through the browser — for example with a popup window monitored through
+`System.Runtime.InteropServices.JavaScript` JSImport/JSExport interop, or an in-app `WebView2`.
```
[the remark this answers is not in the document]
MartinZikmund: Applied verbatim - dropped the WebView2 mention (Windows-only, doesn't apply to WASM) and backtick-wrapped JSImport/JSExport. In 5d237f0a684.

MartinZikmund:
Re Copilot's suppressed note on `src/Uno.CrossTargetting.targets:40` (review 5247993998): fair catch, consumers do get `UNO_REFERENCE_API` (legacy synonym of `HAS_UNO`) and `__WASM__` from the SDK. Reworded in a29c33648e1.

